PR #3545, opened 2021-12-17
Add pagination and summary to Disputes table

mgascam:
Fixes #3386 

#### Changes proposed in this Pull Request

Adds summary information and pagination controls to the Disputes table. 

Screen Shot 2021-12-26 at 19 20 18

#### Testing instructions

1. Add a number of orders with disputes greater than the rows per page defined for the disputes table. This number is set to 25. Alternatively, you can use the `per_page` query parameter to set a smaller number i.e. `http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes&paged=1&per_page=10`
2. You can use [Stripe's test cards to pay for the orders](https://stripe.com/docs/testing#disputes). Using one of this cards will generate a dispute after the order is paid.
3. Note that this PR relies on the pagination functionality developed [here](1415-gh-automattic/woocommerce-payments-server)
4. Follow instructions from 1408-gh-Automattic/woocommerce-payments-server to have disputes saved in the cache
5. Observe that the pagination controls appear at the bottom of the table.
6. Check that the pagination controls work as expected (navigate to next/ prev page, go to page number, select rows per page)
7. Observe that the total number of Disputes is displayed at the bottom of the table

-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Added testing instructions to the [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) (or does not apply)

PR #3654, opened 2022-01-18
Bump plugin version to see if that fixes breaking Disputes E2Es

anu-rock:
🧪 **EXPERIMENT-ONLY: NOT INTENDED TO BE MERGED**

The sole purpose of this PR is to see if bumping up the plugin version to 3.6.0 fixes E2Es that are purported to be [breaking](https://github.com/Automattic/woocommerce-payments/actions/runs/1710384484) after the merging of #3545.

### DIFF
--- a/woocommerce-payments.php
+++ b/woocommerce-payments.php
@@ -12,7 +12,7 @@
  * WC tested up to: 6.1
  * Requires at least: 5.6
  * Requires PHP: 7.0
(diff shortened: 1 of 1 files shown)


anu-rock:
Confirmed that bumping the version will fix the breaking disputes tests:  
https://github.com/Automattic/woocommerce-payments/actions/runs/1711563910

PR #3662, opened 2022-01-18
Platform Checkout - move email field to top of checkout and handle input

mdmoore:

To move the email field to a new section above billing details and handle input, adding the spinner/loading icon.
#### Testing instructions

**Post merge**


includes/class-wc-payments.php
```diff
@@ -778,6 +778,8 @@ public static function maybe_register_platform_checkout_hooks() {
 			add_filter( 'determine_current_user', [ __CLASS__, 'determine_current_user_for_platform_checkout' ] );
 			// Disable nonce checks for API calls. TODO This should be changed.
 			add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+			add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'platform_checkout_remove_default_email_field' ] );
+			add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'platform_checkout_fields_before_billing_details' ], 20 );
```
superdav42, line 782:
```suggestion
			if ( WC_Payments_Features::is_platform_checkout_enabled() ) {
				add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'platform_checkout_remove_default_email_field' ] );
				add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'platform_checkout_fields_before_billing_details' ], 20 );
			}
```
We should make sure are changes are hidden behind feature flag.

client/checkout/utils/platform-checkout.js
```diff
@@ -0,0 +1,28 @@
...
+
+		// Placeholder to simulate request. Replace with request to Platform Checkout email verification endpoint.
+		setTimeout( () => {
+			platformCheckoutEmailContainer.classList.remove( 'is-loading' );
+		}, 3000 );
+	};
+
+	const platformCheckoutEmailInput = document.querySelector( field );
+
+	platformCheckoutEmailInput.addEventListener( 'keyup', ( e ) => {
```
superdav42, line 18: I think the event we want to listen to is the `input` event. keyup will not be triggered on mobile or if inserting content with autocomplete prompts and other ways. `input` should always work. 
```suggestion
	platformCheckoutEmailInput.addEventListener( 'input', ( e ) => {
```

includes/class-wc-payments.php
superdav42: My bad. I didn't see the context.

client/checkout/utils/platform-checkout.js
mdmoore: 👍 Thanks for catching this! Updated.

includes/class-wc-payments.php
```diff
@@ -841,4 +843,44 @@ function( string $cookie_hash ) {
...
 	}
+
+	/**
+	 * Remove default billing email field for Platform Checkout
+	 *
+	 * @param array $fields WooCommerce checkout fields.
+	 * @return array WooCommerce checkout fields.
+	 */
+	public static function platform_checkout_remove_default_email_field( $fields ) {
+		unset( $fields['billing']['billing_email'] );
```
superdav42, line 854: After the discussion on WooPay Sync meeting they raised the question if moving this field will break some theme/plugins. I was thinking we could add a check here to see if `$fields` has changed from it's default and then not doing anything with the email.

superdav42, line 854: Something like this should do it.
```suggestion
		$wc_checkout = WC_Checkout::instance();
		$default_fields = (function () {
			return $this->fields;
		})->bindTo( $wc_checkout, WC_Checkout )();
		if ( $default_fields === $fields ) {
			unset( $fields['billing']['billing_email'] );
		} else {
			remove_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'platform_checkout_fields_before_billing_details' ], 20 );
		}
```

includes/class-wc-payments.php
```diff
@@ -778,6 +778,8 @@ public static function maybe_register_platform_checkout_hooks() {
 			add_filter( 'determine_current_user', [ __CLASS__, 'determine_current_user_for_platform_checkout' ] );
 			// Disable nonce checks for API calls. TODO This should be changed.
 			add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+			add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'platform_checkout_remove_default_email_field' ] );
```
superdav42, line 781: Maybe set a priority so it will happen after any other extensions modifying the email field.
```suggestion
			add_filter( 'woocommerce_checkout_fields', [ __CLASS__, 'platform_checkout_remove_default_email_field' ], 50 );
```

superdav42 (approved):
Style changes look good. Should be much better than the relative width. Thanks for looking at all the potential problems for such a small change. 👍

PR #3673, opened 2022-01-19
Add redirect to view Capital financing offers

### LINKED ISSUE #3655: View Capital financing offer redirect
### Description

Merchants should have access to a wp-admin page that automatically redirects them to the loan offer, after validating that they have the appropriate permissions.

### Acceptance criteria

* A new intermediate page should be created to redirect merchants to the loan offer page
* Only authenticated users should have access to it

### Designs

This page has no design since it's just a redirect. The designs for this flow can be checked in paJDYF-2Yf-p2.

luizreis:
Fixes #3655

Merchants should have a way to view a Capital loan offer from the link in the offer email. This PR adds a handler when an admin request is made with the wcpay-loan-offer GET parameter.

The redirect will only happen if the request is not AJAX, the user has admin permissions and the GET parameter is specified. If we're not able to get an account link for a merchant, they will be redirected to the Overview page and an error message will be displayed.

Note: We don't check for a nonce here because it's an automatic redirect. Navigating to the loan offer page is also not going to accept the loan or take any actions for the customer. The main risk we have here regarding CSRF or clickjacking attacks is having merchants be redirected to Stripe's dashboard or the WC Pay Overview page.

#### Testing instructions

*All tests must be run on the shared dev account, otherwise Capital won't be enabled.*

**Unauthorized user**

1. Navigate to `/wp-admin/admin.php?wcpay-loan-offer` as a non-admin user or guest
2. Assert that the login screen is shown and you're not redirected to the loan offer details

**Failed redirect**

1. Before adding a financing offer to a merchant account, navigate to `/wp-admin/admin.php?wcpay-loan-offer`
2. Assert that you have been redirected to the Overview page
3. Assert that there's an error notice on top informing that there was a problem redirecting you to the loan offer

**Successful redirect**

1. Add a financing offer to the merchant account
2. Navigate to `/wp-admin/admin.php?wcpay-loan-offer`
3. Assert that you have been redirected to the loan offer page
4. Copy the offer page link (to be used in the next test)

**Expired account link**

1. Navigate to the same link from the last test using a incognito tab or another browser
2. Assert that you're redirected to `/wp-admin/admin.php?wcpay-loan-offer` to be reauthenticated
3. Log in to the WP Admin, if not already
4. Assert that you have been redirected to the loan offer page

-------------------

**Post merge**


includes/class-wc-payments-account.php
```diff
@@ -330,6 +331,46 @@ public function get_fraud_services_config() {
...
+	 * Only admins are be able to perform this action. The redirect doesn't happen if the request is an AJAX request.
+	 * This method will end execution after the redirect if the user requests and is allowed to view the loan offer.
+	 */
+	public function maybe_redirect_to_capital_offer() {
+		if ( wp_doing_ajax() ) {
+			return;
+		}
+
+		// Safety check to prevent non-admin users to be redirected to the view offer page.
+		if ( ! current_user_can( 'manage_options' ) ) {
```
luizreis, line 346: Should we use `manage_woocommerce` here instead?

tpaksu: Is there a chance that one user is a WP-admin, but not a WC-admin?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1009,6 +1009,31 @@ public function get_login_data( $redirect_url ) {
...
+	 *
+	 * @param string $type        The type of link to be requested.
+	 * @param string $return_url  URL to navigate back to from the dashboard.
+	 * @param string $refresh_url URL to navigate to if the link expired, has been previously-visited, or is otherwise invalid.
+	 *
+	 * @return array Account link object with create, expires_at, and url fields.
+	 *
+	 * @throws API_Exception When something goes wrong with the request, or there aren't valid loan offers for the merchant.
+	 */
+	public function get_capital_link( $type, $return_url, $refresh_url ) {
```
tpaksu, line 1023: I think this was already created by Marcin here: 1382-gh-Automattic/woocommerce-payments-server

includes/class-wc-payments-account.php
```diff
@@ -330,6 +331,46 @@ public function get_fraud_services_config() {
...
+
+		$return_url  = $this->get_overview_page_url();
+		$refresh_url = add_query_arg( [ 'wcpay-loan-offer' => '' ], admin_url( 'admin.php' ) );
+
+		try {
+			$capital_link = $this->payments_api_client->get_capital_link( 'capital_financing_offer', $return_url, $refresh_url );
+
+			$this->redirect_to( $capital_link['url'] );
+		} catch ( API_Exception $e ) {
+			$error_url = add_query_arg(
```
tpaksu, line 365: We can add a log here, just in case.

tpaksu (changes requested):
Code looks good, thanks for handling the e-mail related issues! There's one thing I've found, a code duplication possibility, which account link retrieval was added before and you wrote it again. And one nitpick, should we log the errors when we fall into the catch block in the account link retrieval code? I didn't test the code execution yet, I'll re-do my review after you check those. Thank you!

includes/wc-payment-api/class-wc-payments-api-client.php
luizreis: I just see the server endpoint added in that PR. I also searched the client for any mention of a request to `capital_links`, but couldn't find anything except this method, which is what makes the request to the server method you mentioned above. Am I missing something here?

includes/class-wc-payments-account.php
luizreis: Most errors are already logged by [`extract_response_body`](https://github.com/Automattic/woocommerce-payments/blob/e33a0ea28d4e466e8443a9782f7788ad9105ad86/includes/wc-payment-api/class-wc-payments-api-client.php#L1711), with the exception being [failing to encode the request body](https://github.com/Automattic/woocommerce-payments/blob/e33a0ea28d4e466e8443a9782f7788ad9105ad86/includes/wc-payment-api/class-wc-payments-api-client.php#L1603). I was afraid logging it again here would clutter the logs quite a bit, and left the logging to the API client.

I'm definitely open to adding a new call here if you don't see any issues with it.

includes/class-wc-payments-account.php
luizreis: I think so. Also, all other checks on the plugin are using `manage_woocommerce`, so I changed it to that in https://github.com/Automattic/woocommerce-payments/pull/3673/commits/19ea8d421db7d0418bd3ac47cb47b1c0fe3e97fd.

includes/wc-payment-api/class-wc-payments-api-client.php
tpaksu: This creates an internal REST request to the endpoint created by Marcin, and that endpoint returns the link to this method as the method output. I believe the server endpoint was created to create a link to fulfill a request from the client, before the designs showed us the e-mail flow, I think it'll be good to eliminate the internal REST request and directly call the Stripe API, how does it sound? /cc @marcinbot

marcinbot: It looks right to me. My PR was just adding a server endpoint, and this one seems to call it correctly.

We can't call the Stripe API directly because the client shouldn't have access to the secret key.

includes/class-wc-payments-account.php
tpaksu: Yes you're right, that seems to log errors already, I was more curious about Stripe errors, like "offer don't exists" or "wrong account ID" etc, but they return back an account link even if the loan doesn't exist, so it won't be a problem not logging it, as we don't have any user entered inputs here. Feel free to skip this :)

tpaksu (approved):
Tests well, got all my answers, thanks for handling this, works perfectly! :shipit:

PR #3674, opened 2022-01-19
Add an endpoint to retrieve an uploaded file

### LINKED ISSUE #3595: Add an endpoint to retrieve an uploaded file
Currently, there's an endpoint to upload a file that until recently was used on the new dispute form for upload evidence functionality. Now, the [card reader receipt settings page](https://github.com/Automattic/woocommerce-payments/issues/3212) also uses it to upload a merchant's icon and logo.

Until now, we didn't need to preview uploaded files. But previewing files on the receipt settings page is highly desirable. For this, we'll need to add a new endpoint to retrieve uploaded files.

I suggest that we create a new controller - `WC_REST_Payments_Files_Controller` - to add the new endpoint.

**Bonus:** The file upload endpoint `POST /payments/file` is presently inside `WC_REST_Payments_Disputes_Controller`, which made sense when it was the only place where file upload functionality was being used. Now that we have more consumers of this API, IMHO we should move this endpoint into the new Files controller. That will require us to refactor all usages of the endpoint:

* https://github.com/Automattic/woocommerce-payments/blob/d184f4611a2db4bb67608092cba9427c022b7e9b/includes/wc-payment-api/class-wc-payments-api-client.php#L787

* https://github.com/Automattic/woocommerce-payments/blob/8613c7a8c4b6d15a3e9642d0b3ca085a8e2cd5af/client/card-readers/settings/file-upload.js#L17

dpaun1985:
Fixes #3595

Depends on 1461-gh-Automattic/woocommerce-payments-server.

* add `GET payments/file` endpoint to retrieve a file content from Stripe
* move upload file logic from `WC_REST_Payments_Disputes_Controller` to the specific files controller 
 

#### Testing instructions

* Upload a file (you can use file upload options from the evidence submission  - disputes section)
* Use `/wp-json/wc/v3/payments/file/<file_id>` to retrieve the uploaded file.

-------------------

**Post merge**


includes/class-wc-payments.php
```diff
@@ -604,6 +604,10 @@ public static function init_rest_api() {
 		$charges_controller = new WC_REST_Payments_Reader_Controller( self::$api_client, self::$card_gateway, self::$in_person_payments_receipts_service );
 		$charges_controller->register_routes();
 
+		include_once WCPAY_ABSPATH . 'includes/admin/class-wc-rest-payments-files-controller.php';
+		$disputes_controller = new WC_REST_Payments_Files_Controller( self::$api_client );
```
anu-rock, line 608: Perhaps you mean `$files_controller`? 😉

tests/unit/admin/test-class-wc-rest-payments-files-controller.php
```diff
@@ -0,0 +1,87 @@
...
+		$this->assertSame( 'test_file_content', $response->get_data() );
+		$this->assertSame( 200, $response->status );
+	}
+
+	public function test_get_file_no_file() {
+		$this->mock_api_client
+			->expects( $this->once() )
+			->method( 'get_file_contents' )
+			->with( 'file_mock_ID' )
+			->willThrowException( new API_Exception( "Error: No such file upload: 'file_mock_ID'", 'resource_missing', 500 ) );
```
anu-rock, line 59: Ideally, such an error would return a 400 response.

includes/admin/class-wc-rest-payments-files-controller.php
```diff
@@ -0,0 +1,99 @@
...
+		);
+	}
+
+	/**
+	 * Create file and respond with file object via API.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 */
+	public function upload_file( $request ) {
+		return $this->forward_request( 'upload_evidence', [ $request ] );
```
anu-rock, line 53: I think it's a good time to rename `upload_evidence` as `upload_file`.

includes/admin/class-wc-rest-payments-disputes-controller.php
```diff
@@ -60,16 +60,6 @@ public function register_routes() {
 			]
 		);
 
-		register_rest_route(
```
anu-rock, line 63: This change will likely impact code in `woocommerce-payments/client/card-readers/settings/file-upload.js`. I hope you've already tested it to be working fine.

includes/admin/class-wc-rest-payments-files-controller.php
```diff
@@ -0,0 +1,99 @@
...
+	 *
+	 * @param WP_REST_Request $request - request object.
+	 *
+	 * @return WP_Error|WP_HTTP_Response
+	 */
+	public function get_file( WP_REST_Request $request ) {
+		$file_id = $request->get_param( 'file_id' );
+		$result  = $this->forward_request( 'get_file_contents', [ $file_id ] );
+
+		if ( $result instanceof WP_Error ) {
```
anu-rock, line 67: [is_wp_error](https://developer.wordpress.org/reference/functions/is_wp_error/) could be helpful here.

includes/admin/class-wc-rest-payments-files-controller.php
```diff
@@ -0,0 +1,99 @@
...
+			'rest_pre_serve_request',
+			function ( bool $served, WP_HTTP_Response $response ) : bool {
+				echo $response->get_data(); // @codingStandardsIgnoreLine
+				return true;
+			},
+			10,
+			2
+		);
+
+		return new WP_HTTP_Response(
```
anu-rock, line 92: Will this send out the response with the Content-Disposition header? If not, wouldn't that be better?

mgascam: Following @anu-rock's comment, I wonder if it would make sense to add a parameter to be able to specify the value of the Content-Disposition header? I mean to set it as `attachment` or `inline`

includes/admin/class-wc-rest-payments-disputes-controller.php
dpaun1985: I used the same path as before ( `payments/file` ), so we should be safe

includes/admin/class-wc-rest-payments-files-controller.php
dpaun1985: by default, Content-Disposition is considered inline, but for clarity I just added also the content disposition header.

anu-rock (approved):
LGTM, though I will confess that I have not tested the endpoint myself.

Thanks for addressing my feedback, you're awesome!

mgascam (approved):
LGTM! I did some manual testing and works fine

includes/admin/class-wc-rest-payments-files-controller.php
```diff
@@ -0,0 +1,102 @@
...
+			]
+		);
+
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<file_id>\w+)',
+			[
+				'methods'             => WP_REST_Server::READABLE,
+				'callback'            => [ $this, 'get_file' ],
+				'permission_callback' => [],
```
kalessil, line 42: Any chance here to keep the permissions checked in generals, but allow free access for branding-related files?
Currently, looks like a potential security issue.

anu-rock: Hmm, good point. We might be able to use the `purpose` attribute of the file object returned by Stripe to allow public access to `business_icon` and `business_logo`, and put all other files secured. Wdyt, @dpaun1985?

kalessil: Might be an option! Then we still need to retrieve the file completely, but at least the endpoint itself will be able to decide if to share the content or forbid.

kalessil (changes requested):
Let's re-iterate with purpose attributes as per the discussion above.

includes/admin/class-wc-rest-payments-files-controller.php
dpaun1985: As discussed in p1643122246349800-slack-C02C00HCJ8J , we will check the permissions based on `purpose` field

includes/class-wc-payments-file-service.php
```diff
@@ -0,0 +1,40 @@
...
+
+
+	/**
+	 * Check if a file purpose is public or needs permissions.
+	 *
+	 * @param string $purpose - file purpose.
+	 *
+	 * @return bool
+	 */
+	public function file_need_access_permissions( string $purpose ) : bool {
```
anu-rock, line 32: I found the name of this method a bit confusing. How about `is_file_restricted` or `is_file_private`? Going by the "deny by default" security principle, it may even be `is_file_public` with internal boolean logic reversed.

includes/admin/class-wc-rest-payments-files-controller.php
```diff
@@ -61,20 +60,33 @@ public function upload_file( $request ) {
...
+		if ( ! $purpose ) {
+			$file = $this->forward_request( 'get_file', [ $file_id ] );
+			if ( is_wp_error( $file ) ) {
+				return $this->file_error_response( $file );
+			}
+			$purpose = $file->data['purpose'];
+			set_transient( WC_Payments_File_Service::CACHE_KEY_PREFIX_PURPOSE . $file_id, $purpose, WC_Payments_File_Service::CACHE_PERIOD );
+		}
+
+		if ( $file_service->file_need_access_permissions( $purpose ) && ! $this->check_permission() ) {
```
anu-rock, line 77: Shouldn't it be an OR condition?

includes/class-wc-payments-file-service.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class which handles files.
+ */
+class WC_Payments_File_Service {
```
anu-rock, line 13: Thanks for creating the File service 👍

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -881,6 +881,18 @@ public function get_file_contents( string $file_id ) : array {
 		return $this->request( [], self::FILES_API . '/' . $file_id . '/contents', self::GET );
 	}
 
+	/**
+	 * Retrieve a file content via API.
```
anu-rock, line 885: Perhaps you mean "file details"?

includes/admin/class-wc-rest-payments-files-controller.php
dpaun1985: no, we need both conditions. `file_need_access_permissions` checks if the file shouldn't be public, and `check_permission` if the user has required permissions. So, if the file isn't public **and** the user does not have required permissions, we should throw access denied response.

anu-rock: I see. Thanks for clarifying. In that case, `check_permission` is named oddly. I thought it meant a user _has_ permission if it returned true.

anu-rock (approved):
LGTM with the new changes 🎉

includes/class-wc-payments-file-service.php
```diff
@@ -0,0 +1,40 @@
...
+
+
+	const FILE_PURPOSE_NO_PERMISSIONS = [
```
kalessil, line 16: `FILE_PURPOSE_PUBLIC` would be more speaking here.

includes/class-wc-payments-file-service.php
```diff
@@ -0,0 +1,40 @@
...
+
+	/**
+	 * Check if a file purpose is public or needs permissions.
+	 *
+	 * @param string $purpose - file purpose.
+	 *
+	 * @return bool
+	 */
+	public function is_file_public( string $purpose ) : bool {
+		if ( in_array( $purpose, static::FILE_PURPOSE_NO_PERMISSIONS, true ) ) {
```
kalessil, line 33: A one-liner here can be used: `return in_array(...)`

kalessil (approved):
LGTM! Left a couple of comments with possible improvements.

PR #3675, opened 2022-01-19
Add/order service

jessepearson:
Fixes issues with repetitive code.

This is a start of needed code cleanup and refactoring. The main purpose of this was to handle updating payments and their related orders and statuses in the same fashion no matter if they come from webhooks, website checkout, or in person payments.

* Create new class of `WC_Payments_Order_Service`.
  * This service now handles the responsibility of updating the order status and adding order notes when payments complete, fail, get disputed, etc.
  * For payments, it will update the order status and add an order note if needed, and it will also update the intent status if the status is available.
  * For disputes, it updates the order status and adds the proper notes.
    * If a dispute is lost, a refund is created. 
* Updated (most) calls to `$order->update_status()` with call to respective method in `WC_Payments_Order_Service`.

### Testing instructions

* All tests should pass.
* All manual tests should be done with the webhook listener running. This is due to there are webhooks that can process orders, and we need to make sure orders are not being double processed.
* Manual tests can be done using UPE checkout, as it uses all methods within the WCPay gateway class.

#### Manual order tests

_Successful order:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be Processing or Completed, depending on if the product is virtual and downloadable.
* Order should have 3 notes:
  * _A payment of TOTAL was started using WooCommerce Payments (INTENT_ID)._ (Will not show if using a saved card.)
  * _Order status changed from Pending payment to Processing._ 
  * _A payment of TOTAL was successfully charged using WooCommerce Payments (CHARGE_ID)._ Total is the order total, and Charge_Id will be a link to the transaction details page.

_Failed order:_
* Add item to cart and check out with `4000000000000002` as your test card.
* Order will fail.
* View order in admin.
* Order should be in Failed status.
* Order should have 3 notes:
  *  _A payment of TOTAL was started using WooCommerce Payments (INTENT_ID)._
  * _Order status changed from Pending payment to Failed._ 
  * _A payment of TOTAL failed using WooCommerce Payments (INTENT_ID)._ Total is the order total, and Intent_Id will be a link to the transaction details page.

_Pending order:_
* Go to Payments > Settings and enable Bancontact as a payment method.
* Add item to cart and go to checkout page.
* Update your currency to Euro via MC widget.
* Select _Use a new payment method_, and then select Bancontact.
* Click _Place order_, but do not complete the order on the next page.
* View order in admin.
* Order should be in Pending payment status.
* Order should have 1 note:
  * _A payment of TOTAL was started using WooCommerce Payments (INTENT_ID)._

_On-hold order:_
* Go to Payments > Settings and then the _Transactions and deposits_ section.
* Enable the _Issue an authorization on checkout, and capture later_ setting.
* Add item to cart and check out with test card.
* View order in admin.
* Order should be in On Hold status.
* Order should have 3 notes:
  *  _A payment of TOTAL was started using WooCommerce Payments (INTENT_ID)._
  * _Order status changed from Pending payment to On hold._ 
  * _A payment of TOTAL was authorized using WooCommerce Payments (INTENT_ID)._ Total is the order total, and Intent_Id will be a link to the transaction details page.

_Capture authorized payment:_
* Using previous order, go to _Order actions_ at the top right of the order in the admin.
* Choose _Capture charge_ and either click the right arrow or _Update_.
* Order should be Processing or Completed, depending on if the product is virtual and downloadable.
* Order should have 2 new notes:
  * _Order status changed from On hold to Processing._ 
  * _A payment of TOTAL was successfully charged using WooCommerce Payments (CHARGE_ID)._ Total is the order total, and Charge_Id will be a link to the transaction details page.

_Cancel authorized payment:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be in On Hold status.
* Choose _Cancel authorization_ and either click the right arrow or _Update_.
* Order should be in Cancelled status.
* Order should have 2 notes:
  * _Order status changed from On hold to Cancelled._ 
  * _Payment authorization was successfully cancelled._

_Capture authorization failure:_
* Enable below snippet in your store, either through theme or Code Snippets.
* Add item to cart and check out with test card.
* View order in admin.
* Order should be in On Hold status.
* Choose _Capture charge_ and either click the right arrow or _Update_.
* Order should remain in On hold status.
* Order should have 1 new note:
  * _A capture of TOTAL failed to complete using WooCommerce Payments (false_id). Error: No such payment_intent: 'false_id'_

```
add_filter( 'woocommerce_order_get_transaction_id', function() {
	return 'false_id';
});
```

### Webhook tests

For these tests, it's easiest to continue to leave authorization on and place orders that are authorized so that the webhooks can act on them. If you need to enable this again: 

* Go to Payments > Settings and then the _Transactions and deposits_ section.
* Enable the _Issue an authorization on checkout, and capture later_ setting.

You will need to get meta data from each order, such as the intent and/or charge ids. Unless you feel like digging into the database, I would recommend a snippet (below) or [plugin to expose this data](https://wordpress.org/plugins/post-meta-inspector/). With either the snippet or plugin enabled, the hidden meta data can be found within the order itself.

```
add_filter( 'is_protected_meta', '__return_false' ); 
```

To get around webhook validation this snippet can be used to post via a tool like Postman:

```
add_filter( 'user_has_cap', function( $caps ) {
	$caps['manage_woocommerce'] = true;
	return $caps;
});
```

Webhook url is:

```
https://YOUR_SITE/wp-json/wc/v3/payments/webhook
```

_Successful payment intent:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be On hold.
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Processing status.
* Order should have 2 new notes:
  * _Order status changed from On hold to Processing._ 
  * _A payment of TOTAL was successfully charged using WooCommerce Payments (CHARGE_ID)._

```
{
    "type": "payment_intent.succeeded",
    "data": {
        "object": {
            "id": "pi_INTENT_ID",
            "status": "succeeded",
            "charges": {
                "data": [
                    {
                        "id": "ch_CHARGE_ID"
                    }
                ]
            }
        }
    }
}
```

_Failed payment intent:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be On hold.
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Failed status.
* Order should have 2 new notes:
  * _Order status changed from On hold to Failed._ 
  * _ A payment of TOTAL failed using WooCommerce Payments (INTENT_ID). The customer’s bank account could not be located._

```
{
    "type": "payment_intent.payment_failed",
    "data": {
        "object": {
            "id": "pi_INTENT_ID",
            "status": "failed",
            "object": "payment_intent",
            "charges": {
                "data": [
                    {
                        "id": "ch_CHARGE_ID",
                        "failure_code": "no_account",
                        "payment_method": "pm_PAYMENT_METHOD_ID",
                        "payment_method_details": {
                            "type": "us_bank_account"
                        }
                    }
                ]
            }
        }
    }
}
```

_Expired charge authorization:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be On hold.
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Cancelled status.
* Order should have 2 new notes:
  * _Order status changed from On hold to Cancelled._ 
  * _ Payment authorization has expired (INTENT_ID)._

```
{
    "type": "charge.expired",
    "data": {
        "object": {
            "id": "ch_CHARGE_ID",
            "charges": {
                "data": [
                    {
                        "id": "pi_INTENT_ID"
                    }
                ]
            }
        }
    }
}
```

_Successful terminal capture:_
* Add item to cart and check out with test card.
* View order in admin.
* Order should be On hold.
* Post the below JSON to your site, you should get a `succeeded` result.
  * **Note:** This webhook requires a different url: 
    * `https://YOUR_SITE/wp-json/wc/v3/payments/orders/ORDER_ID/capture_terminal_payment`
* Order should be in Completed status.
* Order should have 3 new notes:
  * _Order status changed from On hold to Processing._ 
  * _A payment of TOTAL was successfully captured using WooCommerce Payments (CHARGE_ID)._
  * _Order status changed from Processing to Completed._

```
{
    "payment_intent_id": "pi_INTENT_ID",
    "order_id": "ORDER_ID"
}
```

_Dispute created:_
When testing with a card that creates a dispute, my local site receives the webhook so fast that the checkout process does not have time to fully complete. This means the order does not get updated correctly. Sending the webhook manually _after_ the order is created resolves the issue.
* You can now disable the _Issue an authorization on checkout, and capture later_ setting under Payment > Settings.
* Add item to cart and check out with valid test card.
* View order in admin.
* Order should be Processing.
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Cancelled status.
* Order should have 2 new notes:
  * _Order status changed from Processing to On hold._ 
  * _ Payment has been disputed as product_not_received. See dispute overview for more details._
  * _dispute overview_ should link to the dispute overview page.

```
{
    "type": "charge.dispute.created",
    "data": {
        "object": {
            "id": "dp_123",
            "charge": "ch_CHARGE_ID",
            "reason": "product_not_received"
        }
    }
}
```

_Dispute won:_
* The order from the last test can be used.
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Completed status.
* Order should have 2 new notes:
  * _Order status changed from On hold to Completed._ 
  * _Payment dispute has been closed with status won. See dispute overview for more details._
  * _dispute overview_ should link to the dispute overview page.

```
{
    "type": "charge.dispute.closed",
    "data": {
        "object": {
            "id": "dp_123",
            "charge": "ch_CHARGE_ID",
            "status": "won"
        }
    }
}
```

_Dispute lost:_
* The order from the last test can be used (again).
* Post the below JSON to your site, you should get a `success` result.
* Order should be in Refunded status.
* Order should have 2 new notes:
  * _Order status changed from Completed to Refunded._ 
  * _Payment dispute has been closed with status lost. See dispute overview for more details._
  * _dispute overview_ should link to the dispute overview page.
* A refund should be created on the order refunding the full amount. The note on the refund should be _Dispute lost._

```
{
    "type": "charge.dispute.closed",
    "data": {
        "object": {
            "id": "dp_123",
            "charge": "ch_CHARGE_ID",
            "status": "lost"
        }
    }
}
```
-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.8.0#changes-to-how-order-status-and-notes-are-set-within-the-plugin


includes/admin/class-wc-rest-payments-webhook-controller.php
```diff
@@ -512,17 +536,14 @@ private function get_order_from_event_body_intent_id( $event_body ) {
...
+					/* translators: %1: intent ID */
 					__( 'Could not find order via intent ID: %1$s', 'woocommerce-payments' ),
 					$intent_id
 				),
 				'order_not_found'
 			);
 		}
 
-		// Get an updated set of order properties to avoid race conditions when the server sends the paid webhook before we've finished processing the original payment request.
-		$order->get_data_store()->read( $order );
```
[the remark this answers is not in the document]
jessepearson: It happens in the order service itself when the order is being "prepared for processing"
https://github.com/Automattic/woocommerce-payments/pull/3675/files/e23404e5122a26349b5be6b142c5d43e62aea6e4#diff-7114744ea90ad0b842ff7b28874379c3318f36eb1cd82d3266caebc8361aaa9bR463

includes/class-wc-payments-order-service.php
```diff
@@ -0,0 +1,508 @@
...
+	/**
+	 * Adds the success order note, if needed, and additional message, if included.
+	 *
+	 * @param WC_Order $order     Order object.
+	 * @param string   $intent_id The ID of the intent associated with this order.
+	 * @param string   $charge_id The charge ID related to the intent/order.
+	 *
+	 * @return void
+	 */
+	private function add_success_note( $order, $intent_id, $charge_id ) {
```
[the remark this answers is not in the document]
jessepearson: Thanks for noticing this. I had initially had an optional message for all of the notes, but removed them all and left them just for the failures, this needed to be this way _just in case_ there was a separate message passed. I've updated it locally and will include it in the next push.

includes/class-wc-payments-order-service.php
```diff
@@ -0,0 +1,732 @@
...
+
+		// Check if the order is already complete.
+		if ( function_exists( 'wc_get_is_paid_statuses' ) ) {
+			if ( $order->has_status( wc_get_is_paid_statuses() )
+				|| $clone_order->has_status( wc_get_is_paid_statuses() ) ) {
+				return false;
+			}
+		}
+
+		if ( $this->is_order_locked( $order, $intent_id )
```
dmallory42, line 668: Sorry for the drive-through comment, but I _think_ this function would only need to be called on the original order and not the clone order (as far as I can tell it would do the same thing as all it would do is get the order ID, and use that to check if the transient exists. 🙂

includes/class-wc-payments-order-service.php
```diff
@@ -0,0 +1,732 @@
...
+	 *
+	 * @return bool
+	 */
+	private function order_prepared_for_processing( $order, $intent_id ) {
+		if ( ! is_a( $order, 'WC_Order' ) ) {
+			return false;
+		}
+
+		// Read the latest order properties from the database to avoid race conditions if webhook was handled during this request.
+		$clone_order = clone $order;
```
dmallory42, line 657: This looks good, do you think it's worth extracting this part to a private function in case we need to reuse it anywhere else? I was thinking something like `is_order_paid()`.

includes/class-wc-payments-order-service.php
jessepearson: No worries, you are correct here. I just updated locally so that it will be included with the next push.

includes/class-wc-payments-order-service.php
jessepearson: I think this is a good idea. Were you thinking like this?

```
	private function order_prepared_for_processing( $order, $intent_id ) {
		if ( ! is_a( $order, 'WC_Order' ) ) {
			return false;
		}

		if ( $this->is_order_paid( $order ) ) {
			return false;
		}

		if ( $this->is_order_locked( $order, $intent_id ) ) {
			return false;
		}

		// Lock the order.
		$this->lock_order_payment( $order, $intent_id );

		return true;
	}

	/**
	 * Checks to see if the current order, and a fresh copy of the order from the database are paid.
	 *
	 * @param WC_Order $order The order being checked.
	 *
	 * @return boolean True if it has a paid status, false if not.
	 */
	private function is_order_paid( $order ) {
		// Read the latest order properties from the database to avoid race conditions if webhook was handled during this request.
		$clone_order = clone $order;
		$clone_order->get_data_store()->read( $clone_order );

		// Check if the order is already complete.
		if ( function_exists( 'wc_get_is_paid_statuses' ) ) {
			if ( $order->has_status( wc_get_is_paid_statuses() )
				|| $clone_order->has_status( wc_get_is_paid_statuses() ) ) {
				return true;
			}
		}

		return false;
	}
```

dmallory42: Yep, that's exactly what I had in mind!

includes/class-wc-payments-order-service.php
```diff
@@ -0,0 +1,706 @@
...
+	 * Updates a terminal order to completed status, while adding a note with a link to the transaction.
+	 *
+	 * @param WC_Order $order         Order object.
+	 * @param string   $intent_id     The ID of the intent associated with this order.
+	 * @param string   $intent_status The status of the intent related to this order.
+	 * @param string   $charge_id     The charge ID related to the intent/order.
+	 *
+	 * @return void
+	 */
+	public function mark_terminal_payment_completed( $order, $intent_id, $intent_status, $charge_id ) {
```
dmallory42, line 242: I wonder whether it's possible to use `mark_payment_completed` here, rather than having two functions. Perhaps by passing through a bool flag to `mark_payment_completed` such as `$terminal = false` and then setting it to true if it is a terminal payment.

Is there a significant difference between how we mark a terminal payment as completed vs a normal payment? If there is, I think it's fine to have the two functions of course.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1065,7 +1080,7 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 			if ( 'requires_action' === $status && $payment_information->is_merchant_initiated() ) {
 				// Allow 3rd-party to trigger some action if needed.
 				do_action( 'woocommerce_woocommerce_payments_payment_requires_action', $order, $intent_id, $payment_method, $customer_id, $charge_id, $currency );
-				$order->update_status( 'failed' );
+				$this->order_service->mark_payment_failed( $order, $intent_id, $status, $charge_id, '' );
```
dmallory42, line 1083: Extreme Nit Pick: I think we don't need to pass the `''` as the last parameter here, because it's the default value of `$message` if no parameter is passed to the function.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2077,6 +1992,9 @@ public function cancel_authorization( $order ) {
 				)
 			);
 		}
+
+		$order->update_meta_data( '_intention_status', $status );
```
dmallory42, line 1996: It looks like we do this in all of the `mark_payment_*` functions (as part of `complete_order_processing`) - is there a case where this might not happen that means we'd need to do it here as well?

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -48,11 +58,13 @@ public function setUp() {
 		$this->mock_api_client       = $this->createMock( WC_Payments_API_Client::class );
 		$this->mock_gateway          = $this->createMock( WC_Payment_Gateway_WCPay::class );
 		$this->mock_customer_service = $this->createMock( WC_Payments_Customer_Service::class );
+		$this->order_service         = new WC_Payments_Order_Service();
```
dmallory42, line 61: Shouldn't this be a mock (since the Order Controller is what's under test in this file)?

(Please ignore this comment if the tests aren't complete yet!)

dmallory42:
In process of testing now and will update when I'm done, but just to mention that I actually have 2 notes, not 2 when testing the successful order - however, the notes make sense and look correct to me.

Screenshot 2022-02-09 at 12 58 20

dmallory42:
Tested all of the manual order tests, but get this when I try and do the webhook ones - I've probably done something wrong! Let me know and I can finish up the tests and approve. It all looks good so far, I left a few comments but nothing blocking.

Screenshot 2022-02-09 at 14 53 38

jessepearson:
> Tested all of the manual order tests, but get this when I try and do the webhook ones - I've probably done something wrong! Let me know and I can finish up the tests and approve. It all looks good so far, I left a few comments but nothing blocking.

Did you add the filter mentioned at the beginning on the webhook tests section? It allows you to get around the whole authentication process for posting API requests.

```
add_filter( 'user_has_cap', function( $caps ) {
	$caps['manage_woocommerce'] = true;
	return $caps;
});
```

includes/class-wc-payments-order-service.php
jessepearson: Basically all `mark_terminal_payment_completed` is doing at this point is moving the order to the `completed` status. It also updates the intent on the order, like the other methods in the service does, but I believe that's already done in the code for the terminal transaction and in `mark_payment_completed`.

When it's called, [it has already stepped through the process](https://github.com/Automattic/woocommerce-payments/blob/add/order-service/includes/admin/class-wc-rest-payments-orders-controller.php#L146-L151) of calling [`update_order_status_from_intent`](https://github.com/Automattic/woocommerce-payments/blob/add/order-service/includes/class-wc-payment-gateway-wcpay.php#L1288), which has called `mark_payment_completed`. `mark_payment_completed` adds the note and moves the order to processing if it's a physical product, and since the order is supposedly in person, it should be moved to completed assuming the merchant is physically selling the item to the customer.

Since all it's doing is moving the order to the `completed` status, it's not needed, I was just trying to consolidate the calls to `update_status` for orders into this service.

If there was a better/standard intent object being used between all of these transactions, we could have a flag within that to say "this is a terminal transaction" and it could be marked as completed rather than processing initially, but that's something for a future refactor. I went down so many rabbit holes with trying to make things work better, but things like multiple intent types and using arrays and objects as intents becomes a huge issue.

includes/class-wc-payment-gateway-wcpay.php
jessepearson: Yes. The whole method/function this is included in cancels an authorization, and if it is successful, it hands off to the order service to `mark_payment_capture_cancelled`, but if there's some sort of error,[ the next steps are handled there](https://github.com/Automattic/woocommerce-payments/blob/add/order-service/includes/class-wc-payment-gateway-wcpay.php#L1968-L1997). There's an early return after `mark_payment_capture_cancelled` is called so that the meta data isn't updated again if it's called, otherwise the meta data does need to be updated.

includes/class-wc-payments-order-service.php
```diff
@@ -0,0 +1,706 @@
...
+	 * @param string   $charge_id     The charge ID related to the intent/order.
+	 *
+	 * @return void
+	 */
+	public function mark_payment_completed( $order, $intent_id, $intent_status, $charge_id ) {
+		if ( ! $this->order_prepared_for_processing( $order, $intent_id ) ) {
+			return;
+		}
+
+		$this->update_order_status( $order, 'payment_complete', $intent_id );
```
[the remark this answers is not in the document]
jessepearson: I like this and considered it, however, I didn't do this due to the difference between intent and order statuses. I feel like the names would need to be more descriptive like `ORDER_STATUS_COMPLETE`. Also, `payment_complete` isn't an actual order status, it's a flag to mark the payment complete in the order.

I will reach out in the morning to see what we can do here that works well.

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
jessepearson: Some of the tests had mocks, where others used the actual service. This is due to the existing tests either checked for notes being added to the orders or they checked for the changed order status. This does mean that some tests are somewhat duplicated, but I felt it was better that way so function X can be tested that it does call another function to change order status, for example.

jessepearson:
> In process of testing now and will update when I'm done, but just to mention that I actually have 3 notes, not 2 when testing the successful order - however, the notes make sense and look correct to me.

The first `started` one is added due to my changes. IIRC, there are a lot of instances where a customer can trigger an order to be created and it will just sit in Pending status with no notes. For instance the _Pending order_ test using the _Bancontact_ payment method. I am not sure if there's a way to _not_ include that note for other methods, like credit card. 🤔

dmallory42:
> Did you add the filter mentioned at the beginning on the webhook tests section? It allows you to get around the whole authentication process for posting API requests.

I thought I did, but perhaps I didn't add it in the right place 🤦‍♂️ if you wouldn't mind showing me when we chat tomorrow, I'd appreciate it! 

> The first started one is added due to my changes. IIRC, there are a lot of instances where a customer can trigger an order to be created and it will just sit in Pending status with no notes. For instance the Pending order test using the Bancontact payment method. I am not sure if there's a way to not include that note for other methods, like credit card. 🤔

I think it's a sensible change and makes sense, I just wanted to note it to make sure it was expected.

includes/class-wc-payment-gateway-wcpay.php
dmallory42: Thanks for explaining!

includes/class-wc-payments-order-service.php
dmallory42: Thanks for the detailed explanation! I agree that we should look at this when we have better models to pass around our code, it works well for now and this PR is laying the groundwork for future refactors anyway.

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
dmallory42: I think in general we should try and mock everything apart from what we're currently testing - although I realise that  may require a re-think to the tests here, in the past they were probably doing too much and now we have better split the responsibilities we can rely on the Order service unit tests to verify that order status and notes are getting updated properly.

With that in mind, I don't think we need to address it in this PR because it's already quite large - I think maybe we could address it in a follow up PR at some point? It's definitely not high priority or blocking by any means.

jessepearson:
> Tested and all works well, you can :shipit: after taking care of the constants as we discussed in our huddle. Nice work on this issue!

Added the order status constants within the order service, but I _did not_ add any sort of constants for the intent statuses. The reasoning behind that was due to if I added the intent status constants to the intent object, I would feel like they should be used throughout the code and not just in the order service (hope that makes sense).

> I think we should also consider having these tests run as part of the GlobalStep testing for the next release, since it's quite a big refactor to how we process orders.

Added the manual tests to Globalstep testing, but I don't believe they do API testing, so didn't include those.

PR #3682, opened 2022-01-20
New platform checkout user on checkout page

Mayisha:

Description:
In the `short code` checkout page, this PRs inserts a new section which is visible only if the current user is not yet registered to the platform checkout.

### Testing instructions
- Enable `_wcpay_feature_platform_checkout`
- Set short code version of the checkout page
- Return from `isRegisteredUser=false` from `client/components/platform-checkout/hooks/use-platform-checkout-user.js`
- This new section should be placed above the `Place Order` button properly
- The phone number input is not functional yet. Will be implemented separately

-------------------


harriswong:
> The phone number input is not functional yet. Will be implemented separately

@Mayisha would you like me to create a follow-up or have you already created one?

Mayisha:
@harriswong created a separate PR [here](https://github.com/Automattic/woocommerce-payments/pull/3699)

harriswong:
Hey @Mayisha, I tried a few things and I am not sure why I am not seeing the component being rendered. The breakpoint I set in index indicates that it is going through the conditional check:
image

When I check the "Save payment information to my account for future purchases", I get the "component is not rendered" exclamation mark. However, I do see the component in there. Maybe mine didn't mount?
image

What I did on top of the testing instructions:
- I am using docker. I ran the `npm run recreate`
- I created a new WP page `http://localhost:8082/shortcode-checkout/` with only the short code inside `[woocommerce_checkout]`. I also tested this on `http://localhost:8082/checkout/`. Same thing.
- I also disabled UPE with
```diff
 // hook for handling API calls to get and create platform checkout user.
 const usePlatformCheckoutUser = () => {
-       const isRegisteredUser = true;
+       const isRegisteredUser = false;

        return {
                isRegisteredUser,
diff --git a/includes/class-wc-payments-features.php b/includes/class-wc-payments-features.php
index b7ca0573..a127d4a9 100644
--- a/includes/class-wc-payments-features.php
+++ b/includes/class-wc-payments-features.php
@@ -23,7 +23,7 @@ class WC_Payments_Features {
         * @return bool
         */
        public static function is_upe_enabled() {
-               return '1' === get_option( self::UPE_FLAG_NAME, '0' );
+               return false; //'1' === get_option( self::UPE_FLAG_NAME, '0' );
        }

        /**
@@ -32,7 +32,7 @@ class WC_Payments_Features {
         * @return bool
         */
        public static function did_merchant_disable_upe() {
-               return 'disabled' === get_option( self::UPE_FLAG_NAME, '0' );
+               return true; //'disabled' === get_option( self::UPE_FLAG_NAME, '0' );
        }

        /**
@@ -95,7 +95,7 @@ class WC_Payments_Features {
         * @return bool
         */
        public static function is_platform_checkout_enabled() {
-               return '1' === get_option( '_wcpay_feature_platform_checkout', '0' );
+               return true || '1' === get_option( '_wcpay_feature_platform_checkout', '0' );
        }
```

Have you experienced this error before? Am I missing a step? Thanks!

harriswong:
@Mayisha and I had a zoom chat over this issue. We think we found the issue that occurred on my machine. 
In `plugins/woocommerce/assets/js/frontend/checkout.js`, line 373 `// Always update the fragments`, this block of code executes *after* React rendered the `<CheckoutPageSaveUser>` component. As a result, the fragment removed the `<CheckoutPageSaveUser>` component.

There are 2 fragments that load, ` ".woocommerce-checkout-review-order-table": ` and ` ".woocommerce-checkout-payment". ` ".woocommerce-checkout-payment" is the one which erased the react component. Here are some screenshot
We can see that the `<CheckoutPageSaveUser>` component is there prior to the fragment replacement:
image

As soon as it loops to the `.woocommerce-checkout-payment`, it replaces the element and removed the `<CheckoutPageSaveUser>` component:
image

We are not sure why my machine loads the fragments differently than Mayisha's machine.

We think this may be due to the sequence of how scripts are loaded (ie. loading checkout.js first vs last). As a PoC, we added a `setTimeout()` around the `<CheckoutPageSaveUser>` render. Then, I can then see the component. Maybe something to do with the sequence of enqueue scripts?

client/components/platform-checkout/hooks/use-platform-checkout-user.js
```diff
@@ -0,0 +1,10 @@
+// hook for handling API calls to get and create platform checkout user.
+const usePlatformCheckoutUser = () => {
+	const isRegisteredUser = true;
+
+	return {
+		isRegisteredUser,
```
harriswong, line 6: We now have an endpoint for this 🙌 . What do you think about creating a follow-up to make an API call?

client/components/platform-checkout/save-user/style.scss
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	.save-details-form {
+		display: flex;
+		flex-direction: column;
+		font-size: 16px;
+		line-height: 26px;
+		font-weight: 400;
+		background-color: #f8f8f8;
```
harriswong, line 42: Interesting how the colour here is different than the other payment box `#fafafa`. The design does say f8.

Are these UI components something we can use from the `@woocommerce-components` ([link](https://woocommerce.github.io/woocommerce-admin/#/components/))? Maybe we can avoid writing our own CSS?

includes/platform-checkout-user/class-platform-checkout-save-user.php
```diff
@@ -0,0 +1,49 @@
...
+	public function register_checkout_page_scripts() {
+		$script_src_url    = plugins_url( 'dist/platform-checkout.js', WCPAY_PLUGIN_FILE );
+		$style_url         = plugins_url( 'dist/platform-checkout.css', WCPAY_PLUGIN_FILE );
+		$script_asset_path = WCPAY_ABSPATH . 'dist/platform-checkout.asset.php';
+		$script_asset      = file_exists( $script_asset_path ) ? require_once $script_asset_path : [ 'dependencies' => [] ];
+
+		wp_register_style(
+			'WCPAY_PLATFORM_CHECKOUT',
+			$style_url,
+			[ 'wp-components' ],
```
harriswong, line 34: If wc-component is what we used, then this dependency will be different as well.

harriswong (commented):
I tested it on my Chrome, mobile view. LGTM so far! I have 2 comments

- Creating a follow-up to add the actual API call for the `usePlatformCheckoutUser()` custom hook
- Question regarding CSS and WooCommerce components

client/components/platform-checkout/save-user/about-platform-checkout.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+import interpolateComponents from 'interpolate-components';
+
+const AboutPlatformCheckout = () => {
+	return (
+		<div className="about-platform-checkout">
+			{ interpolateComponents( {
+				mixedString: __(
+					'Enter your phone number to save your checkout information. You’ll get {{strong}}secure single-click checkouts{{/strong}} here, and at 1,000s of other stores using Platform Checkout.',
```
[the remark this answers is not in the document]
Mayisha: Good catch. This happened because I directly copied the whole paragraph from Figma. Fixing this :smile:

client/components/platform-checkout/hooks/use-platform-checkout-user.js
Mayisha: Creating a follow-up ticket for this sounds good to me. :+1:

client/components/platform-checkout/save-user/style.scss
Mayisha: As per our discussion, skipping using the `wc-components` for this section as it does not have the required components. To achieve more consistency with the other part of the page using an existing `class` named `place-order` in this section.

harriswong (approved):
The last 2 commits look good. Tested again on desktop and mobile. LGTM!

PR #3683, opened 2022-01-20
Fix disable unavailable payment methods returning associative array

ismaeldcom:

After using `array_intersect` we could end with an associative array, so we need to use `array_values` to ensure it returns a plain array. Otherwise, we could end up having an error in the JS side trying to run `.map()` on an object.

#### Testing instructions
- Look for `COUNTRIES_WITH_GIROPAY` on the server and remove your Stripe's account country from it.
- Clear the account cache from WCPay Dev
- Go to **Payments > Settings**
- You shouldn't see any error notice in the payments methods section.

-------------------


dmallory42 (approved):
LGTM and works as expected! :shipit:

PR #3692, opened 2022-01-21
Notes - do not add notes on admin ajax

marcinbot:
Improves the site performance a bit

- Adds a check and early return for `wp_doing_ajax` before attempting to add any notes

#### Testing instructions

* Add a debug breakpoint after the early return added here, for example line 726
* Load any admin page
* The breakpoint should still be hit, ideally only once

-------------------


jrodger:
This tests as expected 👍

A comment above the check would be good, just to explain why we're doing it for future reference.

I assume we're adding the notes on non-ajax requests because of the 2 we'd expect there to be fewer of them? It doesn't actually matter when we add the notes exactly right? They're just written to the DB and read back when someone visits the dashboard?

jrodger (approved):
Thanks for adding the comments, this looks good to go.

Just need to be mindful of the changelog entries going under the correct version header since there has been a release.

PR #3699, opened 2022-01-24
Phone number input with country code

Mayisha:

Description: 
Creates a phone number input field with country selector. The input field is implemented using the JS plugging `intl-tel-input` to show country dropdown and country code selection.

### Testing instructions
- Enable `_wcpay_feature_platform_checkout`
- Set short code version of the checkout page
- Return from `isRegisteredUser=false` from `client/components/platform-checkout/hooks/use-platform-checkout-user.js`
- Select card payment method
- Select the checkbox
- Phone number input field is present and supports country codes

phone number

-------------------

### DIFF
--- a/client/components/platform-checkout/hooks/use-phone-number-validity.js
+++ b/client/components/platform-checkout/hooks/use-phone-number-validity.js
@@ -1,0 +1,33 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useState } from 'react';
+
+// hook for checking the validity of phone number.
+const usePhoneNumberValidity = ( element, phoneNumber ) => {
+	const [ isValidNumber, setIsValidNumber ] = useState( false );
+
+	useEffect( () => {
+		const checkValidity = () => {
+			const instance = window.intlTelInputGlobals.getInstance( element );
+			setIsValidNumber( instance.isValidNumber() );
+		};
+
+		if ( element ) {
+			checkValidity();
+			element.addEventListener( 'countrychange', checkValidity );
+		}
+
+		return () => {
+			if ( element ) {
+				element.removeEventListener( 'countrychange', checkValidity );
+			}
+		};
+	}, [ element, phoneNumber ] );
+
+	return {
+		isValidNumber,
+	};
+};
+
+export default usePhoneNumberValidity;
--- a/client/components/platform-checkout/save-user/phone-number-input.js
+++ b/client/components/platform-checkout/save-user/phone-number-input.js
@@ -1,0 +1,70 @@
+/**
+ * External dependencies
+ */
+import React, { useEffect, useState } from 'react';
+import { TextControl } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import intlTelInput from 'intl-tel-input';
+
+/**
+ * Internal dependencies
+ */
+// eslint-disable-next-line import/no-unresolved
+import utils from 'iti/utils';
+
+const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
+	const [ inputValue, setInputValue ] = useState( '' );
+	const [ inputInstance, setInputInstance ] = useState( null );
+
+	const handlePhoneNumberInputChange = ( value = inputValue ) => {
+		setInputValue( value );
+		if ( inputInstance ) {
+			handlePhoneNumberChange( inputInstance.getNumber() );
+		}
+	};
+
+	useEffect( () => {
+		let iti = null;
+		const input = document.querySelector( '#phone-number' );
+
+		const handleCountryChange = () => {
+			handlePhoneNumberChange( iti.getNumber() );
+		};
+
+		if ( input ) {
+			iti = intlTelInput( input, {
+				initialCountry: 'US',
+				customPlaceholder: function ( selectedCountryPlaceholder ) {
+					return selectedCountryPlaceholder;
+				},
+				separateDialCode: true,
+				utilsScript: utils,
+			} );
+			setInputInstance( iti );
+
+			input.addEventListener( 'countrychange', handleCountryChange );
+		}
+
+		return () => {
+			if ( iti ) {
+				iti.destroy();
+				input.removeEventListener(
+					'countrychange',
+					handleCountryChange
+				);
+			}
+		};
+	}, [ handlePhoneNumberChange ] );
+
+	return (
+		<TextControl
+			type="tel"
+			id="phone-number"
+			label={ __( 'Mobile phone number', 'woocommerce-payments' ) }
+			value={ inputValue }
+			onChange={ handlePhoneNumberInputChange }
+		/>
+	);
+};
+
+export default PhoneNumberInput;
--- a/client/components/platform-checkout/save-user/phone-number-input.js
+++ b/client/components/platform-checkout/save-user/phone-number-input.js
@@ -25,7 +25,12 @@ const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
 
 	useEffect( () => {
 		let iti = null;
-		const input = document.querySelector( '#phone-number' );
+		const input = document.querySelector(
+			`input[aria-label="${ __(
+				'Mobile phone number',
+				'woocommerce-payments'
+			) }"]`
+		);
 
 		const handleCountryChange = () => {
 			handlePhoneNumberChange( iti.getNumber() );
@@ -59,7 +64,7 @@ const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
 	return (
 		<TextControl
 			type="tel"
-			id="phone-number"
+			aria-label={ __( 'Mobile phone number', 'woocommerce-payments' ) }
 			label={ __( 'Mobile phone number', 'woocommerce-payments' ) }
 			value={ inputValue }
 			onChange={ handlePhoneNumberInputChange }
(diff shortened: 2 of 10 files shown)


client/components/platform-checkout/save-user/test/phone-number-input.test.js
```diff
@@ -0,0 +1,51 @@
...
+		);
+
+		expect( handlePhoneNumberChangeMock ).not.toHaveBeenCalled();
+
+		const input = screen.queryByPlaceholderText( '201-555-0123' ); // placeholder number for us
+		fireEvent.change( input, { target: { value: '201' } } );
+
+		expect( handlePhoneNumberChangeMock ).toHaveBeenCalledWith( '+1201' );
+	} );
+} );
```
Mayisha, line 51: Probably need to mock the calls.

client/components/platform-checkout/save-user/style.scss
```diff
@@ -70,6 +71,55 @@
 	}
 }
 
+#phone-number {
+	height: 42px;
+	background-color: #f2f2f2;
```
FangedParakeet, line 76: I'm not sure if this is even something worth moaning about here, but on a dark background, perhaps with dark mode, this section looks a bit weird. Here's what it looks like to me: I think this is still the standard Storefront, with perhaps a little bit of theme tinkering. 

Dark illegible inputs
_Text illegible with dark background styling_

Here it is if we inherit the background colour on the parent `.save-details-form`.

Legible text fields
_Legible inputs_

I'm not sure if this is worth worrying about in this PR, but I noticed it, so felt like mentioning it.

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -0,0 +1,70 @@
...
+		const handleCountryChange = () => {
+			handlePhoneNumberChange( iti.getNumber() );
+		};
+
+		if ( input ) {
+			iti = intlTelInput( input, {
+				initialCountry: 'US',
+				customPlaceholder: function ( selectedCountryPlaceholder ) {
+					return selectedCountryPlaceholder;
+				},
```
FangedParakeet, line 39: Nice! This is a handy setting. 👍

client/components/platform-checkout/save-user/test/phone-number-input.test.js
FangedParakeet: Did you mean to mock the `intl-tel-input` package? I think this is fine; I think testing some of these visual features somewhat necessitates having the actual package present. That being said, I don't possess much mastery over these front-end tests, so I don't know how feasible mocking out something like that is in the first place. 🤷

FangedParakeet (commented):
Sorry for taking so long to get around to reviewing this one! I think, for the most part, this component looks great and it seems like this mysterious `intl-tel-input` will satisfy our needs quite suitably.

I didn't see any validation on the input and it didn't appear as if anything is actually consuming the `phoneNumber` value; but I'm presuming that might be in another PR somewhere (?), since you're looking to merge this branch into another before diving into `develop`.

Anyways, I don't have too many complaints here, but just left a few comments: if you could just add some clarifications to my questions, I'll likely gladly approve and get this moving to where it needs to be. Cheers!

FangedParakeet:
Compatability tests failing sounds a little sus; I'd consider re-running them, to see if any magic fixes them. But I can get that JS test to fail pretty regularly on my local. I haven't investigated it too thoroughly, but from a cursory glance, it looks like that test should pass...

client/components/platform-checkout/save-user/test/phone-number-input.test.js
Mayisha: My comment was a bit unclear. Some tests were failing in the parent component tests because of `intl-tel-input` so I thought I would need to mock it there. But `npm install` fixed that actually. So we are good without the mock :smile:

client/components/platform-checkout/hooks/use-phone-number-validity.js
```diff
@@ -0,0 +1,33 @@
...
+			}
+		};
+	}, [ element, phoneNumber ] );
+
+	return {
+		isValidNumber,
+	};
+};
+
+export default usePhoneNumberValidity;
```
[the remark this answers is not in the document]
Mayisha: This is not used anywhere yet. This needs to be hooked to the `Place Order` button. When that button is clicked other form fields are validated. 
This will be handled with the issue when we will combine this UI with the backend API to create the user in `woopay` platform. Created this hook in advance to save some work later :smile:

client/components/platform-checkout/save-user/style.scss
Mayisha: Addressed this one by removing the bg and using an existing class which is originally applied to the `place order` section.

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -0,0 +1,70 @@
...
+					handleCountryChange
+				);
+			}
+		};
+	}, [ handlePhoneNumberChange ] );
+
+	return (
+		<TextControl
+			type="tel"
+			id="phone-number"
```
FangedParakeet, line 62:
```suggestion
			aria-label={ __( 'Mobile phone number', 'woocommerce-payments' ) }
```
I think I figured out why that one JS test (`should render the save user form when checkbox is checked`) keeps failing! The input is there, but Jest is unable to locate it.

In the test, we use `screen.queryByLabelText( 'Mobile phone number' )` to find the rendered phone number input. This finds the label containing the described text and then locates its associated input. [Here](https://testing-library.com/docs/queries/bylabeltext) are the patterns that are used to connect label to input. However, `intl-tel-input` appears to do things a little differently, when we feed it our phone input. Here's what it looks like in the HTML source.

Phone number HTML
_`intl-tel-input` rendered HTML_

Note that our label has no `id` property, its `for` property does not match the `id` nor `name` on the input, and the input is sitting outside the label: none of the patterns match what Jest is scanning for, so `screen.queryByLabelText( 'Mobile phone number' )` returns `null`!

I added the property `aria-label` to the input (as you can see in the screenshot above) and now Jest can find the input and the test passes at last! Try it out and let me know if that works for you too.

This does raise another issue that perhaps you have some more thoughts on: the mobile phone number label does not really work as an HTML label at the moment. Its `for` property does not match the `name` nor `id` on the input, so when we click on the label, the phone number input isn't automatically focussed on, as one might expect. 

If we look at the [source code for this `TextControl` component](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/text-control/index.js#L44), it seems that the `id` property which becomes the label's `id` is generated dynamically, whereas the `id` property for the input can be overwritten by our own properties on the component. The only way to keep the `id` on the label and the input the same is to remove the `id` field from the list of properties when we render the `TextControl` element and allow both ids to be generated dynamically.

But if we do that, now we have no selector for the input to feed to the initial `intlTelInput` function to create our magic phone input! I used the below as a substitute selector--and it works--but it is a little ugly.

```js
const input = document.querySelector( `input[aria-label="${ __( 'Mobile phone number', 'woocommerce-payments' ) }"]` );
iti = intlTelInput( input, {
				initialCountry: 'US',
				customPlaceholder: function ( selectedCountryPlaceholder ) {
					return selectedCountryPlaceholder;
				},
				separateDialCode: true,
				utilsScript: utils,
			} );
```

Now the phone input renders correctly _and_ clicking on the label selects the input as well (and the tests pass)! Let me know your thoughts and if you have any smarter ideas for a cleaner solution here! We could add a `className` property to the `TextControl` element as well and then use something like `.className input` to select the input, but I'll leave it up to you to decipher to best solution to this puzzling predicament.

FangedParakeet (changes requested):
Pretty much everything here is working hunky dory and looking fine and dandy...However, I think I solved the mystery of why this one JS test keeps failing and I uncovered a few more puzzling things along the way--all of which I have left in a rather lengthy comment below!

Let me know your thoughts and once that's resolved, I'll gladly approve the rest. Cheers!

zmaglica:
@Mayisha Could you please update the status of the PR/task?

FangedParakeet:
Hey @zmaglica, I just reviewed this yesterday and found one little issue, which shouldn't take too long to resolve (I described it in some length in [a comment above](https://github.com/Automattic/woocommerce-payments/pull/3699#discussion_r799121868)). @Mayisha is AFK today, but we should probably be able to sort this one out and get it merged in sometime early next week.

client/components/platform-checkout/save-user/phone-number-input.js
Mayisha: Thanks for the suggestion @FangedParakeet :+1: 

I also figured out the first part that `int-tel-input` is changing the hierarchy of `label` and `input`. For that, I could no longer find the element with `labelText` and had to use the `test-id` instead. :eyes: 

The idea of `area-label` is a good finding and solves the issue. I also did not realize that adding an explicit `id` to the `input` field is detaching the `label` from the `input`. :woman_facepalming: 

I am updating the PR with your suggestions :tada:

Mayisha:
@FangedParakeet I have pushed 2 updates here

- In [this commit](https://github.com/Automattic/woocommerce-payments/pull/3699/commits/87b0ba1b219d650a0f7cd03e62e70e8740cbd6d1) addressed your feedback from [this comment](https://github.com/Automattic/woocommerce-payments/pull/3699/#discussion_r799121868) and used `aria-label` for the `input` field. Thanks for this idea :tada: 
- In [this commit](https://github.com/Automattic/woocommerce-payments/pull/3699/commits/8ca1b30f0871c52a66cddb748f82b7aa30fb8be3) added `name` to the `checkbox` and `phone number` field which will be used [in the backend](https://github.com/Automattic/woocommerce-payments-server/issues/1559) upon form submission.

FangedParakeet (approved):
Cheers for making those little changes! Tests are green, input looks great, and everything is working fine for me now. LGTM! 🚢

PR #3706, opened 2022-01-25
Update `get_enabled_payment_method_currencies` to support more payment methods

ismaeldcom:
Update`get_enabled_payment_method_currencies` to look for a key in `Payment_Method` enum.

With the inclusion of more payment methods like ACH, BECS and Bacs, we will have more classes where their id doesn't match the name. So rather than adding more conditions like the SEPA one, we could rely on the `Payment_Method` enum to find the key that matches the class name.

#### Testing instructions
Tests should pass.

-------------------


tpaksu (approved):
LGTM! Makes sense to have it from the PaymentMethod enum instead using clauses for separate payment methods. You can 🚢  it!

PR #3586, opened 2021-12-29
Remove the custom Translations Loader initialization

diegocurbelo:
Fixes #3493

#### Changes proposed in this Pull Request

This PR removes the custom translations loader used to merge all language pack JSON files into one and enqueue it with each plugin JS file.

#### Additional context
Before #3486 the language packs did not contain a JSON with the translations for each of the plugin enqueued JS files (webpack generated files in the dist folder), so we relied `WC_Payments_Translations_Loader` to merge all JSON files into one, and the inject that single JSON file each time a JS file was enqueued (using the filter `load_script_translation_file` to always resolve to that single file.

#### Testing instructions
1. Go to **Settings** and change the Site Language to any [supported languages](https://translate.wordpress.com/projects/woocommerce/woocommerce-payments)
2. Go to **Dashboard > Updates** and click **Update translations** (this will be shown in the selected language :P)
3. Go to **Payments > Overview** and verify that the strings are fully translated:

Screen Shot 2022-01-03 at 18 22 02

Screen Shot 2022-01-03 at 18 20 21

The strings from that page are JS only; they are not present in the `.PO` file, only in the `JSON` files. For example the string `Pending balance` is only present in the JSON file: `woocommerce-payments-es_ES-090b05cdc4c33b5b4d778c13de44d8ea.json`

PR #3475, opened 2021-12-06
HACK week: use new PHP 8.0/8.1 functions across codebase and add symfony polyfills to enable this

kalessil:
I'm not advocating merging this PR, rather suggesting to the Platform Team to review if this makes sense at all.

#### Changes proposed in this Pull Request

- add Symfony polyfills, which gracefully add the functions for PHP below the required versions
- migrate codebase to use the new functions

#### Testing instructions

- All CI checks shall pass

-------------------

- [ ] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

PR #3708, opened 2022-01-25
Revert Symfony polyfills

jrodger:
Reverts PR #3475 to restore PHP 7.0 compatibility

#### Details

* Reverts #3475
* Some conflicts were handled, most notably we no longer have one of the affected files thanks to https://github.com/Automattic/woocommerce-payments/pull/3586
* The changes in `phpcs-compat.xml.dist` don't seem completely related. Should we be reverting that as well or not?

#### Testing instructions
* Review changes against the original PR
* Smoke test the plugin on a couple of PHP versions (7.0 and something else)

-------------------


htdat (approved):
The revert looks good.

I manually tested with a PHP 7.0 Jurassic site with these flows: 

- Onboarding.
- Checkout with traditional way (only cards) and UPE.
- Checkout with a card to create disputes.
- Multi-currency activation.
- Admin screens: Overview, Transactions, Deposits, Disputes, and settings.

kalessil (approved):
LGTM!

kalessil:
@htdat : IIUC, we'll re-revert this once PHP 7.0 support is dropped and the change will be back.

jrodger:
There was a mistake in the revert of `composer.json`. I've re-applied the changes by manually removing the polyfill packages with Composer. This also ensures that the lock file is in a sensible state.

haszari:
Tagging in @mattallan to review this on behalf of Helix - to ensure subscriptions is not impacted by the revert (`subscriptions-core` dependency).

mattallan:
> to ensure subscriptions is not impacted by the revert (subscriptions-core dependency).

1.6.2 is the latest version of subscriptions-core and it doesn't look like this revert changes that, so this is all good from a WCPay + Subscriptions perspective :100:

ricardo (approved):
I was asked to give this another look too.

Changes LGTM.

Smoke tested checkout, transactions search, admin pages and multi-currency with PHP 7.0, 7.4 and 8.0 on Docker. Everything seems to work as expected.

However, on PHP 8.1, I see a deprecation warning on package `myclabs/php-enum`.

> Deprecated: Return type of MyCLabs\Enum\Enum::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in [file] on [line].

I also see a couple warnings on the latest version of WooCommerce, so I understand we're probably not there yet. I tried updating `myclabs/php-enum` to the latest version, but it didn't fix the notice.

I think this is out of the scope for this PR, so approving 👍.

Can someone else double-check the notice problem on their environment with PHP 8.1? Then I can open an issue.

Thanks for the fix @jrodger. Also props to @htdat for fixing the dev utils plugin for PHP 7.0 in the meanwhile. 46-gh-Automattic/woocommerce-payments-dev-tools 👏

PR #3714, opened 2022-01-26
Skip flaky subscription tests

dmvrtx:

Update E2E tests to skip flaky subscription tests.

Skipped tests have dependencies on components like Action Scheduler and there is no guarantee in the test environment that it won't be overloaded with other tasks, e.g. image regeneration. Hence, it is better to skip tests until we can find a way to create a "pure" environment without any background tasks.

#### Testing instructions

* E2E tests should pass on the branch while tests "Renew a subscription as a merchant" and "Renew a subscription via Action Scheduler" should be marked as skipped.

-------------------

**Post merge**


deepakpathania (approved):
Thanks for picking this up.  🙌 

Could you add a TODO/comment to the skipped block to unksip these after we figure out how to reliably test this flow? Changes LGTM otherwise, pre approving it.

dmvrtx:
> Could you add a TODO/comment to the skipped block to unksip these after we figure out how to reliably test this flow?

Done in c0d1e4b6

PR #3718, opened 2022-01-27
Fix checkout on block-based themes

RadoslavGeorgiev:

@vbelolapotkov discovered that checkout fields do not appear on the new default theme Twenty Twenty-Two, which comes bundled with the recently released WordPress 5.9. ref: p1643208857030300-slack-C0E1AV8T0

After further investigation we found out that:
1. Both the standard and UPE checkouts do not work with the classic checkout.
2. The issue is not specific to Twenty Twenty-Two, but all block-based themes. Another theme I tested with is called Gutena.
3. Even though there is no deprecation notice about it, localizing scripts within a shortcode (the classic checkout is a shortcode) does not work with those themes. This is evident in https://wordpress.org/support/topic/issue-with-wp_localize_script-or-similar-on-block-themes/ as well. This means that something changed with WP 5.9, and we did not test it.

The good news is that the fix is relatively simple: Instead of localizing scripts within the `payment_fields` method, we can add the action to `wp_footer`. This should not affect WCPay's functionality in any way, because the `wp_footer` hook is triggered by the `wp_footer()` function, the same function which forces `$wp_scripts` to output everything necessary.

#### Testing instructions

1. Update to WP 5.9 and install+activate the Twenty Twenty-Two theme or start with a fresh 5.9 install.
2. Make sure that WC and WCPay are set up.
3. Open the checkout page without this PR and confirm that Stripe's fields are not rendered. The checkout block should not be affected.
4. Check out this PR.
5. Verify that fields to appear and work both on the classic checkout, and in the checkout block.
6. In Dev Tools, disable UPE, and try again. Everything should keep on working.

I'm not covering this with tests, as it can be tested only with E2E tests, and we will discuss the 5.9 scenario separately.

-------------------


zmaglica (approved):
LGTM!

brucealdridge:
Have created an issue to close the loop on this.

closes #3718

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -731,8 +731,7 @@ public function payment_fields() {
 		try {
 			$display_tokenization = $this->supports( 'tokenization' ) && ( is_checkout() || is_add_payment_method_page() );
 
-			wp_localize_script( 'WCPAY_CHECKOUT', 'wcpay_config', $this->get_payment_fields_js_config() );
-			wp_enqueue_script( 'WCPAY_CHECKOUT' );
+			add_action( 'wp_footer', [ $this, 'enqueue_payment_scripts' ] );
```
haszari, line 734: Based on the comments on forum link (https://wordpress.org/support/topic/issue-with-wp_localize_script-or-similar-on-block-themes/), it looks like the recommended approach is to use `wp_add_inline_script` to enqueue script data (in preference to `wp_localize_script`).

@RadoslavGeorgiev should we be using `wp_add_inline_script` ?

PR #3719, opened 2022-01-27
Release 3.6.1

RadoslavGeorgiev:
## Changelog:

= 3.6.1 - 2022-xx-xx =
* Fix - Remove packages not compatible with PHP 7.0
* Security update.


naman03malhotra (approved):
LGTM!

includes/class-wc-payments-account.php
```diff
@@ -415,7 +415,7 @@ public function allowed_redirect_hosts( $hosts ) {
 	 * Handle onboarding (login/init/redirect) routes
 	 */
 	public function maybe_handle_onboarding() {
-		if ( ! is_admin() ) {
+		if ( ! is_admin() || ! current_user_can( 'manage_woocommerce' ) ) {
```
zmaglica, line 418: Are you certain that this will work? For example, the user with `shop_manager` role will not be able to perform onboarding because `is_admin()` will return false. I am not 100% sure about `is_admin()` function and what it will return in this scenario.

PR #3720, opened 2022-01-27
Limit the number of requests for rejected client

marcinbot:
Fixes 1539-gh-Automattic/woocommerce-payments-server

- When the account is rejected, only load the overview page with the status card

#### Testing instructions

- See the testing instructions in 1542-gh-Automattic/woocommerce-payments-server

-------------------

### DIFF
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -111,6 +111,28 @@ class WC_Payments_Account {
 	}
 
 	/**
+	 * Checks if the account has been rejected, assumes the value of $on_error on server error.
+	 *
+	 * @param bool $on_error Value to return on server error, defaults to false.
+	 *
+	 * @return bool True if the account is rejected, false otherwise, $on_error on error.
+	 */
+	public function is_account_rejected( bool $on_error = false ): bool {
+		try {
+			$account = $this->get_cached_account_data();
+
+			if ( empty( $account ) ) {
+				// Empty means no account, so not rejected.
+				return false;
+			}
+
+			return strpos( $account['status'], 'rejected' ) === 0;
+		} catch ( Exception $e ) {
+			return $on_error;
+		}
+	}
+
+	/**
 	 * Checks if the account is connected, throws on server error.
 	 *
 	 * @return bool      True if the account is connected, false otherwise.
--- a/client/overview/index.js
+++ b/client/overview/index.js
@@ -51,6 +51,7 @@ const OverviewPage = () => {
 		'1' === queryParams[ 'wcpay-connection-success' ];
 
 	const showLoginError = '1' === queryParams[ 'wcpay-login-error' ];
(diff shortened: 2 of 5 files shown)


dmallory42 (approved):
LGTM, tested the server PR in conjunction with this and it works well! 🚢

includes/class-wc-payments-account.php
```diff
@@ -110,6 +110,28 @@ public function is_stripe_connected( bool $on_error = false ): bool {
...
+		try {
+			$account = $this->get_cached_account_data();
+
+			if ( empty( $account ) ) {
+				// Empty means no account, so not rejected.
+				return false;
+			}
+
+			return strpos( $account['status'], 'rejected' ) === 0;
+		} catch ( Exception $e ) {
```
dechov, line 130: Is there any particular known way that an exception would be thrown?

marcinbot: Yes, it's either `Account_Not_Found_Exception` or `Account_Suspended_Exception`.

marcinbot: Whoops, sorry, I confused the PRs. Yeah, the API client should only throw `API_Exception` and `get_cached_account_data` should catch it.

This perhaps overly defensive pattern mirrors the existing [`is_stripe_connected`](https://github.com/Automattic/woocommerce-payments/blob/b304edab702e0606d7bb0532ccef38f3b09fcd98/includes/class-wc-payments-account.php#L105-L111). But I see that the `empty()` check in line 124 should return `$on_error` instead of false.

I'd still keep the try/check though with a generic Exception. The methods in this class are quite sensitive and called often at different points of the request lifecycles - if any of them causes a crash, then we risk the entire site crashing, not just the wcpay functionality.

marcinbot: Opened https://github.com/Automattic/woocommerce-payments/pull/3766 to fix

dechov: Hm, I see how it could be considered an error if the account doesn't exist but actually it's not obvious to me that it shouldn't be false, since the function name could possibly be interpreted as "is there a connected account that is rejected" – and that's also closer to what I intended with the name of a new function `is_account_onboarded` in https://github.com/Automattic/woocommerce-payments/pull/3767/commits/275c5077308da15e654cdd524beee907575cb4e5 (from a draft PR).

What do you think? If it does return an error, then the function comment should be expanded beyond "on server error".

marcinbot: I see, what threw me off was the multiple purposes of the empty() function, it will pass if the value is false or an empty array. I tweaked https://github.com/Automattic/woocommerce-payments/pull/3766 to fix it (no account still should be always false, it's not an error) and also to expand the comment.

What do you think?

dechov: Sounds good 👍

PR #3723, opened 2022-01-28
Add customer link in disputes list

### LINKED ISSUE #3698: Link to customer analytics from disputes list
### Description
We are adding `order` property containing `customer_url` both to Transactions and Disputes but only the transactions list table is taking advantage of `order.customer_url` to make the customer name clickable and link to the customers table. 

### Acceptance criteria
The customer field in the Disputes list table should be a link that points to the customer in the customers analytics table the same way how it's currently done for Transactions. 

### Dev notes
[Where we add `customer_url` to the response](https://github.com/Automattic/woocommerce-payments/blob/e33a0ea28d4e466e8443a9782f7788ad9105ad86/includes/wc-payment-api/class-wc-payments-api-client.php#L1822)
[How we do it for transactions](https://github.com/Automattic/woocommerce-payments/blob/develop/client/transactions/list/index.tsx#L242)
[Required changes for disputes](https://github.com/Automattic/woocommerce-payments/blob/develop/client/disputes/index.tsx#L187)

elazzabi:
Fixes #3698 

Adds a link to the customer (if available) in the disputes list.  

#### Testing instructions

* Open Disputes page
* Double check the user name is clickable if the `order.customer_url` is exposed for the dispute
* Double check the user name is not clickable, but still links to the dispute page, if `order.customer_link` is not exposed

-------------------

**Post merge**


elazzabi:
I'm not sure if this deserves its own testing plan in the QA wiki, or if it can be somehow added to the `Introduce sorting on Disputes page`

dpaun1985:
> I'm not sure if this deserves its own testing plan in the QA wiki, or if it can be somehow added to the `Introduce sorting on Disputes page`

I think we can add it later if needed

client/disputes/test/__snapshots__/index.tsx.snap
```diff
@@ -460,10 +460,8 @@ exports[`Disputes list renders correctly 1`] = `
                   class="woocommerce-table__item is-left-aligned"
                 >
                   <a
-                    class="woocommerce-table__clickable-cell"
                     data-link-type="wc-admin"
-                    href="admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes%2Fdetails&id=dp_asdfghjkl"
-                    tabindex="-1"
+                    href="https://shop.local"
```
anu-rock, line 464: Nitpick: It may help to have tests to check both scenarios - `customer_url` and no `customer_url`.

anu-rock (approved):
Short and sweet. Great work with your first PR 🎉 

I have left a small nitpick. Otherwise, LGTM.

client/disputes/test/__snapshots__/index.tsx.snap
elazzabi: @anu-rock I was thinking about this when working on the tests and found two solutions, but couldn't decide what's the best given I'm new to the code-base:

1. Add a test case here, without the `customer_url`. But this assumes any future changes will be added here, which will make the snapshot a little bigger.
2. Maybe expose a smaller component? Like DisputeRow or just Dispute in addition to the currently exposed DisputesList

Both solutions will work. The first is easier for now, but the second one is probably more useful for future changes 🤔 

Anyway, happy to go forward (part of this PR or a new PR) with whichever is the best option

anu-rock: @elazzabi Glad to know you considered all possibilities here 👍 

> Add a test case here, without the customer_url. But this assumes any future changes will be added here, which will make the snapshot a little bigger.

You are right in assuming that more changes will likely happen. Props will be added and removed in time. That said, I do not expect a lot of conditional props (like `customer_url`) to be added. In that sense, it may not hurt to fully cover this prop.

> Maybe expose a smaller component? Like DisputeRow or just Dispute in addition to the currently exposed DisputesList

Good thought. At this point, it may be a bit overkill.

Ultimately, it's just a nitpick and a pretty minor one. I'll leave the decision to you :) It's okay to skip it for now.

elazzabi: That sounds good! I pushed a fix here to add a new test case https://github.com/Automattic/woocommerce-payments/pull/3723/commits/10ee24e5c4ce63795d69ff9a7a417a8eb9b858d3

PR #3303, opened 2021-11-05
Add idempotency key to post requests

chickenn00dle:
Fixes #530 

#### Changes proposed in this Pull Request

This adds an idempotency key to all post requests sent to the server. This also adds request headers to logs so the idempotency key is useful for debugging.

#### Testing instructions

- Trigger a post request by doing something like completing checkout
- Confirm idempotency key is present in headers. Be sure to also check logs.
- Trigger a get or delete request
- Confirm idempotency key is not present in headers. Be sure to also check logs.

-------------------

- [x] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Added testing instructions to the [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) (or does not apply)

Issue #2970, opened 2021-09-22
CI: Fix unit tests for PHP 7.0


vbelolapotkov:
Follow up from #2945 

### Describe the bug

There are issues with loading dependencies in unit tests for combination of WC 4.4, WP 5.4, PHP 7.0.

### To Reproduce

1. Update compatibility.yml file to run tests with combination above.
2. Push PR.
3. See tests are failing.

### Actual behavior

Tests fails see failed checks in #2945 e.g. [this one](https://github.com/Automattic/woocommerce-payments/runs/3675736404)

### Expected behavior

Ideally we should be able to run tests with it as well.

### Additional context

Two ways to fix it:

1. Add imports manually to bootstrap.
2. Add some autoloader for PHPUnit 6.5.14 (not sure if there is a solution for that).

Slack thread: p1632307719018700/1632161800.465400-slack-CGGCLBN58

jrodger:
Since WooCommerce will be increasing the [minimum version of PHP to 7.2](https://developer.woocommerce.com/2022/01/05/new-requirement-for-woocommerce-6-5-php-7-2/), we're not going to spend time getting this going in our CI pipeline.

We have added to our testing notes so that we do at least some of our manual testing on the minimum version of PHP.

PR #3726, opened 2022-01-28
Fix PHPUnit errors on WP 59 and PHP 8.*

### LINKED ISSUE #3712: PHPUnit: Errors for WP 5.9 and PHP 8.0, 8.1
### Describe the bug

Since WP 5.9 is released, these CI tests are failing: 

- Environment - WC beta (beta, latest, latest, 8.0)
- Environment - WC beta (beta, latest, latest, 8.1)

Error: 

> WCPay_Multi_Currency_User_Settings_Tests::test_save_presentment_currency
> ParseError: syntax error, unexpected token “match”, expecting variable
> /home/runner/work/woocommerce-payments/woocommerce-payments/tests/unit/multi-currency/test-class-user-settings.php:36

p1643170695291800-slack-CGGCLBN58

### To Reproduce

Basically, all CI tests are failing. 

### Expected behavior

Expect tests to continue to run.

### Additional context

Core 5.9 includes some changes to PHP8.0 and 8.1 w/ PHPUni: 

- https://make.wordpress.org/core/2022/01/10/wordpress-5-9-field-guide/
- https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/#impact-plugin-theme-integration-tests

Seems like this issue https://github.com/sebastianbergmann/phpunit/pull/4374

Based on this chart, we need to use PHPUnit 9 for WP 5.9
https://make.wordpress.org/core/handbook/references/phpunit-compatibility-and-wordpress-versions/

_Update:_ this change may be relevant https://github.com/Automattic/woocommerce-payments/pull/3475

zmaglica:
Fixes #3712

Update PHPUnit tests and dependencies so pipelines for WP 5.9, PHP version 8.0 and 8.1 doesn't break. 

The current PHPUnit version that we use doesn't support PHP 8.*. PHPUnit version that supports PHP version 8.* doesn't support PHP version 7.0. To make this work, the CI/CD pipeline will detect PHP current runtime PHP version and if the current PHP version is 8, the script will download PHPUnit 9 from the official website and use that to run unit tests. The problem with new version of PHPUnit is that countless functions has been deprecated and removed from the version we currently use.

To resolve this issue, the [PHPUnit polyfills](https://github.com/Yoast/PHPUnit-Polyfills) have been added to the project. The default WP_TestCase has been updated to use TestCase from polyfill, and multiple tests have been fixed by replaced deprecated functions with the ones that are not. Many small warnings and errors has been fixed too.

#### Testing instructions

Make sure that tests pass on all PHP and supported WordPress versions.

-------------------


bin/phpunit.sh
```diff
@@ -4,4 +4,14 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
 	exit
 fi
 
-if [ -f "phpunit.phar" ]; then php phpunit.phar -c phpunit.xml.dist $@; else ./vendor/bin/phpunit $@; fi;
+CURRENT_PHP_MAJOR_VERSION=$(php -r 'echo PHP_MAJOR_VERSION;')
+
+if [ $CURRENT_PHP_MAJOR_VERSION -gt 7 ]; then
```
deepakpathania, line 9: Should we get the hardcoded version numbers from some constant so that they're easier to update for any future releases where we might have to revisit this?

composer.json
```diff
@@ -39,7 +39,7 @@
       "php-stubs/wordpress-stubs": "5.8.2",
       "php-stubs/woocommerce-stubs": "5.9.0",
       "rregeer/phpunit-coverage-check": "0.3.1",
-      "yoast/phpunit-polyfills": "1.0.1",
+      "yoast/phpunit-polyfills": "^1.0",
```
deepakpathania, line 42: Since we use hard coded versions for the rest of the deps, is there a special reason to use caret here?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1597,8 +1597,7 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
 			$url          .= '?' . http_build_query( $params );
 			$redacted_url .= '?' . http_build_query( $redacted_params );
 		} else {
-			$headers['Idempotency-Key'] = $this->uuid();
```
deepakpathania, line 1600: Why have we removed sending this in headers?

includes/wc-payment-api/class-wc-payments-http.php
```diff
@@ -71,6 +71,15 @@ public function remote_request( $args, $body = null, $is_site_specific = true, $
 			$args['url'] = implode( '?', $url );
 		}
 
+		// Generate Idempotency key for PATCH, POST and PUT requests.
+		if (
```
deepakpathania, line 75: This check for request methods for adding the idempotency key is new, right? Maybe we could only refactor in this PR without any functionality changes to isolate any side effects?

tests/unit/multi-currency/compatibility/test-class-woocommerce-product-add-ons.php
```diff
@@ -119,12 +119,12 @@ public function test_get_addons_price_returns_percentage_without_conversion() {
 	}
 
 	public function test_get_addons_price_returns_converted_price() {
-		$this->mock_multi_currency->method( 'get_price' )->with( 50, 'product' )->willReturn( 75 );
-		$this->assertEquals( 75, $this->woocommerce_product_add_ons->get_addons_price( 50, [ 'price_type' => 'flat_fee' ] ) );
+		$this->mock_multi_currency->method( 'get_price' )->with( 50.00, 'product' )->willReturn( 75.00 );
```
deepakpathania, line 122: Have we made any changes that lead to the precision of numbers here? We should ideally get the tests to work with the older values themselves for a refactor IMO. Might be wrong here in case this is a new quirk with the version bump itself so please CMIIAW.

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -191,7 +192,16 @@ public function test_create_setup_intent() {
 			->expects( $this->once() )
 			->method( 'remote_request' )
 			->with(
-				$this->contains( 'https://public-api.wordpress.com/wpcom/v2/sites/%s/wcpay/setup_intents' ),
+				[
```
deepakpathania, line 195: This might be bulky to do for any new tests we add, any way to have the same URL checking functionality only like the original test?

deepakpathania (dismissed):
Thanks for all the great work @zmaglica, left some minor comments requesting a few clarifications, looks good overall. 🙌

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: It was hard to test it because this parameter is unique and mocking it is extremely hard. The point is that in unit tests, the `containts()` is used, but the same function was removed from PHPUnit 8 (or 9) which makes pipelines fails on PHP 8.* versions. To overcome this issue, the header is removed inside HTTP Client library

includes/wc-payment-api/class-wc-payments-http.php
zmaglica: Yes, this checks for idempotency key and if it doesn't exist, it will create a new one.
The problem is that the test will fail on PHP 8 version if we leave UUID generation as it was. The only solution I found so far is to remove that logic from payments api client to wc payments HTTP client because idempotency key is required  there.

deepakpathania: Makes sense but do we need the check for specific request methods here for this to work? I don't see a similar check in the existing flow so maybe we could move it here without adding this conditional check.

tests/unit/multi-currency/compatibility/test-class-woocommerce-product-add-ons.php
zmaglica: You are right. I will fix it.

includes/wc-payment-api/class-wc-payments-http.php
zmaglica: I followed the original concept and added check for specific request methods, as can be seen here:

https://github.com/Automattic/woocommerce-payments/blob/0fd92cf70a3d02e24f6313ad73fef26fa336fc9b/includes/wc-payment-api/class-wc-payments-api-client.php#L1596

deepakpathania: Ah, my bad, missed that somehow. 😅

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
zmaglica: That `contains()` was added [here](https://github.com/Automattic/woocommerce-payments/pull/3303/files ) because of idempotency key. The creation of key was part of the mocked class, and it was impossible to recreate mocking with a value that is always random.

If you open the mentioned commit and compare it with changes from this one, you will see that they are the same. Since includes() function is no longer needed, the test worked before, should work now too.

includes/wc-payment-api/class-wc-payments-api-client.php
chickenn00dle: If I recall correctly, the reason we added the Idempotency Key to GET, DELETE requests was to display in logs and potentially help when debugging. If we move this into the `remote_request` method, I believe we lose this benefit, making the key sort of pointless.

> The point is that in unit tests, the containts() is used, but the same function was removed from PHPUnit 8 (or 9) which makes pipelines fails on PHP 8.* versions.

I haven't looked too far into this, but couldn't we use [the `callback` constraint](https://phpunit.readthedocs.io/en/9.5/test-doubles.html#test-doubles-mock-objects-examples-subjecttest3-php) here to deal with the randomness of the Idempotency-Key header?

bin/phpunit.sh
htdat: I think it's fine to keep it as-is as this is not a frequent change. But it's worth adding a comment to clarify why we have the if-else condition here.

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -5,6 +5,7 @@
  * @package WooCommerce\Payments\Tests
  */
 
+use PHPUnit\Framework\Constraint\TraversableContains;
```
htdat, line 8: This class is not used in this file? So I think we can remove it.

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: > If I recall correctly, the reason we added the Idempotency Key to GET, DELETE requests was to display in logs and potentially help when debugging. If we move this into the remote_request method, I believe we lose this benefit, making the key sort of pointless.

This makes sense. However, since this is critical, I think we can create another issue to add up tests for it to avoid other devs to refactor it without knowing this knowledge. 

> I haven't looked too far into this, but couldn't we use the callback constraint here to deal with the randomness of the Idempotency-Key header?

Yeah, this seems a reasonable solution.

htdat (commented):
I would say an awesome work here to figure out the fix for this issue. 

Though it's not mentioned in the description, this PR follows up the suggestion from WP core. 

https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/#impact-plugin-theme-integration-tests

Also, @deepakpathania's review is spot on, and I only contribute some thoughts into this current comments!

bin/phpunit.sh
zmaglica: This is fixed and refactored.

composer.json
zmaglica: Fixed.

tests/unit/multi-currency/compatibility/test-class-woocommerce-product-add-ons.php
zmaglica: Fixed.

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: @chickenn00dle I manage to fix the issue. Could you please verify it?

changelog/fix-phpunit-errors-on-wp59-php8
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Unit test support for PHP 8 and upgrade to the lastest PHPUnit version
```
htdat, line 4: Nitpick: May clarify PHPUnit version here?

.github/workflows/compatibility.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           - woocommerce: '4.8.0'
             wordpress:   '5.7'
             gutenberg:   '11.4.0' # The latest version supporting WP 5.6.
-            php:         '7.1' # TODO This should be 7.0. Pending to fix in issue #2970.
+            php:         '7.0' # Minimum supported PHP version
```
htdat, line 24: Could you clean up the other line in this file too? 

https://github.com/Automattic/woocommerce-payments/blob/cf4f777e16c146c18bc5823e9dae76474c67fe60/.github/workflows/compatibility.yml#L43-L43

bin/phpunit.sh
```diff
@@ -4,4 +4,21 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
 	exit
 fi
 
-if [ -f "phpunit.phar" ]; then php phpunit.phar -c phpunit.xml.dist $@; else ./vendor/bin/phpunit $@; fi;
+CURRENT_PHP_MAJOR_VERSION=$(php -r 'echo PHP_MAJOR_VERSION;')
+CURRENT_PHP_MINOR_VERSION=$(php -r 'echo PHP_MINOR_VERSION;')
+
+# The PHPUnit version inside composer.json is not compatible with PHP 7.2 and above
```
htdat, line 10: You updated value `config.platform.php` in composer.json from `7.1` to `7.3` so this comment is no longer correct?

bin/phpunit.sh
```diff
@@ -4,4 +4,21 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
...
+SUPPORTED_PHP_MAJOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON=7
+SUPPORTED_PHP_MINOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON=3
+
+if [ $CURRENT_PHP_MAJOR_VERSION -gt $SUPPORTED_PHP_MAJOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON ]; then
+	./vendor/bin/phpunit $@;
+else
+	if [ $CURRENT_PHP_MINOR_VERSION -ge $SUPPORTED_PHP_MINOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON ]; then
+		./bin/phpunit6 -c phpunit.xml.dist $@
+	else
+    	chmod +x ./bin/phpunit6
```
htdat, line 21: I may miss something here. This means no tests are executed for PHP < 7.3?

Also, the following runs are only show `Running the tests.` then finish without any test. I think that is due to this command.

https://github.com/Automattic/woocommerce-payments/runs/5350917490?check_suite_focus=true#step:5:168

https://github.com/Automattic/woocommerce-payments/runs/5350917508?check_suite_focus=true#step:5:168

https://github.com/Automattic/woocommerce-payments/runs/5350917927?check_suite_focus=true#step:5:168

bin/phpunit.sh
```diff
@@ -4,4 +4,21 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
...
+CURRENT_PHP_MAJOR_VERSION=$(php -r 'echo PHP_MAJOR_VERSION;')
+CURRENT_PHP_MINOR_VERSION=$(php -r 'echo PHP_MINOR_VERSION;')
+
+# The PHPUnit version inside composer.json is not compatible with PHP 7.2 and above
+# Update this constant if you wish to bump supported PHP major version
+SUPPORTED_PHP_MAJOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON=7
+SUPPORTED_PHP_MINOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON=3
+
+if [ $CURRENT_PHP_MAJOR_VERSION -gt $SUPPORTED_PHP_MAJOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON ]; then
+	./vendor/bin/phpunit $@;
```
htdat, line 16: `$@` should be bound in double quotes `"$@"`? PHPStorm gives this reference https://github.com/koalaman/shellcheck/wiki/SC2068 

Nitpick: also specifying the configuration file may be better. 


```suggestion
	./vendor/bin/phpunit -c phpunit.xml.dist "$@";
```

bin/phpunit.sh
```diff
@@ -4,4 +4,21 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
...
+else
+	if [ $CURRENT_PHP_MINOR_VERSION -ge $SUPPORTED_PHP_MINOR_VERSION_FOR_PHPUNIT_INSTALLED_VIA_COMPOSER_JSON ]; then
+		./bin/phpunit6 -c phpunit.xml.dist $@
```
htdat, line 19: ^ Similar above regarding the double quotes: 

```suggestion
		./bin/phpunit6 -c phpunit.xml.dist "$@"
```

htdat (commented):
The label for this PR is still `pr: needs author reply` but I see you've added more commits so I've had another look with some comments. 

As I see, you're attempting to fix this issue too #2970? If so, may update the PR description too as it seems a bit out-of-date.

bin/phpunit.sh
zmaglica: I needed to bump the version because of the latest PHPUnit because the latest one doesn't support version below 7.3. If I didn't update this, you would receive an error that you couldn't install PHPUnit because of PHP version. But again, there is a different approach that I can use. I will push new commit to solve this.

bin/phpunit.sh
zmaglica: This is fixed. Thank you for finding this.

changelog/fix-phpunit-errors-on-wp59-php8
zmaglica: Fixed

.github/workflows/compatibility.yml
zmaglica: Fixed.

deepakpathania:
Decided to ditch support for PHP 7.0, we will open a separate issue to support if needed. @zmaglica

bin/phpunit.sh
htdat: Thanks for the explanation 👍 !

htdat (approved):
Awesome work!

✅ LGTM, we can 🚢 it after merging the latest changes from `develop` branch (and maybe fixing other PHPUnit test files).  

After that, it's worth announcing this on P2 about what developers should be aware when writing tests, and we can safely use PHPUnit 9 features 💯

PR #3731, opened 2022-01-31
Show Currency filter in Disputes page

### LINKED ISSUE #3697: Disputes: filtering by currency is missing
By having disputes in USD/JPY I expect a currency filter to be available, unfortunately, it is not.

elazzabi:
Fixes #3697 

Make it possible to filter by currency in the Disputes list. All the work was already there, it only needed few small fixes to turn it on.

Before:
Screen Shot 2022-01-31 at 14 40 55

After:
Screen Shot 2022-01-31 at 14 35 46

#### Testing instructions

* Navigate to Disputes list
* Make sure there are different disputes with different currencies
* Double check the currencies displayed on the filter are for those in the disputes only (ie, if disputes are in USD/JYP, filter should only display those currencies and not EUR for example)
* Filter by a currency
* Double-check disputes list contains disputes in the selected currency
* Refresh your page and double-check the filter is still there (it's using a query param, so it should be there)
* Double-check filter works with advanced filters (like status, and date)

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : [Testing currency filter in disputes page](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.7.0/a3ebbb17ebc8dca11df23bfbac145b7bb096d3ac)


KarlisJ:
LGTM! 

Also tested and it works as expected. 

The only comment would be if tests cover the case for "All currencies"? But I'm not 100% sure if it's required to fully cover the component.

client/disputes/filters/config.ts
```diff
@@ -47,7 +47,7 @@ export const filters: [ DisputesFilterType, DisputesFilterType ] = [
 			'date_after',
 			'date_between',
 		],
-		showFilters: () => false,
+		showFilters: () => true,
```
kalessil, line 50: It was intentionally false. [Here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/deposits/filters/index.js#L28) we enable them based on data fetched from the server side. The intention is to ensure UI elements are not playing hide and show game.

kalessil (changes requested):
LGTM and tests well! I left one comment though, which we need to address before emerging the PR.

client/disputes/filters/config.ts
elazzabi: Aha! The filter wasn't showing at first, but I believe it's because I didn't have disputes in different currencies. Thank you for the catch! Fixed: https://github.com/Automattic/woocommerce-payments/pull/3731/commits/3594571566a45d8e7b19224845c02b1c66b3235a

elazzabi:
> LGTM!
> 
> Also tested and it works as expected.
> 
> The only comment would be if tests cover the case for "All currencies"? But I'm not 100% sure if it's required to fully cover the component.

Thanks for the review @KarlisJ !

Hmm I'm not sure either. I can add another test to catch that, better have extra tests than no test :)

client/disputes/filters/config.ts
kalessil: Yep, thanks!

kalessil (approved):

elazzabi:
@KarlisJ added a default case test here https://github.com/Automattic/woocommerce-payments/pull/3731/commits/019e0db6a65dd1909b6906b51e3d0526f4f6b97e

mgascam (approved):
LGTM!

PR #3736, opened 2022-02-01
Load iframe if correct email is entered

superdav42:
Opens iframe on merchant site for platform checkout.

Test:
Enter valid email address. I frame should popup.
Should work on desktop mobile. 
Back button should close. Clicking outside of iframe should close. Esc key should close.
Finishing the process should login and redirect.


superdav42:
This is all done but I can't get the styles to load in the block checkout. I'll look again later. It seems they only are set to load if UPE is enabled but I think that would have other problems. I need to investigate.

client/checkout/utils/platform-checkout.js
```diff
@@ -6,13 +13,69 @@ export const handlePlatformCheckoutEmailInput = ( field ) => {
...
+		iframe.src =
+			getConfig( 'platformCheckoutHost' ) + '/sms-otp/?email=' + email;
+		parentDiv.insertBefore( iframeWrapper, null );
+		iframe.focus();
+	};
+
+	document.addEventListener( 'keyup', ( event ) => {
+		const key = event.which || event.keyCode;
+
+		if ( 27 === key && closeIframe() ) {
```
brettshumaker, line 57: `event.which` and `event.keyCode` are [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which) and they suggest using `event.key` or `event.code` instead.

```suggestion
		if ( 'Escape' === event.key && closeIframe() ) {
```

client/checkout/utils/platform-checkout.js
```diff
@@ -6,13 +13,69 @@ export const handlePlatformCheckoutEmailInput = ( field ) => {
...
+				if ( data[ 'user-exists' ] ) {
+					openIframe( email );
+				}
+			} )
+			.finally( () => {
+				spinner.remove();
+			} );
 	};
 
 	const validateEmail = ( value ) => {
```
brettshumaker, line 81: I know it's not part of this PR, but the `input.checkValidity()` is not working properly for me. It's returning `true` for any value containing at least one character, followed by a `@`,  followed by at least one character. So the value `a@a` returns `true`. This is causing the `platformCheckoutLocateUser()` function to be called without a proper email address which results in a `400 Bad Request`. Ideally, we'd validate the email address before making that request.

FWIW, it's not working [outside of the codebase](https://jsfiddle.net/e9um57d0/) either.

client/components/platform-checkout/iframe.tsx
```diff
@@ -0,0 +1,34 @@
+
```
brettshumaker, line 1: Was this file meant to be included?

client/checkout/utils/platform-checkout.js
```diff
@@ -6,13 +13,69 @@ export const handlePlatformCheckoutEmailInput = ( field ) => {
...
+			closeIframe();
+		}
+	} );
+
+	const openIframe = ( email ) => {
+		iframe.src =
+			getConfig( 'platformCheckoutHost' ) + '/sms-otp/?email=' + email;
+		parentDiv.insertBefore( iframeWrapper, null );
+		iframe.focus();
+	};
```
brettshumaker, line 52: We should add `overflow: hidden` to `body` (or add a class that achieves the same) when we open the iframe so it can't be scrolled.

client/checkout/utils/platform-checkout.js
```diff
@@ -6,13 +13,69 @@ export const handlePlatformCheckoutEmailInput = ( field ) => {
...
+	iframeWrapper.addEventListener( 'click', closeIframe );
+	window.addEventListener( 'message', ( e ) => {
+		if ( getConfig( 'platformCheckoutHost' ) !== e.origin ) {
+			return;
+		}
+
+		if ( 'close_modal' === e.data.action ) {
+			closeIframe();
+		}
+	} );
```
brettshumaker, line 45: If we add `overflow: hidden` to `body` in `openIframe`, we'll need to also remove it here (or remove the class - whichever method you choose).

brettshumaker (dismissed):
Things are mostly working but I've got a few changes/questions.

Also, this `margin-top` in [`client/checkout/blocks/style.scss`](https://github.com/Automattic/woocommerce-payments/blob/bdf28db8885e79aa85f75a18c6b167591a1156b3/client/checkout/blocks/style.scss#L13) is being applied to the iframe on the shortcode checkout.

> This is all done but I can't get the styles to load in the block checkout.

That file I mentioned above seems to get loaded on both shortcode and block checkout so maybe that'd work?

client/checkout/utils/platform-checkout.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { debounce } from 'lodash';
```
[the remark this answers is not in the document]
superdav42: Somehow I missed pushing my last commit so it didn't have all the changes needed. I updated the [webpack config](https://github.com/Automattic/woocommerce-payments/pull/3736/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fR116) to bundle lodash instead of using it externally. This should only copy the debounce function into the js asset. I look at implementing our own debouce but it's not a trivial function so I decided not to.

client/checkout/utils/platform-checkout.js
superdav42: I think that might need to be in the platform code. I'll play around with it.

client/checkout/utils/platform-checkout.js
superdav42: `a@a` is technically a valid email address that could be used by some systems so the browser html5 based validation works. Ideally we'd use the same validation scheme that is used by WordPress but this is more complicated. I could probably use a simple regex.

brettshumaker: > a@a is technically a valid email address that could be used by some systems
Ah, that's right. As you said, we should probably use the same validation as WordPress since all of our users will have to have an email address that WordPress sees as valid.

client/checkout/utils/platform-checkout.js
brettshumaker: Sorry, I meant that the merchant site can't be scrolled. But maybe not a bad idea for the iframe to have the same restriction

superdav42:
Should have all issues address. Also made a [PR](https://github.com/Automattic/woopay/pull/354/files) on platform checkout side.

client/checkout/utils/platform-checkout.js
```diff
@@ -25,15 +88,48 @@ export const handlePlatformCheckoutEmailInput = ( field ) => {
...
 		}, waitTime );
 	} );
+
+	// Track the current state of the header. This default
+	// value should match the default state on the platform.
+	let iframeHeaderValue = true;
+	const getWindowSize = () => {
+		if ( 768 >= window.innerWidth && ! iframeHeaderValue ) {
+			iframeHeaderValue = ! iframeHeaderValue;
+			iframe.contentWindow.postMessage(
```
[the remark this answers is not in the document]
superdav42: Moved It to the load event of the iframe so it shouldn't be a problem.

brettshumaker (changes requested):
Almost there! Just a few more tweaks and I think we're in good shape to merge.

client/checkout/platform-checkout/style.scss
```diff
@@ -0,0 +1,80 @@
...
+			height: 90vh;
+			max-height: 520px;
+			position: absolute;
+			left: calc( 100% + 30px );
+			top: -250px;
+			width: calc( 100vw - 100% - 94px );
+			max-width: 374px;
+			border-radius: 8px;
+			&.open {
+				top: -45vh;
```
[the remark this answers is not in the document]
superdav42: This should be fixed now. I set the top to half height after it's calculated.

client/checkout/platform-checkout/style.scss
```diff
@@ -66,7 +66,7 @@
 		&::after {
 			content: '';
 			position: absolute;
-			left: calc( 100% + 15px );
+			left: calc( min(calc(100% + 30px), calc( 85vw - max(85vw - 100% - 94px, 300px) ) ) - 15px);
```
brettshumaker, line 69: Looks like the linter is upset about this line - I think because of the `calc` inside of the `min()`.

```suggestion
			left: calc( min( 100% + 30px, 85vw - max(85vw - 100% - 94px, 300px) ) - 15px)
```

brettshumaker (changes requested):
Looks good. Just one issue with the linter, but after that, this is good to go. 👍

brettshumaker (approved):
LGTM

client/checkout/platform-checkout/style.scss
brettshumaker: Looks like I missed a semi-colon on my suggested change. I pushed a commit to fix that and the CSS formatter added a few changes based on line length on commit.

PR #3746, opened 2022-02-03
Use dedicated data attributes for E2E tests

### LINKED ISSUE #3425: E2E Tests: Introduce a dedicated HTML attribute to be used for testing only
### Description
Currently tests use HTML id or class attributes for identifying elements, which lead to flakiness. Introduce a dedicated HTML attribute to be used for testing only. Reference: pcr5Cj-qJ-p2 
Discussion: paJDYF-2PM-p2 

### Acceptance criteria

* E2E test improvements so tests should pass and flakiness in pipeline should be reduced.

deepakpathania:
Fixes #3425 

Some E2E tests currently use nested CSS properties for grabbing elements that might change, introducing flakiness in the tests. This adds custom `data-testid` attributes to elements and use them to select elements in E2E tests.

#### Testing instructions

* Go to Actions->WCPay E2E tests->Run workflow.
* Select this branch: `add/e2e-test-specific-identifiers` and run the workflow.
* All refactored tests should still pass, [reference job](https://github.com/Automattic/woocommerce-payments/actions/runs/1794296607)

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : 'QA Testing Not Applicable'


zmaglica:
Before I run tests locally, I would change this `data-testid` attribute to something else like `data-componentid` or something similar or even other HTML attributes like `id` that can be used in tests. the same `data-testid` can be used to put other HTML attributes like `id` or `name`.

@deepakpathania What do you think about this approach? Adding data attributes is fine but it seems that there are other HTML attributes that we can without adding `data-testid` directly to the HTML element for a sake of testing.

deepakpathania:
@zmaglica The `data-testid` standard is actually used for other places in the repo already for things like multi currency flows, I tried to follow the same to keep it consistent across the codebase.

zmaglica:
> @zmaglica The `data-testid` standard is actually used for other places in the repo already for things like multi currency flows, I tried to follow the same to keep it consistent across the codebase.
> 
> Using a custom attribute for tests make more sense than using existing `id` or `name` so that we don't add any dependency on actual flows with test flows IMO. We already had selectors to reference these elements otherwise.

Make sense.

zmaglica (approved):
LGTM!

PR #3755, opened 2022-02-07
Move class name to the outer level

Mayisha:

On page load, if any other payment method is selected apart from debit/credit card, the checkbox is hidden. But on selecting the card method, multiple checkboxes are rendered.

before

The reason is the class name is attached to the react [component](https://github.com/Automattic/woocommerce-payments/blob/develop/client/components/platform-checkout/save-user/checkout-page-save-user.js#L30). When another payment method is selected, on window load [this condition](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/platform-checkout/index.js#L17) is false, so the component is rendered but [returns null](https://github.com/Automattic/woocommerce-payments/blob/develop/client/components/platform-checkout/save-user/checkout-page-save-user.js#L26) as the payment method is something else. On ajax request completion [this condition](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/platform-checkout/index.js#L17) is false again as the previous component render did not have the class name (but the component is actually attached already). So it attaches the component again. Finally on selecting card method it shows 2/3 checkboxes.

#### Testing instructions

Recreate the issue-
- First place an order with `cod` method (this will set the `cod` as payment method by default in the next order and help reproduce the issue)
- Add some products from the shop and go to the checkout page. The payment method is set to `cod` so the checkbox is absent
- Select card payment. Multiple checkboxes will appear.

This branch-
- Checkout this branch
- - Add some products from the shop and go to the checkout page. The payment method is set to `cod` so the checkbox is absent
- Select card payment. Only one checkbox will appear as expected

after


harriswong:
👀

harriswong:
Hey @Mayisha , I am using the "Recreate the issue" steps. I am using the latest `trunk` and I can't see the multiple "remember your details". I am using the following feature flags:
```php
         * @return bool
         */
        public static function is_upe_enabled() {
-               return '1' === get_option( self::UPE_FLAG_NAME, '0' );
+               return false; //'1' === get_option( self::UPE_FLAG_NAME, '0' );
        }

        /**
@@ -32,7 +32,7 @@ class WC_Payments_Features {
         * @return bool
         */
        public static function did_merchant_disable_upe() {
-               return 'disabled' === get_option( self::UPE_FLAG_NAME, '0' );
+               return true || 'disabled' === get_option( self::UPE_FLAG_NAME, '0' );
        }

        /**
@@ -95,7 +95,7 @@ class WC_Payments_Features {
         * @return bool
         */
        public static function is_platform_checkout_enabled() {
-               return '1' === get_option( '_wcpay_feature_platform_checkout', '0' );
+               return true || '1' === get_option( '_wcpay_feature_platform_checkout', '0' );
        }
```

client/checkout/platform-checkout/index.js
```diff
@@ -23,6 +23,8 @@ const renderSaveUserSection = () => {
 	)?.[ 0 ];
 	const buttonParent = placeOrderButton?.parentNode;
 	const checkoutPageSaveUserContainer = document.createElement( 'div' );
+	checkoutPageSaveUserContainer.className =
+		'platform-checkout-save-new-user-container';
```
harriswong, line 27: I see. Since we use the class `platform-checkout-save-new-user-container` to determine whether or not `saveUserSection` is already created (L14 above), by moving the classname to here, we make sure that the mount point is independent of the React component. 

Regardless of how many payment methods and how many times this is called, we will either have `<div class="'platform-checkout-save-new-user-container'"></div>` or with the component mounted inside.

This looks good to me.

harriswong (approved):
Tested in `fix/multiple-save-user-checkbox-bug`. I checkout with COD so that the second time I checkout, it is COD by default. Confirmed that there is only 1 "Remember your details" section. 👍

PR #3766, opened 2022-02-08
Refactor the check for rejected account 

marcinbot:
See https://github.com/Automattic/woocommerce-payments/pull/3720/files#r799667133

- Check if the stripe is connected before checking for the rejected status
- On error, always return false, remove the `on_error` arg
- Add some tests

#### Testing instructions

* Unit tests should pass, it's a one line change otherwise, where we always expected `false` as the returned value

-------------------

**Post merge**

- QA Testing Not Applicable

### DIFF
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -116,34 +116,6 @@ class WC_Payments_Account {
 	}
 
 	/**
-	 * Checks if the account has been rejected, assumes the value of $on_error on any account retrieval error.
-	 * Returns false if the account is not connected.
-	 *
-	 * @param bool $on_error Value to return on server error, defaults to false.
-	 *
-	 * @return bool True if the account is connected and rejected, false otherwise, $on_error on error.
-	 */
-	public function is_account_rejected( bool $on_error = false ): bool {
-		try {
-			$account = $this->get_cached_account_data();
-
-			if ( false === $account ) {
-				// False means error.
-				return $on_error;
-			}
-
-			if ( is_array( $account ) && empty( $account ) ) {
-				// Empty array means no account, so not rejected.
-				return false;
-			}
-
-			return strpos( $account['status'] ?? '', 'rejected' ) === 0;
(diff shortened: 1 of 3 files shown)


includes/class-wc-payments-account.php
```diff
@@ -116,22 +116,28 @@ public function is_stripe_connected( bool $on_error = false ): bool {
...
-	 * @return bool True if the account is rejected, false otherwise, $on_error on error.
+	 * @return bool True if the account is connected and rejected, false otherwise, $on_error on error.
 	 */
 	public function is_account_rejected( bool $on_error = false ): bool {
 		try {
 			$account = $this->get_cached_account_data();
 
-			if ( empty( $account ) ) {
-				// Empty means no account, so not rejected.
+			if ( false === $account ) {
```
dechov, line 130: I believe the same logic is covered by `try_is_stripe_connected` below – should we just use `is_stripe_connected` for all of the conditions aside from the `rejected` check itself?

On another note, what do you think of moving this function so that those two are adjacent in the code as before?

marcinbot: > I was thinking $on_error would just be passed to is_stripe_connected

The thing is, it would be difficult to tell what on_error returned from this function actually represents. I think if we ever want to handle a case where an errored out connection has to be treated as rejected (which I think is unlikely) we could do separate checks outside of this class :) 

I'll change the PR title.

dechov (approved):
Changes LGTM and tests pass. Errors on PHP 8 should be addressed in https://github.com/Automattic/woocommerce-payments/pull/3726 I think?

I was hoping to test with the instructions in 1542-gh-Automattic/woocommerce-payments-server but haven't gotten the chance

Tagging @dmallory42 (as reviewer of https://github.com/Automattic/woocommerce-payments/pull/3720) in case you have some bandwidth to check this out

dmallory42 (approved):
LGTM! I was able to test using the instructions of 1542-gh-Automattic/woocommerce-payments-server and it worked well.

Issue #3771, opened 2022-02-09
Cannot add a new payment method to existing subscriptions


RadoslavGeorgiev:
## How it currently works

https://github.com/Automattic/woocommerce-payments/pull/2386 allowed subscriptions' payment methods to be changed through UPE. It's recommended to see the PR, but __TL;DR:__ When the "Change payment method" form is submitted:

1. The SetupIntent needs to be confirmed, and for that we use Stripe's `confirmSetup` API.
2. Once the intent has been confirmed, the customer is redirected back to the same page, but with a `redirect_status=succeeded` parameter in the URL.
3. Within UPE's `get_payment_fields_js_config()` method, we'll catch that, and add the new payment method to the customer, and __configure JS to select the new payment method__ when the page is loaded.
4. The gateway's `payment_fields()` will proceed, and include the new payment method within the displayed saved payment methods.
5. When the page is loaded, JS will select the new payment method, and properly submit the form.

As described in the original PR, there's an additional step, which is not ideal. That's the case, because `confirmSetup` redirects back with a GET request, but Subscriptions are expecting a POST request.

## What is broken

In https://github.com/Automattic/woocommerce-payments/pull/3718 we had to localize scripts outside of the checkout shortcode, because localization within shortcodes doesn't work with block-based themes and WordPress 5.9.

This broke steps 4-5 from "How it currently works", because `get_payment_fields_js_config` is called after `payment_fields`, and the newly added payment method does not appear on the page for step 4.

### Steps to reproduce

1. Sign up for a subscription
2. Try to change the subscription's payment method (from My Account -> Subscriptions -> [Subscription] -> Change payment) with a new card.
3. You will see that the new payment method is assigned added to the customer, but not assigned to the subscription.

## Ideal scenario

How can we improve this? There are two paths:

1. __Non-redirect payment__ (card, SEPA, etc.):
	1. Use `confirmSetup()`'s `redirect: 'if_required'` parameter in order to avoid the redirect.
	2. Add the Setup Intent details as fields to the form.
	3. Submit the form.
	2. Use the gateway's `process_payment` method, called by `WC_Subscriptions_Change_Payment_Gateway::change_payment_method_via_pay_shortcode()` to create the WC token and assign it to the customer/subscription/order immediately.
[...]

PR #3772, opened 2022-02-09
Fix subscription PM changes.

RadoslavGeorgiev:
Fixes (only partially) #3771.
Closes https://github.com/Automattic/woocommerce-payments/issues/3765.

Call `get_payment_fields_js_config()` as early as possible within UPE's `payment_fields` in order to allow newly added PMs to be displayed as saved payment methods, and chosen by JavaScript before the "Change payment method" form gets re-submitted.

More detailed reasoning can be found in the issue.

#### Testing instructions

We need to use the following matrix in order to perform those actions:

- Complete the classic checkout
- Complete checkout through the checkout block.
- Add a new payment method to an existing subscription (new as one that doesn't exist in the database yet, not just a different one)

In combinations of the following scenarios:

- When using a block-based theme (ex. Twenty Twenty-Two) or using a classic theme (ex. Storefront)
- When using UPE fields and when **not** using UPE fields

| UPE | Block-based | Checkout block | Classic | Subscription update |
|-----|-------------|----------------|---------|---------------------|
| 0   | 0           | Yes/No         | Yes/No  | Yes/No              |
| 1   | 0           | Yes/No         | Yes/No  | Yes/No              |
| 0   | 1           | Yes/No         | Yes/No  | Yes/No              |
| 1   | 1           | Yes/No         | Yes/No  | Yes/No              |

- [x] No changelog entries needed, as the previous change is not live yet.

### DIFF
--- a/includes/payment-methods/class-upe-payment-gateway.php
+++ b/includes/payment-methods/class-upe-payment-gateway.php
@@ -760,7 +760,19 @@ class UPE_Payment_Gateway extends WC_Payment_Gateway_WCPay {
 		try {
 			$display_tokenization = $this->supports( 'tokenization' ) && ( is_checkout() || is_add_payment_method_page() );
 
-			add_action( 'wp_footer', [ $this, 'enqueue_payment_scripts' ] );
+			/**
+			 * Localizing scripts within shortcodes does not work in WP 5.9,
+			 * but we need `$this->get_payment_fields_js_config` to be called
+			 * before `$this->saved_payment_methods()`.
+			 */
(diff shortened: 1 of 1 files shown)


RadoslavGeorgiev:
FWIW, the matrix is all 🟢 for me :)

shendy-a8c (approved):
Looks good. Tested well against the matrix you provided.
I also assume red PHP 8 compatibility checks are OK per paJDYF-38U-p2.

shendy-a8c:
I just edited this PR's description that it closes https://github.com/Automattic/woocommerce-payments/issues/3765.

PR #3782, opened 2022-02-10
Cache UPE intents for reuse (WC Session edition)

### LINKED ISSUE #3602: Redundant payment intent on Checkout page load
### Describe the bug

I'm not entirely sure but I think this started after I enabled the new payments experience (early access) from the Payments Settings page.

Every time the UPE element is mounted—eg. on every load of the Checkout page—a new payment intent is created:

https://github.com/Automattic/woocommerce-payments/blob/62b3e5ae9a2f72314a1a316c9019cb57f39e0122/client/checkout/classic/upe.js#L298-L300

However, this intent goes unutilized if only the **card** and **cash-on-delivery** payment methods are enabled. The card payment method creates its own new intent. The cash-on-delivery payment method neither needs an intent on checkout nor creates one.

### To Reproduce

1. Enable UPE (I guess via the "Enable the new Stripe checkout experience" option on the Settings page).
2. Keep only card and cash-on-delivery payment methods enabled.
3. Create a new order in WooCommerce frontend.
4. Visit the Checkout page.
5. Look for the `create_payment_intent` network call in your browser's developer tools.

### Actual behavior

A network request is sent to create a new payment intent. The newly created intent may be verified in Stripe's Connect dashboard.

### Screenshots

Screenshot 2022-01-05 at 7 02 56 PM

image

### Expected behavior

A new intent should only be created when needed. Eg. on clicking the "Place order" button.

### Additional context

See #3430 for a related discussion.

anu-rock:
Fixes #3602

Mounting the UPE element requires us to create a payment or setup intent in advance in the front-end code. This happens on the Checkout page. When a non-UPE payment method is used or the page is reloaded, the previously created intent is wasted and a new one is created. This has the potential of resulting in a large number of abandoned intents even for just one merchant.

This PR fixes this behavior by caching the newly created intent in WooCommerce session. Once cached, it is then reused in the next mounting of UPE. The flow looks like this (driven by JS code):

image

Furthermore:

* Since a payment intent is dependent on the cart's total amount, the logic ties an intent with cart hash.

#### Testing instructions

Testing will require you to enable UPE first. Do this by using the "Enable the new Stripe checkout experience" option on the Settings page. Update the store currency as per the payment methods you want to test. Eg. Giropay, iDEAL, etc. will require Euro as the store currency.

##### Payment Intent

* Create a new order in the front-end.
* Open Network tab in browser's dev tools, and visit the Checkout page.
* On this page, notice an XHR call to `create_payment_intent`.
* Reload the page. No fresh XHR call to `create_payment_intent` should be observed.
* Complete the payment. The cached intent should be removed. Verify this by visiting the Checkout page again and observing an XHR call to `create_payment_intent`.
* Also check if the cached intent is removed when a payment fails. When any of the following is used, reloading the Checkout page should generate a fresh XHR call to `create_payment_intent`:
  * a [credit card that is immediately declined](https://stripe.com/docs/testing#cards-responses)
  * a [card that requires further authentication](https://stripe.com/docs/testing#regulatory-cards) and eventually the transaction fails
  * a redirection payment method (e.g. Giropay, SEPA, iDEAL, Sofort, etc.) and eventually the transaction fails
  * a saved payment method that fails

**FYI:** The intent id retrieved from cache or received as a response from `create_payment_intent` is used only when a new payment method is used. Using saved cards (successfully) will create a new intent on the server, resulting in the UPE intent being abandoned.

##### Setup Intent

* Open Network tab in browser's dev tools, and visit My account → Payment methods → Add payment method.
* On this page, notice an XHR call to `init_setup_intent`.
* Reload the page. No fresh XHR call to `init_setup_intent` should be observed.
* Add a new payment method.
* The cached intent should be removed after successfully adding a new PM.
* Go back to the "Add payment method" page. A fresh XHR call to `init_setup_intent` will confirm that the previously cached intent was removed.

#### Scenarios I have tested

* New card - successful payment
* New card - declined
* New card - authentication failed
* Saved card - successful payment
* Saved card - authentication failed
* New SEPA - successful payment
* Saved SEPA - successful payment
* Giropay - successful payment
* Giropay - failed payment

-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt`
- [x] Covered with tests
- [ ] Tested on mobile (does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.8.0#cache-upe-intents-for-reuse


includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1071,8 +1094,12 @@ public function log_payment_error_ajax() {
 			}
 			$order->add_order_note( $order_note );
 
+			self::remove_upe_payment_intent_from_session();
```
anu-rock, line 1097: It would be cleaner if instead this method is called in a hook rather than directly in the "log" method.

I tried this:

```php
add_action( 'wc_ajax_wcpay_log_payment_error', [ 'WCPay\Payment_Methods\UPE_Payment_Gateway', 'remove_upe_payment_intent_from_session' ] );
```

But that apparently didn't work as `remove_upe_payment_intent_from_session` was not called on client-side errors.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -94,6 +98,9 @@ public function __construct( WC_Payments_API_Client $payments_api_client, WC_Pay
 		if ( ! is_admin() ) {
 			add_filter( 'woocommerce_gateway_title', [ $this, 'maybe_filter_gateway_title' ], 10, 2 );
 		}
+
+		add_action( 'woocommerce_order_payment_status_changed', [ 'WCPay\Payment_Methods\UPE_Payment_Gateway', 'remove_upe_payment_intent_from_session' ], 10, 0 );
```
FangedParakeet, line 102:
```suggestion
		add_action( 'woocommerce_order_payment_status_changed', [ __CLASS__, 'remove_upe_payment_intent_from_session' ], 10, 0 );
```
I think you can save a few characters and safely use the `__CLASS__` here. Or at the least, I don't think you need the full absolute class name, since the class is namespaced.

FangedParakeet (approved):
Cool, the approach looks neat and tidy and is working pretty well for me! 

I found one issue, which I'm pretty sure is not related to the changes in this PR, but I was wondering if you could just double-check to confirm. If you check out with a 3DS card that requires authentication, then fail the authentication once the modal comes up, once you are returned to the checkout, if you retry with a valid payment method, it will always fail...but if you now retry _again_ it will succeed this time (as expected).

Here are the steps, in case that was a little confusing.

1. Enter checkout. 
2. Enter 3DS card details. Place order.
3. Modal opens, manually fail authentication.
4. Enter valid card details. Place order.
5. Transaction fails.
6. Place order again.
7. Transaction succeeds.

This works with other payment methods too. For example...

1. Enter checkout. 
2. Enter 3DS card details. Place order.
3. Modal opens, manually fail authentication.
4. Select new payment method (e.g. Giropay). Place order.
5. Transaction fails.
6. Place order again.
7. Transaction succeeds, redirects to Giropay authorisation page.

I don't _think_ this is caused by some cookie confusion (or at least I can't find an obvious place where that might happen), but if you wouldn't mind double-checking just to be sure, that'd be grand!

Anyways, I've made one small code suggestion, but everything here looks happy enough to me, so I'm gonna approve, so that I don't slam the brakes on this train. I'll leave it up to the other reviewers to give the final blessing, but LGTM. Nice one!

includes/payment-methods/class-upe-payment-gateway.php
anu-rock: Fixed in c6ff6fa :)

client/checkout/classic/upe.js
```diff
@@ -12,7 +12,7 @@ import { getConfig, getCustomGatewayTitle } from 'utils/checkout';
 import WCPayAPI from '../api';
 import enqueueFraudScripts from 'fraud-scripts';
 import { getFontRulesFromPage, getAppearance } from '../upe-styles';
-import { getTerms } from '../utils/upe';
+import { getTerms, getCookieValue } from '../utils/upe';
```
kalessil, line 15: Do we still need this?

kalessil (approved):
LGTM (did review only, will do testing tomorrow with more capacity on my end)!

FangedParakeet:
@anu-rock, I'll retest my calamaties against `develop` and create a separate issue if it's aught other than my own fault. I've no other critical complaints with anything within these changes then! :shipit:

client/checkout/classic/upe.js
anu-rock: I am using it to get the cart hash down here. Perhaps there's a better way to get the hash?

https://github.com/Automattic/woocommerce-payments/blob/5ec80784666a545b38bdd09062a81218007eae96/client/checkout/classic/upe.js#L668

kalessil: +1

PR #3785, opened 2022-02-11
Fix not populated loan variable without account cache 

tpaksu:

This PR will fix the bug which is introduced on #3717 that causes the Stripe onboarding CTA page to display "Sorry, you are not allowed to view this page" error, because of a React error. The page doesn't show up because of this error, because the accountLoans property isn't populated when a Stripe account isn't connected, and even it's not populated, the client code tries to read it.

#### Testing instructions

* All tests shall pass.
* Without the fix, go to WCPay dev tools and enable "Force the plugin to act as disconnected from WCPay", then click the "Payments" link on the sidebar. You should get the error mentioned.
* With the fix, do the same and expect to see the Stripe onboarding CTA page.

-------------------

**Post merge**

QA Testing Not Applicable

### DIFF
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -315,7 +315,11 @@ class WC_Payments_Account {
 	 */
 	public function get_capital() {
 		$account = $this->get_cached_account_data();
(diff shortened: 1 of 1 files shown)


mordeth (approved):
LGTM! 👍

Issue #909, opened 2020-09-15
Saved cards: View saved payment methods list for a user


luizreis:
Child of #908

See paJDYF-Kq-p2 for more details.

Merchants should be able to view a list of payment methods for a customer, including:
- Stripe customer ID
- Basic payment method details
- *Click to copy is going to be addressed in another issue*

![image](https://user-images.githubusercontent.com/7714042/93265088-78a15380-f77e-11ea-9d09-4728327b93eb.png)

![image](https://user-images.githubusercontent.com/7714042/93265112-7fc86180-f77e-11ea-8ad2-939271c65bac.png)

![image](https://user-images.githubusercontent.com/7714042/93265254-b56d4a80-f77e-11ea-99d0-01f5c696b60d.png)

luizreis:
Closing for now. More info can be found on p1643975980351219-slack-C02L752DS0G.

PR #3789, opened 2022-02-14
Fix Stripe Level 3 data limit for orders with more than 200 different products

elazzabi:
Fixes [#909](909-gh-Automattic/woocommerce-payments-server)

#### Fix Stripe Level 3 data limit

**Context:**
With Stripe's Level 3 data, we are sending metadata about each order to Stripe. This metadata includes a list of product ids bought, description, price, taxes... With a new limitation by Stripe, the maximum basket size is 200 products. This doesn't impact the majority of orders, but we are fixing it to be compliant with Stripe's API.

Stripe offered two solutions: Split charges into separate payments or aggregate items together. We are going with the second option given it's 1) doesn't take much effort (compared to the other one), and 2) this doesn't touch 99% of cases, and thus not worth the effort.

**Solution:**
This change bundles all products (when basket size is > 200) into one product. I _personally think_ this is enough, but I can re-work this PR if it's not the case.

The old behavior is kept intact, which means we'll only change the logic if the basket size is more than 200 different products. This change contains necessary tests too.

#### Testing instructions

* This is unit tested. No manual tests are needed (unless we can have an order with 200 different products)

-------------------

**Post merge**


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2768,4 +2773,33 @@ public function get_connection_url() {
...
+				function( $cost, $qty ) {
+					return $cost * $qty;
+				},
+				array_column( $items, 'unit_cost' ),
+				array_column( $items, 'quantity' )
+			)
+		);
+
+		return (object) [
+			'product_code'        => (string) substr( uniqid(), 0, 26 ),
```
elazzabi, line 2797: Random `product_code` given this "bundle" is created programmatically and not tied to a specific bundle

anu-rock (commented):
The code changes look great, thanks 👏

But I wonder if we also considered the approach of reducing/bundling only the "excess" items—items beyond 199—into a single item so that the total count would be 200. One benefit is that metadata for the first 199 items will at least not be lost for such big orders.

I'll concede that I'm not sure how often we have to process orders that huge. So the approach you have here might work just fine if it's a rare occurrence.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2128,6 +2128,11 @@ public function get_level3_data_from_order( WC_Order $order ): array {
 		};
 		$items_to_send = array_map( $process_item, $order_items );
 
+		if ( count( $items_to_send ) > 200 ) {
+			// Stripe API max items is 200. If more than that is present, bundle everything in one "product bundle".
+			$items_to_send = [ $this->bundle_level3_data_from_items( $items_to_send ) ];
```
kalessil, line 2133: I'd suggest keeping the first 199 items unchanged and grouping the rest with your approach (hiding behind 'X items more'). how does it sound?

kalessil (changes requested):

includes/class-wc-payment-gateway-wcpay.php
elazzabi: Same as @anu-rock 's suggestion. I think that makes sense!

I'll work on it 👍

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -832,6 +832,9 @@ public function test_level3_data_bundle() {
 		];
 
 		$bundle_data = $this->wcpay_gateway->bundle_level3_data_from_items( $items );
+
+		$this->assertEquals( $bundle_data->product_description, '2 more items' );
```
anu-rock, line 836: We prefer `assertSame` to account for type :)

anu-rock (approved):
LGTM with the new approach 🎉

tests/unit/test-class-wc-payment-gateway-wcpay.php
kalessil: +1 for using `assertSame`

kalessil (approved):
LGTM!

tests/unit/test-class-wc-payment-gateway-wcpay.php
elazzabi: Sounds good! Added in https://github.com/Automattic/woocommerce-payments/pull/3789/commits/5e88b591f6e296ea651c495b23fb219fa1e08592

dpaun1985 (approved):
LGTM

PR #3796, opened 2022-02-14
Jetpack Tracking integration

### LINKED ISSUE #3823: Infrastructure to track Platform Checkout buyer events
### Description
Implement the infrastructure to support tracking requirements of platform checkout

### Acceptance criteria

* Should not track events from site admins
* Tracking should only be enabled when platform checkout is enabled.
* Should be able to track logged-in user events
* Should be able to track logged-out user events

### Additional context

pdpOw2-l8-p2
pdpOw2-mz-p2

malithsen:
Fixes #3823

This PR integrates `jetpack-tracking` package to support the tracking needs of platform checkout.

`Tracking` class that comes from `jetpack-tracking` package is extended to 
* Not track IP address and User-Agent
* Track guest users

**See** pdpOw2-l8-p2, pdpOw2-mz-p2

Proposed changes only implement the infrastructure needed to track C2 events. Actual tracking events will be implemented in subsequent PRs. Event names mentioned in the testing instructions below are for testing purposes only. They might not fire from the place mentioned below and will likely track other properties. 

## Testing instructions

* In your `wp_options` table ensure that
    * `_wcpay_feature_platform_checkout` option is set to 1.
    * `jetpack_tos_agreed` option is set to 1. This will already be set if you have Jetpack plugin installed
* Enable Platform Checkout option as a merchant from Payments -> Settings -> Express checkouts

#### Setup
* Add the following hook to the constructor `add_action( 'woocommerce_add_to_cart', [ $this, 'track_add_to_cart' ], 10, 6 );`

#### As a logged in normal user
* Login to the store as a customer (or as any user except administrator)
* Add a product to the cart
* In ~5 minutes, checks the Tracks Live View for an event called, `woocommerceanalytics_order_checkout_start`
* Expand the event Props section
* Notice the `userId` 

#### As a logged out user
Repeat the above tests and notice that `woocommerceanalytics_order_checkout_start` appears on live view with a different `userId`

#### As site administrator
* Login to the store as site admin 
* Add a product to the cart
* Ensure that no tracking event is created on MC live view after ~5 minutes

#### Disable Platform Checkout
* Disable Platform Checkout option as a merchant from Payments -> Settings -> Express checkouts
* Add a product to the cart as a normal user
* Ensure that no tracking event is created on MC live view after ~5 minutes

### Test tracking from JS
* Add the following tracking line within the `onClick` handler in https://github.com/Automattic/woocommerce-payments/blob/develop/client/components/platform-checkout/index.tsx#L47
```
wcpayTracks.recordUserEvent( 'platform_checkout_registered', null );
``` 
* Make sure wcpayTracks is imported with `import wcpayTracks from 'tracks';` as well
* Add a product to the cart 
* Go to blocks checkout page
* Click "Buy now with platform checkout". 
* In ~5 minutes, checks the Tracks Live View for an event called, `woocommerceanalytics_platform_checkout_registered`
 

-------------------

**Post merge**

cc @brianyu28


client/tracks/index.js
```diff
@@ -28,6 +28,19 @@ function recordEvent( eventName, eventProperties ) {
...
+ * Records events from buyers.
+ *
+ * @param {string}  eventName       Name of the event.
+ * @param {Object?} eventProperties Event properties.
+ */
+function recordUserEvent( eventName, eventProperties ) {
+	if ( ! window.jpTracksAJAX ) {
+		return;
+	}
+	window.jpTracksAJAX.record_ajax_event( eventName, null, eventProperties );
```
a-danae, line 41: @malithsen I'm not 100%, but when I have Jetpack active, looks like the event triggered here following the instructions from the description the recorded event is named `jetpack_platform_checkout_registered`, and when Jetpack is not installed it's named `woocommerceanalytics_platform_checkout_registered`. Can you replicate on your end?

includes/class-platform-checkout-tracker.php
```diff
@@ -0,0 +1,160 @@
...
+	 * @return bool
+	 */
+	public function should_track() {
+		// Track only site pages.
+		if ( is_admin() && ! wp_doing_ajax() ) {
+			return false;
+		}
+
+		// Don't track site admins.
+		if ( is_user_logged_in() && in_array( 'administrator', wp_get_current_user()->roles, true ) ) {
```
a-danae, line 121: Just double checking here that we do want to exclude users with the "administrator" role instead of users that can [manage_woocommerce](https://woocommerce.com/document/roles-capabilities/).

includes/class-wc-payments.php
```diff
@@ -306,6 +310,8 @@ public static function init() {
 			new WC_Payments_Status( self::get_wc_payments_http(), self::get_account_service() );
 		}
 
+		new Platform_Checkout_Tracker( self::get_wc_payments_http() );
```
a-danae, line 313: Should we include and instantiate this class within the `WC_Payments_Features::is_platform_checkout_enabled()` conditional above in line 224, like we're doing for `Platform_Checkout_Save_User`? Thinking that this would only be used in those cases.

a-danae (commented):
Thanks for working on this, the instructions, and for accompanying the changes with tests @malithsen!

The events are recorded as described, and things look good. I've left a few comments with questions but nothing big. Only the `jetpack` <-> `woocommerceanalytics` prefix question might be worth checking before approving this PR, but other than that I think we're good to go.

client/tracks/index.js
malithsen: Hey @a-danae this is a great catch. I could not replicate it, but I noticed your test events on Tracks live view with the incorrect prefixes. I am wondering if another plugin (Jetpack or some other) that uses `jetpack-tracking` is taking precedence over what's implemented here. In which case it will use the default `jetpack` prefix. I will try to see whether this is the case.

a-danae: Welp, I've been trying to replicate again but things look good. I keep seeing `woocommerceanalytics` as the prefix and not `jetpack`. Thanks for checking this on your end as well! I'm marking this as resolved.

a-danae (approved):
I couldn't replicate again the issue with the `woocommerceanalytics` and `jetpack` prefix I mentioned before, I think we're good here.

client/tracks/index.js
malithsen: I was able to replicate this when tracking using JS (via ajax) with a logged-in user. It was because we weren't overriding jetpack-tracking's `wp_ajax_jetpack_tracks` hook [here](https://github.com/Automattic/woocommerce-payments/pull/3796/commits/1f060f8a8f28adddd997063a95a6772532e13aa0#diff-cfa882fcd35a7427cbf7f2d3ae9f54370d02d46ca04342f3f74282df721d3b20R33). It should be fixed with https://github.com/Automattic/woocommerce-payments/pull/3796/commits/240ea404f90420402855b11c017af48d563cb641

With the above change, I also opted to separate ajax hooks specific to this implementation to avoid conflicts with other plugins that may also use `jetpack-tracking` module

includes/class-wc-payments.php
malithsen: Good point @a-danae, adjusted with https://github.com/Automattic/woocommerce-payments/pull/3796/commits/26d5b45b67a2ddbc4e2a8e4701abbbd39da8d74e

includes/class-platform-checkout-tracker.php
malithsen: We collect some other buyer events like [woocommerceanalytics_product_checkout](https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php#L313) in WooCommerce Analytics module, so wanted to keep the tracking conditions close to the [ones enforced there](https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/modules/woocommerce-analytics/class-jetpack-woocommerce-analytics.php#L40) for consistency.

client/tracks/index.js
a-danae: Makes sense. I missed testing with a logged-in user when trying to replicate again, thanks for pushing there! I was able to replicate without the fix, and things now work as expected with the fix 🎉 it looks good to me.

includes/class-platform-checkout-tracker.php
a-danae: Gotcha, thanks for explaining!

includes/class-wc-payments.php
a-danae: Thanks!

a-danae (approved):
Thanks for checking out the comments before @malithsen! Things look good on my end.

includes/class-platform-checkout-tracker.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Generic method to track user events.
+	 *
+	 * @param string $event name of the event.
+	 * @param array  $data array of event properties.
+	 */
+	public function record_event( $event, $data = [] ) {
+
+		if ( ! $this->should_track() ) {
```
tommyshellberg, line 87: When testing I was confused that adding to cart as a logged-in admin actually called `$this->record_event()`. 
I didn't notice at first that `record_event()` would short-circuit because of `! $this->should_track()`. While it is nice to have a single line of code that can break out of recording all events (avoiding duplication), it was not intuitive to me at first. 

I would have expected `track_add_to_cart()` to conditionally call `$this->record_event()` itself based on its own specific conditions. So if you were an admin it would return instead of calling `$this->record_event()`.
I know that `track_add_to_cart()` doing the conditional check will likely result in duplicated code but it seems like doing all of those checks in `record_event()` could end up becoming inflexible as you add more tracks.

An alternative is to go the WooCommerce way and call the method `maybe_record_event()` 😆

tommyshellberg (approved):
Looks good and tests well. Just a comment about how we can conditionally call `record_event()` but this is great work!

includes/class-platform-checkout-tracker.php
malithsen: Thanks for having a look @tommyshellberg. `should_track` contains the base conditions that should be checked for all platform-checkout track events. Having them defined in the `record_event` method will make it less susceptible to accidentally capturing these events when we should not. So I am opting towards having the `should_track` condition as-is. However, I agree that short-circuit logic within `record_event` method is confusing, and as you suggested, adding `maybe_` suffix might clarify things a bit. Updated in https://github.com/Automattic/woocommerce-payments/pull/3796/commits/4684915fae27d304ad56b05b1f2ee2f3ff827a7d. Thanks!

malithsen:
Just noting that I removed the method that was only there to help with testing this PR in https://github.com/Automattic/woocommerce-payments/pull/3796/commits/2db2a26068b8c2006836acf609b9e3e030dcd5a1

PR #3716, opened 2022-01-27
Bump minimum supported and tested up to version for 3.7 release

deepakpathania:
Bumps the tested to version based on any available release candidates for WordPress or WooCommerce in this or the next WCPay release cycle. Updates the minimum required version by increasing the version by 1.

#### Changes proposed in this Pull Request
- Bumped the tested to version based on:
  - [WordPress 5.9 was released on January 25, 2022](https://make.wordpress.org/core/5-9/)
  - [WooCommerce 6.2 release date is February 8, 2022](https://developer.woocommerce.com/2022/01/25/woocommerce-6-2-rc/)
- Updated the minimum required version
  - WordPress 5.6 to 5.7
  - WooCommerce 4.5 to 4.6

#### Testing instructions

- Validate the changes in `woocommerce-payments.php` and `readme.txt`
- CI tests have been updated ( [reference E2E job](https://github.com/Automattic/woocommerce-payments/actions/runs/1755451028) ), and cover all required compatibilities.

-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : [Testing instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.7.0#bumped-tested-up-to-versions-for-wc-and-wp)

PR #3798, opened 2022-02-15
Bump WC minimum supported version for 3.8.0 release

ricardo:

- Bump WC from 4.6.0 to 4.7.0 (still long way behind L-2 on 6.0.0).
- Fixed readme.txt which was inconsistent with plugin header.
- Fixed php-lint-test.yml which was inconsistent with L-2.

No WP updates to catch up on.

#### Testing instructions

* Please check the CI tests in the GH workflows.

-------------------

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -11,13 +11,13 @@ jobs:
       fail-fast:    false
       max-parallel: 10
       matrix:
-        woocommerce: [ '4.6.0', '4.7.0', '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.0-rc.1']
+        woocommerce: [ '4.7.0', '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.0' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
         php:         [ '7.4' ]
         include:
           # Edge case: oldest dependencies compatibility
-          - woocommerce: '4.6.0'
+          - woocommerce: '4.7.0'
             wordpress:   '5.7'
             gutenberg:   '11.4.0' # The latest version supporting WP 5.6.
             php:         '7.1' # TODO This should be 7.0. Pending to fix in issue #2970.
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -36,7 +36,7 @@ jobs:
       fail-fast:     false
       matrix:
         # Min WooCommerce | L-2 | L(atest) | beta
-        woocommerce:   [ '4.6.0', '5.9.0', '6.1.1', 'beta' ]
+        woocommerce:   [ '4.7.0', '6.0.0', '6.2.0', 'beta' ]
         wordpress:     [ 'latest' ]
         php:           [ '7.4' ]
         test_groups:   [ 'wcpay', 'subscriptions', 'blocks' ]
--- a/.github/workflows/php-lint-test.yml
+++ b/.github/workflows/php-lint-test.yml
@@ -5,7 +5,7 @@ on:
 
 env:
   WP_VERSION:        latest
-  WC_VERSION: 5.7.0  # the min supported version as per L-2 policy
+  WC_VERSION: 6.0.0  # the min supported version as per L-2 policy
   GUTENBERG_VERSION: latest
 
 jobs:
--- a/readme.txt
+++ b/readme.txt
@@ -39,7 +39,7 @@ Our global support team is available to answer questions you may have about WooC
 = Requirements =
 
 * WordPress 5.7 or newer.
-* WooCommerce 5.9 or newer.
+* WooCommerce 4.7 or newer.
 * PHP version 7.0 or newer. PHP 7.2 or newer is recommended.
 
 = Try it now =
--- a/woocommerce-payments.php
+++ b/woocommerce-payments.php
@@ -8,8 +8,8 @@
  * Woo: 5278104:bf3cf30871604e15eec560c962593c1f
  * Text Domain: woocommerce-payments
  * Domain Path: /languages
- * WC requires at least: 4.6
- * WC tested up to: 6.1.1
+ * WC requires at least: 4.7
+ * WC tested up to: 6.2.0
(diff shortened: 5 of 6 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       fail-fast:     false
       matrix:
         # Min WooCommerce | L-2 | L(atest) | beta
-        woocommerce:   [ '4.6.0', '5.9.0', '6.1.1', 'beta' ]
+        woocommerce:   [ '4.7.0', '6.0.0', '6.2.0', 'beta' ]
```
asumaran, line 39: Are we sure we want to replace `5.9.0` with `6.0.0` here? I think we should keep the `5.x.x` branch?

.github/workflows/php-lint-test.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 env:
   WP_VERSION:        latest
-  WC_VERSION: 5.7.0  # the min supported version as per L-2 policy
+  WC_VERSION: 6.0.0  # the min supported version as per L-2 policy
```
asumaran, line 8: Can you confirm this change? From what I understand the minimum L-2 version would be `4.9.0`. Since the L-2 [policy](https://developer.woocommerce.com/2021/05/12/woocommerce-payments-is-adopting-a-new-version-support-policy/) [says](https://developer.woocommerce.com/2020/05/08/wordpress-core-support-policy/) that we will support the current version as well as the latest release from the two previous release branches.

asumaran (commented):
A couple of questions for you @ricardo. I'm a bit confused with the changes here according to the L-2 policy. I hope you can respond to my questions.

.github/workflows/e2e-test.yml
ricardo: When the previous PR was merged (#3716), the latest version was `6.1.1`, and the L-2 was `5.9.0`. Now the latest is `6.2.0`, so I had to also bump the L-2 to be 2 versions behind it.

So I basically just followed the previous PR. Does that make sense?

.github/workflows/php-lint-test.yml
ricardo: > Can you confirm this change? From what I understand the minimum L-2 version would be 4.9.0

Can you elaborate on why `4.9.0`?

`5.7.0` didn't make sense to me. WP is at `5.9` right now, this was likely mistaken as WP_VERSION, not WC.

If it were WP, it should be `5.7`, because we must support `5.7`, `5.8` and latest.

> An L-2 version support policy means supporting the latest version and the 2 versions before it. 
paJDYF-1BP-p2

So the "# the min supported version as per L-2 policy" should be 6.0.0, because as per L-2 we should support 6.0.0, 6.1.1 and 6.2.0 (latest).

Any reason

readme.txt
```diff
@@ -39,7 +39,7 @@ Our global support team is available to answer questions you may have about WooC
 = Requirements =
 
 * WordPress 5.7 or newer.
-* WooCommerce 5.9 or newer.
+* WooCommerce 6.0 or newer.
```
ricardo, line 42: @asumaran 

I've just bumped the minimum version here as per this thread p1644907045226379/1644898556.442049-slack-C01BPL3ALGP

.github/workflows/php-lint-test.yml
asumaran: > So the "# the min supported version as per L-2 policy" should be 6.0.0, because as per L-2 we should support 6.0.0, 6.1.1 and 6.2.0 (latest).

Gotcha. Thanks for the explanation 📝

asumaran (approved):
LGTM 🚀  Thanks for the responses @ricardo

.github/workflows/compatibility.yml
```diff
@@ -11,13 +11,13 @@ jobs:
       fail-fast:    false
       max-parallel: 10
       matrix:
-        woocommerce: [ '4.6.0', '4.7.0', '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.0-rc.1']
+        woocommerce: [ '4.7.0', '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.0' ]
```
htdat, line 14: I may add `beta` here too, which would have been done before, and leave a comment to explain which range this list should have: 

```suggestion
        # Minimum support version, ..., Latest version, `beta`.
        woocommerce: [ '4.7.0', '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.0' ]
```

htdat (approved):
LGTM! But let's wait until Friday before merging this PR and see if we can raise more Woo core minimum version.
paJDYF-3fF-p2#comment-9755

.github/workflows/compatibility.yml
ricardo: @htdat should we add `beta` as in [here](https://github.com/Automattic/woocommerce-payments/pull/3798/files#diff-5cc2c864012004f60ee3f9078fa1c4a8721c6b841e386c57f12bea58b692e6e2R39)?

htdat: > should we add `beta` as in [here](https://github.com/Automattic/woocommerce-payments/pull/3798/files#diff-5cc2c864012004f60ee3f9078fa1c4a8721c6b841e386c57f12bea58b692e6e2R39)?

Yes, I see that we have already added it? 

--- 

Also, my suggestion code above missed `beta` at the end, so I've just updated it.

ricardo: Yeah, I was talking about this specific change 👍.

htdat:
> But let's wait until Friday before merging this PR and see if we can raise more Woo core minimum version.

Decision has been made paJDYF-3fF-p2#comment-9727
@ricardo Let's move up one more Woo core version, i.e. from 4.6.0 to 4.8.0, then we can 🚢 this PR.

htdat (approved):
One more place to update Woo core to 4.8 in woocommerce-payments.php, then 🚢 it!

woocommerce-payments.php
```diff
@@ -8,8 +8,8 @@
  * Woo: 5278104:bf3cf30871604e15eec560c962593c1f
  * Text Domain: woocommerce-payments
  * Domain Path: /languages
- * WC requires at least: 4.6
- * WC tested up to: 6.1.1
+ * WC requires at least: 4.7
```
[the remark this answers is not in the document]
ricardo: That one I missed 🤦

PR #3642, opened 2022-01-15
Support payment processing via platform checkout

reykjalin:
#### Changes proposed in this Pull Request

To enable payment processing via platform checkout we need to make some changes to the payment processing, mostly in the form of filters.

Related PRs are listed below via GitHub mentions. You must use the code in the related PRs for the changes in this PR to be effective.

#### Testing instructions

* Follow the testing instruction in PR 280 in the platform repo (linked below).

-------------------

- [ ] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _Add link here / 'QA Testing Not Applicable'_

PR #3811, opened 2022-02-15
Add support for platform checkout server side payment processing

reykjalin:

Supersedes #3642 

This PR adds the necessary functionality for us to support platform payment processing without introducing all the new filters we needed for the first approach we tried in #3642 .

- Add a new filter, `wc_payments_account_id_for_intent_confirmation`, to allow the platform to modify the account used to confirm a payment intent.
    - **Note:** The account ID must match the provided payment intent, otherwise the intent confirmation will fail.
- Add new JS prop to the frontend settings, `accountIdForIntentConfirmation`, which contains the result of running the new filter on an empty string.
- Add the ability to process a payment when the checkout request provides a payment intent.
    - This is so the platform can process the payment and then update the order on the merchant store appropriately.

## Testing instructions

1. Make sure payments work as they normally do in a normal WCPay store. Maybe try some of the [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows)?
2. Follow the testing instructions in the corresponding platform checkout PR (should be mentioned below via GitHub mention).

I have no automated tests for this because I'm not so sure how to actually cover this with automated tests.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : QA Testing Not Applicable


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2193,6 +2201,9 @@ public function update_order_status() {
 			}
 
 			$payment_method_id = isset( $_POST['payment_method_id'] ) ? wc_clean( wp_unslash( $_POST['payment_method_id'] ) ) : '';
+			if ( 'null' === $payment_method_id ) {
+				$payment_method_id = '';
+			}
```
reykjalin, line 2206: See https://github.com/Automattic/woocommerce-payments/pull/3642/files#r786451779

reykjalin:
I'm not so sure this needs a changelog entry. I think the filter should at least have an entry, but I'm not sure about the other stuff.

Would love to hear from the reviewer on that.

asumaran (approved):
LGTM

reykjalin:
@ricardo It's been a good while since the last review, so pinging you here for a review as well.

client/checkout/api/index.js
```diff
@@ -209,9 +227,36 @@ export default class WCPayAPI {
...
+			// When not dealing with a setup intent or platform checkout we need to force an account
+			// specific request in Stripe.
+			return this.getStripe( true );
+		};
+
 		const confirmAction = isSetupIntent
-			? this.getStripe().confirmCardSetup( clientSecret )
-			: this.getStripe( true ).confirmCardPayment( clientSecret );
+			? getStripe().confirmCardSetup( clientSecret )
+			: getStripe().confirmCardPayment( clientSecret );
```
ricardo, line 259: This part is kinda tricky IMO, and even more now with the extra condition for the different account ID. I wonder if we can make it easier to understand what's going on.

Some ideas came to mind:

- Perhaps we should call the `getStripe` function here something else, to help differentiate between this and `WCPayAPI.getStripe`?
- Or, does it make sense to create a function to return the intent confirmation (confirmCardSetup/confirmCardPayment)? We can at least dismiss the extra `isSetupIntent` check.

For example:

```JavaScript
const confirmIntention = () => {
	const { locale, publishableKey } = this.options;
	const accountIdForIntentConfirmation = getConfig(
		'accountIdForIntentConfirmation'
	);

	// If this is a setup intent we're not processing a platform checkout payment so we can
	// use the regular getStripe function.
	if ( isSetupIntent ) {
		return this.getStripe().confirmCardSetup( clientSecret );
	}

	if ( accountIdForIntentConfirmation ) {
		return this.createStripe(
			publishableKey,
			locale,
			accountIdForIntentConfirmation
		).confirmCardPayment( clientSecret );
	}

	// When not dealing with a setup intent or platform checkout we need to force an account
	// specific request in Stripe.
	return this.getStripe( true ).confirmCardPayment( clientSecret );
};

const request = confirmIntention()
```

- Or even, avoid this altogether by overriding `options.accountId` when initializing the WCPay API with `getConfig( 'accountIdForIntentConfirmation' )`.

```JavaScript
// Or could this be overwritten from the back-end instead of adding "accountIdForIntentConfirmation"?
const accountId = getConfig( 'accountIdForIntentConfirmation' ) ? getConfig( 'accountIdForIntentConfirmation' ) : getConfig( 'accountId' );

new WCPayAPI(
	{
		publishableKey,
		accountId,
		forceNetworkSavedCards: getConfig( 'forceNetworkSavedCards' ),
		locale: getConfig( 'locale' ),
	}
```

And we can keep this part of the code as it was before.

```JavaScript
const confirmAction = isSetupIntent
	? this.getStripe().confirmCardSetup( clientSecret )
	: this.getStripe( true ).confirmCardPayment( clientSecret );
```

PS: Haven't tested the code here though 😅.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -616,23 +616,24 @@ public function output_payments_settings_screen() {
...
+			'fraudServices'                  => $this->account->get_fraud_services_config(),
+			'features'                       => $this->supports,
+			'forceNetworkSavedCards'         => WC_Payments::is_network_saved_cards_enabled(),
+			'locale'                         => WC_Payments_Utils::convert_to_stripe_locale( get_locale() ),
+			'isUPEEnabled'                   => WC_Payments_Features::is_upe_enabled(),
+			'isSavedCardsEnabled'            => $this->is_saved_cards_enabled(),
+			'isPlatformCheckoutEnabled'      => WC_Payments_Features::is_platform_checkout_enabled() && 'yes' === $this->get_option( 'platform_checkout', 'no' ),
+			'platformCheckoutHost'           => defined( 'PLATFORM_CHECKOUT_FRONTEND_HOST' ) ? PLATFORM_CHECKOUT_FRONTEND_HOST : 'http://localhost:8090',
+			'platformTrackerNonce'           => wp_create_nonce( 'platform_tracks_nonce' ),
+			'accountIdForIntentConfirmation' => apply_filters( 'wc_payments_account_id_for_intent_confirmation', '' ),
```
ricardo, line 636: Should the platform checkout always initialize Stripe as the platform, except when confirming an intent?

If so, perhaps we could overwrite the `accountId` instead. That would prevent a lot of maybe unnecessary JS changes I mentioned in the other comment.

ricardo (changes requested):

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -616,23 +616,24 @@ public function output_payments_settings_screen() {
...
+			'accountId'                      => $this->account->get_stripe_account_id(),
+			'ajaxUrl'                        => admin_url( 'admin-ajax.php' ),
+			'wcAjaxUrl'                      => WC_AJAX::get_endpoint( '%%endpoint%%' ),
+			'createSetupIntentNonce'         => wp_create_nonce( 'wcpay_create_setup_intent_nonce' ),
+			'createPaymentIntentNonce'       => wp_create_nonce( 'wcpay_create_payment_intent_nonce' ),
+			'updatePaymentIntentNonce'       => wp_create_nonce( 'wcpay_update_payment_intent_nonce' ),
+			'genericErrorMessage'            => __( 'There was a problem processing the payment. Please check your email inbox and refresh the page to try again.', 'woocommerce-payments' ),
+			'fraudServices'                  => $this->account->get_fraud_services_config(),
+			'features'                       => $this->supports,
+			'forceNetworkSavedCards'         => WC_Payments::is_network_saved_cards_enabled(),
```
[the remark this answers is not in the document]
reykjalin: Ah, this was accidental 🤔 
I only meant to add the `accountIdForIntentConfirmation` prop, but something must've messed things up during a merge/rebase.

I'll go through all the options and fix this 👍 

Thanks for catching this!

includes/class-wc-payment-gateway-wcpay.php
reykjalin: > Should the platform checkout always initialize Stripe as the platform, except when confirming an intent?

On the platform checkout site, _yes_. In other words; on the platform checkout site we want to use the platform account (i.e. the store account) for all cases, _except_ when confirming an intent. The reason for that is that the payment intent is created with the merchant account, so we need the same account to confirm the intent.

> If so, perhaps we could overwrite the `accountId` instead

That was the initial approach, but we decided not to go that route to limit the impact of the new filter as much as possible. If we overwrite the `accountId` that would give plugin authors (or anyone with write access to the site) the ability to overwrite the account ID used in the Stripe SDK. By adding a new property we can ensure that the value from the filter is only used when confirming intents.

It's inconvenient and complicates the JS, but provides better encapsulation for the value returned by the filter.

client/checkout/api/index.js
reykjalin: > Perhaps we should call the `getStripe` function here something else, to help differentiate between this and `WCPayAPI.getStripe?`

I did call them the same thing because they sort of serve the same purpose, but I can definitely see how that could be confusing 😅 

> avoid this altogether by overriding `options.accountId` when initializing the WCPay API with `getConfig( 'accountIdForIntentConfirmation' )`.

Unfortunately this won't work since the `WCPayAPI.getStripe()` function "caches" the Stripe SDK instance it creates, so we'll have to recreate the Stripe SDK instance here anyway. The reason being that the SDK should normally be instantiated with the store account ID, but in some cases (e.g. platform checkout intent confirmation) we need a different account ID, so we'll need 2 Stripe JS SDKs initialized at this point.

> does it make sense to create a function to return the intent confirmation (confirmCardSetup/confirmCardPayment)?

Hmm, I like this idea 🤔 

Let's go with that and see if that improves things 👏

includes/class-wc-payment-gateway-wcpay.php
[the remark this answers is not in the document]
reykjalin: Fixed in e7a135899cb6ef7083ed267378e6decb7a0da9bf 🙌

includes/class-wc-payment-gateway-wcpay.php
ricardo: > If we overwrite the accountId that would give plugin authors (or anyone with write access to the site) the ability to overwrite the account ID used in the Stripe SDK.

Got it. Makes sense.

client/checkout/api/index.js
reykjalin: I tried this out in https://github.com/Automattic/woocommerce-payments/pull/3811/commits/c245507af26b1e264a92da94bd5f4e2e74981a5c, but not sure if it looks better 🤔 
What do you think?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1041,20 +1042,28 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
-				$payment_information->is_merchant_initiated(),
-				$additional_api_parameters,
-				$payment_methods
-			);
+			// phpcs:ignore
+			if ( ! empty( $_POST['platform-checkout-intent'] ) ) {
+				// If the intent is included in the request use that intent.
+				// phpcs:ignore
+				$intent_id = wp_unslash( $_POST['platform-checkout-intent'] );
+				$intent    = $this->payments_api_client->get_intent( $intent_id );
```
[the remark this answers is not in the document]
ricardo: Yeah, maybe `sanitize_key` 👍.

includes/class-wc-payment-gateway-wcpay.php
ricardo: Should we try to confirm or fix this bug before merging the PR?

ricardo (approved):
LGTM 👍

> I tried this out in https://github.com/Automattic/woocommerce-payments/commit/c245507af26b1e264a92da94bd5f4e2e74981a5c, but not sure if it looks better 🤔 What do you think?

I think it got a bit clearer 👍. It was mainly confusing to me because of the function name, and now we have a bit less code to process as well.

includes/class-wc-payment-gateway-wcpay.php
reykjalin: I'll see if it's still happening before merging and report back 🤔

includes/class-wc-payment-gateway-wcpay.php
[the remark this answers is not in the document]
reykjalin: Ah, turns out sanitize_key won't work since it filters out upper case characters and Stripe IDs may have upper case characters 😩

I think [sanitize_user( $intent_id, true /* enable strict mode */ )](https://developer.wordpress.org/reference/functions/sanitize_user/) is more appropriate since it will not replace upper case characters 🤔 

Made that change in df263b3e96c4c6f078a16a5ac52c1aaf31b71937

includes/class-wc-payment-gateway-wcpay.php
reykjalin: Just confirmed that this is still an issue. Very easy to reproduce;

Just run a payment. If you set a breakpoint here and inspect the value you'll see that it comes back as `"null"` 🤷‍♂️

PR #3816, opened 2022-02-16
Add tracking for enabling and disabling platform checkout

### LINKED ISSUE #3815: Tracking for enabling and disabling platform checkout
### Description

Record an event in Tracks whenever a merchant enables or disables platform checkout.

### Additional context

pdpOw2-l8-p2

brianyu28:
Fixes #3815

See pdpOw2-l8-p2

#3738 added a setting on the WCPay Settings page to enable and disable platform checkout. This PR records an event in Tracks each time the platform checkout is enabled or disabled.

cc @malithsen in case you'd like to review or have feedback as well.

#### Testing instructions

- In your `wp_options` table, ensure that the `_wcpay_feature_platform_checkout` option is set to `1`.
- In WP Admin, go to **Payments > Settings**.
- Look for the "Platform Checkout" option next to "Express checkouts."
- Enable the checkbox and "Save changes."
- In ~5 minutes, checks the Tracks Live View for an event called `wcadmin_platform_checkout_enabled`.
- Disable the checkbox and "Save changes."
- In ~5 minutes, checks the Tracks Live View for an event called `wcadmin_platform_checkout_enabled`.
- Click "Save changes" again.
- In ~5 minutes, check the Tracks Live View — there shouldn't be any new event, since the setting hasn't changed.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1597,6 +1597,19 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	}
 
 	/**
+	 * Updates whether platform checkout is enabled or disabled.
+	 *
+	 * @param bool $is_platform_checkout_enabled Whether platform checkout should be enabled.
+	 */
+	public function update_is_platform_checkout_enabled( $is_platform_checkout_enabled ) {
+		$current_is_platform_checkout_enabled = 'yes' === $this->get_option( 'platform_checkout', 'no' );
+		if ( $is_platform_checkout_enabled !== $current_is_platform_checkout_enabled ) {
+			wc_admin_record_tracks_event( $is_platform_checkout_enabled ? 'platform_checkout_enabled' : 'platform_checkout_disabled' );
+			$this->update_option( 'platform_checkout', $is_platform_checkout_enabled ? 'yes' : 'no' );
+		}
+	}
+
+	/**
 	 * Init settings for gateways.
 	 */
 	public function init_settings() {
--- a/changelog/add-platform-checkout-enable-disable-tracking
+++ b/changelog/add-platform-checkout-enable-disable-tracking
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Add tracking for enabling and disabling platform checkout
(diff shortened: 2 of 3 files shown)


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1596,6 +1596,19 @@ public function update_option( $key, $value = '' ) {
...
 
+	/**
+	 * Updates whether platform checkout is enabled or disabled.
+	 *
+	 * @param bool $is_platform_checkout_enabled Whether platform checkout should be enabled.
+	 */
+	public function update_is_platform_checkout_enabled( $is_platform_checkout_enabled ) {
+		$current_is_platform_checkout_enabled = 'yes' === $this->get_option( 'platform_checkout', 'no' );
+		if ( $is_platform_checkout_enabled !== $current_is_platform_checkout_enabled ) {
+			wc_admin_record_tracks_event( $is_platform_checkout_enabled ? 'platform_checkout_enabled' : 'platform_checkout_disabled' );
```
a-danae, line 1607: Looks like we're using that prefix for all events in WCPay except for `page_view`. [Here are the ones](https://github.com/Automattic/woocommerce-payments/blob/develop/client/tracks/index.js#L31) recorded on the client-side, for example. So I'm wondering if the event name should be prefixed with `wcpay` so it ends up being `wcadmin_wcpay_platform_checkout_{action}`.

Not sure if this was discussed or we decided not to use it in this case, though. What do you think?

a-danae (approved):
Thanks for working on this @brianyu28!

I've followed the testing instructions and the events are recorded in Tracks as described. I just left a question about the event name, but I'm approving because it's a question we might or might not want to address rather than a change request.

LGTM 👍

includes/class-wc-payment-gateway-wcpay.php
brianyu28: Thanks for raising the question — for these events, we've been excluding `wcpay` from the event name so that platform checkout events could be recorded in other places in the future, even though they'll only be in WCPay for now.

a-danae: Makes sense. Thanks for clarifying!

PR #3818, opened 2022-02-16
Fix email download for deposit transactions.

### LINKED ISSUE #3777: "Deposit Transaction" reports get replaced with "All Deposits" report upon download request
### Describe the bug

The issue began at the same time that WooCommerce Payments began _emailing_ the "All Deposits" instead of directly downloading it in the browser (about 4 weeks ago for this user). I believe that transition is based on file size or total number of transactions.

Instead of giving a CSV of the `Deposit Transactions` report showing the data for a single deposit, they are emailed the full `Deposits` report. 

e.g. The `Deposits Transaction` view shows a total of $3,000, with 10 `transactions` making up the deposit. The report that gets emailed upon request to download does not match this data (i.e. Contains no customer transactions) and only shows all 8000 `Deposits` from WooCommerce Payments instead. 

### To Reproduce

I am personally unable to reproduce it since I don't have a site with enough WooCommerce Payments transactions to trigger the changeover from direct download to emailing of reports.

To download a `Deposit Transactions` report: 
1. Navigate to **Payments → Deposits**
2. Click on the date for the Deposit in question to open the `Deposit Transactions` view.
3. Click the `Download` link in the top right of the list.
4. Review the CSV file that is downloaded or emailed, and confirm it does not match the data in the view.

### Actual behavior

The wrong report is emailed when requesting a download of a "Deposit Transactions" view.

### Screenshots

WCPay-deposit-transactions-example

### Expected behavior

A report containing only the data from that view should be downloaded (or emailed). Considering the small amount of data contained in a single `Deposit Transactions` view on average, I would think direct download should be possible.

### Additional context

4766625-zen

jessepearson:
Fixes #3777 

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/client/transactions/list/index.tsx
+++ b/client/transactions/list/index.tsx
@@ -446,6 +446,7 @@ export const TransactionsList = (
 				type_is: typeIs,
 				type_is_not: typeIsNot,
 			} = params;
+			const depositId = props.depositId ?? '';
 
 			const isFiltered =
 				!! dateAfter ||
@@ -481,6 +482,7 @@ export const TransactionsList = (
 							search,
 							typeIs,
 							typeIsNot,
+							depositId,
 						} ),
 						method: 'POST',
 					} );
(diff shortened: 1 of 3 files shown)


oaratovskyi (approved):
Thanks @jessepearson, it works very well! 
I even went further - tested it with two deposits they both had more than one transaction.
You can :shipit: !

PR #3827, opened 2022-02-18
Webhook Reliability: Fetch failed events and process them

### LINKED ISSUE #3813: Periodically fetch and process failed webhooks from the server
### Description

This PR should work with the response, provided by the server in 1631-gh-Automattic/woocommerce-payments-server.

### Acceptance criteria

- Use the action scheduler to periodically fetch failed webhooks from the server, and process them as if they just arrived.
- This issue does not focus on the idempotency of the incoming webhooks.
- We should specify a limited list of handled events, and only handle one of them as a PoC within this issue.

When testing, any available event from the server should be processed automatically and ideally asynchronously.

htdat:
Fixes #3813

Add a new service to achieve by doing these things: 

- Hook into `woocommerce_payments_account_refreshed` and check the relevant flag, and decide whether to schedule a job `wcpay_webhook_fetch_events`. Relevant convo p1645601664622659/1645086273.563649-slack-C01BPL3ALGP
- `wcpay_webhook_fetch_events` jobs fetches events from the server, then do two things: 
     - Decide whether to schedule another `wcpay_webhook_fetch_events`.
     - Save data and schedule a job for each event.
- `wcpay_webhook_process_event` jobs simply retrieve data and dispatch to `WC_Payments_Webhook_Processing_Service` (see this PR https://github.com/Automattic/woocommerce-payments/pull/3852)

### Testing instructions

Requirement: this server PR 1633-gh-Automattic/woocommerce-payments-server
ActionScheduler management pending page: wp-admin/tools.php?page=action-scheduler&status=pending

#### Test case: properly schedule `wcpay_webhook_fetch_events`

**TODOs:** 

- [x] Wait for the server PR to implement this. Otherwise, you can apply this diff for testing 2bd57-pb/#diff - change true/false to see the difference. 
- [x] Update this value `CONTINUOUS_FETCH_FLAG_ACCOUNT_DATA` based on the server PR.

Steps: 

- Follow the server PR, and ensure that your site has some failed not-yet-processed webhook events.
- Go to ActionScheduler management page and run `wcpay_refresh_account_cache`, which forces to retrieve new account data. 
- Expect: a new job `wcpay_webhook_fetch_events` is scheduled. 

#### Test case: run `wcpay_webhook_fetch_events`

- Ensure that your site has some not-yet-processed failed webhook events.
- Go to ActionScheduler management page and run `wcpay_webhook_fetch_events`.
- Expect: failed events are retrieved from the server, and new jobs `wcpay_webhook_process_event` are scheduled with the `event_id` parameter. 
- Try adding more than 10 not-yet-processed failed webhook events. 
- Go to ActionScheduler management page and run `wcpay_webhook_fetch_events`.
- Expect: similar above plus that another `wcpay_webhook_fetch_events` is scheduled (This is to test the `has_more` flag).
- Expect: Go to wp-content/uploads/wc-logs (if you use Docker, it's under the `docker/wordpress` folder), find the log file for today, and see the relevant log entries. Tip: use `event_id` for search.

#### Test case: run `wcpay_webhook_process_event`. 

- Go to ActionScheduler management page and run a `wcpay_webhook_process_event` job.
- Expect: Also check in the log file and see the relevant log entries for `event_id` attached to the job you've just run. 

-------------------

**Post merge**


includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -0,0 +1,91 @@
...
+
+		add_action( self::WEBHOOK_FETCH_EVENTS_ACTION, [$this, 'fetch_events'] );
+		add_action( self::WEBHOOK_PROCESS_EVENT_ACTION, [$this, 'process_event'] );
+		add_action( 'woocommerce_payments_account_refreshed', [ $this, 'maybe_schedule_fetch_events'] );
+	}
+
+	private function fetch_events() {
+		$payload = $this->payments_api_client->get_failed_webhook_events();
+
+		$remaining_failed_events = (int) $payload['failed_webhook_count'] ?? 0; // TODO - maybe `failed_webhook_count` (int) or `has_more` (bool). See server PR 1633
```
RadoslavGeorgiev, line 57: We will not be using `COUNT(*)` in DB queries, see `$count + 1` in 1633-gh-Automattic/woocommerce-payments-server/files#diff-91becda04a3d12e72f51da7e3f60e14114fc78c8c1090c93af5de85c19b77155R91. `has_more` will simply be boolean in order to know whether to schedule another batch or not.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+
+	public function schedule_process_event( array $event_data ) {
+		$this->action_scheduler_service->schedule_job( time(), self::WEBHOOK_PROCESS_EVENT_ACTION, $event_data );
+	}
+
+	public function process_event( array $event_data ) {
+		// TODO with the following pseudo-code
+		// Verify data?
```
RadoslavGeorgiev, line 86: I don't think we need to verify the data here, except for integrity maybe. We're already using the JP's connection package to reach the server, I think we can trust the response.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -0,0 +1,91 @@
...
+		// Create a request based on this data. Dispatch to Webhook_Controller
```
RadoslavGeorgiev, line 87: IMO it would be better to isolate the core event handling from `WC_REST_Payments_Webhook_Controller::handle_webhook()` into a separate method, which can be called here. The code between and including the `woocommerce_payments_before_webhook_delivery` and `wocommerce_payments_after_webhook_delivery` actions seems like a good candidate. The rest (reading REST properties and error handling) should remain there and be specific to standard webhook delivery. This method (`process_event`) should have it's own event handling, as we're not responding to a call from the server here.

includes/class-wc-payments-webhook-reliability-service.php
htdat: > I don't think we need to verify the data here, except for integrity maybe.

Yes, this is what I mean here, not for the security reason.

includes/class-wc-payments-webhook-reliability-service.php
htdat: > This method (process_event) should have it's own event handling, as we're not responding to a call from the server here.

Agreed! However, it's tricker in the implementation. I even tried to build an instance of `WC_REST_Payments_Webhook_Controller` then call `handle_webhook()`. However, it's not working as the whole REST API code is not loaded when ActionScheduler runs a job.

> The code between and including the woocommerce_payments_before_webhook_delivery and wocommerce_payments_after_webhook_delivery actions seems like a good candidate.

This is a good point and it may require more refactor to `WC_REST_Payments_Webhook_Controller` so that it can work within an AS job too.

htdat: > That said, it's still possible but it requires more large effort to refactor WC_REST_Payments_Webhook_Controller for our purpose.

Here it is https://github.com/Automattic/woocommerce-payments/pull/3852

htdat:
Updated the PR description, ready for the first round of review.

RadoslavGeorgiev (approved):
Once again, awesome work, @htdat! 👏 

Immaculate code without any need for changes. The code itself looks excellent and the tests seem to properly cover all needed scenarios.

In terms of testing, I implemented the proper change to the server, which got commited in 1633-gh-Automattic/woocommerce-payments-server/commits/5876c7e8ddb569bde4f00ff95dd5aaa31bc11b39. Once I did that, everything worked flawlessly: 🎊 

0. Temporarily decreased the amount of events returned by the server endpoint to 5 to ease testing.
1. Enabled the setting in the server's `wp-admin` to intentionally store all webhooks as failed.
2. Made a few payments with a card, which automatically triggers disputes. This added up to 11 webhooks in my local database.
3. Checked the orders to confirm that they appear as paid and not disputed (as the webhooks did not reach the client).
4. Triggered the account refresh, which automatically scheduled the fetch action.
5. Triggered the fetch action once, and 5 `process` actions appeared.
6. Triggered those pending actions one by one while refreshing all of those orders, and their notes started appearing one by one, as expected.
7. Repeated the last few steps two more times to fetch the remaining 5 + 1 webhooks. At this point all orders had their disputes processed.
8. Verified that there are no more fetches scheduled.
9. Paid for another order with the same card before triggering the same set of actions. Everything went well, and at the end the slate was clean both for the client, and the server.

The PR is approved, but don't :shipit: yet. I've marked it as blocked until the server gets ready. If for some reason the server is not ready by release time, we can also deploy dummy endpoint(s) for the server, in order to be able to ship client PRs, and activate them later.

deepakpathania:
Removing the blocked status from this now since the server PR is ready, we'll merge this soon so that GS can test the other flows with this merged - this itself does not have any testing instructions because, without the server changes, we can't test it. cc @htdat

PR #3832, opened 2022-02-21
Fix platform checkout iframe positioning

brettshumaker:

This PR handles positioning the iframe popover positioning. It will position it at the right edge of the email input field unless the iframe window would be < 50px from the right edge of the screen. At that point, it will stick to 50px from the right edge of the screen. If the window is below the mobile breakpoint it resets the `left` and `right` properties and does nothing else. For situations when the position of the email field would cause the iframe popover to be off the top edge of the screen, the window is scrolled down and the `top` property of the iframe popover is adjusted.

#### Testing instructions

1. Add product to cart.
2. Go to checkout.
3. Enter a valid email address from your platform checkout environment.
4. Make sure that as you resize the window that the iframe popover behaves as described above.

### DIFF
--- a/changelog/fix-platform-checkout-iframe-positioning
+++ b/changelog/fix-platform-checkout-iframe-positioning
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Made change to functionality behind a feature flag and is not yet public.
+
+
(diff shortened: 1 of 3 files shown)


frosso (approved):
🧮

PR #3834, opened 2022-02-21
Hook user/exists endpoint to UI for saving data to platform checkout

brettshumaker:

This PR creates and dispatches a new custom event when a response is returned from the `user/exists` platform checkout endpoint with the `isRegisteredUser` set to `data[ 'user-exists' ]`. A listener for that event has also been added to the component that allows users to save their data for faster checkout. If the user exists, that section is not shown. Otherwise, it is.

#### Testing instructions

1. Add a product to your cart.
2. Go to the shortcode checkout.
3. Verify that the "Remember your details?" section is shown.
4. Enter an email address that exists on the platform checkout and verify that the section is **_not_** rendered.
4. Enter an email address that **_does not exist_** on the platform checkout and verify that the section **_is_** rendered.

### DIFF
--- a/client/components/platform-checkout/hooks/use-platform-checkout-user.js
+++ b/client/components/platform-checkout/hooks/use-platform-checkout-user.js
@@ -27,9 +27,7 @@ const usePlatformCheckoutUser = () => {
 		};
 	}, [] );
 
-	return {
-		isRegisteredUser,
-	};
+	return isRegisteredUser;
 };
 
 export default usePlatformCheckoutUser;
--- a/client/components/platform-checkout/hooks/use-platform-checkout-user.js
+++ b/client/components/platform-checkout/hooks/use-platform-checkout-user.js
@@ -1,6 +1,31 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useRef, useState } from 'react';
+
 // hook for handling API calls to get and create platform checkout user.
 const usePlatformCheckoutUser = () => {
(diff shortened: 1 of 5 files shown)


client/components/platform-checkout/hooks/use-platform-checkout-user.js
```diff
@@ -1,6 +1,31 @@
...
+
+		return () => {
+			currentWindowRef.removeEventListener(
+				'PlatformCheckoutUserCheck',
+				handlePlatformCheckoutUserCheck
+			);
+		};
+	}, [] );
 
 	return {
```
frosso, line 30: What are your thoughts on also just returning the boolean value here, rather than an object containing the boolean?
The boolean can be uglified by webpack, whereas an object and its properties can't

frosso (approved):
Gg

client/components/platform-checkout/hooks/use-platform-checkout-user.js
brettshumaker: That's a good point. I sat here for a while trying to think of a use case where we would need to add something to the return value in the future and came up empty. We can always change what's returned in the future anyway. I'll make that update to the hook then update the usage and tests to expect an unnamed boolean. 👍

brettshumaker: Resolved in b1f7ace3

PR #1883, opened 2021-05-21
feat: disable WCPay confirmation dialog

frosso:
Fixes https://github.com/Automattic/woocommerce-payments/issues/1564

#### Changes proposed in this Pull Request

Added a modal to display when the merchant attempts to disable WCPay.
Solution for the event listener has been proposed by @superdav42 here: https://github.com/Automattic/woocommerce-payments/issues/1564#issuecomment-837473858

#### Testing instructions

- Ensure you have the latest version of the dev tools: `cd docker/wordpress/wp-content/plugins/woocommerce-payments-dev-tools && git checkout trunk && git pull`
- Ensure you have the "grouped settings" flag enabled in WCPay Dev Tools
- (optional) Ensure you have the "Add UPE additional payment methods" flag enabled in WCPay Dev Tools
- Go to the list of gateways: http://localhost:8082/wp-admin/admin.php?page=wc-settings&tab=checkout&section
- Attempting to disable WCPay shows the modal
- The modal is not displayed when enabling WCPay
- Cancelling the action on the modal does not disable WCPay
- Confirming the action on the modal disables WCPay

-------------------

- [ ] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

PR #3848, opened 2022-02-23
Remove interactions with the dev tools setting "grouped settings" in E2E tests

### LINKED ISSUE #3845: E2E tests breaking due to grouped settings removed
### Describe the bug
After merging 49-gh-Automattic/woocommerce-payments-dev-tools, we have failing tests.

See [this failed run](https://github.com/Automattic/woocommerce-payments/runs/5298764957?check_suite_focus=true) for reference.

jbordonado:
Fixes #3845 

Remove any interactions in our E2E tests with the old "Enable grouped settings" option in the WCPay Dev tools page.

This control was removed both from the plugin and dev tools because it was not used anymore.

#### Testing instructions

`npm run test:e2e-dev tests/e2e/specs/wcpay/merchant/merchant-payment-gateways-confirmation.spec.js` works fine after the changes (it was the only file breaking).

-------------------

**Post merge**


jbordonado:
I triggered a manual run which was successful: https://github.com/Automattic/woocommerce-payments/actions/runs/1888506436

deepakpathania:
@jbordonado I checked the changes and can't get the test to pass locally, I get the following error message:
```
    TimeoutError: Element [aria-label='The "WooCommerce Payments" payment method is currently enabled'] (text: "Yes") not found
```

I think the check you removed was used to disable the confirmation dialog in [this PR](https://github.com/Automattic/woocommerce-payments/pull/1883) which the test cannot find now based on the error message above. Could you check this once?

jbordonado:
> @jbordonado I checked the changes and can't get the test to pass locally, I get the following error message:
> 
> ```
>     TimeoutError: Element [aria-label='The "WooCommerce Payments" payment method is currently enabled'] (text: "Yes") not found
> ```
> 
> I think the check you removed was used to disable the confirmation dialog in [this PR](https://github.com/Automattic/woocommerce-payments/pull/1883) which the test cannot find now based on the error message above. Could you check this once?

I think this is not related to my changes because all I did was remove the initial step where we go the WCPay Dev tools to enable the grouped settings (which has no impact on the plugin, we don't rely anymore on this setting which is why we deleted the option from dev tools).

This error means that the test expected that the WooCommerce Payments toggle is "on" (enabled), and it isn't.

To be honest, when I checkout the branch to check this and run directly the test, I had the same errors as you (yesterday it worked fine, and [on the CI it ran fine too](https://github.com/Automattic/woocommerce-payments/actions/runs/1888506436)). I noticed that from the beginning, WCPay is disabled and that is not what the test expects so of course it fails.

I'm wondering if there are some cleanup steps missing in the afterEach() of some tests, because it seems like we end up with the extension disabled after running some tests, hence when we try again to run the tests, it won't work anymore.

If I do again a fresh setup (`npm run test:e2e-setup`), then it works fine.

Can you confirm on your side?

deepakpathania (approved):
I reset the test setup once and it seems to consistently pass for me now, I guess we could monitor this in CI for some time to see how it behaves - I am approving this for now. @jbordonado 

cc @achyuthajoy in case you want to verify the same as well once.

changelog/fix-3845-update-breaking-e2e-tests-grouped-settings
```diff
@@ -0,0 +1,5 @@
+Significance: patch
```
achyuthajoy, line 1: A change log entry may not be required for the change since it's related to devops. Also the change is based on the Dev Tools plugin which is not public as well

achyuthajoy (approved):
Left a comment regarding change log. I'll leave it to you whether or not to remove it 😄 

Changes approved. I've already tested your changes on another CI branch. Works as expected - https://github.com/Automattic/woocommerce-payments/actions/runs/1920522059

changelog/fix-3845-update-breaking-e2e-tests-grouped-settings
jbordonado: Indeed, I removed it now. Thanks

PR #3869, opened 2022-02-28
Check the merchant's account before showing the multi-currency inbox note

### LINKED ISSUE #3806: Do not show multi-currency inbox note until merchant has successfully set up WCpay
We're showing an inbox note when WooCommerce payments is installed, but setup is not complete. We should not show the inbox note until the merchant as successfully created a WooCommerce Payments account. If the note is clicked, you're taken to a page that says "You're not allowed to access this page"

Screen Shot 2022-02-15 at 11 35 28 AM
Screen Shot 2022-02-15 at 11 36 03 AM

@vbelolapotkov

jbordonado:
Fixes #3806 

Do not show the multiple currencies inbox note if the merchant doesn't have an account set up yet.

We need this change because the merchant cannot set up multiple currencies if he doesn't have a connected account set up. At the moment, if he clicks on the note, he will be redirected to a page that says "Sorry, you are not allowed to access this page". After this change, he will not be shown this note before he actually has an account set up.

At first, I thought of adding the condition in `MultiCurrency.php > add_woo_admin_notes()`.
But then, I ran into another inbox note (`class-wc-payments-notes-additional-payment-methods.php`) that had the same requirement and checked the account only during the `get_note()` step so I followed the same style.

#### Testing instructions

- Create a fresh JN site with WooCommerce and WCPay Dev Tools plugins
- Install the WooCommerce Payments plugin from this branch
- Acknowledge that there is NOT a multi currency note in the inbox
- Set up your WCPay account
- Acknowledge that there is now a multi currency note in the inbox

-------------------

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.9.0


includes/multi-currency/Notes/NoteMultiCurrencyAvailable.php
```diff
@@ -28,10 +29,22 @@ class NoteMultiCurrencyAvailable {
...
+	 * @var WC_Payments_Account
+	 */
+	private static $account;
+
 	/**
 	 * Get the note.
 	 */
 	public static function get_note() {
+		// Don't show the note if the merchant didn't create a WCPay account yet.
+		if ( ! self::$account->is_stripe_connected() ) {
```
htdat, line 44: Moving this condition to `possibly_add_note` makes more sense and ensure that this method is a pure getter without any side effects. 

The same pattern is being used in `WC_Payments_Notes_Loan_Approved::possibly_add_note` - [src](https://github.com/Automattic/woocommerce-payments/blob/1d5ce3ce89c9a2513c036c23fd5a5c4fa6d266c4/includes/notes/class-wc-payments-notes-loan-approved.php#L88-L88).

Also, there is a chance that `self::$account` is not set yet at this point, so checking it too makes code a bit more defensive. 

```suggestion
		if ( is_null ( self::$account ) || ! self::$account->is_stripe_connected() ) {
```

htdat (changes requested):
Great work and research on this issue. 

Followed tests in `Testing instructions`, and confirmed the PR works correctly.

I have only one comment to make code more logical and a bit more defensive around the logic ;)

includes/multi-currency/Notes/NoteMultiCurrencyAvailable.php
jbordonado: Agreed that it would be best to move the check into possibly_add_note, done.
Also added an additional check on the account as you suggested to be sure it won't break.

includes/multi-currency/Notes/NoteMultiCurrencyAvailable.php
```diff
@@ -51,4 +59,30 @@ public static function get_note() {
...
 	}
+
+	/**
+	 * Add the note if it passes predefined conditions.
+	 */
+	public static function possibly_add_note() {
+		// Don't add the note if the merchant didn't create a WCPay account yet.
+		if ( is_null( self::$account ) || ! self::$account->is_stripe_connected() ) {
+			return;
+		}
```
htdat, line 70: However, I see that you've removed relevant tests too. I think it's worth adding tests again to cover the logic here. 

We can use `NoteTraits::note_exists()` to verify if the note has been added or not. WDYT?

htdat (commented):
Thanks for making the change. I have a question regarding tests. It would be great if we can add them back.

includes/multi-currency/Notes/NoteMultiCurrencyAvailable.php
jbordonado: Sounds good, I added back some tests to cover this piece of logic, thanks!

htdat (approved):
🚢 it! Thanks for adding back the tests.

Issue #3807, opened 2022-02-15
Remove platform checkout code that affects checkout block


harriswong:
### Description
We have `isPlatformCheckoutEnabled` feature flag wrapped around code for platform checkout functionalities. Some of these flags are integrated with the "checkout block".

Checkout block no longer needs to support platform checkout, we should remove the flag and the code inside. For example, this can be removed. 
https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe.js#L60-L65

### Acceptance criteria
- Remove https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe.js#L60-L65
- Find other places where `isPlatformCheckoutEnabled` is **used with checkout block**, remove those.

harriswong:
@frosso, I want to double check with you regarding this issue. The checkout block code is functional, but we would still like to remove it because we only want to support shortcode checkout right?

cc @brettshumaker

frosso:
@harriswong that's correct, let's remove everything related to modifications in the "checkout" block made in the WC Payments plugin 👍

PR #3870, opened 2022-03-01
Remove platform checkout feature flag for upe

harriswong:
Fixes https://github.com/Automattic/woocommerce-payments/issues/3807

Remove platform checkout feature flagged codes for UPE checkout block.

#### Testing instructions
1. With "shortcode checkout", when you finished typing the email address in the checkout form, you should see an API request that goes out to check if there is a user in platform checkout. 
2. With "block" checkout, when you finished typing the email address in the checkout form, there is no API request going out.

-------------------

**Post merge**


harriswong:
@superdav42 @waclawjacek We are removing the platform checkout feature flag around UPE block. Please let me know if there are other places we should also remove.

superdav42:
> @superdav42 @waclawjacek We are removing the platform checkout feature flag around UPE block. Please let me know if there are other places we should also remove.

Why are we doing this? Is there something in UPE that is not compatible with platform checkout?

superdav42:
> @superdav42 @waclawjacek We are removing the platform checkout feature flag around UPE block. Please let me know if there are other places we should also remove.

Why are we doing this? Is there something in UPE that is not compatible with platform checkout?

harriswong:
> Is there something in UPE that is not compatible with platform checkout?

@superdav42 It's from [this](https://github.com/Automattic/woocommerce-payments/issues/3807#issuecomment-1047342578). 

@frosso is it because block checkout is not fully supported with platform checkout, so that when we enable the feature flag, block checkout remains untouched?

brettshumaker (changes requested):
Looks like what you have here covers UPE, but we still need to remove it from the blocks checkout ([here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/index.js))?

harriswong:
Thanks @brettshumaker. I removed the flags from checkout/block/index.js. The only one left should be the classic one
image
Please take a look again.

brettshumaker (changes requested):
Awesome, thanks for taking that out.

I did just notice that we're still rendering the "Remember your details?" section when UPE is enabled. Should we be?

frosso:
> Why are we doing this? Is there something in UPE that is not compatible with platform checkout?
> @frosso is it because block checkout is not fully supported with platform checkout, so that when we enable the feature flag, block checkout remains untouched?

There are a couple of things here:
- In order for the platform checkout to work, we'll need to explore some changes to the UPE implementation ( paJDYF-3gI-p2 )
- The `_wcpay_feature_platform_checkout` flag controls both the "shortcode" and the "blocks" checkout. We'll start only supporting the "shortcode" checkout and make some changes to support the "blocks" checkout later

harriswong:
> I did just notice that we're still rendering the "Remember your details?" section when UPE is enabled. Should we be?

Thanks @brettshumaker. No. "Remember your details" should be behind the feature flag. I will create a separate issue for this.

brettshumaker (approved):
🚢

Issue #3862, opened 2022-02-25
WCPay Admin Pages doesn't load with WooCommerce 4.8.0


achyuthajoy:
### Describe the bug

We've recently merged a [PR](https://github.com/Automattic/woocommerce-payments/pull/3838) that fixes picking up the correct WordPress & WooCommerce versions with setting up the WCPay E2E environment. After the merge, we noticed that WCPay merchant tests for WooCommerce 4.8.0 are failing along with the tests for WC Blocks.

### Testing Instructions

* On your test site, install WooCommerce 4.8.0 along with the latest version of WooCommerce Payments. 
* Access any WCPay pages from backend to see a blank page with the error mentioned below.

### Actual behavior

With WooCommerce 4.8.0 installed on a JN site along with the latest version of WCPay, accessing any WCPay admin page returns blank with the error:

```
Uncaught TypeError: can't access property "plugins", window.wcSettings is undefined
    297 http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    a http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    561 http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    561 http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    a http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    app http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
    <anonymous> http://localhost:8084/wp-content/plugins/woocommerce/packages/woocommerce-admin/dist/app/index.js?ver=1.7.3:2
```

For failures related to WC Blocks can be ignored since WC Blocks requires a higher version of WooCommerce.

### Screenshots

Screencast - https://d.pr/v/rjgPqj

### Expected behavior

Since WooCommerce 4.8.0 is still the minimum supported version for WCPay, all WCPay pages should load as expected.

jessepearson:
[...]

PR #3879, opened 2022-03-02
E2E - Update GH workflow to fix compatibility with WC 4.8.0

achyuthajoy:
### Problem

WooCommerce 4.8.0 has compatibility issues with WooCommerce Blocks package (#3862) causing the E2E pipeline to fail. Since we install the latest version of WooCommerce blocks package during E2E environment setup, we need to update the workflow to skip specific tests based on the WooCommerce version.

Introduces a new step on GH workflow to add environment variable `SKIP_E2E_WORKFLOW` based on test versions and configured rest of the steps to check for the same. If present, the steps will be skipped marking the test as successful in GH actions.

For WC 4.8.0, WCPay - Merchant & Blocks - Merchant tests are skipped. Also during setup, WC Blocks package won't be installed.

Additionally, `WP_DEBUG` errors causes problems with tests on WC 4.8.0. To fix this problem, `WP_DEBUG_DISPLAY` is set to `false` & `WP_DEBUG_LOG` is set to `true` when `DEBUG` flag is not present. The workflow is also configured to upload `debug.log` as an artifact.

### Testing changes

* Trigger a  new E2E Test against this branch from Actions tab. Reference Job - https://github.com/Automattic/woocommerce-payments/actions/runs/1920915076
* Confirm that `WC - 4.8.0 | wcpay - merchant` & `WC - 4.8.0 | blocks - shopper` tests passes but steps are skipped.
* Confirm that rest of the tests for WC 4.8.0 are passing (requires #3848 to be merged). Reference Job - https://github.com/Automattic/woocommerce-payments/actions/runs/1920522059

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -56,11 +56,18 @@ jobs:
       E2E_BRANCH: ${{ matrix.test_branches }}
 
     steps:
-      # Set conditional ENV Variables
-      - name: Set conditional variables
-        if: ${{ matrix.woocommerce == '4.8.0' }}
+      # Conditionally skip workflow. Remove/update based on min supported WC version.
+      - name: Conditionally skip workflow
         run: |
-          echo "SKIP_WC_BLOCKS_TESTS=1" >> $GITHUB_ENV
+          if [[ $E2E_WC_VERSION == '4.8.0' ]]; then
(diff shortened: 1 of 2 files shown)


deepakpathania (changes requested):
Changes look good and works as intended, thanks for working on this, left one small comment.

deepakpathania (approved):
LGTM 🚢

PR #3881, opened 2022-03-02
Fix woocommerce_order_actions when order is empty

### LINKED ISSUE #2704: Fatal error when the `woocommerce_order_actions` filter is evaluated out of "edit order" context
`\WC_Payment_Gateway_WCPay::add_order_actions()` which is hooked to `woocommerce_order_actions` always expects a global `$theorder` variable to be an instance of an order. 

This will cause a fatal error if the `woocommerce_order_actions` filter is evaluated by some other plugin or custom code. The value of `$theorder` should be type checked before its used.

First discovered in AutomateWoo: 1017-gh-woocommerce/automatewoo

alefesouza:
Fixes #2704

* Fix fatal error when `woocommerce_order_actions` is called by a plugin or custom code without an order.

See 1017-gh-woocommerce/automatewoo and 1031-gh-woocommerce/automatewoo for more information.

#### Testing instructions

To test it we will need to add `do_action( 'woocommerce_order_actions', [] );` to the theme functions.php file, or create a new plugin that call `woocommerce_order_actions`.

* Create a `dummy-plugin` folder on the `wp-content/plugins`.
* Create a `dummy-plugin.php` file.
* Add this content to the file:

```php
<?php

/**
 * Plugin Name:       My Basics Plugin
 */

do_action( 'woocommerce_order_actions', [] );
```

* Enable the plugin on WordPress admin.
* Without these modifications, it will throw a error and not allow to enable the plugin.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -17,6 +17,7 @@ use WCPay\Constants\Payment_Initiated_By;
 use WCPay\Constants\Payment_Capture_Type;
 use WCPay\Tracker;
 use WCPay\Payment_Methods\UPE_Payment_Gateway;
+use WC_Order;
 
 /**
  * Gateway class for WooCommerce Payments
@@ -1856,6 +1857,10 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	public function add_order_actions( $actions ) {
 		global $theorder;
 
+		if ( ! is_object( $theorder ) ) {
+			$theorder = new WC_Order();
+		}
+
 		if ( $this->id !== $theorder->get_payment_method() ) {
 			return $actions;
 		}
@@ -1868,6 +1873,9 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			'capture_charge'       => __( 'Capture charge', 'woocommerce-payments' ),
 			'cancel_authorization' => __( 'Cancel authorization', 'woocommerce-payments' ),
 		];
+
+		// Clear the dummy order.
+		$theorder = null;
 
 		return array_merge( $new_actions, $actions );
 	}
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -17,7 +17,6 @@ use WCPay\Constants\Payment_Initiated_By;
 use WCPay\Constants\Payment_Capture_Type;
 use WCPay\Tracker;
 use WCPay\Payment_Methods\UPE_Payment_Gateway;
-use WC_Order;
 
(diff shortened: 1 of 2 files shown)


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1856,6 +1856,10 @@ public function validate_account_statement_descriptor_field( $key, $value ) {
 	public function add_order_actions( $actions ) {
 		global $theorder;
 
+		if ( ! is_object( $theorder ) ) {
```
leonardola, line 1859: This can lead to a bug in the future in case the next if is changed or code is added between the ifs. It is probably better to actually validate if the order is an object and in case it is not just return the $actions instead of trying to execute the rest of the code with an invalid or dummy order. This conditional can be added inside the next if as it already returns

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1869,6 +1873,9 @@ public function add_order_actions( $actions ) {
 			'cancel_authorization' => __( 'Cancel authorization', 'woocommerce-payments' ),
 		];
 
+		// Clear the dummy order.
```
leonardola, line 1876: in case the changes proposed above are applied then this line must be removed.

leonardola (changes requested):

includes/class-wc-payment-gateway-wcpay.php
alefesouza: The code is inspired on the 1031-gh-woocommerce/automatewoo code changes, I think that when another plugin call `woocommerce_order_actions` on WCPayments it wants to call the `capture_charge` and the `cancel_authorization` that is added on the following code even when we do not a global order, what do you think?

leonardola: That PR in AutomateWoo repository is just a patch meant to temporary fix the error without having to wait for a fix on WC Payments. Basically it creates a dummy order just so when the filter `woocommerce_order_actions` is called the function `add_order_actions` has an order to check and return without doing anything. 

As we are fixing WC Payments we don't need the dummy order anymore. We actually want to check if the `$the_order` exists and if it does not just return the $actions for the next function to use. This way they can remove the dummy order in a future PR (which we will need to create an issue for on AutomateWoo). 

> I think that when another plugin call woocommerce_order_actions on WCPayments it wants to call the capture_charge and the cancel_authorization that is added on the following code even when we do not a global order, what do you think?

As it is right now with the dummy order the `capture_charge` and `cancel_authorization` will not be added as the payment method is empty. So basically we are not using the dummy order and returning before removing it. So this could have some side effects.

leonardola (approved):

leonardola:
Nice work on this one. LGTM 🚢

PR #3891, opened 2022-03-03
Clear cart for fraudulent payments

KarlisJ:
Fixes 1659-gh-Automattic/woocommerce-payments-server#

If the server returns an error response indicating that payment is fraudulent we'll invalidate the current session.

image

-------------------

**Post merge**

### DIFF
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -9,6 +9,7 @@ defined( 'ABSPATH' ) || exit;
 
 use WCPay\Exceptions\API_Exception;
 use WCPay\Exceptions\Amount_Too_Small_Exception;
+use WCPay\Exceptions\Fraudulent_Payment_Exception;
 use WCPay\Logger;
 use Automattic\WooCommerce\Admin\API\Reports\Customers\DataStore;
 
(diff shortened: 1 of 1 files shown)


mordeth (approved):
LGTM!

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1817,6 +1817,10 @@ protected function extract_response_body( $response ) {
 					$response_code
 				);
 			} elseif ( isset( $response_body['error'] ) ) {
+				if ( isset( $response_body['error']['decline_code'] ) && 'fraudulent' === $response_body['error']['decline_code'] ) {
+					WC()->cart->empty_cart();
+					WC()->session->set( 'reload_checkout', true );
```
kalessil, line 1822: Sorry for being late for the review. What does `WC()->session->set( 'reload_checkout', true );` do?

KarlisJ: ```WC()->session->set( 'reload_checkout', true );```

The front end upon receiving a response will refresh the page. It basically displays the error messages.

KarlisJ: > Also, we should consider keeping the cart content, would it be possible?

If we trust the fraudulent response then emptying the cart is the best option to do as it forces to perform several steps. 

WP nonce is tied to session and time. We can't regenerate a new nonce that could be used with the existing cart.

kalessil: I see. This could be an option, but we need to do a little bit more here: particularly we need a setting that will enable/disable the protection.

Hence, let's revert the PR and re-open the issue for the next iteeration.

PR #3893, opened 2022-03-03
Fix fatal error when a subscription renews automatically

### LINKED ISSUE #3892: Fatal error when trying to remove intent from session during `process_payment`
### Describe the bug
[`UPE_Payment_Gateway::remove_upe_payment_intent_from_session`](https://github.com/Automattic/woocommerce-payments/blob/3.8.1/includes/payment-methods/class-upe-payment-gateway.php#L1129-L1134) is called via the 
`woocommerce_order_payment_status_changed` action hook, which is called when there's a status change on an order. When subscriptions are renewed, this action happens, but there is no session to act upon.

> 2022-03-03T01:30:05+00:00 CRITICAL Uncaught Error: Call to a member function __unset() on null in .../wp-content/plugins/woocommerce-payments/includes/payment-methods/class-upe-payment-gateway.php:1133

We can fix this by checking to see if a session exists first.

### Additional context
p1646321924932389-slack-C7U3Y3VMY

jessepearson:
Fixes #3892

* Add a check to make sure there's a session before trying to unset a value in it.
* As far as I can tell, no other methods that modify the session are called without the possibility of a session existing.

#### Testing instructions

* Install WooCommerce Subscriptions. This can be obtained from the `all-plugins` repo.
* Enable UPE checkout if not already enabled.
* Create a subscription product that renews daily, make sure it has a price (like 10).
* Add subscription to your cart and check out.
* Modify the `woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/assets/js/admin/meta-boxes-subscription.js` file by commenting out lines 206-232. This allows you to set the subscription to renew in less than an hour.
* Go to Subscriptions, and then into your new subscription.
* Set the _Next payment_ date for the current day and about 2 minutes into the future.
* Log out of WordPress.
* Set a breakpoint in the [`remove_upe_payment_intent_from_session` method](https://github.com/Automattic/woocommerce-payments/pull/3893/files#diff-b541c0a287f40f82a1d0e43c52c7c7a2b67ae233f261655cbb437726fec83feeR1133) and wait for the renewal to process.
* When the renewal processes, it should find no session and should skip over calling `__unset`. 

-------------------

**Post merge**


ricardo (approved):
🚢 LGTM, tests well.

I was able to reproduce. I manually changed the `wp_actionscheduler_actions` table to set the date of the next renewal.

PR #3896, opened 2022-03-04
Drop `wordpress-element` package

### LINKED ISSUE #709: Drop own wordpress-element dependency
cc @marcinbot can we do this?

ismaeldcom:
Fixes #709

Drop `wordpress-element` in favor of `@wordpress/element` and update all affected code.

Our minimum WordPress version is now 5.7 and includes `@wordpress/element@2.19.1` that already exports `createInterpolateElement`. So we can replace all the `__experimentalCreateInterpolateElement ` calls with it. And stop bundling it.

#### Testing instructions
- Checkout the branch and run `npm run start`
- Any of the affected components that use `createInterpolateElement` should render without issues. For example: deposit status, connect account page, account status settings, etc.

-------------------

### DIFF
--- a/changelog/remove-709-wordpress-element
+++ b/changelog/remove-709-wordpress-element
@@ -1,0 +1,5 @@
+Significance: patch
+Type: dev
+Comment: Drop an old dependency
+
+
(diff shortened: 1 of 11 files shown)


luizreis (approved):
LGTM! Unit tests and affected screens are working fine, and the `wordpress-element` dependency has been removed :shipit:

PR #3898, opened 2022-03-04
Exclude Level3 data when capturing a terminal payment

### LINKED ISSUE #3861: In-Person Payments: exclude L2/L3 data from capturing process
### Description

Previously we added the data injection in this [PR](https://github.com/Automattic/woocommerce-payments/pull/2438).
The data are not required for In-Person Payments, so we can exclude it (also brings more privacy for buyers).

### Acceptance criteria

When capturing In-Person Payments, no L2/L3 data is shared.

### Dev notes

Perhaps we could revert the referenced PR as the most straightforward approach.

### Additional context

paJDYF-39U-p2

anu-rock:
Fixes #3861

We currently collect level 3 data for payments captured via terminal (in-person payments). Based on a recent discussion, this is no longer required. Thus, this PR aims to exclude level 3 data while capturing a terminal payment.

##### In-person payment

* Create a new order in WooCommerce with "Cash on delivery" as the payment method.

* Create a payment intent in Stripe Dashboard at `https://dashboard.stripe.com/test/connect/accounts/<account_id>/payments/new`. Make sure to have the "Capture funds later" option selected under "Manually enter card information."

* Make a POST request to `/payments/orders/<order_id>/capture_terminal_payment`. Refer the [docs](https://github.com/woocommerce/woocommerce/wiki/Getting-started-with-the-REST-API) to set up authorization correctly.

* On the intent details page `https://dashboard.stripe.com/test/connect/accounts/<account_id>/payments/<intent_id>`, scroll down to the "Events and logs" section. In the latest event data, verify that `level3` is null.

##### Capture later payment

* On the page WP Admin → Payments → Settings, check the option "Issue an authorization on checkout, and capture later." This will force all payments on the site to be captured later instead of right away.

* Create a new order in WooCommerce with "Credit card" as the payment method. After successfully checking out, notice that money from the card is not deducted yet.

* Go to the detail page (in WP Admin) for the just-created order. The status of this order should be "On hold," indicating that the payment hasn't been captured yet.

* You will also see this message in order notes: "A payment of $18.00 USD was authorized using WooCommerce Payments (_intent_id_)". Take a note of this intent id.

* From the "Order actions" dropdown in the right sidebar, select the "Capture charge" option.

* On the intent details page `https://dashboard.stripe.com/test/connect/accounts/<account_id>/payments/<intent_id>`, scroll down to the "Events and logs" section. In the latest event data, verify that `level3` is populated with some data.

-------------------

- [x] Covered with tests
- [x] Tested on mobile

**Post merge**


anu-rock:
@kalessil I've excluded level 3 data altogether. Reverting the changes in #2438 as you suggested would only impact the `customer_reference` field. Let me know if there's a gap in my understanding.

changelog/update-3861-ipp-exclude-l2-l3-data
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: tweak
```
kalessil, line 2: Can we use 'update' here?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -358,17 +358,15 @@ public function refund_charge( $charge_id, $amount = null ) {
...
 	 * @throws API_Exception - Exception thrown on intention capture failure.
 	 */
-	public function capture_intention( $intention_id, $amount, $level3 = [] ) {
+	public function capture_intention( $intention_id, $amount ) {
 		$request                      = [];
 		$request['amount_to_capture'] = $amount;
-		$request['level3']            = $level3;
 
-		$response_array = $this->request_with_level3_data(
+		$response_array = $this->request(
```
kalessil, line 369: This will change the behavior of manually capturing payments from the admin area.
Instead we should supply empty level3 data when capturing in-person payments.

kalessil (changes requested):
Noted a side-effect we are causing by following the [proposed approach.

includes/wc-payment-api/class-wc-payments-api-client.php
anu-rock: With empty level3 data, won't the `request_with_level3_data` method simply pass through the request to the `request` method just like my direct call here?

https://github.com/Automattic/woocommerce-payments/blob/fa960bcd98377f53b51adbf55b854e27e1c7bc8c/includes/wc-payment-api/class-wc-payments-api-client.php#L1717-L1721

changelog/update-3861-ipp-exclude-l2-l3-data
anu-rock: Fixed in 104d889 ✅

includes/wc-payment-api/class-wc-payments-api-client.php
anu-rock: > This will change the behavior of manually capturing payments from the admin area.

Would love to know more about it :) If I'm not wrong, the method `WC_Payments_API_Client::capture_intention` is used only in the `WC_Payment_Gateway_WCPay::capture_charge` method.

kalessil: Correct, and `WC_Payment_Gateway_WCPay::capture_charge` is also used in `add_action( 'woocommerce_order_action_capture_charge', [ $this, 'capture_charge' ] );`.

In WCPay settings we can enable authorizing payments and the from order action manually capture the payment.

kalessil: > With empty level3 data, won't the request_with_level3_data method simply pass through the request to the request method just like my direct call here?

Yes, that's the point. If we have a way from IPP capture payment endpoint pass in empty level 3 data, we are good.

anu-rock: Addressed the side effect in 3a7906a

kidinov (approved):
Tested on mobile - looks good!

p91TBi-7Iv-p2#comment-9101

PR #3899, opened 2022-03-04
Fee Display Causing Confusion for Merchants

### LINKED ISSUE #3666: Fee Display Causing Confusion for Merchants
### Description

While on janitorial, I've encountered two instances of merchants who are confused by how we display credit card fees in WCPay. I wanted to open an issue for discussion and consideration of whether we should adjust how we display this.

Links to the relevant issues:
p1642143941000300-slack-C7U3Y3VMY 
p1642564542115300-slack-C7U3Y3VMY 

In both instances, merchants who were only selling domestically saw the following:

Screenshot 2022-01-19 at 10 21 44

The merchants were both only selling domestically, and were confused at why the rate was so high (it should be 2.9% + $0.30 in both the US and Canada). They also thought it may somehow be related to the new checkout experience.

The rate displayed on the interface includes international charges, and while a breakdown can be displayed by hovering over the rate, both merchants did not realise this and thought this rate was too high/incorrect. I worry this could put some merchants off, as in this case the merchant asked us about the rate, but other merchants may see this rate and think it is a blocker to using WCPay.

I wonder whether we should either:
a. Only display the domestic rate if the user is *not* using Multi Currency
b. Display the rate in a way that makes it more clear what the *domestic* rate is for the user's country, with some additional design to show that additional rates would be incurred for international payments.

Would appreciate any thoughts on this. Thanks!
 
cc @LevinMedia @jarekmorawski

zmaglica:
Fixes #3666

There were cases when users got confused with high fees. This PR solves that by letting user know that fees start from a specific value (base rate) instead of displaying total possible transaction fee.

**Before**
image

**After**
image

#### Testing instructions

1. Go to WooCommerce Payments settings page (wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments)
2. Scroll to the area where payments methods are listed
3. You should see fee prices where prices have "From" prefix
4. For payments methods where base fee percentage doesn't exit, the lowest percentage fee should be displayed instead (iDeal might be a good example here, but it might not be in your case).
5. Mouse over the price should show tooltip or modal where fees are explained in detail (like it was before)

-------------------

### DIFF
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -222,17 +222,17 @@ export const formatAccountFeesDescription = (
 		...customFormats,
 	};
 
+	let displayFeePercentageRate = baseFee.percentage_rate;
+	if ( displayFeePercentageRate <= 0 ) {
+		displayFeePercentageRate =
+			additionalFee.percentage_rate < fxFee.percentage_rate
+				? additionalFee.percentage_rate
+				: fxFee.percentage_rate;
+	}
 	const feeDescription = sprintf(
 		formats.fee,
-		formatFee(
-			baseFee.percentage_rate +
-				additionalFee.percentage_rate +
-				fxFee.percentage_rate
-		),
-		formatCurrency(
-			baseFee.fixed_rate + additionalFee.fixed_rate + fxFee.fixed_rate,
-			baseFee.currency
-		)
+		formatFee( displayFeePercentageRate ),
+		formatCurrency( baseFee.fixed_rate )
 	);
 
 	const isFormattingWithDiscount =
@@ -281,7 +281,7 @@ export const formatAccountFeesDescription = (
 		} );
 	}
 
-	return feeDescription;
+	return sprintf( __( 'From %s', 'woocommerce-payments' ), feeDescription );
 };
 
 export const formatMethodFeesDescription = (
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -222,6 +222,7 @@ export const formatAccountFeesDescription = (
 		...customFormats,
 	};
 
+	// Some payment methods doesn't have base percentage rate. In this case, the lowest rate will be shown as a start value
 	let displayFeePercentageRate = baseFee.percentage_rate;
 	if ( displayFeePercentageRate <= 0 ) {
 		displayFeePercentageRate =
@@ -232,9 +233,8 @@ export const formatAccountFeesDescription = (
 	const feeDescription = sprintf(
 		formats.fee,
 		formatFee( displayFeePercentageRate ),
-		formatCurrency( baseFee.fixed_rate )
+		formatCurrency( baseFee.fixed_rate, baseFee.currency )
 	);
-
 	const isFormattingWithDiscount =
 		currentFee.percentage_rate !== baseFee.percentage_rate ||
 		currentFee.fixed_rate !== baseFee.fixed_rate ||
@@ -281,7 +281,7 @@ export const formatAccountFeesDescription = (
 		} );
 	}
 
-	return sprintf( __( 'From %s', 'woocommerce-payments' ), feeDescription );
+	return feeDescription;
 };
 
 export const formatMethodFeesDescription = (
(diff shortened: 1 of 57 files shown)


client/utils/account-fees.tsx
```diff
@@ -222,17 +222,17 @@ export const formatAccountFeesDescription = (
 		...customFormats,
 	};
 
+	let displayFeePercentageRate = baseFee.percentage_rate;
+	if ( displayFeePercentageRate <= 0 ) {
```
deepakpathania, line 226: Can we add a comment explaining this logic for clarity?

client/utils/account-fees.tsx
```diff
@@ -222,17 +222,17 @@ export const formatAccountFeesDescription = (
...
-			baseFee.percentage_rate +
-				additionalFee.percentage_rate +
-				fxFee.percentage_rate
-		),
-		formatCurrency(
-			baseFee.fixed_rate + additionalFee.fixed_rate + fxFee.fixed_rate,
-			baseFee.currency
-		)
+		formatFee( displayFeePercentageRate ),
+		formatCurrency( baseFee.fixed_rate )
```
deepakpathania, line 235: I think we still need to pass the currency as a second param, it defaults to `USD` but for multi-currency scenarios, it should be explicitly passed IMO.

client/utils/account-fees.tsx
```diff
@@ -281,7 +281,7 @@ export const formatAccountFeesDescription = (
 		} );
 	}
 
-	return feeDescription;
+	return sprintf( __( 'From %s', 'woocommerce-payments' ), feeDescription );
```
deepakpathania, line 284: Since we're wrapping `feeDescription` in `sprintf` above, can we try to do it in a single place and keep this line as plain return only?

deepakpathania (changes requested):
Thanks for working on this. Added some comments, please also add some screenshots in the PR description for easier visualization of the changes.

client/utils/account-fees.tsx
zmaglica: Fixed. Added 'From' prefix when function is called, so we don't have to hardcode it. This way, the custom formatting won't interfere with 'From' prefix.

client/utils/account-fees.tsx
zmaglica: Added.

zmaglica:
> Thanks for working on this. Added some comments, please also add some screenshots in the PR description for easier visualization of the changes.

Thank you for your comment. Screenshots have been added to the PR description.

client/utils/account-fees.tsx
```diff
@@ -222,19 +222,19 @@ export const formatAccountFeesDescription = (
...
-			baseFee.percentage_rate +
-				additionalFee.percentage_rate +
-				fxFee.percentage_rate
-		),
-		formatCurrency(
-			baseFee.fixed_rate + additionalFee.fixed_rate + fxFee.fixed_rate,
-			baseFee.currency
-		)
+		formatFee( displayFeePercentageRate ),
+		formatCurrency( baseFee.fixed_rate, baseFee.currency )
```
deepakpathania, line 236: Is sending only `fixed_rate` here an intended change? I am not sure about this change tbh and would appreciate another pair of eyes on this. cc @jessy-p who might have more context.

deepakpathania:
The original issue also shows a tooltip with the breakdown on hovering over the fee bubble, is that not added yet here?

client/utils/account-fees.tsx
jessy-p: To explain simply, the change is that, instead of showing fee as "baseFee.percentage_rate + additionalFee.percentage_rate + fxFee.percentage_rate" we will show "From baseFee.percentage_rate", so number doesn't look too high. By using the keyword "From" we are saying, it could be more too.

jessy-p: Similarly for fixed_rate, we use the most common and usually applied baseFee.fixed_rate instead of adding all the rates and showing that value

jessy-p (approved):

zmaglica:
> The original issue also shows a tooltip with the breakdown on hovering over the fee bubble, is that not added yet here?

Tooltip is already implemented. As @jessy-p mentioned, the only change is to show base fee (or the lowest fee if base fee is not set) and add "From" prefix.

PR #3433, opened 2021-11-26
IPP Styled Receipts: Print receipt

mgascam:
Fixes #3021 and #3224

#### Changes proposed in this Pull Request

This PR exposes a new REST API endpoint to generate an IPP receipt for printing. The new endpoint will collect all the data needed to generate the receipt, and will return an HTML version of it. The client can then use the returned HTML to print the receipt.

We did some tests using a thermal printer to validate the implementation (thanks again @malinajirka ). After some tweaks, this is how the receipt looks like:

I added the new endpoint to the payment readers controller because the receipts are strictly related to the readers. A client can do a `GET` request to `payments/readers/receipts/payment_id ` to fetch the receipt. 

Initially, we had the idea of caching the receipt, to get some performance savings. But during development we decided to change the approach. The generation of the receipt in HTML does not seem to be a costly operation and it will be printed just once most of the times. With these things in mind, the complexity of caching the receipt in the database seemed to me not necessary at the moment.

 

#### Testing instructions

To test this feature in your local environment you will need:

- Connect your local WC Pay client and server.
- Create an order with Cash on Delivery as payment method. Please use the proxied url of your local site.
- Start listening for webhooks in your local WC Pay server. You'll need this to refresh your order status after next step.
- Use the WooCommerce Mobile App to pay the order. You can follow this great guide to set it up locally: pdjVC1-9M-p2
- Once the payment is suceeded, you'll need to note the payment id reference i.e: you can find it in the order details
- Using Postman (note that you'll need to set up Authentication as explained in this guide: PbTz5e-12J-p2) make a GET request to `/wp-json/wc/v3/payments/readers/receipts/<payment_id>`.
- You should be able to check the HTML response within Postman. Hint: There is a preview button that let's you see the rendered HTML

Screen Shot 2021-12-01 at 20 07 19
-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Added testing instructions to the [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) (or does not apply)

PR #3901, opened 2022-03-04
Fix Printed receipt endpoint throws error when called from Jetpack API

### LINKED ISSUE #3900: Printed receipt endpoint throws error when called from Jetpack API
### Describe the bug

I noticed that there is an error when the endpoint that we created to generate a printed receipt is called from the Jetpack API.

### To Reproduce

1. Be sure that you have at least one IPP payment available in your test store. Follow instructions on [this PR](https://github.com/Automattic/woocommerce-payments/pull/3433) to learn how to do it. 
2. Make a `GET` request to `payment/readers/receipts/<payment_intent_id>` through Jetpack API i.e.  where `payment_intent_id` is the one you created in previous step. Please find [here](pdjVC1-o4-p2) instructions to use WP.COM API developer console to do it.

### Actual behavior

There is a 500 error thrown.

### Screenshots

Screen Shot 2022-03-04 at 10 59 58

### Expected behavior

There is no error when calling this endpoint and the HTML of the receipt is present in the response:

Screen Shot 2022-03-04 at 11 15 30

mgascam:
Fixes #3900

Change the content type of the response of the endpoint to generate a printed receipt.

When the endpoint is called from the Jetpack API, it throws a 500 error because the API expects `application/json` type. 

Screen Shot 2022-03-04 at 10 59 58

#### Expected outcome
Screen Shot 2022-03-04 at 11 15 30

#### Testing instructions

1. Be sure that you have at least one IPP payment available in your test store. Follow instructions on https://github.com/Automattic/woocommerce-payments/pull/3433 to learn how to do it.
2. Make a GET request to `payments/readers/receipts/<payment_intent_id> ` through Jetpack API i.e. where `payment_intent_id` is the one you created in previous step. Please find [here](pdjVC1-o4-p2) instructions to use WP.COM API developer console to do it.
3. Check that there are no errors in the response.
4. Check that the printed receipt HTML is present in the json response.

-------------------

**Post merge**


anu-rock (commented):
The code changes look good to me.

But I could not successfully test the endpoint in WP Developer Console. I kept getting this response (both for existing and new payments):

```json
{
    "code": "generate_print_receipt_error",
    "message": "Payment method details needs to be provided.",
    "data": {
        "status": 500
    }
}
```

In fact, I got this error even with localhost endpoint URL.

mgascam:
> The code changes look good to me.
> 
> But I could not successfully test the endpoint in WP Developer Console. I kept getting this response (both for existing and new payments):
> 
> ```json
> {
>     "code": "generate_print_receipt_error",
>     "message": "Payment method details needs to be provided.",
>     "data": {
>         "status": 500
>     }
> }
> ```
> 
> In fact, I got this error even with localhost endpoint URL.

Thanks for checking it @anu-rock.
I did some checks on my end, using `capture_terminal_payment` endpoint, and I am getting the same error. But it works as expected when I use the mobile app with the emulator.

If I compare two charges (one linked to a payment captured with `capture_terminal_payment` and the other with the mobile app) I see that the `payment_method_details` are different. In one case, the type is set to `card_present` and in the other it is set to `card`. The charge linked to the one I captured with the emulated card reader also has information like `dedicated_file_name` or `application_preferred_name`, that is missing in the other charge.

This error is thrown when we validate the charges (in this [line](https://github.com/Automattic/woocommerce-payments/blob/7814dd81cd3a65a9377362a321fb50f4329654cd/includes/in-person-payments/class-wc-payments-in-person-payments-receipts-service.php#L122)) so I'd say this is the expected behavior, or at least intentional. We could stop throwing this error if the `card_present` method is not specified in the charge, but then we would not be able to print in the receipt some compliance information like the `application_name`, I think.

Perhaps there is room for improvement in the error message. It does not seem to be very descriptive :)

Please let me know your thoughts when you have a moment.

mgascam:
As per our conversation offline, we decided to explore how to make it easier to test this functionality without the help of the mobile app. We will tackle this into a different ticket, since it is not strictly related to the issue at hand. Also we do not think this is a blocker, since this can be tested using the mobile app.

We still need to investigate what would be the best option to do this.
cc @anu-rock

kalessil (approved):
LGTM! Please ensure to sync with folks from Mobile teams that the fix works.

anu-rock (approved):
LGTM too!

malinajirka:
Awesome, thank you for working on this 🙇 !

PR #3912, opened 2022-03-07
REST API documentation 

dmvrtx:

1. Initial documentation for our API endpoints.
2. A new Github workflow to publish API docs to https://automattic.github.io/woocommerce-payments.

The documentation static site generation is based on [Slate](https://github.com/slatedocs/slate). The theming and flow are heavily inspired by [woocommerce-rest-api-docs](https://github.com/woocommerce/woocommerce-rest-api-docs).

image

Docs are added as markdown files in `docs/rest-api/source/includes/wp-api-v3`. This PR ships only with the `Orders` controller docs.

#### Testing instructions

* Check the documentation markdown files.
* Check the documentation site at https://automattic.github.io/woocommerce-payments.

-------------------

- [x] Covered with tests (does not apply)


docs/rest-api/source/includes/wp-api-v3/introduction.md
```diff
@@ -0,0 +1,82 @@
...
+WooCommerce (WC) 2.6+ is fully integrated with the WordPress [REST](http://en.wikipedia.org/wiki/Representational_State_Transfer) API. This allows WC data to be created, read, updated, and deleted using requests in JSON format and using WordPress REST API Authentication methods and standard HTTP verbs which are understood by most HTTP clients.
+
+The current WP REST API integration version is `v3` which takes a first-order position in endpoints.
+
+## Requirements
+
+To use the latest version of the REST API you must be using:
+
+-   WooCommerce 3.5+.
+-   WordPress 4.4+.
```
KarlisJ, line 12: I know we have a compatibility file somewhere. Could it be possible to somehow automate that these versions are pulled from there or updated? 

My general worry with the approach is that documentation will go out of date unless some strong practises are put in place.

docs/rest-api/source/includes/wp-api-v3/order.md
```diff
@@ -0,0 +1,141 @@
+# Order
+
+The order API allows you to perform operations such as capture a terminal payment, create an intent, create a customer, etc. for a particular order.
+
+## Create a terminal intent
+
+Create a new in-person payment intent for the given order ID without [confirming](https://stripe.com/docs/api/payment_intents/confirm) it.
```
KarlisJ, line 7: If we plan to make this public then we shouldn't link to Stripe

docs/rest-api/source/includes/wp-api-v3/order.md
```diff
@@ -0,0 +1,141 @@
+# Order
+
+The order API allows you to perform operations such as capture a terminal payment, create an intent, create a customer, etc. for a particular order.
+
+## Create a terminal intent
```
KarlisJ, line 5: Overall, I wasn't able to understand this section. 

- The request has `<order_id>` in it's path.
- There's a post param `payment_intent_id: string`
- example doesn't pass `payment_intent_id` 

I think that 
```
POST params
payment_intent_id: string
```

is mistakenly put in this section but it's missing from next one - "Capture a terminal payment"

docs/rest-api/source/includes/wp-api-v3/order.md
```diff
@@ -0,0 +1,141 @@
...
+		"status": 500
+	}
+}
+```
+
+## Create customer
+
+Returns customer id from order. Create or update customer if needed.
+
+If the order is created for a registered customer, we extract it's Stripe customer ID.
```
KarlisJ, line 103: Again, a reference to "Stripe" that shouldn't be in public docs.

KarlisJ (commented):
First of, this looks really good in browser and if we can get the automation working then it's huge improvement for our documentations

I left some comments, but mostly I have two concerns:
 1. We reference Stripe in documentation that's supposed to be public. Ideally, we could provide a way to add a directive that would be used only when used internally, so we could "extend" public documentation with Stripe-related information for internal use.
 2. The JSON response examples doesn't have all the fiels in them. Don't know if that's intentional or temporal. But overall, IMHO, if there is a documentation then it should be full and precise.

docs/rest-api/source/includes/wp-api-v3/order.md
anu-rock: Fixed in 5ecd764

docs/rest-api/source/includes/wp-api-v3/order.md
anu-rock: Fixed in 5ecd764

docs/rest-api/source/includes/wp-api-v3/order.md
anu-rock: Indeed. It was earlier fixed in 1d20d47 but the deployed docs were not regenerated later. In fact, docs will only be regenerated when this PR is merged.

anu-rock:
> The JSON response examples doesn't have all the fiels in them. Don't know if that's intentional or temporal. But overall, IMHO, if there is a documentation then it should be full and precise.

AFAIK, the success responses are mentioned in full. Which ones do you think are missing fields?

docs/rest-api/source/includes/wp-api-v3/introduction.md
anu-rock: You're right. At the moment, I'm not sure of a foolproof solution. Perhaps we can revisit this later?

kalessil (commented):
With input on this PR being a foundation for further documenting IPP-related endpoints, this PR is awesome! Great work everyone!

I did only a very brief review, hence several points from me:
- not mentioning Stripe and server-side of things
- cross-linking to existing docs on the WooCommerce side (introduction, authentication perhaps)
- it'd be great for each endpoint to have 'since' information and mini-changelog describing changes there per release version

The first one seems to be raised already and I'd appreciate it being addressed today or on Monday.

mgascam (approved):
LGTM! I am not familiar with how Slate works, so I cannot provide any meaningful feedback there. But the documentation looks great ✨ Amazing job

anu-rock:
@kalessil Thank you for your detailed review :)

> not mentioning Stripe and server-side of things

This was addressed on Friday. There are no more such references.

> cross-linking to existing docs on the WooCommerce side (introduction, authentication perhaps)

Our intro and auth docs are abridged versions of the Woo docs. The auth docs link to Woo docs to help with more details.

> it'd be great for each endpoint to have 'since' information and mini-changelog describing changes there per release version

Do you think our JavaDoc-like code comments would be better suited for "since" and "changelog" information? That information would then be picked up when generating Swagger docs. I feel that manually maintaining this information in markdown files could get painful soon.

dmvrtx:
> I feel that manually maintaining this information in markdown files could get painful soon.

I agree. Maybe it is a good challenge for the next HACK week?

kalessil (approved):
LGTM!

PR #3914, opened 2022-03-07
Do not enqueue admin assets if user cannot manage_woocommerce

jessepearson:

* Add `current_user_can( 'manage_woocommerce' )` checks when `is_admin()` is called so as not to include classes or assets that aren't available to non-admins when in `/wp-admin/`.

#### Testing instructions

* Tests should pass.
* WCPay pages in the admin should still remain functional when you're an admin: Settings, Transactions, Multi-Currency.
* WordPress pages should not break when in `/wp-admin/` as a non-admin, like an Author.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : Testing would be covered with standard critical flow tests.


ismaeldcom:
Hey @jessepearson, can you provide more details about the following point?
>WordPress pages should not break when in /wp-admin/ as a non-admin, like an Author.

I tried following the testing instructions, but couldn't see any issue in WP Admin like an Author without this PR.

jessepearson:
> Hey @jessepearson, can you provide more details about the following point?
> I tried following the testing instructions, but couldn't see any issue in WP Admin like an Author without this PR.

There is not an existing problem with `wp-admin` for non-admins other than the JS may be included in their pages when it doesn't need to be. Meaning, nothing is currently broken. Other admin files are also no longer included if the user cannot `manage_woocommerce`. 

I thought I had included this above, but this came from this little Slack conversation: p1645795914033389-slack-C01B8KNUYSW . Basically, I noticed something just wasn't being done well and wanted to fix it.

ismaeldcom (approved):
LGTM! Let's 🚢  it.

>There is not an existing problem with `wp-admin` for non-admins other than the JS may be included in their pages when it doesn't need to be. Meaning, nothing is currently broken. Other admin files are also no longer included if the user cannot `manage_woocommerce`. 

My bad! I was thinking on an existing issue. Now it got the point. I've checked how with the PR it doesn't load `WCPAY_MULTI_CURRENCY_ANALYTICS_JS` in WP Dashboard for an Author. Thanks for fine-tuning the implementation 🙂

PR #3916, opened 2022-03-08
HACK week: migration to TypeScript (client/card-readers/list)

mgascam:
Migration to TS of `client/card-readers/list` folder component.

#### Testing instructions

- Enable the Card Readers menu item (you can force `calculate_card_present_eligibility` to return true [in the server](https://github.com/Automattic/woocommerce-payments-server/blob/e481ea6d1c570d5a6d3872c4446bb20d1f7b57f5/server/wp-content/rest-api-plugins/endpoints/wcpay/services/class-account-service.php#L1052))
- Update the `_wcpay_feature_card_readers` option to `1`.
- Go to Payments > Card Readers
- You should see the list of available card readers


client/card-readers/declarations.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module 'wcpay/settings/settings-section' {
+	interface SettingsSectionProps {
+		/* eslint-disable-next-line @typescript-eslint/naming-convention */
+		Description: () => JSX.Element;
```
kalessil, line 4: Can we follow naming conventions and use `Description`?

kalessil (changes requested):
Looks nice! I left a few comments (did a quick review only).

client/card-readers/list/index.tsx
```diff
@@ -48,7 +48,7 @@ const ReadersList = () => {
 							</div>
 						</CardBody>
 						<CardDivider />
-						<CardBody size={ null }>
+						<CardBody>
```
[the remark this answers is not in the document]
mgascam: Indeed, the type for size does not allow `null` :), it can be `undefined` though

client/card-readers/declarations.d.ts
mgascam: Good point, I think I am going to migrate `SettingsSectionProps` also.

client/card-readers/list/index.tsx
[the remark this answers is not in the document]
kalessil: +1

client/settings/test/settings-section.tsx
```diff
@@ -33,7 +33,7 @@ describe( 'SettingsSection', () => {
 
 	test( 'renders children', () => {
 		render(
-			<SettingsSection title="Foo" description="Bar">
```
mgascam, line 36: I removed this because there is no property `description` in the `SettingsSection` component. The correct property name is `Description`

client/data/card-readers/hooks.ts
```diff
@@ -3,9 +3,13 @@
  * External dependencies
  */
 import { useSelect } from '@wordpress/data';
+import { CardReaders, CardReaderStats } from 'wcpay/types/card-readers';
```
kalessil, line 6: This should be in the internal dependencies block.

client/settings/test/settings-section.tsx
mgascam, line 36: Removing this since `description` cannot be of `string` type

client/data/card-readers/hooks.ts
```diff
@@ -26,19 +30,19 @@ export const useCardReaderStats = ( chargeId, transactionId ) =>
 		[ chargeId, transactionId ]
 	);
 
-export const useReaders = ( limit ) =>
+export const useReaders = ( limit: number ): CardReaders =>
```
kalessil, line 33: Let's inline the type here.

client/data/card-readers/hooks.ts
```diff
@@ -3,9 +3,13 @@
...
  */
 import { useSelect } from '@wordpress/data';
+import { CardReaders, CardReaderStats } from 'wcpay/types/card-readers';
 import { STORE_NAME } from '../constants';
 
-export const useCardReaderStats = ( chargeId, transactionId ) =>
+export const useCardReaderStats = (
+	chargeId: string,
+	transactionId: string
+): CardReaderStats =>
```
kalessil, line 12: Let's inline the type here as well.

client/settings/test/settings-section.tsx
```diff
@@ -33,7 +33,7 @@ describe( 'SettingsSection', () => {
...
+			<SettingsSection title="Foo">
```
kalessil, line 36: Why do we drop `description` here?

client/card-readers/list/index.tsx
```diff
@@ -9,8 +9,8 @@ import { Card, CardBody, CardDivider } from '@wordpress/components';
 /**
  * Internal dependencies
  */
-import SettingsSection from 'wcpay/settings//settings-section';
-import SettingsLayout from 'wcpay/settings//settings-layout';
+import SettingsSection from 'wcpay/settings/settings-section';
+import SettingsLayout from 'wcpay/settings/settings-layout';
 import LoadableSettingsSection from 'wcpay/settings//loadable-settings-section';
```
kalessil, line 14: Can wee replace `//` to `/` here as well?

kalessil (changes requested):
Great work! I added some comments to further improve the PR.

client/card-readers/declarations.d.ts
mgascam: @kalessil I went for a different approach here. We have a TS linter rule that enforces camelCase or snake_case for property names. I refactored the `SettingsSection` component to use `description` as the name of the property. It is worth mentioning that in the `SettingsSection` component I had to alias it to `Description`. This is because `<description />` throws a warning from React. Instead, I can alias it and use it like `<Description />`.

Please let me know your thoughts.

client/settings/test/settings-section.tsx
kalessil: Got it, thanks!

client/settings/test/settings-section.tsx
mgascam: I think this was a mistake. Originally the `SettingsSection` component did not have a `description` property. In my understanding, this was not making any effect.

After this PR, I refactored the original name of the property `Description` to `description`, to comply with the TS linter naming-conventions rule. And now the type of description is `React.FunctionComponent`, so setting it to `string` is not allowed. It is worth mentioning that there were no cases were the `Description` was passed as `string`.

Hopes this makes sense but let me know otherwise

client/data/card-readers/hooks.ts
mgascam: Done

client/data/card-readers/hooks.ts
mgascam: Done

client/data/card-readers/hooks.ts
mgascam: Done

client/card-readers/list/index.tsx
mgascam: Done

kalessil (approved):
LGTM!

kalessil:
@mordeth: do you have a dev. environment with terminals? For me, the changes look good, but I'd like someone to give the PR a manual test.

mordeth (changes requested):
Changes looks good. There is an extra gap on the readers list screen. Looks like the extra space comes from the additionally added `is-size-medium` class to `components-card__body` wrapper.

Before:
Screenshot 2022-03-11 at 14 01 06

After:
Screenshot 2022-03-11 at 14 01 20

mgascam:
> Changes looks good. There is an extra gap on the readers list screen. Looks like the extra space comes from the additionally added `is-size-medium` class to `components-card__body` wrapper.
> 
> Before: Screenshot 2022-03-11 at 14 01 06
> 
> After: Screenshot 2022-03-11 at 14 01 20

Great catch @mordeth! This must come from removing `size=null` from `CardBody`. I'll look into it

mordeth (approved):
LGTM!

PR #3925, opened 2022-03-09
Place WCPay subscriptions on-hold prior to renewal

### LINKED ISSUE #3923: Subscription action not firing on successful renewals
### Describe the bug
The `woocommerce_subscription_renewal_payment_complete` action is not firing on successful subscription renewals. This causes issues with extensions that rely on this action.

### To Reproduce

1. Create a small plugin that creates a log entry when the `woocommerce_subscription_renewal_payment_complete` action runs. (see below for example)
2. As a shopper, purchase a subscription product.
3. Use the WCPay Dev Tools test clock feature to force a renewal.
4. Observe, the log entry is not created.

```php
add_action(
	'woocommerce_subscription_renewal_payment_complete',
	function( $subscription ) {
		error_log( sprintf( 'Subscription #%d renewal payment complete!', $subscription->get_id() ) );
	}
);
```

### Actual behavior
The `woocommerce_subscription_renewal_payment_complete` action is not fired on successful subscription renewals. 

### Expected behavior
The `woocommerce_subscription_renewal_payment_complete` action is fired on successful subscription renewals. 

### Additional context
Discussed here p1646718001301219-slack-C02BW3Z8SHK

aprea:
Fixes #3923

Temporarily place the subscription on-hold to imitate the normal subscription renewal flow. This ensures the downstream effects take place, e.g. a payment status order note is added and the `woocommerce_subscription_payment_complete` action is fired.

#### Testing instructions

1. Create a small plugin that creates a log entry when the `woocommerce_subscription_renewal_payment_complete` action runs. (see below for example)
2. As a shopper, purchase a subscription product.
3. Use the WCPay Dev Tools test clock feature to force a renewal.
4. Observe, the log entry is created (this log entry **is not** created on `develop`).

```php
add_action(
	'woocommerce_subscription_renewal_payment_complete',
	function( $subscription ) {
		error_log( sprintf( 'Subscription #%d renewal payment complete!', $subscription->get_id() ) );
	}
);
```

-------------------

**Post merge**


includes/subscriptions/class-wc-payments-subscriptions-event-handler.php
```diff
@@ -119,7 +119,18 @@ public function handle_invoice_paid( array $body ) {
...
+			/*
+			 * Temporarily place the subscription on-hold to imitate the normal subscription renewal flow.
+			 * This ensures the downstream effects take place, e.g. a payment status order note is added and the
+			 * 'woocommerce_subscription_payment_complete' action is fired.
+			 */
+			remove_action( 'woocommerce_subscription_status_on-hold', [ $this->subscription_service, 'suspend_subscription' ] );
+			$subscription->update_status( 'on-hold' );
+			add_action( 'woocommerce_subscription_status_on-hold', [ $this->subscription_service, 'suspend_subscription' ] );
+
+			remove_action( 'woocommerce_subscription_status_on-hold_to_active', [ $this->subscription_service, 'reactivate_subscription' ] );
```
aprea, line 131: I've added this to prevent the WCPay Server API call which attempts to reactive the subscription.

This is unnecessary as the WCPay subscription is already active and does not need to be reactivated.

haszari: Might be worth commenting that in code?

aprea: Code comment added in 57c3d1c9eb6f0ecce74d0c0969b8c3eb6a7acf61

tests/unit/helpers/class-wc-helper-subscription.php
```diff
@@ -231,4 +231,8 @@ public function is_manual() {
 	public function set_requires_manual_renewal( $bool ) {
 		$this->manual = $bool;
 	}
+
+	public function update_status( $status ) {
```
shendy-a8c, line 235: Why is this needed?

shendy-a8c (commented):
After fixing my local setup so that local wcpay server forwards webhook event correctly to my wcpay client / my local merchant store, I can confirm that it works as per testing instructions.

Out of curiosity, I also tested with WC Subscriptions active and I see `woocommerce_subscription_payment_complete` is fired just fine even without the change from this PR. This is expected. The only difference when testing this is that subscriptions from WCS can be renewed from the admin edit subscription page > Subscriptions actions > Process renewal directly. No need test clock.

Another minor concern that I can think of from this bug fix is that if there is code out there that notices `woocommerce_subscription_renewal_payment_complete` isn't firing for WCPay Subscriptions and then decide to hook on to `woocommerce_renewal_order_payment_complete` without removing the callback from `woocommerce_subscription_renewal_payment_complete`, after this fix, that code might be executed more times than expected, but this is very unlikely.

tests/unit/helpers/class-wc-helper-subscription.php
shendy-a8c: That makes sense. I missed that.

shendy-a8c (approved):

PR #3926, opened 2022-03-09
Fix guest session

superdav42:
###
Fixes platform checkout when customer is not logged into merchant site. 

Something changed in WC or WCPay so the session was initialized before the user was loaded. Therefore we need to register the `woocommerce_cookie` filter earlier in the init process. 

Without this change platform checkout site would say the cart was empty.

### DIFF
--- a/changelog/fix-guest-session
+++ b/changelog/fix-guest-session
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix checkout as guest on platform checkout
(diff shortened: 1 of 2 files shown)


Mayisha (commented):
Changes look good to me. But I can not regenerate the empty cart scenario in my `localdev` (in branch `develop`) :thinking: 

I tried these steps in `develop`:
- logged out of merchant site
- restarted WCPay 
- From WCPay checkout page went to Woopay via OTP verification
- If I am logged in to Woopay then the cart data loads perfectly ( no empty cart issue )

superdav42:
> Changes look good to me. But I can not regenerate the empty cart scenario in my `localdev` (in branch `develop`) thinking
> 
> I tried these steps in `develop`:
> 
> * logged out of merchant site
> * restarted WCPay
> * From WCPay checkout page went to Woopay via OTP verification
> * If I am logged in to Woopay then the cart data loads perfectly ( no empty cart issue )

It may be dependent on the version of WC Core installed on the merchant site. I tracked it down to loading current currency from the session so may also be dependent on multi-currency settings. Ether way it's a matter of the filter `woocommerce_cookie` maybe triggered before `determine_current_user` so we need to register the filter always.

Mayisha (approved):
:100:

PR #3940, opened 2022-03-10
Fix the conversion amount in the transaction list

### LINKED ISSUE #3939: Conversion indicator shows wrong amount on transaction list
### Describe the bug
When a transaction is displaying a converted value, it shows the same amount in both currencies, they should differ where the currencies are different (if the rate is different)

### To Reproduce
Assuming you have an USD account:
- Enable a currency on Payments > Settings > Multi-Currency which has a different rate than 1.
- On your store, change the currency to the new currency and sell something.
- Go to Payments > Transactions, and see the transaction
- Hover the conversion indicator icon, and on the tooltip you see this:

image

As you can see, both have the same amount, which they shouldn't.

### Expected behavior
The conversion indicator should have the original amount and currency, the converted amount should be different if the rate is not 1.

The fix may be related with this line having the amount: https://github.com/Automattic/woocommerce-payments/blob/develop/client/transactions/list/index.tsx#L286

jessepearson:
Fixes #3939

* Update the `fromAmount` in the transactions list to be the amount the conversion was from, meaning the amount the customer paid in their currency.

#### Testing instructions

* Add another currency to your store under WooCommerce > Settings > Multi-Currency.
* Add a product to your cart and go to checkout.
* Change the currency using the currency selector widget.
* Check out.
* View your transactions list and confirm the converted from amount shows correctly.

Screen Shot 2022-03-10 at 9 29 05 AM

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : QA Testing Not Applicable

### DIFF
--- a/changelog/fix-3939-incorrect-conversion-in-transaction-list
+++ b/changelog/fix-3939-incorrect-conversion-in-transaction-list
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix the conversion from amount in the transactions list.
--- a/client/transactions/list/index.tsx
+++ b/client/transactions/list/index.tsx
@@ -276,6 +276,7 @@ export const TransactionsList = (
 		const dataType = txn.metadata ? txn.metadata.charge_type : txn.type;
 		const formatAmount = () => {
 			const amount = txn.metadata ? 0 : txn.amount;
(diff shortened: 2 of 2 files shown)


tpaksu (approved):
Tested and works great! Thanks for fixing this. LGTM 🚢  it!

PR #3947, opened 2022-03-10
Use docker named volume instead of mount for MySQL data

### LINKED ISSUE #3946: `Permission denied` issues with mysql data and mounted (docker) volume
### Describe the bug
I recently tried re-creating my WCPay docker environment from scratch, and I hit an issue. The MySQL / `db` container had permissions issues with the data volume.

In our current docker environment, the [data volume is a mounted folder (from the host OS)](https://github.com/Automattic/woocommerce-payments/blob/develop/docker-compose.yml#L36). I think this causes the permission issue I'm seeing (maybe intermittent?). 

There's a workaround – to mount the volume with a user id mapping, so the container user has same access as host OS user. 

I'd recommend switching to a [named volume](https://docs.docker.com/storage/volumes/), which avoids the permissions issue altogether. This may also have other benefits since it's completely isolated inside the container. Note named volumes still persist across container recreate, so we still have a persistent database (and can reset it at will by deleting the volume - [`docker compose down --volumes`](https://docs.docker.com/compose/reference/down/)).

I notice that we [use a mounted volume for WordPress as well](https://github.com/Automattic/woocommerce-payments/blob/develop/docker-compose.yml#L22). I personally don't need this, perhaps if we move to a named volume it will simplify things and maybe even help with performance. 

My understanding is that mounted volumes are best for sharing source code and config files into a container; not sure about using them to share large amounts of data "out".

Note if we do switch to named volumes, developers can easily o[verride this if they have different needs, using `docker-compose.override.yml`](https://github.com/Automattic/woocommerce-payments/pull/2801). For example, if a developer needs to hack on WordPress core at the same time, they could clone locally and map into the container. 

### To Reproduce
- Clone repo. 
- Put stuff in `./docker/data` or remove write permissions (to force the issue!).
- `npm install`
- `npm run up:recreate`

### Actual behavior
Lots of errors in database container like the following, and then it bails. No database container, unhappy WordPress.

```
woocommerce_payments_mysql | chown: changing ownership of '/var/lib/mysql/': Permission denied
woocommerce_payments_mysql | chown: changing ownership of '/var/lib/mysql/ibdata1': Permission denied
woocommerce_payments_mysql | chown: changing ownership of '/var/lib/mysql/wcpay_tests': Permission denied
woocommerce_payments_mysql | chown: changing ownership of '/var/lib/mysql/wcpay_tests/wptests_wc_product_meta_lookup.frm': Permission denied
```

### Expected behavior
- Quick and robust launch or recreate of dev env under a wide range of circumstances. (I believe using named volumes will make the env more reliable.)

- macOS Monterey, Intel

haszari:
Fixes #3946

This PR uses [named docker volumes](https://docs.docker.com/storage/volumes/) for the MySQL data ~~and WordPress installation~~. These were previously mounted to the host OS, which can cause issues (and may not be needed IMO).

See above issue for further discussion / rationale.

Note that if this merges, WCPay devs may need / want to recreate their existing environments. The existing environments are based on the old config (with mounts) so there may be issues (would be good to test this to confirm how). Recreating is the cleanest way to adopt, but that might be inconvenient.

**Update** I've reverted the `wordpress` folder back to using a mount. This is less disruptive, and wasn't causing issues for me. This allows access to logs and convenient debugging of PHP code across WP core and plugins. The PR is now focused on the MySQL data specifically. Thanks @htdat  for feedback about these dev flows 🙌 

Note that I have committed some changes to ensure various mounted folders exist in a clean clone. In my testing the initial `npm run up` (docker-compose up) fails for each folder that doesn't exist. Creating the empty folders ensures that the initial `npm run up`[`:recreate`] works first time.

##### Bonus: `sleep 5`

This PR also changes a `sleep` call in `docker-setup.sh`. When I run this (on macOS) I get the following error/warning: 

```
usage: sleep seconds
```

During the initial setup of the dev env (e.g. `npm run up:recreate`) this means it polls without any delay, and makes the terminal output much noisier. 

I've changed this to `sleep 5` which should do the same thing on most platforms, I believe the default is seconds (I tried it inside container and it worked well).

##### Bonus: documented `dockerdirectory` workaround
We use a named volume to prevent all the `./docker` subfolders from clogging up the plugin folder. I didn't understand this until I tested / explored so I added some comments to clarify. 

##### out of the box devx
- Clean clone of wcpay code. 
- Install dependencies `npm install` 
- Start up dev environment `npm run up:recreate`
- Should be able to hack on WCPay with ease.

##### stopping and restarting dev env
- Existing clone or same one as above.
- Your WordPress content and images etc are set up how you want for some specific testing scenario.
- `npm run down` or reboot your actual host OS, etc, go AFK
- return from AFK, install your updates etc then `npm run up` to bring it back up
- Your WordPress content and images etc are **still** set up how you want for some specific testing scenario.

##### miscellaneous dev flows
Please test your own hot dev flows and use cases to ensure this doesn't break em.

-------------------

- [x] Covered with tests (or have a good reason not to test in description ☝️) n/a
- [x] Tested on mobile (or does not apply) n/a

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _'QA Testing Not Applicable'_


brucealdridge (commented):
Have tested the build / recreate steps and working well.

I believe that this change will reset developers environments which makes me hesitant. 
I didn't previously have docker setup so this did not affect me.

I wonder if there is a straightforward way to migrate the database to prevent this happening when this change is merged.

naman03malhotra (commented):
Thanks, Rua for working on this and also for fixing the "usage: sleep seconds" warning along with it.

The changes look good to me, however as soon as this gets merged it will break the local dev environment for everyone and can slow down the progress of ongoing projects a bit.

IMO, we can merge a change like this when there are fewer active projects and in AFK season unless urgent, well that's my opinion, would like to hear what others have to say.

shendy-a8c:
> I don't see a particular reason why we should maintain long-lived state in our dev databases.

I personally use a long-lived state as local development and I think a lot of others too, CMIIW.
For short-lived state, I personally use JN sites. Usually to work with onboarding or testing on a clean store.

Can your change be applied to `docker-compose.override.yml` instead?

haszari:
> The changes look good to me, however as soon as this gets merged it will break the local dev environment for everyone and can slow down the progress of ongoing projects a bit.

> IMO, we can merge a change like this when there are fewer active projects and in AFK season unless urgent, well that's my opinion, would like to hear what others have to say.

That's good feedback @naman03malhotra , I don't want to disrupt projects or release train more than necessary.

Here's what I'm thinking:

1. I work with @shendy-a8c to confirm easy steps to backup/restore a local dev site. Also will explore any other gotchas or impact on devs.
2. Decide on a date to merge that will reduce impact on release train – e.g. end of release week. 
3. P2 the plan, and the steps for saving/restoring state and gather feedback. This is another opportunity for feedback on the PR itself, or other approaches to restoring state.

Then if that all goes well we can proceed, if not we can step back and rethink. 

> I personally use a long-lived state as local development and I think a lot of others too, CMIIW. That is so that I don't have to keep repopulating my local store with products, etc.
For short-lived state, I personally use JN sites. Usually to work with onboarding or testing on a clean store.

I can understand that – thanks for clarifying why that state is important to you @shendy-a8c . There are a few ways to preserve and restore this state if you choose to – e.g. backup / restore or manually copy data folders and WordPress media folder. 

At the same time I recommend not relying on your local store state too much – if there's useful testing data or setup that you need regularly, find a way to reproduce it in a clean store (e.g. restore backup or run a CLI script).

> Can your change be applied to docker-compose.override.yml instead?

I'm advocating that we move to named volumes because (IMO) it's a better fit for our needs and would make the environment more robust. If that's true, then it should be the default.

The override file is not persisted anywhere (by design), so it's not a good fit for implementing this change.

aprea:
> I work with @shendy-a8c to confirm easy steps to backup/restore a local dev site. Also will explore any other gotchas or impact on devs.

FWIW this should be pretty straightforward with WP CLI.

e.g.

Export DB:

```
docker exec -ti woocommerce_payments_wordpress wp db export db-backup.sql --add-drop-table --allow-root
```

Import DB:

```
docker exec -ti woocommerce_payments_wordpress wp db import db-backup.sql --allow-root
```

shendy-a8c:
Upon closer look, I think there are more changes needed. For example:

1. e2e stuffs. It might be good to create a separate PR based off this PR's branch.

Search for `docker` in [./tests/e2e/env/docker-compose.yml](https://github.com/Automattic/woocommerce-payments/blob/8f8c3f1c78208453cb7f13ca3496249aa8f4ee84/tests/e2e/env/docker-compose.yml).

I also see reference to the wordpress docker folder [here](https://github.com/Automattic/woocommerce-payments/blob/8f8c3f1c78208453cb7f13ca3496249aa8f4ee84/.github/workflows/e2e-test.yml#L163).

`./tests/e2e/env/setup.sh` setting up folder permission [here](https://github.com/Automattic/woocommerce-payments/blob/8f8c3f1c78208453cb7f13ca3496249aa8f4ee84/tests/e2e/env/setup.sh#L123-L124) won't be necessary. Maybe this step is needed because of the host volume approach.

2. In `.gitignore` and `.eslintignore`. Remove any reference to the docker folders. Just search for `docker`.

This is minor. Can maybe just left them there.

After removing from `.gitignore`, those docker folders will show up as untracked files. One just need to move those folders out or remove them.

3. Updating `CONTRIBUTING.md`.

4. `docker-setup.sh` needs to install wcpay dev tool in the named volume.

[`npm run up:recreate` will run `docker-setup.sh`](https://github.com/Automattic/woocommerce-payments/blob/8f8c3f1c78208453cb7f13ca3496249aa8f4ee84/package.json#L43) and it [clones wcpay dev tool to the docker folder](https://github.com/Automattic/woocommerce-payments/blob/8f8c3f1c78208453cb7f13ca3496249aa8f4ee84/bin/docker-setup.sh#L131).

haszari:
Thanks @shendy-a8c for picking up those issues! I'll take a look and make the necessary changes.

jrodger (approved):
I don't have anything to add beyond what has already been discussed on P2.

I'm onboard with the general idea and plan for implementing it though!

haszari:
@htdat  suggested that it's useful to have these logs available (and mapped) by default:

- wp-content/debug.log (file)
- wp-content/uploads/wc-logs (folder)

I'm planning on updating the PR to selectively map these folders.

docker/wordpress/empty-folder.md
```diff
@@ -0,0 +1,3 @@
+This folder is required by docker setup. 
+
+This file is here to ensure the folder is present in a clean clone.
```
haszari, line 3: I've added these placeholder files to ensure that the folders are present in a clean clone of the repo.

If the folders aren't present, `npm run up` / `docker-compose up` can fail with errors like this:

```
ERROR: for woocommerce_payments_wordpress Cannot start service wordpress: error while creating mount source path '/Users/rua/_automattic/transact-dev/wcpay-docker-testing/woocommerce-payments/docker/wordpress': chown /Users/rua/_automattic/transact-dev/wcpay-docker-testing/woocommerce-payments/docker/wordpress: permission denied
```

haszari:
Manually running the relevant GitHub Action workflow to test if e2e is impacted by these changes:

https://github.com/Automattic/woocommerce-payments/actions/runs/2068144731

docker-compose.yml
```diff
@@ -24,7 +26,6 @@ services:
       - .:/var/www/html/wp-content/plugins/woocommerce-payments
       - ./docker/wc-payments-php.ini:/usr/local/etc/php/conf.d/wc-payments-php.ini
       - dockerdirectory:/var/www/html/wp-content/plugins/woocommerce-payments/docker
-      - ./docker/bin:/var/scripts
```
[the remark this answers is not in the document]
shendy-a8c: I believe `./docker/bin` is used for JT setup. References:
- https://github.com/Automattic/woocommerce-payments/blob/develop/package.json#L57
- https://github.com/Automattic/woocommerce-payments/blob/develop/bin/jurassic-tube-setup.sh

docker/wordpress/empty-folder.md
shendy-a8c: nit: I usually add an empty folder to Git [this way](https://stackoverflow.com/questions/115983/how-can-i-add-a-blank-directory-to-a-git-repository), that is adding `docker/wordpress/.gitignore` with content
```
# Ignore everything in this directory
*
# Except this file
!.gitignore
```
So it's a hidden file and more standard. Description as you have it in empty-folder.md can be set as comment in the `.gitignore`.

Just a suggestion. Your current approach is fine too =)

shendy-a8c:
As I'm trying to make setting up e2e tests locally works, I see that we need to do the same changes here to WCPay server. Hence, 1968-gh-Automattic/woocommerce-payments-server.

bin/docker-setup.sh
```diff
@@ -31,7 +31,7 @@ set +e
 cli wp db check --path=/var/www/html --quiet > /dev/null
 while [[ $? -ne 0 ]]; do
 	echo "Waiting until the service is ready..."
-	sleep 5s
+	sleep 5
```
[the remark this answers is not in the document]
shendy-a8c: Addressed 70276a94de298b319f283de790dbe9a4e454540e.

docker/wordpress/empty-folder.md
shendy-a8c: I use `.gitignore` for WCPay server changes. It doesn't have to be but are you OK to make it consistent here too?

haszari: > I use .gitignore for WCPay server changes. It doesn't have to be but are you OK to make it consistent here too?

Yes please! `.gitignore` is a much better approach than an "empty folder" file. Feel free to make this change @shendy-a8c – thanks for the tip.

haszari: .. and now I have switched to local `.gitignore`! 🕚  25ecd52376ff9225818bb3ce46447cad386d8985

haszari:
I've dropped the ball on this PR – I keep deferring it. Requested a new review from Harmony team so we can align on any blockers or risks and get this merged (finally).

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `3947` or branch name `fix/3946-use-docker-named-volumes` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/3946-use-docker-named-volumes) :rocket:

---

Build info: 

- Latest commit: 1222a9ee7a74a6632e357e76b6ebb81a54a31726
- Build time: 2023-09-19 23:52:55 UTC

docker-compose.yml
```diff
@@ -3,6 +3,8 @@ version: '3'
 volumes:
   ## Kludge for not having the ./docker directory bound recursively
   dockerdirectory:
```
oaratovskyi, line 6: we may not need this as we have `mysql_data` from this PR

oaratovskyi (approved):
@haszari everything works good! A few observations:
- `docker/data` is not created as expected
- `woocommerce-payments_mysql_data` is successfully created
- `woocommerce-payments_dockerdirectory` is still created but seems like it's empty (see the image below). The size is 8KB and the contents are empty

image
 
To test this PR I've did next:
- dumped my data with `mysqldump` that is coming from MySQL Workbench
- removed the containers with `npm run down` and removed the volume `woocommerce-payments_dockerdirectory` + removed manually the `docker/data` folder
- run `npm run up:recreate`
- and restored with `mysql` utility coming from Workbench too.

I've checked basic functions, successfully onboarded an account, bought an item from the store.

Added my suggestion to get rid of the unused volume created in the comment

:shipit:

haszari:
Thanks for the thoughtful and thorough review @oaratovskyi ! I'll take a look at some tweaks based on your feedback, and then look at steps to merge (with appropriate advance notice to all devs).

docker-compose.yml
haszari: I have never understood why this is here … thanks for asking 😁 

I did some investigation to find out the current behaviour and what would happen if i removed.

### current behaviour – kludged `/plugins/woocommerce-payments/docker`
- The kludge prevents everything in `./docker` from being mapped into the container. 
- There's a lot of subfolders, logs etc, and these are mapped elsewhere. 
- So I think this kludge is to keep the plugin folder neat & tidy, and not have stuff redundantly mapped.

Screenshot 2023-09-20 at 11 34 59 AM

### experiment: remove `dockerdirectory` volume, no kludge
- All the subfolders of `./docker` are mapped into `/plugins/woocommerce-payments/docker`.
- It's a bit messy and redundant.

Screenshot 2023-09-20 at 11 31 58 AM

So I think this kludge is worth keeping. I'll update the comment(s) to make the purpose clearer – TIL!

haszari: The workaround remains, I've added comments and reordered mappings to make this clearer. 1222a9ee7a74a6632e357e76b6ebb81a54a31726

haszari:
Next steps for this PR:

- [ ] Test auto tests and any other docker/container dev flows to ensure not broken.
- [ ] Post a P2 announcement to Woo Payments dev team with planned date of merge and how to prepare
- [ ] Wait, address any feedback, or (worst case) abandon if major blockers
- [ ] Merge on planned date 

Side note – I'll plan to merge this at least 1 week before code freeze to avoid any possible impact on release process. This might delay merge a few weeks.

oaratovskyi:
@haszari thanks for detailed response! Gotcha about kludge - nice find 👍

haszari:
Closing, this is stale and I am not prioritising getting it over the line. Workaround with docker-compose works for me.

This PR is a good place to start if there's interest in switching to named volumes :)

PR #3948, opened 2022-03-10
Load currencies from WooCommerce Core

### LINKED ISSUE #2251: Migrate away from deprecated currency format data
In the Payments dashboard, we'll need to start using [locale info from WC core](https://github.com/woocommerce/woocommerce/blob/77127f93af6edd0384810a3e440fbddcbea5a751/i18n/locale-info.php#L220) instead of the **deprecated** (and more limited) info we're currently using from `@woocommerce/currency`, as we realized in review in https://github.com/Automattic/woocommerce-payments/pull/2171#discussion_r652127055.

Looks like `getCurrencyData` [[source](https://github.com/woocommerce/woocommerce-admin/blob/e0ace9a65868a28f9f6765d9538931e4c8f189bb/packages/currency/src/index.js#L206-L330)] is slated to be _removed_ in half a year or less. The function was deprecated in [this PR](https://github.com/woocommerce/woocommerce-admin/pull/5969) – note the "Is this too much data to preload" consideration there, in the context of using it in the client.

See also https://github.com/Automattic/woocommerce-payments/issues/2246

alefesouza:
Fixes #2251 

* Use the currencies registered on WooCommerce Core instead of the deprecated `getCurrencyData` method from `@woocommerce/currency` package. 

#### Testing instructions

The best way I found to test it is to edit the `locale-info.php` on the WooCommerce plugin directly:

* Open `wp-contents/plugins/woocommerce/i18n/locale-info.php`.
* Find the first occurrence of the `USD` currency, to me it's the `AS` country.
* Change the `short_symbol` key to anything else, such as `$$`.

Screen Shot 2022-03-10 at 10 22 19 PM

* Open the WooCommerce Payments Overview page, the symbol you entered should be showing there.

Screen Shot 2022-03-10 at 7 01 56 PM

It's worth noting that the EUR symbol € on the WooCommerce Core file has the `currency_pos` property set to `right_space`, so all EUR currencies on WooCommerce Payments will change from `€ 100,00` to `100,00 €`.

-------------------

**Post merge**


alefesouza:
After some research I noticed that the € symbol comes before the value in English, [but comes after it in most other European languages](https://en.wikipedia.org/wiki/Euro_sign), maybe it's worth to also use the user locale to determine the position?

malithsen:
Nice work here so far @alefesouza 👍 

> It's worth noting that the EUR symbol € on the WooCommerce Core file has the currency_pos property set to right_space,

Only some of the entries in `locale-info.php` have `currency_pos` property set to `right_space`.
I think the reason it formats the symbol differently is that there are multiple entries with `'currency_code'  => 'EUR',` in `locale-info.php`, whereas in `@woocommerce/currency` there's only a single entry. We seem to be looking up the symbols from the `currency_code` key [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/utils/currency/index.js#L42) using loadash `find`, which will return the first entry with `'currency_code': 'EUR'` in the `currencyData` array. This entry happens to belong to Andorra (`AD`) with the preferred `currency_pos` as `right_space`.

To get around this, I think one option would be to change this [method](https://github.com/Automattic/woocommerce-payments/blob/09428183e768dfe920ff0b84ed72fb9166510ae3/client/utils/currency/index.js#L39) to get the currency from store's country_code, I am not sure whether that's desirable though.

cc @dechov

alefesouza:
Thanks @malithsen!

> To get around this, I think one option would be to change this [method](https://github.com/Automattic/woocommerce-payments/blob/09428183e768dfe920ff0b84ed72fb9166510ae3/client/utils/currency/index.js#L39) to get the currency from store's country_code

I've checked and noticed that is possible to use the store country code, but what if the store is using a currency that is not from its country? For example is possible to set the store country to UK but the currency to Euro, should we get the first EUR occurrence from locale-info.php in this case?

malithsen:
Good point @alefesouza. I am not familiar with the currency related things in WCPay, but looking at the docs [here](https://woocommerce.com/document/payments/currencies/#section-5) it seems like WCPay account currency is different from store currency and it's tied to the WCPay account's country. So I think WCPay account's country is what we want to use. But we'd want to verify this approach. 

Would like to request a review from @dechov as well. Let us know if we should assign it to someone else instead.

daquinons:
> @daquinons Based on adjacent prior work, is there anyone from your team who would be available and positioned well to review?

Thanks for the ping! Let me ask a few folks, we can definitely add a pair more of eyes to review 👍

ismaeldcom (approved):
The current implementation LGTM! Thanks for working on it @alefesouza.

>I've checked and noticed that is possible to use the store country code, but what if the store is using a currency that is not from its country? For example is possible to set the store country to UK but the currency to Euro, should we get the first EUR occurrence from locale-info.php in this case?

As @malithsen pointed out, we use WC Pay account currency rather than the store one. Using store currency can lead to tricky conversions, specially on the deposits and transactions pages. You can see more about currency guidelines here: paJDYF-20W-p2.

The example you provided about different EUR formatting rules per country, is something that was happening before, partially mitigated by [`composeFallbackCurrency`](https://github.com/Automattic/woocommerce-payments/blob/09428183e768dfe920ff0b84ed72fb9166510ae3/client/utils/currency/index.js#L237) JS NumberFormat.

Now that we have access to all countries and currencies using `locale-info.php`, might be a good option to update `getCurrency` to use country currency from `wcpaySettings.connect.country` as `baseCurrency` if it's not set. That way we ensure merchant currency formatting rules are used. But I think that will be better suited in another PR, because will need proper testing of different screens and accounts countries.

alefesouza:
@ismaeldcom thanks! I'll merge and start to work on the PR to use `wcpaySettings.connect.country` as `baseCurrency`.

PR #3949, opened 2022-03-10
Fix/platform checkout my account email

mdmoore:

Platform Checkout removes the core email field. It was also being removed from the My Account > Addresses page. This ensures that the field is only affected on the Checkout page.

#### Testing instructions

* Enable Platform Checkout
* Using shortcode checkout, add a product to the cart and proceed to checkout
* Check that email field appears at the top of the checkout page
* Navigate to My Account > Addresses and add/edit a Billing Address
* Check that the email field appears on the page

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -960,10 +960,8 @@ class WC_Payments {
 	 */
 	public static function filter_woocommerce_form_field_platform_checkout_email( $field, $key, $args, $value ) {
 		$class = $args['class'][0];
-		if ( is_checkout() ) {
-			if ( false === strpos( $class, 'platform-checkout-billing-email' ) ) {
-				$field = '';
-			}
+		if ( false === strpos( $class, 'platform-checkout-billing-email' ) && ( is_checkout() || is_checkout_pay_page() ) ) {
+			$field = '';
 		}
 		return $field;
 	}
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -960,8 +960,10 @@ class WC_Payments {
 	 */
 	public static function filter_woocommerce_form_field_platform_checkout_email( $field, $key, $args, $value ) {
 		$class = $args['class'][0];
-		if ( false === strpos( $class, 'platform-checkout-billing-email' ) ) {
-			$field = '';
+		if ( is_checkout() ) {
+			if ( false === strpos( $class, 'platform-checkout-billing-email' ) ) {
+				$field = '';
+			}
 		}
 		return $field;
 	}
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments.php
```diff
@@ -960,8 +960,10 @@ public static function platform_checkout_fields_before_billing_details() {
 	 */
 	public static function filter_woocommerce_form_field_platform_checkout_email( $field, $key, $args, $value ) {
 		$class = $args['class'][0];
-		if ( false === strpos( $class, 'platform-checkout-billing-email' ) ) {
-			$field = '';
+		if ( is_checkout() ) {
+			if ( false === strpos( $class, 'platform-checkout-billing-email' ) ) {
```
ricardo, line 964: Nit: I personally would prefer these two conditions in the same `if` block, just because it's equivalent with one less instruction:

```php
if (
    false === strpos( $class, 'platform-checkout-billing-email' ) &&
    is_checkout()
) {
```

ricardo (commented):
LGTM, tests well 👍.

Do you know if it's possible to have the "Pay for order" page ever ask for billing details? When you manually create an order from wp-admin, and then go to the "Customer payment page →" link.

If that's ever possible, I think `is_checkout()` will also return `true` for that page, and we may want to add an additional check for `! is_wc_endpoint_url( 'order-pay' )`.

includes/class-wc-payments.php
mdmoore: Good call! I can't think of any scenarios within core that would load the checkout form on the Pay for Order page, but it must be possible since [is_checkout_pay_page()](https://github.com/woocommerce/woocommerce/blob/0a381f422b94e6c9866b73d5790e2bace8186ca6/plugins/woocommerce/includes/wc-conditional-functions.php#L123) exists. I'll combine those in the same block as mentioned above.

mdmoore:
@ricardo I've updated to ignore the Pay for Order page now. Let me know if you think of anything else.

includes/class-wc-payments.php
ricardo: Just to clarify, we want to hide the old billing email field from the shortcode checkout, but not from the "pay for order" page, if it's ever possible to have a billing form there, right?

I've tested this:

- `is_checkout()` ✅ for the classic checkout page, ✅ for the pay for order page.
- `is_checkout_pay_page()` ❌ for the classic checkout page, ✅ for the pay for order page.

In that case, I guess we would want to make sure we're on the checkout page, but not on the pay for order page:

```php
if ( false === strpos( $class, 'platform-checkout-billing-email' ) && is_checkout() && ! is_checkout_pay_page() ) {
```

mdmoore: >we want to hide the old billing email field from the shortcode checkout, but not from the "pay for order" page

Yep. This was a silly mistake on my part. Thanks for catching it!

PR #3951, opened 2022-03-11
KYC: New Onboarding Flow

### LINKED ISSUE #3792: KYC: Add new onboarding flow
### Description
Implement a new onboarding flow that will be shown before redirecting the merchant to Stripe KYC. It will be a dynamic form that will require business country, type, and structure, based on the selected options. And then show a summary of the required details that will be requested by Stripe.

The form data will be sent to the server to create the Stripe account with them.

The data to populate the form and summary will be handled in 1609-gh-Automattic/woocommerce-payments-server.

We should create another entry point in webpack to allow it to be easily injected in the installation process.

We will also need to handle the summary result asynchronously, requesting it from the server.

### Acceptance criteria
- The form and summary should update based on the selection.
- The entered details should be sent to the server.

### Designs
[Figma file](ka1kY306hQ3jcKBlQxVrfH-fi-596%3A19933)
|Initial|Select|Details|
|-|-|-|
|Screenshot 2022-02-14 at 16 34 53|Screenshot 2022-02-14 at 16 35 06|Screenshot 2022-02-14 at 16 35 16|

dmallory42:
Fixes #3792  

#### Testing instructions

* Check out 1773-gh-Automattic/woocommerce-payments-server.
* Go to `http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fonboarding`
* Ensure the country value is pre-populated with the WC Admin store currency (if the country is a WCPay supported country).
* Wait for the `<LoadableBlock>` to disappear and then try changing the country, business type, and structure values. Ensure the selects appear/disappear depending on whether the business type selected has structures.
* If all values are selected (either country & type, or country & type & structure), then the task should be marked as completed.

Known issues:

- Currently, the custom OnboardingSelectControl throws an error when a value is selected.
- Need to add placeholder "Please Select" to the selects.
- Need to provide strings for the description of the business type.
 

-------------------

**Post merge**

TODO:
- [ ] Remove the onboarding link from the WC Payments sidebar.


client/components/onboarding-select-control/index.js
```diff
@@ -0,0 +1,195 @@
...
+				</label>
+			}
+			<Button
+				{ ...getToggleButtonProps( {
+					// This is needed because some speech recognition software don't support `aria-labelledby`.
+					'aria-label': label,
+					'aria-labelledby': undefined,
+					className: 'components-custom-select-control__button',
+					isSmall: true,
+					// describedBy: getDescribedBy(),
```
dmallory42, line 140: I needed to comment out this line to prevent a console error - not sure why, as this was directly copied from the [CustomSelectControl](https://github.com/WordPress/gutenberg/tree/7aa042605ff42bb437e650c39132c0aa8eb4ef95/packages/components/src/custom-select-control).

client/components/onboarding-select-control/index.js
```diff
@@ -0,0 +1,195 @@
+/** @format */
+
```
dmallory42, line 2: Need to add a comment here linking to the original [CustomSelectControl](https://github.com/WordPress/gutenberg/tree/7aa042605ff42bb437e650c39132c0aa8eb4ef95/packages/components/src/custom-select-control) code.

client/data/onboarding/hooks.ts
```diff
@@ -0,0 +1,17 @@
+/** @format */
+
+/**
+ * External dependencies
+ */
+import { useSelect } from '@wordpress/data';
+import { STORE_NAME } from '../constants';
+
+export const useBusinessTypes = (): unknown =>
```
dmallory42, line 9: Should add a return type here.

client/data/onboarding/resolvers.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Internal dependencies
+ */
+import { NAMESPACE } from '../constants';
+import { updateBusinessTypes } from './actions';
+import { Country } from './types';
+
+/**
+ * Retrieve business types from the site's REST API.
+ */
+export function* getBusinessTypes(): unknown {
```
dmallory42, line 19: Add return type here.

client/onboarding/strings.js
```diff
@@ -0,0 +1,38 @@
+/** @format */
+
+/* eslint-disable max-len */
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+export default {
```
dmallory42, line 9: One thing we'll need to figure out here is that we'll need to store the descriptions for each business type (as shown on the KYC onboarding on Stripe).

I think at a minimum, we should probably just have the `US` specific ones and then a fallback generic one for other countries - but we can check and confirm with Magda.

includes/admin/class-wc-payments-admin.php
```diff
@@ -166,6 +166,18 @@ public function add_payments_menu_for_treatment() {
 				],
 			]
 		);
+		wc_admin_register_page(
```
dmallory42, line 169: Remove this later as we don't want the Onboarding to appear in the WC Admin menu.

client/components/onboarding-select-control/index.js
ismaeldcom: That prop should be `aria-describedby`, maybe our version of `@wordpress/components` doesn't have `describedBy` prop defined in `<Button/>`.

client/onboarding/add-business-info-task/index.js
```diff
@@ -0,0 +1,171 @@
...
+		( countryCode ) => {
+			return {
+				label: availableCountries[ countryCode ],
+				value: countryCode,
+			};
+		}
+	);
+
+	const [ businessCountry, setBusinessCountry ] = useState( country );
+	const [ businessType, setBusinessType ] = useState();
```
ismaeldcom, line 36: To avoid `OnboardingSelectControl` to throw a warning, you'll need to provide a default value here. Otherwise it will complain about changing a uncontrolled prop to controlled. That's because using an empty `useState` here results in it being `undefined`.

client/onboarding/add-business-info-task/index.js
```diff
@@ -0,0 +1,171 @@
...
+
+			<p className="wcpay-wizard-task__description-element is-muted-color">
+				{ strings.onboarding.countryDescription }
+			</p>
+
+			{ isLoading ? null : (
+				<OnboardingSelectControl
+					className="wcpay-onboarding-select-business-types"
+					label={ __( 'Business type', 'woocommerce-payments' ) }
+					value={ businessType }
```
ismaeldcom, line 138: The select doesn't work because we are using the received object as value here. We need to access `selectedItem`. You can for example destructure it with something like:
```jsx
<OnboardingSelectControl onChange={ ( { selectedItem } ) => handleBusinessTypeUpdate( selectedItem ) }>
```

client/onboarding/strings.js
ismaeldcom: Yes, that was my main thought about storing them in the client to easily include them like translatable strings. But we can always copy those with defined labels from the server..

client/data/onboarding/resolvers.ts
ismaeldcom: I think we should go with `unknown` for now. We will need types definitions from `@wordpress/data-controls` to properly type resolvers generators.

client/data/onboarding/hooks.ts
ismaeldcom: For return type of hooks you should use the types of the returned object, in this case it should be:
```js
{
	businessTypes: Country[];
	isLoading: boolean;
} 
```

dmallory42: Addressed in 0960b83e7e20849ec2ae6d19965aa716632ae6f8

client/components/onboarding-select-control/index.js
dmallory42: Thanks! Addressed in c7dd051342c8a1e6fde44fa9ebc488e41479ce30.

client/onboarding/add-business-info-task/index.js
```diff
@@ -0,0 +1,171 @@
...
+		setCompleted( false );
+	};
+
+	return (
+		<WizardTaskItem
+			className="complete-business-info-task"
+			index={ 1 }
+			title={ strings.onboarding.heading }
+		>
+			{ isLoading ? <p>Loading...</p> : '' }
```
ismaeldcom, line 117: You can use the `<LoadableBlock/>` component to wrap everything that should be hidden while it loads. This will avoid calling/render any unneeded method until it loads. Something like:
```jsx
<LoadableBlock isLoading={ isLoading } numLines={ 20 }>
```
And simplify all the get* and format* functions to not check if it's loading.

dmallory42: Thanks! I had put a really bad workaround in place for this while I fixed other issues. Will update to use the `LoadableBlock` solution 🙂

client/onboarding/add-business-info-task/index.js
```diff
@@ -0,0 +1,171 @@
...
+
+	const [ businessCountry, setBusinessCountry ] = useState( country );
+	const [ businessType, setBusinessType ] = useState();
+	const [ businessStructure, setBusinessStructure ] = useState();
+	const [ displayStructures, setDisplayStructures ] = useState( false );
+
+	const getBusinessTypesForCountry = ( countryCode ) => {
+		if ( isLoading || ! businessTypes.hasOwnProperty( 'data' ) ) {
+			return false;
+		}
```
ismaeldcom, line 43: Dealing with objects can get a bit tricky. We could probably simplify this by storing an array of countries in `getBusinessTypes` resolver. We're declaring a type of `Countries[]` but it's an object nested in `data`.

With an array of countries we can get rid of the checks of `data`, and store the selected one directly, not the key to access it later.

client/onboarding/add-business-info-task/index.js
```diff
@@ -0,0 +1,171 @@
...
+		if ( isLoading || ! businessTypeData ) {
+			return [];
+		}
+
+		return Object.keys( businessTypeData ).map( ( key ) => ( {
+			name: strings.businessTypes[ key ],
+			description: 'Test',
+			key: key,
+		} ) );
+	};
```
ismaeldcom, line 62: Following up with my previous comment. We can then move this function and all that are similar outside the component.

With functional components we should think that all the component will be called again, meaning that all of these function will be declared every time.

That's when `useCallback` and `useMemo` can be useful. But if we can just move them outside, will be better 🙂

client/onboarding/add-business-info-task/index.js
dmallory42: Thanks! Resolved in 9bd42017c7ebd3476557100b7ff1175cc5ba2c9b.

client/onboarding/add-business-info-task/index.js
dmallory42: Thanks, I like this idea! I want to clarify though, does this mean we'd only pass the currently selected country through to our component? How would we handle when the country is changed via the dropdown?

I'm assuming this is dealt with by the resolver somehow, but the picture is a bit unclear in my mind 😅

client/onboarding/add-business-info-task/index.js
dmallory42: Updated in 3df4f0d0fc11527e705063fd02d49d051a3e9fd4.

However, this is a little imperfect because the select briefly displays before the LoadableBlock element appears - I'm guessing this is because `isLoading` is false for a brief moment before the resolver starts loading the data. Have you encountered this before?

client/onboarding/add-business-info-task/index.js
ismaeldcom: Sorry, I wasn't clear enough! 🙂 

My idea is to make the resolver store an array of countries. Currently, it stores an object like `{data: {AE: {...}, ...}}` and I'm thinking on `['AE': {...}, ...]`. You'll probably need to update the new server PR to return the new format. And you'll receive the array while using the selector, like you already defined with types.

Then you can iterate on that array to create the select options, and when you select a country rather than storing the select country key, you can store the full country object. And later access it to look for the types and structures.

Let me know if it's clearer now, I can make a quick code example if you need it 🙂

client/onboarding/add-business-info-task/index.js
ismaeldcom: Yep! It returns false before start loading it. We can use `hasFinishedResolution` to mitigate that. Something like:
```js
isLoading: isResolving( 'getBusinessTypes' ) || ! hasFinishedResolution( 'getBusinessTypes' )
```

client/onboarding/add-business-info-task/index.js
dmallory42: As we discussed on Slack, we needed to use `''` as the placeholders, to prevent the warnings getting thrown.

includes/class-wc-payments-onboarding-service.php
```diff
@@ -0,0 +1,128 @@
...
+		$business_types_cache = [
+			'business_types' => $business_types,
+			'expires'        => time() * $expiration,
+		];
+
+		if ( false === get_option( self::BUSINESS_TYPES_OPTION ) ) {
+			$result = add_option( self::BUSINESS_TYPES_OPTION, $business_types_cache, '', 'no' );
+		} else {
+			$result = update_option( self::BUSINESS_TYPES_OPTION, $business_types_cache, 'no' );
+		}
```
ismaeldcom, line 102: I wonder if we need to use check it and use `add_option` or `update_option`. I see the following sentence in `update_option` docs 🤔
> If the option does not exist, it will be created.

client/onboarding/index.scss
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	.components-custom-select-control__item {
+		display: block;
+	}
+
+	& .components-custom-select-control__item-icon {
+		margin-left: 0 !important;
+	}
+}
```
ismaeldcom, line 24: You were right, it looks a bit strange wirth the tick displaced. And the `flex-direction: row-reverse` solution doesn't work because the selected `<li>` doesn't have any unique class added.

What do you think about using the `OnboardingSelectControl` for all of them? We can make the description optional and keep all of them consistent. And also add a comment about it being used while WP Components one doesn't provide a way to add a custom option component.

ismaeldcom (commented):
Great work @dmallory42! 🙂 I really like the cache error rate limit solution and how you structured the JS code and files.

It's working as expected and looks almost done. I see the following remaining points:
- Provide strings for the description of the business type.
- Remove `wc_admin_register_page`
- Send form details
- Show requirements based on form

But feel free to move any or all of them to new issues, to not overload this PR 🙂

includes/class-wc-payments-onboarding-service.php
dmallory42: Good catch! I thought the same, but I noticed in the account caching code we are doing this check. But I tested and it works fine just using `update_option`, so I adjusted this in 9d7e89aaa33c6c55e134f244cb69b3afc75c3e4d.

client/onboarding/index.scss
dmallory42: Yes, I thought about this and even added an [issue](https://github.com/Automattic/woocommerce-payments/issues/4028) to track it because I think we'll need to add some additional logic to the `OnboardingSelectControl` to make it display nicely.

For now, I will adjust them to use `OnboardingSelectControl`, then address a follow up PR where I add more flexibility to how the OnboardingSelectControl works!

dmallory42: Updated in 9f2a435b532fc55251b6e217fd1c959357f642cb.

dmallory42:
Thanks @ismaeldcom for all of the reviews! I have made an update to the tests using the suggestions you made to me as a base, and I think I've addressed the outstanding comments that were left over.

I also opened some [new issues](https://github.com/Automattic/woocommerce-payments/issues?q=is%3Aissue+is%3Aopen+label%3A%22component%3A+kyc+optimization%22) to track the remaining bits we need to resolve for the onboarding flow after this PR is merged.

Let me know if you think this is good to go!

ismaeldcom (approved):
Thank you for your perseverance! 🙂

LGTM! Tests are good now and all the remaining work will be handled in the new issues. Thanks for creating them.

Merge it! 🚀

PR #3955, opened 2022-03-11
Display notices when redirecting to platform checkout is in progress/fails

waclawjacek:

If clicking the "Buy now with platform checkout" button triggers an invalid response from the back-end, the customer gets redirected to `http://example.org/undefined`. The changes proposed in this PR prevent the redirection and cause displaying an error.

#### Testing instructions

1. Break the AJAX response by adding the following code above [this line](https://github.com/Automattic/woocommerce-payments/blob/8e1cd1871f8f57f766f96a20fce9778ac8e6755e/includes/class-wc-payments.php#L863):
   ```php
   wp_send_json( [ 'foo' => 'bar' ] );
   ```
1. See the redirection error after clicking the "Buy now with platform checkout" button.
1. Remove the above code.
1. See the redirection succees message after clicking the button.

#### Screenshots

Before clicking:

Markup on 2022-03-11 at 19:21:12

On error:

Markup on 2022-03-11 at 19:21:24

On success:

Markup on 2022-03-11 at 19:21:33

### DIFF
--- a/client/components/platform-checkout/index.tsx
+++ b/client/components/platform-checkout/index.tsx
@@ -4,15 +4,18 @@
 import React from 'react';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { Notice } from '@wordpress/components';
 import interpolateComponents from 'interpolate-components';
 
 /**
  * Internal dependencies
  */
 import { getConfig } from 'utils/checkout';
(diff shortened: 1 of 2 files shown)


waclawjacek:
Closing. With the SMS OTP modal displaying messaging about the redirection, this is no longer needed.

PR #3959, opened 2022-03-14
Fix `is_platform_payment_method` flag on blocks checkout

ricardo:
In #3762 we introduced platform payment methods in the shortcode checkout when the platform checkout is enabled.

More specifically in https://github.com/Automattic/woocommerce-payments/commit/7847fbfcb5e3efa18dd3c7d3fefbf1777943173f, I did some cleanup and removed a server-side check to see if a given payment method was created in the platform, but that ended up breaking the blocks and PRB checkout (when platform checkout is enabled) since both use the same `process_payment_for_order`, and I didn't anticipate that `$this->should_use_stripe_platform_on_checkout_page()` would be `true` for blocks.

Even though we have this check: `is_checkout() && ! has_block( 'woocommerce/checkout' )`, it will return `true` for the endpoint triggered from the blocks checkout.

Unfortunately, the requests coming from payment request button (express checkout), blocks and shortcode, are too similar to make use of any logic to differentiate between these types of checkouts, and let the server know which type of payment method is being used.

So the solution to this was to add an input field with the return from `$this->should_use_stripe_platform_on_checkout_page()`.

Additionally, this PR also fixes the "pay for order" page.

- Prevent Stripe JS from being initialized as platform on order-pay page: `is_wc_endpoint_url( 'order-pay' )`.
- Add form field `wcpay-is-platform-payment-method` to differentiate requests coming from blocks, shortcode and PRB.

#### Testing instructions

Note: This PR works with the `trunk` branch of the server.

- Re-build JS from this branch `npm start`.
- Make sure **Settings > Express checkouts > Platform checkout** is enabled.
- Notice shortcode, blocks and payment request button checkouts work.
- Notice the "pay for order" page works:
  - Go to wp-admin.
  - Go to **WooCommerce > Orders > Add order**.
  - Add an item to the order, and click "Create".
  - Click "Customer payment page →".
  - Checkout with a new card.
- Extra: Notice UPE still works.

-------------------

**Post merge**

- [x] QA Testing Not Applicable

### DIFF
--- a/changelog/fix-is-platform-payment-method-flag
+++ b/changelog/fix-is-platform-payment-method-flag
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix `is_platform_payment_method` flag on blocks checkout
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -818,7 +818,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				<div id="wcpay-card-element"></div>
 				<div id="wcpay-errors" role="alert"></div>
 				<input id="wcpay-payment-method" type="hidden" name="wcpay-payment-method" />
-
+				<input type="hidden" name="wcpay-is-platform-payment-method" value="<?php echo esc_attr( $this->should_use_stripe_platform_on_checkout_page() ); ?>" />
 			<?php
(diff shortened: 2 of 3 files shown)


lovo-h (approved):
Tested the following:
* Checkout through the "pay for order" page.
* Checkout through the shortcode.
* Checkout through the blocks.
* Checkout through PRB.
* Checkout through UPE (giropay).

All the tests worked as intended and the code changes look good 👍

PR #3962, opened 2022-03-15
Fix wrong update when payment intent is attached to multiple order ids in UPE

### LINKED ISSUE #3784: WC could duplicate an order between payments when checked out with UPE
### Describe the bug

Sometimes WooCommerce duplicates an order during checkout. When the successful payment webhook is processed, it incorrectly marks the initial order as completed.
In the orders mentioned in p1644259767706289/1644215203.183149-slack-C7U3Y3VMY, it looks like what caused the duplication was the coupon code not being applied on the second order, which changed the order total. But the root cause is that two orders have the same payment intent ID.

### To Reproduce

1. As a merchant, make sure you have UPE enabled and card checkbox is on.
2. As a merchant, create coupon with 1 limit usage.
3. As a shopper, buy some products, add to the cart and go to checkout.
4. Enter coupon into the input and use `4000000000009995` that should fail with insufficient funds.
5. After that, enter a valid card and do successful checkout.

### Actual behavior
As a merchant (in admin), you should see two orders with the same payment intent ID, both in status Completed.

### Expected behavior
As a merchant (in admin), you should see two orders, first (unsuccessful attepmt) is `Failed` and second one (paid-for) is `Completed`. Similar to non-UPE checkout.

### Additional context
Slack discussion where the bug was reproduced p1644341905654969/1644336878.912519-slack-C01B8KNUYSW.

jessy-p:
Fixes #3784 

When using UPE, the same payment intent can be attached to multiple orders, where only the last order is successful. When the payment intent succeeds the failed order gets updated. This change fixes this issue by picking
the last order.

- Updates the query used to pick the order to use the last or max order number.

Assumes that the last or largest order number to be updated is the one that is successful.  Looking at the [WooCommerce code](https://github.com/woocommerce/woocommerce/blob/ae2a7af95171a120d915f7a9c653b84c84aab42f/plugins/woocommerce/includes/class-wc-checkout.php#L334-L348), this is the correct assumption.

#### Testing instructions

- As a merchant, make sure you have UPE enabled and card checkbox is on.
- As a merchant, create coupon with 1 limit usage.
- As a shopper, buy some products, add to the cart and go to checkout.
- Enter coupon into the input and use 4000000000009995 that should fail with insufficient funds.
- After that, enter a valid card and do successful checkout. (coupon will get removed and order total will change)

On `trunk`, above would result in two orders, with the older failed order getting updated with the successful payment and the newer (actually successful) order not getting updated.
On this branch, the latest order should get updated

Alternatively, any steps that create a new order Id between the unsuccessful and successful attempt can be used e.g. switch the shipping rate.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-3.9.0#fix-for-failed-order-updated-when-multiple-orders-submitted-in-upe


oaratovskyi (approved):
Works like a charm!
image

I think you can ship it! :shipit: 
Thanks for fixing it 💪

Issue #3814, opened 2022-02-16
Ensure that webhooks do not cause duplicate actions


RadoslavGeorgiev:
### Description

Once https://github.com/Automattic/woocommerce-payments/issues/3813 is implemented, we'll be processing webhooks even more asynchronously than we normally would. This might cause issues, so we should address it on webhook-type-basis.

### Acceptance criteria

Build on top of https://github.com/Automattic/woocommerce-payments/issues/3813, go through [event types](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-rest-payments-webhook-controller.php#L118-L154) one by one, enable them for webhook polling, and make sure that they don't have any side-effects.

htdat:
> Once https://github.com/Automattic/woocommerce-payments/issues/3813 is implemented, we'll be processing webhooks even more asynchronously than we normally would. This might cause issues, so we should address it on webhook-type-basis.

Removing block from ZenHub as the PR for this issue https://github.com/Automattic/woocommerce-payments/pull/3827 is `ready to merge`.

htdat:
I have two following ideas to approach this issue, @RadoslavGeorgiev @deepakpathania would love to see your thoughts. Maybe @marcinbot too. 

### Idea 1 

In every Webhook_processing::process() - [src](https://github.com/Automattic/woocommerce-payments/blob/d1a65f79294d105016a39158e2de69489ac9bd95/includes/class-wc-payments-webhook-processing-service.php#L89-L89), we create a transient with 24-hour expiration to mark that we have already processed that event, something like `wcpay_process_event_evt_xxxxxxxx`. 

Before processing any other event, we check if this event exists, and only process if this transient does not exist. 

Pro: 

- We are absolutely (or at least, very close to absolutely) sure that this can prevent processing duplicate events

Cons: 

- For a huge site with a lot of transactions, aka many webhook events too, this idea can create a lot of transients. 
- Also, most of the sites are still processing these webhooks one time via Webhook_Controller. And this creates a lot of unnecessary data for most of the sites - see the stats here p1645601664622659/1645086273.563649-slack-C01BPL3ALGP


### Idea 2 

Similarly above, but we will only do the additional save and check `wcpay_process_event_evt_xxxxxxxx` when flag `has_more_failed_events` is `true` in the account data.

Pro: 

- We only check for sites with potential duplicate actions, which is more effective. 

Con: 

[...]

PR #3965, opened 2022-03-15
Fix/handle async failed refunds

### LINKED ISSUE #2285: Identify and delete a refund when an async refund failure occurs
Currently, when refunds fail asynchronously, the WC refund is still present in the Order Details page. We should try to delete the WC refund so it's not present on the order after a refund has failed.

> I think can probably (identify and) delete the refund when an async refund failure comes in, as is done in the sync refund failure case in WC core [[source](https://github.com/woocommerce/woocommerce/blob/9e9a124f528de266ede7a89bdfaa064b6b78d190/includes/wc-order-functions.php#L597-L602)] – we can open a separate issue for that though, since it's not strictly related to the change in this PR.
>
> _Originally posted by @dechov in https://github.com/Automattic/woocommerce-payments/pull/2167#issuecomment-864244565_

I looked into this briefly, and I'm pretty sure we'll need to add some meta data to either the main `WC_Order` or the `WC_Order_Refund` so we can identify the refund we should be deleting.

## Reproduction steps

1. Pay for an order with the `4000 0000 0000 5126` testing card
2. Open that order's details page
3. Refund any non-zero amount for the order
4. The refund will fail asynchronously which means:
    - You have to wait 5-10 seconds and then reload the page to see the refund failure.
    - You have to make sure webhooks work
5. Now that you've reloaded the page notice that the refund failed (there should be an order note) and the WC refund still exists in the list of items associated with the WC Order.

achyuthajoy:
Fixes #2285 

At the moment when a WCPay refund fails and the webhook request is received, we update the order meta `_wcpay_refund_status` to `failed`. However, the refund is not deleted. This PR enables fetching the last WC Refund and saving the WCPay refund id as meta, which is then used to delete WC Refunds when the failed webhook request is received.

This PR also updates the refund order notes to include the WCPay refund id and changing the order status to `failed` if the order is fully refunded. For partially refunded orders, the status is not changed.

#### Testing instructions

* Follow the steps mentioned in #2285 to replicate the problem.
* Switch to the branch `fix/handle-async-failed-refunds`.
* The changes can be tested locally, but the Jurassic Tube tunnel needs to be enabled to receive webhook requests.
* Create a new order from front-end with the card `4000 0000 0000 5126`
* Fully refund the order from WP-ADMIN and confirm that the order note added includes the WCPay refund id - `re_xxxxxxxxxxxxxxxxxxxxxxxx`
* Refresh the page to see the new order note indicating refund failure.
* Confirm that the refund create is no longer present on the order and the order status is changed to `Failed`
* Create another order using the same card details above and do a partial refund.
* Confirm that the refund create is no longer present on the order and the order status is the same as before.

-------------------

**Post merge**


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1463,19 +1465,37 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
...
-				WC_Payments_Explicit_Price_Formatter::get_explicit_price( wc_price( $amount, [ 'currency' => $currency ] ), $order )
+				WC_Payments_Utils::esc_interpolated_html(
+					/* translators: %1: the successfully charged amount, %2: refund id */
+					__( 'A refund of %1$s was successfully processed using WooCommerce Payments (<code>%2$s</code>).', 'woocommerce-payments' ),
+					[
+						'code' => '<code>',
+					]
+				),
+				WC_Payments_Explicit_Price_Formatter::get_explicit_price( wc_price( $amount, [ 'currency' => $currency ] ), $order ),
+				$refund['id']
```
htdat, line 1476: Yup, great idea to add the refund ID for relevant notes.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -202,6 +202,17 @@ private function process_webhook_refund_updated( $event_body ) {
 			),
 			$refund_id
 		);
+
+		// Delete refund based on WCPay refund id.
+		$wc_refunds = $order->get_refunds();
```
htdat, line 207: Nitpick: to aid IDE

```suggestion
		/** @var $wc_refunds WC_Order_Refund[] */
		$wc_refunds = $order->get_refunds();
```

htdat (changes requested):
Great job in figuring out proper places to add the relevant actions. 

As discussed via DMs, the main purpose of introducing `_wcpay_active_refund_id`is that during process_refund , we can not know the current refund ID. 

However, the related WC_Order_Refund object should be created before `process_refund`, we can explore using the last refund. That would eliminate introducing `_wcpay_active_refund_id` and make the code change more clear.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -202,6 +202,17 @@ private function process_webhook_refund_updated( $event_body ) {
...
+		foreach ( $wc_refunds as $wc_refund ) {
+			$wcpay_refund_id = $wc_refund->get_meta( '_wcpay_refund_id', true );
+			if ( $refund_id === $wcpay_refund_id ) {
+				$wc_refund->delete();
+				break;
```
htdat, line 212: @achyuthajoy - I am working on this issue https://github.com/automattic/woocommerce-payments/issues/3814#issuecomment-1066691553 and since you're working here, I think it would integrate it for handling `charge.refund.updated` event. 

More specifically, if there is no refund ID having the `refund_id` from the webhook data, let's stop right away without processing further.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2710,4 +2730,23 @@ function( $cost, $qty ) {
...
+	 * @param WC_Data_Store   $store WC Data Store object.
+	 */
+	public function add_refund_id_to_order_meta( $refund, $store ) {
+		$refund_id       = $refund->get_id();
+		$parent_order_id = $refund->get_parent_id();
+
+		$order = wc_get_order( absint( $parent_order_id ) );
+		if ( ! is_object( $order ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
achyuthajoy: Code updated. Removed object check.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -351,6 +351,8 @@ public function __construct(
 
 		// Update the current request logged_in cookie after a guest user is created to avoid nonce inconsistencies.
 		add_action( 'set_logged_in_cookie', [ $this, 'set_cookie_on_current_request' ] );
+
+		add_action( 'woocommerce_after_order_refund_object_save', [ $this, 'add_refund_id_to_order_meta' ], 10, 2 );
```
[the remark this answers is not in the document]
achyuthajoy: @htdat This action is triggered here - https://github.com/woocommerce/woocommerce/blob/9eca3a58250ac3b03f47dc26677c5539b1b653ae/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L200

I've removed the usage of this action based on the new approach.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1463,19 +1465,37 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
...
 				WC_Payments_Explicit_Price_Formatter::get_explicit_price( wc_price( $amount, [ 'currency' => $currency ] ), $order ),
+				$refund['id'],
 				$reason
 			);
 		}
 
+		// Get active WC refund id from order and save WCPay refund id as meta.
+		$active_refund_id = $order->get_meta( '_wcpay_active_refund_id', true );
+		if ( ! empty( $active_refund_id ) ) {
+			update_post_meta( $active_refund_id, '_wcpay_refund_id', $refund['id'] );
```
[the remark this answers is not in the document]
achyuthajoy: Code updated to use built-in methods

includes/class-wc-payment-gateway-wcpay.php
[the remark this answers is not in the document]
htdat: Gotcha! Thank you. It's really hard to catch up for this kind of variable action name.

includes/class-wc-payments-webhook-processing-service.php
achyuthajoy: Sure @dat.

Just wanted to confirm that we need to skip processing the refund event, if `_wcpay_refund_status` is already marked as failed or if we can't find a refund associated with the wcpay refund id. Correct?

htdat: Thank you. 

> if we can't find a refund associated with the wcpay refund id.

It would be only this condition :)

htdat: After discussing further, we decided to withdraw this idea and use the same approach `order_note_exists` in this PR https://github.com/Automattic/woocommerce-payments/pull/3975

Going to mark this resolved.

achyuthajoy:
@htdat I've added the unit tests for the changes. Based on the discussion (p1648029332849129-slack-CGGCLBN58) regarding setting order status, I've added commits to update the order status after deleting the refund in question.

Ready for your review again :)

achyuthajoy:
Testing instructions updated.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1485,19 +1486,39 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
...
+				),
 				WC_Payments_Explicit_Price_Formatter::get_explicit_price( wc_price( $amount, [ 'currency' => $currency ] ), $order ),
+				$refund['id'],
 				$reason
 			);
 		}
 
+		// Get the last created WC refund from order and save WCPay refund id as meta.
+		$wcpay_db       = new WC_Payments_DB();
+		$wc_last_refund = $wcpay_db->last_refund_from_order_id( $order_id );
```
htdat, line 1516: I think it would be better to move the util method `last_refund_from_order_id` to a public static method in `WC_Payments_Utils` due to 2 reasons: 

- Initializing this service `WC_Payments_DB` is a bit contrast with our effort of currently using it and plug into it the relevant services in the entry class `WC_Payments`. 
- The method itself is not tightly coupled with WCPay, and it can be used by anywhere else in our codebase. Hence, a public static function `WC_Payments_Utils` is a good fit.

includes/class-wc-payments-db.php
```diff
@@ -110,4 +110,24 @@ private function order_id_from_meta_key_value( $meta_key, $meta_value ) {
 	public function order_from_order_id( $order_id ) {
 		return wc_get_order( ( $order_id ) );
 	}
+
+	/**
+	 * Retrieve last WC refund from order ID.
+	 *
+	 * @param string $order_id WC Order ID.
+	 *
+	 * @return bool|WC_Order|WC_Order_Refund
```
htdat, line 119: To ease the usage of this util function, returning either `null` or `WC_Order_Refund` seems a better option. 

Also, I think we may try our best to avoid using the direct SQL command. IMHO, `wc_get_orders` is still a good choice to use here. It provides [enough arguments (limit, orderby, and type)](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query) to get the same result and ensure the compatibility for [the future of a custom order table](https://developer.woocommerce.com/2022/01/17/the-plan-for-the-woocommerce-custom-order-table/).

htdat: We may add tests to cover this util function too.

htdat (commented):
✅ Thanks for adding solid tests.
✅ Manual tests look great. 

💬 I've left some comments to improve the util function and the logic of updating order status a bit.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -202,6 +202,30 @@ private function process_webhook_refund_updated( $event_body ) {
...
+				$wcpay_refund_id = $wc_refund->get_meta( '_wcpay_refund_id', true );
+				if ( $refund_id === $wcpay_refund_id ) {
+					// Delete WC Refund.
+					$wc_refund->delete();
+
+					// Update order status if order is fully refunded.
+					$current_status = $order->get_status();
+					if ( 'refunded' === $current_status ) {
+						$order->update_status( 'failed' );
+					}
```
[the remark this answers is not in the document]
achyuthajoy: Moved status change logic outside loop.

includes/class-wc-payments-db.php
achyuthajoy: Updated the PHPdoc comment and replaced DB query with `wc_get_orders`. I've originally added a [commit](https://github.com/Automattic/woocommerce-payments/pull/3965/commits/a3a8935376de6c56338f804b0164b2f917e049d1) that uses `wc_get_orders` and dropped it to use DB query.

achyuthajoy:
That for the detailed reviews @htdat 👍

PR #2324, opened 2021-06-26
Enable payment methods when appropriate

FangedParakeet:
Fixes #2192 

### Refactoring UPE payment methods by extracting them from the gateway

> The goal of this issue is to provide a means to enable/disable certain pay methods within the UPE depending on the store and cart. Since Giropay is the first pay method outside of credit card in UPE we'll be using, this will be laying the foundation for the other pay methods as they become available. While the leading thought is to create a separate class (NOT a gateway) for these pay methods to centralize the functionality, other approaches can be explored as well.

This PR separates the internal logic for any UPE payment methods from the gateway itself. This makes the gateway ignorant of any payment method logic, removing any branching logic or switch cases dependent on the payment method type (Stripe key). This also allows us to define logic about whether payment methods can be saved or shown when subscription products are present in one place on a per payment method basis. 

Since all payment methods also inherit most functionality from a base class, you'll see that this will mean that we will be able to add new UPE payment methods very easily, simply by creating a lightweight new class and making the gateway aware of its existence. 

#### Testing instructions

1. Please pull these changes, as well as the latest version of the Payments Server and the WCPay Dev Utils plugin. Using the latter plugin, please ensure that `Enabled grouped settings`, `Enable UPE checkout`, `Enable UPE settings changes`, and `Add UPE additional payment methods` are all checked.
2. Either add an item from your store to cart and proceed to the regular checkout or create an order and proceed to a customer's Order Pay page to complete payment for the order (or try both, if you're feeling dangerous). Note that if you try to use the customer's Order Pay page to pay for an Order containing a subscription item (create a new Subscription, then create a parent order from this subscription, and finally visit the Order Pay page for this order), you will actually be sent to the regular checkout page instead of the Order Pay page. Anyways, from either of these pages, ensure that the regular payment flow still works and then you can focus on the following new features.
3. Within the UPE element, if you toggle between payment methods, check that the `Save payment information` checkbox (when logged into an account) is only visible and selectable when the payment method is reusable (e.g. Card and Sofort).
4. If the cart at checkout contains a Subscription product, check that only reusable payment methods are present within the UPE.

-------------------

- [x] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

PR #3968, opened 2022-03-15
POC: Multiple instances of single-method UPE at checkout

FangedParakeet:

# UPE Payment Methods as Separate Payment Gateways

The changes in this PR show a glimpse of a radical yet visionary future, where the enabled UPE payment methods exist as separate payment gateways. Please read [this P2 post](paJDYF-3gI-p2) for more context on the reasoning for this. Essentially, this makes payment options at checkout as more visually understandable for customers: no more are all existing payment methods crammed under the WooCommerce Payments' "Popular Payment Methods" umbrella at checkout, but now they can be hosted as separate payment gateways, in the same way that any other payment method would be.

UPE payment gateways at checkout
_UPE reimagined and reloaded_

In order to achieve this I have broken the existing UPE Payment Gateway into a class that supports a single payment method--as opposed to multiple--in its constructor, managing the lifetime of payment intents for only a single payment method at a time. This creates a new gateway at checkout for each payment method with a separate payment element; however, since payment elements must be primed with the client secret from a valid payment intent, this implies that for each checkout we must create a separate payment intent for each payment method, necessitating separate requests to the Payments Server to create each intent. Furthermore, once the checkout is completed, only one payment intent is updated and the rest will remain in their initial state in limbo for all eternity. Since [payment intents are now tied to the WC Session](https://github.com/Automattic/woocommerce-payments/pull/3782), a tradition I have continued in this PR, this reduces the number of excess payment intents somewhat, as refreshing the page will not create new payment intents; but they will still now always be unused payment intents at checkout.

To maintain the previous gateway for the card method, so that this gateway can use Stripe Elements instead of the UPE, when the UPE is enabled I create an instance of both the `UPE_Payment_Gateway` and `WC_Payment_Gateway_WCPay`, the latter representing the legacy card gateway that uses Stripe Elements. Whenever the `$card_gateway` is called upon, I use the `UPE_Payment_Gateway` instance, so that the new UPE settings are used; however, I insert the legacy `WC_Payment_Gateway_WCPay` instance instead into the array of available gateways when registering the gateway. This allows us to use the UPE settings while in the admin pages, while using the Stripe Elements from the previous gateway at checkout. This switcharoo probably breaks a few things, but I'm not looking for them just yet--the work here is just to prove that this is possible!

## Side-Effects

So far I have only really tested these changes with the regular checkout flow. Here are a few more things that I have either not fixed or not really tested with this refactoring of the UPE.

 - Saving payment methods / using saved payment methods.
 - Using payment methods that require additional 3DS authentication.
 - Adding payment methods from the Accounts page.
 - Purchasing items using the Pay for Order page.
 - Processing Subscription payments.
 - Blocks checkout.

This is not to say that these things are not possible--**on the contrary, I think they are all definitely very achievable!** However, they are unique processes that require specific flows, which I have not aimed to cater towards in this PR, since I only wanted to prove that we could complete a regular checkout with the UPE existing in separate payment gateways.

Additionally, adding the UPE payment gateways as separate gateways does not add the payment gateways to the main settings screen. This means that we can continue to manage these payment methods using the new settings within WCPay's own settings. Note that I haven't yet implemented those latter changes here, so if that works currently, it is another accident. 

Main payments settings
_Settings for `All payment methods` only contains WooCommerce Payments_

WCPay Settings
_WCPay main settings can function as normal_

However, the `UPE_Payment_Gateway` class inherits from the main `WC_Payment_Gateway_WCPay` class and usually there is only expected to be one of these gateways present at any time. In this implementation, each payment method is a separate gateway, which is in turn a separate instance of `UPE_Payment_Gateway`, continuing to extend its parent, so that I can make use of the `process_payment` function, as well as of the settings from `WC_Payment_Gateway_WCPay`. This means that many features become duplicated, since there will now always be more than one instance of this class at any time: this includes things such as hooks, [admin notices](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L336-L338), and AJAX functions from `WC_Payment_Gateway_WCPay`. 

You will find the most obvious occurence of this redundant replication inside the WCPay settings, where the top of the page is filled up with all of these multiplied notices. 

Notices in WCPay settings
_Notice pollution_

This is a side-effect of the main design choice of this PR in inheriting and multiplying these classes, so this will need to be solved somehow.

Also, there are no tests here and all the existing tests are probably going to be broken. Boo!

#### Testing instructions

1. Please pull the changes from this PR and ensure that you have the UPE enabled with more than one payment method enabled. 
2. Add an item to your cart and proceed to checkout.
3. Notice that the "Popular payment methods" gateway has disappeared and there is now a new separate tab for each payment method.
4. Fill in your details, select a payment method, enter your payment information, cross your fingers, and click "Place order".
5. Hopefully, the transaction completes.
6. That's all for now!

-------------------

**Post merge**


includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -852,15 +868,13 @@ function() use ( $payment_fields ) {
 			?>
 
 			<fieldset id="wc-<?php echo esc_attr( $this->id ); ?>-upe-form" class="wc-upe-form wc-payment-form">
-				<div id="wcpay-upe-element"></div>
-				<div id="wcpay-upe-errors" role="alert"></div>
-				<input id="wcpay-payment-method-upe" type="hidden" name="wcpay-payment-method-upe" />
-				<input id="wcpay_selected_upe_payment_type" type="hidden" name="wcpay_selected_upe_payment_type" />
```
FangedParakeet, line 858: Are these two inputs used or relied upon anywhere in the codebase? I couldn't find any reference to their names nor ids.

mdmoore:
Tracking the "Side Effect" type issues for this POC here for now: 

### Saving payment methods / using saved payment methods

Confirmed that cards can't be yet saved. We get this error: 

```Error: `setup_future_usage` cannot be used with one or more of the values you specified in `payment_method_types`. Please remove `setup_future_usage` or remove these types from `payment_method_types`: ["eps", "giropay", "p24"]```

### Notice Pollution
 
See the [image above](https://camo.githubusercontent.com/7d88b209ff77f0363a2912ff5f77bbc7e987cf1d53a45c350085e326a039510f/68747470733a2f2f63646e2d7374642e64726f706c722e6e65742f66696c65732f6163635f313130343230362f52714b334b42).

**Comment by harriswong** [comment]
image
I get an error clicking "add payment methods". Is this the same error you experience in the "Saving payment methods / using saved payment methods" section?

**Comment by mdmoore** [comment]
@harriswong I don't have the same error when adding payment methods. I've just tried again, enabling those specific methods and the settings save as expected.

**Comment by waclawjacek** [comment]
> ### ✅ Saving payment methods / using saved payment methods
> 
> Confirmed that cards can't be yet saved. We get this error:
> 
> `` Error: `setup_future_usage` cannot be used with one or more of the values you specified in `payment_method_types`. Please remove `setup_future_usage` or remove these types from `payment_method_types`: ["eps", "giropay", "p24"] ``

Could you please share some background on this? Does this differ from what happens in WCPay prior to this PR? Judging from this error message, if the payment method types mentioned would be removed from that UPE instance, could we then save cards?

**Comment by mdmoore** [comment]
@waclawjacek FWIW, the ✅ items were issues that have now been solved. But to expand on the issue, prior to this PR, this error wasn't occurring. This PR registers multiple gateways, whereas WCPay normally only registered either the WCPay gateway OR UPE gateway. With this PR, the intent creation was attempting to use all enabled payment types while most of those types aren't compatible reusable methods that can be used with `setup_future_usage`. 

>if the payment method types mentioned would be removed from that UPE instance, could we then save cards?

Yep. The solution is to not send all enabled payment types and only send the selected type: https://github.com/Automattic/woocommerce-payments/pull/3968/commits/f11e1556f6ae114e3abcec971639e7337e05b724

File: includes/payment-methods/class-upe-payment-gateway.php [hunk 2]
```diff
@@ -86,36 +86,41 @@ public function __construct(
...
 
-		if ( ! is_admin() ) {
-			add_filter( 'woocommerce_gateway_title', [ $this, 'maybe_filter_gateway_title' ], 10, 2 );
-		}
+			add_action( 'woocommerce_order_payment_status_changed', [ __CLASS__, 'remove_upe_payment_intent_from_session' ], 10, 0 );
+			add_action( 'woocommerce_after_account_payment_methods', [ $this, 'remove_upe_setup_intent_from_session' ], 10, 0 );
 
-		add_action( 'woocommerce_order_payment_status_changed', [ __CLASS__, 'remove_upe_payment_intent_from_session' ], 10, 0 );
-		add_action( 'woocommerce_after_account_payment_methods', [ $this, 'remove_upe_setup_intent_from_session' ], 10, 0 );
+			if ( is_admin() ) {
```
harriswong, line 120: @FangedParakeet @mdmoore Is this supposed to be `! is_admin()`?

mdmoore: I feel like this had something to do with preventing the UPE gateway titles from being changed but ill defer to @FangedParakeet.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -86,37 +86,42 @@ public function __construct(
 		WC_Payments_Customer_Service $customer_service,
 		WC_Payments_Token_Service $token_service,
 		WC_Payments_Action_Scheduler_Service $action_scheduler_service,
-		array $payment_methods,
+		UPE_Payment_Method $payment_method,
```
harriswong, line 89: @FangedParakeet I have a question. 

The `UPE_Payment_Gateway` provides UPE's payment methods. We had no control over what UPE provides and so we introduced https://github.com/Automattic/woocommerce-payments/pull/2324/. This allows us to inject a list of "payment methods" into the UPE gateway such that the UPE gateway will only display the payment methods the merchant enabled.

With this PR, we are changing the behaviour -- UPE gateway will always have 1 UPE's payment method. 

Question: Instead of injecting 1 payment method into the `UPE_Payment_Gateway` class, have we thought about not having any payment method class (the logics), but having separate UPE gateways instead? ie. `UPE_Sepa_Payment_Gateway extends UPE_Payment_Gateway`, `UPE_Giropay_Payment_Gateway extends UPE_Payment_Gateway`, etc. I wonder if it would be easier to integrate with Woo core's payment gateway because "UPE method" is now 1-to-1 to its own "UPE specific gateway".

includes/payment-methods/class-upe-payment-gateway.php
FangedParakeet: I think this _is_ actually supposed to be `is_admin()`. What happened here is that I repurposed the function `maybe_filter_gateway_title`, but neglected to update the comment describing it.

Previously the function `maybe_filter_gateway_title` was intended to be used at checkout. Since there was only a single payment gateway, containing multiple payment methods (in the UPE), we used `maybe_filter_gateway_title` to title the single gateway dynamically. If there were no payment methods enabled, we'd default to the card title, if only a single payment method were enabled we'd title the gateway the name of that payment method, and if multiple payment methods were enabled the gateway would be titled "Popular payment methods".

However, in this brave new realisation, each payment method has its own gateway at checkout, so we don't need to rename anything at checkout. Instead I am renaming the gateway in the WP Admin settings to be titled "WooCommerce Payments (`{{ Payment method title }}`)". That's why `maybe_filter_gateway_title` basically returns the property `method_title` instead of the actual `title`. 

Having explained that, I actually think this entire function is now redundant. When I was working on this, I think I originally had all the payment methods appearing as separate gateways in WP Admin (i.e. if you go to **Payments > Settings > All payment methods**). This is how the payment methods appear in the Stripe gateway, but I think after chatting to @LevinMedia (RIP), I changed this (or I did nothing and it happened by accident) so that only WooCommerce Payments would appear on this screen and users could only manage the enabled gateways through the WCPay Settings. 

**TL;DR**
Basically this wasn't a typo, but I think we can delete this whole thing--function included--anyways and that'll also be one less unit test you'll have to worry about fixing (lol).

includes/payment-methods/class-upe-payment-gateway.php
FangedParakeet: > have we thought about not having any payment method class (the logics), but having separate UPE gateways instead? ie. UPE_Sepa_Payment_Gateway extends UPE_Payment_Gateway, UPE_Giropay_Payment_Gateway extends UPE_Payment_Gateway

I am totally open to this design approach! You may not believe me, but when I was initially working on this POC that was how I initially designed this implementation--with separate named classes (e.g. `UPE_Sepa_Payment_Gateway`, `UPE_Card_Payment_Gateway`, etc.) each extending the parent `UPE_Payment_Gateway` class. However, when I was actually writing out the code, I found that there every single function and property used ended up defined in the parent class and there wasn't really anything unique I could store in these separate child classes. The individual child classes were all left empty and the only thing different about each child class was the Stripe payment method ID and the `UPE_Payment_Method` class that each gateway relied on. 

Consequently, I decided not to create separate child classes, since none of the child classes really added any unique functionality to the parent class, and instead simply created multiple instances of the parent class (`UPE_Payment_Gateway`) each with their own unique `UPE_Payment_Method`. The other benefit of this is that now we can add a new payment gateway to the list, simply by creating a `UPE_Payment_Method` class, with matching properties, and adding it to [this list](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L312-L323). We don't need to manage a separate array of all the `UPE_Payment_Gateway` classes and ensure that they match up with the `UPE_Payment_Method` classes. Payment gateway enablement is then automatically controlled by the logic inside the `WC_Payments::register_gateway` function, without having to create explicit logic around the separate `UPE_Payment_Gateway` classes.

I basically thought this saved us space, since we didn't need to create those separate class files, and we could spare a few more lines of code in the `WC_Payments` class. However, maybe it makes more sense not be so thrifty and to use explicitly named, separate classes after all. This was all mostly written quite late at night and fuelled by dangerous amounts of caffeine, so I definitely wouldn't stop anyone from redesigning this structure, if ultimately it made more sense another way! 😅

harriswong: > I found that there every single function and property used ended up defined in the parent class and there wasn't really anything unique I could store in these separate child classes. The individual child classes were all left empty and the only thing different about each child class was the Stripe payment method ID and the UPE_Payment_Method class that each gateway relied on.
> ...none of the child classes really added any unique functionality to the parent class

Thanks! These are the things I was wondering about.

The reason why I was asking is because of the unit tests. I was wondering if I should update the tests to test different "payment methods" inside the same UPE gateway; or if they should be tested as different gateways.  But yea, I agree that having 1 UPE gateway is better than splitting it up.

includes/payment-methods/class-upe-payment-gateway.php
harriswong: Sweeeet! Thanks for the explanation! I will take a look and remove it

includes/class-wc-payments.php
```diff
@@ -754,6 +799,41 @@ public static function get_file_version( $file ): string {
...
 
+	/**
+	 * Returns payment method instance by Stripe ID.
+	 *
+	 * @param string $payment_method_id Stripe payment method type ID.
+	 * @return UPE_Payment_Method Matching UPE Payment Method instance.
+	 */
+	public static function get_payment_method_by_id( $payment_method_id ) {
+		if ( ! isset( self::$upe_payment_method_map[ $payment_method_id ] ) ) {
+			return false;
```
harriswong, line 810: 👋  @FangedParakeet @mdmoore, this is falgged by `npm run psalm`.  For the `get_payment_method_by_id()`, we now update it to `return false` if the payment method isn't found.  But there are a few calls to this function that don't handle `false` very well. Can we assume the `self::$upe_payment_method_map[ $payment_method_id ]` will always exist and will return a `UPE_Payment_Method`? 

Or, should all calls to this function assume that the payment method may not exist and the caller will need to handle error properly? ie. skip the payment method, throw exception , etc

harriswong: [This](https://github.com/Automattic/woocommerce-payments/pull/4557/commits/bde68cd09312b4317f8f19e27c61ac517dd26b38) is what I did to address the psalm errors. So in terms of dealing with the psalm's errors, it will be fixed.

I was wondering more around if these functions do return `false`, what's gonna happen to all the code that calls them. Maybe we will need to go add `(if !false && ...` in various places.

FangedParakeet: > Can we assume the self::$upe_payment_method_map[ $payment_method_id ] will always exist and will return a UPE_Payment_Method?

As far as I can see, the `WC_Payments::get_payment_method_by_id()` function is only fed payment method IDs that are returned by `get_upe_enabled_payment_method_ids()` or `get_payment_method_ids_enabled_at_checkout()`, which both basically rely on these values being stored as a meta value in the `options` table. I _think_ the only way to get something invalid in there would be to manually edit those meta values, so that should be a pretty negligible notion. 

I think it's just overly defensive programming, to prevent the situation where some versions of PHP might swallow the undefined offset error and others might throw an exception. Perhaps we could throw our own exception instead, but I don't think this is really a case that can be reasonably encountered, in any case.

tommyshellberg: >Can we assume the self::$upe_payment_method_map[ $payment_method_id ] will always exist and will return a UPE_Payment_Method?
> Or, should all calls to this function assume that the payment method may not exist and the caller will need to handle the error properly? ie. skip the payment method, throw exception, etc

I'm curious about how to handle things like this as well. At the shallow level, does returning `null` make more sense than `false` here since we're trying to get something that may not exist? Because `true/false` implies the question you're asking is binary (does this thing exist?) rather than _return me the thing I'm asking for_. 
Maybe a pedantic question but I'm curious about it as a newer engineer. 🤔  

At a deeper level, I wonder about the suggestion @FangedParakeet made about throwing an Exception if we don't find a payment method. Throwing an exception could make sense if we call the method incorrectly with an illegitimate payment method ID (we want to fail hard if we screw up rather than "hide" that with defensive programming and conditionals). This assumes that we could know beforehand if it was illegitimate.

But it could be a design decision that when we loop over the available payment methods and encounter a payment method that doesn't exist (when `WC_Payments::get_payment_method_by_id()` returns `false`, we simply short-circuit and don't add it to the list of usable payment methods). Seems like it would be nice to have those pruned away to avoid it happening over and over again 🤔 

By the way, it seems possible that we'd accidentally be [calling methods on a non-object here](https://github.com/Automattic/woocommerce-payments/pull/3968/files#diff-b541c0a287f40f82a1d0e43c52c7c7a2b67ae233f261655cbb437726fec83feeR1076) if we return `false`:

```
$payment_method  = WC_Payments::get_payment_method_by_id( $payment_method_id );
	$settings[ $payment_method_id ] = [
	'isReusable'           => $payment_method->is_reusable(),
	'title'                => $payment_method->get_title(),
    ];
```

It's probably all pretty shallow and wouldn't cause us any headache; I think I'm mostly curious from a learning perspective if anyone has thoughts on it.

harriswong: WCPay initializes the `WC_Payments::$upe_payment_method_map` in its `WC_Payments::init()` function. This loops all our supported [payment methods](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L312-L325).

@FangedParakeet pointed out in the [earlier comment](https://github.com/Automattic/woocommerce-payments/pull/3968#discussion_r947091062) that the `$payment_method_id` passed to `get_payment_method_ids_enabled_at_checkout()` here is controlled by capability things:
```php
		$capturable_payment_methods = $capture ? $this->get_upe_enabled_payment_method_ids() : [ 'card' ];
		$enabled_payment_methods    = [];
		$active_payment_methods     = $this->get_upe_enabled_payment_method_statuses();
		foreach ( $capturable_payment_methods as $payment_method_id ) {
			$payment_method_capability_key = $this->payment_method_capability_key_map[ $payment_method_id ] ?? 'undefined_capability_key';
			if ( false !== $this->wc_payments_get_payment_method_by_id( $payment_method_id ) ) {

```
And essentially `$payment_method_id` is determined by `get_upe_enabled_payment_methods_ids()` which is from the option table:
```php
	public function get_upe_enabled_payment_method_ids() {
		return $this->get_option(
			'upe_enabled_payment_method_ids',
			[
				'card',
			]
		);
	}
```

So we know that it is either `card` or something from the `options` table. With all these, the only way for `WC_Payments::get_payment_method_by_id( $payment_method_id );` to be `false` is when the option table contain payment method other than the ones we listed in our `WC_Payments::init()`, which seems unlikely. 

Because this case is unexpected, I think I would choose to throw an `Exception` inside `WC_Payments::get_payment_method_by_id()` rather than failing gracefully by skipping over `false` or `null`. 

Also, because the payment method should always be one of the payment methods we created in `WC_Payments::init(), I think we don't need this condition anymore: `if ( false !== $this->wc_payments_get_payment_method_by_id( $payment_method_id ) ) {`

client/checkout/constants.js
```diff
@@ -1,7 +1,11 @@
 export const PAYMENT_METHOD_NAME_CARD = 'woocommerce_payments';
-export const PAYMENT_METHOD_NAME_BECS = 'woocommerce_payments_becs';
+export const PAYMENT_METHOD_NAME_BANCONTACT = 'woocommerce_payments_bancontact';
+export const PAYMENT_METHOD_NAME_BECS = 'woocommerce_payments_au_becs_debit';
+export const PAYMENT_METHOD_NAME_EPS = 'woocommerce_payments_eps';
 export const PAYMENT_METHOD_NAME_GIROPAY = 'woocommerce_payments_giropay';
-export const PAYMENT_METHOD_NAME_SEPA = 'woocommerce_payments_sepa';
+export const PAYMENT_METHOD_NAME_IDEAL = 'woocommerce_payments_ideal';
+export const PAYMENT_METHOD_NAME_P24 = 'woocommerce_payments_p24';
+export const PAYMENT_METHOD_NAME_SEPA = 'woocommerce_payments_sepa_debit';
```
harriswong, line 8: I just saw these in the code: 
https://github.com/Automattic/woocommerce-payments/blob/23a3b08ba18f5dc855798d233267dd44baf1ac3b/includes/payment-methods/class-upe-payment-gateway.php#L979-L987

It looks like the existing code base preferred "class constants" over "a constant file". Maybe we should take these out and use the class constant approach?

changelog/fix-4396-filter-reusable-add-payment-methods
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: This is a minor fix to a larger PR that is forthcoming.
```
harriswong, line 3: We should probably remove this changelog.

client/checkout/classic/index.js
```diff
@@ -462,30 +456,13 @@ jQuery( function ( $ ) {
...
-			let paymentMethodDetails = cardPayment;
-			if ( isWCPaySepaChosen() ) {
-				paymentMethodDetails = sepaPayment;
-			} else if ( isWCPayGiropayChosen() ) {
-				paymentMethodDetails = giropayPayment;
-			} else if ( isWCPaySofortChosen() ) {
-				sofortPayment.sofort = {
-					country: $( '#billing_country' ).val(),
-				};
-				paymentMethodDetails = sofortPayment;
```
harriswong, line 486: @FangedParakeet Are we deleting these sepa, giropay, and sofort logics because classic checkout no longer needs to handle these payment methods? I am wondering because there is another place with a similar if-then clause. I am also wondering if we can remove the function `isWCPaySepaChosen` itself.

FangedParakeet: I _think_ these functions (`isWCPaySepaChosen`, `isWCPayGiropayChosen`, and `isWCPaySofortChosen`) are completely redundant now, because the test gateways on which they relied have been removed. I think I was just cleaning up the code here, because this branching logic was confusing me a little when I was trying to figure out the chain of flow. 

There are probably related pieces that can also be removed along with this, but I think this file should only be used support card payments now.

harriswong: Perfect! Thanks!

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1178,6 +1165,9 @@ private function add_upe_payment_intent_to_session( string $intent_id = '', stri
 	public static function remove_upe_payment_intent_from_session() {
 		if ( isset( WC()->session ) ) {
 			WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT );
+			foreach ( WC_Payments::get_payment_method_map() as $id => $payment_method ) {
+				WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT . '_' . $payment_method->get_id() );
+			}
 		}
 	}
```
timur27: I was going through this function today, and I noticed that we are currently performing two actions here. 

As a first step, we **try** to remove a particular payment method from the cache (line 1167). After that, we are taking all the payment methods and payment intents for them all (lines 1168-1170). 

The thing to which I paid attention first was that 
```
WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT );
``` 

won't bring any effect since the `KEY_UPE_PAYMENT_INTENT` is only a part of the key and is lacking `_${payment_method_name}.` I think that removing the payment intent from the session works mainly thanks to the following lines, which remove everything. I would guess, that making `WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT )` line work well might be a bit tricky due to the fact, that the method is static, and I am happy to hear other opinions on this in case you already discussed it!

Not sure if unsetting everything (lines 1168-1170) is a long-term solution. But if yes, there are two things I am curious about. 

1. As we work with the split UPE, we tend to treat `class-up-payment-gateway` as an individual gateway. At the same time, currently, this method seems to have a different responsibility - it works with a set of gateways. 
2. Regarding performance, I thought today that removing unused payment intents after another method was eventually used can be avoided. This will help us to keep payment intents, let's say, for P24 and giropay (which were opened by a user on the checkout page), after the SEPA payment method was used. Thus, we'll decrease the number of requests to create a new payment intent.

FangedParakeet: > The thing to which I paid attention first was that `WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT );` won't bring any effect since the KEY_UPE_PAYMENT_INTENT is only a part of the key and is lacking _${payment_method_name}. 

Yeah, keen eye! It looks like [line 1167](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L1167) is basically doing nothing. The only relevant situation is if a user were to switch from the legacy UPE (where `self::KEY_UPE_PAYMENT_INTENT` is the cache key) to the split-UPE--but even in this case, we would never use this legacy transient again, so clearing it doesn't do us a whole lot of good. 

Basically, I absolutely agree and think that this line can be safely deleted. 

> I would guess, that making WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT ) line work well might be a bit tricky due to the fact, that the method is static, and I am happy to hear other opinions on this in case you already discussed it!

This function is static because it is also called from the legacy `WC_Payment_Gateway_WCPay` class as well ([here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payment-gateway-wcpay.php#L951) and [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payment-gateway-wcpay.php#L1023)--though, actually I think we might be able to remove this second call as well 🤔 ). This line is necessary in the scenario wherein a user selects the (legacy, non-UPE) card payment method at checkout and we would like to clear the cache for all UPE intent data that is now redundant, since another payment method has been selected instead.

Anyways, this case is essentially why this function unsets every single intent. If another non-WCPay payment method is selected, we use this function again to unset all cached UPE intent data via the hook up [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L118). In this scenario again, we would like to unset the cached data for _all_ UPE payment methods, which is why we use this static function to nuke all the transient stores. 

> As we work with the split UPE, we tend to treat class-up-payment-gateway as an individual gateway. At the same time, currently, this method seems to have a different responsibility - it works with a set of gateways.

Yep, this is the biggest problem with the split-UPE implementation. In the previous implementation, regardless of whether the UPE was enabled or not, there was only _one_ instance of the WCPay gateway class--either `WC_Payment_Gateway_WCPay` or `UPE_Payment_Gateway`. The split-UPE approach mutliplies these instances, but there are still places in the plugin where other services would like to request one gateway to perform certain functions. 

We can either use  [`WC_Payments::get_gateway()`](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments.php#L847) to return the card gateway, which will return [an instance of the UPE card gateway or legacy card gateway](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments.php#L353-L355) depending on whether the UPE is enabled or not, or we can use the static `UPE_Payment_Gateway` functions in the case where we need to perform an action on behalf of all UPE payment methods. As of now, I think the case described above is the only situation where the latter occurs.

> Regarding performance, I thought today that removing unused payment intents after another method was eventually used can be avoided. This will help us to keep payment intents, let's say, for P24 and giropay (which were opened by a user on the checkout page), after the SEPA payment method was used.

This is an interesting idea and would definitely conserve payment intents. We basically only clear all cached payment intents when a checkout has completed, but an alternate non-UPE payment method is selected. We could maintain the transient caches in this situation and continue to use the unconfirmed payment intents. However, I feel this would be inappropriate, because this intent would contain all the history and updates from a previous checkout session. This would be very confusing from a store-owner or Happiness perspective, because upon reviewing an intent's details from a store's Stripe dashboard, one would find all the amount and metadata updates to the intent from completely unrelated checkouts. 

Anyways, those are my thoughts on your revelations, but let me know if I managed to make any sense or perhaps misunderstood the points you were making here instead. 🙏

timur27:
Today I noticed that when paid with a UPE method on the classic checkout page and the payment went through successfully, this payment method is shown for users by default after they open a checkout page for the next time within the session, instead of the legacy card payment method. Is it expected, and if not, do we want to create an issue to fix this (maybe the issue already exists, I'm not sure)? @FangedParakeet @mdmoore 

https://user-images.githubusercontent.com/22319444/201338932-816fe464-597d-4b32-bcf0-5f72aaac176b.mov

mdmoore:
>this payment method is shown for users by default after they open a checkout page for the next time within the session, instead of the legacy card payment method.

This the default behavior of WooCommerce, to [set and select the `chosen_payment_method`](https://github.com/woocommerce/woocommerce/blob/f18799229bfd50d7796f05a745e590937b125dbe/plugins/woocommerce/includes/class-wc-checkout.php#L957). I'm not sure if there was ever discussion around forcing the card element to be selected. Keeping the previously used payment method for the session seems to make sense to me.

timur27:
Awesome! Then, there's no point in creating any issue for this.
Thanks for clarifying, @mdmoore!

asumaran:
Just wanted to let you know that there's some scrolling happening when the viewport is small.

frosso (approved):
Getting this in, after a long time

frosso:
Thanks for the report @asumaran , I created a ticket for it: https://github.com/Automattic/woocommerce-payments/issues/5557

PR #3971, opened 2022-03-16
Fix/3970 show currencies based on store country

### LINKED ISSUE #3970: Show currencies based on store currency
### Description
With the changes implemented on #3948 we can load all currencies from WC Core but there are some problems about currency formatting, for example the EUR symbol € on some locales can come before or after the currency value.

Based on the [comments on this PR](https://github.com/Automattic/woocommerce-payments/pull/3948#pullrequestreview-910041566), we can implement a way to show the currency based on the store country base currency, this way we will show the currencies based on the user expected way.

### Acceptance criteria

* If the currency is EUR, show € before the value if the store country is Netherlands and after the value if the store country is France.

### Designs

N/A

### Testing instructions

TBD

### Dev notes

* Get the store country from `wcpaySettings.connect.country`, then get the country currency from `wcpaySettings.currencyData`, and use it as base currency on the `getCurrency` function if there isn't anything defined.

### Additional context

PR #3948

alefesouza:
Fixes #3970

* Show the currencies on WooCommerce Payments based on the Country setting on WC Core, this way we can show the currencies in the right format.

See #3948 

#### Testing instructions

We change to change the settings on WC Core to a country that uses a different way to format currencies.

* Open WC Payments Overview.
* If the store country is US, the currency symbol should be on the left side using , as thousand separator.
* Access WooCommerce > Settings.
* Change the Country to France.
* Open WC Payments Overview.
* The currency symbol should be showing on the right ride, using a space as thousand separator.

Screen Shot 2022-03-15 at 10 03 21 PM

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-3970-show-currencies-based-on-store-country
+++ b/changelog/fix-3970-show-currencies-based-on-store-country
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Show currencies based on store country currency
--- a/client/components/account-status/account-fees/test/index.js
+++ b/client/components/account-status/account-fees/test/index.js
@@ -17,6 +17,9 @@ describe( 'AccountFees', () => {
 	beforeEach( () => {
 		global.wcpaySettings = {
 			zeroDecimalCurrencies: [],
+			connect: {
+				country: 'US',
+			},
 			currencyData: {
 				US: {
 					code: 'USD',
@@ -73,6 +76,8 @@ describe( 'AccountFees', () => {
 	} );
 
 	test( 'renders non-USD base fee', () => {
+		global.wcpaySettings.connect.country = 'FR';
+
 		const { container: accountFees } = renderAccountFees( [
 			{
 				payment_method: 'card',
@@ -124,6 +129,8 @@ describe( 'AccountFees', () => {
 	} );
 
 	test( 'renders discounted non-USD base fee', () => {
+		global.wcpaySettings.connect.country = 'UK';
+
 		const { container: accountFees } = renderAccountFees( [
 			{
 				payment_method: 'card',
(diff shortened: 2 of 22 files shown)


leonardola:
@ismaeldcom according to [this comment](https://github.com/Automattic/woocommerce-payments/pull/3948#pullrequestreview-910041566) we are not supposed to use the store country but the WC Pay account country. But `wcpaySettings.connect.country` uses [the store country not the account country](https://github.com/Automattic/woocommerce-payments/blob/17d81f20516dab05bb9fe5ec234d355401f80b79/includes/admin/class-wc-payments-admin.php#L405).

Can't this lead to problems if the store country is different from the account country for some reason?(Is it even possible in the real world?) So we would need to convert the values from one currency to another. 

Pinging you just to make sure we are loading the correct data here.

ismaeldcom:
>Can't this lead to problems if the store country is different from the account country for some reason?(Is it even possible in the real world?) So we would need to convert the values from one currency to another.

That's a good question! Sorry for not being clear enough with my previous comment, it's a bit confusing 🙂 And yes, it's possible.

I meant that we use WC Pay account currency (not country) as main currency in WC Pay, even if WC currency doesn't match. That's to avoid any conversion between merchant's payment account amounts and those shown in WC Pay transactions, deposits, etc.

And at the same time, as our current guidelines (paJDYF-20W-p2), we should format currencies using the store country/currency. Quoting the guideline:
> This will help a merchant scan and comprehend prices more quickly by displaying them in the format they’re most familiar with.

An example, I can have a Stripe account in USD and use Germany as country in WC. Amounts in WC Pay (admin) should be displayed like `1.000,24 $` rather than `$ 1,000.24`.

Hope this makes it a bit more clear 🙂  Thanks for taking care of it.

leonardola (approved):

PR #3972, opened 2022-03-16
Allow filtering API request params

dechov:

Makes a filter available for manipulating params, prior to making any WCPay API request.

Not covered by tests as the code change is trivial, but open to ideas.

#### Testing instructions

Add a filter like this and verify that the request is modified (e.g. that the statement descriptor is set on the payment):

```php
add_filter( 'wcpay_api_request_params', function( $params, $api, $method ) {
    if ( $api === WC_Payments_API_Client::INTENTIONS_API && $method === WC_Payments_API_Client::POST ) {
        $params['statement_descriptor'] = 'overridden descriptor';
    }
    return $params;
}, 10, 3 );
```

-------------------

**Post merge**


marcinbot (approved):
Looks good, as you mentioned, the change is trivial, so let's :shipit: when we're ready.

PR #3975, opened 2022-03-17
Prevent webhook duplicate actions

### LINKED ISSUE #3814: Ensure that webhooks do not cause duplicate actions

htdat:
Fixes #3814 
See more details in this Rado's comment https://github.com/Automattic/woocommerce-payments/issues/3814#issuecomment-1066691553

I did research for all webhook event types below. 

To resolve for most potential duplicate actions, I created a new method: `WC_Payments_Order_Service::order_note_exists`, and only allowed processing data if the note does not exist. 

Note: 

- I did not only check for a unique string as it's not reliable. For example, if I use the dispute ID to check if it exists in the order notes, multiple dispute webhooks with different notes can not be added. 

#### Research 

| Event Type | Before  | After |
| ------- | ------- | --------|
| charge.refund.updated | Focus on `failed` charge only, and no check to avoid duplicate actions | ~[Asking here](https://github.com/Automattic/woocommerce-payments/pull/3965/files#r827772755), handled by @achyuthajoy~ Fixed with `order_note_exists` |
| charge.dispute.created |  Add a note and update status, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| charge.dispute.closed |  Add a note and update status, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| charge.dispute.funds_reinstated |  Only add a note, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| charge.dispute.funds_withdrawn |  Only add a note, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| charge.dispute.updated |  Only add a note, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| charge.expired |  Add a note and update status, but no check to avoid duplicate actions  |  Fixed with `order_note_exists` |
| account.updated | It's OK to have duplicate actions, i.e. refresh account data  |  No change |
| wcpay.notification | Current checks have already prevented duplicate actions. See further in `put_note` and `can_note_be_added` of WC_Payments_Remote_Note_Service  |  No change |
| payment_intent.payment_failed |  Seems not enough as it applies for two payment types: `au_becs_debit` and `us_bank_account`, and checks the existing order status. |  Fixed with `order_note_exists` |
| payment_intent.succeeded |  Seems not enough and it checks the order status before processing only |  Fixed with `order_note_exists` |
| invoice.upcoming |  It seems not under my eyes but needs the check from Helix |  Ask Helix |
| invoice.paid |  It seems not under my eyes but needs the check from Helix |  Ask Helix |
| invoice.payment_failed |  It seems not under my eyes but needs the check from Helix |  Ask Helix |

#### Testing instructions

* Turn on your webhook listen for your local server.
* Use the dispute test cards as mentioned [here](https://stripe.com/docs/testing#disputes). 
* See the expected note is created in the order management (in WP Admin).
* Try to resend webhooks for your server.
* Confirm: no new note is created for the same webhook.

-------------------

**Post merge**

- [ ] Create an issue for Helix to handle subscription events.


htdat:
Ready to have the initial review to see if the idea of using `WC_Payments_Order_Service ::add_unique_note` is reasonable. 

@jessepearson - added you here too as this PR touches the order service created by you on #3675

includes/class-wc-payments-order-service.php
```diff
@@ -201,9 +203,10 @@ public function mark_payment_dispute_created( $order, $dispute_id, $reason ) {
 			return;
 		}
 
-		$this->update_order_status( $order, self::STATUS_ON_HOLD );
-		$this->add_dispute_created_note( $order, $dispute_id, $reason );
-		$order->save();
+		if ( $this->add_dispute_created_note( $order, $dispute_id, $reason ) ) {
+			$this->update_order_status( $order, self::STATUS_ON_HOLD );
+			$order->save();
+		}
```
RadoslavGeorgiev, line 209: I'm curious though what you think about this approach though:

AFAICT most `add_*_note` methods (ex. `add_dispute_created_note`) only generate the note and add it to the order. Instead of doing that, maybe they could just return the note, with us having it within a variable and being able to check whether it exists already? Then we can decouple the check for uniqueness from the addition of the note, ex.:

```suggestion
		$note = $this->generate_dispute_created_note( $order, $dispute_id, $reason );

		if ( $this->order_note_exists( $order, $note ) ) {
			return;
		}

		$this->update_order_status( $order, self::STATUS_ON_HOLD );
		$order->save();
		$order->add_note( $note );
```

I know that this looks a bit longer and cumbersome, but I like the idea of separating concerns. This way other developers wouldn't be expected to know that `add_dispute_created_note` would return `false` if the note already exists.

RadoslavGeorgiev (commented):
The changes look good so far, I like the `add_unique_note` approach, but I have a suggestion:

includes/class-wc-payments-order-service.php
htdat: Yes, makes sense. I originally went to this route but I withdrew it as it's kinda long. I am going to make the change towards this idea as it has the benefits of separating concerns and reading code easier.

htdat: >  I am going to make the change towards this idea as it has the benefits of separating concerns and reading code easier.

Done this!

htdat:
Updated the code and PR description to use `order_note_exists`. Ready to review again. 
For subscription events, I will create an issue for @Automattic/helix

tests/unit/test-class-wc-payments-webhook-reliability-service.php
```diff
@@ -201,7 +201,7 @@ public function test_fetch_events_save_data_and_schedule_process_jobs( array $pa
 				...array_map(
 					function ( $event_id ) {
 						return [
-							$this->lessThanOrEqual( time() ),
+							$this->greaterThanOrEqual( time() ),
```
htdat, line 204: See the reason for this change p1647362351781909-slack-C01BPL3ALGP

I am seeing it again in this PR https://github.com/Automattic/woocommerce-payments/runs/5599144294?check_suite_focus=true

jessepearson (approved):
The duplicate note issue was something I had considered could happen in a passing thought, but didn't look into. I like this solution and it's another step forward to cleaning up the nest of code that this plugin has become.

Thanks for taking the time to do this 😁

htdat:
> Create an issue for Helix to handle subscription events.

The issue is here https://github.com/Automattic/woocommerce-payments/issues/4032

PR #3976, opened 2022-03-17
Platform Checkout: Removes dependency on script related to express checkout methods.

brettshumaker:

This PR switches away from using `getPaymentRequestAjaxURL()` when building an AJAX URL to `buildAjaxURL()` because this feature shouldn't rely on express checkout methods being active to function. `getPaymentRequestAjaxURL()` tries to use a global variable that isn't present when express checkout methods are inactive and it was causing a js error.

#### Testing instructions
- Check out this PR.
- Make sure that `Apple Pay / Google Pay` is unchecked in settings.
- Add a product to your cart.
- Go to the shortcode checkout.
- Enter a valid platform checkout email address.
- Complete the SMS OTP challenge.
- There should be no js errors and you should be redirected.

-------------------

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : QA Testing Not Applicable

### DIFF
--- a/changelog/fix-platform-checkout-remove-express-payment-dependency
+++ b/changelog/fix-platform-checkout-remove-express-payment-dependency
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: Update to platform checkout functionality - feature not yet released.
+
+
--- a/client/checkout/api/index.js
+++ b/client/checkout/api/index.js
@@ -588,9 +588,9 @@ export default class WCPayAPI {
 
 	initPlatformCheckout() {
 		return this.request(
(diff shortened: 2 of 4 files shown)


FangedParakeet (approved):
Nice one, cheers for picking this up! This is exactly how I thought this could be fixed by relying on our own cheery little WCPay config. I found that I had to enable the UPE in order to disable the payment request buttons, which was a little cruel and unusual, but I'll investigate this separately.

Changes here do what they say on the tin and the redirect functions normally now even with the payment request buttons disabled. LGTM. 🚢

Issue #3872, opened 2022-03-01
Add "Enable fraud protection" option in the admin panel


dpaun1985:
### Description

A new toggle option should be added in the admin panel to enable/disable the fraud protection. This option will have an extra fee per transaction that should be displayed near the toggle.

### Acceptance criteria
- Add a new checkbox to enabled/disable the fraud protection under Woo Payments settings -> General.

PR #3979, opened 2022-03-17
Add fraud protection token to checkout

KarlisJ:
Fixes 1659-gh-Automattic/woocommerce-payments-server#

Introduce a new hidden field in the checkout form which is refreshed after a fraudulent attempt and validated on checkout.

This change will force users to refresh the checkout page for potential fraudulent carding attempts. 

It's achieved by rendering a hidden input containing a token that is saved in the session. Upon fraudulent payment exception, the token is regenerated and the page is refreshed. The refresh of the page ensures that incorrectly labeled payments are handled gracefully as the refresh will renderer the regenerated token. 

#### Implementation notes

Code contains TODO comments where we will read if fraud protection is enabled implemented in [#3872](https://github.com/Automattic/woocommerce-payments/issues/3872)

The current implementation does not invalidate the token after a successful purchase. 

#### Testing instructions

1. Open checkout in 2 tabs
2. Try fraudulent checkout using CC 4100000000000019
3. Assert that payment failed and the page reloaded
4. Assert that it's not possible to make payment in the 2nd tab (since the token has been refreshed)
5. Assert that valid CC number passes payment in 1st tab.

-------------------

**Post merge**


KarlisJ:
This whole functionality revolves around `WC()->session`. Is that testable somehow?

kalessil:
@KarlisJ:

> This whole functionality revolves around WC()->session. Is that testable somehow?

Nope, we need to introduce properly testable classes here with the possibility to inject session objects via a constructor.

> But I'm confused on where to put it in the directory structure?

`includes/fraud-protection`, I'd also suggest introducing a service (with a singleton approach I guess) to implement the methods (but with a focus on the token, leaving HTML in template).

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -826,6 +827,12 @@ public function payment_fields() {
 			?>
 
 			</fieldset>
+
+
+			<?php if ( Fraud_Prevention_Service::instance()->is_enabled() ) : ?>
+				<input type="hidden" id="wcpay-fraud-prevention-token" name="wcpay-fraud-prevention-token" value="<?php echo esc_attr( Fraud_Prevention_Service::instance()->get_token() ); ?>">
```
kalessil, line 833: Could we omit `id` attribute here?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -863,6 +870,17 @@ public function process_payment( $order_id ) {
 		$order = wc_get_order( $order_id );
 
 		try {
+			$fraud_prevention_service = Fraud_Prevention_Service::instance();
+			if ( $fraud_prevention_service->is_enabled() ) {
+				// phpcs:ignore WordPress.Security.NonceVerification.Missing
+				if ( ! $fraud_prevention_service->verify_token( $_POST ) ) {
```
kalessil, line 876: I'd preferer to merge the bot ifs into one.

kalessil, line 876: Let's also dispatch here the token value instead of complete array.

includes/class-wc-payment-gateway-wcpay.php
kalessil, line 833: W have a constant for value in `name` attribute.

includes/fraud-prevention/class-fraud-prevention-service.php
```diff
@@ -0,0 +1,106 @@
...
+	 * @param null $session Session instance.
+	 * @return Fraud_Prevention_Service
+	 */
+	public static function instance( $session = null ) {
+		if ( null === self::$instance ) {
+			if ( ! $session ) {
+				$session = WC()->session;
+			}
+
+			self::$instance = new self( $session );
```
kalessil, line 54: We could try this: `$session ?? WC()->session`

kalessil (changes requested):
Overall looks great! The singleton implementation is correct and I left some proposals for further improving the codebase.

includes/class-wc-payment-gateway-wcpay.php
KarlisJ: From one side I also didn't like the idea of passing whole `$_POST` but if we pass token value then we need to pull out the `isset( $post_data['wcpay-fraud-prevention-token'] )` into something like
```
if ( isset( $_POST[Fraud_Prevention_Service::REQUEST_PARAM] && $fraud_prevention_service->verify_token( $_POST[ Fraud_Prevention_Service::REQUEST_PARAM ] ) )
```
which to me looks like something that should be refactored into one method.

includes/class-wc-payment-gateway-wcpay.php
KarlisJ: Similarly to https://github.com/Automattic/woocommerce-payments/pull/3979/files#r832202589 I feel that having to use `Fraud_Prevention_Service` twice to create the input indicates that whole input creation should the the responsibility of `Fraud_Prevention_Service` with a method `get_hidden_input`.

includes/class-wc-payment-gateway-wcpay.php
kalessil: Not really: `->verify_token ( $POST[...] ?? null)`. And verify token should ensure null is not accepted as a valid value.

includes/class-wc-payment-gateway-wcpay.php
kalessil: I see. Makes sense, let's stick to the current approach then.

KarlisJ: I'm not sure about it. I agree with you that the name of field should be a constant on `Fraud_Prevention_Service` so that we are using the same value when creating the input and pulling it out from `$_POST`. 

What I don't like is having to use same class twice in the same line which is what's happening for cases. That's why I feel that `Fraud_Prevention_Service` should create the whole input as well as verify token using whole `$_POST`  because the the constant is completely isolated into the service.

kalessil: I understood, the key point here is the separation of concerns. This service has enough responsibilities, adding html-generation is not teething we should there.

That's why my last comment was to use the string and not constant.

KarlisJ:
@kalessil 
>  introducing a service (with a singleton approach

I've managed to implement unit tests for the service but I don't see any good way to assert that the singleton methods are called properly. 

I can either couple the code with tests by allowing to use `public static function set_instance` so that we can set mock instance before calling the class that uses the singleton. 

Another option is to ditch singleton and use proper dependency injection.

kalessil:
@KarlisJ, NP - testing `instance` for returning the same instance is not required.

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -1992,6 +1994,50 @@ public function test_process_payment_caches_mimimum_amount_and_displays_error_up
...
+			->with( 'incorrect-token' )
+			->willReturn( false );
+
+		$_POST['wcpay-fraud-prevention-token'] = 'incorrect-token';
+
+		$this->expectException( Exception::class );
+		$this->expectExceptionMessage( 'Your payment was not processed.' );
+		$this->wcpay_gateway->process_payment( $order );
+	}
+
```
KarlisJ, line 2040: I feel that this is missing the "success case" but then there are 2 options:
- Setting everything up for success case. It's a very complicated setup and seems to be the reason there isn't a test covering that so far.
- create a partial mock for gateway class and assert that `process_payment_for_order` is called (passed the fraud prevention checks). I gave it a try and say that the setup is complicated here as well since we need to mock things until that method like ` $this->failed_transaction_rate_limiter->is_limited()` 
- Make ` $this->failed_transaction_rate_limiter->is_limited()` fail and checking that method fail because of rate limiter. This seems doable but then the test relies on the order of execution. Don't know if that's a major issue.

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -1992,6 +1994,50 @@ public function test_process_payment_caches_mimimum_amount_and_displays_error_up
...
+		$fraud_prevention_service_mock
+			->expects( $this->once() )
+			->method( 'verify_token' )
+			->with( 'incorrect-token' )
+			->willReturn( false );
+
+		$_POST['wcpay-fraud-prevention-token'] = 'incorrect-token';
+
+		$this->expectException( Exception::class );
+		$this->expectExceptionMessage( 'Your payment was not processed.' );
```
KarlisJ, line 2037: Ideally, I'd like to assert exception code as well but it seems impossible. 

We are setting the [code when initially throwing the exception ](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L872)
```
throw new Process_Payment_Exception(
	__( 'Your payment was not processed.', 'woocommerce-payments' ),
	'rate_limiter_enabled'
);
```
And then in the `catch`	block we're[ re-throwing it but using only the message](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L952).
```
```
Is there a reason we shouldn't add the error code when re-throwing the exception?

tests/unit/test-class-wc-payment-gateway-wcpay.php
kalessil: > Ideally, I'd like to assert exception code as well but it seems impossible.

Correct, that's due to the custom exception layer.

> Is there a reason we shouldn't add the error code when re-throwing the exception?

Looking into codebase and git history, no clues on my end.

tests/unit/test-class-wc-payment-gateway-wcpay.php
kalessil: > Ideally, I'd like to assert exception code as well but it seems impossible.

Correct, that's due to the custom exception layer.

> Is there a reason we shouldn't add the error code when re-throwing the exception?

Looking into codebase and git history, no clues on my end.

tests/unit/test-class-wc-payment-gateway-wcpay.php
kalessil: I like the second option (`create a partial mock for gateway class ...`), can you give it a try.

KarlisJ: Implemented in [883ee4b06e0deeddb8adc7dc1e7358f33b7aa049](https://github.com/Automattic/woocommerce-payments/pull/3979/commits/883ee4b06e0deeddb8adc7dc1e7358f33b7aa049)

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -863,6 +870,15 @@ public function process_payment( $order_id ) {
 		$order = wc_get_order( $order_id );
 
 		try {
+			$fraud_prevention_service = Fraud_Prevention_Service::get_instance();
+			// phpcs:ignore WordPress.Security.NonceVerification.Missing,WordPress.Security.ValidatedSanitizedInput.MissingUnslash,WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
KarlisJ, line 874: I've listed all 3 ignores here but I'm not sure if that's the best approach.

Looking at other similar cases where we access `$_POST` variables I don't see anything specific being done as well as phpcs:ignore statements.

tests/unit/test-class-wc-payment-gateway-wcpay.php
kalessil: Look good!

includes/class-wc-payment-gateway-wcpay.php
kalessil: Overall this looks safe.

But we could improve the solution security:
- wp_generate_password is by default generates 12 characters long password, I'd suggest increasing the length to 16
- instead if using `===` when comparing tokens we should try using `hash_equals` (hash timing attacks prevention, when someone sends loooong values intentionally to bring the server down).

tests/unit/fraud-prevention/test-class-fraud-prevention-service.php
```diff
@@ -0,0 +1,115 @@
...
+		$this->session_mock = $this->getMockBuilder( 'WC_Session' )
+			->getMock();
+
+		Fraud_Prevention_Service::set_instance( null );
+		$this->fraud_prevention_service = Fraud_Prevention_Service::get_instance( $this->session_mock );
+	}
+
+	public function test_it_checks_if_enabled() {
+		// TODO: depends on #3994.
+		$this->markTestSkipped( 'depends on #3994' );
```
dpaun1985, line 41: I think, now we can implement the test here, #3994 is in develop branch

kalessil: +1

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -560,6 +562,30 @@ function ( $output ) {
...
+			->expects( $this->once() )
+			->method( 'get_token' )
+			->willReturn( $token_value );
+
+		// Use a callback to get and test the output (also suppresses the output buffering being printed to the CLI).
+		$this->setOutputCallback(
+			function ( $output ) use ( $token_value ) {
+				$result = preg_match_all( '/.*<input.*type="hidden".*name="wcpay-fraud-prevention-token".*value="' . $token_value . '".*\/>.*/', $output );
+
+				$this->assertEquals( 0, $result );
```
kalessil, line 582: `assertSame`?

kalessil (approved):
LGTM! Also left a couple of comments.

PR #3997, opened 2022-03-21
Compatibility fixes for themes with platform checkout.

### LINKED ISSUE #3808: Theme compatibility with platform checkout
### Description
When UPE is disabled and platform checkout is enabled, make sure the shortcode checkout is displayed properly in the following themes (in the order of highest priority to lowest):
- storefront 2 columns
- twenty-twenty 2 columns
- storefront 1 column
- twenty-twenty 1 column
- A dark theme, make sure elements are displayed properly and readable

We should timebox this no more than 1 week. If there are many things to address/fix, please create a follow-up and we will tackle it in a future sprint.

### Acceptance criteria
- Make sure shortcode checkout works with various themes described above
- Create follow-up if we can't cover all themes above.

superdav42:
Fixes #3808

Moves the email field to be inside the billing fields instead of it's own section. Fixes a few problems most obvious was with twenty twenty-two. Now it look like this in twenty twenty-two: 
image

#### Testing instructions

* Activate twenty twenty-two theme. Ensure email field is the full width of the column and the spinner shows up when entering an email address.
* Activate twenty twenty theme. Ensure iframe is full width when on small screen and the screen is resized.

-------------------

**Post merge**

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -12,13 +12,13 @@ jobs:
       max-parallel: 10
       matrix:
         # Minimum support version, ..., Latest version, `beta`.
-        woocommerce: [ '4.8.0', '4.9.0', '5.0.0', '5.1.0', '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.1', 'beta' ]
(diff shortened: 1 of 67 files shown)


superdav42:
@mdmoore I think there was a reason you used the `woocommerce_checkout_before_customer_details` action but it seems to work better this way. Can you double check?

mdmoore:
>I think there was a reason you used the `woocommerce_checkout_before_customer_details` action but it seems to work better this way. Can you double check?

I was initially trying the `woocommerce_before_checkout_billing_form` hook, but it's inside the Billing Details section. `woocommerce_checkout_billing` is perfect.

superdav42:
@mdmoore That's odd. Maybe we can do a 1:1 session to find what the difference may be.

PR #4002, opened 2022-03-21
Bump version and tested up to

leonardola:
This PR bumps the version of Tested Up to and Minimum required versions of WC and WP. It also removed the now unnecessary conditional to skip tests on older versions of WC

#### Testing instructions

Make sure all versions are correct
Make sure all tests pass


brianyu28 (approved):
This looks good to me, thanks @leonardola!

PR #4005, opened 2022-03-21
Handle the case when some attributes from WC's currency data is null.

### LINKED ISSUE #4001: Payment details page shows errors for subsription transaction with WC 4.8.0
### Describe the bug

With WC 4.8.0 and subscription order, payment detail page (`/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Ftransactions%2Fdetails&id={ch_xxx}`) shows error.

### To Reproduce

1. Purchase a subscription product.
2. Open the payment detail page of that order. You can click the link on the edit order admin page as shown below.

Screen Shot 2022-03-22 at 02 20 24

### Actual behavior

Parts of the payment detail page do not render. Instead, it shows errors `TypeError: Cannot read property 'toString' of null`.

### Screenshots

screenshot_of_failed_test

### Expected behavior

No error.

### Additional context
I think this commit 02259ac19510d6d7cfc86f826fc24e22fc1092cc where the problem starts. Commit before it doesn't give the error.

Update: I understand more of the problem now.
In commit 02259ac19510d6d7cfc86f826fc24e22fc1092cc, specifically changes in `client/utils/currency/index.js`, we stop getting currency data from `@woocommerce/currency`'s `getCurrencyData()` but instead getting it from `wcpaySettings.currencyData`, which is passed from PHP `includes/admin/class-wc-payments-admin.php`.

We get the currency data from Woo's `i18n/locale-info.php`.
[WC 4.8's i18n/locale-info.php](https://github.com/woocommerce/woocommerce/blob/4.8.3/i18n/locale-info.php) doesn't have `short_symbol`.
Not sure when `short_symbol` is added but the [latest version](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/i18n/locale-info.php) has it.

The error is because symbol is null and [woocommerce-admin's CurrencyFactory just toString() it without checking for null](https://github.com/woocommerce/woocommerce-admin/blob/main/packages/currency/src/index.js#L27).

Also, maybe this error doesn't only happen for subscriptions order but for any kind of order.

Screen Shot 2022-03-22 at 04 35 42

shendy-a8c:
Fixes #4001

Make sure we use empty string when some of the attributes from WC's currency data are null.

I put more detailed of the problem and my investigation in the issue.

Since this PR is fixing a problem that hasn't been released yet, I do not need to add a changelog (right?).

#### Testing instructions

0. Use WC 4.8.
1. Purchase a subscription product.
2. Open the payment detail page of that order.

With head branch, expect no error.
With base branch, expect an error.

Also, expect green E2E tests for WC 4.8.
With base branch, `WC - 4.8.0 | subscriptions - merchant` [failed](https://github.com/Automattic/woocommerce-payments/runs/5622680470?check_suite_focus=true).

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _Add link here / 'QA Testing Not Applicable'_


shendy-a8c:
Is the changelog necessary and how do I not submit a changelog without failing GH checks?

aprea:
Hey @shendy-a8c, it looks like yesterday we bumped the minimum WooCommerce version to 5.0 😄 https://github.com/Automattic/woocommerce-payments/pull/4002

These changes look pretty harmless so we can probably still proceed with this PR, just wanted to give you a heads up.

aprea (approved):
I haven't been able to test this in my local environment with WC 4.8 😢

Every WooCommerce admin page is rendering as completely blank for me and there's an error in the console. It looks related to `woocommerce-admin`.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'plugins')
    at Object.297 (index.js?ver=1648011344:2:129300)
    at a (index.js?ver=1648011344:2:1209)
    at Module.<anonymous> (index.js?ver=1648011344:2:216093)
    at Module.561 (index.js?ver=1648011344:2:217380)
    at a (index.js?ver=1648011344:2:1209)
    at wc.app.0 (index.js?ver=1648011344:2:3963)
    at index.js?ver=1648011344:2:3974
```

With that said, the code change looks harmless and I haven't observed any adverse side effects on newer versions of WooCommerce.

As such, I'm approving this PR.

leonardola:
Just so you know. The error you were seeing in the console is related to [WooCommerce Blocks](https://woocommerce.com/products/woocommerce-gutenberg-products-block/) which is not compatible with WC versions older than 6.0.0. Disabling WooCommerce Blocks fixes it.

PR #4006, opened 2022-03-21
Replace the "colors" dependency with chalk

### LINKED ISSUE #3961: Remove our "colors" dependency
Some context about it: https://www.theverge.com/2022/1/9/22874949/developer-corrupts-open-source-libraries-projects-affected?scrolla=5eb6d68b7fedc32c19ef33b4 (there's also an internal P2 about it).

We don't use the `color` dependency in any mission-critical code, just in the release script. In any case, migrating to [chalk](https://www.npmjs.com/package/chalk) is trivial, so we should do that and get it out of the way.

DanReyLop:
Fixes #3961

Replace the "colors" dependency with [chalk](https://github.com/chalk/chalk). We only use it to style the output of a few messages on the `release` script, so this change has no risk. I was tempted to just drop the dependency and let the success & error messages at the end of the release script be boring white-on-black, but whatever, we're already pulling the `chalk` dependency transitively.

#### Testing instructions

- Run `npm run build:release`.
- Notice that the script doesn't crash.
- Notice that the script prints a green message at the end.


ismaeldcom (approved):
LGTM! Thanks for keeping the project tidy 🙂

Issue #3810, opened 2022-02-15
Subscription products compatibility with platform checkout 


harriswong:
### Description
~Investigate if subscription products are compatible with platform checkout. If not compatible, can we fix it? Should we support subscription products?~
Subscription is not compatible with platform checkout. We need a way to disable platform checkout when there are subscription products involved in checkout. These "subscription product" could come from WCPay's subscription or woocommerce-subscriptions. 

## Suggestions
- When subscription products in cart, then maybe we don't enqueue scripts for platform checkout?

### Acceptance criteria
- If platform checkout is enabled, and if subscription product is in cart, WCPay should still checkout without platform checkout, and it should not throw error.

harriswong:
@frosso I want to double check with you on this issue. Is the description accurate?

frosso:
@harriswong that looks good! 👍

PR #4020, opened 2022-03-23
Disable platform checkout if subscription product is in cart

harriswong:
Fixes https://github.com/Automattic/woocommerce-payments/issues/3810

On the `/checkout` page, if there are any subscription products in the cart, then we do not load the platform checkout features. 

#### Testing instructions
1. Enable platform checkout feature.
2. Checkout product that are not subscription products. Confirm "email address", "Remember your details" shows up in checkout page.
image
3. Create a subscription product, add it to cart and checkout
4. The "email address" at the top and "Remember your details" are no longer there.
image

**Post merge**


includes/class-wc-payments.php
```diff
@@ -239,21 +239,15 @@ public static function init() {
...
 		include_once __DIR__ . '/class-wc-payments-webhook-reliability-service.php';
+		include_once __DIR__ . '/platform-checkout/class-platform-checkout-utilities.php';
 
 		// Load customer multi-currency if feature is enabled.
 		if ( WC_Payments_Features::is_customer_multi_currency_enabled() ) {
 			include_once __DIR__ . '/multi-currency/wc-payments-multi-currency.php';
 		}
 
-		// Load platform checkout save user section if feature is enabled.
-		if ( WC_Payments_Features::is_platform_checkout_eligible() ) {
```
harriswong, line 249: I want to point out here that the code inside this condition is used to load in the `plugins_loaded` hook. Now, it will be loaded after the cart content is populated.

includes/class-wc-payments.php
```diff
@@ -296,8 +290,6 @@ public static function init() {
 			self::$card_gateway = new $card_class( self::$api_client, self::$account, self::$customer_service, self::$token_service, self::$action_scheduler_service, self::$failed_transaction_rate_limiter, self::$order_service );
 		}
 
-		self::maybe_register_platform_checkout_hooks();
```
harriswong, line 299: This code used to load in the `plugins_loaded` hook. Now, it will be loaded after the cart content is populated, inside the `init_platform_checkout()` function along with other platform checkout things.

brettshumaker (approved):
LGTM. Everything works as expected and tests well.

includes/class-wc-payments.php
superdav42: I'm worried moving it till after the 'woocommerce_cart_loaded_from_session' will cause problem because I think some of those hooks need to be registered sooner. It may work for you now but things could change or just happen in different orders on different sites.

harriswong: Good point. Instead of loading the entire `maybe_register_platform_checkout_hooks` in the  `woocommerce_cart_loaded_from_session` hook, maybe load only the `woocommerce_checkout_billing` action and `woocommerce_form_field_email` filter after cart is loaded.

This way, only the action/filter related to the email field is pushed off till the cart items are loaded. The rest of the platform checkout cookie script can stay where it is.

includes/class-wc-payments.php
```diff
@@ -852,8 +846,6 @@ public static function maybe_register_platform_checkout_hooks() {
 			add_filter( 'woocommerce_cookie', [ __CLASS__, 'determine_session_cookie_for_platform_checkout' ] );
 			// Disable nonce checks for API calls. TODO This should be changed.
 			add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
-			add_action( 'woocommerce_checkout_billing', [ __CLASS__, 'platform_checkout_fields_before_billing_details' ], -50 );
-			add_filter( 'woocommerce_form_field_email', [ __CLASS__, 'filter_woocommerce_form_field_platform_checkout_email' ], 20, 4 );
```
harriswong, line 856: These 2 action and filter will load after the cart is loaded; after the `woocommerce_cart_loaded_from_session` hook.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -631,7 +634,7 @@ public function get_payment_fields_js_config() {
 			'locale'                         => WC_Payments_Utils::convert_to_stripe_locale( get_locale() ),
 			'isUPEEnabled'                   => WC_Payments_Features::is_upe_enabled(),
 			'isSavedCardsEnabled'            => $this->is_saved_cards_enabled(),
-			'isPlatformCheckoutEnabled'      => WC_Payments_Features::is_platform_checkout_eligible() && 'yes' === $this->get_option( 'platform_checkout', 'no' ),
+			'isPlatformCheckoutEnabled'      => $platform_checkout_util->should_enable_platform_checkout( $this ),
```
harriswong, line 637: This flag handles the checkout email input https://github.com/Automattic/woocommerce-payments/blob/0ab629c971bfe7ab61646ca96cbde786ff121b6f/client/checkout/classic/index.js#L561-L563.

It will now also check if subscription product is in cart before enabling.

harriswong:
Hi @brettshumaker @superdav42, I updated the code a bit. I now load only hook the "email input" related codes to the `woocommerce_cart_loaded_from_session` hook. The original location of `self::maybe_register_platform_checkout_hooks();` remains unchanged.

includes/class-wc-payments.php
harriswong: Updated the code, it should be better. Let me know what you think!

harriswong: @superdav42 bump

superdav42 (approved):
LGTM!

PR #4022, opened 2022-03-24
Allow file upload to select same file after error

### LINKED ISSUE #460: File upload does not start when selecting the same file
The issue is in the evidence form with FileUpload component which doesn't trigger onChange event when the same file is selected.

### Steps to reproduce

1. Go to dispute evidence form.
2. Select a file which is not allowed to upload, e.g. .docx. The upload should fail.
3. Select the same file. Upload won't start again.

Potential solution: introduce error prop in the FileUpload component, and reset input value if error is passed.

alefesouza:
Fixes #460 

* Fixes an error that occur when a user upload a file and have some connection or general error, if select the same file again it does not emit the onChange event and the user get stuck until select a different file.

Needed to use a workaround to empty the file input because WordPress/gutenberg#39268 will be implemented on the next Gutenberg release and WCPay still uses the version 5.1.2 of it.

#### Testing instructions

* Buy something on a merchant store and [add a payment that generates a dispute](https://woocommerce.com/document/payments/testing/#section-15).
* Access the Transactions section on the left panel and open the latest transaction, if it does not work (this page is always empty to me), open the latest order and click on the "pi_hash" link.

Screen Shot 2022-03-23 at 9 45 36 PM

* On this page there is a "View dispute" link, click on it.

Screen Shot 2022-03-23 at 9 47 39 PM

* Click on "Challenge dispute".
* Change the "Product type".
* The file input button will be on this page.

Screen Shot 2022-03-23 at 9 49 49 PM

* Now we need to make an error occur, to do it, just select a file that is not a PDF/JPG/PNG file.
* Now select the same file again, it will show the loading animation, before this PR the user would get stuck because the onChange event was not called.
* You can also disable the internet on the browser DevTools and try to upload a file, enable the internet again and selected the same file, before this PR the "You are offline" message would be showing until the user select another file.

-------------------

**Post merge**

### DIFF
--- a/client/components/file-upload/index.tsx
+++ b/client/components/file-upload/index.tsx
@@ -45,7 +45,11 @@ export const FileUploadControl = ( {
 		event: React.MouseEvent< HTMLButtonElement >,
 		openFileDialog: () => void
 	) => {
-		// Get file input next to the button element and clear it's value, allowing to select the same file again in case of connection or general error or need to select it again
+		// Get file input next to the button element and clear it's value,
+		// allowing to select the same file again in case of
+		// connection or general error or just need to select it again.
+		// This workaround is useful until we update @wordpress/components to a
+		// version the supports this: https://github.com/WordPress/gutenberg/issues/39267
 		const fileInput:
 			| HTMLInputElement
 			| null
--- a/client/components/file-upload/index.tsx
+++ b/client/components/file-upload/index.tsx
@@ -41,6 +41,25 @@ export const FileUploadControl = ( {
 		/>
 	);
 
+	const handleButtonClick = (
+		event: React.MouseEvent< HTMLButtonElement >,
+		openFileDialog: () => void
+	) => {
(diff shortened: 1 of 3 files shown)


client/components/file-upload/index.tsx
```diff
@@ -41,6 +41,25 @@ export const FileUploadControl = ( {
 		/>
 	);
 
+	const handleButtonClick = (
+		event: React.MouseEvent< HTMLButtonElement >,
+		openFileDialog: () => void
+	) => {
+		// Get file input next to the button element and clear it's value, allowing to select the same file again in case of connection or general error or need to select it again
```
brianyu28, line 48: I'd suggest leaving a comment here that links to https://github.com/WordPress/gutenberg/issues/39267, so that it's clear why this workaround is being used. That'll also make it clearer later when we might want to use `onChange` directly on `FormFileUpload` once it's supported.

brianyu28 (approved):
This looks good to me, thanks @alefesouza! I left one suggestion about leaving a comment documenting the relevant Gutenberg issue, otherwise this looks good and works well.

PR #4041, opened 2022-03-28
Prevent WCPay server refund of interac_present

### LINKED ISSUE #4105: [Card Present Payments] Allow refunds of `interac_present` payments to be done via the apps
Interac is used commonly in Canada, and Interac require that an in-person refund flow is used. To quote [Stripe's docs](https://stripe.com/docs/terminal/features/refunds#client-side):

> In-person refunds are mandatory for Interac transactions in Canada. You can’t create refunds in the API or in the Dashboard for these payments; you must use the client-side methods. In this flow, the reader prompts the cardholder to present the card used in the original charge. After the card details are read successfully, your application processes the refund.

Calls to the `refunds` endpoint for `interac_present` refunds should update the order details with notes as normal, but not actually process the refund. They should only succeed when the refund has already been successfully processed by Stripe, via the mobile apps.

joshheald:
Fixes #4105

Depends on https://github.com/Automattic/woocommerce-payments-server/pull/1864

1. Prevent `refund_charge` call to WCPay Server for `interac_present` refunds – these should be refunded in the mobile apps.
2. Add notes as if the refund had been performed in this case regardless. 
3. Prevent refunds from wp-admin of `interac_present` refunds 

#### Outstanding questions

There is an `Order-refund` object created as part of this flow. Does anything need to be updated on this object with data from the Stripe refund? For example, do we need to set `refunded_payment` to true?

This change allows the apps to process refunds of `interac_present` transactions, and still update the Order with the details of those refunds.

Interac is used commonly in Canada, and Interac require that an in-person refund flow is used. To quote [Stripe's docs](https://stripe.com/docs/terminal/features/refunds#client-side):

> In-person refunds are mandatory for Interac transactions in Canada. You can’t create refunds in the API or in the Dashboard for these payments; you must use the client-side methods. In this flow, the reader prompts the cardholder to present the card used in the original charge. After the card details are read successfully, your application processes the refund.

The client-side (app-side) refunds that are created in the apps are passed directly to Stripe from the apps, as that is how the StripeTerminal SDK works. The apps will then call the WCPay Refund API as normal, which leads to the `process_refund` function.

This change identifies these `interac_present` refund calls, and avoids calling `refund_charge` for them, as that would error and risk a double-refund situation. 

Instead, it then checks for the existence of the refund with Stripe. If the refund is found, it records the notes of the refund on the order and proceeds as normal. If the refund is not found, it assumes that the API call was made from wp-admin and the refund has not been processed in the apps. It provides a 404 (as the refund was not found) with an error message specific to why the refund failed.

##### App refund
N.B. the refund flow for interac is not yet finished in the apps. To test, a build of [this branch](https://github.com/woocommerce/woocommerce-ios/tree/test/6445-stripe-ipp-client-side-refunds-hacked-refund-flow) should be used, until a PR that fixes https://github.com/woocommerce/woocommerce-ios/issues/5983 has been merged.

1. Use a store set up for Canada
4. In the app, add an order from the Orders tab, using the `+` button 
5. Take payment for the order using a WISEPAD_3 and Interac test card
6. Open the order, and tap `Issue Refund`
7. Go through the refund flow and observe that the refund is successful
8. Open the order in WP-Admin and observe that the refund has been recorded.

https://user-images.githubusercontent.com/2472348/162271616-88ec6d2c-0ed1-4272-9c32-3a71b8f7acad.mp4

##### WP-admin refund attempt
1. Use a store set up for Canada
2. In the app, add an order from the Orders tab, using the `+` button 
3. Take payment for the order using a WISEPAD_3 and Interac test card
4. Go to the order in wp-admin
5. Attempt to refund it and observe that you're shown an error message informing you that this must be done in the app.

Interac in-person refund failure

-------------------

**Post merge**


tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -426,6 +426,101 @@ public function test_process_refund_with_reason_non_usd() {
 		$this->assertTrue( $result );
 	}
 
+	public function test_process_refund_interac_present() {
```
joshheald, line 429: This test fails due to [the addition of the refund ID to the notes](https://github.com/Automattic/woocommerce-payments/blob/0af1f4305851f30aa38a0f598d79d6442288f8be/includes/class-wc-payment-gateway-wcpay.php#L1520), I'm looking at how to resolve that for `interac_present` transactions, and whether we should fetch a refund from Stripe on the server to store somewhere too.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1554,6 +1558,20 @@ public function has_refund_failed( $order ) {
 		return 'failed' === $order->get_meta( '_wcpay_refund_status', true );
 	}
 
+	/**
+	 * Gets the payment method type used for an order, if any
+	 *
+	 * @param WC_Order $order The order to get the payment method type for.
+	 * @return string
+	 */
+	protected function get_payment_method_type_for_order( $order ) {
```
kalessil, line 1567: I'd suggest making the method stricter: `private function get_payment_method_type_for_order( $order ): string`

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1554,6 +1558,20 @@ public function has_refund_failed( $order ) {
...
+		$payment_method_id = $order->get_meta( '_payment_method_id', true );
+
+		$payment_method_details = $this->payments_api_client->get_payment_method( $payment_method_id );
+		$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : 'unknown';
```
kalessil, line 1571: `return $payment_method_details['type'] ?? 'unknown'` would be easier to read.

kalessil (commented):
Overall looks great! Left a couple of comments with improvement suggestions.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1479,13 +1480,35 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
...
+						return 'succeeded' === $r['status'] && $stripe_expected_refund_amount === $r['amount'];
+					}
+				);
+
+				if ( isset( $filtered_refunds[0] ) ) {
+					$refund = $filtered_refunds[0];
+				} else {
+					$interac_refund_error = __( 'Interac in-person payments must be refunded using the mobile app, with the customer present.', 'woocommerce-payments' );
+					return new WP_Error( 'wcpay_edit_order_interac_present_refund_failure', $interac_refund_error, [ 'status' => 404 ] );
+				}
```
joshbetz, line 1502: Minor: Reverses our check that a refund exists, which arguably makes this easier to read.

```suggestion
				if ( ! isset( $filtered_refunds[0] ) ) {
					$interac_refund_error = __( 'Interac in-person payments must be refunded using the mobile app, with the customer present.', 'woocommerce-payments' );
					return new WP_Error( 'wcpay_edit_order_interac_present_refund_failure', $interac_refund_error, [ 'status' => 404 ] );
				}
				
				$refund = $filtered_refunds[0];
```

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -363,6 +363,28 @@ public function refund_charge( $charge_id, $amount = null ) {
...
+	 * List refunds
+	 *
+	 * @param string $charge_id - The charge to retrieve the list of refunds for.
+	 * @param string $payment_intent_id - The payment intent to retrieve the list of refunds for.
+	 *
+	 * @return array
+	 * @throws API_Exception - Exception thrown on request failure.
+	 */
+	public function list_refunds( string $charge_id = null, string $payment_intent_id = null ) {
+		$query = [ 'limit' => 100 ];
```
joshbetz, line 376: Minor: I suspect this doesn't matter since we're asking for a specific charge_id, but do you think we can set this limit to 1?

joshbetz (commented):
Looks good from my perspective, but I will defer to someone who is more familiar with the codebase.

includes/wc-payment-api/class-wc-payments-api-client.php
joshheald: We shouldn't limit it to one, since Stripe will return all the refunds for the charge, including failures, or previous partial refunds, which may not match the refund we're looking for.

You've made me realise that there is a potential for a charge to have lots of refunds on it and then become impossible to refund here, but I think that's very unlikely for an interac_present refund.

kalessil (approved):
LGTM! I'll wrangle a couple more manual tests and will merge it on Monday.

PR #3994, opened 2022-03-21
Add fraud protection section in settings page

elazzabi:
Fixes #3872 

#### Changes proposed in this Pull Request

This PR adds a fraud protection section in the settings page, currently **behind a feature flag**.

#### Testing instructions

1. This feature is currently behind a feature flag, make sure to turn the `isFraudProtectionSettingsEnabled` flag on in the `client/settings/settings-manager/index.js` file
2. Visit the WCPay settings page, and scroll to the bottom
3. You should see a new section like the screenshot below
4. Turn it on and save the settings
5. Reload the page
6. Check that the setting is still on (saved as WP option)

Screen Shot 2022-03-21 at 10 56 33

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) :'QA Testing Not Applicable - Feature is behind a feature flag

PR #4047, opened 2022-03-29
Add data fingerprinting for fraud prevention in intent creation/update

### LINKED ISSUE #3875: Send fraud protection related fingerprinting data on payment intent meta
### Description

Given the following data we want to store for fingerprinting: paJDYF-3ja-p2

 1 Evaluate hashing algorithm we could use (with respect to min supported PHP version; using salt vs not using salt; consider performance)
2 If fraud protection is enabled, inject the necessary hashed data into payment intent (lowercasing the input data make sense; also supply fingerprinting_avaialable=true as part of the metadata).

Context

If fraud protection is enabled, we'll be sending fingerprinting data as part of the intent.

elazzabi:
Fixes #3875 

## Add data fingerprinting for fraud prevention in intent creation/update
Following our work to reduce fraud on WC Payments, we'll hash some customer data and use it afterward to decide if the payment is fraudulent or not.

#### Testing instructions

This code isn't currently used given the `is_fraud_prevention_enabled_for_store` function returns `false`. This is a way of keeping it behind a feature flag :)

-------------------

**Post merge**


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2063,4 +2086,25 @@ public function get_active_loan_summary() : array {
...
+	}
+
+
+	/**
+	 * Checks if fraud prevention is enabled for this store.
+	 *
+	 * @return bool True if store has fraud prevention enabled.
+	 */
+	public function is_fraud_prevention_enabled_for_store() {
+		return false;
```
elazzabi, line 2108: This is a way of making this behind a "feature flag" :)

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -263,6 +263,19 @@ public function create_intention(
 		$request['payment_method_types'] = $payment_methods;
 		$request['capture_method']       = $capture_method;
 
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fingerprinting_avaialable'] = true;
```
dmvrtx, line 267: I suggest to replace `fingerprint` with `fraud_prevention` here and further to make the purpose of data collection clear to customers.

elazzabi: I agree, I think that makes more sense. I'll work on it

includes/wc-payment-api/class-wc-payments-api-client.php
kalessil: With https://github.com/Automattic/woocommerce-payments/pull/3994 merged, I believe we can refactor this.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -263,6 +263,20 @@ public function create_intention(
 		$request['payment_method_types'] = $payment_methods;
 		$request['capture_method']       = $capture_method;
 
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+			$order               = wc_get_order( $order_id );
+			$billing_email_parts = explode( '@', $order->get_billing_email() );
+			$email_domain        = array_pop( $billing_email_parts );
```
kalessil, line 270: `$email_domain = explode( '@', $order->get_billing_email(), 2 )[1];`

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -263,6 +263,20 @@ public function create_intention(
...
+
+			$request['metadata']['fraud_prevention_data'] = [
```
kalessil, line 272: Similar to https://github.com/Automattic/woocommerce-payments/pull/3979/files. we should introduce a `Buyer_Fingerprinting_Service` which will take care of the metadata generation for the order.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -318,6 +332,16 @@ public function update_intention(
 			$request['setup_future_usage'] = 'off_session';
 		}
 
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+
+			$request['metadata']['fraud_prevention_data'] = [
```
kalessil, line 338: Also to be moved into a service class.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2150,4 +2174,25 @@ public function get_active_loan_summary() : array {
...
+		return hash( 'sha512', $data );
+	}
+
+
+	/**
+	 * Checks if fraud prevention is enabled for this store.
+	 *
+	 * @return bool True if store has fraud prevention enabled.
+	 */
+	public function is_fraud_prevention_enabled_for_store() {
```
kalessil, line 2195: Also to be moved into a service class.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2150,4 +2174,25 @@ public function get_active_loan_summary() : array {
...
 	}
+
+	/**
+	 * Hashes customer data for the fraud prevention.
+	 *
+	 * @param string $data The data you want to hash.
+	 *
+	 * @return string Hashed data.
+	 */
+	public function hash_data_for_fraud_prevention( $data ) {
```
kalessil, line 2185: Also to be moved into a service class.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -263,6 +263,20 @@ public function create_intention(
...
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+			$order               = wc_get_order( $order_id );
+			$billing_email_parts = explode( '@', $order->get_billing_email() );
+			$email_domain        = array_pop( $billing_email_parts );
+
+			$request['metadata']['fraud_prevention_data'] = [
+				'shopper_ip_hash'        => $this->hash_data_for_fraud_prevention( $order->get_customer_ip_address() ),
+				'shopper_useragent_hash' => $this->hash_data_for_fraud_prevention( $order->get_customer_user_agent() ),
+				'shopper_email_hash'     => $this->hash_data_for_fraud_prevention( $order->get_billing_email() ),
```
kalessil, line 275: Email is injected into other meta-data: `customer_email`, so we should be able to calculate the hash on the server side when inserting fingerprint into DB. But I'm not sure which approach is better. @dmvrtx ?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -263,6 +263,20 @@ public function create_intention(
...
 
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+			$order               = wc_get_order( $order_id );
+			$billing_email_parts = explode( '@', $order->get_billing_email() );
+			$email_domain        = array_pop( $billing_email_parts );
+
+			$request['metadata']['fraud_prevention_data'] = [
+				'shopper_ip_hash'        => $this->hash_data_for_fraud_prevention( $order->get_customer_ip_address() ),
+				'shopper_useragent_hash' => $this->hash_data_for_fraud_prevention( $order->get_customer_user_agent() ),
```
kalessil, line 274: We should lowercase user agent and email.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -318,6 +332,16 @@ public function update_intention(
 			$request['setup_future_usage'] = 'off_session';
 		}
 
+		if ( $this->is_fraud_prevention_enabled_for_store() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+
+			$request['metadata']['fraud_prevention_data'] = [
+				'payment_intent_id_hash' => $this->hash_data_for_fraud_prevention( $intention_id ),
+				'payment_method_hash'    => '' !== $selected_upe_payment_type ? $this->hash_data_for_fraud_prevention( $selected_upe_payment_type ) : null,
```
kalessil, line 340: I believe we have to hash e.g. 'card:<card-number>' here. It could be also done on the server side, where we could extract payment details from the transaction object. This pushes us toward having the hashing partially on server side.

Let;s keep this approach and when we address the comments, we'll follow up on this.

kalessil (changes requested):
I like the approach, left some comments we need to address.

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: That's a good point, and it also reduces exposure of attributes used. Let's move it to the server side.

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: Actually, related to the comment above about server- vs client-side, I think we can calculate this block on the server-side if needed. Front-end should only provide us with data that are not reliable to retrieve otherwise, e.g. IP, User-Agent, etc.

includes/wc-payment-api/class-wc-payments-api-client.php
elazzabi: I think whenever we can offload something to the server, we should do it to keep the client light

includes/wc-payment-api/class-wc-payments-api-client.php
elazzabi: That makes sense. I'll remove this block

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @package WCPay\Fraud_Prevention
+ */
+
+namespace WCPay\Fraud_Prevention;
+
+/**
+ * Class Buyer_Fingerprinting_Service
+ */
+class Buyer_Fingerprinting_Service {
```
kalessil, line 13: Oh boy, this looks slim and nice. Can we also add a constructor, where we inject `Fraud_Prevention_Service` and turn all methods into non-static, please? On top, I'd suggest using a singleton pattern as `Fraud_Prevention_Service` does.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @return [string] An array of hashed data for an order.
+	 */
+	public static function get_hashed_data_for_order( $order_id ) {
+		$order        = wc_get_order( $order_id );
+		$email_domain = explode( '@', $order->get_billing_email(), 2 )[1];
+
+		return [
+			'shopper_ip_hash'        => self::hash_data_for_fraud_prevention( $order->get_customer_ip_address() ),
+			'shopper_useragent_hash' => self::hash_data_for_fraud_prevention( strtolower( $order->get_customer_user_agent() ) ),
+			'shopper_email_hash'     => self::hash_data_for_fraud_prevention( strtolower( $order->get_billing_email() ) ),
```
kalessil, line 48: `shopper_email_hash`, `shopper_email_domain` can be also picked up on server side.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,52 @@
...
+	public static function is_fraud_prevention_enabled_for_store() {
+		return Fraud_Prevention_Service::get_instance()->is_enabled();
+	}
+
+	/**
+	 * Returns fraud prevention data for an order.
+	 *
+	 * @param string $order_id The WC order id.
+	 *
+	 * @return [string] An array of hashed data for an order.
```
kalessil, line 39: `[string]` -> `string[]`

includes/fraud-prevention/class-buyer-fingerprinting-service.php
kalessil: And we need tests for the class.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -264,6 +265,11 @@ public function create_intention(
 		$request['payment_method_types'] = $payment_methods;
 		$request['capture_method']       = $capture_method;
 
+		if ( Buyer_Fingerprinting_Service::is_fraud_prevention_enabled_for_store() ) {
```
kalessil, line 268: We'd need a new test case for `create_intention` which ensures the metadata injected when fraup protection is enabled.

kalessil (approved):
Sweet! Next feedback round =)

includes/fraud-prevention/class-buyer-fingerprinting-service.php
elazzabi: The `is_fraud_prevention_enabled_for_store` function was only a wrapper around the `Fraud_Prevention_Service::get_instance()->is_enabled()` so I believe we can use the later directly instead.

When using it, the class now is independent and doesn't need a `Fraud_Prevention_Service` part of the constructor. I transformed it to a singleton, but I think it's only a group of helper functions and we don't need it to be a singleton?

elazzabi: After a talk on Slack p1649330319655879-slack-C02C00HCJ8J, I think we can keep it :)

elazzabi:
@kalessil / @dmvrtx this should be ready for another pair of eyes 👀 

Thank you 🙇

changelog/add-data-fingerprinting
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Add data fingerprinting for fraud prevention part of intent creation/update
```
kalessil, line 4: fraud -> card testing

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	private static $instance;
+
+	/**
+	 * Fraud prevention service instance.
+	 *
+	 * @var Fraud_Prevention_Service
+	 */
+	private $fraud_prevention_service;
+
```
kalessil, line 27: Extra empty line can be dropped.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Hashes customer data for the fraud prevention.
+	 *
+	 * @param string $data The data you want to hash.
+	 *
+	 * @return string Hashed data.
+	 */
+	public function hash_data_for_fraud_prevention( $data ) {
```
kalessil, line 69: We should make the APIs strictly typed: 'public function hash_data_for_fraud_prevention( string $data ): string'

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Returns fraud prevention data for an order.
+	 *
+	 * @param string $order_id The WC order id.
+	 *
+	 * @return string[] An array of hashed data for an order.
+	 */
+	public function get_hashed_data_for_order( $order_id ) {
```
kalessil, line 80: public function get_hashed_data_for_order( int $order_id ): string

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -103,6 +104,68 @@ public function test_create_intention_success() {
...
+
+		$result = $this->payments_api_client->create_intention(
+			$expected_amount,
+			'usd',
+			'pm_123456789',
+			1
+		);
+
+		$this->assertEquals( $expected_amount, $result->get_amount() );
+		$this->assertEquals( $expected_status, $result->get_status() );
```
kalessil, line 166: Can we use here and on the previous line 'assertSame'?

kalessil (approved):
LGTM! Left some suggestions on how to make the PR even better.

changelog/add-data-fingerprinting
kalessil: ` creation/update` can be skipped

includes/fraud-prevention/class-buyer-fingerprinting-service.php
elazzabi: I think the return type here is `array`, so I used that instead of `string`

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,87 @@
...
+
+	/**
+	 * Hashes customer data for the fraud prevention.
+	 *
+	 * @param string $data The data you want to hash.
+	 *
+	 * @return string Hashed data.
+	 */
+	public function hash_data_for_fraud_prevention( string $data ): string {
+		return hash( 'sha512', $data );
```
dmvrtx, line 69: I'd suggest to explicitly set string output by adding third argument set to `false`. Method itself can be defined as a static method.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @var Buyer_Fingerprinting_Service
+	 */
+	private static $instance;
+
+	/**
+	 * Fraud prevention service instance.
+	 *
+	 * @var Fraud_Prevention_Service
+	 */
+	private $fraud_prevention_service;
```
dmvrtx, line 26: What is the reason for having `Fraud_Prevention_Service` here? I don't see any usage.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * Returns fraud prevention data for an order.
+	 *
+	 * @param int $order_id The WC order id.
+	 *
+	 * @return string[] An array of hashed data for an order.
+	 */
+	public function get_hashed_data_for_order( int $order_id ): array {
```
dmvrtx, line 79: It appears to be a candidate for a static method, removing the need for singleton.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,87 @@
...
+		$order = wc_get_order( $order_id );
+
+		return [
+			'shopper_ip_hash'        => self::hash_data_for_fraud_prevention( $order->get_customer_ip_address() ),
```
dmvrtx, line 83: `hash_data_for_fraud_prevention` is called as static, but is not defined as such.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,87 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Sets a instance to be used in request cycle.
+	 * Introduced primarily for supporting unit tests.
+	 *
+	 * @param Buyer_Fingerprinting_Service|null $instance Instance of self.
+	 */
+	public static function set_instance( self $instance = null ) {
```
dmvrtx, line 57: This service exposes two idempotent public methods. Is there an actual need to override it for testing?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -265,6 +266,11 @@ public function create_intention(
 		$request['payment_method_types'] = $payment_methods;
 		$request['capture_method']       = $capture_method;
 
+		if ( Fraud_Prevention_Service::get_instance()->is_enabled() ) {
+			$request['metadata']['fraud_prevention_data_available'] = true;
+			$request['metadata']['fraud_prevention_data']           = Buyer_Fingerprinting_Service::get_instance()->hash_data_for_fraud_prevention( $order_id );
```
dmvrtx, line 271: Should method here be `get_hashed_data_for_order`?

tests/unit/fraud-prevention/test-class-buyer-fingerprinting-service.php
```diff
@@ -0,0 +1,56 @@
...
+
+		$this->fraud_prevention_service_mock = $this->createMock( Fraud_Prevention_Service::class );
+
+		Buyer_Fingerprinting_Service::set_instance( null );
+		$this->buyer_fingerprinting_service = Buyer_Fingerprinting_Service::get_instance( $this->fraud_prevention_service_mock );
+	}
+
+	public function test_it_hashes_using_sha512() {
+		$result                 = $this->buyer_fingerprinting_service->hash_data_for_fraud_prevention( 'test_string' );
+		$expected_hashed_string = hash( 'sha512', 'test_string' );
```
dmvrtx, line 41: It would be also good to add third argument `false` here and also to test `$this->assertTrue( is_string( $result ) );`

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -103,6 +104,68 @@ public function test_create_intention_success() {
...
+			->method( 'is_enabled' )
+			->with()
+			->willReturn( true );
+
+		$mock_fingerprinting
+			->expects( $this->exactly( 1 ) )
+			->method( 'hash_data_for_fraud_prevention' )
+			->with( 1 );
+
+		$this->set_http_mock_response(
```
dmvrtx, line 134: It appears that this mock response and `assertSame` part below is irrelevant to the purpose of the test. Calling `create_intent` should be enough to verify that fingerprinting service has been called.

dmvrtx (commented):
Thanks for a meticulous work @elazzabi. I've been a bit late with my review, but I think it is worth checking comment in `includes/wc-payment-api/class-wc-payments-api-client.php` .

includes/fraud-prevention/class-buyer-fingerprinting-service.php
kalessil: That was my request, so it's aligned with other fraud services we introduced previously. I suggest keeping the current approach but discussing on our team chat which approaches to use in the future. @dmvrtx, can you add an item agenda for the team chat, please?

elazzabi:
Thank you @kalessil and @dmvrtx for your feedback, I worked on the feedback in this PR https://github.com/Automattic/woocommerce-payments/pull/4120

kalessil:
@elazzabi, out of curiosity, why in a separate PR?

elazzabi:
@kalessil just because this one is merged. I think I can't re-open a PR once it's merged? Or is there another way I don't know of 🤔

kalessil:
I see @elazzabi. All good, thanks for pointing out =)

PR #4054, opened 2022-03-30
Tweak onboarding select control

### LINKED ISSUE #4027: KYC Optimisation: Add default states to the select boxes
### Description

Follow on from #3792.

We should add default placeholders to the selects, and have the states get reset when the form is updated. For example, if a new country is selected the selected type/structure values may be invalid.

### Acceptance criteria

* The form matches the Figma designs (see #3792), and the states are reset appropriately when a value is changed.

### LINKED ISSUE #4028: KYC Optimisation: Improve the styling of the select dropdowns
### Description

Follow on from #3792 

Improve the way the CustomSelectControls display. Because we use an outdated version of `@wordpress/components`, the way it displays puts the tick icon _before_ the text. We tried to get around this using `display: flex; flex-direction: reverse`, but this also reverses all other (unselected) elements in the select, making them display right-to-left.

One thing we could consider is to fix this in our `OnboardingSelectControl` component, make the description field optional, and use this component throughout our onboarding. 

Screenshot 2022-03-24 at 13 34 21

ismaeldcom:
Fixes #4028, Fixes #4027

- Adjust `LoadableBlock` to make it more similar to the form.
- Move country description to where it belongs.
- Update `OnboardingSelectControl`. Rename it to `CustomSelectControl` and make it usable anywhere by allowing the use of a custom component to render options, and add a placeholder. Plus some minor style changes.
- Update add business task to use the new `CustomSelectControl`, adding placeholders and a custom option component for the business type select. And some minor style improvements to meet the Figma design.
- Optimize first `businessCountry` assignment, to avoid calling `.find()` after loading `businessTypes`.

#### Testing instructions
Tests should pass, and the implemented design should match the Figma design.
__Note__: I've chosen to keep the select control gap to 12px rather than 16px to keep it consistent with other WC Pay and WC Admin controls.

#### Screenshots
| Before | After |
|-|-|
|Screenshot 2022-03-30 at 13 06 45|Screenshot 2022-03-30 at 13 04 56|
|Screenshot 2022-03-30 at 13 07 04|Screenshot 2022-03-30 at 13 05 27|
|Screenshot 2022-03-30 at 13 07 14|Screenshot 2022-03-30 at 13 05 46|

-------------------

**Post merge**
QA Testing Not Applicable


dmallory42 (approved):
LGTM, matches the designs and works really well. I like the changes you made to the select component!

One small thing I noticed which is unrelated to the changes but seems relevant. It looks like the country select is currently sorted by country code - would it make more sense to sort by country name? For example, United Arab Emirates (AE) is at the top of the list, United Kingdom (GB) is somewhere in the middle, and United States (US) is at the end.

ismaeldcom:
Thanks @dmallory42! And good catch with the list of countries. Although it's unrelated, I added it in 64dd6f6, as it's only one line 🙂

PR #4061, opened 2022-03-31
Add handling for server export for disputes CSV download

### LINKED ISSUE #3846: Extend disputes CSV export
### Description
Similar to transactions export implemented in #1484 implement full disputes export.
Use email and add notice as described in https://github.com/Automattic/woocommerce-payments/issues/3751#issuecomment-1039536778

### Dev notes

Extra server work might be necessary as well.

### Additional context

p1645630566831739/1643675501.909209-slack-C7U3Y3VMY

deepakpathania:
Fixes #3846 

Trigger request to the server when requesting a download of disputes table while using filters. Otherwise, only download disputes presented in the table view.

#### Testing instructions

Note this PR relies on this server PR: 1794-gh-Automattic/woocommerce-payments-server

- Via WP-Admin go to Payments > Disputes
- Click the download link.
- Ensure a CSV containing the disputes listed in the table view immediately downloads
- Select any filter
- Ensure a notice appears letting you know an export will be emailed to you
- Confirm you receive an email containing a download link for an export containing disputes that meet the criteria (look at server PR's testing instructions for reference)
- Repeat steps 3-6 with different combinations of filters

-------------------

**Post merge**

### DIFF
--- a/client/disputes/index.tsx
+++ b/client/disputes/index.tsx
@@ -39,7 +39,6 @@ import { DisputesTableHeader } from 'wcpay/types/disputes';
 import { getDisputesCSV } from 'wcpay/data/disputes/resolvers';
 
 import './style.scss';
-import { title } from 'process';
 
 const getHeaders = ( sortColumn?: string ): DisputesTableHeader[] => [
 	{
@@ -250,6 +249,7 @@ export const DisputesList = (): JSX.Element => {
 
 	const onDownload = async () => {
 		setIsDownloading( true );
+		const title = __( 'Disputes', 'woocommerce-payments' );
 		const downloadType = totalRows > rows.length ? 'endpoint' : 'browser';
 		const userEmail = wcpaySettings.currentUserEmail;
 
@@ -259,18 +259,16 @@ export const DisputesList = (): JSX.Element => {
 				date_after: dateAfter,
 				date_between: dateBetween,
 				match,
-				search,
-				type_is: typeIs,
-				type_is_not: typeIsNot,
+				status_is: statusIs,
+				status_is_not: statusIsNot,
 			} = getQuery();
 
 			const isFiltered =
 				!! dateBefore ||
 				!! dateAfter ||
 				!! dateBetween ||
-				!! typeIs ||
-				!! typeIsNot ||
-				!! search;
+				!! statusIs ||
+				!! statusIsNot;
 
 			const confirmThreshold = 1000;
 			const confirmMessage = sprintf(
@@ -296,9 +294,8 @@ export const DisputesList = (): JSX.Element => {
 							dateBefore,
 							dateBetween,
 							match,
-							search,
-							typeIs,
-							typeIsNot,
+							statusIs,
+							statusIsNot,
 						} ),
 						method: 'POST',
 					} );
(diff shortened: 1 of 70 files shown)


client/disputes/index.tsx
```diff
@@ -241,58 +248,137 @@ export const DisputesList = (): JSX.Element => {
...
+
+					createNotice(
+						'success',
+						__(
+							'Your export will be emailed to you.',
+							'woocommerce-payments'
+						)
+					);
+
+					wcpayTracks.recordEvent( 'wcpay_transactions_download', {
```
jbordonado, line 312: Should it be `disputes` instead of `transactions`? Same goes for the fields below

Similar to l.374

includes/admin/class-wc-rest-payments-disputes-controller.php
```diff
@@ -135,6 +144,18 @@ public function close_dispute( $request ) {
 		return $this->forward_request( 'close_dispute', [ $dispute_id ] );
 	}
 
+	/**
+	 * Initiate disputes export via API.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 */
+	public function get_disputes_export( $request ) {
+		$user_email = $request->get_param( 'user_email' );
```
jbordonado, line 153: I don't see where we add this parameter, is this set somewhere? Otherwise, it looks like it will always fallback to the value of the blog option `admin_email` in the server handling.

client/disputes/index.tsx
deepakpathania: Good catch, fixed this.

includes/admin/class-wc-rest-payments-disputes-controller.php
deepakpathania: Started sending it in params now, missed it earlier.

client/disputes/index.tsx
```diff
@@ -241,58 +248,143 @@ export const DisputesList = (): JSX.Element => {
...
 
-		const csvColumns = [
-			{
-				...headers[ 0 ],
-				label: __( 'Dispute Id', 'woocommerce-payments' ),
-			},
-			...headers.slice( 1 ),
-		];
+		if ( 'endpoint' === downloadType ) {
+			const {
```
jbordonado, line 257: I think `type_is` should be `status_is` and `type_is_not` should be `status_is_not`.

As for the `search` filter, I'm not sure it is available on the disputes page? Couldn't find how to trigger it at least, maybe you know.. But I think we'll be able to remove this one as I can't see it being used in `getDisputesCSV` > `formatQueryFilters` anyway

jbordonado (changes requested):
Left more comments.

FYI, I also noticed a very slight formatting difference on the amount so I thought I'd mention it anyway: client export shows amount = 16 while server export shows amount = 16.00

Otherwise, testing went well in combination with the server PR when I didn't try to use filters. I was able to export CSV files as expected, with a notice and a link to download it in the email, and the formatting was 99% identical to the client export.

There's some work to do on the filters though, as the status is not taken into account for now

jbordonado (approved):
I was able to download an export from the server with some filters as expected.
Everything seems to work fine on my end.

Well done 🥳

PR #4066, opened 2022-03-31
Prevent platform checkout iframe appear when go back

### LINKED ISSUE #4065: Prevent platform checkout iframe appear when go back
### Describe the bug

Currently when you access platform checkout and go back to merchant store it shows the platform checkout again, it would be good to implement a query parameter or a way to detect if user came from it, preventing it to show again.

### Acceptance criteria

* When you access platform checkout and go back, its iframe should not be shown again until the user change their email address.
* If the email is pre-filled, it should show the platform checkout iframe unless there's a query parameter that means that the user came from the platform checkout.

alefesouza:
Fixes #4065

* Fix a bug that always show the platform checkout when go back from it.

#### Testing instructions

* Enable platform checkout.
* Go to checkout page and fill out you email address.
* Once redirected, go back with browser button.
* It should not show the platform checkout iframe.
* Remove a letter from your email address and add it again.
* Once redirected, click the back button on platform checkout.
* It should not show the platform checkout iframe.

-------------------

**Post merge**


client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -244,10 +244,26 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
+		const historyTraversal =
+			event.persisted ||
+			( 'undefined' !== typeof performance &&
+				'back_forward' ===
+					performance.getEntriesByType( 'navigation' )[ 0 ].type );
+
+		const searchParams = new URLSearchParams( window.location.search );
+
+		if (
+			! historyTraversal &&
```
a-danae, line 259: When coming back through the browser's back button, I still see the SMS OTP modal open with the loading state. Which is pretty much the last thing displayed before getting redirected. Do you get that on your end as well?

I think that'd make sense because the page is loaded directly from the bfcache - instead of loading it normally and invoking all of the JS flow for rendering it - so I don't think this approach will solve that issue.

Because we're changing `javascript:history.back()` to the actual merchant's site URL and using history only as a fallback, we could reduce the scope of the issue we're addressing here to cover the behavior of the buttons in the UI only, and have another issue opened for handling the "back" navigation via history. Thinking that skipping the bfcache for this case might be tricky and doesn't seem to have a high priority: you can always close the modal with the X button and the current behavior doesn't seem to affect the UX, even less when the behaviors of the buttons of the UI are addressed. My 2c really. Glad to ping the ones in the conversation about the original issue to chime in.

a-danae (changes requested):
Hey! Thanks for working on this, @alefesouza. I can confirm that the modal doesn't get opened when coming back from the platform checkout using the URLs with the special param. I left a comment about a behavior I noticed and an idea if you can also replicate it.

client/checkout/platform-checkout/email-input-iframe.js
alefesouza: Got it, I've noticed that the browser back button event works in Chrome but do not work in Safari, removed it on c255b8b, I'll investigate it better and open another PR later on.

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -244,8 +244,15 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
 	};
 
-	// Check the initial value of the email input and trigger input validation.
-	if ( validateEmail( platformCheckoutEmailInput.value ) ) {
+	// Prevent show platform checkout iframe if the page comes from
+	// the back button on platform checkout itself.
+	const searchParams = new URLSearchParams( window.location.search );
+
+	if (
+		'true' !== searchParams.get( 'skip_platform_checkout' ) &&
```
asumaran, line 252: Not really necessary but I think it'd be nice to remove the param from the URL (using `history.replaceState()`) if the `skip_platform_checkout` param is found.

asumaran (approved):
Code looks good. I also found the same odd behavior where the user uses the browser's "Back" button (The iframe dialog is still displayed).

I'm just leaving a non-blocking comment.

a-danae (approved):
Thanks for the update here @alefesouza ! LGTM.

client/checkout/platform-checkout/email-input-iframe.js
alefesouza: Good suggestion, added on c74f0343.

asumaran (approved):
LGTM! 🚀

PR #4069, opened 2022-04-01
Add a namespace to Session_Rate_Limiter

marcinbot:
Noticed that `Session_Rate_Limiter` has no namespace and isn't prefixed with `WC_Payments_`. If any other plugin on a site declares a class with a similar name, it could lead to some nasty errors.

Add a `WCPay` namespace to the class, update the files that use it.

#### Testing instructions

* Process a payment - it should succeed
* Unit tests should pass
* Verify that all files that use `Session_Rate_Limiter` now import it at the top of the file as `WCPay\Session_Rate_Limiter`

-------------------

### DIFF
--- a/changelog/fix-rate-limiter-namespace
+++ b/changelog/fix-rate-limiter-namespace
@@ -1,0 +1,5 @@
+Significance: patch
+Type: dev
+Comment: Minor code cleanup
+
+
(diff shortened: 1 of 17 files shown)


dmallory42 (approved):
LGTM, tested and all working as expected. :shipit:

PR #4074, opened 2022-04-01
Track document downloaded events

luizreis:
Fixes 1791-gh-Automattic/woocommerce-payments-server

This PR uses `wc_admin_record_tracks_event` to register document downloaded events, along with the ID for the requested document and gateway mode.

The change added doesn't use the existing `Tracker` or `WC_Tracks` classes because WooCommerce core only includes Tracks in admin, not the REST API, so it'd be better to use the `wc_admin_record_tracks_event` method that includes `WC_Tracks` in case it's not loaded.

#### Testing instructions

* Add a test document to your account if it doesn't have one yet
* Navigate to the Documents section and click on the "Download" link for a document
* Open the Tracks Live View, filter for `wcadmin_wcpay_document_downloaded`, and make sure an event was registered with the document id and gateway mode
    * _It might take a while for the event to show up in the live view_

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _QA Testing Not Applicable_

### DIFF
--- a/changelog/add-1791-documents-usage-tracking
+++ b/changelog/add-1791-documents-usage-tracking
@@ -1,0 +1,5 @@
+Significance: patch
+Type: add
+Comment: These changes are behind the documents feature flag and will be reported when the main feature is launched
+
+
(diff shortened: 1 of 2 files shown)


jessepearson (approved):
LGTM, and it works as described:

```
mode:
test
products_count:
44
wc_version:
6.4.0
document_id:
test_document_IQHe1Nn4xO6JGo9ed_JQ9A
url:
https://wcpaydev.test
```

PR #4076, opened 2022-04-01
Checkout page style tweaks

harriswong:
Fixes platform issue 640

Address all the following style tweaks:

- [x] "Remember my details" headline should match the h3 styles from the rest of the checkout
- [x] Form input styles should be inherited from the theme (checkbox and mobile phone number entry field)
- [x] Body copy in the expanded section should be 0.875 times the size of the body copy
- [x] Body copy and icons should inherit the body copy colors from the checkout
- [x] Ensure vertical spacing between elements is consistent and of reasonable proportions across a variety of themes

The credit for this work is from @brettshumaker. I take his `.diff` file and `git apply` to latest `develop`.

#### Testing instructions
1. Use this branch. Checkout an item.
2. Go to http://localhost:8082/checkout/ and make sure all the bullets above are addressed.
3. Click "Use a new payment method" 
4. Click "Remember your details", enter phone number
5. Proceed to checkout.

-------------------

**Post merge**


client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -19,7 +18,8 @@ const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
 	const [ inputInstance, setInputInstance ] = useState( null );
 	const [ isValid, setIsValid ] = useState( true );
 
-	const handlePhoneNumberInputChange = ( value = inputValue ) => {
+	const handlePhoneNumberInputChange = ( e ) => {
```
harriswong, line 21: I updated this, otherwise it shows something like `[Object] [object]` in the phone field.

harriswong (commented):
Tested with another theme against the bullet items. 👍  But I notice the checkbox is now looking strange. 
image

client/components/platform-checkout/save-user/phone-number-input.js
brettshumaker: Thanks! I forgot to mention that was broken in the diff I sent you. 🙏

client/components/platform-checkout/save-user/test/checkout-page-save-user.test.js
```diff
@@ -19,6 +19,10 @@ jest.mock( 'utils/checkout', () => ( {
 	getConfig: jest.fn(),
 } ) );
 
+// Mock the SVG icons.
+jest.mock( '../../icons/phone-icon', () => () => null );
+jest.mock( '../../icons/lock-icon', () => () => null );
```
frosso, line 24: Hehehe a big middle finger to those icons, I see ;)

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -40,28 +38,38 @@ const CheckoutPageSaveUser = () => {
...
-				value="true"
-			/>
+			<span>
+				<label htmlFor="save_user_in_platform_checkout">
+					<input
+						type="checkbox"
+						checked={ isSaveDetailsChecked }
+						onChange={ () => {
+							setIsSaveDetailsChecked( ! isSaveDetailsChecked );
+						} }
```
frosso, line 48: (minor) on a `useState` in these kinds of scenarios where the state would need to be toggled, a more appropriate way to update state would be via a functional update:
```
setIsSaveDetailsChecked( v => !v );
```

This is because _technically_, multiple state updates can be queued and processed in a batch. When they are processed, you want them to be processed in their correct order and with the correct state value. More on this: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

frosso (approved):
I tested a few scenarios locally, the changes test well!

client/components/platform-checkout/save-user/checkout-page-save-user.js
brettshumaker: After a brief misunderstanding - fixed in [a99015c](https://github.com/Automattic/woocommerce-payments/pull/4076/commits/a99015cd1a08dc9c5eae1b28faff8df8ba4de7d8)

PR #4078, opened 2022-04-04
Fix onboarding form country getting reseted

ismaeldcom:
Check if `businessCountry` is not empty before applying the account settings country, to avoid resetting user selected one after context update.

#### Testing instructions
- Select any country other than the default selected one
- Select **Individual** as **Business type**
- Should keep the selected country and not revert to the initial one.

-------------------

**Post merge**

- [x] QA Testing Not Applicable


dmallory42 (approved):
LGTM and works well! :shipit:

PR #4085, opened 2022-04-04
Payment Request Button - Do not set WC session cookie and HTML container on the product page if not enabled

### LINKED ISSUE #4084: Payment Request Button - do not set the session cookie and HTML container on the product page if not enabled
See pd3Vha-37-p2#comment-498

When the Payment Request button is enabled, but set not to be visible on the product page:
Screenshot 2022-04-04 at 17 53 20

The [session cookie](https://github.com/Automattic/woocommerce-payments/blob/2751a8380866ae101e944924f881cac0b8032d95/includes/class-wc-payments-payment-request-button-handler.php#L157) and [the html containers](https://github.com/Automattic/woocommerce-payments/blob/2751a8380866ae101e944924f881cac0b8032d95/includes/class-wc-payments-payment-request-button-handler.php#L74-L75) should not be set/rendered.

ricardo:
Fixes #4084 

TL;DR: The VIP platform uses Varnish HTTP cache with cookie exceptions, such as `wp_woocommerce_session_*`. [Src](https://docs.wpvip.com/technical-references/caching/cookies/#:~:text=A%20WooCommerce%20session%20cookie).

That prevents product pages from being cached, even though express checkouts might be disabled for product pages.

- Do not set session cookie if product page is not enabled under settings.
- Check if HTML wrapper and separator should be displayed, [similar to Stripe extension](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/a8e04bcb4fd30d2a69b92067e10b4ace13849f0d/includes/payment-methods/class-wc-stripe-payment-request.php#L809) - this was probably ignored when customizable button locations was introduced.

#### Testing instructions

With Google Pay configured in your browser:

1. Make sure your site uses HTTPS.
2. As a merchant, make sure "Express checkouts" is enabled under the WCPay settings.
3. Under **"Express checkouts > Apple Pay / Google Pay"**, click "Customize".
4. Uncheck "Product page" and click "Save changes".
5. As a guest, navigate to a product page and remove all cookies (you can use a browser extension such as [EditThisCookie](https://chrome.google.com/webstore/detail/editthiscookie/fngmhnnpilhplaeedifhccceomclgfbg?hl=pt-BR), or **Chrome DevTools > Application > Cookie**).
6. Refresh the product page, and ensure the `wp_woocommerce_session_*` cookie is not created.
7. As a merchant, enable "Apple Pay / Google Pay" on the product page.
8. As a guest, navigate to a product page and ensure the `wp_woocommerce_session_*` cookie was created.
9. Proceed with Google Pay checkout and make sure it works.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -150,8 +150,10 @@ class WC_Payments_Payment_Request_Button_Handler {
 	 * @return void
 	 */
 	public function set_session() {
+		// Don't set session cookies on product pages to allow for caching when payment request
+		// buttons are disabled. But keep cookies if there is already an active WC session in place.
 		if (
-			! ( $this->is_product() && $this->is_available_at( 'product' ) )
+			! ( $this->is_product() && $this->should_show_payment_request_button() )
 			|| ( isset( WC()->session ) && WC()->session->has_session() )
 		) {
 			return;
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -150,7 +150,10 @@ class WC_Payments_Payment_Request_Button_Handler {
 	 * @return void
 	 */
 	public function set_session() {
-		if ( ! $this->is_product() || ( isset( WC()->session ) && WC()->session->has_session() ) ) {
+		if (
+			! ( $this->is_product() && $this->is_available_at( 'product' ) )
+			|| ( isset( WC()->session ) && WC()->session->has_session() )
+		) {
 			return;
 		}
 
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -662,6 +662,9 @@ class WC_Payments_Payment_Request_Button_Handler {
 	 * Display the payment request button.
 	 */
 	public function display_payment_request_button_html() {
+		if ( ! $this->should_show_payment_request_button() ) {
+			return;
+		}
 		?>
 		<div id="wcpay-payment-request-wrapper" style="clear:both;padding-top:1.5em;display:none;">
 			<div id="wcpay-payment-request-button">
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -150,7 +150,10 @@ public function get_total_label() {
 	 * @return void
 	 */
 	public function set_session() {
-		if ( ! $this->is_product() || ( isset( WC()->session ) && WC()->session->has_session() ) ) {
+		if (
+			! ( $this->is_product() && $this->is_available_at( 'product' ) )
```
vbelolapotkov, line 154: @ricardo I've noticed there is another condition for product page in `should_show_payment_request_button` method [here](https://github.com/Automattic/woocommerce-payments/blob/d48df11bcc2c50ce931a699371390eb7ef7a2cac/includes/class-wc-payments-payment-request-button-handler.php#L431-L435), should we take that into account as well? So this line would look like this:

```
! ( $this->is_product() && $this->should_show_payment_request_button() )
```

IMO that would keep all the logic in the same place.

vbelolapotkov: Another thing worth doing here is to add a comment explaining why there is such a special treatment for product page. Something like: `Don't set sessions cookies on product pages to allow caching when payment request buttons are disabled. But keep cookies if there is already active WC session in place.`

vbelolapotkov (commented):
I haven't tested the changes, but the code looks good to me (see a couple of comments). I'll ask someone else to run tests.

includes/class-wc-payments-payment-request-button-handler.php
ricardo: Good call. Makes sense to use `should_show_payment_request_button`.

Made the change in [8ff6ba9](https://github.com/Automattic/woocommerce-payments/pull/4085/commits/8ff6ba97e2cc006684775da76599718bab2c7d3a).

vbelolapotkov (approved):
Approving since the code looks good to me, but I haven't tested. So I'd expect someone else to test it properly and approve before merging.

cesarcosta99 (approved):
LGTM and tests well! :shipit:

PR #4088, opened 2022-04-05
Send Store API URL to platform site

Mayisha:
Fixes platform checkout issue 666

Retrieving the Store API namespace from the blocks of woocommerce core and sending this to the platform with other store data.

#### Testing instructions
- Install different versions of `woocommerce-blocks` on the merchant site and verify the URL on platform site
- From `woocommerce-blocks` >= 7.1.0 the value of `store_api_url` should be `wc/store/v1`
- For older versions of  `woocommerce-blocks` the value of `store_api_url` should be `wc/store`


includes/class-wc-payments.php
```diff
@@ -891,6 +893,7 @@ public static function ajax_init_platform_checkout() {
 
 		$platform_checkout_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'http://host.docker.internal:8090';
 		$url                    = $platform_checkout_host . '/wp-json/platform-checkout/v1/init';
+		$store_api_url          = Package::container()->get( RoutesController::class )->get( 'cart' )->get_namespace();
```
waclawjacek, line 896: I recall working on [this code](https://github.com/Automattic/woocommerce-payments/blob/abbe046497c841633325d7b8dc57b6f4c75493b2/tests/unit/admin/test-class-wc-rest-payments-settings-controller.php#L523) where depending on the version of WC Blocks installed, the availability of `Package::container()` (or its methods) would differ. We should also account for that here.

Maybe try installing versions of WC Blocks earlier than 2.6.0 and checking if this PR works?

Once we get that ironed out, we should also make sure that even if `Package::container()` is present:
* is `Package::container()->get( RoutesController::class )` present?
* is `Package::container()->get( RoutesController::class )->get( 'cart' )` present?
* is `Package::container()->get( RoutesController::class )->get( 'cart' )->get_namespace()` present?

waclawjacek (changes requested):
Fantastic solution! Clicking Request changes because of compatibility with WC Blocks < 2.6.0.

includes/class-wc-payments.php
```diff
@@ -906,7 +908,13 @@ public static function ajax_init_platform_checkout() {
 
 		$platform_checkout_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'http://host.docker.internal:8090';
 		$url                    = $platform_checkout_host . '/wp-json/platform-checkout/v1/init';
-		$body                   = [
+
+		if ( class_exists( 'RoutesController' ) ) {
```
Mayisha, line 912: @waclawjacek I have tested with some versions and added these checks as per your feedback.
The current minimum version of `WooCommerce` is `5.0.0` which has the `Package` class but not `RoutesController`. So I added these checks and used 'wc/store' as the fallback here :eyes:

Mayisha: Actually found an issue with this check. Fixing that now. :eyes:

Mayisha: Even if the `RoutesController` class exists, in older versions it may not be registered in the container which throws an error. Added a try-catch [here](https://github.com/Automattic/woocommerce-payments/pull/4088/commits/ef4a7cda9c8907bb5f4966fb6a94556115aa13e1) to safely handle this scenario.

waclawjacek (approved):
Works great on the random version of WC Blocks `2.5.0 – 2019-11-19`, no Blocks installed (= using the version of Blocks bundled with WC 6.3.1), and WC Blocks `7.3.0` and it worked brilliantly.

Great work, @Mayisha! :tada: :100:

PR #4096, opened 2022-04-06
New onboarding flow refactor

ismaeldcom:
- Migrate all code to TypeScript.
- Update `CustomSelectControl` to handle `undefined` values and use a generic item type.
- Implement `useApiFetch` hook.
- Replace the use of `wp/data` with `useApiFetch`.
- Simplify types, and try to reuse them across different files.

The main change here is to simplify API requests and get rid of `wp/data` if it's not needed, using a new hook called `useApiFetch` that provides a similar functionality.

Sometimes we only need to fetch some data, and show it on a component, without storing it in the store. For this use case, dealing with the store and creating a reducer, actions, resolvers and selectors, can overcomplicate the implementation and make it harder to implement and understand.

`useApiFetch` hook uses `@wordpress/api-fetch` under the hood and returns the data from an API response and the loading state. With cache based on `path` value and support for showing error notices. It's a quite simple implementation that can be extended to support more cases.

**Example:**
```ts
const { data, isLoading } = useApiFetch< DataType >( {
    path: '/some/url',
    errorMessage: 'Error retrieving some data'
} );
```

#### Testing instructions
There is no functional changes, so we should unsure that it keeps working as before.

-------------------

**Post merge**
- QA Testing Not Applicable


client/onboarding/tasks/add-business-info-task/test/index.tsx
```diff
@@ -1,35 +1,45 @@
...
-import React from 'react';
+import { mocked } from 'ts-jest/utils';
 
 /**
  * Internal dependencies
  */
 import AddBusinessInfoTask from '../';
-import { useBusinessTypes } from 'wcpay/data/onboarding';
+import { useBusinessTypes } from 'onboarding/hooks';
+import { BusinessType } from 'onboarding/types';
```
dmallory42, line 14: It looks like this is no longer being used in this file and can be removed!

client/onboarding/tasks/add-business-info-task/test/required-verification-info.tsx
```diff
@@ -1,18 +1,18 @@
...
+import { render } from '@testing-library/react';
+import { mocked } from 'ts-jest/utils';
 
 /**
  * Internal dependencies
  */
 import RequiredVerificationInfo from '../required-verification-info';
-import { useRequiredVerificationInfo } from 'wcpay/data/onboarding';
+import { useRequiredVerificationInfo } from 'onboarding/hooks';
+import { BusinessType } from 'onboarding/types';
```
dmallory42, line 13: Looks like this can be removed here since it is no longer being used.

dmallory42 (approved):
LGTM - thanks for working on this!

I really like the new approach, I find it a lot simpler for the use case where a value needs to be fetched from the API but we don't need to store it in the store. I think both use cases can exist side-by-side as it may be appropriate to use either depending on the need.

I would be happy to merge this in, not sure whether we need any additional eyes on it. From my perspective, the approach suits the need, also benefits the overall codebase and it works well upon testing.

client/onboarding/tasks/add-business-info-task/test/index.tsx
ismaeldcom: Good catch! Addressed in 8bd2f99

PR #4111, opened 2022-04-08
[WIP] Unset receipt_email before capturing an intent

### LINKED ISSUE #4048: Unset receipt_email before capturing an intent
Before our implementation of IPP email receipt is merged (#4014), we should prepare ourselves to tackle the scenario in which a merchant is using the latest plugin version with an old version of the Mobile app (which will send an appropriately set `receipt_email` field). In this case, the plugin will send a Woo-branded email and the app will trigger a Stripe-branded email. This will result in duplicate-ish emails that are likely to confuse a buyer.

One approach is to adjust the `update_intention_metadata` method so to unset the `receipt_email` field while updating the intent:

https://github.com/Automattic/woocommerce-payments/blob/49fbba763016b0bb2d29bc4e495d6880c6fd7256/includes/class-wc-payment-gateway-wcpay.php#L1988-L1991

This may require renaming the method to something that better reflects the fact that both metadata AND receipt email fields will be updated.

mgascam:
Fixes #4048 

This is WIP, please feel free to amend this description as you see fit. 

I think it is going in the good direction. As stated in the description of the issue, we need to update the payment intent to remove the `receipt_email` to prevent [Stripe from sending the email on their end](https://stripe.com/docs/receipts#automatically-send-receipts-when-payments-are-successful). 

There is an issue with testing this approach, since Stripe does not send email receipts in Test mode. However, I can see that after this change, the `receipt_email` is set to null in the payment events in Stripe's Dashboard:

**Before this change**

Screen Shot 2022-04-08 at 18 55 42

**After this change**

Screen Shot 2022-04-08 at 18 56 40

If we trust the Stripe Docs, **there should be no email sent since the `receipt_email` is set to null.** 

**TODOs**

- [ ] Refactor the name of the method to reflect that we are also updating the `receipt_email`
- [ ] Amend / Add unit tests if needed
- [ ] Manual test that emails are not sent?

#### Testing instructions

* You will need to use the mobile emulator or use the capture_charge endpoint to test this

-------------------

**Post merge**

### DIFF
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -338,14 +338,20 @@ class WC_Payments_API_Client {
 	 *
 	 * @param string $intention_id - The ID of the intention to update.
 	 * @param array  $metadata     - Meta data values to be sent along with payment intent creation.
+	 * @param bolean $remove_receipt_email - Whether to remove the receipt_email.
 	 *
 	 * @return WC_Payments_API_Intention
 	 * @throws API_Exception - Exception thrown on intention creation failure.
 	 */
-	public function update_intention_metadata( $intention_id, $metadata ) {
+	public function update_intention_metadata( $intention_id, $metadata, $remove_receipt_email = false ) {
 		$request = [
 			'metadata' => $metadata,
 		];
+
+		if ( $remove_receipt_email ) {
+			// I tried to set this to null but then Stripe did not remove the email.
+			$request['receipt_email'] = '';
+		}
 
 		$response_array = $this->request_with_level3_data( $request, self::INTENTIONS_API . '/' . $intention_id, self::POST );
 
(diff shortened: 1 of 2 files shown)


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -338,15 +338,21 @@ public function update_intention(
...
 	 */
-	public function update_intention_metadata( $intention_id, $metadata ) {
+	public function update_intention_metadata( $intention_id, $metadata, $remove_receipt_email = false ) {
 		$request = [
 			'metadata' => $metadata,
 		];
 
+		if ( $remove_receipt_email ) {
+			// I tried to set this to null but then Stripe did not remove the email.
+			$request['receipt_email'] = '';
```
kalessil, line 353: Can we apply `$request['metadata']['receipt_email']` here and add a comment that it's for BC with older Mobile apps? That would allow us to omit other changes.

kalessil (changes requested):

anu-rock (commented):
The overall direction looks good. Thanks for leaving a detailed summary, testing instructions, and next steps.

I've pushed a few changes that:
* rename the method to better reflect the new changes
* adjust existing unit tests

I tested the changes with the latest mobile app and found everything to be working as expected. However, that's not the ideal test. We want to test it with the combination of latest plugin and old mobile app. That I haven't been able to do because of Jetpack issues. I'll continue my testing tomorrow.

mordeth:
I have tested it, and the `receipt_email` is set to `null`. Great job!

Issue #3729, opened 2022-01-29
Customer (subscriber) should be able to easily update default payment method for all subscriptions


dhjw:
### Describe the bug
Managing payment methods in My Account > Payment Methods is not updating active subscriptions as expected.

### Reproduction / Behavior
1. Set up a subscription using a saved credit card.
2. Save a second credit card using a different expiry date so you can tell them apart.
3. In My Account > Payment Methods, click Make Default until you choose the card not applied to the subscription. Click Yes to update the subscription payment method to the new default. View the subscription details and see that nothing changed - it's still using the original card.
4. In My Account > Payment Methods, delete the card that is currently used for the subscription. A notice says the subscription was automatically updated to the _deleted_ card instead of the remaining card. The subscription then has no payment method applied and just says "Via credit card / debit card" instead of "Visa ending in ..." and renewal will surely fail.

### Expected behavior
When changing the default card and clicking Yes the subscription should change to use the new default.
When deleting the card currently in use the subscription should change to use the remaining card.

### Screenshots

Clicking `Make Default` on the Amex card ending in 0005. Presents the prompt.
![Screen Shot 2022-03-23 at 12 59 15 PM](https://user-images.githubusercontent.com/57298/159597470-55b65ab8-1eeb-4254-b253-049000b26218.png)
Clicking `Yes` does not update the payment method for the subscription
![Screen Shot 2022-03-23 at 12 59 35 PM](https://user-images.githubusercontent.com/57298/159597584-3cb09261-115b-42f3-8d56-0b451c4f4838.png)

haszari:
Labelling as high priority as this is silently failing, and changing payment method is a pretty fundamental subscriber flow.

haszari:
@zmaglica @dhjw – Do you recall if this is affecting WCPay Subscriptions (built in) or WC Subscriptions (extension integration) or both? Please update labels/reproduce steps to clarify. 

Assuming both for now.

dhjw:
My report is based on WCPay Subscriptions. I did not pay for and can't use WC Subscriptions and don't know if it's affected.

haszari:
Thanks for the info @dhjw – and thanks also for reporting 🙌

melek:
Report in 5454898-zen. The merchant is using WC Subscriptions.

alexander7567:
> Report in 5454898-zen. The merchant is using WC Subscriptions.

This is me if you need any more information.

mattallan:
[...]

PR #4119, opened 2022-04-11
Fixed issue with subscription payment switch by adding flag to UPE pa…

### LINKED ISSUE #3771: Cannot add a new payment method to existing subscriptions

zmaglica:
Fixes #3771 

I did a lot of testing, tried different scenarios, and proposed changes in code that @RadoslavGeorgiev suggested, but at the end of the day, the simple fix that is included in this PR solves all issues. I performed various tests and even recorded a whole process using the latest woocommerce/wordpress version with two different block-based themes, and everything worked well on my end.

#### Testing instructions

1. Make sure that you have subscription with added payment method
2. Go to page where you manage subscription and change payment method
3. Enter new credit card (you can use Stripe testing cards) into form where you're supposed to add a new card
4. After submitting form, double check is new credit card added to the existing subscription

*

-------------------

**Post merge**


achyuthajoy (commented):
Hey @zmaglica 👋. I couldn't replicate the bug on latest `develop`. I followed the steps mentioned in #3771. 

Here are the steps I followed:
* Enabled UPE Payment methods
* Created a new simple subscription product and checked out with card.
* Tried to update the payment method via My Account -> Subscriptions using another card.

Instead of the problem, I noticed that the card is added successfully to the subscription and also updated on subscription details in Stripe dashboard. Subscription details on WP-ADMIN confirms that the card is updated to a new one.

Please let me know if I'm missing anything.

brucealdridge (approved):
Thanks @zmaglica !!
Have tested updating subscription payment method with this PR and confirmed that it is working.
Have also tested the flow for updating all subscriptions (see #3729) however this PR does not fix this.

zmaglica:
@brucealdridge I will close this PR since the issue cannot be reproduced on the current develop branch. The issue that you linked, I will address in another PR

PR #4122, opened 2022-04-11
Send CVC confirmation to server for fraud prevention

### LINKED ISSUE #4121: Send CVC confirmation to server for fraud prevention
### Description

We need to send CVC confirmation on Platform Checkout to prevent fraud if the user has a bad rating on the fraud prevention service.

### Acceptance criteria

* A new flag CVC confirmation is sent to the server to process it when the transaction is being created.

### Additional context

pdpOw2-z7-p2

alefesouza:
Fixes #4121

* Send a new CVC confirmation flag to the server to process it when the transaction is being created.

#### Testing instructions

* As we still are implementing it on the Platform Checkout and need it for the WCPay 4.0 launch, just test the normal user flow (add something to cart and place order), everything should work normally.

-------------------

**Post merge**


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -183,6 +183,7 @@ public function create_charge( $amount, $source_id ) {
 	 * @param bool   $off_session                     - Whether the payment is off-session (merchant-initiated), or on-session (customer-initiated).
 	 * @param array  $additional_parameters           - An array of any additional request parameters, particularly for additional payment methods.
 	 * @param array  $payment_methods                 - An array of payment methods that might be used for the payment.
+	 * @param string $cvc_confirmation The CVC confirmation for this payment method.
```
brianyu28, line 190: Nit: This line is misaligned with the other parameter descriptions.

includes/class-payment-information.php
```diff
@@ -264,6 +275,26 @@ public static function get_token_from_request( array $request ) {
...
+	/**
+	 * Extract the payment CVC confirmation from the provided request.
+	 *
+	 * @param array $request Associative array containing payment request information.
+	 *
+	 * @return string|NULL
+	 */
+	public static function get_cvc_confirmation_from_request( array $request ) {
+		$payment_method    = $request['payment_method'] ?? null;
+		$token_request_key = 'wc-' . $payment_method . '-payment-cvc-confirmation';
```
brianyu28, line 287: In the event that `$payment_method` is `null`, do we need to still generate the `$token_request_key` here? Or can we just return `null` in that case?

includes/class-payment-information.php
```diff
@@ -264,6 +275,26 @@ public static function get_token_from_request( array $request ) {
...
+		if (
+			! isset( $request[ $token_request_key ] ) ||
```
brianyu28, line 289: Is there supposed to be a `"wc-woocommerce_payments-payment-cvc-confirmation"` key in the request data? I'm not seeing that in the request data, so this is always turning out to be `null` for me. Is there some flow that I should test where this will have a value?

brianyu28 (changes requested):
Thanks @alefesouza! My main question is whether there's supposed to be a value for `wc-woocommerce_payments-payment-cvc-confirmation` yet — right now that key doesn't exist, but maybe that's something you're planning to add on the platform site later?

If that's the expected key and it's supposed to be null for now, then this looks okay, just left a couple small comments.

includes/wc-payment-api/class-wc-payments-api-client.php
alefesouza: Fixed on d8b4cfa.

includes/class-payment-information.php
alefesouza: This method was based on the upper method, but fixed on 8ba6fb8.

includes/class-payment-information.php
alefesouza: I'm working to send this variable on Platform Checkout to validate the CVC number.

brianyu28 (approved):
Looks good, thanks @alefesouza!

PR #4129, opened 2022-04-13
Pass along the platform checkout user session key

cesarcosta99:

This change is needed to allow the plugin to receive a user session key that comes from the OTP authentication. Considering the platform checkout flow requirements, it's also sending the key back during the platform checkout initialization.

See the discussion for more details: p1649281169619379-slack-C022WMN88KG.

#### Testing instructions

Test instructions in the related PR.

-------------------

**Post merge**


client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -285,7 +285,9 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
 				wcpayTracks.recordUserEvent(
 					wcpayTracks.events.PLATFORM_CHECKOUT_OTP_COMPLETE
 				);
-				api.initPlatformCheckout().then( ( response ) => {
+				api.initPlatformCheckout(
+					e.data.platformCheckoutUserSession
+				).then( ( response ) => {
```
lovo-h, line 290: One thing that is not a part of this PR but we may want to handle now, is adding a catch-block in case something goes wrong with the `api.initPlatformCheckout(...)` call. Something like the following:
```javascript
api.initPlatformCheckout( e.data.platformCheckoutUserSession )
	.then( ( response ) => {
		if ( 'success' === response.result ) {
			window.location = response.url;
		} else {
			closeIframe();
		}
	} )
	.catch( ( error ) => {
		// Handle error.
	}
```

This way, things don't silently fail.

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -285,7 +285,9 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
 				wcpayTracks.recordUserEvent(
 					wcpayTracks.events.PLATFORM_CHECKOUT_OTP_COMPLETE
 				);
-				api.initPlatformCheckout().then( ( response ) => {
+				api.initPlatformCheckout(
+					e.data.platformCheckoutUserSession
```
waclawjacek, line 289: Since this value is delivered over `window.postMessage()`, I am concerned that it can be captured by the merchant site and abused. Is it possible to transmit the `user_session` from the platform checkout site to the merchant site without the merchant site being aware of what it is?

waclawjacek (changes requested):

client/checkout/platform-checkout/email-input-iframe.js
cesarcosta99: The concern is valid and I believe we have it sorted by making the session a one-time use token – it has been addressed in the platform checkout end. I've replied your question in the platform checkout's PR, can we continue this discussion there?

includes/class-wc-payments.php
```diff
@@ -970,6 +970,8 @@ public static function ajax_init_platform_checkout() {
 				'store_api_url'     => self::get_store_api_url(),
 				'account_id'        => $account_id,
 			],
+			// Maybe we should be checking the nonce in this request?
```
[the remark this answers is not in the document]
cesarcosta99: Added it in 0f42939.

client/checkout/platform-checkout/email-input-iframe.js
cesarcosta99: I created #4157 for this. I talked to David in p1650326845599849-slack-C022WMN88KG and thought we could address that in a different PR. Thanks for raising this.

includes/class-wc-payments.php
```diff
@@ -976,6 +985,7 @@ public static function ajax_init_platform_checkout() {
 				'store_api_url'     => self::get_store_api_url(),
 				'account_id'        => $account_id,
 			],
+			'user_session'         => isset( $_REQUEST['user_session'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['user_session'] ) ) : null,
```
waclawjacek, line 988: If I understand correctly, if `user_session` is missing, the request will never ever succeed. If so, let's bail early even before making the HTTP request. E.g. like this:

```php
if ( empty( $_REQUEST['user_session'] ) ) {
	wp_send_json_error(
		__( 'User session ID is missing', 'woocommerce-payments' ),
		500
	);
}
```

waclawjacek (approved):
LGTM! :shipit:

includes/class-wc-payments.php
cesarcosta99: Not exactly. If `user_session` is missing, it simply won't be added in the return URL. Besides, we might need to make some changes in how we handle the `user_session` to be able to redirect existing customers without going through the SMS OTP flow. I'll keep that in mind, though 👍

waclawjacek: @cesarcosta99 Sorry for the late response here but since this is a call to the platform checkout server to start a session, don't we have to have a `user_session` to authenticate?

PR #4133, opened 2022-04-13
Fix error in refunding in-person payments

### LINKED ISSUE #4100: [In Person Payments] Refunding IPP payments incorrectly reports that they are not captured.
### Describe the bug

When attempting to refund an IPP payment, the API is returning an error: 

`This payment is not captured yet. To cancel this order, please go to 'Order Actions' > 'Cancel authorization'. To proceed with a refund, please go to 'Order Actions' > 'Capture charge' to charge the payment card, and then trigger a refund via the 'Refund' button.`

This happens when refunding the IPP order from either the app, or the web. 

Following the error message's instructions does not work either. When cancelling the authorisation in wp-admin, we get the following error:

`Error: You cannot cancel this PaymentIntent because it has a status of succeeded. Only a PaymentIntent with one of the following statuses may be canceled: requires_payment_method, requires_capture, requires_confirmation, requires_action, processing.`

When capturing the charge (again), no change is apparent other than the successful `Order updated` banner, but further attempts to refund still result in the original error message.

### To Reproduce

1. In the mobile app, use the `+` button to add a simple payment order and take payment.
2. Open the order, and use the `Issue refund` button to start a refund to the card
3. Confirm the following steps

### Actual behavior
An error is reported in the app due to the above API error:
refund-error

### Screenshots

https://user-images.githubusercontent.com/2472348/162177061-132ad1f7-c811-446c-8dbe-34d554e274cd.mp4

### Expected behavior

Charges should be captured (and it looks like they are) and that reported to the apps so that they can be refunded from the app.

* iOS and Android (device or simulator)
* iOS/Android App, Firefox on the desktop for wp-admin testing

### Additional context

Multiple mobile engineers have seen this now, and though we can't pinpoint when it started, it's pretty common. We've tested this flow thoroughly in the past and this seems like a regression.

It's not 100% reproducible, but it happens much more often than not. Sometimes, older transactions can be refunded... but even that is inconsistent.

I have an old testing store running WCPay 3.2.3, and I can't reproduce it on that store.

anu-rock:
Fixes #4100

Update the `capture_terminal_payment` endpoint to use the correct intent status while marking the payment as completed. In the current state, all completed orders have their `_intention_status` meta key set as "requires_capture" when it should be "succeeded". This leads to undesired behavior such as error in refunding IPP orders.

#### Testing instructions

1. Create a new order in WooCommerce with “Cash on delivery” as the payment method.
2. Create a payment intent in Stripe Dashboard at `https://dashboard.stripe.com/test/connect/accounts/<account_id>/payments/new`. Make sure to have the “Capture funds later” option selected under “Manually enter card information.”
3. Make a POST request to `/payments/orders/<order_id>/capture_terminal_payment` using the intent id from step 2. Refer the [docs](https://github.com/woocommerce/woocommerce/wiki/Getting-started-with-the-REST-API) to set up authorization correctly.
4. Go to the order details page in WP Admin and refund the order.
5. Ensure that the order gets refunded successfully.
6. Before this change, you would get an alert dialog with the error "This payment is not captured yet. To cancel this order, please go to...".

-------------------

- [ ] Covered with tests (does not apply)
- [x] Tested on mobile

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -176,7 +176,7 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			// Store receipt generation URL for mobile applications in order meta-data.
 			$order->add_meta_data( 'receipt_url', get_rest_url( null, sprintf( '%s/payments/readers/receipts/%s', $this->namespace, $intent->get_id() ) ) );
 			// Actualize order status.
-			$this->order_service->mark_terminal_payment_completed( $order, $intent_id, $intent_status, $charge_id );
+			$this->order_service->mark_terminal_payment_completed( $order, $intent_id, $result['status'], $charge_id );
 
 			return rest_ensure_response(
 				[
--- a/changelog/fix-4100-ipp-refund-error
+++ b/changelog/fix-4100-ipp-refund-error
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Fix error in refunding in-person payments
+Fix an error in refunding In-Person Payments
--- a/changelog/fix-4100-ipp-refund-error
+++ b/changelog/fix-4100-ipp-refund-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix error in refunding in-person payments


includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -176,7 +176,7 @@ public function capture_terminal_payment( $request ) {
 			// Store receipt generation URL for mobile applications in order meta-data.
 			$order->add_meta_data( 'receipt_url', get_rest_url( null, sprintf( '%s/payments/readers/receipts/%s', $this->namespace, $intent->get_id() ) ) );
 			// Actualize order status.
-			$this->order_service->mark_terminal_payment_completed( $order, $intent_id, $intent_status, $charge_id );
+			$this->order_service->mark_terminal_payment_completed( $order, $intent_id, $result['status'], $charge_id );
```
anu-rock, line 179: @jessepearson @kalessil  Do we need this method call when we have a similar one in the `WC_Payment_Gateway_WCPay::capture_charge` method?

https://github.com/Automattic/woocommerce-payments/blob/cbf2927962e26b81d8b0e245678222254cc993a7/includes/class-wc-payment-gateway-wcpay.php#L2091-L2093

kalessil: IIRC, we need it here due to `$result = $is_intent_captured ? $result_for_captured_intent : $this->gateway->capture_charge( $order, false );`.

kalessil (approved):
LGTM!

includes/admin/class-wc-rest-payments-orders-controller.php
anu-rock: Right, that makes perfect sense :)

joshheald:
🎉 I've tested this in the iOS app post-release, and it's working well. Thanks for including it in 4.0!

PR #4139, opened 2022-04-14
Added payment methods cache cleaning when account is updated via webhook

### LINKED ISSUE #4138: Cached payment methods data persists after account is updated via webhook
### Describe the bug

When `account.updated` webhook request hits the WooCommerce Payments plugin, the account data is updated and cached again. The cache data for saved payment methods that are related to the account are not cleared. If an account is re-onboarded or switched, the stored payment methods will point to the old account, and the end-user will see stored credit cards for an invalid WooCommerce Payments account. The end-user will not be able to buy anything using stored payments methods and in the most cases, the end-user will not be able to delete them. The only way (currently) is to wait for cache to expire, and it can take up to 24 hours to delete, which is the default expiration time. 

### To Reproduce

1. Add new payment method to account or make sure that you have at least one payment method stored.
2. Re-onboard or create a new WooCommerce Payments account
3. Go to the previous mentioned account, and you will see stored payment methods that belong to old account
4. Deleting them won't do much because they belong to a different account, and the only solution is to manually delete cache data or let it expire by default.

### Actual behavior
When a webhook account update request is made, the saved payment methods cache data is not deleted

### Expected behavior
When a webhook account update request is made, the saved payment methods cache data needs to be deleted.

zmaglica:
Fixes #4138

When the account is updated via webhook, the webhook triggers a function that clears account cache data. In addition to the previous described functionality, new code is added that schedules an action that will clear cache for stored payment methods

#### Testing instructions

1. Make sure to have WordPress account with at least one payment methods that is added within 24 hours.
2. Buy random product, so we are certain that stored payment method is working, and it is cached.
3. Update account data via Stripe or re-onboard
4. Open wp_options table, and you should not see any transients that start with wcpay_pm_[CUSTOMER_ID]

-------------------

**Post merge**


includes/class-wc-payments-customer-service.php
```diff
@@ -327,6 +328,26 @@ public static function map_customer_data( WC_Order $wc_order = null, WC_Customer
...
+	 * Delete all cached payment methods. Used when account data is updated from WooCommerce Payments server.
+	 *
+	 * @return void
+	 */
+	public function handle_delete_all_cached_payment_methods() {
+		global $wpdb;
+
+		// Get last active users where payment methods cache key might be added. The payment method cache key expires after.
+		// 24 hours, so we will fetch active users from last 24 hours.
+		$users = $wpdb->get_results( $wpdb->prepare( "SELECT `user_id` FROM $wpdb->usermeta WHERE `meta_key` = %s AND `meta_value` >= %d", 'wc_last_active', strtotime( '-1 day' ) ), ARRAY_A );
```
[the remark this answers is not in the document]
zmaglica: That is a valid point. I will move it inside this class and inject it in service

includes/class-wc-payments-account.php
```diff
@@ -981,6 +989,7 @@ function () {
 	 * @return array|bool|string Either the new account data or false if unavailable.
 	 */
 	public function refresh_account_data() {
+		$this->maybe_clear_payment_methods_cache();
```
marcinbot, line 992: I don't think this should belong in `WC_Payments_Account`. Seems more of a `WC_Payments_Customer_Service` kind of thing.

includes/class-wc-payments-account.php
```diff
@@ -40,15 +40,23 @@ class WC_Payments_Account {
...
 	 *
 	 * @param WC_Payments_API_Client $payments_api_client Payments API client.
 	 * @param Database_Cache         $database_cache      Database cache util.
 	 */
 	public function __construct( WC_Payments_API_Client $payments_api_client, Database_Cache $database_cache ) {
-		$this->payments_api_client = $payments_api_client;
-		$this->database_cache      = $database_cache;
+		$this->payments_api_client      = $payments_api_client;
+		$this->database_cache           = $database_cache;
+		$this->action_scheduler_service = new WC_Payments_Action_Scheduler_Service( $this->payments_api_client );
```
marcinbot, line 59: Thanks for refactoring this! I think this should be passed as an argument to the constructor from the `WC_Payments` class. Similar to how we do it for the [gateway](https://github.com/Automattic/woocommerce-payments/blob/19f661dd27f97597e98db73a35bf01be4050a091/includes/class-wc-payments.php#L316) and [other](https://github.com/Automattic/woocommerce-payments/blob/19f661dd27f97597e98db73a35bf01be4050a091/includes/class-wc-payments.php#L320) classes.

marcinbot (commented):
Looks fine, but the code might need to be moved somewhere else.

I also wonder if we could make a refactor here to use the new `Database_Cache` class to cache the saved payment methods. But that might require an investigation of what happens during ajax requests and some special handling of dynamic keys.

includes/class-wc-payments-account.php
zmaglica: It makes sense to put it inside `WC_Payments_Customer_Service `. I will do that.

The reason why I put it here is that I wanted to avoid injecting `WC_Payments_Action_Scheduler_Service` inside `WC_Payments_Customer_Service ` and then inject that service inside `WC_Payments_Accounts` which will results in broken tests that needs to be fixed because constructor needs to be updated.

includes/class-wc-payments-account.php
zmaglica: You are right. I can't believe that I put new statement inside constructor... I will fix that.

includes/class-wc-payments-account.php
marcinbot: Ah, do we even need to use the `WC_Payments_Action_Scheduler_Service`? The account service refreshes the cache synchronously and that hasn't been much of an issue, even before the recent improvements.

It has been further improved now that the refresh request doesn't affect concurrent requests, but that's only a minor point in favor of migrating the method caching to `Database_Cache`.

includes/class-wc-payments-account.php
marcinbot: Just a reminder about this outstanding item :)

includes/class-database-cache.php
```diff
@@ -16,6 +16,10 @@ class Database_Cache {
 	const ACCOUNT_KEY        = 'wcpay_account_data';
 	const BUSINESS_TYPES_KEY = 'wcpay_business_types_data';
 	const CURRENCIES_KEY     = 'wcpay_multi_currency_cached_currencies';
+	/**
+	 * Payment methods transient. Used in conjunction with the customer_id to cache a customer's payment methods.
```
marcinbot, line 20: Let's not use the `transient` terminology since it's an already-established term in WP. `Payment methods key prefix` should work here.

includes/class-database-cache.php
```diff
@@ -16,6 +16,10 @@ class Database_Cache {
...
+	 */
+	const PAYMENT_METHODS_KEY = 'wcpay_pm_';
```
marcinbot, line 22: Related to the above, maybe we could name this `PAYMENT_METHODS_KEY_PREFIX`

includes/class-database-cache.php
```diff
@@ -132,6 +136,29 @@ public function disable_refresh() {
...
+	 *
+	 * @param string $key Cache key prefix to delete.
+	 *
+	 * @return void
+	 */
+	public function delete_by_prefix( string $key ) {
+		/**
+		 * Protection against accidentally deleting all options or options that are not related to WcPay caching.
+		 */
+		if ( ! in_array( $key, [ self::ACCOUNT_KEY, self::BUSINESS_TYPES_KEY, self::CURRENCIES_KEY, self::PAYMENT_METHODS_KEY ], true ) ) {
```
marcinbot, line 150: I think here we could just check for `self::PAYMENT_METHODS_KEY(_PREFIX)` since that's the only prefix we support so far. Others are just plain strings.

phpunit.xml.dist.bak
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
```
marcinbot, line 1: Is this file intended to be committed?

includes/class-wc-payments-account.php
```diff
@@ -935,6 +943,9 @@ function () {
 					// below re-create it if the server tells us on-boarding is still disabled.
 					delete_transient( self::ON_BOARDING_DISABLED_TRANSIENT );
 
+					// Delete all saved payment methods cache.
+					$this->database_cache->delete_by_prefix( Database_Cache::BUSINESS_TYPES_KEY );
```
marcinbot, line 947: I think this should be a separate method on the customer service and the account service shouldn't care about the payment methods.

Also, this should be `PAYMENT_METHODS_KEY` not `BUSINESS_TYPES_KEY`, right?

In addition, this will mess with the methods cache ttl. It seems to be intended to be 24h, but more frequent account updates will trigger it to be 2h or more. If nothing else, it can lead to some difficult-to-debug issues.

phpunit.xml.dist.bak
zmaglica: Nope. I will remove it.

includes/class-wc-payments-account.php
zmaglica: This is fixed.

phpunit.xml.dist.bak
zmaglica: Removed.

includes/class-database-cache.php
zmaglica: This is refactored.

includes/class-database-cache.php
zmaglica: Noted. The constant is changed.

includes/class-database-cache.php
zmaglica: Fixed.

includes/class-database-cache.php
```diff
@@ -132,6 +136,31 @@ public function disable_refresh() {
...
+	 * @param string $key Cache key prefix to delete.
+	 *
+	 * @return void
+	 */
+	public function delete_by_prefix( string $key ) {
+		/**
+		 * Protection against accidentally deleting all options or options that are not related to WcPay caching.
+		 * Since only one cache key prefix is supported, we will check only this one by checking does key starts with payment method key prefix.
+		 * Feel free to update this statement if more prefix cache keys you are planning to add.
+		 */
```
marcinbot, line 151: Nitpick: let's use the regular `//` comments rather than phpdoc-style comments.

marcinbot (commented):
Looks good! Let's just move the delete function to the proper service.

includes/class-database-cache.php
zmaglica: Fixed.

includes/class-wc-payments-account.php
```diff
@@ -977,6 +986,15 @@ function () {
...
 
+	/**
+	 * Delete all saved paymenent methods that are stored inside database cache driver.
+	 *
+	 * @return void
+	 */
+	public function delete_cached_payment_methods() {
+		$this->database_cache->delete_by_prefix( Database_Cache::PAYMENT_METHODS_KEY_PREFIX );
+	}
+
```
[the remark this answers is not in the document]
zmaglica: Fixed.

PR #4166, opened 2022-04-20
Fix missing `customer_id` on platform checkout

asumaran:
Fixes platform checkout issue 822

Assigns the recently created customer Id to `$customer_id`. It seems that was the initial intention.

#### Testing instructions

To reproduce the error you'll need to run the following steps while using the `dev` branch for the WC Payments plugin on your merchant store.

- You need to log in on the platform first with your customer account.
- Delete all cookies on the **merchant store** (while still logged on the platform).
- Add at least a product to the cart.
- Go to the checkout page.
- Enter your customer email.
- Enter the code (00000) in the popup.
- Click the "Place Order" button
- You'll see the error banner and the error in the log.

To test the fix:

Run the same steps above but in the end you'll see no error and no error should be logged.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -812,6 +812,17 @@ class WC_Payments {
 	}
 
 	/**
+	 * Sets the customer service instance. This is needed only for tests.
+	 *
+	 * @param WC_Payments_Customer_Service $customer_service_class Instance of WC_Payments_Customer_Service.
+	 *
+	 * @return void
+	 */
+	public static function set_customer_service( WC_Payments_Customer_Service $customer_service_class ) {
+		self::$customer_service = $customer_service_class;
+	}
+
+	/**
 	 * Registers the payment method with the blocks registry.
 	 *
 	 * @param Automattic\WooCommerce\Blocks\Payments\PaymentMethodRegistry $payment_method_registry The registry.
@@ -942,7 +953,7 @@ class WC_Payments {
 		if ( null === $customer_id ) {
 			// create customer.
 			$customer_data = WC_Payments_Customer_Service::map_customer_data( null, new WC_Customer( $user->ID ) );
-			self::$customer_service->create_customer_for_user( $user, $customer_data );
+			$customer_id   = self::$customer_service->create_customer_for_user( $user, $customer_data );
 		}
 
 		$account_id = self::get_account_service()->get_stripe_account_id();
--- a/tests/unit/test-class-wc-payments.php
+++ b/tests/unit/test-class-wc-payments.php
@@ -91,6 +91,40 @@ class WC_Payments_Test extends WP_UnitTestCase {
 		$this->assertEquals( 'woocommerce_rest_missing_nonce', $response->get_data()['code'] );
 	}
 
+	public function test_ajax_init_platform_checkout_sends_correct_data() {
+		// Necessary in order to prevent die from being called.
+		define( 'DOING_AJAX', true );
+
+		$customer_id = 'cus_123456789';
+
+		$pre_http_request_cb = function ( $preempt, $parsed_args, $url ) use ( $customer_id ) {
+			$body = json_decode( $parsed_args['body'] );
+			$this->assertEquals( $customer_id, $body->customer_id );
+			return [ 'body' => wp_json_encode( [] ) ];
+		};
+
+		$wp_die_ajax_handler_cb = function () {
+			return function ( $message, $title, $args ) {};
+		};
+
+		add_filter( 'pre_http_request', $pre_http_request_cb, 10, 3 );
+		add_filter( 'wp_die_ajax_handler', $wp_die_ajax_handler_cb );
+
+		$mock_customer_service = $this->getMockBuilder( 'WC_Payments_Customer_Service' )
+									->disableOriginalConstructor()
+									->getMock();
+		$mock_customer_service
+			->expects( $this->once() )
+			->method( 'create_customer_for_user' )
+			->with( $this->anything(), $this->anything() )
+			->will( $this->returnValue( $customer_id ) );
+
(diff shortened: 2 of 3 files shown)


lovo-h (approved):
This fix looks good to me. Good find!

**Issue**:
* WCPay on `dev`, Platform repo on `trunk` branch.
* Log in on platform site, delete all cookies in merchant site, add product, checkout, enter customer email, enter OTP, click "Place Order" was able to replicate the issue (see image).

image


**Fix**:
* WCPay on `as-fix-missing-customer_id`, Platform repo on `trunk` branch.
* Log in on platform site, delete all cookies in merchant site, add product, checkout, enter customer email, enter OTP, click "Place Order", resulted in a successful checkout (see image).

image

As we discussed, we may want to add a test for this in order to avoid a similar issue, moving forward. Although, I am not sure if we want to add the test here on a separate PR.

tests/unit/test-class-wc-payments.php
```diff
@@ -167,6 +167,41 @@ function () {
 		$this->assertNotEquals( 'woocommerce_rest_missing_nonce', $response->get_data()['code'] );
 	}
 
+	public function test_temp() {
+		// Necessary in order to prevent wp_die() from being called.
```
lovo-h, line 171: This was my typo. I didn't notice this yesterday but I meant to say the following:
```
// Necessary in order to prevent die from being called.
```

Since defining `DOING_AJAX` to `true` forces `wp_die()` to be called.

tests/unit/test-class-wc-payments.php
```diff
@@ -167,6 +167,41 @@ function () {
...
+		define( 'DOING_AJAX', true );
+
+		$self        = $this;
+		$customer_id = 'cus_123456789';
+
+		$pre_http_request_cb = function ( $preempt, $parsed_args, $url ) use ( $self, $customer_id ) {
+			$body = json_decode( $parsed_args['body'] );
+			$self->assertEquals( $customer_id, $body->customer_id );
```
lovo-h, line 179: I also missed this yesterday but I did a quick test and found that `$self` can be removed and `$this` can be used in its place.

lovo-h (changes requested):
The test looks good to me. Just a few typos that we may want to take care of before merging.

Also, we may want to get an additional pair of eyes 👀  on this test since we both worked on it 🙂

tests/unit/test-class-wc-payments.php
```diff
@@ -167,6 +167,41 @@ function () {
 		$this->assertNotEquals( 'woocommerce_rest_missing_nonce', $response->get_data()['code'] );
 	}
 
+	public function test_temp() {
```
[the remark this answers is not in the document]
asumaran: heh. I missed this one too.

ricardo (commented):
Just a quick update: I spent some time debugging my local environment, but this tests well.

I'll do some more testing around the `customer_id` assignment before I submit the review.

tests/unit/test-class-wc-payments.php
```diff
@@ -167,6 +167,40 @@ function () {
 		$this->assertNotEquals( 'woocommerce_rest_missing_nonce', $response->get_data()['code'] );
 	}
 
+	public function test_ajax_init_platform_checkout_sends_correct_data() {
```
ricardo, line 170: Since we're only asserting the Customer id in this test, should we name it something like "test_ajax_init_platform_checkout_creates_new_customer"?

includes/class-wc-payments.php
```diff
@@ -844,6 +844,17 @@ public static function update_plugin_version() {
...
 	}
 
+	/**
+	 * Sets the customer service instance. This is needed only for tests.
+	 *
+	 * @param WC_Payments_Customer_Service $customer_service_class Instance of WC_Payments_Customer_Service.
+	 *
+	 * @return void
+	 */
+	public static function set_customer_service( WC_Payments_Customer_Service $customer_service_class ) {
```
ricardo, line 854: Nit: Maybe move this after `get_customer_service`? So it's more grouped like `set_database_cache` and `set_account_service`.

ricardo (approved):
LGTM and tests well. Nice work!

tests/unit/test-class-wc-payments.php
lovo-h: I also considered this but we may want to add additional data points that we would like to test, other than just customer ID (e.g. `user_id`, `email`, etc.). If so, we could add them here. 

Although, at the moment we _are_ only testing customer ID so `test_ajax_init_platform_checkout_sends_correct_customer_id` works just as well and we can always rename it in the future.

Thoughts?

ricardo: > Although, at the moment we are only testing customer ID so `test_ajax_init_platform_checkout_sends_correct_customer_id` works just as well and we can always rename it in the future.

I agree 👍. Let's keep the name consistent with its current state.

asumaran: This test only focus on the `customer_id` creation. I think it will involve additional logic to check on the other data.

includes/class-wc-payments.php
asumaran: done

tests/unit/test-class-wc-payments.php
ricardo: I don't think we need to check on other data right now, but renaming the test sounds good.

asumaran: done [d7e94e2](https://github.com/Automattic/woocommerce-payments/pull/4166/commits/d7e94e2f0346a8267d0d7a14c2f64ca202de0522)

lovo-h (approved):
LGTM :shipit:

PR #4140, opened 2022-04-14
Add missing VAT information prompt when downloading invoice

luizreis:
Fixes #3933

#### Changes proposed in this Pull Request

This PR adds a new `hasSubmittedVatData` flag to the `wcpaySettings.accountStatus` object, based on the value we get from the basic account info call. This value is then used when merchants click on the download button for a VAT invoice.

If the data has already been submitted, the document should be downloaded right away. If the data hasn't been submitted, a modal should pop up with the form to submit the VAT details.

We're only handling the modal as part of this PR. The actual form and data submission will be handled in a separate PR for #3934.

Note: A few small fixes like removing unused props from the Document type, removing a few extra `break;` statements, and re-adding the document data methods to the `wordpress/data` store have also been included in this PR.

#### Testing instructions

* Create a new `vat_invoice` for your account if you don't have one by running the following command with 1803-gh-Automattic/woocommerce-payments-server
    ```
    wp wcpay add_document_for_account {Your account id} vat_invoice --date="2022-03-01" --period_from="2022-02-01 00:00:00" --period_to="2022-02-28 23:59:59"
    ```
* Apply this PR's branch and navigate to the documents list
* Click on the download button for the VAT invoice and make sure a modal pops up
    * _The modal is pretty simple for now and should just say VAT details/VAT form_
* Apply 1896-gh-Automattic/woocommerce-payments-server to your local server or sandbox, and refresh the cached account data
* Refresh the document's list page and click on the download button
    * If your VAT data hasn't been submitted, the same modal should pop up
        * Please use the instructions on the server PR to add the VAT details for your account and test the next step
    * If you have already submitted the data, the invoice should be downloaded
        * _The download will fail while 1803-gh-Automattic/woocommerce-payments-server is not merged, but trying to access the page should be enough to test this PR_ 

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _QA Testing Not Applicable_

PR #4168, opened 2022-04-20
Add VAT details form when downloading invoice

### LINKED ISSUE #3934: Tax-related information: data update form
### Description

We should add a screen for merchants to update their tax-related information or provide it for the first time.

### Acceptance criteria

* EU merchants can provide us with their tax-related information

### Designs

Designs can be checked in paJDYF-3mi-p2 and more details around scope are available in paJDYF-3sv-p2.

There are currently no designs for this form. We'll update this issue when they're available.

luizreis:
Fixes #3934 

This PR builds on top of #4140, adding the actual VAT details form to the modal. The form is composed of two steps: in the first, we ask merchants for their VAT number — in case they have one — and validate it; in the second step we use the data we were able to get when validating the VAT number to ask for the company name and address.

Merchants are able to submit the details without specifying a VAT number, but business name and address are required. Once the details are stored in the server, the `hasSubmittedVatData` flag is set to true on the client to allow invoices to be downloaded without requiring a page refresh. The account cache is also refreshed in the background, so that the new status is reflected right away, instead of having to wait for an automatic refresh.

As soon as the form is submitted and the flag is updated, merchants will be redirected to the invoice document, without having to click on Download a second time. Further downloads should be processed directly, without asking for the details again.

#### Testing instructions

_We don't have a way to validate VAT numbers for some countries, like the US. In these cases, an error will be displayed when attempting to validate it. Merchants should be able to proceed without informing a VAT number in such cases._

_If using an account from the UK, the local server and sandbox will use a test API. You can find valid VAT numbers to test it [here](https://github.com/hmrc/vat-registered-companies-api/blob/main/public/api/conf/1.0/test-data/vrn.csv)._

_If using an account from the EU, the VIES production service will be used, so you'll need to use a real VAT number._

* Apply this PR's branch and 1803-gh-Automattic/woocommerce-payments-server on your local server or sandbox
* Create a new `vat_invoice` for your account if you don't have one by running the following command
    ```
    wp wcpay add_document_for_account {Your account id} vat_invoice --date="2022-03-01" --period_from="2022-02-01 00:00:00" --period_to="2022-02-28 23:59:59"
    ```
* Navigate to the documents list and click on the download button for the VAT invoice
    * Make sure a modal pops up with the VAT details form
    * Check "I’m registered for a VAT number" and make sure your account's country prefix is used in the input
        * Input an invalid VAT number, click on Continue, and assert that a notice is displayed with the error
        * Input a valid VAT number, click on Continue, and assert that the form proceeds to the next step with the details from the VAT check
            * Modify or keep the placeholder values for business name and address and click on Confirm
            * Make sure the form is submitted correctly and the document is downloaded automatically
            * Check the account_meta table on the server to make sure the data was saved correctly
* Remove the VAT details from the server and clear the account cache on the client
    * Navigate to the documents list and click on the download button for the VAT invoice
    * Click on Continue without checking "I’m registered for a VAT number"
        * Make sure the Confirm button is disabled until a business name and address are provided
    * Provide a business name and address and click on Confirm
        * Make sure the form is submitted correctly and the document is downloaded automatically 
        * Check the account_meta table on the server to make sure the data was saved correctly

-------------------

**Post merge**

### DIFF
--- a/client/vat/form-modal/index.tsx
+++ b/client/vat/form-modal/index.tsx
@@ -25,8 +25,6 @@ const VatFormModal = ( {
 	return isModalOpen ? (
 		<Modal
 			title={ __( 'VAT details', 'woocommerce-payments' ) }
-			isDismissible={ false }
-			shouldCloseOnClickOutside={ true }
 			onRequestClose={ () => setModalOpen( false ) }
 		>
 			<VatForm onCompleted={ onCompleted } />
--- a/client/vat/form-modal/index.tsx
+++ b/client/vat/form-modal/index.tsx
@@ -11,13 +11,16 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import VatForm from '../form';
+import { VatFormOnCompleted } from '../types';
 
 const VatFormModal = ( {
 	isModalOpen,
 	setModalOpen,
+	onCompleted,
 }: {
 	isModalOpen: boolean;
 	setModalOpen: ( value: boolean ) => void;
+	onCompleted: VatFormOnCompleted;
 } ): JSX.Element | null => {
 	return isModalOpen ? (
 		<Modal
@@ -26,7 +29,7 @@ const VatFormModal = ( {
 			shouldCloseOnClickOutside={ true }
 			onRequestClose={ () => setModalOpen( false ) }
 		>
-			<VatForm />
+			<VatForm onCompleted={ onCompleted } />
 		</Modal>
 	) : null;
 };
--- a/client/vat/form-modal/index.tsx
+++ b/client/vat/form-modal/index.tsx
@@ -10,6 +10,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import VatForm from '../form';
 
 const VatFormModal = ( {
 	isModalOpen,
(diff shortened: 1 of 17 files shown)


jessepearson:
@luizreis Do you think the address entry should be a textarea instead of an input field? If we leave it as an input, then the address has to be one line and that's how it will also show on the invoice.

Example:
> 123 North Street, Miami, FL 33333 United States

vs

> 123 North Street
> Miami, FL 33333
> United States

Screen Shot 2022-04-21 at 10 49 44 AM

client/vat/form-modal/index.tsx
```diff
@@ -25,7 +29,7 @@ const VatFormModal = ( {
 			shouldCloseOnClickOutside={ true }
```
jessepearson, line 29: Should we do `isDismissible={ true }` instead since there's not a _Cancel_ button? Just so there's some sort of UI to dismiss the modal?

jessepearson (approved):
Pre-approving this, since it works as expected and I just have a question on a minor UI change.

client/vat/form-modal/index.tsx
luizreis: No worries, GH is bit quirky when commenting on code that hasn't been touched in the current PR. I've removed the `isDismissible={ false }` prop, as modals default that to true, and added a test in https://github.com/Automattic/woocommerce-payments/pull/4168/commits/57402b44d14c57ced68247f9d563483b476036b7.

PR #4174, opened 2022-04-21
Fix KYC treatment mode API and menu issues

### LINKED ISSUE #4171: KYC treatment mode not working properly
### Describe the bug
There are two reported issues with the new KYC flow:
1. An onboarded user has **WC Pay > Settings** menu pointing to onboarding page. Probably we're still adding onboarding menu entry even after being onboarded for some reason. In fact, they could access the onboarding page, when they shouldn't.
2. While trying to onboard in treatment mode the country list doesn't load, the server request returns `data: []`.

More details in #woo-devs request: p1650470113900559-slack-C7U3Y3VMY

### To Reproduce
Still not able to reproduce the 1st issue.

2nd issue
- Launch a new JN site with WC Pay in test mode.
- Enable tracking in **WooCommerce > Settings > Advanced > Allow usage of WooCommerce to be tracked**
- Get the treatment bookmarklet from Abacus `woo_wcpayments_tasklist_click_introducing_select_business_type_202203_v1` experiment.
- Run the bookmarklet in the site admin.
- Go to onboarding
- You'll see the country select empty, and the server request returning `data: []`

### Additional context
WC Pay v4.0.1

ismaeldcom:
Fixes #4171

- Avoid redirecting the settings menu to onboarding page.
- Redirect WC Admin task to onboarding page rather than settings, that then will redirect to connect page.
- Use user token for new onboarding requests. I'm still concerned about this one, not sure why it was working fine before. Although looking at the server controller looks like we should be using it from the beginning.

#### Testing instructions
- Get treatment bookmarklet from `woo_wcpayments_tasklist_click_introducing_select_business_type_202203_v1` experiment in Abacus.
- You'll need a non onboarded local site with **WCPay Dev > Proxy WPCOM requests** and **WooCommerce > Settings > Advanced > Allow usage of WooCommerce to be tracked** enabled.
- Run the treatment bookmarklet, it should show a success message.
- Go to **WooCommerce > Set up payments** task, click on **Finish Setup** and then on **Set up**.
- Should be redirected to new onboarding page with the KYC form.
- The form should load countries and filling it should render the required verification details.
- Complete the onboard.
- Check that the settings menu **Payments > Settings** points to `?page=wc-settings&tab=checkout&section=woocommerce_payments` and not `?page=wc-admin&path=/payments/onboarding`

-------------------

**Post merge**
QA Testing Not Applicable


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1148,7 +1148,9 @@ public function get_onboarding_business_types() {
 		return $this->request(
 			[],
 			self::ONBOARDING_API . '/business_types',
-			self::GET
+			self::GET,
+			true,
+			true
```
oaratovskyi, line 1153: maybe it would be a good idea to add a test that checks that the method was called with user_token = true, won't be easy to break it again :)

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1176,7 +1178,9 @@ public function get_onboarding_required_verification_information( string $countr
 		return $this->request(
 			$params,
 			self::ONBOARDING_API . '/required_verification_information',
-			self::GET
+			self::GET,
+			true,
```
oaratovskyi, line 1182: maybe it would be a good idea to add a test that checks that the method was called with user_token = true, won't be easy to break it again :)

includes/class-wc-payments-account.php
```diff
@@ -476,6 +476,12 @@ public function maybe_redirect_to_onboarding() {
 			return false;
 		}
 
+		// Redirect directly to onboarding page if come from WC Admin task and are in treatment mode.
+		$http_referer = sanitize_text_field( wp_unslash( $_SERVER['HTTP_REFERER'] ?? '' ) );
```
oaratovskyi, line 480: I think it would be good to cover this if in the unit test as well!

oaratovskyi (approved):
LGTM and works well :shipit: 
I am pre-approving it and left some comments that are not blocking the PR from being shipped!

includes/wc-payment-api/class-wc-payments-api-client.php
ismaeldcom: Addressed in 0df59ed

includes/wc-payment-api/class-wc-payments-api-client.php
ismaeldcom: Addressed in 0df59ed

includes/class-wc-payments-account.php
ismaeldcom: 🤔  I can't think of an easy way to test it, because `is_in_onboarding_treatment_mode` is a static method, and we can't mock it, right?

We made it static as all the other utils methods, to easily call it from different places. Probably could be a better option to make `Experimental_Abtest` a singleton. But AFAIK we're using it as a temporal replacement while it's being included in Jetpack.

oaratovskyi: Yeah, you're right, it would require additional moves.
p.s. I did not know that `Experimental_Abtest` is temporary!

ismaeldcom:
Hey @oaratovskyi  I've added a new change to mitigate those cases where the user is not connected to Jetpack. By checking if the server is connected before doing the onboarding redirect, we could redirect to the jetpack connection page first.

oaratovskyi (approved):
Tested it, works correct.
:shipit:

PR #4180, opened 2022-04-21
Add support for direct document access

luizreis:
Fixes 1838-gh-Automattic/woocommerce-payments-server

This PR modifies the document download button to use a helper method and window.open instead of relying on the
`<a>` href property to open the document. This is useful to tidy up the retry mechanism and also to allow documents to be downloaded without having a button on the page.

It also adds a handler to the documents list that will automatically download a document, or go through the VAT validation process, if `document_id` and `document_type` are present in the query string. This is used to create a direct link from the new document email to the actual report.

Direct links will attempt to be opened in the current tab. However, depending on the content-disposition header, files might be automatically downloaded. In such cases, the current page will remain the documents list.

#### Testing instructions

To be tested in conjunction with 1926-gh-Automattic/woocommerce-payments-server.

-------------------

**Post merge**

### DIFF
--- a/client/documents/list/test/index.tsx
+++ b/client/documents/list/test/index.tsx
@@ -4,7 +4,7 @@
  * External dependencies
  */
 import * as React from 'react';
-import { createEvent, fireEvent, render, screen } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import user from '@testing-library/user-event';
 import { getQuery, updateQueryString } from '@woocommerce/navigation';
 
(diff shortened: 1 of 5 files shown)


client/documents/list/test/index.tsx
```diff
@@ -6,7 +6,7 @@
 import * as React from 'react';
 import { createEvent, fireEvent, render, screen } from '@testing-library/react';
```
jessepearson, line 7: > 'createEvent' is defined but never used
> 'fireEvent' is defined but never used

jessepearson (approved):
Seems much "cleaner", if that makes sense. Pre-approving pending the changes that GitHub has suggested about unused functions/methods.

client/documents/list/test/index.tsx
luizreis: Good catch! Removed in https://github.com/Automattic/woocommerce-payments/pull/4180/commits/b78577bd2bb14ffd9247e38df5ff07f51ee57ef8

PR #4193, opened 2022-04-25
Added mode check on order tracking

### LINKED ISSUE #4194: Fix order tracking when WooCommerce Payments Plugin mode is changed
### Describe the bug
The good example of this issue can be explained when you have recurring subscription. When you create a subscription, the order is created. For every subscription renewal, the new order is created. When an order is created, the order tracking job is fired. When mode is changed or switched from test to prod and vice versa, the orders that are scheduled for tracking will fail because WCPay server will look for customers data to an entirely different mode and will throw exception which will raise a lot of flaky and invalid errors.

### To Reproduce

1. Buy new subscription
2. Switch WooCommerce Payments mode
3. Wait for subscription renewal that will trigger new order
4. You will see errors because cron task failed

### Actual behavior
When mode is changed, the order tracking job keeps tracking that order, which is invalid at this point and will result in many invalid errors.

### Expected behavior
When mode is changed, order tracking job will stop track that order

* OS: All
* Browser N/A
* Version All

zmaglica:
Fixes #4194

When mode is changed on Stripe (from test to prod and vice versa), the order tracking becomes invalid because orders are linked to specific mode (production or test). Order tracking is a scheduled job and changing mode can cause scenarios where the order is invalid because it is related to an entirely different mode with own payment methods and orders (on Stripe). This PR fixes it by adding a metadata value to order that will tell in what environment or mode the order has been created.

#### Testing instructions

1. Create new order (subscription is the best since it creates recurring orders)
2. Switch WooCommerce Payments mode
3. Check crons and make sure that order tracking for previous order is not scheduled or will not trigger an error

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-1453-switch-test-live-mode-issue
+++ b/changelog/fix-1453-switch-test-live-mode-issue
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed issue with order tracking when mode is changed
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1071,6 +1071,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		$payment_method = $payment_information->get_payment_method();
 		$order->update_meta_data( '_payment_method_id', $payment_method );
 		$order->update_meta_data( '_stripe_customer_id', $customer_id );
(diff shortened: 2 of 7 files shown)


zmaglica:
> Looks good. As discussed, I think `_mode` suffix might be better than `_env`. I.e. `_order_mode` and `current_mode`.
> 
> Also needs a changelog, but ready to ship otherwise

Thank you for your comment. I added proposed changes.

marcinbot (approved):
:shipit:

PR #4203, opened 2022-04-27
Add a check for if the SMS OTP iframe modal is off the bottom of the screen.

### LINKED ISSUE #4108: Can't scroll to SMS OTP modal on merchant site
### Describe the bug
When visiting the checkout page after saving the account for faster checkout the SMS OTP modal loads automatically but if the site header is too big the modal is partially out of screen and it blocks the scroll. Meaning the customer will have to close it before scrolling and then edit the email to trigger the modal again. I doubt customers will know they need to do that so we should make sure the modal is fully visible or at least allow scroll when it is being shown.

### To Reproduce
- With a small screen (tested on a laptop here) Go to the platform checkout merchant demo store
- Buy any product and check `saving the account for faster checkout`
- Return to the store and add another product to the cart
- Go to the checkout page
- Notice the SMS OTP page is out of the screen and the scroll is disabled

### Screenshots
Screen Shot 2022-04-08 at 11 08 47 AM

### Expected behavior
- Customer should either be allowed to scroll or we should make sure the modal is fully visible at all times

* OS: Mac OS
* Browser chrome

brettshumaker:
Fixes #4108

We had an existing check for the SMS OTP modal being off the top edge of the screen, but not for the bottom. This PR introduces that check.

#### Testing instructions

1. Check out this PR and build the assets.
2. Go to the shortcode checkout page on a store configured to use the platform checkout.
3. Adjust your browser window to be about 1000x600 and make sure the email field is near the bottom of the screen.
4. Enter a valid platform checkout email address.
5. Verify that when the SMS OTP modal pops up, the window adjusts its scroll position to show the entire modal.

-------------------

**Post merge**

### DIFF
--- a/client/checkout/platform-checkout/email-input-iframe.js
+++ b/client/checkout/platform-checkout/email-input-iframe.js
@@ -84,8 +84,18 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
 			return;
 		}
 
-		// Check if the iframe is off the top of the screen and scroll back into view.
-		if ( 0 >= iframe.getBoundingClientRect().top ) {
+		/**
+		 * If the iframe is off the top of the screen
+		 * OR the iframe is off the bottom of the screen
+		 * scroll the window so the iframe is in view.
+		 */
+		if (
+			0 >= iframe.getBoundingClientRect().top ||
+			0 >=
+				window.innerHeight -
+					iframe.getBoundingClientRect().height +
+					iframe.getBoundingClientRect().top
+		) {
 			const topOffset = 50;
 			const scrollTop =
 				document.documentElement.scrollTop +
--- a/changelog/fix-4108-sms-otp-modal-off-screen
+++ b/changelog/fix-4108-sms-otp-modal-off-screen
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Ensure platform checkout SMS OTP iframe modal is always visible.


client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -84,8 +84,18 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
+		/**
+		 * If the iframe is off the top of the screen
+		 * OR the iframe is off the bottom of the screen
+		 * scroll the window so the iframe is in view.
+		 */
+		if (
+			0 >= iframe.getBoundingClientRect().top ||
+			0 >=
+				window.innerHeight -
+					iframe.getBoundingClientRect().height +
```
leonardola, line 96: This should be a subtraction not an addition. Let's say we have a 460px high screen and 420px high modal and 210px top

the way it was written will result in 460-420+211 = 250 but it should actually be 460-420-210 = -164 or we could also do 460 - ( 420 + 211 ) = -164

leonardola (changes requested):
This was not working for me. So I debugged and found out we are adding a number where we should be subtracting.

client/checkout/platform-checkout/email-input-iframe.js
brettshumaker: Ah, you're right. I think I was trying to move some parentheses around to appease the linter and accidentally removed a pair. I'll get that patched up.

brettshumaker: Fixed in [a1e0ac7](https://github.com/Automattic/woocommerce-payments/pull/4203/commits/a1e0ac7e37fba769f4e9c3cdcc8667c85a7f0ff9)

leonardola (approved):
Now it works as expected. Thanks for fixing this one

PR #4206, opened 2022-04-28
fix proposal for Apple pay issue w/ WordPress in subfolder

### LINKED ISSUE #3565: Apple Pay verification fails when using an alternate folder structure
### Describe the bug

There is a problem with Apple Pay verification when using a different folder structure. (Bedrock for example – https://roots.io/bedrock/)

The variable $domain_name is taken from the site_url, which may contain a sub-directory, in this [line](https://github.com/Automattic/woocommerce-payments/blob/cbf5b9e441960058fd33eb7c3542d9586daede6c/includes/class-wc-payments-apple-pay-registration.php#L69)

### To Reproduce

Install WordPress on an alternate folder stucture, such as `www.example.com/wp`

### Actual behavior

An alternate folder structure results in the error:

`You passed an invalid URI for your domain name. It must not contain a port, path, query, or fragment, and should be just a bare top-level domain, such as https://www.example.com`

### Expected behavior

I would expect the domain name to be passed correctly without port, path, query, or fragment

### Additional context

Raised in forums [here](https://wordpress.org/support/topic/apple-pay-invalid-uri-with-custom-folder-structure/#post-15186119).

ironicmoka:
Fixes #3565 

- Parse URL from `get_site_url()`, stripping down any paths or subfolders, so the actual domain is passed down to the Apple Pay verification request.
- Better document things in the Apple Pay verification error notice.

#### Testing instructions

Test things breaking with the latest release of WooCommerce Payments, and then update the plugin at the end to see the fix in action:

1. Go to https://jurassic.ninja > More options > Launch Multisite on subdirs.
2. Wait for the creation of a multisite WP installation, then at the top bar, go to **My Sites > Network Admin > Plugins**, and network install these plugins from WP.org:
    - WooCommerce.
    - WooCommerce Payments.
    - WooCommerce Payments Dev Tools (download the zip from the repository and upload it to plugins).
3. Create a new site on a subfolder under **Sites > Add new**.
4. Go to the dashboard of the recently added site, and navigate to the WCPay Dev plugin.
5. Make sure all options are disabled, except for "Enable API request redirection".
6. Expose your server to a public URL using Ngrok or JT and update the "Redirect API requests to:" field to point to your public server URL. (Something like `https://{SERVER_URL}/wp-json/`).
7. Make sure your server is configured with live secrets from the `WCPay Dev (Express)` Stripe platform account, since the Apple domain verification does not work in test mode.
8. Since the verification doesn't work with a dev mode account, you'll need to connect to a live mode account, feel free to use mine (Ricardo's): [`acct_1ITcPcR03O9UZoSU`](https://dashboard.stripe.com/connect/accounts/acct_1ITcPcR03O9UZoSU).
    1. Click "Finish setup" from WooCommerce Payments onboarding, and finish the Jetpack authorization.
    2. Don't proceed further with KYC.
    3. From your local server's DB (`wcpay_accounts` table) or WP CLI, connect your blog ID (located in the dev tools plugin) with the `acct_1ITcPcR03O9UZoSU` Stripe account. The account record needs to be set in live mode (`is_live` = `1`).
    4. From the WCPay Dev plugin, clear the account cache to confirm you're successfully connected to `acct_1ITcPcR03O9UZoSU` live account.
9. As soon as you enable WooCommerce Payments, after you refresh the page, you should see the error notice below:
    - Apple Pay domain verification error notice
    - This is because WCPay is passing the value from `get_site_url()`, which has a subfolder in it.
10. Now, with all the information that caused the issue, build the fix from this fork: 
     - `git clone https://github.com/ironicmoka/woocommerce-payments.git && cd woocommerce-payments`
     - `git checkout apple-pay-folder-wordpress-fix`
     - `npm install`
     - and `npm run build:release` to create a new zip for WCPay.
12. Upload the generated zip into the network plugins: **My Sites > Network Admin > Plugins**.
13. When prompted, replace the existing plugin with the new uploaded zip, and then activate WooCommerce Payments.
14. Trigger the domain registration again by disabling express checkouts, saving settings, and then enabling it again.
15. Error notice should be gone and domain should be verified (Apple Pay button should now show up on Safari).


ricardo:
@ismaeldcom Sorry, I'll not be able to review this thoroughly this week as I'm at a local meetup.

ricardo:
Thanks for proposing this fix @ironicmoka 🙌.

Unfortunately this doesn't solve the issue 😞.

`get_bloginfo( 'url' );` should still return the blog's home URL with a subdirectory as per [documentation](https://developer.wordpress.org/reference/functions/get_bloginfo/):

> ‘url’ – The Site address (URL) (set in Settings > General)

Apple requires us to verify the domain itself by hosting the [`.well-known/apple-developer-merchantid-domain-association`](https://github.com/Automattic/woocommerce-payments/blob/develop/apple-developer-merchantid-domain-association) file in the root of the domain, and that's where the problem lies:

If the WP installation is on a subfolder, WooCommerce Payments is not able to serve that file from the root of the domain.

I guess the problem will need to be fixed by:
1. Forcing WooCommerce Payments to pass the domain to Apple, stripping down any paths, ports, etc.
2. Better documentation to allow users to serve the [`apple-developer-merchantid-domain-association`](https://github.com/Automattic/woocommerce-payments/blob/develop/apple-developer-merchantid-domain-association) file from the domain's `.well-known`.

ricardo:
I pushed a few changes so that it always passes down the domain name to Apple, and not the site URL. This should work well for multisites on subdirectories, since the `.well-known/apple-developer-merchantid-domain-association` file should be hosted from the root of the domain, as WooCommerce Payments will also be available at the top-level/main site.

On the other hand, there's not much we can do about single WP installations on subdirectories, so I've added a link to the [documentation](https://woocommerce.com/document/payments/apple-pay/#triggering-domain-registration) on how to host the domain association file from the root of the domain via FTP. This should at least point the user in the right direction.

rafaelzaleski (approved):
LGTM! 🎉

PR #4215, opened 2022-04-29
Ignore cache expiry check when checking platform_checkout eligibility

### LINKED ISSUE #4214: Platform checkout ajax tracking requests sometimes return 404s
### Describe the bug
When tracking ajax request is received, wcpay checks whether platform checkout feature flag is set via `is_platform_checkout_eligible`. This method looks at the flag in Account cache. If the cache has expired `is_platform_checkout_eligible` will return in a falsey value and no attempt is made to refresh the cache. For Ajax requests, cache is expired after [2 hours](https://github.com/Automattic/woocommerce-payments/blob/21183fd912904eb368be1835320239812cdd801b/includes/class-database-cache.php#L276), where as for non-admin, non-ajax requests it will expire after 24 hours. So there's a window from 2hrs to 24hrs where `is_platform_checkout_eligible` will always return `false`. This results in `Platform_Checkout_Tracker` class to not be instentiated for Ajax requests. 
### To Reproduce

1. To make it easier to reproduce, set `$ttl = MINUTE_IN_SECONDS` in this [line](https://github.com/Automattic/woocommerce-payments/blob/1b6827884c53767f43c51057a174c56098f29d99/includes/class-database-cache.php#L281) 
2. As a customer, add a product to the cart
3. Go to checkout page
4. Enter an email address that is already registered to platform checkout
5. Open dev tools network tab and look for `admin-ajax` requests
6. When the OTP iframe opens, notice that the `admin-ajax` request to track `platform_checkout_otp_prompt_start` event returns a 400

### Actual behavior
Ajax event to track platform checkout events returns a 400

### Expected behavior
Tracks requests from ajax events should not fail.

### Additional context
Discussed in https://github.com/Automattic/woocommerce-payments/issues/4109#issuecomment-1113530133

malithsen:
Fixes #4214 

Introduce an optional argument to database cache `get` method to ignore the expiry check. This is utilized when determining the platform checkout eligibility.

#### Testing instructions

1. To make it easier to reproduce, set `$ttl = MINUTE_IN_SECONDS` in this [line](https://github.com/Automattic/woocommerce-payments/blob/1b6827884c53767f43c51057a174c56098f29d99/includes/class-database-cache.php#L281) 
2. As a customer, add a product to the cart
3. Go to the checkout page
4. Enter an email address that is already registered to platform checkout
5. Open dev tools network tab and look for `admin-ajax` requests
6. When the OTP iframe opens, notice that the `admin-ajax` request to track `platform_checkout_otp_prompt_start` returns a 200 response.

-------------------

**Post merge**

### DIFF
--- a/includes/class-database-cache.php
+++ b/includes/class-database-cache.php
@@ -94,10 +94,7 @@ class Database_Cache {
 	public function get( string $key, bool $force = false ) {
 		$cache_contents = get_option( $key );
 		if ( is_array( $cache_contents ) && array_key_exists( 'data', $cache_contents ) ) {
-			if ( true === $force ) {
-				return $cache_contents['data'];
-			}
-			if ( $this->is_expired( $key, $cache_contents ) ) {
+			if ( ! $force && $this->is_expired( $key, $cache_contents ) ) {
 				return null;
 			}
 
--- a/includes/class-database-cache.php
+++ b/includes/class-database-cache.php
@@ -87,12 +87,16 @@ class Database_Cache {
 	 * Gets a value from the cache.
 	 *
 	 * @param string $key The key to look for.
+	 * @param bool   $force if set return from the cache without checking for expiry.
 	 *
 	 * @return mixed The cache contents.
 	 */
-	public function get( string $key ) {
+	public function get( string $key, bool $force = false ) {
 		$cache_contents = get_option( $key );
 		if ( is_array( $cache_contents ) && array_key_exists( 'data', $cache_contents ) ) {
+			if ( true === $force ) {
+				return $cache_contents['data'];
+			}
 			if ( $this->is_expired( $key, $cache_contents ) ) {
 				return null;
 			}
--- a/includes/class-database-cache.php
+++ b/includes/class-database-cache.php
@@ -87,7 +87,7 @@ class Database_Cache {
 	 * Gets a value from the cache.
 	 *
 	 * @param string $key The key to look for.
-	 * @param bool   $force if set return from the cache without checking for expiry.
+	 * @param bool   $force If set, return from the cache without checking for expiry.
 	 *
 	 * @return mixed The cache contents.
 	 */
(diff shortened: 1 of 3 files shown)


includes/class-database-cache.php
```diff
@@ -87,12 +87,16 @@ public function get_or_add( string $key, callable $generator, callable $validate
 	 * Gets a value from the cache.
 	 *
 	 * @param string $key The key to look for.
+	 * @param bool   $force if set return from the cache without checking for expiry.
```
marcinbot, line 90: Super-nitpick, sorry: let's keep the capitalization consistent: `If set, return from the cache without checking for expiry.`

includes/class-database-cache.php
```diff
@@ -87,12 +87,16 @@ public function get_or_add( string $key, callable $generator, callable $validate
...
 	 * @return mixed The cache contents.
 	 */
-	public function get( string $key ) {
+	public function get( string $key, bool $force = false ) {
 		$cache_contents = get_option( $key );
 		if ( is_array( $cache_contents ) && array_key_exists( 'data', $cache_contents ) ) {
+			if ( true === $force ) {
+				return $cache_contents['data'];
+			}
 			if ( $this->is_expired( $key, $cache_contents ) ) {
```
marcinbot, line 100: If we converted the second condition to something like `if ( ! $force && $this->is_expired( $key, $cache_contents ) )` then we could just rely on `return $cache_contents['data']` below. This would reduce the maintenance burden in the future.

marcinbot (approved):
Looks good, two nitpicks, but then feel free to :shipit:

includes/class-database-cache.php
malithsen: Thanks! Adjusted https://github.com/Automattic/woocommerce-payments/pull/4215/commits/1f284c94bdbcb9212d09fb5290d1e2e77bea548c

includes/class-database-cache.php
malithsen: Adjusted in https://github.com/Automattic/woocommerce-payments/pull/4215/commits/bf12e72e3b8cf3fbf53f5920c77648322f813592

brianyu28 (approved):
Thanks, @malithsen, I tested this as well and it looks good.

alefesouza (approved):
LGTM :shipit:

PR #4222, opened 2022-05-04
add docs about updating siteurl with ngrok tunnel

haszari:

This PR adds a line to the dev readme about `siteurl` and `home` options. When using a tunnel (e.g. ngrok), I believe these need to be updated (temporarily) to the tunnel url.

I hit this when regenerating my local environment, so I figured it might block others in future. 

#### Testing instructions

Read through the readme – check if it's helpful and clear 😁

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -41,5 +41,7 @@ ngrok http 8082
 
 You will see it give a forwarding address like this one:
  http://e0747cffd8a3.ngrok.io
+ 
+You may need to set your `siteurl` and `home` `wp_option`s to the new url. You can do this with phpMyAdmin or WP-CLI.
 
 Visit the `<url>` , login and setup WCPay.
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -41,5 +41,7 @@ ngrok http 8082
 
 You will see it give a forwarding address like this one:
  http://e0747cffd8a3.ngrok.io
+ 
+You may need to set your `siteurl` and `home` `wp_option`s to the new url. You can do this with phpMyAdmin or WP-CLI.
```
brucealdridge, line 45: Seems like this may only happen in some circumstances. LTGM.
I think we may need to call out that this should only be a temporary change while setting up WCPay.

```suggestion
You may need to temporarily set your `siteurl` and `home` `wp_option`s to the new url. You can do this with phpMyAdmin or WP-CLI.
```

brucealdridge (approved):

docker/README.md
haszari: Good tip about "temporary", as I understand it the ngrok tunnel is only needed for initial onboarding connection.

haszari:
Merging without checks passing – this is a docs-only change so there's no risk of regressions or flow impact.

PR #4229, opened 2022-05-05
Release branch for 4.1.0

shendy-a8c:
Release branch for WooCommerce Payments 4.1.0 release.


.github/workflows/e2e-test.yml
```diff
@@ -44,6 +44,12 @@ jobs:
         exclude:
           - test_groups: 'blocks'
             test_branches: 'merchant'
+          - woocommerce: '5.2.0'
+            test_groups: 'blocks'
+            test_branches: 'shopper'
+          - woocommerce: '6.1.0'
+            test_groups: 'blocks'
+            test_branches: 'shopper'
```
achyuthajoy, line 52: @shendy-a8c This approach has some problems. This does skip the specified tests. However the setup script still installs and activates WC Blocks on other tests with WC `5.20` & `6.1.0` which will cause problems.

That why I added the conditional workflow script, which skips installation of WC blocks for specified WC versions.

.github/workflows/compatibility.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         woocommerce: [ 'beta' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
-        php:         [ '7.1', '8.0', '8.1' ]
+        php:         [ '7.2', '8.0', '8.1' ]
```
achyuthajoy, line 58: Should we update to PHP `7.4` since it's the least supported version based on https://www.php.net/supported-versions.php

achyuthajoy (changes requested):
@shendy-a8c I've let a few comments regarding changes related to GH actions workflow.

.github/workflows/e2e-test.yml
achyuthajoy: Whoops. My bad. I think I made these changes 😂 to fix the pipeline. 

The idea is to hide the tests are not running from the list.

achyuthajoy (changes requested):
@shendy-a8c Coverage tests still uses WC 6.1.0. This needs to be update to latest - https://github.com/Automattic/woocommerce-payments/blob/d8a0fd5b22daabc1242ef44e9492abb0628bdf1e/.github/workflows/coverage.yml#L14

E2E workflow WC version needs to be updated along with related changes - https://github.com/Automattic/woocommerce-payments/blob/d8a0fd5b22daabc1242ef44e9492abb0628bdf1e/.github/workflows/e2e-test.yml#L39

achyuthajoy:
Version updates to GH actions workflows will be handled on a different PR - p1651748957569029/1651735937.887419-slack-CGGCLBN58

achyuthajoy (approved):
LGTM 🚢

shendy-a8c:
Created https://github.com/Automattic/woocommerce-payments/issues/4230 and https://github.com/Automattic/woocommerce-payments/issues/4231 to address @achyuthajoy's suggestions above.

shendy-a8c:
[6fe502a](https://github.com/Automattic/woocommerce-payments/pull/4229/commits/6fe502acd7d5e17645f3fbc014023afa69e7c3e0) copies `4.1.0` changelog to readme.txt

PR #4232, opened 2022-05-05
Show WooPay info on Success Page

leonardola:
This PR adds WooPay specific info on success page when a customer pays via WooPay. This PR needs to be reviewed alongside with 911-gh-Automattic/woopay and #4233

#### Testing instructions
- Check instructions in 911-gh-Automattic/woopay

-------------------

**Post merge**


woocommerce-payments.php
```diff
@@ -287,3 +287,5 @@ function wcpay_get_jetpack_idc_custom_content(): array {
 
 	return $custom_content;
 }
+
+add_filter( 'jetpack_offline_mode', '__return_false' );
```
brianyu28, line 291: This looks like it may have been left in unintentionally.

includes/admin/class-wc-payments-admin.php
```diff
@@ -836,4 +837,25 @@ public function maybe_redirect_to_onboarding() {
...
 		$this->account->redirect_to_onboarding_page();
 	}
+
+	/**
+	 * Add woopay as a payment method to the edit order on admin.
+	 *
+	 * @param int $order_id order_id.
+	 */
+	public function show_woopay_payment_method_name_admin( $order_id ) {
+		$order = new WC_Order( $order_id );
```
brianyu28, line 847: Here, I'm thinking we may want to use [`wc_get_order`](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-order-functions.html#function_wc_get_order) instead, so that it will return `false` instead of throwing an exception if the order doesn't exist. Then we can also add a check for `! $order` to return early in that case too.

includes/class-wc-payments-order-success-page.php
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+
+	/**
+	 * Add a WooPay logo and card last 4 to the payment method name in the success page.
+	 *
+	 * @param int $order_id the order id.
+	 */
+	public function show_woopay_payment_method_name( $order_id ) {
+		$order = new WC_Order( $order_id );
```
brianyu28, line 46: Here too, I think it's better to use `wc_get_order` and then add a check for if it returns `false` in the conditional below.

includes/admin/class-wc-payments-admin.php
```diff
@@ -836,4 +837,25 @@ public function maybe_redirect_to_onboarding() {
...
+		if ( ! $order->get_meta( 'is_woopay' ) ) {
+			return;
+		}
+		?>
+		<div class="wc-payment-gateway-method-name-woopay-wrapper">
+			<?php echo esc_html_e( 'Paid with ', 'woocommerce-payments' ); ?>
```
brianyu28, line 853: For here and the `'Card ending in '` string below, can we move the trailing whitespace outside of the internationalized string? I think generally it's best for strings that we expect to be translated to not have trailing whitespace that we rely on, since that might not get included in translation.

includes/class-payment-information.php
```diff
@@ -223,6 +223,12 @@ public static function from_payment_request(
 		$token            = self::get_token_from_request( $request );
 		$cvc_confirmation = self::get_cvc_confirmation_from_request( $request );
 
+		if ( isset( $request['is_woopay'] ) && $request['is_woopay'] ) {
+			$order->add_meta_data( 'is_woopay', true );
+			$order->add_meta_data( 'last4', $request['last4'] );
```
brianyu28, line 228: I mentioned this in the other PR reviews, but I'm seeing `last4` appear twice in the order meta for WooPay orders, when I use both PRs' changes on both the WooPay site and the merchant site. Noting that here too since this line may be related. We may also want to check to make sure that `last4` is set before using it to update order metadata.

brianyu28 (changes requested):
Thanks @leonardola! This is working well, just left a few small comments.

includes/admin/class-wc-payments-admin.php
leonardola: Fixed with 6a1f0bb66b58fa48d70cf228c4221554baed3a51

includes/class-wc-payments-order-success-page.php
leonardola: Fixed with bfa9b7a5fea1faed1f54312ff1bcd004032a09db

includes/admin/class-wc-payments-admin.php
leonardola: Fixed with 0b5b5e2383e41972e8e3f3546416d4acef915bea

woocommerce-payments.php
leonardola: Fixed with 46b99d1d18edefed06d571559ae8487e2e1e8bf0

includes/class-payment-information.php
leonardola: I found another solution here. So we don't need to save the last4 in WooPay and then send it to WC Payments. Now we only send the is_woopay flag.

Fixed with 8b9748e6f2b4fee6c30728fdc77ff50f53f0b62e

includes/admin/class-wc-payments-admin.php
```diff
@@ -836,4 +837,25 @@ public function maybe_redirect_to_onboarding() {
...
 	}
+
+	/**
+	 * Add woopay as a payment method to the edit order on admin.
+	 *
+	 * @param int $order_id order_id.
+	 */
+	public function show_woopay_payment_method_name_admin( $order_id ) {
+		$order = wc_get_order( $order_id );
+		if ( $order || ! $order->get_meta( 'is_woopay' ) ) {
```
brianyu28, line 848: Should this `$order` be `! $order`?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1267,6 +1267,11 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 		if ( $payment_needed ) {
 			$payment_method_details = $intent->get_payment_method_details();
 			$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
+
+			if ( $order->get_meta( 'is_woopay' ) ) {
+				$order->add_meta_data( 'last4', $payment_method_details['card']['last4'], true );
```
brianyu28, line 1272: I think for now the WooPay payment method will always be card payment, but should we add a check that `$payment_method_type` is `'card'` before adding the metadata, just in case?

includes/class-wc-payments-order-success-page.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * Class WC_Payments_Order_Service
```
brianyu28, line 3: Should be `WC_Payments_Order_Success_Page`

brianyu28 (changes requested):
Thanks @leonardola! Just a few last comments and then this should be ready to merge.

includes/class-wc-payments-order-success-page.php
leonardola: Fixed with 592c4a815d7be01ae5b0916ebeb3f9a7d50262bd

includes/class-wc-payment-gateway-wcpay.php
leonardola: Fixed with 37e3d23d2f6ace956ea1526021e53779c32ad830

includes/admin/class-wc-payments-admin.php
leonardola: Fixed with 69cd2cce3e84a73031c82fd7094fe1dd38544312

leonardola:
I haven't tested the changes yet as I am not able to checkout after the last git pull. Please don't review the changes yet.

includes/class-wc-payments-order-success-page.php
```diff
@@ -0,0 +1,86 @@
...
+		if ( ! $order || ! $order->get_meta( 'is_woopay' ) ) {
+			return;
+		}
+		?>
+		<ul class="woocommerce-order-overview woocommerce-thankyou-order-details order_details woopay">
+			<li class="woocommerce-order-overview__payment-method method">
+				<?php esc_html_e( 'Payment method:', 'woocommerce-payments' ); ?>
+				<strong>
+					<div class="wc-payment-gateway-method-name-woopay-wrapper">
+						<img src="<?php echo esc_url_raw( plugins_url( 'assets/images/woopay.svg', WCPAY_PLUGIN_FILE ) ); ?>">
```
brianyu28, line 56: Same comment about `alt` text here.

brianyu28 (approved):
Thanks @leonardola, this works well now! I added a couple final comments about adding `alt` attributes to the WooPay `img` tags for accessibility, but looks good to merge after that.

PR #4233, opened 2022-05-05
added card last 4 to order meta

leonardola:
This PR adds the card last 4 to the order to allow WooPay to send it to the merchant store. This PR needs to be reviewed alongside with 911-gh-Automattic/woopay and #4232

#### Testing instructions
- Check instructions in 911-gh-Automattic/woopay

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1267,6 +1267,11 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		if ( $payment_needed ) {
 			$payment_method_details = $intent->get_payment_method_details();
 			$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
+
+			if ( $payment_method_details && 'card' === $payment_method_details['type'] ) {
+				$order->add_meta_data( 'last4', $payment_method_details['card']['last4'] );
+				$order->save_meta_data();
+			}
 		} else {
 			$payment_method_details = false;
 			$payment_method_options = isset( $intent['payment_method_options'] ) ? array_keys( $intent['payment_method_options'] ) : null;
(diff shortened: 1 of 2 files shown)


kalessil:
Out of curiosity @leonardola , is that WooPay specific or other card payment method shall do this as well?

Laso, shouldn't;t wee start documenting the metal intention somewhere @Automattic/transact-architecture

leonardola:
> Out of curiosity @leonardola , is that WooPay specific or other card payment method shall do this as well?

Only for WooPay

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1267,6 +1267,11 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 		if ( $payment_needed ) {
 			$payment_method_details = $intent->get_payment_method_details();
 			$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
+
+			if ( $payment_method_details && 'card' === $payment_method_details['type'] ) {
+				$order->add_meta_data( 'last4', $payment_method_details['card']['last4'] );
```
brianyu28, line 1272: This works fine when I check out this PR only on the WooPay site, but if I check it out on _both_ the WooPay site and on the merchant site, then `last4` ends up appearing twice in the metadata with the same value — I assume this is happening because it's first coming from this payment data and it's also getting added by WCPay on the merchant site. Can we make sure the order meta only gets added once?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1267,6 +1267,11 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
+				$order->save_meta_data();
```
brianyu28, line 1273: Possibly related to the above comment, this adds a `last4` metadata item for all orders paid with credit card, even if the order isn't paid with WooPay. If we only need it for WooPay orders, can/should we avoid adding the metadata for non-WooPay orders?

brianyu28 (changes requested):

PR #4242, opened 2022-05-07
Do not show country code on Platform Checkout opt-in

### LINKED ISSUE #4241: Platform Checkout opt-in shows country code instead of flag only
### Describe the bug
Currently when you write a phone in the billing address, and then opt-in to Platform Checkout, the Platform Checkout phone field selects the country correct but shows the country code in the input field. 

167046098-965890a4-ab17-4984-ac03-d9b3d3d9b0e7

### To Reproduce

1. Enable Platform Checkout.
2. On the checkout page billing address, write a phone number starting with + and country code.
3. Opt-in to Platform Checkout.
4. The country code will show in the opt-in field.

### Expected behavior

Select the correct country code flag, but do not show the code in the input field.

alefesouza:
Fixes #4241

* Prevent show country code text in the Platform Checkout opt-in phone when it's pre-filled, showing only its flag.

#### Testing instructions

1. Enable Platform Checkout.
2. On the checkout page billing address, write a phone number starting with + and country code.
3. Opt-in to Platform Checkout.
4. The country code will not show in the opt-in phone field.

-------------------

**Post merge**

### DIFF
--- a/client/components/platform-checkout/save-user/phone-number-input.js
+++ b/client/components/platform-checkout/save-user/phone-number-input.js
@@ -16,7 +16,6 @@ const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
 		document.getElementById( 'billing_phone' )?.value ?? ''
 	);
 	const [ inputInstance, setInputInstance ] = useState( null );
-	const [ countryCode, setCountryCode ] = useState( null );
 	const [ isValid, setIsValid ] = useState( true );
 
 	const handlePhoneNumberInputChange = ( e ) => {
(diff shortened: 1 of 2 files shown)


client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -123,7 +126,7 @@ const PhoneNumberInput = ( { handlePhoneNumberChange } ) => {
 				) }
 				label={ __( 'Mobile phone number', 'woocommerce-payments' ) }
 				name="platform_checkout_user_phone_field[no-country-code]"
-				value={ inputValue }
+				value={ inputValue.replace( countryCode, '' ) }
```
brianyu28, line 129: Just noting that this `replace` gets called every time the input field changes, and means that something like `+1` will disappear if the user ever types it in the middle of the phone number field — granted, that would make it an invalid phone number, but it might still be a strange user experience for something the user types to just disappear. Should we instead have the `replace` call happen just once when the input field loads for the first time?

brianyu28 (approved):
This works well, thanks @alefesouza — just one question about the expected behavior if the user types a country code manually, but not a blocker if this is the intended behavior.

PR #4258, opened 2022-05-13
Allow merchant to set store logo on Platform Checkout settings

### LINKED ISSUE #4257: Allow merchant to set Platform Checkout logo on settings page
### Describe the bug

Currently we use Stripe account branding logo as the store logo on Platform Checkout, but the merchant can only set it on `Payments > Card Reader > Receipt details`, it would be good to have this options on Platform Checkout settings too.

alefesouza:
Fixes #4257

* Add a new file field on Platform Checkout settings that updates merchant store logo.

#### Testing instructions

* Access WCPay settings.
* On Platform Checkout, click on `Customize`.
* On Store logo section, click `Upload custom logo`.
* After upload finish, click `Save changes`.
* Reload the page, it should be persisted.

-------------------

**Post merge**


client/settings/express-checkout-settings/platform-checkout-settings.js
```diff
@@ -4,20 +4,32 @@
...
+	CheckboxControl,
+	TextControl,
+	BaseControl,
+} from '@wordpress/components';
 
 /**
  * Internal dependencies
  */
 import CardBody from '../card-body';
+import BrandingFileUpload from '../../card-readers/settings/file-upload';
```
malithsen, line 18: This component doesn't look like a generic component. Should we have our own component extending the generic `file-upload` component?

malithsen (commented):
@alefesouza I left a comment on the use of card-reader branding file upload component. Testing that further, it seems like updating the logo from woopay settings also changes the logo in [card reader receipts ](http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fcard-readers) which I don't think is an expected outcome?

One thing I noticed is that the logo introduced here is not the same custom store logo in the design files cO37XZJMyjQweZ6RhEeDhX-fi-31%3A20894. The one in the above figma screen updates the store's logo and this one updates the order thumbnail. This is not an issue, I just wanted to highlight it because I was initially confused thinking they are the same.

> Fixes https://github.com/Automattic/woocommerce-payments/issues/4257

Finally, we might want to make sure that merging this PR does not close the original issue as it involves adding merchant site names to the order page as well. We can also create a separate issue for the store name part.

client/settings/express-checkout-settings/platform-checkout-settings.js
alefesouza: Currently WCPay Server sends the Stripe account branding logo to WooPay, that's the same logo that we set on Card Reader receipts, so I thought there was no problem to use the same component.

alefesouza: Oh I didn't know about this Figma file, I'll change the PR to follow this design.

alefesouza:
@malithsen I've updated this following the design and saving the logo as an option that will be send during platform checkout initialization as occur with Platform Checkout Custom Message instead of using Stripe account that it's used by Card Receipts, I still need to add some front-end tests but I think you can review.

malithsen:
Nice work here! Just a couple of minor UI issues I noticed.
Screen Shot 2022-05-18 at 10 25 12 AM


1. Upload button's description is different from the designs.
2. There's a gray border around the button which is not in the designs

includes/class-wc-payments.php
```diff
@@ -998,7 +998,7 @@ public static function ajax_init_platform_checkout() {
 			'session_cookie_value' => wp_unslash( $_COOKIE[ $session_cookie_name ] ?? '' ), // phpcs:ignore WordPress.Security.ValidatedSanitizedInput
 			'store_data'           => [
 				'store_name'        => get_bloginfo( 'name' ),
-				'store_logo'        => wp_get_attachment_image_src( get_theme_mod( 'custom_logo' ), 'full' )[0] ?? '',
+				'store_logo'        => add_query_arg( 'as_account', '0', get_rest_url( null, 'wc/v3/payments/file/' . self::get_gateway()->get_option( 'platform_checkout_store_logo' ) ) ),
```
alefesouza, line 1001: Now we send full merchant url, this way we can ensure that Plain permalinks work.

alefesouza:
@malithsen fixed, please take another look.

includes/class-wc-payments.php
malithsen, line 1001: We'd want to check the presence of the `platform_checkout_store_logo` and not send the URL if it's not there.

malithsen (changes requested):
Nice improvements again Alefe. Just one more issue to take care,
If I remove the image, WCPay sends a partial URL that doesn't contain the file ID, so the images will appear broken on the Platform Checkout site.

malithsen (approved):
Thanks for the changes Alefe! 
Looks good to me now 👍

PR #4270, opened 2022-05-18
Handle a new partly restricted account status

zmaglica:
#### Handle partly restricted account status

New account status has been introduced that tells the end customer that some payment methods or deposits has been disabled. This PR handles the new status and show correct message to the end user.

Screenshot 2022-05-20 at 17 39 22

#### Testing instructions

- Make sure that you one of the payment methods doesn't have all necessary documents. The best way to achieve this is to submit for new payment method and stop it at the middle of setup.
- On your ~status~ overview page you should see a message that says that some payment methods or deposits has been disabled.

If you'd like to simulate the new status, you can apply this snippet locally:

```diff
diff --git a/includes/class-wc-payments-account.php b/includes/class-wc-payments-account.php
index 196858fb7..a39b50e36 100644
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -186,6 +186,8 @@ class WC_Payments_Account {
                        ];
                }
 
+               $account['status'] = 'restricted_partially';
+
                return [
                        'email'               => $account['email'] ?? '',
                        'country'             => $account['country'] ?? 'US',
```

-------------------

- [x] Covered with tests

### DIFF
--- a/changelog/fix-3783-webhook-idempotency
+++ b/changelog/fix-3783-webhook-idempotency
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Preventing duplicate order notes and emails by clearing the cache before checking order status.
--- a/changelog/fix-store-api-url
+++ b/changelog/fix-store-api-url
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix store api url used by platform checkout to work on different permalink preferences
--- a/changelog/subscriptions-core-2.0.0
+++ b/changelog/subscriptions-core-2.0.0
@@ -1,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Update subscriptions-core to 2.0.0.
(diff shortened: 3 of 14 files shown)


RadoslavGeorgiev (approved):
There was an extra `else if` statement at the end, which is not removed. To be on the safe side, I also added tests to both places where this appears. If you agree with the changes, I've pre-approved the PR, so you can go ahead and merge it. :shipit: 

Note: Now I see that this PR is connected to 2021-gh-Automattic/woocommerce-payments-server, so I assume that I need both to test, however I'm not sure how to simulate `requirements.disabled_reason`. Nevertheless, that is controlled on the server, so it can be altered later if needed.

zmaglica:
> There was an extra `else if` statement at the end, which is not removed. To be on the safe side, I also added tests to both places where this appears. If you agree with the changes, I've pre-approved the PR, so you can go ahead and merge it. :shipit:
> 
> Note: Now I see that this PR is connected to 2021-gh-Automattic/woocommerce-payments-server, so I assume that I need both to test, however I'm not sure how to simulate `requirements.disabled_reason`. Nevertheless, that is controlled on the server, so it can be altered later if needed.

Looks goot to me. I will merge it.

PR #4272, opened 2022-05-19
Add exception handling to prevent fatal errors when fetching payment methods on the checkout block

### LINKED ISSUE #1566: Checkout block see fatal error when account is suspended.
This is because `get_payment_methods_for_customer()` might throw an exception but not caught. here https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L118.

To reproduce:
1. Suspend your account.
2. Open checkout block.

EduardoUmpierre:
Fixes #1566 

Prevent fatal errors when fetching payment methods on the checkout block

The `get_payment_methods_for_customer` method was causing a fatal error to the application, exposing the stack trace directly to the client. It was caused by the absence of exception handling on the `get_payment_methods_for_customer` method as it rethrows the error to the upper level.

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>
Before
</td>
<td>
After
</td>
</tr>
</table>

#### Testing instructions

To be able to reproduce this bug you will need to suspend your account:
* Open your account's RC
* Go to the "WooCommerce Payments" section and search for the "Account tool" subsection
* On the dropdown select "Hard Block Account" and then click on "Apply Account Action"
* Clear your account cache on WCPay Dev

After suspending your account, you will need to publish a page using the Checkout Block:
* Install the "WooCommerce Blocks" plugin in case you don't have it installed yet
* Create a new page accessing "Pages > Add new"
* Type `/` in the editor and search by "checkout"
* Add the "Checkout" block and publish the new page

On the store:
* Add a product to your cart
* Access the page that you have just created
* You should see the following error

Before

In case it doesn't show the error, try to clear the payment methods cache by adding those 2 lines to `includes/class-wc-payments-customer-service.php` at line 212:
```php
	public function get_payment_methods_for_customer( $customer_id, $type = 'card' ) {
		if ( ! $customer_id ) {
			return [];
		}

+		$this->clear_cached_payment_methods_for_user( get_current_user_id() );
+
		$cache_payment_methods = ! WC_Payments::is_network_saved_cards_enabled();
		$cache_key             = Database_Cache::PAYMENT_METHODS_KEY_PREFIX . $customer_id . '_' . $type;
```

-------------------

**Post merge**

- [x] QA Testing Not Applicable as it may be already in the critical flows

### DIFF
--- a/changelog/fix-1566-checkout_block_account_suspended
+++ b/changelog/fix-1566-checkout_block_account_suspended
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add exception handling to prevent fatal error
--- a/changelog/fix-1566-checkout_block_account_suspended
+++ b/changelog/fix-1566-checkout_block_account_suspended
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Add exception handling to prevent fatal error
+Prevent fatal errors when fetching payment methods on the checkout block
--- a/tests/unit/test-class-wc-payments-token-service.php
+++ b/tests/unit/test-class-wc-payments-token-service.php
@@ -328,6 +328,21 @@ class WC_Payments_Token_Service_Test extends WP_UnitTestCase {
 		$this->assertEquals( [ new WC_Payment_Token_CC() ], $result );
 	}
 
+	public function test_woocommerce_get_customer_payment_tokens_failed_to_load_payment_methods_for_customer() {
+		$this->mock_customer_service
+			->expects( $this->once() )
+			->method( 'get_customer_id_by_user_id' )
+			->willReturn( 'cus_12345' );
+
+		$this->mock_customer_service
+			->expects( $this->once() )
+			->method( 'get_payment_methods_for_customer' )
+			->willThrowException( new Exception( 'Failed to get payment methods.' ) );
+
+		$result = $this->token_service->woocommerce_get_customer_payment_tokens( [ new WC_Payment_Token_CC() ], 1, 'woocommerce_payments' );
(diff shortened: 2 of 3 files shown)


changelog/fix-1566-checkout_block_account_suspended
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add exception handling to prevent fatal error
```
luizreis, line 4: This changelog is added to the public changelog of WC Pay, so what do you think of including more details about the issue here? IMO something along the lines of "Prevent fatal errors when fetching payment methods on checkout block" would be a bit more informative (feel free to reword it 😄).

luizreis: Similar to the comment above, I think the PR name could use some more details about what error is being fixed.

tests/unit/test-class-wc-payments-token-service.php
```diff
@@ -328,6 +328,21 @@ public function test_woocommerce_get_customer_payment_tokens_no_customer() {
 		$this->assertEquals( [ new WC_Payment_Token_CC() ], $result );
 	}
 
+	public function test_woocommerce_get_customer_payment_tokens_failed_to_load_payment_methods_for_customer() {
```
luizreis, line 331: Thanks for adding a test! 🎉

luizreis (approved):
Thanks for working on this @EduardoUmpierre! The code looks good and it fixes the issue.

I left a small comment about the changelog entry and the PR name, but I'm pre-approving the PR. Feel free to :shipit: once the comment is addressed.

PR #4275, opened 2022-05-19
Add support for (custom) order numbers in API calls

### LINKED ISSUE #4142: Suggestion: Improve compatibility with plugins replacing order numbers
### Describe the bug

Currently, we pass the order's `post_id` to Stripe along with the transactions. 

Eg: If the order number is `#2033` and the `post_id` of that order is `1012` then on our Stripe dashboard, it will display that the transaction was for `order #1012`.

This makes debugging harder for HEs to find the transaction along with the correct order number. 

Can we pass the order number itself to Stripe to improve compatibility with plugins replacing order numbers?

Discussion: p1649147095821479-slack-C7U3Y3VMY

RadoslavGeorgiev:
Fixes #4142

WooCommerce Orders have IDs, which are the same as their post ID, and order numbers, which by default are also the same, but can be customized through extensions. Those are retrieved through `$order->get_id()` and `$order->get_order_number()` respectively.

Within WCPay we seem to use `get_order_number` everywhere within the site, but when we're sending information to Stripe, we only included the order/post ID. **This PR** adds the number to the metadata we're sending to Stripe, and updates the description of the payment to use the order number instead of ID.

Here's the description (with this PR):

image
and the updated metadata:
Screenshot 2022-05-19 at 15 12 47

This was done, implemented, and manually tested in the following scenarios:

- UPE standard card
- UPE SCA card
- UPE redirect
- classic checkout
- express checkout

#### Testing instructions

0. Implement a manual filter, or setup an extension like [Sequential Order Number for WooCommerce](https://wordpress.org/plugins/wt-woocommerce-sequential-order-numbers/). This way you'll be able to see the difference between IDs and numbers, otherwise they would be the same.
1. Make payments in various ways (the ones listed above, and any others if you think of them)
2. Check the payments in Stripe's dashboard, and make sure that the metadata includes both order ID and number, and that the description uses the number.

-------------------

- [x] Covered with tests


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2313,17 +2313,17 @@ private function deserialize_intention_object_from_array( array $intention_array
 	/**
 	 * Returns a formatted intention description.
 	 *
-	 * @param  int $order_id The order ID.
-	 * @return string        A formatted intention description.
+	 * @param  string $order_number The order number (might be different from the ID).
+	 * @return string               A formatted intention description.
 	 */
-	private function get_intent_description( int $order_id ): string {
+	private function get_intent_description( $order_number ): string {
```
zmaglica, line 2319: I think that we should include order ID if the order number is not set to maintain backward compatibility. For the orders that are created before this PR, the order number will not exist, so my suggestion is to refactor a code a bit on line 223 and 259 to include something like this: `$metadata['order_number'] ?? $metadata['order_id'] ?? 0`

RadoslavGeorgiev: @zmaglica do you think that we need a fallback here? AFAICT the method will only be called for new orders, which need a description for their intent. FWIW I've checked all code paths I could think of, and the only place where we would not call `$order->get_order_number()` in advance is [here](https://github.com/Automattic/woocommerce-payments/pull/4275/files#diff-b541c0a287f40f82a1d0e43c52c7c7a2b67ae233f261655cbb437726fec83feeR276), and for that situation, there is a fallback to zero, in case the order does not exist.

zmaglica: Makes sense. I will resolve this comment.

zmaglica (approved):
Tested and works perfectly.

PR #4276, opened 2022-05-19
Add a check on credit card being enabled to determine whether WCPay gateway is available for checkout

### LINKED ISSUE #3969: Unchecking Credit card / debit card in Payments accepted on checkout does not remove it at checkout
### Describe the bug
When Credit card / debit card in Payments accepted on checkout is unchecked and Express checkouts is enabled, the option to pay with a card still appears to the customer. 

### To Reproduce

1. Go to Payments > Settings
2. Uncheck the option to display credit/debit cards. 
3. You'll see a warning message that customers will no longer be able to use this method for payment. 
4. Visit checkout
5. The credit card fields are still visible and can be used for payment. 

Tested with and without UPE. 

### Actual behavior

Some merchants only want to use WooCommerce Payments for Express Checkout features. While Credit card / debit card can be deselected, this doesn't seem to have any impact on availability at checkout. 

### Screenshots

Screenshot of warning when the payment method is removed: 
Screenshot on 2022-03-15 at 15-37-20

Screenshot of the checkout in Chrome, Express is available but credit card is also. 
Markup 2022-03-15 at 15 39 01

### Expected behavior

Since credit/debit has been disabled, we expect that it would not appear. 
If it cannot be fully disabled, I would expect the checkbox to not appear for that payment method. 

* OS: [e.g. iOS] Mac OS 12.2.1
* Browser [e.g. chrome, safari] Chrome 
* Version [e.g. 22] Version 99.0.4844.51

### Additional context

Brought up in 4862961-zen 
Previously discussed in https://github.com/Automattic/woocommerce-payments/issues/2560#issuecomment-896222148
And: https://wordpress.org/support/topic/only-apple-google-pay-on-woocommerce/

jbordonado:
Fixes #3969 

#### Testing instructions

Classic checkout & Blocks checkout

- Visit the store, add a product to the cart and go to the checkout page: you should see the credit card payment method fields
- Go to Payment > Settings and disable the payment method `Credit card / debit card`
- Visit the store, add a product to the cart and go to the checkout page: you should not see the credit card payment method fields
- Go back to Payment > Settings and click on `Enable in your store` to enable UPE (add the `Bancontact` method)
- Switch to the General tab and set the currency of your store to US dollar
- Visit the store, add a product to the cart and go to the checkout page: you should not see the credit card payment method fields and you should also not see the UPE method you added earlier (since it is not compatible with the store's currency)
- Now, go back to wp-admin and change the currency of your store to EUR
- Visit the store, add a product to the cart and go to the checkout page: you should not see the credit card payment method fields BUT now you should see the `Bancontact` UPE method option. You should be able to place an order using whatever payment method is available to you (while credit card is not).
- Just to be sure, if you now enable again  the payment method `Credit card / debit card` in Payments > Settings, visit the store, add a product to the cart and go to the checkout page: you should see the credit card option along with the UPE one.

-------------------

**Post merge**


includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -919,7 +919,7 @@ function() use ( $payment_fields ) {
 	 */
 	public function get_payment_method_ids_enabled_at_checkout( $order_id = null, $force_currency_check = false ) {
 		$capture                    = empty( $this->get_option( 'manual_capture' ) ) || $this->get_option( 'manual_capture' ) === 'no';
-		$capturable_payment_methods = $capture ? $this->get_upe_enabled_payment_method_ids() : [ 'card' ];
+		$capturable_payment_methods = $capture ? $this->get_upe_enabled_payment_method_ids() : [];
```
htdat, line 922: Changing to an empty array seems not appropriate. We may filter and return `card` if `$this->get_upe_enabled_payment_method_ids()` array has `card` value.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -537,7 +537,7 @@ public function test_create_payment_intent_with_manual_capture() {
 			->with(
 				5000,
 				'usd',
-				[ 'card' ],
+				[],
```
htdat, line 540: Consider reverting this test, or adding more test case.

htdat (commented):
I am leaving some initial reviews. 

Besides inline comments, But in the meantime, I am seeing an issue in this case: UPE enabled but with only `sofort` and `giropay`. - `card` is disabled. Something like this: 

- ✅ Traditional checkout: only `sofort` and `giropay` display => expected.
- ❌ Block checkout: no payment method. 

Markup 2022-05-24 at 14 54 28

includes/admin/class-wc-rest-upe-flag-toggle-controller.php
```diff
@@ -125,13 +127,19 @@ private function update_is_upe_enabled( WP_REST_Request $request ) {
 		update_option( WC_Payments_Features::UPE_FLAG_NAME, 'disabled' );
 
 		// resetting a few other things:
-		// removing the UPE payment methods and _only_ leaving "card",
+		// removing the UPE payment methods and _only_ leaving the rest,
```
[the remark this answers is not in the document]
jbordonado: Done

includes/payment-methods/class-upe-payment-gateway.php
jbordonado: Yes, that was not correct.

I introduced a new function `WC_Payment_Gateway_WCPay->get_default_upe_enabled_payment_method_ids()` with the logic suggested above (in_array), so that it can be reused in the three places where it's needed.

includes/admin/class-wc-rest-upe-flag-toggle-controller.php
```diff
@@ -125,13 +127,19 @@ private function update_is_upe_enabled( WP_REST_Request $request ) {
...
+			function ( $method_id ) {
+				return Payment_Method::CARD === $method_id;
+			}
+		);
 		$this->wcpay_gateway->update_option(
 			'upe_enabled_payment_method_ids',
-			[
-				'card',
-			]
+			$new_enabled_payment_method_ids
```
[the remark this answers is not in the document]
jbordonado: Done, used this logic in the new `WC_Payment_Gateway_WCPay->get_default_upe_enabled_payment_method_ids()`

jbordonado:
> I am leaving some initial reviews.
> 
> Besides inline comments, But in the meantime, I am seeing an issue in this case: UPE enabled but with only `sofort` and `giropay`. - `card` is disabled. Something like this:
> 
> * ✅ Traditional checkout: only `sofort` and `giropay` display => expected.
> * ❌ Block checkout: no payment method.
> 
> Markup 2022-05-24 at 14 54 28

For now, I have a hard time reproducing your case.
Classic checkout and Blocks checkout both display `sofort` and `giropay` on my end, in that scenario.
We'll need further debugging on your test site to understand what's going on..

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2720,6 +2721,15 @@ public function enable() {
 		$this->update_option( 'enabled', 'yes' );
 	}
 
+	/**
+	 * Returns the default list of enabled payment method types for UPE.
+	 *
+	 * @return string[]
+	 */
+	public function get_default_upe_enabled_payment_method_ids(): array {
```
htdat, line 2729: IMO. The `default` name is not really correct, and it's confusing when reading other functions using this method. 

The better name may be exactly what it does, something like: `filter_card_payment_method_id_enabled`? 

Also, we may narrow down the scope of this function to `protected` as it's being used in the inheritance class `UPE_Payment_Gateway` too but nowhere else. 

With the current confusing way we're currently constructing these classes: `CC_Payment_Gateway` (empty code inside) and `UPE_Payment_Gateway` are child classes of `WC_Payment_Gateway_WCPay`, I think we may want to make this function `final` too, i.e. not allowing to modify the function in `UPE_Payment_Gateway`.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
htdat: Resolved! Good tests anyway.

includes/admin/class-wc-rest-upe-flag-toggle-controller.php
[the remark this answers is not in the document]
htdat: Good idea to introduce this function for sure. However, I am having some comments regarding it. See my next inline comment.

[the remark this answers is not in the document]
htdat: Resolved this comment now.

htdat:
> For now, I have a hard time reproducing your case.
> Classic checkout and Blocks checkout both display sofort and giropay on my end, in that scenario.
> We'll need further debugging on your test site to understand what's going on.

As mentioned here p1653898213089389/1653896075.267169-slack-C01BPL3ALGP, this can be replicated by using the following steps: 

1. In WP Admin > WooCommerce > Settings, choose `USD` as the default currency.
2. Add `EUR` as one of currencies in the multi-currency feature. 
3. In the block checkout, choose `EUR` then the issue happens. 

Other notes: 

- If I switch to `develop` branch, this issuer no longer happens. 
- From my debugging, it's coming from this validation: 

https://github.com/Automattic/woocommerce-payments/blob/7a26eafb8251bb4414810d77b009d5c0d5daff8d/includes/payment-methods/class-upe-payment-method.php#L128-L130

- Normal checkout: `get_woocommerce_currency()`returns `EUR`, and this `is_currency_valid` returns `true`.
- Block checkout: `get_woocommerce_currency()`returns `USD`, and this `is_currency_valid` returns `false`.

Here is the place MultiCurrency feature modifies the value of `get_woocommerce_currency()`: 
https://github.com/Automattic/woocommerce-payments/blob/7a26eafb8251bb4414810d77b009d5c0d5daff8d/includes/multi-currency/FrontendCurrencies.php#L105

jessy-p:
**Issue faced with Multicurrency & payment methods when card is disabled.**

**Testing Instructions**
  - Set the default currency as GBP/USD
  - Disable card and enable giropay in Payments > Settings
  - Set the currency via MultiCurrency to EUR

**Expected Behavior:** Giropay should be displayed.

**Current Behavior:** Giropay is not displayed.
 
**More details:** The payment method script, in this case, `upe-blocks-checkout.js` is not included on the page. `wc.settings` contains Giropay, which is correct.

**Reason:** Multicurrency sets get_woocommerce_currency to correct value 'EUR' AFTER the block scripts are enqueued. 

**Fix:** [Change the priority of init action in Multicurrency.php Line 217](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/MultiCurrency.php#L217)
`add_action( 'init', [ $this, 'init' ] , 9);`. works.

**Note:** Issue was not seen earlier because card was always enabled, even when disabled from Settings. As the card payment method is not dependent on currency, the block would always be displayed and settings being correct, giropay was correctly shown.

jbordonado:
I tested again the scenario with Blocks and it seemed to work well now, just like classic checkout ✅

includes/class-wc-payment-gateway-wcpay.php
jessy-p: Renamed to be less confusing.

zmaglica:
I pulled the latest code and when I try to remove Credit Card as a payment option, I am getting this error:

```
{
   "code":"rest_invalid_param",
   "message":"Invalid parameter(s): enabled_payment_method_ids",
   "data":{
      "status":400,
      "params":{
         "enabled_payment_method_ids":"enabled_payment_method_ids[0] is not one of card, bancontact, eps, giropay, ideal, sofort, p24, and link."
      },
      "details":{
         "enabled_payment_method_ids":{
            "code":"rest_not_in_enum",
            "message":"enabled_payment_method_ids[0] is not one of card, bancontact, eps, giropay, ideal, sofort, p24, and link.",
            "data":null
         }
      }
   }
}
```

I even tried to reonboard and the same error is presented, which makes it untestable for me. Has anyone else had the same experience?

jessy-p:
Copying from [Slack thread](https://a8c.slack.com/archives/C01BPL3ALGP/p1654492928190699),
1) @zmaglica  Error above is not replicable
2) @deepakpathania  Error is replicable. Steps:
- enable UPE methods only, add a product and visit checkout
- enable credit card, and go directly to checkout (without adding a new product)

That way, I see my saved credit card, but there’s no credit card option in the list of new payment method (only the UPE ones)
If you now add a product, then go back to Checkout, everything is shown as expected, including the credit card option
Using that same “behaviour”, if you :
- enable CC + UPE method, add a product and visit checkout
- disable CC and visit checkout directly
I see this JS error:
```
?ver=3.0:1 Uncaught TypeError: Cannot read properties of undefined (reading 'isReusable')
    at upe.js:352:52
```
I believe this is related to the payment intent and the fact that without adding a new product (hence a new price in the cart), the intent in session will be reused and the payment methods are shown based on the intent’s response.

jessy-p:
Some more flows that need to be tested for this use-case:

- Subscriptions’s behaviour in case of renewal with a disabled card.
- Adding a new card as payment method with existing items in cart while admin toggles this check.
- Stripe’s behaviour in case of a card saved there but disabled from our interface.

Question: Should we allow cards to be disabled at all ? If yes, we should test and solve all above issues. If no, we can implement disabling the ability to uncheck cards.

To be discussed on P2.

jessy-p:
paJDYF-4kY-p2

jessy-p:
Based on direction provided on paJDYF-4kY-p2, will be listing out the various scenarios to be tested and fixed on top of this PR.

deepakpathania:
@jessy-p to create a PRD for the issues.

PR #4281, opened 2022-05-24
Switch to `@woocommerce/dependency-extraction-webpack-plugin`

### LINKED ISSUE #3517: Switch to using `@woocommerce/dependency-extraction-webpack-plugin`
### Description
We currently use the `@wordpress/dependency-extraction-webpack-plugin` to declare the packages we want to use globally versions of and omit from our built JavaScript bundle. There is a WooCommerce version of this package available (`@woocommerce/dependency-extraction-webpack-plugin`) which we could be using instead.

The steps for doing this would be something like:
* Remove the WordPress package and add the WooCommerce one.
* Remove any `woocommerce/*` configuration entries in the webpack plugin configuration.
* Ensure the built package still has the same packages in it.
* (Spoiler) Having tried this, the built package is a lot smaller, so we'll need to figure out what we were bundling that we weren't intending to bundle.
* Once completed update the [documentation](https://github.com/Automattic/woocommerce-payments/blob/develop/docs/renovate.md).

### Testing instructions
If we get the build to behave as before then the required testing is limited. If we find packages are now being removed from the bundle we'll want to make sure that hasn't had any impact.

ismaeldcom:
Fixes #3517

There were some issues while using WC 5.2.0, but now that we require at least 6.8, it's safe to proceed with this. 

- Replace `@wordpress/dependency-extraction-webpack-plugin` with `@woocommerce/dependency-extraction-webpack-plugin`.
- Update `webpack.config.js` to remove all entries of `@woocommerce/*` packages.
- Update renovate documentation.
- Remove `onboarding` entry from webpack, as it is not used anymore.

Analyzing both bundles, the main difference (as expected) are included `@woocommerce/*` packages, in the following screenshot we can see all the included packages in our current build. With this PR none are included. 

Screenshot 2022-05-24 at 10 26 36

#### Testing instructions
I didn't see any visual difference or console error testing main functionalities. We're just unbundling some imported WC packages that were not added to the dependency extraction plugin.

Smoke test functionalities using `@woocommerce/*` packages (components, currency,...) deposits/transactions and disputes are good ones. Will be good to include it early in our next release to have it deeply tested by GS.

-------------------

**Post merge**
- General QA testing.


oaratovskyi (approved):
Great work, @ismaeldcom !

I've tested basic stuff like:
- cart + checkout regular payment (EUR, USD)
- multi-currency setup
- dispute payment (GBP)
- view transactions
- view disputes

It works as expected and feels like it works even faster 👍 
Let's 🚢 it after GS tests in on Tuesday!

webpack/shared.js
```diff
@@ -111,20 +110,12 @@ module.exports = {
...
 		new MiniCssExtractPlugin( { filename: '[name].css' } ),
-		new WordPressExternalDependenciesPlugin( {
+		new WooCommerceDependencyExtractionWebpackPlugin( {
 			injectPolyfill: true,
 			requestToExternal( request ) {
 				switch ( request ) {
 					case '@wordpress/components':
 						return null;
-					case '@woocommerce/components':
-						return [ 'wc', 'components' ];
```
oaratovskyi, line 121: yay!

PR #4284, opened 2022-05-24
Revert "Fix webhook idempotency (for payment intents) (#4278)"

reykjalin:
This reverts commit 92265b288a4607acb4105c1c1074c6062f2addc3.

Fixes #

The changes applied in #4278 broke checkout in WooPay because it changed the order of several key operations required for WooPay to work.

In WooPay we rely on order information being up to date and complete when the `woocommerce_payment_complete` hook is fired so we can accurately process payments. However, the change introduced in #4278 made it so that the `woocommerce_payment_complete` hook is fired _before_ the necessary order meta has been attached to the order which causes all WooPay checkouts to fail.

#### Testing instructions

Before applying this change WooPay checkout is broken.
After applying the change WooPay checkout should work.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-3783-webhook-idempotency
+++ b/changelog/fix-3783-webhook-idempotency
@@ -1,4 +1,0 @@
-Significance: patch
-Type: fix
-
-Preventing duplicate order notes and emails by clearing the cache before checking order status.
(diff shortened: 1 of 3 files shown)


naman03malhotra (approved):
LGTM!

PR #3628, opened 2022-01-12
Handle payment_failed payment intent webhooks.

jessepearson:
Fixes #3620

#### Changes proposed in this Pull Request

* Add handling for `payment_intent.payment_failed` webhooks.
  * Add `process_webhook_payment_intent_failed` method that updates the order to the _Failed_ status and adds the reason why.
* Add `get_order_from_event_body_intent_id` method to get the `WC_Order` object from the payment intent id.
  *  Refactor `process_webhook_payment_intent_succeeded` to use the new method.
* Add `get_failure_message_from_event` method to get the `failure_code` from the webhook event, then provide a translated error message from the code.
  * Used in `process_webhook_payment_intent_failed` to add the note to the order
* Add test for `process_webhook_payment_intent_failed` and remove unused stub tests.

#### Testing instructions

* Tests should pass.

I found the easiest way to manually test is to use Postman, unless you have another method for sending in requests to the REST API.

* Use filter below to get around authentication, be sure to remove it afterwards.
* Set to POST to `https://YOURSITE/?rest_route=/wc/v3/payments/webhook`.
* Use the json below as the body of the post, updating where needed.
  * It's best to use an existing order, and get the payment intent id out of it, then add it to the json data.
  * Set the order to On Hold, Pending, etc.
* After posting, the order should be moved to Failed status, and there should be a note in the order that the account was not found.

filter:
```
add_filter( 'user_has_cap', function( $caps ) {
	$caps['manage_woocommerce'] = true;
	return $caps;
});
```

json:
```
{
    "type": "payment_intent.payment_failed",
    "data": {
        "object": {
            "id": "ADD_INTENT_ID_HERE",
            "object": "payment_intent",
            "charges": {
                "data": [
                    {
                        "failure_code": "no_account"
                    }
                ]
            }
        }
    }
}
```

-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _ 'QA Testing Not Applicable'_

PR #4287, opened 2022-05-24
Fix fatal error by checking for valid order before processing webhook

### LINKED ISSUE #4286: Fatal error when webhook processing service cannot find order
### Describe the bug

Please see p1653399315701929-slack-C7U3Y3VMY for more information. 

Complete error stack trace:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_meta() on bool in /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-webhook-processing-service.php:365
Stack trace:
#0 /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-webhook-processing-service.php(164): WC_Payments_Webhook_Processing_Service->process_webhook_payment_intent_succeeded(Array)
#1 /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/admin/class-wc-rest-payments-webhook-controller.php(80): WC_Payments_Webhook_Processing_Service->process(Array)
#2 /wordpress/core/6.0/wp-includes/rest-api/class-wp-rest-server.php(1143): WC_REST_Payments_Webhook_Controller->handle_webhook(Object(WP_REST_Request))
#3 /wordpress/core/6.0/wp-includes/rest-api/class-wp-rest-server.php(990): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/wc/v3/payments...', Array, NULL)
#4 /wordpress/core/6.0/wp-includes/rest-api/class-wp-rest-server.php(414): WP_REST_Server->dispatch(Object(WP_REST_Request))
#5 /wordpress/core/6.0/wp-includes/rest-api.php(394): WP_REST_Server->serve_request('/wc/v3/payments...')
#6 /wordpress/core/6.0/wp-includes/class-wp-hook.php(307): rest_api_loaded(Object(WP))
#7 /wordpress/core/6.0/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#8 /wordpress/core/6.0/wp-includes/plugin.php(524): WP_Hook->do_action(Array)
#9 /wordpress/core/6.0/wp-includes/class-wp.php(398): do_action_ref_array('parse_request', Array)
#10 /wordpress/core/6.0/wp-includes/class-wp.php(770): WP->parse_request('')
#11 /wordpress/core/6.0/wp-includes/functions.php(1330): WP->main('')
#12 /wordpress/core/6.0/wp-blog-header.php(16): wp()
#13 /wordpress/core/6.0/index.php(17): require('/wordpress/core...')
#14 {main}
  thrown in /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-webhook-processing-service.php on line 365
```

The above error was spotted causing fatal errors on Atomic sites. I'm not sure how to specifically reproduce it, but I can see how it would be caused. On [this line](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-webhook-processing-service.php#L365), if the function `get_order_from_event_body_intent_id` is unable to retrieve an order and returns `false` to `$order`, this will cause the call to `get_meta` upon this object to return the above error.

FangedParakeet:
Fixes #4286 

# Fixing Fatal Errors From Fractured Webhooks

Please read the details within #4286 for more context on the need for this fix. 

If the event body of the `process_webhook_payment_intent_succeeded` webhook event is not associated with a valid order ID for whatever reason, this fix will prevent a fatal error. All other functions within `WC_Payments_Webhook_Processing_Service` already have similar safety nets to handle mysteriously indecipherable orders and with these changes the `process_webhook_payment_intent_succeeded` handler will fall in line by throwing the same `Invalid_Webhook_Data_Exception` if an order is not to be found.

#### Testing instructions

1. Please pull the changes in this PR.
2. We will need to trigger the webhooks REST endpoint with an event type of `payment_intent.succeeded`. I'm not sure if there is a better way to do this, but I managed it by making a `POST` request to the endpoint using Postman. The target endpoint should be `/wp-json/wc/v3/payments/webhook`. Here is an example of a test request body. All of these fields will be required and the object `id` (here `pi_123`) must not match anything in your orders table.
```json
{
  "id": "evt_123",
  "object": "event",
  "data": {
    "object": {
      "id": "pi_123",
      "object": "payment_intent",
      "charge": "ch_123",
      "status": "processing",
      "invoice": "inv_123",
      "metadata": [],
      "charges": {
          "data": [
              {
                  "id": "ch_123"
              }
          ]
      }
    }
  },
  "type": "payment_intent.succeeded"
```
3. If we make the above request without these changes, the response should be an `internal_server_error` that states that there has been a critical error on the website. However, with the changes we should now be greeted with a marginally friendlier `400` `bad_request` error instead. 
4. Double-check that we threw the correct Exception by checking the WCPay status/error logs. There should be an error printed similar to the following: "ERROR WCPay\Exceptions\Invalid_Webhook_Data_Exception: Could not find order via charge ID: ch_123 in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-webhook-processing-service.php:365".

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-webhook-processing-service.php
+++ b/includes/class-wc-payments-webhook-processing-service.php
@@ -361,6 +361,15 @@ class WC_Payments_Webhook_Processing_Service {
 		$charges_data  = $this->read_webhook_property( $event_charges, 'data' );
 		$charge_id     = $this->read_webhook_property( $charges_data[0], 'id' );
 
+		if ( ! $order ) {
+			throw new Invalid_Webhook_Data_Exception(
+				sprintf(
+				/* translators: %1: charge ID */
+					__( 'Could not find order via charge ID: %1$s', 'woocommerce-payments' ),
+					$charge_id
+				)
+			);
+		}
 		// update _charge_id meta if it doesn't exist - happens when maybe_process_upe_redirect fails sometimes.
 		if ( $charge_id && ! $order->get_meta( '_charge_id' ) ) {
 			$order->update_meta_data( '_charge_id', $charge_id );
--- a/includes/class-wc-payments-webhook-processing-service.php
+++ b/includes/class-wc-payments-webhook-processing-service.php
@@ -362,13 +362,7 @@ class WC_Payments_Webhook_Processing_Service {
 		$charge_id     = $this->read_webhook_property( $charges_data[0], 'id' );
 
 		if ( ! $order ) {
-			throw new Invalid_Webhook_Data_Exception(
-				sprintf(
-				/* translators: %1: charge ID */
-					__( 'Could not find order via charge ID: %1$s', 'woocommerce-payments' ),
-					$charge_id
-				)
-			);
+			return;
 		}
 		// update _charge_id meta if it doesn't exist - happens when maybe_process_upe_redirect fails sometimes.
 		if ( $charge_id && ! $order->get_meta( '_charge_id' ) ) {
--- a/changelog/fix-4286-fix-webhook-service-fatal-error
+++ b/changelog/fix-4286-fix-webhook-service-fatal-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixes fatal error on payment intent succeeded webhook.


includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -361,6 +361,15 @@ private function process_webhook_payment_intent_succeeded( $event_body ) {
...
 
+		if ( ! $order ) {
+			throw new Invalid_Webhook_Data_Exception(
+				sprintf(
+				/* translators: %1: charge ID */
+					__( 'Could not find order via charge ID: %1$s', 'woocommerce-payments' ),
+					$charge_id
+				)
+			);
+		}
```
vbelolapotkov, line 372: Looking into code it looks like the bug is caused by [this line](https://github.com/Automattic/woocommerce-payments/blob/5de214390589f707a8b0dc7472c680d06f8892c3/includes/class-wc-payments-webhook-processing-service.php#L577), according to it missing order is valid case and execution should continue, rather than throw exception. However, this change stops execution and throws the exception.

@htdat looks like you were the author of `get_order_from_event_body_intent_id` ([code](https://github.com/Automattic/woocommerce-payments/blob/5de214390589f707a8b0dc7472c680d06f8892c3/includes/class-wc-payments-webhook-processing-service.php#L559)), could you please clarify how the code is supposed to behave in fallback case.

Side note, this the second time when I see that the fatals are caused by mixed return type from the method.

ismaeldcom: I think the bug comes from [this commit](https://github.com/Automattic/woocommerce-payments/commit/eb591d24146989485ce9ec65151e3f2e7bb4df95) where `$order` is used without checking it.

`get_order_from_event_body_intent_id` [was created here](https://github.com/Automattic/woocommerce-payments/commit/d619c4d5464899ba36cb187f8cfc7eebe7615238), but it just contains the code used in `process_webhook_payment_intent_succeeded` and was created to avoid duplicated code.

And as @vbelolapotkov mentions, we should not throw there, because that will prevent [`invoice.paid`](https://github.com/Automattic/woocommerce-payments/blob/5de214390589f707a8b0dc7472c680d06f8892c3/includes/class-wc-payments-webhook-processing-service.php#L169) to run.

Probably the simpler approach is to just check `$order` before using it in  `process_webhook_payment_intent_succeeded`. Although I don't know if that could bring any other issue. There are some new additions since I reviewed [the failed payments PR](https://github.com/Automattic/woocommerce-payments/pull/3628).

zmaglica: Without Jessy PR, the code would break on a couple of lines bellows. I will create a PR that will resolve this problem. @ismaeldcom I think that the simplest way is to use your approach.

ismaeldcom: Yes! `process_webhook_payment_intent_succeeded` was way shorter before, there were no additional uses of `$order`, but now we should wrap all of them within it or do an early return.

https://github.com/Automattic/woocommerce-payments/blob/d619c4d5464899ba36cb187f8cfc7eebe7615238/includes/admin/class-wc-rest-payments-webhook-controller.php#L287-L294

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -361,6 +361,15 @@ private function process_webhook_payment_intent_succeeded( $event_body ) {
 		$charges_data  = $this->read_webhook_property( $event_charges, 'data' );
 		$charge_id     = $this->read_webhook_property( $charges_data[0], 'id' );
 
+		if ( ! $order ) {
```
zmaglica, line 364: I would move order fetching at the begging of the function, and use this if statement right after it to prevent getting all this data for nothing if the order is false

includes/class-wc-payments-webhook-processing-service.php
zmaglica: > Side note, this the second time when I see that the fatals are caused by mixed return type from the method.

Either you return `false/null` or throw exception in this case. One of the problems is that I failed to see, as a reviewer, that this function can return multiple types. This should be covered with tests where every case that can occur during the execution of the function can occur.

htdat: > I think the bug comes from [this commit](https://github.com/Automattic/woocommerce-payments/commit/eb591d24146989485ce9ec65151e3f2e7bb4df95) where `$order` is used without checking it.

Agreed this point! 

>`get_order_from_event_body_intent_id` [was created here](https://github.com/Automattic/woocommerce-payments/commit/d619c4d5464899ba36cb187f8cfc7eebe7615238), but it just contains the code used in `process_webhook_payment_intent_succeeded` and was created to avoid duplicated code.

But looking further, the side effect coming from this commit as it introduces `false` for this condition branch: 
https://github.com/Automattic/woocommerce-payments/blob/d619c4d5464899ba36cb187f8cfc7eebe7615238/includes/admin/class-wc-rest-payments-webhook-controller.php#L485-L488

The original commit [ref](https://github.com/Automattic/woocommerce-payments/commit/89766ab37e0b51af22497d8f14825a8b176528da) adding this condition means to return early and ignore this webhook:

https://github.com/Automattic/woocommerce-payments/blob/89766ab37e0b51af22497d8f14825a8b176528da/includes/admin/class-wc-rest-payments-webhook-controller.php#L273-L276

htdat: > Probably the simpler approach is to just check `$order` before using it in `process_webhook_payment_intent_succeeded`.

Also agreed with this! And more specifically, we may want to return early for the `! $order` condition like the way we're handling in `process_webhook_payment_intent_failed`. 

```suggestion
		if ( ! $order ) {
			return;
		}
```

ismaeldcom: >But looking further, the side effect coming from this commit as it introduces false for this condition branch:

I prefer null over false for those cases. But in the end it will throw an error if we don't check if `$order` is valid, right? 🤔

htdat: @ismaeldcom - yes, returning `null` is ideal. But in this specific case, it just introduces another type for `get_order_from_event_body_intent_id` as in the body of this function, it's already having: `$this->wcpay_db->order_from_intent_id()` and `$this->wcpay_db->order_from_order_id()`, which are returning these types: `boolean|WC_Order|WC_Order_Refund`. 

For this specific issue, I think keeping the change is as light as possible, i.e. my suggestion above. For ideal return types, we may create an issue for it, and much better have a clear RFC/guideline to avoid this type of issue in the future.

ismaeldcom: Sure! I was thinking on the first cause of the issue, but for fixing it now, I'd stick with an early check 🙂

htdat: Added this commit https://github.com/Automattic/woocommerce-payments/pull/4287/commits/1c26bbd4e395e6cd197e7c989f42952073e06f03

I will take care of creating another issue later.

htdat:
Updated the PR https://github.com/Automattic/woocommerce-payments/pull/4287#discussion_r881584124

Can someone give a final review before I merge it?

ismaeldcom (approved):
LGTM! Thanks for taking care of it 🙂

includes/class-wc-payments-webhook-processing-service.php
htdat: > For this specific issue, I think keeping the change is as light as possible, i.e. my suggestion above. For ideal return types, we may create an issue for it, and much better have a clear RFC/guideline to avoid this kind of issue in the future.

FYI. Added the issue here https://github.com/Automattic/woocommerce-payments/issues/4293

PR #4172, opened 2022-04-21
Update GH Workflow to skip setting up WC Blocks for incompatible WC versions

achyuthajoy:
Fixes #4173 

#### Changes proposed in this Pull Request

We've recently noticed failures on E2E pipeline (discussion - p1649998135891069-slack-CQ0Q6N62D). The problem was caused due to WooCommerce Checkout Blocks version [support policy](https://developer.woocommerce.com/2021/07/27/announcing-a-change-to-the-woocommerce-blocks-version-support-policy/). The minimum supported WooCommerce version for the latest version of WC Blocks is `6.3`.

Since our E2E tests currently runs against lower WooCommerce versions `5.2.0` and `6.1.0`, installing WC Blocks on these versions causes the tests to fail because checkout blocks will fail to load.

Screenshot

This PR disables setting up WC Blocks on incompatible WC versions, skipping blocks related tests. Also, this PR adds page refresh between running WC Blocks to fix flakiness while reading WC error notices.

#### Testing instructions

* Create a new E2E workflow against this branch from Actions tab. Reference job - https://github.com/Automattic/woocommerce-payments/runs/6107673722?check_suite_focus=true
* Confirm that the pipeline is Green

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_

PR #4289, opened 2022-05-25
Bump WC minimum supported version for 4.2.0 release 

zmaglica:

Bump minimum version of WooCommerce from 5.2.0 to 5.4.0
Bump tested version of WooCommerce to 6.5.1

#### Testing instructions

- Validate the version changes.
- Make sure all CI tests are passing.

-------------------

**Post merge**


woocommerce-payments.php
```diff
@@ -8,11 +8,11 @@
...
  * Text Domain: woocommerce-payments
  * Domain Path: /languages
- * WC requires at least: 5.2
- * WC tested up to: 6.4.1
+ * WC requires at least: 5.4
+ * WC tested up to: 6.5.1
  * Requires at least: 5.7
  * Requires PHP: 7.0
- * Version: 4.1.0
+ * Version: 4.2.0
```
htdat, line 15: Don't change this line! It was supposed to handle by `woorelease` command. PCYsg-z2T-p2#release-the-plugin-with-woorelease

.github/workflows/compatibility.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       max-parallel: 10
       matrix:
         # Minimum support version, ..., Latest version, `beta`.
-        woocommerce: [ '5.2.0', '5.3.0', '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.1', '6.2.2', '6.3.0', '6.3.1', 'beta' ]
+        woocommerce: [ '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.1', '6.2.2', '6.3.0', '6.3.1', '6.4.0', '6.4.1', '6.5.0', '6.5.1' 'beta' ]
```
htdat, line 15: The latest minor version for each major version is enough. 

```suggestion
        woocommerce: [ '5.4.0', '5.5.0', '5.6.0', '5.7.0', '5.8.0', '5.9.0', '6.0.0', '6.1.0', '6.1.1', '6.2.2', '6.3.1', '6.4.1', '6.5.1' 'beta' ]
```

.github/workflows/e2e-test.yml
```diff
@@ -44,10 +44,10 @@ jobs:
         exclude:
```
htdat, line 44: I can not comment on line 39 so leaving the comment here. I think this line 39 should be updated too? So it should be: 

```
        woocommerce:   [ '5.4.0', '6.3.1', '6.5.1', 'beta' ]
```

.github/workflows/e2e-test.yml
```diff
@@ -44,10 +44,10 @@ jobs:
...
           - test_groups: 'blocks'
             test_branches: 'merchant'
-          - woocommerce: '5.2.0'
+          - woocommerce: '5.4.0'
             test_groups: 'blocks'
             test_branches: 'shopper'
-          - woocommerce: '6.1.0'
+          - woocommerce: '6.5.1'
```
htdat, line 50: This does not sound correct to me. @achyuthajoy and @deepakpathania - can you have a look to confirm this change? 

Also, I think it's worth commenting why we exclude these combinations?

htdat (dismissed):

woocommerce-payments.php
zmaglica: OMG. Fixed it.

.github/workflows/compatibility.yml
zmaglica: Fixed.

.github/workflows/e2e-test.yml
zmaglica: Fixed.

.github/workflows/e2e-test.yml
zmaglica: I wasn't sure why it was excluded in the previous release, so I used the same pattern.

.github/workflows/e2e-test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       # Conditionally skip WC Blocks tests. Remove/update based on min supported WC version by the blocks checkout plugin.
       - name: Conditionally skip WC Blocks tests
         run: |
-          SKIP_WC_VERSIONS=('5.2.0' '6.1.0')
+          SKIP_WC_VERSIONS=('5.4.0' '6.5.1')
```
achyuthajoy, line 66:
```suggestion
          SKIP_WC_VERSIONS=('5.4.0')
```

We only need to skip running WC Blocks tests against WC `5.4.0` since the minimum supported WC version for WC Blocks is `6.3`

.github/workflows/e2e-test.yml
achyuthajoy: @htdat @zmaglica These test groups were excluded because the minimum supported WC version for WC Blocks. If we're running Blocks tests against an unsupported WC version, the tests will fail and hence they're excluded from testing.

At the moment, the minimum supported WC version for Blocks is `6.3` - https://github.com/woocommerce/woocommerce-blocks/blob/07a0b46a568e6187981cfd89467c78600c82f2af/woocommerce-gutenberg-products-block.php#L12

achyuthajoy: @zmaglica You can remove exclusion for WC versions higher than `6.3`. So in this case, keep exclusion for `5.4.0` which is lower than the minimum supported WC version for blocks.

achyuthajoy: Here is the PR that adds this change - https://github.com/Automattic/woocommerce-payments/pull/4172

achyuthajoy (changes requested):
@zmaglica Changes looks good except for the E2E change. I've requested a few changes

.github/workflows/e2e-test.yml
zmaglica: Fixed.

achyuthajoy (approved):
Changes look good to me @zmaglica 

LGTM 🚢

Issue #504, opened 2020-03-08
Pre-fill dispute evidence form fields


dechov:
Use order details to pre-fill as many dispute evidence fields as possible, and perhaps the product type selector as well.

(See https://github.com/Automattic/woocommerce-payments/pull/500#pullrequestreview-370647844 for initial discussion.)

vbelolapotkov:
Worth discussing the issue details with design team first.

LevinMedia:
@vbelolapotkov generally I'm in favor of pre-filling as much as possible, although given what we know or don't about an order / shipping details, I'm not sure how much more we'll be able to pre-fill than what we're already doing. 

If an order is disputed with only a single product, we could pre-fill product information from the store's product description. Having universal shipping statuses and information in core would allow us to pre-fill additional information, but for now, that information could be coming from any number of sources, so it's going to be a challenge to get anything meaningful.

vbelolapotkov:
@LevinMedia taking into account number of dispute reasons and lack of clarity what/how can be pre-filled this issue looks like high effort - low impact, so I'd close it for now and reopen ones there is a higher need from merchants. WDYT?

LevinMedia:
Sounds good 👍

PR #4294, opened 2022-05-26
Auto redirect a logged in platform user to platform checkout

Mayisha:
Fixes platform checkout issue 504.

## Description
If a user is already logged in to the platform and lands on a merchant's checkout page having platform checkout enabled, the user should be automatically redirected to the platform checkout page by initiating a valid checkout session.

#### Changes in this Pull Request

- When a user first lands on the checkout page, we are checking for a valid platform checkout login session.
- If a valid login session is present, redirecting the user to the platform. Triggers the OTP flow otherwise.
- Tracking the auto redirect event 
- Sending `platform_user_id` in the `session_init` request which is received in the first response.

### Related issue(s)
#504 

### Steps to test
Steps are mentioned in the other related PR.

### DIFF
--- a/client/checkout/platform-checkout/email-input-iframe.js
+++ b/client/checkout/platform-checkout/email-input-iframe.js
@@ -15,6 +15,10 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
 		return;
 	}
 
+	// Make a overlay for the page to behave disabled
+	const overlay = document.createElement( 'div' );
+	overlay.classList.add( 'disabled-page' );
+
 	const spinner = document.createElement( 'div' );
 	const parentDiv = platformCheckoutEmailInput.parentNode;
 	spinner.classList.add( 'wc-block-components-spinner' );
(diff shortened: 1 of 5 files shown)


client/checkout/platform-checkout/style.scss
```diff
@@ -45,6 +45,17 @@
 	}
 }
 
+.disabled-page {
```
Mayisha, line 48: This is not part of any approved design. I added this as a solution to prevent the user from doing anything on the checkout page while the request for checking a platform user session is going on.

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -204,6 +208,19 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
 		}
 	} );
 
+	const platformCheckoutCheckSession = () => {
+		return fetch(
+			`${ getConfig(
+				'platformCheckoutHost'
+			) }/wp-json/platform-checkout/v1/user/session`,
```
harriswong, line 215: Does this needs to be done on the front-end? Or can this be done via a hook of some sort on the PHP side? 

I am thinking something like, as soon as `/checkout` is loaded, PHP sends the request to `/wp-json/platform-checkout/v1/user/session` before we render anything. This way, we don't need to disable the screen and the checkout redirection can be more fluid.

harriswong, line 215: Does this needs to be done on the front-end? Or can this be done via a hook of some sort on the PHP side? 

I am thinking something like, as soon as `/checkout` is loaded, PHP sends the request to `/wp-json/platform-checkout/v1/user/session` before we render anything. This way, we don't need to disable the screen and the checkout redirection can be more fluid.

Issue #4280, opened 2022-05-23
Multi-Currency Analytics: Only apply join clauses on Orders queries


dmallory42:
### Description

When Multi-Currency is enabled, queries made to the Analytics section of WC Admin are modified to join in order to fetch the original currency of the order and the exchange rate. This is in order to display the original currency an order was placed in. However, we can only do this on the `Orders` analytics page, meaning we are doing unnecessary joins on the other analytics pages.

We can make this more efficient by checking to see which context we're in before making the joins.

### Acceptance criteria

* We should only add the extra joins in MC Analytics if in the context of the Orders page.

### Additional Context

p1653313090124909-slack-C01B8KNUYSW

dmallory42:
This might be a little more complex than I initially thought. I made a quick spike into it and realised that we make [replacements](https://github.com/Automattic/woocommerce-payments/blob/7f58d28449bf987e884a180e71e47d63abfffcd7/includes/multi-currency/Analytics.php#L349) in the SQL using the joins in the case of calculating the `discount_amount`, `product_net_revenue`, and `product_gross_revenue` -  this is because these values are looked up using the *original* order values (stored in `wc_orders`) - rather than the stats tables.

The issue with this is that the original order values would be stored in the currency they were purchased in. For this reason, it is necessary for us to make this conversion with the SQL query.

It may be possible to optimise and work out which cases we need to apply these transformations, but it wouldn't be possible to optimise as much as I initially thought without modifying how analytics works in WC Admin.

joshbetz:
Some questions from Slack:

How do we know if a store supports multiple currencies? Is that opt-in or enabled automatically?

Can we detect when a store has only ever received orders from a single currency and disable some of this?

dmallory42:
> How do we know if a store supports multiple currencies? Is that opt-in or enabled automatically?

Enabled automatically currently, but it may be worth exploring whether it's safe/possible to disable this functionality in the case where a user has not onboarded with WCPay yet.

> Can we detect when a store has only ever received orders from a single currency and disable some of this?

Good question! Again, I think that's something worth exploring.

joshbetz:
I think we can close this now?

PR #4296, opened 2022-05-27
Prevent expensive JOINs when there aren't multiple currencies used

joshbetz:

Prevent expensive JOINs when there aren't multiple currencies used

Looks up the number of currencies that have been used on this site. If
there are more than 1, we need to reflect that in the analytics.
Otherwise we should avoid the expensive JOINs.

See #4280

#### Testing instructions (updated by @dmallory42)

Checking the MC Analytics work when a MC order has been made

- On a site where you have previously made a multi-currency order, go to the analytics section and verify (either using step debugging, or query monitor to view the queries that are being run) to ensure that when the `Analytics` code is initialised, it does *not* return at [this point](https://github.com/Automattic/woocommerce-payments/blob/a4b6b377ab64fe0e389f74bfbfd142b7193a3979/includes/multi-currency/Analytics.php#L72). 
- Make an order with a multi-currency, and view the analytics for today to ensure that the order appears, and the figures (converted into your store default currency) look sensible.

Checking Analytics still loads when no MC order has been made

- Edit the `has_multi_currency_orders` function to return false, then check again - this time , the initialisation of MC Analytics should return early, before adding the filters. Check the analytics and verify that nothing is broken, all pages load without errors etc. (note that the total figures *may* be incorrect in this case, because the filters were not added and the store has MC orders, in prod this should not happen since we'd only return early if there are no MC orders).

-------------------

**Post merge**

### DIFF
--- a/includes/multi-currency/Analytics.php
+++ b/includes/multi-currency/Analytics.php
@@ -66,32 +66,16 @@ class Analytics {
 
 		add_filter( 'woocommerce_analytics_update_order_stats_data', [ $this, 'update_order_stats_data' ], self::PRIORITY_LATEST, 2 );
 
-		// If we aren't making a REST request, return before adding these filters.
+		// If we aren't making a REST request, or no multi currency orders exist in the merchant's store,
+		// return before adding these filters.
 		if ( ! WC()->is_rest_api_request() || ! $this->has_multi_currency_orders() ) {
 			return;
-
 		}
 
 		$this->set_sql_replacements();
 
 		add_filter( 'woocommerce_analytics_clauses_select', [ $this, 'filter_select_clauses' ], self::PRIORITY_LATE, 2 );
 		add_filter( 'woocommerce_analytics_clauses_join', [ $this, 'filter_join_clauses' ], self::PRIORITY_LATE, 2 );
-	}
-
-	public function has_multi_currency_orders() {
-		global $wpdb;
-		$query = <<<EOD
-			SELECT COUNT(DISTINCT `meta_value`) as `count`
-			FROM `{$wpdb->postmeta}`
-			INNER JOIN `{$wpdb->posts}` ON (`{$wpdb->postmeta}`.`post_id` = `{$wpdb->posts}`.`id`)
-			WHERE `{$wpdb->posts}`.`post_type` = 'order'
-				AND `meta_key` = '_order_currency'
-				AND `meta_value` <> ''
-				AND `meta_value` <> NULL;
-			EOD;
-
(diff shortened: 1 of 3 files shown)


joshbetz:
@dmallory42 I think this will miss a big use-case I was thinking of where multi-currency is enabled, but only 1 currency has ever been used. I suspect this is true for many stores--especially any stores that primarily do in person payments or local delivery. It seems like the multi-currency support is not useful in that case?

dmallory42:
>  I think this will miss a big use-case I was thinking of where multi-currency is enabled, but only 1 currency has ever been used. I suspect this is true for many stores--especially any stores that primarily do in person payments or local delivery. It seems like the multi-currency support is not useful in that case?

Hi! That use case should be covered by this implementation. 🙂 In the case that orders have only ever been placed in the store default currency, the metadata key `_wcpay_multi_currency_order_exchange_rate` will not be present, therefore `has_multi_currency_orders` should return `false` and the JOINs will not be added.

includes/multi-currency/Analytics.php
```diff
@@ -332,6 +333,32 @@ private function generate_case_when( string $variable, string $then, string $els
...
+		$orders = intval(
+			$wpdb->get_var(
+				"
+				SELECT
+					COUNT(DISTINCT orders.ID)
+				FROM
+					{$wpdb->posts} AS orders
+				INNER JOIN
+					{$wpdb->postmeta} order_meta ON order_meta.post_id = orders.ID
+					AND order_meta.meta_key = '_wcpay_multi_currency_order_exchange_rate';
```
ismaeldcom, line 354: Not a big improvement, but thinking on squeezing every ms, maybe we could avoid the `JOIN` and just look for the meta. In the end we only want to know if there are orders with exchange rate.

```suggestion
				SELECT
					COUNT(post_id)
				FROM
					{$wpdb->postmeta}
				WHERE
					meta_key = '_wcpay_multi_currency_order_exchange_rate';
```

ismaeldcom (approved):
Thanks @joshbetz  and @dmallory42 for working on this, it's a big performance improvement for those sites using only the store currency while having Multi-Currency enabled.

Looks good to me! Just a tiny check to prevent the performance hit when possible. I've only added a nitpick, but feel free to merge without it 🙂

includes/multi-currency/Analytics.php
dmallory42: Good idea! That would achieve the same result in a more efficient way. Committed the suggestion!

joshbetz:
Awesome. Looks good to me.

PR #4302, opened 2022-05-31
Fixes for unit tests on WPCOM

millerf:
See p81Rsd-ND-p2

- Rename WP_UnitTestCase to WCPAY_UnitTestCase so it can be overriden on WPCOM
- Rewrite URLs with `'sites/3/'`
- Manually init WC_Payments_Explicit_Price_Formatter
- Do not pass `null` as request for rest_request_before_callbacks
- Skip `test_rest_endpoints_validate_nonce`
- Cast option to `bool`
- Expect locale to be `en`
- Enforce uppercase for transients
- ignore PHPCS warnings about modifying gloabl vars

### DIFF
--- a/changelog/tests-fixes-for-wpcom
+++ b/changelog/tests-fixes-for-wpcom
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+
--- a/changelog/tests-fixes-for-wpcom
+++ b/changelog/tests-fixes-for-wpcom
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Fix compatibility with WPCOM test environment
--- a/changelog/tests-fixes-for-wpcom
+++ b/changelog/tests-fixes-for-wpcom
@@ -1,4 +1,0 @@
-Significance: patch
-Type: dev
-
-Fix compatibility with WPCOM test environment
--- a/tests/WCPAY_UnitTestCase.php
+++ b/tests/WCPAY_UnitTestCase.php
@@ -11,4 +11,7 @@
  * Class WP_UnitTestCase
  */
 class WCPAY_UnitTestCase extends \Yoast\PHPUnitPolyfills\TestCases\TestCase {
+	protected function is_wpcom() {
+		return defined( 'IS_WPCOM' ) && IS_WPCOM;
+	}
 }
--- a/tests/WCPAY_UnitTestCase.php
+++ b/tests/WCPAY_UnitTestCase.php
@@ -10,6 +10,5 @@
  *
  * Class WP_UnitTestCase
  */
-class WCPAY_UnitTestCase extends WP_UnitTestCase {
-
+class WCPAY_UnitTestCase extends \Yoast\PHPUnitPolyfills\TestCases\TestCase {
 }
--- a/tests/WCPAY_UnitTestCase.php
+++ b/tests/WCPAY_UnitTestCase.php
@@ -10,7 +10,7 @@
  *
  * Class WP_UnitTestCase
  */
-class WCPAY_UnitTestCase extends \Yoast\PHPUnitPolyfills\TestCases\TestCase {
+class WCPAY_UnitTestCase extends WP_UnitTestCase {
 	protected function is_wpcom() {
 		return defined( 'IS_WPCOM' ) && IS_WPCOM;
 	}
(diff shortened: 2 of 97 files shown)


tests/WCPAY_UnitTestCase.php
```diff
@@ -10,6 +10,6 @@
  *
  * Class WP_UnitTestCase
  */
-class WP_UnitTestCase extends \Yoast\PHPUnitPolyfills\TestCases\TestCase {
```
marcinbot, line 13: I might be missing something, but was `WP_UnitTestCase` declared somewhere else? What about the polyfills? I'm surprised the tests didn't start failing.

changelog/tests-fixes-for-wpcom
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Fix compatibility with WPCOM test environment
```
marcinbot, line 4: I think we could skip a changelog entry since this is for internal purposes. Let's not advertise WPCOM too much.

tests/unit/admin/test-class-wc-rest-payments-accounts-controller.php
```diff
@@ -34,7 +34,11 @@ public function set_up() {
 
 		// Set the user so that we can pass the authentication.
 		wp_set_current_user( 1 );
-		WC_Payments::get_gateway()->update_option( 'test_mode', 'yes' );
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
```
marcinbot, line 37: There's a bunch of `defined( 'IS_WPCOM' ) && IS_WPCOM` - I think this warrants a helper method in `WCPAY_UnitTestCase`

In fact, I see that we override the base class in WPCOM. We could have a helper method whose name doesn't advertise WPCOM, the base class here returns false, while the base class in WPCOM would always return true. That way we could avoid exposing the inner workings of our systems, even if it's just one constant.

changelog/tests-fixes-for-wpcom
millerf: Some checks would trigger an error if I didn't add any changelog...

tests/unit/admin/test-class-wc-rest-payments-accounts-controller.php
millerf: Sure. Good idea.

changelog/tests-fixes-for-wpcom
marcinbot: I think as long as you do `npm run changelog` and then select `patch` as significance, you will be able to provide an empty entry. It will still generate an empty file, but it won't be put in the CL on release.

tests/WCPAY_UnitTestCase.php
marcinbot: Would be good to keep inheriting from `\Yoast\PHPUnitPolyfills\TestCases\TestCase`, especially since we have a separate override in wpcom.

changelog/tests-fixes-for-wpcom
millerf: I removed it and it seems to be fine

millerf: Nevermind. I had to add it. Added an empty changelog.

tests/WCPAY_UnitTestCase.php
millerf: I really don't understand why tests fail when using the polyfill...

millerf:
@marcinbot I added your feedback.

tests/WCPAY_UnitTestCase.php
marcinbot: I know why :) We never used the polyfill. It wasn't imported anywhere. The import was only added to the `bootstrap.php` file in this PR. In that case, I think we can consider this resolved for now and add refactors separately if we ever want to use the polyfill.

marcinbot (approved):
Looks good :shipit:

tests/WCPAY_UnitTestCase.php
millerf: I see (and I think I noticed it when I was working on this). The whole [WP_UnitTestCase.php](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/WP_UnitTestCase.php) is not actually used. The `WP_UnitTestCase` class used across the unit tests is the [WP Core one](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/factory.php)....

PR #4309, opened 2022-06-01
Add UPE appearance filter

### LINKED ISSUE #3604: Add filters to modify the css styling for credit card fields
### Description
There are no filters to manually modify the fields on WooCommerce Payments UPE or legacy card fields. 

Currently, WCPay fields use Stripe's Universal Payment Element (UPE) as opposed to the previous card element and scans fonts and styles from existing inputs on the page in order to match existing themes as much as possible - but for some users they will want more control over the display.

For Legacy checkout, WC Pay lacks an equivalent to the Stripe plugin's `wc_stripe_elements_styling` filter ([described in the Stripe docs here](https://woocommerce.com/document/stripe-styling-fields/#section-4)), which allows site owners to add their own styling to the legacy form. In other words, disabling UPE doesn't allow for more styling options, unlike Stripe.

### Acceptance criteria
Having a filter to either change the CSS selector used to compute styles and/or a filter to manually set styling properties for the Stripe element.

Though I'm sure some developers would like this for the legacy fields, I think it would be acceptable to focus on the UPE for this request.

### Dev notes
For reference, Stripe's card element allows styling options in a [different format](https://stripe.com/docs/js/appendix/style) to the newer [UPE styling format](https://stripe.com/docs/stripe-js/appearance-api#rules), so it's not trivial to copy the legacy filter over the the new checkout. 

#1960 would allow merchants to choose the CSS selector that we use to dynamically scan CSS properties; however, this would still not allow merchants to manually set individual properties themselves, without changing the theme.

### Additional context
* 4647359-zen
* p1640947060027100-slack-C7U3Y3VMY

ismaeldcom:
Fixes #3604

Add `wcpay_upe_appearance` filter for both classic and block based. Apply the filter on the current `get_transient` will load the custom appearance directly, without running the style generation.

Any wrong property provided will be evaluated by Stripe and show a warning log in the browser console, but still loads all the valid ones.

#### Screenshot
Screenshot 2022-06-01 at 11 52 38

_Result using rules `.Label { color: green }`_

#### Testing instructions

Add the following filter, you can try any other properties [from Stripe's doc](https://stripe.com/docs/elements/appearance-api).
```php
add_filter(
	'wcpay_upe_appearance',
	function() {
		return [
			'theme' => 'night',
		];
	}
);
```

- Having UPE enabled.
- For both classic checkout and block based one.
- Your custom appearance should load correctly.

-------------------

**Post merge**

- [ ] Update WCPay public documentation.

### DIFF
--- a/includes/payment-methods/class-upe-payment-gateway.php
+++ b/includes/payment-methods/class-upe-payment-gateway.php
@@ -698,8 +698,8 @@ class UPE_Payment_Gateway extends WC_Payment_Gateway_WCPay {
 		$payment_fields['paymentMethodsConfig']     = $this->get_enabled_payment_method_config();
 		$payment_fields['saveUPEAppearanceNonce']   = wp_create_nonce( 'wcpay_save_upe_appearance_nonce' );
 		$payment_fields['testMode']                 = $this->is_in_test_mode();
-		$payment_fields['upeAppearance']            = get_transient( self::UPE_APPEARANCE_TRANSIENT );
-		$payment_fields['wcBlocksUPEAppearance']    = get_transient( self::WC_BLOCKS_UPE_APPEARANCE_TRANSIENT );
+		$payment_fields['upeAppearance']            = apply_filters( 'wcpay_upe_appearance', get_transient( self::UPE_APPEARANCE_TRANSIENT ) );
+		$payment_fields['wcBlocksUPEAppearance']    = apply_filters( 'wcpay_upe_appearance', get_transient( self::WC_BLOCKS_UPE_APPEARANCE_TRANSIENT ) );
 		$payment_fields['checkoutTitle']            = $this->checkout_title;
 		$payment_fields['cartContainsSubscription'] = $this->is_subscription_item_in_cart();
 		$payment_fields['logPaymentErrorNonce']     = wp_create_nonce( 'wcpay_log_payment_error_nonce' );
--- a/tests/unit/payment-methods/test-class-upe-payment-gateway.php
+++ b/tests/unit/payment-methods/test-class-upe-payment-gateway.php
@@ -1571,6 +1571,24 @@ class UPE_Payment_Gateway_Test extends WP_UnitTestCase {
 		);
 	}
 
+	public function test_upe_appearance_filter() {
+		$expected = [
+			'theme' => 'night',
+		];
+
+		add_filter(
+			'wcpay_upe_appearance',
+			function() use ( $expected ) {
+				return $expected;
+			}
+		);
+
+		$payment_fields = $this->mock_upe_gateway->get_payment_fields_js_config();
+		$this->assertSame( $expected, $payment_fields['upeAppearance'], );
+		$this->assertSame( $expected, $payment_fields['wcBlocksUPEAppearance'], );
+	}
+
+
 	public function maybe_filter_gateway_title_data_provider() {
 		$method_title   = 'WooCommerce Payments';
 		$checkout_title = 'Popular payment methods';


FangedParakeet:
Thanks for whisking up this PR! I have a few comments on the approach, but I'd definitely appreciate your thoughts. 

First of all, it's obvious worth noting that any properties added in this filter must conform to Stripe's [accepted list of properties](https://stripe.com/docs/elements/appearance-api), which is unfortunately and inexplicably less CSS-like than their [previous API](https://stripe.com/docs/js/appendix/style), which the [corresponding Stripe filter](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L1855) takes advantage of instead.

Secondly, the Stripe plugin makes no attempt to alter the styling of elements without the use of the supplied filter. WCPay, on the other hand, [adds hidden elements into the page and uses existing selectors on the page](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L7-L36) to scan these elements for their CSS properties. These properties can be fed into Stripe's Elements API to attempt to automatically match the theming of the payment element to the current theme. You can see that there are [quite a lot of properties](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/upe-styles.js) for which we attempt to set styling properties for on the payment element. When using the filter as prescribed in this PR, we will be replacing _all_ of these properties with whatever is contained within the filter. This means that even if, for example, a customer examined and understood Stripe's Elements API and desired to simply change the font-colour of the text labels, they would not be able to do so without removing all of the existing styling that we set to these elements. It is also practically impossible for them to examine and copy the styles that we already set without altering our plugin code, so this might be frustrating for an end-user. 

Could we possibly merge the results of the filter and our own computations, so that the filter could add or subtract from the styles that are already computed rather than replacing them entirely? The user would still have to comply exactly with Stripe's format for rules, knowing and accepting that if they added some faultily formatted property that broke the Elements API, all styles would be removed and they would only have themselves to blame.

Thirdly and lastly, to avoid us computing these aforementioned automatically scanned styling properties on every page load, we save them to a transient that can be recalled more readily. There are actually two separate transients that are used for the regular and blocks checkout (`UPE_APPEARANCE_TRANSIENT` and `WC_BLOCKS_UPE_APPEARANCE_TRANSIENT` respectively). You can see [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L13-L36) that there are different selectors and targets used for the classic and blocks checkouts that result in quite different styles being returned. Overwriting both values with the same filter would be more problematic for this reason as well.

While merging the filter styles with the transient/computed styles mitigates the damage mentioned above, perhaps we should consider allowing separate filters to allow users to add separate custom style parameters to their blocks and regular checkouts? 

Also, this has nothing to do with your changes, but it is just me or are we missing a closing bracket over [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L36-L37)? 🧐

Please let me know your thoughts on any and all of the above. Cheers!

ismaeldcom:
Thanks for your detailed input! 🙂 

Glad to confirm that it works in the way I understood reading the code, and kudos for us because it was quite easy to read.

>Could we possibly merge the results of the filter and our own computations, so that the filter could add or subtract from the styles that are already computed rather than replacing them entirely?

My bad, I tested it early and was receiving the generated rules in the filter, but when the transients are cleared, as we're retuning new rules with the filter, the style generation didn't get call (as expected).

You were right about it being a bit more complex than expected 🙂 We could proceed with the filter overriding the rules or look for a way to allow extending them. I think the latter would be better, but we should at least offer the former.

>While merging the filter styles with the transient/computed styles mitigates the damage mentioned above, perhaps we should consider allowing separate filters to allow users to add separate custom style parameters to their blocks and regular checkouts?

Good question! I made it to myself too, and my last thought was that it would be uncommon to use both types of checkout at the same time. But yeah, makes sense to have 2 filters, they can always just point the same method to both of them.

>Also, this has nothing to do with your changes, but it is just me or are we missing a closing bracket over [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L36-L37)? 🧐

We're closing it [later here](https://github.com/Automattic/woocommerce-payments/blob/5615b8854175b81f580a905159f2225884b73e59/client/checkout/upe-styles/index.js#L83), because `updateSelectors` and `getSelectors` are declared inside `appearanceSelectors`, right?

FangedParakeet:
Sorry, I hope progress here didn't pause in anticipation of my response, because I completely forgot to return to this issue! 😣

> 🤔 Maybe an option could be having a new property (something like custom_upe_appearance) and merge it with the stored/generated appearance. Will be pretty straightforward to implement, keeping our current implementation almost intact.

This sounds like a really good idea and I think it's pretty much what I had in mind (not that I'm trying to steal the credit 😅).

> I made it to myself too, and my last thought was that it would be uncommon to use both types of checkout at the same time.

Ooh yes, I suppose this is true. I'm not sure why or how a store owner would offer the Blocks and Classic checkout as separately available options for customers. I think the only thing that I was getting at was that one value should not overwrite the other, because the computed styles for each checkout can be quite drastically different. Since we cache two separate transients for these styles, I felt it would be appropriate to maintain two separate filters as well, but, as you suggest, since it's unlikely these will be used in parallel, perhaps a single filter will suffice. Store owners would just have to be aware that upon switching between these checkout styles, the custom styles they enter in the filter will also likely need to be updated.

> We're closing it [later here](https://github.com/Automattic/woocommerce-payments/blob/5615b8854175b81f580a905159f2225884b73e59/client/checkout/upe-styles/index.js#L83), because updateSelectors and getSelectors are declared inside appearanceSelectors, right?

D'oh, right you are! Those functions are just properties on the same object--false alarm, my bad! 😁

ismaeldcom:
Thanks for your input! 🙂 And don't worry, I just didn't have time to work again in this issue. I'll try to implement that solution.

jakewhiteley:
Please reserrect this issue - not being able to affect field styling is very annoying.

**edit* Styling can be edited by modifying the transient used by the plugin which is passed into `upeAppearance`, its not a particularly nice way of doing this though.

dmvrtx:
The styling approach has been updated recently, so I'm closing this one.

PR #4318, opened 2022-06-03
add check to prevent fatal errors when $order is null on success page

leonardola:
Fixes #

This PR adds a check to prevent fatal errors when $order is null. 

#### Testing instructions
- Buy any product and checkout
- You should see no errors and the default success page should show up
- Checkout again but this time using the WooPay
- You should see no errors and the WooPay success page should show up(Green notice and WooPay logo_
- Change [this line](https://github.com/woocommerce/woocommerce/blob/35b58245af09042934d9c726786cb3dfcb246838/plugins/woocommerce/templates/checkout/thankyou.php#L24) to `if(false) {` 
- Repeat the process. Now both times you should only see the default thank you notice and no other message or error.

-------------------


alefesouza (approved):
LGTM 🚢

leonardola:
Closing this because it needs to be a patch. So I'll open another one using trunk

PR #4320, opened 2022-06-03
added check to prevent fatal errors when no $order is null

### LINKED ISSUE #4317: Fatal error on checkout without order
A customer complained that after the last update of WC Payments they've started to see critical errors on their checkout. Which is a custom checkout created by **WooFunnels: by BuildWooFunnels** plugin  ( ticket at 5266568-zd-woothemes ). Apparently the bug happens because at [this line](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/checkout/thankyou.php#L84) we don't send an order to the `woocommerce_thankyou_order_received_text` filter. To prevent these types of errors we need to add a check [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-order-success-page.php#L32) to make sure the `$order` is not null before using it.

leonardola:
fix #4317 

This PR adds a check to prevent fatal errors when $order is null. 

#### Testing instructions
- Buy any product and checkout
- You should see no errors and the default success page should show up
- Checkout again but this time using the WooPay
- You should see no errors and the WooPay success page should show up(Green notice and WooPay logo_
- Change [this line](https://github.com/woocommerce/woocommerce/blob/35b58245af09042934d9c726786cb3dfcb246838/plugins/woocommerce/templates/checkout/thankyou.php#L24) to `if(false) {` 
- Repeat the process. Now both times you should only see the default thank you notice and no other message or error.

-------------------

### DIFF
--- a/changelog/fix-4317-fatal-error-on-checkout-without-order
+++ b/changelog/fix-4317-fatal-error-on-checkout-without-order
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+added check to prevent fatal errors on checkout
--- a/includes/class-wc-payments-order-success-page.php
+++ b/includes/class-wc-payments-order-success-page.php
@@ -23,13 +23,13 @@ class WC_Payments_Order_Success_Page {
 	/**
 	 * Return the WooPay thank you notice when the order was created via WooPay
 	 *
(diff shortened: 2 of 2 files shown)


alefesouza (approved):
LGTM :shipit:

PR #4328, opened 2022-06-07
Auto redirect a logged in platform user to platform checkout without OTP (take 2 via Iframe)

Mayisha:
Fixes platform checkout issue 504.

#### Description
If a user is already logged in to the platform and lands on a merchant's checkout page having platform checkout enabled, the user should be automatically redirected to the platform checkout page by initiating a valid checkout session.

- When a user first lands on the checkout page, we are checking for a valid platform checkout login session with a hidden iframe. If a valid session exists, the iframe is made visible and shows a loader while the user is redirected to the platform checkout.
- If a valid login session is absent, it triggers the OTP flow as before.
- Tracking the auto redirect event

#### Testing instructions
Steps are mentioned in the other related PR.


client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -252,14 +276,40 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
 	// the back button on platform checkout itself.
 	const searchParams = new URLSearchParams( window.location.search );
 
 	if ( 'true' !== searchParams.get( 'skip_platform_checkout' ) ) {
-		// Check the initial value of the email input and trigger input validation.
-		if ( validateEmail( platformCheckoutEmailInput.value ) ) {
-			platformCheckoutLocateUser( platformCheckoutEmailInput.value );
+		// Check if user already has a WooPay login session.
+		if ( ! hasCheckedLoginSession ) {
+			openLoginSessionIframe();
```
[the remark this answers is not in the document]
Mayisha: It is intended on my end. In this implementation, if the `login-session` does not exist we call `closeLoginSessionIframe` and there I am [calling](https://github.com/Automattic/woocommerce-payments/pull/4328/files#diff-d65a9401ef14963cb2c9f8d930e98c69f81e5832f50fe794934471c4fd1a8c4fR300-R302) this `platformCheckoutLocateUser` if the email input has valid value.

Let me know your thoughts. :eyes:

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -19,13 +20,36 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
 	spinner.classList.add( 'wc-block-components-spinner' );
 
-	// Make the iframe wrapper.
+	// Make the login session iframe wrapper.
+	const loginSessionIframeWrapper = document.createElement( 'div' );
+	loginSessionIframeWrapper.setAttribute( 'role', 'dialog' );
+	loginSessionIframeWrapper.setAttribute( 'aria-modal', 'true' );
+	loginSessionIframeWrapper.classList.add(
+		'platform-checkout-login-session-iframe-wrapper'
+	);
```
FangedParakeet, line 29: This wrapper will block the entire checkout screen for _all_ users regardless of WooPay account status: this seems like it might get a little confusing for non-WooPay users, since the entire checkout will be blocked indefinitely until the request to check for an active session is completed. Is there any reason why we need to block the whole checkout while this request is being made or can we not wait until the redirect response returns `true` to bring the curtains down?

Mayisha: Probably we can ask for a design decision on this. It is absolutely true that this experience would be confusing for the non-WooPay user. My initial thought on this was if the page is not blocked and the user is a logged in WooPay user, they can start editing the fields on the checkout page and in the middle of their edit suddenly the redirection will start :sweat_smile:

FangedParakeet: Cool, I don't really have a huge issue with this approach, but would be nice to get some design feedback on it! 

In my humble opinion, I wouldn't mind that a WooPay user starts filling in their details before being interrupted and whisked away, because they would only be sitting and waiting in the checkout anyways and at least they wouldn't have to fill in this information twice. However, forcing a non-WooPay user to sit and wait for the request is a little bit of a waste of their time, since there will be no reward received by them for this time spent.

But what do I know about UX/UI? (Answer: not a lot. 😁)

Mayisha: I had a chat with @rogermattic and decided to go for the approach you suggested to unblock most of the users. Fixed in [8aa1f0d](https://github.com/Automattic/woocommerce-payments/pull/4328/commits/8aa1f0db9f8e22080527802affd2856c8e1ae8e0), now we are not blocking the user from editing the checkout fields while the request to check for a login session is in progress.

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -297,6 +351,29 @@ export const handlePlatformCheckoutEmailInput = ( field, api ) => {
...
+				hasCheckedLoginSession = true;
+				loginSessionIframeWrapper.classList.add(
+					'platform-checkout-login-session-iframe-wrapper'
+				);
+				loginSessionIframe.classList.add( 'open' );
+				wcpayTracks.recordUserEvent(
+					wcpayTracks.events.PLATFORM_CHECKOUT_AUTO_REDIRECT
+				);
+				api.initPlatformCheckout(
+					e.data.userEmail,
```
harriswong, line 364: Do we still have `e.data.userEmail` for `auto_redirect_to_platform_checkout`?

Mayisha: No actually. I can send `""` there as well.

FangedParakeet (approved):
I think the changes here look fine to me, pending any more modifications following the review of the sibling PR in the platform checkout plugin side. 👍

PR #4355, opened 2022-06-16
Use country-phone input component for terminal settings phone field

### LINKED ISSUE #3986: Use the new country-phone input for terminal receipt phone number field
A new input type for phone numbers was introduced in the plugin codebase recently that comes with an integrated country code selector.

We should replace the support phone number field on Card Reader receipt settings page (#3410) with the new input type.

phone number

elazzabi:
Fixes #3986 

Use country-phone component instead of a normal input field.

#### Testing instructions

No regression between the old component and the new one.

Before:

Screen Shot 2022-06-16 at 16 44 18

After:

Screen Shot 2022-06-16 at 16 17 12

-------------------

**Post merge**


elazzabi:
⚠️ 
**Limitations**:
The component was designed for checkout and a lot of the decisions/code is tight to the checkout process: hard-coded class names, etc... I tried to keep it as it is, so I don't break the checkout process, but if we want to reuse that component in different places, it needs to be decoupled from the checkout process.

So far, the biggest limitation is the styling of the component and its interaction with elements on the page (hard-coded class names...)

kalessil:
@elazzabi : so we managed to make it work here without side effects but the component is not as reusable as might be?

elazzabi:
@kalessil Yes, it does work, but there are a few problems:
- we can't style it to make it look like the other inputs on the page (the width of the component is different from others on the page for example)
- it's still trying to interact with other inputs on the page via hardcoded classes names, but those do not exist (exists only in the checkout process)

client/card-readers/settings/sections/test/index.test.js
```diff
@@ -71,7 +71,7 @@ describe( 'Card Reader Contact Details section', () => {
 		const email = screen.getByLabelText( 'Support email' );
 		expect( email ).toBeInTheDocument();
 
-		const phone = screen.getByLabelText( 'Support phone number' );
+		const phone = screen.getByLabelText( 'Mobile phone number' );
```
kalessil, line 74: Is this change due to the component usage (I guess a hardcoded value there)?

kalessil:
@elazzabi : IIRC, under the hood, it uses a third-party component. How much effort it would be to use the third-party one here directly?

client/card-readers/settings/sections/test/index.test.js
elazzabi: Yeah, the label `Mobile phone number` is hardcoded inside the component, so I can't pass one...

elazzabi:
@kalessil good call. I believe most of the code added to the component is to make it work with the checkout process, so I don't think it will be hard to use the component directly. I'll double-check and update this PR accordingly.

kalessil:
Super, thanks @elazzabi !

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -1,144 +1,111 @@
...
-					'+' + iti.getSelectedCountryData().dialCode,
-					''
-				)
-			);
+			setIsValid( iti.isValidNumber() );
+			onValidationChange( iti.isValidNumber() );
 
 			// Focus the phone number input when the component loads.
-			input.focus();
+			inputRef.current.focus();
```
elazzabi, line 63: I'm not sure why this is used. I believe we shouldn't automatically focus on the mobile input when the page is loaded, but maybe I'm missing something...

Should I just go and remove it?

elazzabi:
@kalessil @mordeth Finally made the component decoupled from that page. Changes introduced are:
1. Removed all hard-coded values, class names, or others. They are now either passed via props or taken care of by the parent component
1. Moved the error handling to parent component. This way we can display errors depending on the page-styling, etc...
1. Removed a bunch of internal state that can be replaced by the parent component. The PhoneInput component is now used for representation only + communicating if the phone is valid or not with the parent for styling/error-handling.

What's missing from this PR (I'll work on it next):
1. TypeScript support
2. Fixing tests (if any broken ones)
3. Running some manual tests for checkout phase

anu-rock:
Also invited Mayisha to review as she had originally added the tel input component in #3699.

@Mayisha - We are trying to adapt the tel input component on card reader settings page. We realized that the component was customized for checkout, so we took some steps to make it more generic. Do let us know your thoughts.

mordeth (commented):
Looks good so far! I have added few comments. Something I would change is the location of the component. As it is no longer Platfrom checkout specific, I think `settings` is much better place for it.

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -14,18 +14,50 @@ import AdditionalInformation from './additional-information';
...
+		const formSubmitButton = document.querySelector(
+			'form.woocommerce-checkout button[type="submit"]'
+		);
+
+		if ( ! formSubmitButton ) {
+			return;
+		}
+
+		const updateFormSubmitButton = () => {
+			if ( isPhoneValid ) {
```
Mayisha, line 46: `isPhoneValid` is `null` initially and it does not change until the user clicks the `Save my information for faster checkouts` checkbox or edits the phone number input field. For the `null` value the `Place Order` button is disabled though I have all the required fields filled up in the original checkout blocks.

disabled

The button gets enabled as soon as I select the checkbox.

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -1,144 +1,111 @@
...
-		document.getElementById( 'billing_phone' )?.value ?? ''
-	);
+const PhoneNumberInput = ( {
+	onValueChange,
+	value,
+	onValidationChange = () => {},
+	inputProps = {},
+} ) => {
 	const [ inputInstance, setInputInstance ] = useState( null );
 	const [ isValid, setIsValid ] = useState( true );
```
Mayisha, line 22: We now have 2 state variables indicating whether the phone number is valid or not, one in the parent component and another in this. How about we consolidate the variables and have just one?

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -1,144 +1,111 @@
...
-		const input = document.querySelector(
-			`input[aria-label="${ __(
-				'Mobile phone number',
-				'woocommerce-payments'
-			) }"]`
-		);
 
 		const handleCountryChange = () => {
-			handlePhoneNumberChange( iti.getNumber() );
+			onValueChange( iti.getNumber() );
```
Mayisha, line 47: Let's call the `onValidationChange( iti.isValidNumber() )` also here. Currently if I change the country from the error message does not appear even if it is an invalid phone number for the newly selected country.

client/components/platform-checkout/save-user/phone-number-input.js
```diff
@@ -1,144 +1,111 @@
...
+				'countrychange',
+				handleCountryChange
+			);
 		}
 
 		return () => {
 			if ( iti ) {
 				iti.destroy();
-				input.removeEventListener(
+				inputRef.removeEventListener(
```
Mayisha, line 74: Small typo here, should be `inputRef.current`. This gives an error in the console currently.

Mayisha (changes requested):
I have tested the flow on the checkout page and found a couple of small issues that need to be fixed. :eyes: 
Apart from that, I agree with @mordeth about moving the component to a generic place as an independent component. :+1:

client/components/platform-checkout/save-user/checkout-page-save-user.js
elazzabi: Can you please let me know how to reproduce this? I can't reproduce it on my end:
Screen Shot 2022-06-28 at 11 48 30

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -14,18 +14,50 @@ import AdditionalInformation from './additional-information';
...
 		isWCPayChosen,
 		isNewPaymentTokenChosen,
 	} = useSelectedPaymentMethod();
 
+	// This runs once to populate the phone number state if already present in the DOM.
+	useEffect( () => {
+		setPhoneNumber(
+			document.getElementById( 'billing_phone' )?.value ?? ''
+		);
+	}, [] );
```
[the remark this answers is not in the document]
elazzabi: Do you mean the phone input field is not populated and we can get rid of this, or this gets fired before the population of the field? 🤔

elazzabi:
Thank you @Mayisha for the review! Sorry, it took me a while to respond as I was AFK.

I fixed the bugs you raised, and I asked a few questions for the remaining one. Thank you for your help!

client/components/platform-checkout/save-user/checkout-page-save-user.js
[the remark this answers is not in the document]
Mayisha: >  or this gets fired before the population of the field? 

Yes, this statement is right.

So now it's fired on the checkout page load only. Previously it used to fire on mount the phone input component. If I check/uncheck the `save user` checkbox the phone input component mounts/unmounts. 

Consider this flow on trunk-
- checkout page loads and the phone field (the one in the billing address section) is empty initially. The checkbox is also unchecked
- User adds a phone number
- User selects the checkbox
- The phone number input is populated. This happens because the phone number input component mounts on selecting the checkbox, the hook is triggered and takes the value of the `billing_phone`.

Now in this branch-
- checkout page loads and the phone field (the one in the billing address section) is empty initially. The checkbox is also unchecked. But on the checkbox component the hook is triggered and received empty value as `billing_phone`
- User adds a phone number
- User selects the checkbox
- The phone number input field is empty now because the hook does not get triggered on this child component mount and shows the empty string that was initially obtained from the `billing_phone` value on it's parent component.

So the behavior is deviated from how it is on the trunk. Does this make sense? :thinking:

client/components/platform-checkout/save-user/checkout-page-save-user.js
Mayisha: To see the `Remember your details?` section, you need to have `WooPay` enabled on your merchant site. 

First, you need to set it up in WCPay server DB. Follow any one of the following steps-
- Run this command `wp wcpay enable_woopay <merchant-blog-id>` in your WCPay Server's container.

OR

- In WCPay Server's `wcpay_account_meta` table, add a meta key `platform_checkout_eligible` for your WCPay account, with a meta value of `1`. To find your WCPay account ID, look at the `wcpay_accounts` table.

Then on your merchant site UI enable **wp-admin > Payments > Settings > Express checkouts > WooPay**

woopay

Finally on the merchant's checkout page input an email id and select credit card as payment method. The checkbox will appear now. :tada:

client/components/platform-checkout/save-user/checkout-page-save-user.js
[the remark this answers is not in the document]
elazzabi: Thank you for the detailed steps, I can reproduce it now and I think it's fixed by https://github.com/Automattic/woocommerce-payments/pull/4355/commits/9b934c8939f9e893b96441dd9e72b9799e1cec08

Mayisha:
Thanks @elazzabi for addressing the issues :rocket: 

I have tested this again on the checkout page and all the mentioned issues have been addressed and working perfectly. :tada: 

I noticed one new issue, I can not clear the phone field up to the last digit now in order to input a completely new number. It would be great if this one is also handled :eyes: 

can not clear

elazzabi:
Thank you @Mayisha for the review and catching the issue! I think this one is fixed now

https://user-images.githubusercontent.com/12542046/177749408-53184b6b-cca0-4952-9478-b32e3b408cda.mov

Also:
- I switched the component to TypeScript
- I fixed tests and added a new one

client/settings/phone-input/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+		label: string;
+		ariaLabel: string;
+		name: string;
+	};
+}
+
+const PhoneNumberInput = ( {
+	onValueChange,
+	value,
+	onValidationChange = ( validation ) => validation,
```
elazzabi, line 30: This technically could be an empty function, but TypeScript keeps complaining about an empty function

client/settings/phone-input/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+	// Wrapping this in a div instead of a fragment because the library we're using for the phone input
+	// alters the DOM and we'll get warnings about "removing content without using React."
+	return (
+		<div>
+			<input
+				type="tel"
+				ref={ inputRef }
+				value={ removeInternationalPrefix( value ) }
+				onChange={ handlePhoneNumberInputChange }
+				aria-label={
```
elazzabi, line 112: FYI: After using TypeScript I ditched the `label` attribute. It's not a correct HTML attr for the `input` element and it was raised by TypeScript. The correct one is `aria-label`

client/settings/phone-input/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+ */
+import React, { useEffect, useState, useRef } from 'react';
+import { __ } from '@wordpress/i18n';
+import intlTelInput from 'intl-tel-input';
+import './style.scss';
+
+/**
+ * Internal dependencies
+ */
+// @ts-ignore: Cannot find module
```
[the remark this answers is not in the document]
Mayisha: Probably adding the path in `tsconfig.json` could help? :thinking:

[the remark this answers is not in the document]
mgascam: It seems that this dependency is not using TS and the compiler is failing to find the type definition. There is [a way to declare it and make the compiler happy](https://www.typescriptlang.org/docs/handbook/modules.html#working-with-other-javascript-libraries).

You can add a `declarations.d.ts` in the folder of the component. In this file put the following:

`declare module 'iti/utils';`

Note that this is the shorthand notation for module declaration. Anything imported from that module will be considered to be of `any` type. You can also define the types that the module exports if you need. There are examples in the codebase you can check like [here](https://github.com/Automattic/woocommerce-payments/blob/a703cbd14f375bfa2bc93d8b1d1b3cdddd716ab8/client/components/deposits-information/declarations.d.ts)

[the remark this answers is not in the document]
elazzabi: Thanks @mgascam ! It looks like it's a two step process, and the second part involves what you shared ^

So, first part was adding a `path` to tell TypeScript that `iti/utils` => `../node_modules/intl-tel-input/build/js/utils`.
The second part is declaring the module given no type is exported by the package.

Added here https://github.com/Automattic/woocommerce-payments/pull/4355/commits/cb980c3b3e7b8a1dc6baa8765d4255f1fc604860

client/settings/phone-input/iti.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'iti/utils';
```
mgascam, line 1: These files are usually named `declarations.d.ts` in the code base. Maybe you'll want to stick to that naming?

client/settings/phone-input/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+				customPlaceholder: () => '',
+				separateDialCode: true,
+				hiddenInput: 'full',
+				utilsScript: utils,
+			} );
+			setInputInstance( iti );
+			onValidationChange( iti.isValidNumber() );
+
+			// Focus the phone number input when the component loads.
+			currentRef.focus();
```
mgascam, line 82: In the Card Readers page, when the form loads  the input phone component gets the focus. It feels a bit weird since the phone input is not the first in the form. Not sure how hard it can be to handle it in a different way, without giving the focus to the component?

client/settings/phone-input/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		return phone;
+	};
+
+	useEffect( () => {
+		let iti: intlTelInput.Plugin | null = null;
+		const currentRef = inputRef.current;
+
+		const handleCountryChange = () => {
```
mgascam, line 63: I think I found a bug related to the country change. Steps to reproduce:

1. Go to Payments > Card Readers > Receipt details
2. Remove some characters from the input component so you see the invalid phone error
3. Click on the country selector and change the country
4. Observe the issue

**Actual behavior**
When the country changes the phone number is rendered incorrectly. It includes the phone prefix as part of the number. 
There is a warning thrown to the browser console:
`react_devtools_backend.js:4026 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
 `
In some ocassions I noticed that the page becomes unresponsive and eventually crashes. Unfortunately I could not find the exact combination of steps to reproduce, but I think it can be reproduced playing with the component (edit phone number and change country several times)

I recorded this [video](https://d.pr/v/RlpfR5) that shows the issue in action. Please let me know if I can give any further details

client/settings/phone-input/index.tsx
elazzabi: Oh yes, I agree. This was part of [another review](https://github.com/Automattic/woocommerce-payments/pull/4355#discussion_r903421420) but I missed it. Thank you for pointing it out!

client/settings/phone-input/index.tsx
elazzabi: Aha, weird bug. Thank you for pointing it out!

That doesn't happen on the Checkout page, so I believe it's something specific to the Card Readers page.

Looking...

mgascam: Hey @elazzabi I think the problem could be related to the useEffect dependency array. Now we have:

```
useEffect( () => {
		// ...
	}, [ onValueChange, onValidationChange ] );
```
Since both dependencies are functions I suspect the reference changes on every render, causing the useEffect to run again. I used a different approach that seems to work, maybe you'd like to give it a try:

```
useEffect( () => {
		// ...
	}, [ inputRef ] );
```
With this approach, the effect will run if inputRef changes which I think is correct.

elazzabi: Thank you for the pointers @mgascam! I took a look and, changing the dependencies of `useEffect` to `inputRef` fixes it. However, that fails the lint process given `useEffect` uses both `onValueChange` and `onValidationChange`, and thus, they need to appear there.

So, I left them both but looked deeper into what was causing the issue. It appears that `onValidationChange` is changing on every render, and thus makes `useEffect` re-run on every render.

I wrapped a few functions with `useCallback` to avoid the issue. Now it passes, but the new problem is the input loses focus on every key stroke 😅

client/card-readers/settings/sections/contacts-details.js
```diff
@@ -69,21 +85,28 @@ const ContactDetailsSection = ( { setSaveDisabled } ) => {
 				value={ accountBusinessSupportEmail }
 				onChange={ setAccountBusinessSupportEmail }
 				type="email"
-				onBlur={ updateSaveButtonAvailability }
```
elazzabi, line 72: This is moved from here to the `useEffect` ^

client/settings/phone-input/index.tsx
elazzabi: Ok, back to this.

The problem with this is that the input gets re-rendered on each keystroke. This is derived from a long chain of changes:

Changes to the input are tracked via the `accountBusinessSupportPhone` variable. Any change to this variable will trigger a re-calculation of the updateSaveButtonAvailability [function](https://github.com/Automattic/woocommerce-payments/pull/4355/commits/a8b78b0adc91c15a6d5d85cb65f30799e1a83826#diff-606705683552d06907190d85b12172ed4482010bf9f80f5c39ebc94636458055R67). Each change to this function will trigger a re-calculation of the `onValueChange` [prop](https://github.com/Automattic/woocommerce-payments/pull/4355/commits/a8b78b0adc91c15a6d5d85cb65f30799e1a83826#diff-606705683552d06907190d85b12172ed4482010bf9f80f5c39ebc94636458055R107). And finally, each change to the prop will re-trigger the `useEffect` in the input component. Inside the useEffect, we attach the library to the input. So every time we execute it, the library is re-attached and the focus is lost.

The real problem is I made the `onValueChange` make so many things, which is not ideal, my bad. Especially, [the part where it checks if the Save button should be enabled](https://github.com/Automattic/woocommerce-payments/pull/4355/commits/75602c6dc37f9063349da4d104e3ab99c64081b0#diff-606705683552d06907190d85b12172ed4482010bf9f80f5c39ebc94636458055L105). Now, it is fixed to only change the value. Checking for the button enable/disable state is [happening outside](https://github.com/Automattic/woocommerce-payments/pull/4355/commits/75602c6dc37f9063349da4d104e3ab99c64081b0#diff-606705683552d06907190d85b12172ed4482010bf9f80f5c39ebc94636458055R59). This has the benefit of being independent and will be used by [other inputs too](https://github.com/Automattic/woocommerce-payments/pull/4355/commits/75602c6dc37f9063349da4d104e3ab99c64081b0#r919136794).

client/card-readers/settings/sections/contacts-details.js
```diff
@@ -69,21 +85,28 @@ const ContactDetailsSection = ( { setSaveDisabled } ) => {
...
-			<TextControl
-				className="card-readers-business-phone-input"
-				label={ __( 'Support phone number', 'woocommerce-payments' ) }
-				value={ accountBusinessSupportPhone }
-				onChange={ setAccountBusinessSupportPhone }
-				type="tel"
-				onBlur={ updateSaveButtonAvailability }
-			/>
+			<div className="card-readers-business-phone-input">
+				<span>
```
mgascam, line 95: I guess this could be `label`. I noticed there is less margin between text and the input:

Screen Shot 2022-07-13 at 10 44 27

elazzabi: Oh, missed this!

I wanted to hard-code a margin, but then made a quick search and found the `BaseControl` exposed from `@wordpress/components`. I ended up wrapping the component with it https://github.com/Automattic/woocommerce-payments/pull/4355/commits/b331184c9a2f5367faf0668906bffcf4b48362ad

client/settings/phone-input/index.tsx
mgascam: Nice! I am not able to reproduce the bug anymore, well done

mgascam (approved):
LGTM! Tested the new component on Card Readers settings page and WooPay checkout ✅

mordeth (approved):
LGTM! ✅

PR #4357, opened 2022-06-17
Include WCPay's version when making requests to the Platform Checkout

a-danae:
Fixes #1013 from Platform Checkout

Include WCPay's version when making requests to the Platform Checkout.

This PR adds a parameter with the WCPay version to all requests we make to the Platform Checkout.

I evaluated sending it as a header, but we needed to pass this custom header to the `rest_allowed_cors_headers` on the Platform Checkout's end to allow it. I ended up not doing it because it looks like a request param would be enough and would require fewer changes.

#### Testing instructions

* Make sure the platform checkout is enabled in WCPay
* Add a product to the cart and proceed to checkout
* Open the network tab in the browser's dev tools
* Add an email address that exists in the platform checkout
* Look for the request to `/user/exists`
  * Confirm it has a `wcpay_version` parameter with the WCPay version as its value
* Inspect the OTP iframe and look for its `src` attribute
  * Confirm it has a `wcpayVersion` parameter with the WCPay version as its value
* Enter a valid OTP and wait for the redirection
  * Confirm the call to the `init` endpoint includes a `wcpay_version` parameter in its body with the WCPay version as its value. You can check this in the platform checkout's callback for the `/init` endpoint 

Also, please confirm we're not making other requests to the platform checkout. All requests should include this parameter.

-------------------

**Post merge**

### DIFF
--- a/changelog/add-wcpay-version-to-platform-checkout-requests
+++ b/changelog/add-wcpay-version-to-platform-checkout-requests
@@ -1,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Include the WCPay version in the requests to the Platform Checkout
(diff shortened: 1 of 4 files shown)


alefesouza (approved):
LGTM :shipit:

PR #4383, opened 2022-06-30
Fix display of UPE payment methods when manual capture is enabled

### LINKED ISSUE #4382: Payment methods not correctly shown when first enabling UPE with manual capture enabled
### Describe the bug
The UPE methods are not displayed correctly when we try to enable UPE while manual capture is enabled.

### To Reproduce
- Enable manual capture
- Enable UPE: you are shown a page "Add additional payment methods" 
-  Notice that the payment methods are not displayed correctly in the "Boost your sales with payment methods" section
   - the name is missing
   - the logo is missing
   - the tooltip is showing [Object object] instead of the payment method's name

This is a side-effect of https://github.com/Automattic/woocommerce-payments/pull/4245.

Another issue I noticed, which is not related to the above PR: we have the same tooltip issue when a payment method is pending approval (`pending` status for capability). This can also be fixed.

### Actual behavior
Add additional payment methods

### Expected behavior
We are able to see each payment method's name and logo when manual capture is enabled. 
The tooltip should use the payment method's name.

jbordonado:
Fixes #4382 

Manage correctly the UPE payment methods display when it has to be visually hidden and UPE just got enabled.

The Payment > Settings page already manages that display (see https://github.com/Automattic/woocommerce-payments/pull/4245) but we missed the initial page shown when we first enable UPE.

That means:
- we show the right text in the tooltip
- we display the logo and name of each payment method
- we add an overlay to make it clear that the payment methods won't be available right away

### Manual capture enabled / UPE being enabled:

Before:
image

After:
image

### Manual capture enabled / UPE being enabled with payment methods pending verification/activation

Before:
image

After:
image

#### Testing instructions

- Go to Payment > Settings and enable manual capture
- Click on "Enable in your store" to enable UPE
- You should see the name and logo of each payment method, and the corresponding tooltip attached to the warning sign also shows the right name

In `client/components/payment-methods-checkboxes/payment-method-checkbox.js`, add `status = upeCapabilityStatuses.PENDING_VERIFICATION;` line 67 to simulate that the UPE methods are pending activation.

If you follow the same steps as above, the display should also be correct (check the tooltip attached to "Pending activation").

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-4382-upe-payment-method-display-with-manual-capture
+++ b/changelog/fix-4382-upe-payment-method-display-with-manual-capture
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Correctly show UPE payment methods when UPE is first enabled while manual capture is already enabled
--- a/client/components/payment-methods-checkboxes/payment-method-checkbox.js
+++ b/client/components/payment-methods-checkboxes/payment-method-checkbox.js
@@ -64,13 +64,13 @@ const PaymentMethodCheckbox = ( { onChange, name, checked, fees, status } ) => {
 
 	const [ isManualCaptureEnabled ] = useManualCapture();
 	const paymentMethod = PaymentMethodsMap[ name ];
-	const hasOverlay =
+	const needsOverlay =
 		isManualCaptureEnabled && ! paymentMethod.allows_manual_capture;
 
 	return (
 		<li
 			className={ classNames( 'payment-method-checkbox', {
-				'has-overlay': hasOverlay,
+				overlay: needsOverlay,
 			} ) }
 		>
 			<LoadableCheckboxControl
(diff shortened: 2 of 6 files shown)


jessy-p:
Tested and works fine. LGTM!
Nitpick: The classname 'has-overlay' confused me because we are already on a pop-up or overlay. Maybe something like 'disable' or even 'overlay' (so its not about checking an overlay) ?

jbordonado:
> Tested and works fine. LGTM! Nitpick (feel free to ignore, might be just me) : The classname 'has-overlay' confused me because we are already on a pop-up or overlay. Maybe something like 'disable' or even 'overlay' (so its not about checking an overlay) ?

Rewording the class doesn't hurt, I changed it from `has-overlay` to `overlay` in both components that used it (sadly, the code is duplicated, a refactor would be nice in another issue) in https://github.com/Automattic/woocommerce-payments/pull/4383/commits/04708e1b133166c902e0b1c02bf43271312906f2.

Do you mind checking again just in case? And then I'll go ahead and merge it.
Thanks!

PR #4384, opened 2022-06-30
Update explicit currency formats in admin

ismaeldcom:
Fixes #

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-explicit-price-formatter.php
+++ b/includes/class-wc-payments-explicit-price-formatter.php
@@ -46,50 +46,37 @@ class WC_Payments_Explicit_Price_Formatter {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


dmallory42:
> Especially thinking of those merchants that can place a couple of orders using other currencies or just decide to not use them anymore, and would like to get rid of explicit format. What do you think?

I think that even better is that we give the user a choice on whether to display explicit formatting (but force this behaviour when more than one currency is enabled)?

But adding that seems outside the scope of this particular issue, so I think if we chose to do that it could be a future follow up.

dmallory42 (approved):
LGTM, tested and worked as describes.

Thanks for working on this, let's 🚢 !

PR #4390, opened 2022-07-01
Add Customer Currency filter to transactions advanced filters

### LINKED ISSUE #1278: Transactions list: Add customer currency filter
Depends on 498-gh-Automattic/woocommerce-payments-server

Allow filtering by customer currency via advanced filters:

customer-currency-filter

@LevinMedia in https://github.com/Automattic/woocommerce-payments/issues/1202#issuecomment-763108784:

> The drop down filter should only list currencies that have been used by customers on the store - IE there's a transaction record of such and such a currency.
> 
> Bonus points if we can disable selections if that currency isn't present in the currently selected time period. That way people aren't hunting and pecking through currencies to see what may or may not be there...

As I'd noted in a subsequent [comment](https://github.com/Automattic/woocommerce-payments/issues/1202#issuecomment-765233464) there, we'll need to query all of the orders (or possibly transactions, on the server) for _distinct currencies_, either on page load or when the filter is added. I'm not sure if there's any precedent for the filters UI in that component being dynamically computed (and depending on other filters that have been added) though it seems like it should be _possible_ via the [`onAdvancedFilterAction` prop](https://github.com/woocommerce/woocommerce-admin/blob/d830378e8ca783449c739ab5b770fa819542ade4/packages/components/src/filters/index.js#L192-L195). (Alternatively, the Multi-currency extension – for instance – has its own list of _enabled_ currencies, but that could diverge from currencies that have actually been used.)

dmallory42:
Fixes #1278

- Adds a way to fetch a list of all the currencies that have ever been used in a customer's store
- Adds a filter to allow filtering by customer currency.

#### Testing instructions

- Checkout 2176-gh-Automattic/woocommerce-payments-server on your local server and run `npm run listen`.
-`npm run watch` on the client
- Make a couple of transactions on your store in various currencies.
- Go to `Payments > Transactions`, open the advanced filters menu and test the `Customer currency` filter, using both the `is` and `is not` filters, ensuring you get the expected result (only orders made in the selected currency should be shown, and when using `is not` only orders made in a different currency from the one selected should be shown.
- Test the other advanced filters to ensure they still work as expected

Will add tests, but wanted to get feedback on the proposed solution first.

#### Questions

- I added the 'Is Not' filter as well. I was thinking this could be useful to see all transactions not in the default store currency, for example. Do we want this?
- I tried to refactor the logic to make it more efficient, but there's a few things I'm unsure about. It would be nice if the customer currencies list is only populated on first page load (rather than being repopulated every time a search is made) since these values will likely not change after the initial load. (although, in theory, more transactions could be made in this time and new currencies may be added to the list).
- Also, the deposit currency filter option repopulates every time the search is made, which I don't think is necessary (to observe this, change the second argument passed to `getFilters` to `true` in the `transactions/filters/index.tsx` file.

-------------------

**Post merge**


client/transactions/filters/index.tsx
```diff
@@ -8,45 +8,46 @@ import { getQuery } from '@woocommerce/navigation';
 /**
  * Internal dependencies
  */
-import { filters, advancedFilters, TransactionsFilterType } from './config';
+import { getFilters, getAdvancedFilters } from './config';
 import { formatCurrencyName } from '../../utils/currency';
 import './style.scss';
+import { __ } from '@wordpress/i18n/build-types';
```
ismaeldcom, line 14: It's not used anymore, so we can remove it.

client/transactions/filters/index.tsx
```diff
@@ -8,45 +8,46 @@ import { getQuery } from '@woocommerce/navigation';
...
-			}
-		} );
-		return filtersConfiguration;
-	};
+	const customerCurrencyOptions = customerCurrencies?.map(
+		( currencyCode: string ) => ( {
+			label: formatCurrencyName( currencyCode ),
+			value: currencyCode,
+		} )
+	);
```
ismaeldcom, line 30: This is a good example where `useMemo` could be used to avoid doing complex operations on every render. It's not that heavy, but we are mapping the array and calling the get filter method to generate it.

We can store it as a constant using a code similar to the following one, and the same will apply to `filters` with `storeCurrencies`.

Be aware, that we should evaluate when `useMemo` should be used, as we can end having a performance hit while trying to improve it, by just using `useMemo` or similar hooks everywhere.

```js
const advancedFilters = useMemo(
    () =>
        getAdvancedFilters(
            customerCurrencies.map( ( currencyCode: string ) => ( {
                label: formatCurrencyName( currencyCode ),
                value: currencyCode,
            } ) )
        ),
    [ customerCurrencies ]
);
```

client/transactions/filters/index.tsx
```diff
@@ -8,45 +8,46 @@ import { getQuery } from '@woocommerce/navigation';
...
+
+	const currencies = storeCurrencies || [];
+	const depositCurrencyOptions = currencies?.map(
+		( currencyCode: string ) => ( {
+			label: formatCurrencyName( currencyCode ),
+			value: currencyCode,
+		} )
+	);
```
ismaeldcom, line 38: If we look at [transactions list definition for `storeCurrencies`](https://github.com/Automattic/woocommerce-payments/blob/7cb4bdaec540b6912d7d45041473bdd35c33e78b/client/transactions/list/index.tsx#L607) we're already providing an empty array as fallback value.

I think we could simplify `TransactionsFilters` code by making both props non-optional and just adding `|| []` for [`customerCurrencies` definition in transactions list.](https://github.com/Automattic/woocommerce-payments/blob/7cb4bdaec540b6912d7d45041473bdd35c33e78b/client/transactions/list/index.tsx#L608)

client/transactions/filters/config.ts
```diff
@@ -52,169 +52,228 @@ const loanSelectionOptions = loanDefinitions.map( ( loanDefinition ) => {
...
-					value: 'is',
-					/* translators: Sentence fragment, logical, "Is" refers to searching for transactions matching a chosen transaction type. */
-					label: _x(
-						'Is',
-						'transaction type',
+	];
+};
+
+/**
+ * TODO: Add an interface here for advanced filters.
```
ismaeldcom, line 113: It could also be solved by adjusting `explicit-module-boundary-types` rule, letting us use TS inferred type. I think we can mention that for future TypeScript work. Something like `... or adjust ESLint rules to allow using inferred type`. More info here: paJDYF-4tS-p2#comment-12406

ismaeldcom (commented):
_I hope I'm not missing anything, because I've written the comments twice. Not sure why GH discarded them after the latest merge commit_ 🤷

Thanks for working on it! Working as expected, and the new filters approach looks quite good 🙂 

I left a few comments, but nothing that really affects the functionality.

BTW I'm not sure if you've tested deposit filters, I couldn't make them to work with or without your PR 🤔 I was just trying to verify if your refactor affected the functionality, but it seems to be broken.

client/transactions/filters/index.tsx
dmallory42: Good catch!

client/transactions/filters/index.tsx
dmallory42: Thanks, I tested this out and it seems to improve the performance (at least from a UI POV) because it means it doesn't reload the list when just choosing a currency on the customer currency filter and clicking filter.

Updated in 453d71d5c510c20e4a47881f4d23fa62c2fb0683.

client/transactions/filters/index.tsx
dmallory42: Thanks, that simplifies the code nicely! I updated and also moved to using `useMemo` in 453d71d5c510c20e4a47881f4d23fa62c2fb0683

client/transactions/filters/config.ts
dmallory42: Thanks! Added a comment in 1ee4f397. I think this is probably one that's better to wait for the planned work to initially define some types (or used inferred types) across the board.

ismaeldcom:
> I added the 'Is Not' filter as well. I was thinking this could be useful to see all transactions not in the default store currency, for example. Do we want this?
I think it could be useful, and as you already added it, I don't see any reason to remove it 🙂 

About the other 2 questions, as we already discussed in Slack (p1657278821767129/1657188059.196539-slack-C03KTTK2YMA), we could move them to a new issue, where we can cache filters between requests.

dmallory42:
> About the other 2 questions, as we already discussed in Slack (p1657278821767129/1657188059.196539-slack-C03KTTK2YMA), we could move them to a new issue, where we can cache filters between requests.

Agreed! I'll create a new issue for this later this afternoon.

And thanks for the review! I've now added the unit tests and I think it's now hopefully ready to :shipit:.

client/transactions/filters/index.tsx
```diff
@@ -1,51 +1,54 @@
...
+
+	const filters = useMemo(
+		() =>
+			getFilters(
+				storeCurrencies?.map( ( currencyCode: string ) => ( {
+					label: formatCurrencyName( currencyCode ),
+					value: currencyCode,
+				} ) ),
+				true
```
ismaeldcom, line 42: Should we keep the previous logic to hide it when there is no additional currencies?

ismaeldcom (approved):
LGTM! I left one comment about the filter, but apart from that everything looks good. Thanks again @dmallory42 for working on it and making all those little changes I suggested.

Let's 🚢  it!

client/transactions/filters/index.tsx
dmallory42: Great catch! I must have switched it over to test and forgot to switch back! 

Updated in 137f29fc and also caught a bug, previously we checked for `> 2` but we now need to check for `> 1` because we don't include the placeholder option at this point (The `All currencies` option).

PR #4392, opened 2022-07-04
Add payment channel information + payment identifiers to transactions

### LINKED ISSUE #4101: In-Person Payments: ensure the proper logo is displayed in listings for IPP/Interac
### Description

We need to clarify and ensure which logo should be displayed when paying with Visa/Master card (incl. Interac).
Context: I couldn't find any related discussions and would like to re-evaluate the status here.

elazzabi:
Fixes #4101 

This PR introduces two changes:
- Channel information (In-Person or Online) to the transactions list + individual transaction info. All transactions marked as `card_present` or `interac_present` are marked as `In-Person`. All others are marked as `Online`.
- Each payment method is now displayed with an identifier (last 4 digits for Cards, Bank details for others)

#### Testing instructions

* This PR should be tested with PR `2215` from the server
* Execute all steps necessary in the server's PR
* Connect an account with some in-person transactions
* Open the Transactions list view
* Make sure transactions have identifiers on payment methods (ie, last 4 digits for card...)
* Make sure there is a new column called 'Channel' and contains either `In-Person` or `Online`
* Click on a transaction and double-check if there is a mention of `In-Person` or `Online` on the individual transaction page

Screen Shot 2022-07-18 at 14 51 16

individual_transaction

-------------------

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : 


client/style.scss
```diff
@@ -111,7 +111,7 @@
 }
 
 .payment-method-details {
-	display: flex;
+	display: inline-flex;
```
elazzabi, line 114: This is to make sure all items are aligned left, instead of center on the page:

Before:
Screen Shot 2022-07-19 at 09 28 35

After:
Screen Shot 2022-07-19 at 09 28 25

client/transactions/list/test/index.tsx
```diff
@@ -613,28 +646,31 @@ describe( 'Transactions list', () => {
 			expect( displayFirstTransaction[ 1 ] ).toBe(
 				csvFirstTransaction[ 2 ]
 			); // type
+			expect( displayFirstTransaction[ 2 ] ).toBe(
+				csvFirstTransaction[ 3 ]
+			); // channel
 			expect(
-				getUnformattedAmount( displayFirstTransaction[ 2 ] ).indexOf(
```
elazzabi, line 617: All columns here are offset by 1 given the new "Channel" column

client/utils/charge/index.ts
```diff
@@ -120,3 +121,15 @@ export const getChargeAmounts = ( charge: Charge ): ChargeAmounts => {
 
 	return balance;
 };
+
+export const getChargeChannel = ( type: string ): string => {
+	if (
+		type === 'card_present' ||
```
elazzabi, line 127: A small detail here. This method is called in two places: The individual transaction page, and the list of transactions page.

In the individual transaction page, we are getting the data from Stripe, so we pass the `transaction.type` which can be `card_present` or `interac_present` for In-Person payments.

In the list of transactions, the `type` holds the brand of the payment method, so we aren't passing it. Instead, we pass the `transaction.channel` directly, which might be `in_person|online`. The argument is called type, which might feel incorrect in this case, so just wanted to point it out.

client/transactions/list/test/index.tsx
```diff
@@ -555,6 +587,7 @@ describe( 'Transactions list', () => {
 				'"Transaction Id"',
 				'"Date / Time"',
 				'Type',
+				'Channel',
```
kalessil, line 590: @elazzabi: since we introduced a new column, is it included in CVS export (both client-side and server-side)?

client/utils/charge/index.ts
```diff
@@ -120,3 +121,15 @@ export const getChargeAmounts = ( charge: Charge ): ChargeAmounts => {
...
+		type === 'interac_present' ||
+		type === 'in_person'
```
kalessil, line 129: Not sure where `in_person` might come from, can you elaborate a little bit @elazzabi ?

elazzabi: Yes! I added a small comment on the line above https://github.com/Automattic/woocommerce-payments/pull/4392#discussion_r924218971. I can change course (or add a code comment) if it can clear things up

client/transactions/list/test/index.tsx
elazzabi: Yes. It is covered by the test here, but I made a manual test
Screen Shot 2022-07-19 at 14 49 49

client/utils/charge/index.ts
kalessil: Thanks! Can we place the comment right in the codebase?

elazzabi: Sounds good! I added a comment to the method https://github.com/Automattic/woocommerce-payments/pull/4392/commits/54d99d10acac7529dbb6374de186599658bbe71b

mgascam:
@elazzabi I am not sure if I found a bug. I am testing using the server PR and this one. I followed testing instructions and I am able to see the new columns in db. I was also able to sync my test transactions. But I noticed that, after synching the transactions, all of them have the Channel set to 'Online', even the In person ones. I took some screenshots:

**Transaction is correctly stored in the DB**
Screen Shot 2022-07-20 at 18 31 29

**It has wrong Channel (Online) in the transactions page**
Screen Shot 2022-07-20 at 18 30 47

**Channel shows correctly in the detail view**
Screen Shot 2022-07-20 at 18 31 06

Can you have a look please?

elazzabi:
@mgascam Oh interesting, I couldn't reproduce this on my end 🤔

Two things I noticed in your screenshots, however:
- The transactions list only shows the brand logo, where it should show the logo + the last 4 digits
- The default behavior is to show `Online` for transactions

All this to say that _maybe_ the data isn't getting correctly to the frontend. Or, the frontend isn't correctly built 🤔 

Can you please double check the Network tab and see if the corresponding  transaction has a `channel: in_person`? Or force build with `npm start`?

Thank you!

mgascam:
> @mgascam Oh interesting, I couldn't reproduce this on my end 🤔
> 
> Two things I noticed in your screenshots, however:
> 
> * The transactions list only shows the brand logo, where it should show the logo + the last 4 digits
> * The default behavior is to show `Online` for transactions
> 
> All this to say that _maybe_ the data isn't getting correctly to the frontend. Or, the frontend isn't correctly built 🤔
> 
> Can you please double check the Network tab and see if the corresponding transaction has a `channel: in_person`? Or force build with `npm start`?
> 
> Thank you!

@elazzabi thanks for the insights. I think the problem was the front end. I restarted everything and now it shows the correct Channel and also the logo + last 4 digits:

Screen Shot 2022-07-21 at 11 06 10

It's a bit odd, since I was able to see the Channel column in the table but maybe something was cached?

About logo + last 4 digits, this is how a Link payment looks like:

Screen Shot 2022-07-21 at 11 11 51

I am not sure if we should omit the last 4 in this case?

mgascam (approved):
LGTM! I followed instructions and tests well. 

I left a comment about how the StripeLink is displayed that you may want to look into.

kalessil:
I noticed a couple of things:

- On the transaction details page under FF layout in the, ''bottom sections is not looking good:
Screenshot from 2022-07-21 15-17-03
- Also, in the third block, the 'Fingerprint' field should be removed

Other than that that works great!

client/data/transactions/hooks.ts
```diff
@@ -37,6 +37,8 @@ export interface Transaction {
 	transaction_id: string;
 	date: string;
 	type: 'charge' | 'refund' | 'financing_payout' | 'financing_paydown';
+	channel: 'in_person' | 'online';
+	source_identifier: string;
 	source: string;
```
kalessil, line 42: Should we list here possible options?

client/types/charges.d.ts
```diff
@@ -28,7 +28,7 @@ interface ChargeRefunds {
 
 export interface PaymentMethodDetails {
 	card?: any;
-	type: 'card';
+	type: 'card_present' | 'card';
```
kalessil, line 31: What about 'interac_present'?

elazzabi:
Thank you for the review @kalessil 

I can't reproduce the alignment problem, it looks like this on my end

Screen Shot 2022-07-21 at 16 12 04

Regarding removing the fingerprint, is this part of the work on this PR?

client/data/transactions/hooks.ts
elazzabi: Oh, that would be amazing! Our source is either the card brand or the type of payment, so that would be: `amex`, `diners`, `discover`, `jcb`, `mastercard`, `unionpay`, `visa`, `link`, `au_becs_debit`, `bancontact`, `eps`, `giropay`, `ideal`, `p24`, `sofort`, `sepa_debit`. Right?

client/types/charges.d.ts
elazzabi: I've found [this](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-details/payment-method/index.js#L47) and [this](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-details/payment-method/index.js#L24).

So it looks like it's missing more than interac. Should I add all of them? 🤔

kalessil:
> Regarding removing the fingerprint, is this part of the work on this PR?

Yes.

> I can't reproduce the alignment problem, it looks like this on my end

I can retry later today. Also, note that 'ID' value overlaps into the 2nd column.

client/types/charges.d.ts
kalessil: Add as many as you are aware of =)

client/data/transactions/hooks.ts
kalessil: See https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-type.
There `card`, `card_present`, `interac_present` we replace with card brands (as you list them in your list) + we support `link` ofcourse.

anu-rock: We have an ENUM in the corresponding column in DB that would also require updating to keep things truly in sync. However, I think these are too many options (above). And the list may change sooner than later. How about we keep things more flexible here by dropping the use of enums?

client/transactions/list/index.tsx
```diff
@@ -73,6 +77,26 @@ interface Column extends TableCardColumn {
...
+		case 'p24':
+			return (
+				<Fragment>
+					{ p24BankList[ txn.source_identifier ] ?? '' }
+				</Fragment>
+			);
+		default:
+			return (
+				<Fragment>
+					&nbsp;&bull;&bull;&bull;&bull;&nbsp;{ ' ' }
```
anu-rock, line 93: Can we replace these html entities with CSS? ;)

client/data/transactions/hooks.ts
```diff
@@ -37,6 +37,8 @@ export interface Transaction {
 	transaction_id: string;
 	date: string;
 	type: 'charge' | 'refund' | 'financing_payout' | 'financing_paydown';
+	channel: 'in_person' | 'online';
+	source_identifier: string;
```
anu-rock, line 41: Perhaps we can add a small comment to explain this field?

client/data/transactions/hooks.ts
kalessil: The source column in DB is `varchar(20)`, and their enums are not needed exactly because of how often we add new values here.

> How about we keep things more flexible here by dropping the use of enums?

@anu-rock, can you elaborate a little bit more?

anu-rock: I'm sorry, I think I mixed channel and source. Please ignore my last comment :)

client/transactions/list/index.tsx
elazzabi: I saw we use HTML in other places, so I think it's better to stick to that if there is no big difference in performance...

elazzabi:
@mgascam thank you for the review. That led me to a rabbit hole of debugging StripeLink but I at least have a fix 😅. It looks like StripeLink doesn't expose card data, and I believe other payment providers do not too. That's why I added a check to return an empty element if no information is available https://github.com/Automattic/woocommerce-payments/pull/4392/commits/9ce32ac6da4ff4e8cb002c15cd095daa6eb7b81e

 @kalessil Thanks for the review! I removed the Fingerprint field, still debugging the overlap in the ID field

@anu-rock Thanks for the review! I fixed most of them :)

kalessil (approved):
LGTM!

EduardoUmpierre:
> Thank you for the review @kalessil
> 
> I can't reproduce the alignment problem, it looks like this on my end
> 
> Screen Shot 2022-07-21 at 16 12 04
> 
> Regarding removing the fingerprint, is this part of the work on this PR?

It used to look like this on 4.5.0:

image

I'm going to address this issue by creating different classes for the transaction list and the transaction details page. It will be part of [Remove Charge request from Transaction Details page](https://github.com/Automattic/woocommerce-payments/pull/4516#top).

kalessil:
Super, thank you @EduardoUmpierre !

PR #4427, opened 2022-07-11
Register only reusable payment methods when adding a payment method

### LINKED ISSUE #4396: Multiple UPE - payment method does not support recurring payments error in my account
### Description

This is one of the TODOs from https://github.com/Automattic/woocommerce-payments/pull/3968#issuecomment-1156928898. 

When we use the `poc/upe-instances-multiplied` branch, this error appears on the my account page:

`This payment method does not support recurring payments.`

This is due to all payment gateways appearing on the page when only the card element should appear. We'll need to filter out gateways that don't support tokenization when [register gateways](https://github.com/Automattic/woocommerce-payments/blob/f11e1556f6ae114e3abcec971639e7337e05b724/includes/class-wc-payments.php#L491). We can check if we're on the Add Payment Page and if the payment method is reusable.

This ticket is to investigate and fix the error.

### Acceptance Criteria
- `This payment method does not support recurring payments.` should not show up.
- A new payment method can be successfully added.
- changes made should be merged into `poc/upe-instances-multiplied`, not `develop`.

mdmoore:
With multiple gateways being registered, the Add Payment Method page is broken with the error, `This payment method does not support recurring payments.` and new payment methods cannot be added. 

Fixes #4396 

This ensures that only reusable payment methods are available on the **My Account > Payment Methods > Add Payment Method** page rather than all UPE methods. 

#### Testing instructions

1. After checking out this branch, run `npm run start`, enable WooPay and the UPE.
2. Enable a few UPE payment methods in the WCPay settings.
3. Add a product to the cart and verify that both the CC and enabled UPE methods are present on checkout
4. Go to **My Account > Payment Methods > Add Payment Method** and verify that no error is present on the page and that a new payment method can be added.

-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe.js
+++ b/client/checkout/classic/upe.js
@@ -172,6 +172,8 @@ jQuery( function ( $ ) {
 		const gatewayCardId = getUPEConfig( 'gatewayId' );
 		let selectedGatewayId = null;
 
+		// Handle payment method selection on the Checkout page or Add Payment Method page where class names differ.
+
 		if ( $( 'li.wc_payment_method' ).length ) {
 			selectedGatewayId = $(
 				'li.wc_payment_method input.input-radio:checked'
--- a/client/checkout/classic/upe.js
+++ b/client/checkout/classic/upe.js
@@ -170,11 +170,24 @@ jQuery( function ( $ ) {
 	 */
 	const getSelectedGatewayPaymentMethod = () => {
 		const gatewayCardId = getUPEConfig( 'gatewayId' );
-		const selectedGatewayId = $(
-			'li.wc_payment_method input.input-radio:checked'
-		).attr( 'id' );
+		let selectedGatewayId = null;
+
+		if ( $( 'li.wc_payment_method' ).length ) {
+			selectedGatewayId = $(
+				'li.wc_payment_method input.input-radio:checked'
+			).attr( 'id' );
+		} else if ( $( 'li.woocommerce-PaymentMethod' ).length ) {
+			selectedGatewayId = $(
+				'li.woocommerce-PaymentMethod input.input-radio:checked'
+			).attr( 'id' );
+		}
+
+		if ( 'payment_method_woocommerce_payments' === selectedGatewayId ) {
+			selectedGatewayId = 'payment_method_woocommerce_payments_card';
+		}
 
 		let selectedPaymentMethod = null;
+
 		for ( const paymentMethodType in paymentMethodsConfig ) {
 			if (
 				`payment_method_${ gatewayCardId }_${ paymentMethodType }` ===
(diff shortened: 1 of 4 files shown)


harriswong:
👀

harriswong:
Also, do we need to worry about the failed tests, or is this something we need to address separately in the `poc/upe-instances-multiplied` branch?

harriswong:
In my http://localhost:8082/my-account/add-payment-method/, 
image

I see credit card and SEPA, but I don't see giropay and sofort. My checkout page does show those though. Is this the same case for you? 
image

mdmoore:
It turns out that the selectors used for payment gateways on the Add Payment Method page differs from those used on checkout and other pages, so `getSelectedGatewayPaymentMethod()` wasn't returning the correct payment method. 

@harriswong After rebuilding JS, this should be working well now, as in, both credit card and SEPA payment methods can be added. It does surface another issue where SEPA tokens are appearing in the list of credit card tokens(where they've always appeared because it was the only place) when they should appear inside the SEPA gateway. That warrants a separate issue so I think we can push forward with this one.

client/checkout/classic/upe.js
```diff
@@ -170,11 +170,24 @@ jQuery( function ( $ ) {
...
-		const selectedGatewayId = $(
-			'li.wc_payment_method input.input-radio:checked'
-		).attr( 'id' );
+		let selectedGatewayId = null;
+
+		if ( $( 'li.wc_payment_method' ).length ) {
+			selectedGatewayId = $(
+				'li.wc_payment_method input.input-radio:checked'
+			).attr( 'id' );
+		} else if ( $( 'li.woocommerce-PaymentMethod' ).length ) {
```
harriswong, line 179: Ah I see, this is used inside http://localhost:8082/my-account/add-payment-method/. I was trying to look for this on the http://localhost:8082/checkout/ but I can't find it. 

I wonder if it's worth adding a comment here so we don't accidentally remove it in the future.

client/checkout/classic/upe.js
```diff
@@ -743,15 +755,6 @@ jQuery( function ( $ ) {
 
 	// Handle the add payment method form for WooCommerce Payments.
 	$( 'form#add_payment_method' ).on( 'submit', function () {
-		if (
-			'woocommerce_payments' !==
-			$(
-				"#add_payment_method input:checked[name='payment_method']"
```
harriswong, line 749: Are we removing this because in http://localhost:8082/my-account/add-payment-method/, it will no longer be only "Credit Card/debit card"?

image

Do you know why we needed to `return;` early before?

includes/class-wc-payments.php
```diff
@@ -491,15 +491,27 @@ public static function add_plugin_links( $links ) {
...
+				if ( $upe_payment_method->is_reusable() ) {
+					$reusable_methods[] = $upe_gateway;
+				}
+
+				$gateways[] = $upe_gateway;
+
+			}
+
+			if ( is_add_payment_method_page() ) {
+				return $reusable_methods;
```
harriswong, line 514: I think it makes sense that http://localhost:8082/my-account/add-payment-method/ only display payment methods that are saved, but not every single payment method we support. 👍

harriswong (approved):
I tested this on both checkout and my `my-account/add-payment-method/`. I have some questions but LGTM otherwise!

client/checkout/classic/upe.js
mdmoore: Yes! This was a head-scratcher for me as well. [Comment added](https://github.com/Automattic/woocommerce-payments/pull/4427/commits/a5e6b957163c012a190fbbf7704dc1dafe67a555).

PR #4429, opened 2022-07-12
Make adding fee details to order notes async

### LINKED ISSUE #4439: Make adding fee details to order notes async
### Description

Currently, we add fee details to order notes in a sync manner during checkout flow which in turn makes a timeline call, which is expensive. Make this flow async to remove the timeline call during the checkout flow.

### Acceptance criteria

* Get timeline call shouldn't happen during checkout.

### Additional context

Discussion: pdpOw2-Sb-p2#comment-1140

deepakpathania:
Fixes #4439 

- Makes adding fee details to order notes async.
- Removes the get timeline call from the checkout flow for performance benefits, ref: pdpOw2-Sb-p2#comment-1140
- We are adding a new order note instead of updating the existing one since:
  - it might be inconsistent behavior for the end-user to see a note with different information after refreshing in case they visit before and after the async job runs - the queue can take some time to pick the scheduled tasks even if we schedule it immediately.
  - From what I could see, we also don't have any precedence of updating notes in the codebase, but please CMIIAW here.

#### Testing instructions

- Follow the normal checkout flow and look at the order details screen in the WooCommerce tab.
- Ensure there is a separate order note added to the order. Go to Tools > Scheduled actions > search for this job with `wcpay_add_fee_breakdown_to_order_notes` and select `Run` for running the action.

Before:

Screenshot 2022-07-12 at 5 20 36 PM

After:

Screenshot 2022-07-12 at 5 08 10 PM

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : [Testing instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-4.5.0#make-adding-fee-details-to-order-notes-async)

### DIFF
--- a/includes/class-wc-payments-order-service.php
+++ b/includes/class-wc-payments-order-service.php
@@ -23,7 +23,7 @@ class WC_Payments_Order_Service {
 	const STATUS_ON_HOLD   = 'on-hold';
 	const STATUS_PENDING   = 'pending';
 
-	const ADD_FREE_BREAKDOWN_TO_ORDER_NOTES = 'wcpay_add_fee_breakdown_to_order_notes';
+	const ADD_FEE_BREAKDOWN_TO_ORDER_NOTES = 'wcpay_add_fee_breakdown_to_order_notes';
 
 	/**
 	 * Client for making requests to the WooCommerce Payments API
@@ -40,7 +40,7 @@ class WC_Payments_Order_Service {
 	public function __construct( WC_Payments_API_Client $api_client ) {
 		$this->api_client = $api_client;
 
-		add_action( self::ADD_FREE_BREAKDOWN_TO_ORDER_NOTES, [ $this, 'add_fee_breakdown_to_order_notes' ], 10, 2 );
+		add_action( self::ADD_FEE_BREAKDOWN_TO_ORDER_NOTES, [ $this, 'add_fee_breakdown_to_order_notes' ], 10, 2 );
 	}
 
 	/**
@@ -67,7 +67,7 @@ class WC_Payments_Order_Service {
 		// Update the note with the fee breakdown details async.
 		WC_Payments::get_action_scheduler_service()->schedule_job(
 			time(),
-			self::ADD_FREE_BREAKDOWN_TO_ORDER_NOTES,
+			self::ADD_FEE_BREAKDOWN_TO_ORDER_NOTES,
 			[
 				'order_id'  => $order->get_id(),
 				'intent_id' => $intent_id,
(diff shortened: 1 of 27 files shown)


includes/class-wc-payments-order-service.php
```diff
@@ -54,12 +59,26 @@ public function mark_payment_completed( $order, $intent_id, $intent_status, $cha
...
 
-		$note = $this->generate_payment_success_note( $intent_id, $charge_id, $this->get_order_amount( $order ) );
+		$note = $this->generate_payment_success_note( $charge_id, $this->get_order_amount( $order ) );
 
 		if ( $this->order_note_exists( $order, $note ) ) {
 			return;
 		}
 
+		// Dispatch call here to update the order note with fee breakdown.
+		$action_scheduler_service = new WC_Payments_Action_Scheduler_Service( $this->api_client );
```
htdat, line 69: I think we might avoid initializing `WC_Payments_Action_Scheduler_Service` here as it can create a side effect (see its `__construct` that includes `add_action` functions). 

IMO, there are two ways for avoiding this: 

(1) Create a static function in `WC_Payments::get_action_scheduler_service` (which is similar to multiple static functions like `WC_Payments::get_localization_service`, and use it here.


(2) Add `WC_Payments_Action_Scheduler_Service` as a dependency for this Order_Service class, and register it in WC_Payments  - [src](https://github.com/Automattic/woocommerce-payments/blob/72f6bb76ecd432f09b598e2485aaf012bcf9a6d7/includes/class-wc-payments.php#L303). This requires multiple changes in our tests too because the class signature has changed. But the good thing is that it will be helpful when we need to refactor our code towards DI. 

I think using approach (1) is reasonable here, though (2) is more favorable in the long term.

deepakpathania: Updated this to use approach 1.

htdat (approved):
Thanks for address my feedback. 

Tested again and it worked well.

LGTM!

includes/class-wc-payments-order-service.php
```diff
@@ -23,6 +23,8 @@ class WC_Payments_Order_Service {
 	const STATUS_ON_HOLD   = 'on-hold';
 	const STATUS_PENDING   = 'pending';
 
+	const ADD_FREE_BREAKDOWN_TO_ORDER_NOTES = 'wcpay_add_fee_breakdown_to_order_notes';
```
dechov, line 26: Typo "FREE" in constant name

deepakpathania: Good catch, fixed.

PR #4434, opened 2022-07-13
Fix multicurrency shipping tax conversion

### LINKED ISSUE #4417: Wrong shipping tax conversion when using multi-currency
H/T @reykjalin for reporting this in paJDYF-4vd-p2.

### Describe the bug

By default, shipping costs are converted using the currency rounding setting. Shipping taxes, however, are converted directly based on the value calculated with the price in the default store currency. This leads to an incorrect shipping tax when the shipping costs are affected by rounding.

### To Reproduce

1. Add a flat rate shipping method that costs 5 USD
2. Navigate to "WooCommerce > Settings > General" and check "Enable tax rates and calculations"
3. Navigate to "WooCommerce > Settings > Tax" and check "No, I will enter prices exclusive of tax"
4. On that same screen, click on the "Standard rates" tab and insert a new row with a 10% tax
5. Navigate to "WooCommerce > Settings > Multi-currency" and add the Euro currency with a rate of `0.91` and a rounding of `1.00`
6. Switch to Euros and add a physical product to your cart
7. Navigate to the cart or checkout pages
8. Check that the shipping cost is `5 EUR` but the taxes are `0.46 EUR`, instead of the expected `0.5 EUR`

### Actual behavior

The shipping taxes do not take into account any rounding and charm pricing applied when converting the shipping cost.

### Screenshots

Cart with wrong shipping tax

### Expected behavior

The shipping taxes should be correctly applied based on the converted shipping cost, taking any rounding or charm pricing into account.

### Additional context

When working on a fix for this issue, we should make sure that [Per Product Shipping](https://woocommerce.com/products/per-product-shipping/) still works for multiple products with different tax rates.

A potentially simple fix for this would be modifying the [tax conversion from here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/FrontendPrices.php#L149) to calculate the tax rate based on the original cost and apply it to each entry in the `taxes` array, like the untested solution outlined below.

```
$rate          = clone $rate;
$previous_cost = $rate->cost;
...
		function ( $tax ) use ( $rate, $previous_cost ) {
			$tax_rate = $tax / $previous_cost;
			return $rate->cost * $tax_rate;
		},
...
```

ismaeldcom:
Fixes #4417 

Use another filter to ensure shipping rates conversion. With `woocommerce_package_rates` we need to convert both the cost and every tax, and clone the object to avoid multiple conversions. Instead, we could use `woocommerce_shipping_method_add_rate_args` to only convert the cost and let the shipping class calculate the converted tax.

#### Testing instructions
**WooCommerce > Settings**
- **General**
  - Enable tax rates and calculations
  - **Currency**: `United States (US) dollar $`
- **Shipping > Everywhere > Flat Rate**
  - **Tax status**: `Taxable`
  - **Cost**: `5`
- **Multi-currency > Euro**
  - **Exchange rate**: `Manual`
  - **Manual rate**: `0.91`
  - **Price rounding**: `1.00 (recommended)`
  - **Charm pricing**: `none`
- **Tax**
  - **Tax options**
    - **Prices entered with tax**: `No, I will enter prices exclusive of tax `
    - **Display prices during cart and checkout**: `Excluding tax`
  - **Standard rates**

| Country Code | State code | Postcode / ZIP | City | Rate % | Tax name | Priority | Compound | Shipping |
|- |- |- |- |- |- |- |- |- |
| * | * | * | * | 10.0 | Tax | 1 | ☐ | ☑ |

- Add a product to your cart and go there.
- Switch currency to EUR.
- Flat rate shipping should be `5,00 €`.
- Tax should be `0,50 €`.
- Check that shipping and shipping taxes also work as expected with:
  - Subscriptions
  - Advanced shipping rates cost, using formulas like `10 + [fee percent="10" min_fee="5"]`.

#### Screenshots

Screenshot 2022-07-14 at 15 22 08

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : [Validate multicurrency shipping tax conversion](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-4.5.0#validate-multicurrency-shipping-tax-conversion)


dmallory42 (approved):
Looks good to me! Tested and it worked well in all of the scenarios. Also like how elegant the solution is!

Let's :shipit: !

PR #4436, opened 2022-07-13
Ask for login when buying a subscription with WooPay

### LINKED ISSUE #4435: WooPay redirects to merchant home page when buying a subscription while logged out
### Describe the bug

Currently when buying a subscription with WooPay with an existing user while logged out, you get redirected to the home screen and the cart doesn't empty because the merchant site requires to be logged in to complete a subscription.

### To Reproduce

1. Enable subscriptions for WooPay on merchant site.
2. Buy a subscription on merchant site.
3. Access My Account > Logout.
5. Buy another subscription.
6. Checkout with WooPay.

### Actual behavior

You'll be redirected to merchant home page and the cart will not be empty.

### Expected behavior

When you checkout with an existing e-mail address on merchant site, a notice appear asking the user to login, it would be great if this notice appear before checkout if there's a subscription product in cart and the user types an existing e-mail address on WooPay field when logged out, then they can log in on the merchant site and complete the transaction with WooPay.

alefesouza:
Fixes #4435

* Show a login notice before checkout when a existing logged out user buys a subscription and WooPay is enabled.

#### Testing instructions

* Enable subscriptions for WooPay on merchant site.
* Buy a subscription on merchant site.
* Access My Account > Logout.
* Buy another subscription.
* Type a valid WooPay e-mail.
* Instead of showing the SMS OTP, it should show the same notice when you checkout a subscription on merchant site while having a merchant site account, asking the user to login (notice that the "Please login" link on that notice is a `#` and don't redirect you to the `My account` page even when you checkout on merchant site with a existing account).
* Do the same flow while logged in or without subscriptions on the cart, it should show the SMS OTP normally.

-------------------

**Post merge**


a-danae (changes requested):
Hey, thanks for your work here @alefesouza!

The notice shows up as expected when a subscription is in the cart and a user that exists in the merchant is logged out.

It's also showing up when it's not expected, though, when the user is logged out and there's no subscription in the cart. Video [here](https://d.pr/v/tkFyPO).

Everything else seems to work as expected. 

~Why does the "Log in" link have `#` instead of an actual URL? I see this is expected as per the PR description, but wondering why not have an actual link there since we're already working around here.~ Scratch that, we're doing the same with WCPay subscriptions. It's worth its own issue if we want to update it.

As an improvement, we might want to remove the notice when the email address changes. I tried using two emails that exist on WooPay: first, one that has an account on the merchant site, and next, one that doesn't. The notice persisted when changing the email address. I think a better UX would be removing the notice when the email address changes to avoid this probably confusing scenario. Video [here](https://d.pr/v/Wpo5r4).

includes/admin/class-wc-rest-user-exists-controller.php
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Retrieve if a user exists by email address.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_REST_Response
+	 */
+	public function user_exists( WP_REST_Request $request ): WP_REST_Response {
+		$email   = $request->get_param( 'email' );
+		$user    = get_user_by( 'email', $email );
```
a-danae, line 61: This works fine, just an idea/alternative and maybe nit: 
We have this [email_exists()](https://developer.wordpress.org/reference/functions/email_exists/) function which might be lighter to determine whether the user exists than retrieving the whole user object.
This function is also the one used by [wc_create_new_customer()](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-user-functions.html#source-view.46) to trigger the error we want to display, so it may be more consistent too.

client/globals.d.ts
```diff
@@ -30,3 +30,6 @@ declare const wcpaySettings: {
 };
 
 declare const wcTracks: any;
+
+declare const wcPayPlatformCheckoutSubscriptions: any;
+declare const wcPayUserExistsUrl: string;
```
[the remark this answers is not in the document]
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/4436/commits/50040245210894ffe81691890bc88da6afd5ea11.

includes/admin/class-wc-rest-user-exists-controller.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/4436/commits/31a38c144fa7cbd2d447eeb4910c04fe9a4777ee.

includes/platform-checkout/class-platform-checkout-utilities.php
```diff
@@ -25,8 +25,21 @@ class Platform_Checkout_Utilities {
...
 		$is_platform_checkout_enabled  = 'yes' === $gateway->get_option( 'platform_checkout', 'no' );
-		$disable_for_subscription      = ! WC_Payments_Features::is_platform_checkout_subscriptions_enabled() && $this->is_subscription_item_in_cart();
+		$has_subscription_in_cart      = $this->is_subscription_item_in_cart();
+		$disable_for_subscription      = ! WC_Payments_Features::is_platform_checkout_subscriptions_enabled() && $has_subscription_in_cart;
 
-		return $is_platform_checkout_eligible && $is_platform_checkout_enabled && ! $disable_for_subscription;
+		$enable_platform_checkout = $is_platform_checkout_eligible && $is_platform_checkout_enabled && ! $disable_for_subscription;
+
+		if ( $enable_platform_checkout && $has_subscription_in_cart && ! is_user_logged_in() ) {
+			wp_localize_script(
```
a-danae, line 34: I wouldn't expect a method for calculating whether the platform checkout should be enabled to call `wp_localize_script()`. I also don't think we want to tie this side-effect to checking whether the platform checkout should be enabled. 

So, I'd say to move this localization to a context where it fits better. Maybe in [class-wc-payment-gateway-wcpay.php](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php) where we're handling scripts? Just throwing a quick idea for the location.

a-danae (changes requested):
Thanks for the adjustments here and for exploring alternatives Alefe! I just left a request for a small change. Things are working as expected.

includes/platform-checkout/class-platform-checkout-utilities.php
alefesouza: Makes sense, fixed on https://github.com/Automattic/woocommerce-payments/pull/4436/commits/bdcd132af496beb9d115b8594e658b9fc5d3b822.

a-danae (approved):
LGTM 🙂 thanks Alefe!

PR #4441, opened 2022-07-14
Add Support for filtering by customer currency in analytics

### LINKED ISSUE #2535: Multi-Currency Analytics: Add support for filtering the currency
For phase 2 of Multi-Currency, we'll need to add support for filtering options for currency in analytics as described here: https://wcpay.wordpress.com/2021/04/16/multi-currency-analytics/

Based on the work in #2457, I think this will hopefully be quite straightforward. We will need to add a filter to the front-end as described here: https://developer.woocommerce.com/2020/02/20/extending-wc-admin-reports/, and then in the backend pick up the `?GET` parameters and apply a `WHERE` clause to the query using the available filters.

The other thing we'll need to do is adjust the existing logic which converts the order amounts in place as part of the SQL query. If we only want orders matching a certain currency, then we don't need to do these conversions because we can just use the WHERE clause to filter out orders which do not match the selected currency. 

The only case where we would still want to do the conversions is if the user wants to filter by multiple currencies and display it in their store currency.

dmallory42:
Fixes #2535

- Adds a `Customer currency` filter to the Analytics page.

#### Testing instructions

* Make a few orders on your store in various different currencies, and wait a few minutes to ensure the action scheduler jobs which add the order info to the stats tables get processed
* Go to `Analytics > Orders` and verify you can see the `Customer currency` dropdown in the filters menu.
* Ensure you can filter by Customer Currency, and that selecting the `All options` option removes the filters. Make sure the total figures make sense (each value should reflect only the orders made in the selected currency).
* Quickly check the other analytics pages, they will **not** have this filter available, but just ensure the pages load and the data looks sensible.

Questions:

* This implements functionality that is (IMO) both usable and useful, but it doesn't implement *all* of the proposed changes in the original [project thread](https://wcpay.wordpress.com/2021/04/16/multi-currency-analytics/). I have split some of the additional functionality into follow-up issues. With that in mind, is it still okay to deploy this?

-------------------

**Post merge**


includes/multi-currency/Analytics.php
```diff
@@ -98,6 +112,34 @@ public function register_admin_scripts() {
...
+	 */
+	public function register_customer_currencies() {
+		$currencies           = $this->multi_currency->get_all_customer_currencies();
+		$available_currencies = $this->multi_currency->get_available_currencies();
+		$currency_options     = [];
+
+		foreach ( $currencies as $currency ) {
+			if ( ! isset( $available_currencies[ $currency ] ) ) {
+				continue;
+			}
```
ismaeldcom, line 128: What do you think about creating another method like `get_available_customer_currencies` to move this there and avoid duplicated code in the future if we need to access them? I don't want to overload `Multi_Currency` class with so many methods, but this one could be handy to avoid looping over currencies to filter out those that are not available anymore.

includes/multi-currency/Analytics.php
```diff
@@ -98,6 +112,34 @@ public function register_admin_scripts() {
...
+
+			$currency_details   = $available_currencies[ $currency ];
+			$currency_options[] = [
+				'label' => $currency_details->get_name(),
```
ismaeldcom, line 132: Looks like we will need to decode HTML entities here, I'm seeing `Icelandic kr&oacute;na` for ISK.

dmallory42: Great catch! Updated in fa1cbf57.

I also lowered the cache time, because it felt like too much to have 1 hour, especially if new currencies are being used on the store for the first time. I think this is still a good amount, as it prevents the query being made on every page load.

includes/multi-currency/Analytics.php
ismaeldcom: Make sense, better to create that method if it's needed and avoid a second loop for now 🙂 And yes, the possibility is quite remote, I was thinking of a currency which doesn't exist anymore, like a recent issue we had with a currency changing the code.

ismaeldcom:
I have a small nitpick 🙂 What do you think about sorting the currency options? Looks like they're not sorted by country nor currency code. I don't think there will be too many merchants with a lot of currencies enabled, but is an easy improvement anyway.

dmallory42:
Thanks @ismaeldcom, great suggestion! I've updated in 96e0dc56 to make that change!

PR #4442, opened 2022-07-14
Avoid saving a session cookie when the currency is changed because of geolocation

### LINKED ISSUE #3761: Examine Setting Cookies with Multi-Currency
### Describe the bug
When multi-currency is enabled, visiting the site while logged out will set a cookie.  This can affect performance since platforms like Pressable and Atomic will skip the page cache if a cookie that starts with `wp` or `wordpress` is set.

### To Reproduce
1. Go to Payments > Settings and enable multi-currency.
2. Next go to WooCommerce > Settings > Multi-currency and make sure the "Automatically switch customers to their local currency if it has been enabled" option is enabled.
3. Open up the site in a private browser window and observe that a cookie is set.

### Actual behavior
With these settings, a `wp_woocommerce_session` cookie will be set when logged out customers first visit the site.  

### Screenshots
automatically-switch-currency

cookie

### Expected behavior
It would be great if cookies weren't set until the customer selects a different currency from the store or is needed in some other way.  

* OS: MacOS
* Browser: Chrome
* Version: 97.0.4692.99

### Additional context
4726637-zen
https://a8c.slack.com/archives/C4JPKKQBF/p1644245890344629

daquinons:
Fixes #3761

As you can see in #3761, some caching systems get invalidated when the option "Automatically switch customers to their local currency if it has been enabled" is enabled because we store that change in the WC session cookie just on the first load without any user interaction.

With this change, `update_selected_currency` function gets a new parameter, `persist_change`, that if `true`, will create a cookie in case it doesn't exist. If the cookie exists, it's updated as normal. If `false`, the cookie is not stored in the user browser if it doesn't exist yet.

This parameter allows us to pass it as `false` in the `update_selected_currency_by_geolocation` function where it's called, making sure that a first-time visitor that has its currency changed via geolocation, doesn't get a cookie in its system until a different action is made (adds a product to the cart, changes manually its currency, or logs in).

#### Testing instructions

* Set the default currency as a different one to the one in your current location
* Go to Payments - Settings - Multi-Currency  and enable `Automatically switch customers to their local currency if it has been enabled`
* Clean the cookies in your browser/open an incognito window
* Visit the store and notice a blue banner in the bottom announcing the change of currency to your local one
* Open the storage info in your browser and see that no cookie has been set
image
- Make a purchase and make sure that the currency is the one that got changed because of your geolocation.

- Clear cookies again, and enter in the store

- See the notice announcing the change of currency again

- Choose manually a currency different to the one of your location

- See the cookie is created normally to store that change

image

-------------------

**Post merge**


tests/unit/multi-currency/test-class-multi-currency.php
```diff
@@ -380,13 +380,13 @@ public function test_update_selected_currency_by_url_updates_session_when_curren
 		$this->assertSame( 'GBP', WC()->session->get( WCPay\MultiCurrency\MultiCurrency::CURRENCY_SESSION_KEY ) );
 	}
 
-	public function test_update_selected_currency_by_geolocation_does_not_set_session_when_currency_not_enabled() {
+	public function test_update_selected_currency_by_geolocation_does_not_set_session() {
```
ismaeldcom, line 383: I think we should keep the old name, but also add `$this->mock_localization_service->method( 'get_country_locale_data' )->with( 'CL' )->willReturn( [ 'currency_code' => 'CLP' ] );` to make it similar to other tests. Although we're already covering it while testing `update_selected_currency` directly.

ismaeldcom (approved):
Thanks for working on it, I think it's a good way to avoid invalidating cache due to cookie, without overcomplicating our implementation 🙂 

It works as expected, the only thing I miss is having a test that covers the cookie, but it will be a bit tricky to test as we use `headers_sent()`. I also left a comment on the updated test.

Apart from that I'm approving it now! Let's 🚢  it!

tests/unit/multi-currency/test-class-multi-currency.php
daquinons: @ismaeldcom The reason of modifying this test was because now we are not setting the session for any currency, no matter if it exists or not, so I thought checking also a nonexisting currency was a bit redundant. I changed it to CA because CAD is one of the currencies added into the test. What do you think? 🤔

ismaeldcom: I think the test is working because we're not mocking `get_country_locale_data`, if you add that line it will fail. Because we're setting the session, even after this change, [here](https://github.com/Automattic/woocommerce-payments/blob/54c638f9f3536d5520b4af6bcc4633712c468cc6/includes/multi-currency/MultiCurrency.php#L640). We're only skipping setting the cookie 🙂 

That's why I think testing the cookie existence will be better, but as I mentioned, it will be tricky to do because we rely on `headers_sent()`.

PR #4450, opened 2022-07-18
Add missing intent info when when stripe.intent webhook request is made

### LINKED ISSUE #4313: UPE orders do not have payment_method_id meta when webhook arrives before redirect
### Describe the bug

UPE orders are left without the `_payment_method_id` meta when the payment intent succeeded webhook is processed before the redirect.

Since UPE payments only have the `payment_method` information after Stripe's JS confirms the payment, we rely on `process_redirect_payment` to attach the intent info to orders. However, due to [this line](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L585), `process_redirect_payment` could be skipped if the order is already marked as paid.

In case the payment intent succeeded webhook comes in before `process_redirect_payment` runs, the order [will be marked as paid](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-webhook-processing-service.php#L372), skipping `process_redirect_payment`. This is a problem because the webhook handler does not attach the intent info to the order, leaving some orders with missing data.

### To Reproduce

TBD.

### Actual behavior

In case a payment intent succeeded webhook is processed before the UPE redirect, only partial intent info is attached to the order in the `process_payment` [method](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L408).

### Expected behavior

The full intent info should be attached to the order, even if the payment intent succeeded webhook is processed before the UPE redirect.

### Additional context

p1654102516546879/1653331519.159749-slack-CGGCLBN58

zmaglica:
Fixes #4313 

In most cases, the intent info is updated during the UPE redirect, but sometimes Stripe webhook that delivers information about Stripe Intent can process faster than UPE redirect, which results in missing order metadata like `payment_method_id`. Since this webhook updates the status of the order, some metadata is skipped during the redirect. This PR solves that issue by updating the metadata that can be obtained from the webhook request. Every metadata value is checked and if the value doesn't exist, and it is not empty, the metadata will be stored. There is a function that can be used to update the metadata (`attach_intent_info_to_order`), but this function requires `customer_id` which doesn't exist inside webhook request, and some metadata values can be empty which will results in metadata with empty values.

#### Testing instructions

The easiest way on how to create an order and process it (finish payment) in a normal way. Go to the order and delete order metadata with keys like `_payment_method_id`, `_intent_id` or `_charge_id`. Go to the Stripe dashboard and rerun the `payment_intent.succeeded` webhook, and the metadata that you deleted should be there.

-------------------

**Post merge**


jessy-p (approved):
Tested and working fine. 
Need to add changelog
Personally, I prefer a meaningful variable like `order_changed` rather than `is_dirty`

zmaglica:
> Tested and working fine. Need to add changelog Personally, I prefer a meaningful variable like `order_changed` rather than `is_dirty`

Good call on `is_dirty`. I will update the variable name to one you proposed.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -366,18 +366,34 @@ private function process_webhook_payment_intent_succeeded( $event_body ) {
...
+		$meta_data_to_update = [
+			'_intent_id'         => $intent_id,
+			'_charge_id'         => $charge_id,
+			'_payment_method_id' => $payment_method_id,
+			WC_Payments_Utils::ORDER_INTENT_CURRENCY_META_KEY => $currency,
+		];
+
+		$order_changed = false;
+		foreach ( $meta_data_to_update as $key => $value ) {
+			if ( $value && ! $order->get_meta( $key ) ) {
```
luizreis, line 390: Hey @zmaglica - Sorry for chiming in a bit late, but what's the reasoning behind skipping the update if the order has something in the meta?

I ask because of p1654285164749429/1654031318.105099-slack-CGGCLBN58, in which orders were left with a stale `_charge_id` instead of the successful one, because of the `! $order->get_meta( '_charge_id' )` check. Is checking if they match, and updating the meta if they don't, a feasible approach here?

zmaglica: I wasn't sure where what is the correct source of data, so I skipped updating if the value existed. What I did here is the same this as it was before, but I added checks for all meta keys that can be found in intent.

For example:

Before this, the code that was updating the charge_id was:

```
if ( $charge_id && ! $order->get_meta( '_charge_id' ) ) {
			$order->update_meta_data( '_charge_id', $charge_id );
}
```
As you can see, the code is the same, since `charge_id` is the part of meta keys that needs to be updated. I can easily remove `$order->get_meta( $key )` if you think that value that comes from webhook is always correct.

luizreis: > As you can see, the code is the same, since charge_id is the part of meta keys that needs to be updated.

Yep, I had noticed that. The previous `! $order->get_meta( '_charge_id' )` check is what caused the old charge ID to not be updated in the case I investigated in that Slack thread.

> I can easily remove $order->get_meta( $key ) if you think that value that comes from webhook is always correct.

I've taken a look at the code, and when processing the order we always seem to override the meta, like [here](https://github.com/Automattic/woocommerce-payments/blob/114f6126ace6968837a0045315a6561ed3fc67ee/includes/payment-methods/class-upe-payment-gateway.php#L646) when called from `maybe_process_upe_redirect` for example.

I think it's likely that Stripe will have the most up-to-date data, especially considering that most UPE processing happens on their end. However, I'm not sure if we make some assumption somewhere in the plugin that could break if we start updating the data when Stripe does, so perhaps this needs more investigating or even some architecture's guild input. I'll create a new issue to track this and ask for some more feedback before we remove the check.

zmaglica: We can update charge_id for every webhook request if needed, and leave the reaming ones to be updated if missing. Will that be a viable option for this use case?

PR #4465, opened 2022-07-21
Add Multi-Currency compatibility with Points & Rewards plugin

### LINKED ISSUE #2848: Compatibility Multi-Currency with Points & Rewards
### Description
At the moment, the multi-currency feature is not compatible with Points & Rewards. 

* Install [WooCommerce Points and Rewards](https://woocommerce.com/products/woocommerce-points-and-rewards/?aff=10486&cid=1131038)
* With the default settings, 1 of store currency = 1 point
* Add a product to the cart (e.g. worth 100 EUR) - checkout will display 100 points
* Change currency to ZAR (or another one that has a large difference) - checkout will display that 1700 points will be awarded. 
* If I then change my currency back for the next purchase I can use a ton of points to make purchases

(Given that this is also a compatibility issue with [WooCommerce Multi-Currency](https://woocommerce.com/products/multi-currency/?aff=10486&cid=1131038) and the Aelia Multi-currency, this may also best be fixed on the Points & Rewards side.)

ismaeldcom:
Fixes #2848 

- Hook into earn and redeem points ratios and adjust them by currency rate to ensure it matches the store currency.
- Skip conversion on `WC_Points_Rewards_Discount->get_discount_data` to avoid making it twice, as we're already doing it for coupons in MC.
- As we might have rounding and/or charming enabled for a currency. The reward calculation could be higher after conversion and this is expected. The customer will receive points based on store currency after formatting rules are applied. E.g. with `1:1` earn ratio and EUR rate of `1` plus rounding of `10`.
  - `$15 USD` will be `15 points`
  - `€20 EUR` will be `20 points`

I've initially thought of applying filters to prices, but Points & Rewards doesn't have enough filters to do it right. And using points ratio looks like a cleaner solution which also simplifies double conversion issues.

#### Testing instructions
You'll need [WooCommerce Points and Rewards](https://woocommerce.com/products/woocommerce-points-and-rewards/) plugin.

**WooCommerce**
- **Settings**
  - **General**
    - **Currency**: `United States (US) dollar $`
    - **Number of decimals**: `2`
  - **Multi-currency**
    - **Euro** (€ EUR) 
      - **Exchange rate**: `Manual`
      - **Manual rate**: `0.91`
      - **Price rounding**: `1.00 (recommended)`
      - **Charm pricing**: `none`
    - **Icelandic króna** (kr. ISK)
      - **Exchange rate**: `Manual`
      - **Manual rate**: `133.16
      - **Price rounding**: `1.00 (recommended)`
      - **Charm pricing**: `none`
  - **Points & Rewards**
    - **Earn Points Conversion Rate**: `1`  **Points  =  $**: `1`
    - **Earn Points Rounding Mode**: `Round to nearest integer`
    - **Redemption Conversion Rate**: `10`  **Points  =  $**: `1`

**Store**
- Add a `$99 USD` product to cart and go there.
- The `Complete your order and earn X Points...` message should match:
  - **USD**: `99`
  - **EUR**:  `100`
  - **ISK**: `99`
- Place an order to obtain 99 points.
- Add any product to cart and go there.
- The `Use X Points for a Y discount...` message should match:
  - **USD**: `99 Points for a $9.90`
  - **EUR**: `100 Points for a 9,01 €` Yes, it shows more points, but It's something from the plugin rounding behavior. Not related with the compatibility.
  - **ISK**: `99 Points for a 1.318 kr.`
- Set up`Points earned` in a product
  - For a value of `20` it should reward with `20 points` no matter which currency is selected.
  - For a value of `100%` or `50%` the reward should match the converted numbers mentioned previously.
- Being in the cart press `Apply Discount`
- `Points redemption` in cart totals should match the previously shown discount.
- Remove it, and apply again but in the checkout page.
- Should be applied correctly too.

**REST API**
- Install `WooCommerce Blocks` plugin if you don't have it already.
- Get a key from **WooCommerce > Settings > Advanced > REST API**
- Make a GET request to `https://wcpay.test/wp-json/wc/store/v1/cart?consumer_key=ck_123&consumer_secret=cs_123` using your key
- Look for `extensions['points-and-rewards']`
- The values should be converted like in the store.

#### Screenshots

| USD | EUR | ISK |
| - | - | - |
|Screenshot 2022-07-21 at 12 52 33|Screenshot 2022-07-21 at 12 52 53|Screenshot 2022-07-21 at 12 53 04|

-------------------

**Post merge**

### DIFF
--- a/includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
+++ b/includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Class WooCommercePreOrders
+ * Class WooCommercePointsAndRewards
  *
  * @package WCPay\MultiCurrency\Compatibility
  */
@@ -65,7 +65,7 @@ class WooCommercePointsAndRewards extends BaseCompatibility {
 		$value  = (float) ( $ratio[1] ?? 0 );
 
 		$rate  = $this->selected_currency->get_rate();
-		$value = (float) $value * $rate;
+		$value = $value * $rate;
 
 		return "$points:$value";
 	}
--- a/includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
+++ b/includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
@@ -7,10 +7,26 @@
 
 namespace WCPay\MultiCurrency\Compatibility;
 
+use WCPay\MultiCurrency\Currency;
+
 /**
  * Class that controls Multi Currency Compatibility with WooCommerce Points & Rewards Plugin.
  */
 class WooCommercePointsAndRewards extends BaseCompatibility {
+
+	/**
+	 * Default Currency Code.
+	 *
+	 * @var string
+	 */
+	private $default_currency_code;
+
+	/**
+	 * Selected Currency Code.
+	 *
+	 * @var Currency
+	 */
+	private $selected_currency;
 
 	/**
 	 * Init the class.
@@ -20,14 +36,6 @@ class WooCommercePointsAndRewards extends BaseCompatibility {
 	protected function init() {
 		// Add needed filters if Points & Rewards is active and it's not an admin request.
 		if ( is_admin() || ! class_exists( 'WC_Points_Rewards' ) ) {
-			return;
-		}
-
-		$default_currency  = $this->multi_currency->get_default_currency();
-		$selected_currency = $this->multi_currency->get_selected_currency();
-
-		// Skip it if selected currency and default store currency match.
-		if ( $default_currency->code === $selected_currency->code ) {
 			return;
 		}
 
@@ -42,6 +50,11 @@ class WooCommercePointsAndRewards extends BaseCompatibility {
 	 * @return string Converted points ratio.
 	 */
 	public function convert_points_ratio( string $ratio = '' ): string {
+		// Skip conversion if selected and default currencies are the same.
+		if ( $this->selected_and_default_currency_match() ) {
+			return $ratio;
+		}
+
 		// Skip conversion on discount to avoid doing it twice.
 		if ( $this->utils->is_call_in_backtrace( [ 'WC_Points_Rewards_Discount->get_discount_data' ] ) ) {
 			return $ratio;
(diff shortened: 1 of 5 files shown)


ismaeldcom:
Updated code to fix filters not being applied after selecting a currency coming from the default one for the first time. Moving the skip logic to the rate conversion prevents the race condition.

I've also added a private method to call `get_selected_currency()` only once reducing the performance hit.

Thanks for catching the issue @oaratovskyi!

includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/**
+ * Class WooCommercePreOrders
```
oaratovskyi, line 3:
```suggestion
 * Class WooCommercePointsAndRewards
```

includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
```diff
@@ -0,0 +1,88 @@
...
+		if ( $this->utils->is_call_in_backtrace( [ 'WC_Points_Rewards_Discount->get_discount_data' ] ) ) {
+			return $ratio;
+		}
+
+		$ratio  = explode( ':', $ratio );
+		$points = (float) ( $ratio[0] ?? 0 );
+		$value  = (float) ( $ratio[1] ?? 0 );
+
+		$rate  = $this->selected_currency->get_rate();
+		$value = (float) $value * $rate;
```
oaratovskyi, line 68: I think we can omit casting here, since $rate is `float`, because `$this->selected_currency->get_rate()` is type-hinted. And $value is type casted a bit above.

oaratovskyi (approved):
Thanks for working on it and fixing the race condition issue!
I've tested it one more time and now it works correct ✅ . The code looks good to me, I've left a few nitpicks, but nothing blocking, approving it! 
:shipit:

includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
ismaeldcom: Thanks for catching it! Addressed in 34e2338

includes/multi-currency/Compatibility/WooCommercePointsAndRewards.php
ismaeldcom: You're right, I had it before doing the cast in the previous definition. Addressed in 34e2338

Issue #4395, opened 2022-07-04
Multiple UPE - Checkout can't be completed from the Pay for Order page


harriswong:
### Description

This is one of the TODOs from https://github.com/Automattic/woocommerce-payments/pull/3968#issuecomment-1156928898. 

When we use the `poc/upe-instances-multiplied` branch, checkout can't be completed from the "Pay for Order page". This error is appearing in the console:
```
Uncaught (in promise) TypeError: upeComponents is undefined
    _callee$ upe.js:230
    tryCatch runtime.js:63
    invoke runtime.js:293
    defineIteratorMethods runtime.js:118
    Babel 2
    node_modules upe_checkout.js:10985
    node_modules upe_checkout.js:10974
    mountUPEElement upe.js:224
    <anonymous> upe.js:424
    jQuery 13
upe.js:230:6
    Babel 4
    mountUPEElement upe.js:224
    <anonymous> upe.js:424
    jQuery 13
```

This seems to be related to the change in the way the payment method type is handled.

This ticket is to find out what causes the error and to resolve it.

### Acceptance Criteria
- fix the error
- changes made should be merged into `poc/upe-instances-multiplied`, not `develop`.

harriswong:
If I go to http://localhost:8082/my-account/orders/, then the error is:
```
Uncaught (in promise) IntegrationError: In order to create a payment element, you must pass a valid PaymentIntent or SetupIntent client secret when creating the Elements group.

  e.g. stripe.elements({clientSecret: "{{CLIENT_SECRET}}"})
    at Yn (?ver=3.0:1:222954)
    at rr (?ver=3.0:1:230188)
    at new n (?ver=3.0:1:234345)
    at e.<anonymous> (?ver=3.0:1:237627)
    at e.create (?ver=3.0:1:132577)
    at _callee$ (upe.js:344:25)
    at tryCatch (runtime.js:63:16)
    at Generator.invoke [as _invoke] (runtime.js:293:1)
    at Generator.throw (runtime.js:118:1)
    at asyncGeneratorStep (asyncToGenerator.js:3:1)
```

harriswong:
Merged into `poc/upe-instances-multiplied`. Closing

csmcneill:
Reopening this issue as I can reproduce this issue.

### Steps to Reproduce

1. Enable UPE.
2. Disable multi-currency (as this can allow for the order to be paid with a different currency depending on geolocation settings).
4. Set the site's default currency to EUR.
[...]

PR #4469, opened 2022-07-21
Remove mountUPEElement() because we are creating one for each payment…

harriswong:
Fixes https://github.com/Automattic/woocommerce-payments/issues/4395

### Description 

We are mounting UPE element for each payment method type (giropay, sepa, etc). The `mountUPEElement( useSetUpIntent );` code returns an error because `useSetUpIntent` is passed in as `paymentMethodType` in the new parameter signature for `mountUPEElement(paymentMethodType, upeDOMElement, isSetupIntent = false)`. 

I don't think we need this line anymore because we are already mounting a UPE element for each of the payment types.

#### Testing instructions
1.  Checkout an item and then update the order status to "Pending payment". This allows us to click on the "Customer payment page" link. 
image
2. Click the "Customer payment page" link to open the "Pay for order" page. 
3. You should see the multiple checkout elements without the `Uncaught (in promise) TypeError: upeComponents is undefined` error in console. 
image

-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe.js
+++ b/client/checkout/classic/upe.js
@@ -422,7 +422,6 @@ jQuery( function ( $ ) {
 					useSetUpIntent
 				);
 			}
-			mountUPEElement( useSetUpIntent );
 		}
 	}
 


mdmoore (approved):
@harriswong I discovered this last night as well. It looks like it was leftover during the refactor of the function. It makes sense to remove the extra call here and solves the error. 🚢

FangedParakeet (approved):
Whoops, that shouldn't be there anymore! It looks like all the mounting should be handled in the DOM elements loop above and the function signature doesn't even match here anymore either. Good riddance with this line of code. LGTM. :shipit:

PR #4494, opened 2022-07-27
Update currencies modal height

### LINKED ISSUE #4491: MultiCurrency selection modal gets cut out in smaller resolutions
### Describe the bug
When you go to Payments - Settings - Multi-currency, and you click on the "Add currencies" button, a modal appears with a list of currencies.

When you open it on a 720p screen (not mobile), the modal gets cut out not showing the "Update" button, which needs to be scrolled to be found.

image

That you need to scroll on the modal is not obvious, as the modal has another scrollable area which is the currency list itself.

And this is how you see it on a bigger screen

image

### To Reproduce

1. Go to Payments - Settings - Multi-currency
2. Click on "Add currencies" button
3. Open your browser dev tools and modify the resolution to 1400x720
4. See "Update selected" button hidden in the bottom area of the modal.

### Actual behavior
The modal is built in a way that the header is fixed, with the footer being hidden under the list if the size of the modal is not big enough.

### Expected behavior
I expect to see at a first sight the footer with the action button at any resolution.

* OS: MacOS
* Browser Firefox & Chrome

hsingyuc:
Fixes #4491 

This PR changes the currencies modal height so the footer, as well as the search input box, can be seen at a first sight at any resolution.

#### Videos
*Before*
https://user-images.githubusercontent.com/56378160/181353092-c74fbb0e-158a-4fb2-b9bf-f430c069c3ba.mov

*After*
https://user-images.githubusercontent.com/56378160/181350340-79b9a4b2-9537-4524-82e6-652458c7910c.mov

#### Testing instructions
1. Go to Payments - Settings - Multi-currency
1. Click on the "Add currencies" button
1. Open your browser dev tools and modify the resolution to 1400x720 and any other sizes
1. See the "Update selected" button and the search input box render in the currencies modal.

*

-------------------

- [X] Tested on mobile (or does not apply)

**Post merge**

### DIFF
--- a/changelog/fix-4491-currencies-modal-height
+++ b/changelog/fix-4491-currencies-modal-height
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Update currencies modal height
--- a/client/components/confirmation-modal/styles.scss
+++ b/client/components/confirmation-modal/styles.scss
@@ -6,6 +6,8 @@
 
 	// to ensure that the separator extends all the way, even with different versions of Gutenberg
 	.components-modal__content {
+		display: flex;
+		flex-direction: column;
 		padding: 0 $grid-unit-30 $grid-unit-30;
 	}
 
--- a/client/multi-currency/multi-currency-settings/enabled-currencies-list/style.scss
+++ b/client/multi-currency/multi-currency-settings/enabled-currencies-list/style.scss
@@ -172,8 +172,6 @@
 
 	&__content {
 		overflow-y: scroll;
-		min-height: 350px;
-		max-height: 350px;
 		flex: 1;
 		display: flex;
 		flex-direction: column;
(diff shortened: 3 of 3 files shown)


brianyu28 (changes requested):
Thanks for working on this, @hsingyuc!

Decreasing the height of the modal definitely helps when the height of the window is 720 pixels tall, though it does mean that on larger screens you can't see as many currencies anymore. I also noticed that if a user were to decrease their window size even further (smaller than 720 pixels), the buttons still get cut off:

Screen%20Capture%20on%202022-07-27%20at%2015-43-30

Instead of decreasing the height, could we modify the styles such that the modal as a while keeps the same height as it does now, but the buttons stay visible even if the window gets smaller than 720 pixels tall?

hsingyuc:
@brianyu28  Thank you for the feedback!  I changed it to flex-based on the components modal.  Take a look and let me know. Thank you!

brianyu28:
Looks great @hsingyuc, thanks! Confirmed that the buttons now always show even when resizing the window.

Just to preserve the modal resizing behavior, can we add back the `max-height: 350px` restriction, so that the modal doesn't get too tall? But other than that this looks good to me!

hsingyuc:
My mistake.  I pushed the change, thank you, @brianyu28!

brianyu28 (approved):
This looks good to me now. Thanks, @hsingyuc!

PR #4527, opened 2022-08-03
Add thousand separator to the transactions count

### LINKED ISSUE #3728: Number of results and page count isn't formatted correctly
### Describe the bug
On pages with tables, the number of results and the page count aren't formatted correctly.

Screenshot 2022-01-28 at 17 12 47

In this screenshot I would expect the number of transactions and the number of pages to have thousand separators in them.

### To Reproduce
Tested against the transactions page, but probably happens on other screens as well.

Tested on:
* WP 5.9
* WC 6.1.1
* WCPay 3.6.1

### To Reproduce
1. Open a page (for example, transactions) that has enough results to cause more than 10,000 pages to be displayed (or temporarily hard code the summary endpoint to return a high record count)
2. See the numbers lack thousand separators

Tested on, but probably present on other configurations:
* OS: iOS
* Browser: Chrome
* Version: 97

Potentially related to https://github.com/Automattic/woocommerce-payments/issues/3727

fashuvo:
Fixes #3728

When there are more than 999 transactions, the transactions count does not display the thousand separator. This PR installs the Number package and adds the appropriate thousand separator based on the site language.

### Before

This is how the transactions count looked like before adding the thousand separator (on an English language site):

before-thousand-operator

### After

This is how the transactions count looks like after adding the thousand separator: 

On an **English** language site:

after-thousand-operator

On a **German** language site:

Image 2022-08-10 at 11 48 10 AM

And on a **French** language site:

Image 2022-08-10 at 11 49 19 AM

### Testing instructions

1. Clone the `fix/page-result-count-formatting` branch in your local machine
2. Navigate to `woocommerce-payments/client/transactions/list/index.tsx`
3. Run `npm run watch` in the terminal
4. Find this line in the index.tsx file: ``value:  `${ applyThousandSeparator(transactionsSummary.count as number) }` ``
5. Change `transactionsSummary.count` to any number greater than 999 and save changes
6. Refresh wp-admin > Payments > Transactions and you should see the thousand separator in the transaction count.
7. Go to wp-admin > Settings > General and change site language to `Deutsch`.
8. Refresh wp-admin > Payments > Transactions and you should see `dot` as the thousand separator.
9. Go to wp-admin > Settings > General and change site language to `Français`.
10. Refresh wp-admin > Payments > Transactions and you should see `space` as the thousand separator.

-------------------

**Post merge**

### DIFF
--- a/client/transactions/list/index.tsx
+++ b/client/transactions/list/index.tsx
@@ -54,9 +54,20 @@ import DownloadButton from 'components/download-button';
 import { getTransactionsCSV } from '../../data/transactions/resolvers';
 import p24BankList from '../../payment-details/payment-method/p24/bank-list';
 
+const siteLang = document.documentElement.lang;
 const siteNumberOptions = {
 	thousandSeparator: ',',
 };
+
+if ( [ 'fr-CA', 'pl-PL', 'fr-FR', 'fr-BE' ].includes( siteLang ) ) {
+	siteNumberOptions.thousandSeparator = ' ';
+} else if (
+	[ 'de-AT', 'nl-BE', 'de-DE', 'it-IT', 'nl', 'es', 'pt-BR' ].includes(
+		siteLang
+	)
+) {
+	siteNumberOptions.thousandSeparator = '.';
+}
 
 const formatStoreNumber = partial( numberFormat, siteNumberOptions );
 interface TransactionsListProps {
--- a/client/transactions/list/index.tsx
+++ b/client/transactions/list/index.tsx
@@ -53,23 +53,6 @@ import wcpayTracks from 'tracks';
 import DownloadButton from 'components/download-button';
 import { getTransactionsCSV } from '../../data/transactions/resolvers';
 import p24BankList from '../../payment-details/payment-method/p24/bank-list';
-
-const siteLang = document.documentElement.lang;
-const siteNumberOptions = {
-	thousandSeparator: ',',
-};
-
-if ( [ 'fr-CA', 'pl-PL', 'fr-FR', 'fr-BE' ].includes( siteLang ) ) {
-	siteNumberOptions.thousandSeparator = ' ';
-} else if (
-	[ 'de-AT', 'nl-BE', 'de-DE', 'it-IT', 'nl', 'es', 'pt-BR' ].includes(
-		siteLang
-	)
-) {
-	siteNumberOptions.thousandSeparator = '.';
-}
-
-const formatStoreNumber = partial( numberFormat, siteNumberOptions );
 interface TransactionsListProps {
 	depositId?: string;
 }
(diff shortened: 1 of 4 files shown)


client/transactions/list/index.tsx
```diff
@@ -612,7 +619,7 @@ export const TransactionsList = (
 					transactionsSummary.count as number,
```
deepakpathania, line 619: Do we need to format this here as well? If yes, I'd recommend storing the formatted value as a variable above and then using that directly.

deepakpathania (commented):
Changes look good overall, left a single comment for clarification.

client/transactions/list/index.tsx
fashuvo: `transactionsSummary.count` is passed as an argument to the `_n` function. And the function [expects an integar](https://developer.wordpress.org/reference/functions/_n/#parameters) there. So I don't think we need to apply formatting there.

jessy-p (commented):
If I change the WordPress site language settings to German, the Orders screen respects the German formatting (dot) for thousand seperator
image

The transaction screen uses the comma seperator irrespective of language setting. Should we be using something like number_format_i18n to format the numbers?

fashuvo:
@jessy-p Can you please clarify which Orders screen you were referring to? I am not seeing any total/summary section on the wp-admin > WooCommerce > Orders page.

client/transactions/list/index.tsx
```diff
@@ -52,6 +54,22 @@ import DownloadButton from 'components/download-button';
 import { getTransactionsCSV } from '../../data/transactions/resolvers';
 import p24BankList from '../../payment-details/payment-method/p24/bank-list';
 
+const siteLang = document.documentElement.lang;
+const siteNumberOptions = {
+	thousandSeparator: ',',
+};
+
+if ( [ 'fr-CA', 'pl-PL', 'fr-FR', 'fr-BE' ].includes( siteLang ) ) {
```
deepakpathania, line 62: Few notes:
- How do we know this list is inclusive? A comment on the docs might be helpful for reference.
- We could use something like `startsWith` to avoid having cases like `fr-CA` and `fr-BE` separately. If it starts with `fr` we could consider it valid for this case, similarly for others.
- Encapsulating this logic to a separate function and then calling that would read better instead of having it in the global scope at the beginning of the file.

deepakpathania (changes requested):
Left a comment with a few changes, once addressed please update the PR description with a different language page as well for easier reference.

client/transactions/list/index.tsx
fashuvo: > How do we know this list is inclusive?

WCPay is now available in [18 countries](https://woocommerce.com/payments/#:~:text=Where%20in%20the%20world%3F,Kingdom%2C%20and%20the%20United%20States.). I got the locale for these languages from the WordPress.org [translate page](https://make.wordpress.org/polyglots/teams/). And then checked the thousand separator by changing the site language to each of these languages.

client/transactions/list/index.tsx
```diff
@@ -77,6 +78,28 @@ interface Column extends TableCardColumn {
...
 }
 
+const applyThousandSeparator = ( trxCount: number ) => {
+	const siteLang = document.documentElement.lang;
+	const siteNumberOptions = {
+		thousandSeparator: ',',
+	};
+
+	if ( [ 'fr', 'pl' ].some( ( lang ) => siteLang.startsWith( lang ) ) ) {
+		siteNumberOptions.thousandSeparator = ' ';
```
deepakpathania, line 88: I am not sure about using space and `'` as a separator tbh. 🤔 Could you start a thread and confirm this in the team channel once?

deepakpathania (commented):
Sorry for mentioning this late in the review but have a doubt about the separators, please confirm this once with someone from design team in the channel.

client/transactions/list/index.tsx
deepakpathania: Resolving based on Slack thread: p1660118893096929-slack-C01BPL3ALGP

deepakpathania (approved):
Thanks for addressing the comments, changes LGTM now.

PR #4533, opened 2022-08-04
Avoid execution context errors for E2E tests

jbordonado:
Fixes #

In the E2E tests reports, we can come across the following kind of errors: `Execution context was destroyed, most likely because of a navigation.`.
As the error message suggests, it probably means we're trying to use an element that doesn't exist anymore because the browser started to navigate elsewhere.

An interesting topic on Stackoverflow regarding this: https://stackoverflow.com/questions/55877263/puppeteer-execution-context-was-destroyed-most-likely-because-of-a-navigation

Essentially, we must avoid defining elements too early compared to when we use them. We should grab them the moment we want to use them in order to make sure they are not obsolete.

#### Testing instructions

- Trigger a new E2E tests workflow
- Make sure you don't see any failing scenarios that relates to the execution context error

Latest run (0 errors related to that): https://github.com/Automattic/woocommerce-payments/actions/runs/2795204044

-------------------

**Post merge**

### DIFF
--- a/changelog/dev-fix-e2e-tests-execution-context-errors
+++ b/changelog/dev-fix-e2e-tests-execution-context-errors
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Avoid execution context errors during E2E tests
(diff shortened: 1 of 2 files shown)


htdat (approved):
LGTM! But the CI run `Check changelog` is failing, you might want to run `npm run changelog`.

PR #2456, opened 2021-07-16
feat: remove non-UPE methods when UPE is disabled

frosso:
Fixes #

#### Changes proposed in this Pull Request

When UPE is disabled, the backend removes non-UPE payment methods from the gateway: https://github.com/Automattic/woocommerce-payments/blob/5aefa3eb392b990a5fde0d76e3c40aa323f8808b/includes/admin/class-wc-rest-upe-flag-toggle-controller.php#L126-L131
Which leaves the frontend with an inconsistent state, until the page is refreshed.

Duplicating the same logic to the backend on the frontend, to ensure non-UPE payment methods are removed when the flag is disabled.

#### Testing instructions

- Ensure you have "Enable UPE settings changes" enabled
- Ensure you have "Enable UPE checkout" enabled
- Go to http://localhost:8082/wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments
- Add a non-UPE payment method
- In the console, you should see it when querying `wp.data.select('wc/payments').getEnabledPaymentMethodIds()`
- Disable UPE
- The non-UPE payment methods is removed from the store
- You can check this by looking at the browser's console: `wp.data.select('wc/payments').getEnabledPaymentMethodIds()`

-------------------

- [ ] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

PR #4546, opened 2022-08-08
Update available payment methods to card only after disabling UPE

### LINKED ISSUE #4422: Enabling and disabling UPE - payment method list not updated correctly
### Describe the bug
If you have UPE enabled and then disable it, the payment method list still shows the additional payment methods.

### To Reproduce
- Enable UPE and add an additional payment method like Bancontact
- Now, disable UPE
- You can still see all the additional payment methods in the list instead of just the Credit card
- If you refresh the page, the payment method list is correct and only shows the Credit card

### Actual behavior
image

### Expected behavior
image

hsingyuc:
Fixes #4422 

Update available payment methods to card only after disabling UPE
> Were there other solutions you considered? 

I was planning on calling `invalidateResolutionForStoreSelector( 'getSettings' )` to keep a single source of truth, but we already manually set enabled payments to `[ 'card' ]` when disabling.

> Why did you choose to pursue this solution?

I found we have used the set data store method to set the enabled payment method to `card`.  To keep it consistent, I add `updateAvailabledPaymentMethodIds` to update the card to the available payment method

<hr>

#### Screenshots
*Before*
before

*After*
after

#### Testing instructions
1. Enable UPE, add an additional payment method like Bancontact, and save changes
2. Disable UPE
3. Expect to see only the credit card method and the UPE setup banner

*

-------------------

- [X] Covered with tests (or have a good reason not to test in description ☝️)
- [X] Tested on mobile (or does not apply)

**Post merge**


changelog/fix-4422-set-available-method-to-card-after-disabling-upe
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Update id card to available payment method after disabling UPE
```
brianyu28, line 4: What do you think about something like "Hide UPE payment methods after disabling UPE" to make the changelog entry a little clearer as to what's been fixed?

brianyu28 (approved):
Thanks @hsingyuc, this works well!

Since we already have the separate front-end update for enabled payment methods, doing the same to fix the available payment methods seems reasonable. I read some of the discussion in #2456 for more context there, they had similar discussions about the tradeoffs between these approaches.

Left a small suggestion on the wording of the changelog entry, just to make clear what's changed in the fix. After that, looks good to merge :shipit:

PR #4534, opened 2022-08-04
Try using E2E_RETRY_TIMES with woocommerce/e2e-environment v0.3.0

htdat:
Fixes #

#### Changes proposed in this Pull Request

- See https://github.com/automattic/woocommerce-payments/issues/4421#issuecomment-1203834586

#### Testing instructions

*

-------------------

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4549, opened 2022-08-09
E2E GitHub Workflow: Re-run Failed Test Files.

### LINKED ISSUE #4421: Flaky E2E test - wcpay/shopper/shopper-myaccount-save-card-and-checkout.spec.js
### Describe the bug
The test `wcpay/shopper/shopper-myaccount-save-card-and-checkout.spec.js` fails randomly with multiple errors.

* `Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Error:`
* `waitForFunction failed: frame got detached.`
* `TimeoutError: Element label (text: "Visa ending in 3220 (expires 04/36)") not found`
* `TypeError: expect(...).toClick is not a function`

Reference job - https://github.com/Automattic/woocommerce-payments/runs/7226350436?check_suite_focus=true

### Actual behavior
The test fails randomly.

### Expected behavior
The test should pass consistently.

htdat:
Fixes #4421 and maybe other similar issues. 

- Export the first E2E test result to a JSON file - [Jest doc ref](https://jestjs.io/docs/cli#--outputfilefilename).
- Check this result file to decide whether to run the second run with only failed test files. 

### Rationale for this approach: 

- When monitoring https://github.com/Automattic/woocommerce-payments/issues/4421 and other flaky E2E issues, I noticed that many of them failed due to loading the client site with error status 503 (Service Unavailable) or 429 (Too Many Requests). 
- With our test nature, many tests rely on previous tests in the same spec/test file. E.g: tests/e2e/specs/wcpay/shopper/shopper-checkout-save-card-and-purchase.spec.js, OR wcpay/shopper/shopper-myaccount-save-card-and-checkout.spec.js. 
- Also, many flaky results have only one failed test file. 
- With three reasons above, we think that re-running only failed test files will allow the client site having time to recover from failure and reduce the flakiness. 

Also note that, I also tried to use the built-in [jest.retryTimes()](https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options) but it does not serve our specific test nature well because Jest will retry failed tests in sequence right away before it goes to the next test. 

### Rationale for changing in the GitHub workflow: 

- First, I proposed this approach https://eng.wealthfront.com/2022/04/21/retrying-e2e-test-suites-with-jest/, i.e. writing a custom NodeJS script and running them. 
- However, I faced some challenges: 
     - The article uses `runCLI`, which is NOT documented in Jest docs: https://jestjs.io/docs/ though we can look at its definition in [the Jest code package](https://github.com/facebook/jest/blob/240587bde5dae1467ced0fdeee2e668e01caf896/packages/jest-core/src/cli/index.ts#L37).
     - This approach means running Jest Puppeteer directly without a wrapper provided by `wp-scripts` through `npm run test:e2e`. This results in providing many unexpected tweaks in the NodeJS script, which is not ideal.
     - Last bust not least, when running the NodeJS script, this error pops up `Cannot use import statement outside a module`. I tried a few away to fix it but no way is promsing. It seems an issue with the specific Node version and it's not really easy to deal with our current setup. 
- With my approach in this PR, it is more elegant, and we just need to touch GitHub Actions, which is what we want finally. 

Relevant discussion: p1659950516959449-slack-C01BPL3ALGP

### Testing instructions

I think these two E2E tests for this PR branch `dev/e2e-retry-failed-test-files` can show how effectively this approach is. 

#### First example: https://github.com/Automattic/woocommerce-payments/actions/runs/2829780984

- In this link, there are two tests failed in the first run: 

Markup on 2022-08-10 at 10:43:52

- However, they were successfully in the second run with only over 1 minute (rather than 6 minutes for the first run). Looking into them in detail, they all failed due to this error `Failed to load resource: the server responded with a status of 429`. 

Markup on 2022-08-10 at 10:47:19

#### Second example https://github.com/Automattic/woocommerce-payments/actions/runs/2826082997

- There were even more failed tests in the first run.
- But many of them were fixed in the second run.

-------------------

**Post merge**

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -64,7 +64,7 @@ jobs:
       E2E_WC_VERSION: ${{ matrix.woocommerce }}
       E2E_GROUP:  ${{ matrix.test_groups }}
       E2E_BRANCH: ${{ matrix.test_branches }}
-      E2E_RESULT_FILEPATH: './tests/e2e/results.json'
+      E2E_FIRST_RUN_RESULT_FILEPATH: 'tests/e2e/results.json'
 
     steps:
       # Conditionally skip WC Blocks tests. Remove/update based on min supported WC version by the blocks checkout plugin.
@@ -150,9 +150,9 @@ jobs:
         # Use +e to trap errors when running E2E tests.
         shell: /bin/bash +e {0}
         run: | 
-          npm run test:e2e -- --json --outputFile="$E2E_RESULT_FILEPATH"
+          npm run test:e2e -- --json --outputFile="$E2E_FIRST_RUN_RESULT_FILEPATH"
           
-          E2E_NUM_FAILED_TEST_SUITES=$(cat "$E2E_RESULT_FILEPATH" | jq '.numFailedTestSuites')
+          E2E_NUM_FAILED_TEST_SUITES=$(cat "$E2E_FIRST_RUN_RESULT_FILEPATH" | jq '.numFailedTestSuites')
 
           echo "::set-output name=FIRST_RUN_FAILED_TEST_SUITES::$(echo $E2E_NUM_FAILED_TEST_SUITES)"
 
@@ -165,7 +165,7 @@ jobs:
         if: ${{ steps.first_run_e2e_tests.outputs.FIRST_RUN_FAILED_TEST_SUITES > 0  }}
         # Filter failed E2E files from the result JSON file, and re-run them.
         run: |
-          cat "$E2E_RESULT_FILEPATH" | jq '.testResults[] | select(.status == "failed") | .name' | xargs npm run test:e2e
+          cat "$E2E_FIRST_RUN_RESULT_FILEPATH" | jq '.testResults[] | select(.status == "failed") | .name' | xargs npm run test:e2e
 
       # Archive screenshots if any
       - name: Archive e2e test screenshots & logs
@@ -176,5 +176,6 @@ jobs:
             path: |
               tests/e2e/screenshots
               tests/e2e/docker/wordpress/wp-content/debug.log
+              ${{ env.E2E_FIRST_RUN_RESULT_FILEPATH }}
             if-no-files-found: ignore
             retention-days: 5
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -147,17 +147,20 @@ jobs:
 
       # Run Tests
       - name: Run E2E Tests
+        id: first_run_e2e_tests
+        # Use +e to trap errors when running E2E tests.
         shell: /bin/bash +e {0}
         run: | 
           npm run test:e2e -- --json --outputFile="$E2E_RESULT_FILEPATH"
           E2E_NUM_FAILED_TEST_SUITES=$(cat "$E2E_RESULT_FILEPATH" | jq '.numFailedTestSuites')
+          echo "::set-output name=E2E_NUM_FAILED_TEST_SUITES::$(echo E2E_NUM_FAILED_TEST_SUITES)"
           if [[ ${E2E_NUM_FAILED_TEST_SUITES} -gt 0 ]]; then
-            echo 'Some tests failed but we will try them again in the next job.'
+            echo '::notice Some tests failed but we will try them again in the next job.'
             exit 0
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -144,9 +145,27 @@ jobs:
...
+        shell: /bin/bash +e {0}
+        run: | 
+          npm run test:e2e -- --json --outputFile="$E2E_RESULT_FILEPATH"
+          
+          E2E_NUM_FAILED_TEST_SUITES=$(cat "$E2E_RESULT_FILEPATH" | jq '.numFailedTestSuites')
+
+          echo "::set-output name=FIRST_RUN_FAILED_TEST_SUITES::$(echo $E2E_NUM_FAILED_TEST_SUITES)"
+
+          if [[ ${E2E_NUM_FAILED_TEST_SUITES} -gt 0 ]]; then
+            echo '::notice::Some E2E tests failed in the first run but we will try them again in the second run.'
```
deepakpathania, line 160: Might be nice to use the exact number of tests that failed in the message since we have that available here.

.github/workflows/e2e-test.yml
```diff
@@ -144,9 +145,27 @@ jobs:
...
+            exit 0
+          fi
+      
+      - name: Re-try Failed E2E Files
+        if: ${{ steps.first_run_e2e_tests.outputs.FIRST_RUN_FAILED_TEST_SUITES > 0  }}
```
deepakpathania, line 165: Do we need to add a step to archive the generated output file or is that handled by GH actions?

deepakpathania (changes requested):
Overall the approach looks good to me, I would be interested in how effective the [retryTimes](https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options) option is comparatively if set globally at the file level, as described [here](https://stackoverflow.com/a/60129634), left some basic comments for this approach.

htdat:
@deepakpathania - Thanks for your review! I also addressed your great inline feedback too.

> I would be interested in how effective the [retryTimes](https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options) option is comparatively if set globally at the file level

I gave it a try here https://github.com/Automattic/woocommerce-payments/pull/4534

You can have a look at this run https://github.com/Automattic/woocommerce-payments/runs/7688604355?check_suite_focus=true#step:15:550

This test was run two times: 

```
  ● Payment Methods › when using a 3ds card › should add the card as a new payment method

    TimeoutError: Element a (text: "Add payment method") not found

      waiting for function failed: timeout 500ms exceeded

  ● Payment Methods › when using a 3ds2 card › should add the card as a new payment method

    TimeoutError: Element a (text: "Add payment method") not found

      waiting for function failed: timeout 500ms exceeded
``` 

And it's even created this kind of error, which did not happen before AFAIK. Ref: p1659691831781659-slack-CQ0Q6N62D - you can see even more similar errors for this run on Slack report.

screenshot_of_failed_test

As I mentioned in the PR description, retrying a failed test right away is not a good fit for our test nature.

.github/workflows/e2e-test.yml
htdat: Addressed https://github.com/Automattic/woocommerce-payments/pull/4549/commits/ac705b939c80e2cc949a3df8fedaaf27ac56279a

.github/workflows/e2e-test.yml
htdat: Addressed https://github.com/Automattic/woocommerce-payments/pull/4549/commits/c1941ea604691e0a337be16859ed19a35ac317f2

deepakpathania:
> retrying a failed test right away is not a good fit for our test nature.

ack, makes sense. I was thinking whether doing it at the global level for a test file as I shared above would mean we wait for all the tests to finish in the file before retrying, but that doesn't seem to be the case. This approach seems better then.

deepakpathania (approved):
Thanks for addressing the comments, changes LGTM now. We can wait for @achyuthajoy's review as well but I am approving for now. ✅

achyuthajoy (approved):
Changes look good to me @htdat. Thanks for working on this 👍 

LGTM 🚢

PR #4577, opened 2022-08-12
E2E setup tests: Ignore Docker build ./wordpress-xdebug

### LINKED ISSUE #4585: E2E - Docker build fails when running inside GH actions
See p1660267127466569-slack-CGGCLBN58

#### Issue 

Sometimes we see the `wordpress` Docker build process failed like [here](https://github.com/Automattic/woocommerce-payments/runs/7800300481?check_suite_focus=true#step:14:35):

```
===> Starting CLIENT containers
Creating network "env_default" with the default driver
Pulling db (mariadb:10.5.8)...
Building wordpress
The command '/bin/sh -c pecl install xdebug-2.9.8 	&& echo 'xdebug.remote_enable=1' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_port=9000' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_host=host.docker.internal' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_autostart=0' >> $PHP_INI_DIR/php.ini 	&& docker-php-ext-enable xdebug > /dev/null' returned a non-zero code: 1
Service 'wordpress' failed to build : Build failed
```

htdat:
Fixes #4585 

See p1660267127466569-slack-CGGCLBN58

#### Issue 

Sometimes we see the `wordpress` Docker build process failed like [here](https://github.com/Automattic/woocommerce-payments/runs/7800300481?check_suite_focus=true#step:14:35):

```
===> Starting CLIENT containers
Creating network "env_default" with the default driver
Pulling db (mariadb:10.5.8)...
Building wordpress
The command '/bin/sh -c pecl install xdebug-2.9.8 	&& echo 'xdebug.remote_enable=1' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_port=9000' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_host=host.docker.internal' >> $PHP_INI_DIR/php.ini 	&& echo 'xdebug.remote_autostart=0' >> $PHP_INI_DIR/php.ini 	&& docker-php-ext-enable xdebug > /dev/null' returned a non-zero code: 1
Service 'wordpress' failed to build : Build failed
``` 

The purpose of the Dockerfile build `tests/e2e/env/wordpress-xdebug/Dockerfile` is to add xDebug, some other commands like `subversion mariadb-client less`, and WP-CLI. But we're not using them or they have already been available in the WordPress docker image: 

- We disabled xDebug for CI in the `setup.sh` script after building a new image for it: 
https://github.com/Automattic/woocommerce-payments/blob/2db0e16ddb0d5932888d9f8d7760abd6d2fd4063/tests/e2e/env/setup.sh#L108-L112

- We also use `wordpress:cli` in this `setup.sh` script so no need to embed it in the Dockerfile build:
https://github.com/Automattic/woocommerce-payments/blob/2db0e16ddb0d5932888d9f8d7760abd6d2fd4063/tests/e2e/env/setup.sh#L139-L140

My approach here is pretty aggressive, as I am completely removing the Dockerfile file for build. Though, it might be still useful if we need to eable xDebug for local E2E test sites. If so, we might update this PR to have more flexibility between "a build" for local vs "an image" for CI. 

An added benefit is that without building this Dockerfile, the `Prepare test environment` step is faster between 30-60s. 

#### Testing instructions

- Manually run E2E tests for this branch `e2e/ignore-docker-build-for-client-site` on https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-test.yml
- Confirm that all tests run fine.

Here is a run: https://github.com/Automattic/woocommerce-payments/actions/runs/2845444886

-------------------

**Post merge**


tests/e2e/env/docker-compose.yml
```diff
@@ -2,8 +2,7 @@ version: '3'
 
 services:
   wordpress:
-    build: ./wordpress-xdebug
-    image: wcp_e2e_wordpress
+    image: wordpress:5.6.2-php7.3
```
achyuthajoy, line 5: Could we also update the image to one with PHP v7.4? We specify PHP 7.4 for E2E tests, but the docker images still uses PHP 7.3.

tests/e2e/env/setup.sh
```diff
@@ -105,12 +105,6 @@ if [[ -z $CI ]]; then
 	docker-compose -f "$E2E_ROOT"/env/docker-compose.yml up --build --force-recreate -d phpMyAdmin
 fi
 
-if [[ -n $CI ]]; then
```
achyuthajoy, line 108: @htdat Like you mentioned, on previous modifications, I added this condition to allow xDebug for debugging locally. I'm OK with this change, let see if other have more queries.

achyuthajoy (commented):
Thanks for working on this @htdat and I've left some comments. The changes look good to me but haven't approved it. Let hear from other as well.

Also noting that this is the first time I'm seeing the docker build failing. Manual re-run solved the issue and I believe that the failure could be related to GH actions.

tests/e2e/env/docker-compose.yml
htdat: This image ID is from the to-be-deleted `Dockerfile`:
https://github.com/Automattic/woocommerce-payments/blob/5d19129845e132624bb08cf8f7bb325803f04516/tests/e2e/env/wordpress-xdebug/Dockerfile#L1

But yeah, I can add it to PHP 7.4. I will update the WP version to `5.8` too as it's our minimum requirement now:
https://github.com/Automattic/woocommerce-payments/blob/5d19129845e132624bb08cf8f7bb325803f04516/woocommerce-payments.php#L13-L14

achyuthajoy: That would be great. Thanks 🤝

htdat:
> Also noting that this is the first time I'm seeing the docker build failing. 

This is not my first time though. However, even so, letting Docker build a new image and then not use additional features is kinda... wasteful. 

In the past, we might need these additional features but from what I see, we do not need them, at least with our current E2E tests.

> Manual re-run solved the issue and I believe that the failure could be related to GH actions.

Looking at the error I quoted above, I think it's an issue when trying to install `xdebug`:

```
The command '/bin/sh -c pecl install xdebug-2.9.8 [...] returned a non-zero code: 1
```

tests/e2e/env/docker-compose.yml
htdat: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/4577/commits/075050509db89848271367782d8b03c868c93468

htdat:
Addressed feedback. Ready for another review. 

@achyuthajoy - since I will be off soon, please help me merge this PR if we get consensus about this change.

tests/e2e/env/docker-compose.yml
achyuthajoy: Awesome. Thanks 👍

achyuthajoy:
> Addressed feedback. Ready for another review.
> 
> @achyuthajoy - since I will be off soon, please help me merge this PR if we get consensus about this change.

Sure @htdat. I'll merge the PR when approved.

tests/e2e/env/docker-compose.yml
```diff
@@ -2,8 +2,7 @@ version: '3'
 
 services:
   wordpress:
-    build: ./wordpress-xdebug
-    image: wcp_e2e_wordpress
+    image: wordpress:5.8.3-php7.4
```
deepakpathania, line 5: I am ok with the change as well, only concern is whether updating this image as we bump versions would become an additional task for us. Any way to get the `latest` we support here?

achyuthajoy: I don't think it's an issue @deepakpathania. We update the WP & WC versions before running tests.

htdat: > We update the WP & WC versions before running tests.

Yes, that's right. Also, if using `latest`, I am not sure if we can run the downgrade with this command: 
https://github.com/Automattic/woocommerce-payments/blob/5d19129845e132624bb08cf8f7bb325803f04516/tests/e2e/env/setup.sh#L152-L155

Based on the description here https://developer.wordpress.org/cli/commands/core/update/, it's for `new versions`.

achyuthajoy:
Closing the PR based on the discussion to keep development & test environment in sync. With the suggested change we would be using a different docker image from what's present on development environment.

PR #4587, opened 2022-08-15
Duplicated payment methods after paying for a subscription via WooPay

### LINKED ISSUE #4571: Prevent non-attached payment methods from showing up in the My Account page
### Description

Requires 2359-gh-Automattic/woocommerce-payments-server

Under the merchant site -> My Account -> Payment methods page, there can be duplicates of the same payment method even though only one instance is attached to the customer object. This happens when coming back from paying for a subscription via WooPay using a card that already exists on the merchant site.

### Acceptance criteria

* No more than a single instance of the same payment method must be shown under My Account -> Payment methods
* Subscriptions and renewals must continue to work

### Testing instructions

1. Create a subscription product in the merchant site
2. Purchase it via WooPay (use an account that already exists in WooPay)
3. On the merchant site, go to My Account -> Payment methods
  Notice the card you selected in WooPay is now listed
4. Purchase a subscription again via WooPay. Select the same card you selected before
5. On the merchant site, go to My Account -> Payment methods
  Notice that now there are two instances of the same card

### Dev notes
If you go to the Stripe dashboard -> Merchant account -> customer you just used, you can see that only a single instance of the payment method exists in the customer. This means that the problem here is likely on how WCPay handles the payment methods rather than how WooPay is attaching them.

a-danae:
Fixes #4571

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1300,12 +1300,32 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 
 			if ( $save_payment_method_to_store && ! $intent_failed ) {
 				try {
+					$token = null;
+
 					// Setup intents are currently not deserialized as payment intents are, so check if it's an array first.
 					// For payment intents, we may provide a platform payment method from `$payment_information`, but we need
 					// to return a connected payment method. So we should always retrieve the payment method from the intent.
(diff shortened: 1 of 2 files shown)


brianyu28 (approved):
Thanks @a-danae! This fix works, and the changes look good to me. Thanks for the especially detailed test cases, too.

While testing this, I realized that it's possible to use WooPay to pay for early renewal payments via the subscription management screen, regardless of whether the original subscription was bought using WooPay. I tested renewals both with and without WooPay, and both work well with the changes in this PR.

PR #4595, opened 2022-08-17
Fix charge id validation for UPE transactions

### LINKED ISSUE #3122: Use the PaymentIntent ID in the Transaction Details URL instead of the Charge ID
### Description

This issue is created for [this comment](https://github.com/Automattic/woocommerce-payments/issues/540#issuecomment-607588704) on #540 

Right now, URLs for Transaction Detail screens use the charge ID (e.g. `admin.php?page=wc-admin&path=%2Fpayments%2Ftransactions%2Fdetails&id=ch_xxxx`). [This PR](https://github.com/Automattic/woocommerce-payments/pull/2937) replaces the Charge ID at the top of the card with the Payment Intent ID. So, it makes sense that the Payment Intent ID should be used in the URL too.

### Acceptance criteria

* The Transaction Details page must have the Payment Intent `pi_xxx` in the URL instead of the Charge ID `ch_xxx`
* Accessing the Transaction Details page using the Charge ID `ch_xxx` should redirect to the URL using the Payment Intent ID `pi_xxx`
* The plugin's links pointing to the Transaction Details page should use the Payment Intent
* Of course, everything should continue to load and work fine after the switch

### Testing instructions

1. Go to the Transactions page
2. Click on a transaction to go to its Transactions Details page
3. The URL should have the Payment Intent ID `pi_xxx` instead of the Charge ID `ch_xxx`

There are other scenarios to test. To be added.

### Dev notes

Extracted from the original issue:
* Related #353
* When updating the URL we should also update the gateways get_transaction_url method as added in [this PR](https://github.com/Automattic/woocommerce-payments/pull/837)

### Additional context

Quoting Brent's comment:

> Of note, in the Stripe dashboard:
> 
> * The Connect Accounts dashboard uses a Charge ID under the Payments tab, e.g. https://d.pr/i/pZRNRM
> * Clicking one of those rows takes you to a URL for the account with the Charge ID, e.g. `https://dashboard.stripe.com/acct_xxxx/test/payments/ch_xxxx`
> * Stripe redirects that URL to a URL using the Payment Intent ID, e.g. `https://dashboard.stripe.com/acct_xxxx/test/payments/pi_xxxx`
> * The dashboard for an individual account display's Payment Intent IDs under the Payments section e.g. https://d.pr/i/lU3PxC

Also, we should add a redirect for backward compatibility. When going to the Transaction Details URL using the Charge ID `ch_xxx`, the user should be redirected to the URL using the PaymentIntent URL `pi_xxx`.

That's an old issue, but I've confirmed with the team in slack p1631830055418100-slack-CGGCLBN58 that we still want to do this move.

EduardoUmpierre:
Fixes #3122

An issue was added in 4.5.0 as the Transaction Details Page was invalidating `py_` charge ids.

This commit adds the `py_` charge id variant as a valid id.

#### Testing instructions

- Git checkout the server to use the [latest commit](https://github.com/Automattic/woocommerce-payments-server/commit/6f3beb6eaecdd5594724086be19b6268d056527c)
- Enable a UPE payment method like Bancontact, P24, and iDeal
- Enable Multi-Currency and the currency switcher in the WC Pay settings
- Use the currency switcher to pay in EUR on the client site and add an item to the cart
- Go through checkout using one of the payment methods above
- An error should pop up when processing the payment redirect
- Open the transactions page and click on the new transaction
- Make sure the transaction details page is loaded successfully

-------------------

**Post merge**

### DIFF
--- a/client/data/payment-intents/hooks.ts
+++ b/client/data/payment-intents/hooks.ts
@@ -10,7 +10,7 @@ import { ChargeResponse } from '../charges/types';
 import { STORE_NAME } from '../constants';
 
 export const getIsChargeId = ( id: string ): boolean =>
-	-1 !== id.indexOf( 'ch_' );
+	-1 !== id.indexOf( 'ch_' ) || -1 !== id.indexOf( 'py_' );
 
 export const usePaymentIntentWithChargeFallback = (
 	id: string
--- a/changelog/fix-upe-charge-id
+++ b/changelog/fix-upe-charge-id
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix an issue that was added in 4.5.0 as the Transaction Detail Page was invalidating py_ charge ids.
--- a/changelog/fix-upe-charge-id
+++ b/changelog/fix-upe-charge-id
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Fix an issue that was added in 4.5.0 as the Transaction Detail Page was invalidating py_ charge ids.
+Fix an issue while loading the Transaction Detail Page with py_ charge ids.


client/data/payment-intents/hooks.ts
```diff
@@ -10,7 +10,7 @@ import { ChargeResponse } from '../charges/types';
 import { STORE_NAME } from '../constants';
 
 export const getIsChargeId = ( id: string ): boolean =>
-	-1 !== id.indexOf( 'ch_' );
+	-1 !== id.indexOf( 'ch_' ) || -1 !== id.indexOf( 'py_' );
```
ricardo, line 13: Should we still use this validation, or remove it like in the server?

EduardoUmpierre: We need this in the client to perform the fallback behavior of loading the transaction details from the charge id. It will fetch the payment intent data from the charge id and then redirect to the equivalent `pi_` id.

EduardoUmpierre: We could rely on the payment intent id and assume that if it doesn't include `pi_` then it's a charge id.

luizreis: In this case, newer orders shouldn't generate the `charge_id` links, as everything will switch to using the payment intent ID. The check here is to provide the fallback to old links and we need it to identify if it is using an intent or charge to request the page.

luizreis (approved):
LGTM! Tested with a few orphan transactions and the details page loaded properly. :shipit:

PR #4500, opened 2022-07-29
Supporting changes for WooPay Trial Subscriptions

malithsen:
This PR introduces changes to support processing trial subscriptions in platform-store. Why these changes are required is documented in 1158-gh-Automattic/woopay

#### Changes proposed in this Pull Request

* Add a `save_in_platform_account` field to the setup-intent request
* Introduce a hook to override intent-less payment completions. This is used in the platform-store to process free trial subscriptions.
* Accept setup-intents from platform-store to complete orders in the merchant store.

#### Testing instructions

Please refer to the instructions in 1167-gh-Automattic/woopay

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

Issue #2370, opened 2021-07-07
Update Preview Mode Experiment Name


adrianduffell:
Change `wcpay_empty_state_preview_mode` to `wcpay_empty_state_preview_mode_v1`.

PR #4601, opened 2022-08-18
Fix adding payment tokens for platform-created setup intents

brianyu28:
Addition to #4500 to fix a bug related to 1158-gh-Automattic/woopay

When subscriptions with free trials are purchased via WooPay, currently no payment token is added to the customer's account, so renewals don't process correctly. This is because we attempt to add the payment method with the wrong ID — we're adding the platform payment method (present in the payment information) instead of the connected payment method (present on the intent). This PR fixes the logic for setup intents to behave the same way as payment intents, so that the correct payment method is added.

cc @ricardo — since this is core payment processing code, out of caution I'm asking for another pair of eyes and I see that you were the one who originally wrote the lines in question for payment intents. Essentially what this PR aims to do is to do the same for setup intents so that we're always using the connected payment method instead of the platform payment method.

#### Testing instructions

**Setup**

- Check out 1167-gh-Automattic/woopay and 2225-gh-Automattic/woocommerce-payments-server
- Check "Enable WooPay for subscriptions (Work in progress)" in WCPay Dev Tools
- Install and activate the WooCommerce Subscriptions extension
- Create a simple subscription product with a non-zero free trial.

**Test**
- As the customer, add the subscription product (which has a free trial) to the cart and check out via WooPay.
- On the WooPay checkout screen, verify that the initial total is $0 but the recurring total is the correct recurring amount.
- Complete checkout.
- As the merchant, go to WooCommerce > Subscriptions and click on the subscription. Verify that it is active and was created successfully.
- Click the pencil next to "Billing" to see subscription details, you should see the correct payment method selected automatically under "Saved payment method"
- Under "Subscription actions" in the upper right, "Process renewal". Verify that the renewal processes and the customer is charged.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-woopay-trial-subscription-payment-tokens
+++ b/changelog/fix-woopay-trial-subscription-payment-tokens
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Fixes adding payment tokens for platform-created setup intents
+Fix adding payment tokens for platform-created setup intents
--- a/changelog/fix-woopay-trial-subscription-payment-tokens
+++ b/changelog/fix-woopay-trial-subscription-payment-tokens
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixes adding payment tokens for platform-created setup intents
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1323,9 +1323,9 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 					$token = null;
 
 					// Setup intents are currently not deserialized as payment intents are, so check if it's an array first.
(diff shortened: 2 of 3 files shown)


leonardola (approved):
Works as described. Thanks for the quick solution @brianyu28

ricardo (approved):
LGTM and tests well.

> `$payment_information->get_payment_method()`

IIRC, before #3762, this worked for both payment intents and setup intents. I just refactored it for payment intents to use `$intent->get_payment_method_id()` because setup intents weren't supposed to work at the time, but I guess I would have done the same thing now 👍.

I also noticed that free trial subscriptions were creating duplicate payment methods on the merchant site, but there's already an existing issue for this #2370 😎.

brianyu28:
Thanks for the review, @leonardola and @ricardo!

PR #3597, opened 2022-01-05
Upgrade WP from 5.4 to 5.6 for the edge case in the compatibility CI tests

htdat:
Coming from this comment https://github.com/Automattic/woocommerce-payments/pull/3592#discussion_r778521350

#### Issue 

The current process of setting up test sites, we install the latest version 12.0.x of Gutenberg, which is only compatible with WP 5.7+. 

This issue has already existed for a while but the behavior is different among WP versions: 

- With WP < 5.4, it's just not activating Gutenberg without yelling any error. See [this test run](https://github.com/Automattic/woocommerce-payments/runs/4705827838?check_suite_focus=true#step:5:139).
> Warning: Failed to activate plugin. Current WordPress version does not meet minimum requirements for Gutenberg.
> Success: Installed 1 of 1 plugins.

- Since WP 5.5 ([src](https://github.com/WordPress/wordpress-develop/commit/289c28637ccaa75b5711b82904fff05e2c89bd36#diff-679826964bdb1599506cdfa3508ccd982e2f21f3de25e6f686b5aefc9175fa0dR272) - [core trac](https://core.trac.wordpress.org/ticket/9757)), it stops installing the plugin and throw this error. [See this test run](https://github.com/Automattic/woocommerce-payments/runs/4705683866?check_suite_focus=true#step:5:137).
> Warning: The package could not be installed. "Your WordPress version is 5.6, however the uploaded plugin requires 5.7."
> Plugin installation failed.

#### Changes proposed in this Pull Request

- Add Gutenberg versions in GitHub workflows and relevant bash scripts.
- Upgrade WP from 5.4 to 5.6 for the edge case in the compatibility CI tests.

#### Testing instructions

* Ensure that there is no CI test failure.

-------------------

- [x] Added changelog entries to both `readme.txt` and `changelog.txt` (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Added testing instructions to the [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) (or does not apply)

PR #4615, opened 2022-08-22
Bump WC tested up to and WC minimum version for 4.7.0 release

a-danae:
Fixes #

- Bump minimum WC version from 6.2.2 to 6.4.1
- Add WC 6.9.0 (beta to be released on Tuesday 23) as the "WC tested up to" version

No WP updates to catch up on.

#### Testing instructions

- Verify that wherever we used 6.2 as a minimum before, we now use 6.4. Update. Workflows to be updated in a separate PR as per [this conversation](https://github.com/Automattic/woocommerce-payments/pull/4615#discussion_r951952678).
- Please check the CI tests in the GH workflows.

-------------------

**Post merge**

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -12,13 +12,13 @@ jobs:
       max-parallel: 10
       matrix:
         # Minimum support version, ..., Latest version, `beta`.
-        woocommerce: [ '6.2.2', '6.3.1', '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
+        woocommerce: [ '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
         php:         [ '7.4' ]
         include:
           # Edge case: oldest dependencies compatibility
-          - woocommerce: '6.2.2'
+          - woocommerce: '6.4.1'
             wordpress:   '5.8'
             gutenberg:   '11.4.0' # The latest version supporting WP 5.6.
             php:         '7.1' # Minimum supported PHP version
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -12,13 +12,13 @@ jobs:
       max-parallel: 10
       matrix:
         # Minimum support version, ..., Latest version, `beta`.
-        woocommerce: [ '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
+        woocommerce: [ '6.2.2', '6.3.1', '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
         php:         [ '7.4' ]
         include:
           # Edge case: oldest dependencies compatibility
-          - woocommerce: '6.4.1'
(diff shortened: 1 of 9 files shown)


.github/workflows/compatibility.yml
```diff
@@ -12,13 +12,13 @@ jobs:
...
+        woocommerce: [ '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
         php:         [ '7.4' ]
         include:
           # Edge case: oldest dependencies compatibility
-          - woocommerce: '6.2.2'
+          - woocommerce: '6.4.1'
             wordpress:   '5.8'
             gutenberg:   '11.4.0' # The latest version supporting WP 5.6.
```
a-danae, line 23: @brianyu28 do you know if we should update the Gutenberg version to the latest one supporting our oldest compatible WP version? In this case, looks like the latest version supporting 5.8 is [13.6.0](https://plugins.trac.wordpress.org/browser/gutenberg/tags/13.6.0/gutenberg.php).

brianyu28: It looks like the [reason](https://github.com/Automattic/woocommerce-payments/pull/3597) for specifying the `gutenberg` version here was to avoid installing a version of Gutenberg that required a more recent WP version, and before that we would always use the latest version. So updating this to the 13.6.0, the latest version that we can, makes sense to me.

brianyu28 (approved):
Looks good, thanks @a-danae! Updated versions look accurate to me — I think it would make sense to update the `gutenberg` version here, though the tests seem to be running okay even on the older version. I see one test failing in the GH workflows but it seems unrelated to this PR.

.github/workflows/compatibility.yml
```diff
@@ -12,13 +12,13 @@ jobs:
...
         # Minimum support version, ..., Latest version, `beta`.
-        woocommerce: [ '6.2.2', '6.3.1', '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
+        woocommerce: [ '6.4.1', '6.5.1', '6.6.1', '6.7.0', 'latest', 'beta' ]
         wordpress:   [ 'latest' ]
         gutenberg:   [ 'latest' ]
         php:         [ '7.4' ]
         include:
           # Edge case: oldest dependencies compatibility
-          - woocommerce: '6.2.2'
+          - woocommerce: '6.4.1'
```
a-danae, line 21: This combination is failing in the PHP unit tests. I can't replicate locally. I get a fatal error when using Gutenberg 11.4.0.
For newer versions, I do get the `{"success":false,"data":"You aren\u2019t authorized to do that."}{"result":"failure"}` error but my test doesn't fail with `Failed asserting that an array is empty.`.

@brianyu28  I'm thinking of updating the workflows in a separate PR so we can move forward and prepare the package for testing while sorting things out. Do you think that'd be reasonable?

brianyu28: Yeah, I agree it makes sense to get the testing package out first. The CI test updates I believe are meant to happen during the first week of the release cycle but doesn't need to be on Monday.

a-danae: Cool. Reverting the commit that updates the workflows for now. Thanks!

.github/workflows/compatibility.yml
a-danae: Nice. I'll update the `gutenberg` version here when updating the workflows in a separate PR. Thanks!

Issue #4579, opened 2022-08-15
Product Variations without Prices Result in Critical Error on Product Page in PHP version 8.0


druesome:
### Describe the bug

When a variable product or subscription is created with no prices added to any variation, the product page results in a Critical error in PHP version 8.0.

### To Reproduce

1. On an Atomic site, go to Settings > Hosting Config and set PHP to version 8.0
2. Create a variable product or subscription and leave all prices empty
3. Publish Product
4. Visit the product page

### Actual behavior

The product page displays a critical error.  In CLI, the full error log looks like this:

```
PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string + int in /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-payment-request-button-handler.php:295
Stack trace:
#0 /srv/htdocs/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-payment-request-button-handler.php(656): WC_Payments_Payment_Request_Button_Handler->get_product_data()
#1 /wordpress/core/6.0.1/wp-includes/class-wp-hook.php(307): WC_Payments_Payment_Request_Button_Handler->scripts('')
#2 /wordpress/core/6.0.1/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#3 /wordpress/core/6.0.1/wp-includes/plugin.php(476): WP_Hook->do_action(Array)
#4 /wordpress/core/6.0.1/wp-includes/script-loader.php(2156): do_action('wp_enqueue_scri...')
#5 /wordpress/core/6.0.1/wp-includes/class-wp-hook.php(307): wp_enqueue_scripts('')
#6 /wordpress/core/6.0.1/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#7 /wordpress/core/6.0.1/wp-includes/plugin.php(476): WP_Hook->do_action(Array)
#8 /wordpress/core/6.0.1/wp-includes/general-template.php(3042): do_action('wp_head')
#9 /wordpress/core/6.0.1/wp-includes/template-canvas.php(17): wp_head()
#10 /wordpress/core/6.0.1/wp-includes/template-loader.php(106): include('/wordpress/core...')
#11 /wordpress/core/6.0.1/wp-blog-header.php(19): require_once('/wordpress/core...')
#12 /wordpress/core/6.0.1/index.php(17): require('/wordpress/core...')
#13 {main}
[...]

PR #4624, opened 2022-08-23
Added numeric validation check to product price

zmaglica:
Fixes #[4579](https://github.com/Automattic/woocommerce-payments/issues/4579)

When the product doesn't have price, the exception will be thrown because the function that is used for formatting the price with currency expects the number as a parameter inside request payment button handler.

This PR will prevent that by validating a price before the function that is used for formatting is called. If the product doesn't have a valid price, the function will exit without product data.

#### Testing instructions

1. Create a variable product or subscription and leave all prices empty
2. Publish product
3. Visit product and make sure that no errors are shown

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -239,11 +239,6 @@ class WC_Payments_Payment_Request_Button_Handler {
 
 		/** @var WC_Product_Variable $product */ // phpcs:ignore
 		$product  = $this->get_product();
-
-		$product_price = $this->get_product_price( $product );
-		if ( ! is_numeric( $product_price ) ) {
-			return false;
-		}
 		$currency = get_woocommerce_currency();
 
 		if ( 'variable' === $product->get_type() || 'variable-subscription' === $product->get_type() ) {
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -239,6 +239,11 @@ public function get_product_data() {
 
 		/** @var WC_Product_Variable $product */ // phpcs:ignore
 		$product  = $this->get_product();
+
+		$product_price = $this->get_product_price( $product );
+		if ( ! is_numeric( $product_price ) ) {
```
vbelolapotkov: @zmaglica I doubt this is the right place to do the check. `get_product_price` will probably throw error if you create subscription product without price anyways.

`should_show_payment_request_button` method seem to be much better place to add the logic: https://github.com/Automattic/woocommerce-payments/blob/ed958c921d8c54e5e2a7d1bced59ea222f845d81/includes/class-wc-payments-payment-request-button-handler.php#L475-L475

zmaglica: Fixed.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -515,6 +515,11 @@ public function should_show_payment_request_button() {
 			Logger::log( 'Items in the cart have unsupported product type ( Payment Request button disabled )' );
 			return false;
 		}
+		$product_price = $this->get_product_price( $this->get_product() );
```
vbelolapotkov, line 518: @zmaglica have you checked if it works for subscription product?

PR #4627, opened 2022-08-23
added meta on payment tokens being used by subscriptions

leonardola:

This PR adds a meta to payment tokens that are used by subscriptions so we can show the correct message when deleting it via WooPay. Should be tested alongside with 1218-gh-Automattic/woopay

#### Testing instructions

- Follow the instructions on 1218-gh-Automattic/woopay

-------------------


a-danae (approved):
LGTM 👍

PR #4637, opened 2022-08-25
Add bundle size check to PR's

### LINKED ISSUE #4619: Add bundle size report to PR
### Description

Similar to [this PR](https://github.com/Automattic/woopay/pull/371), we want to introduce a bundle size reporting feature to WCPay as well. We can use [compressed-size-action](https://github.com/preactjs/compressed-size-action) to achieve this.

When implemented, this feature will add a report to each PR showing the size changes introduced by that PR.

fashuvo:
Fixes #4619

This PR adds a GH action for bundle size reporting. This is helpful to detect any unexpected size bump caused by the PR.

Here's a screenshot showing the size changes in a draft PR:

Image 2022-08-25 at 2 30 37 PM

#### Testing instructions

1. Clone the `add/4619-add-bundle-size-check` branch in your local machine.
2. Create a new branch on top of that.
3. Delete/add some code from some file(s).
4. Push that new branch.
5. Create a draft PR.
6. The size change report should be visible once the action completes.

-------------------

**Post merge**


.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,21 @@
...
+    build:
+        name: Bundle size check
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
+            - name: Use desired version of NodeJS
+              uses: actions/setup-node@v2
+              with:
+                  node-version: '12'
```
achyuthajoy, line 17: @fashuvo Instead of specifying a specific node version, we can make use of the version specified in `.nvmrc` file. So we are in sync with the node version of the project.

Here is a similar implementation - https://github.com/Automattic/woocommerce-payments/blob/5b9809465e0fb9ec3476d40f82b85daf82d512e9/.github/workflows/e2e-test.yml#L103-L107

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,21 @@
...
+                  cache: npm
+            - uses: preactjs/compressed-size-action@v2
+              with:
+                  pattern: '{release/**/*.js,release/**/*.tsx,release/**/*.css,release/**/*.php}'
```
achyuthajoy: @fashuvo Do we want to check size of PHP files here? JS, TS & CSS seems fine.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+    build:
+        name: Bundle size check
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
+            - name: Use desired version of NodeJS
+              uses: actions/setup-node@v2
```
achyuthajoy, line 15: @fashuvo Instead of specifying a specific node version, we can make use of the version specified in .nvmrc file. So we are in sync with the node version of the project.

Here is a similar implementation - https://github.com/Automattic/woocommerce-payments/blob/5b9809465e0fb9ec3476d40f82b85daf82d512e9/.github/workflows/e2e-test.yml#L103-L107

achyuthajoy (changes requested):
@fashuvo The changes looks good. I've left a few comments

.github/workflows/bundle-size.yml
fashuvo: Thanks! Updated the PR to use node version from .nvmrc.

.github/workflows/bundle-size.yml
fashuvo: I wasn't sure about that as well, but I thought we track size changes of PHP files too.

Why shouldn't we check the PHP file size? Is that because we don't work with PHP that often, or is there any other reason?

achyuthajoy: > Why shouldn't we check the PHP file size? 

I don't think we use much external lib/helpers with PHP, so a size increase that way is less likely to happen. Usually, bundle size checks are required for front-end assets like JS, CSS, Images etc. PHP files are called and executed only when required.

@RadoslavGeorgiev what do you think about checking bundle size for PHP files?

achyuthajoy:
@fashuvo It seems this check is taking a lot of time - https://github.com/Automattic/woocommerce-payments/runs/8017810298?check_suite_focus=true

I'm not sure whether it's related to node version change. We use `14` on `.nvmrc`. And it is the suggested version for building WCPay.

fashuvo:
> @fashuvo It seems this check is taking a lot of time - https://github.com/Automattic/woocommerce-payments/runs/8017810298?check_suite_focus=true
> 
> I'm not sure whether it's related to node version change. We use `14` on `.nvmrc`. And it is the suggested version for building WCPay.

Yeah, I have also noticed that the check takes about 20 minutes to complete: https://github.com/Automattic/woocommerce-payments/actions/workflows/bundle-size.yml

The duration was the same with Node V.12 as well. Do you think we can do anything to reduce this, @achyuthajoy ?

.github/workflows/bundle-size.yml
anu-rock: I agree with @achyuthajoy's suggestion. As this check targets the client-side bundle generated by webpack, it's safe to exclude what is not considered an asset file. Anything that is statically downloaded by the browser is an asset file - JS, CSS, HTML, images, videos, etc. PHP files are executed on the server side, and are not static assets. In fact, it may be reasonable to include only `release/assets` and `release/dist` in this check.

Through this check, we'd like to keep the total size of files (assets) downloaded by the browser under reasonable limits so that a user's browsing experience (especially on slow devices such as mobiles) is not hampered because of unnecessary files.

`.tsx` files normally get compiled as .`js` files. I wonder if we need them here.

fashuvo: Thanks for your detailed explanation @anu-rock 🙇 

Your explanation makes sense, and I have removed checks for PHP and .tsx files.

fashuvo:
In a recent call with @zmaglica, we tried to make it faster by excluding the vendor folder, but it didn't make any difference. We found out that the building process takes 90% of the time because of the multiple building steps. And it is not possible to avoid this step. The bundle size check takes similar or more time in [other repos as well](https://github.com/Automattic/woopay/actions/workflows/bundle-size.yml).

Do you have any suggestion about reducing the size check duration, @achyuthajoy?

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,23 @@
...
+              with:
+                  node-version-file: '.nvmrc'
+                  cache: npm
+            - uses: preactjs/compressed-size-action@v2
```
achyuthajoy: @fashuvo Just wanted to note that`preactjs/compressed-size-action` is a 3rd party action and I've noticed that some of our repos uses this action. I don't prefer using 3rd party runners on our setup mostly because it can access repository secrets and if compromised, could access other private repositories via the GH token supplied. More info on this here - https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

If we don't have an alternative method, we could go with the current approach, but we would need to watch out for security issues.

achyuthajoy (commented):
@fashuvo I've left a comment regarding usage of 3rd party actions. Regarding the test speed, I don't have a workaround.

The current build process removes and installs all dependencies, which takes about 7 - 10 minutes and needs to run twice - for building the PR branch & `develop` before comparing the changes. Unless we can introduce a new command & make use of dependency caching, it won't be possible to reduce the build time at this time.

I haven't approved the changes yet. Let me know once you go through the comment regarding use of 3rd party actions.

.github/workflows/bundle-size.yml
achyuthajoy: @fashuvo Based on our discussion regarding the security concern, let's lock `preactjs/compressed-size-action` to the latest commit SHA. So it would be `preactjs/compressed-size-action@b03ec15e1a93cb5c062d66a1c263f9ddbd2f81ef`.

This prevents the action from including newer changes during our run. And if we want to update a new version, we can do an audit of the changes made after this commit and update the commit SHA, which is much safer.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,23 @@
...
+              with:
+                  repo-token: '${{ secrets.GITHUB_TOKEN }}'
```
achyuthajoy: I don't see `GITHUB_TOKEN` secret under repository secrets. Have you added it? 

If you haven't, try creating it with access only to public repositories. I haven't tested it, but let me know how it goes :) 

Screenshot 2022-09-22 at 9 57 32 PM

achyuthajoy (changes requested):
@fashuvo I've left a couple of comments on how we can address the security concern. Let me know if you have any suggestions/questions.

.github/workflows/bundle-size.yml
fashuvo: I have updated the check to use the latest commit SHA.

.github/workflows/bundle-size.yml
fashuvo: I have created a token with public repo access and added it to the repo secrets.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,23 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v2
+            - name: Use desired version of NodeJS
+              uses: actions/setup-node@v2
+              with:
+                  node-version-file: '.nvmrc'
+                  cache: npm
+            - uses: preactjs/compressed-size-action@b03ec15e1a93cb5c062d66a1c263f9ddbd2f81ef
```
achyuthajoy: Can we add a step name here, so it won't print out `Run preactjs/compressed-size-action@b03ec15e1a93cb5c062d66a1c263f9ddbd2f81ef` on the job logs.

Something like "Checking for file changes"

achyuthajoy (approved):
@fashuvo I've left one more comment. However, it's not a blocker for merging the changes and I've approved the PR.

Thanks for working on this 👍

.github/workflows/bundle-size.yml
fashuvo: Great suggestion! I have added a step name.

PR #4638, opened 2022-08-25
Support new platform checkout user from checkout blocks

### LINKED ISSUE #4614: Support creating new Platform Checkout user on Checkout Blocks
### Description
Currently, if the checkout page is using shortcode-checkout only then Platform Checkout is supported and a new user is given the opportunity to save their details for using Platform Checkout on their future purchases. Similar support needs to be added when the checkout page is using Checkout Blocks too.

### Acceptance criteria
- Run checks for an existing platform checkout user when a valid email is provided. This will be done in #4613 
- If the user does not exist, show the `Remember your details?` section.
- If the checkbox is selected, when the user clicks `Place Order` and the checkout is successful, create a Platform Checkout user.

Mayisha:
Fixes #4614 

### Description
Added support to create new platform checkout user from Checkout Blocks.
- Inserting the existing `CheckoutPageSaveUser` in blocks checkout page if platform checkout is enabled.
- On blocks checkout page the status of `save user` checkbox and the phone number do not make their way to the backend on POST request. On shortcode, adding a proper name to the element was enough to piggyback them with the POST request but blocks only sends [this set of data](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/assets/js/base/context/providers/cart-checkout/checkout-processor.js#L185-L193) in request body. To get the values in backend used [extensionCartUpdate](https://github.com/woocommerce/woocommerce-blocks/blob/00a1174aebefae2b37178a3ded68cc6c8d56bbf4/docs/third-party-developers/extensibility/rest-api/extend-rest-api-update-cart.md#registering-a-callback-to-run-when-the-cartextensions-endpoint-is-hit) as the solution.
- As part of the above solution, registered a callback for `cart/extensions` endpoint and calling the `extensionCartUpdate` function with desired data when the `Place Order` button is clicked.
- As we need the value of `save_user_in_platform_checkout` selection in multiple places in backend (basically in multiple steps in a checkout flow), saved the value in `WC` session.

### Testing steps 
- Use `[shortcode-checkout]` in the checkout page. Ensure that nothing is broken there, new and existing user flow works same as before.
- Use `Checkout` block in the checkout page
- Go to the store, add some product and go to the blocks checkout page.
- Adding a new platform checkout user email in the email field will show the `Remember your details` section. Select the `save user` checkbox, complete the checkout with this new email. This should create a new platform checkout user.
- Adding an existing platform checkout user in the email field will hide the `Remember your details` section.
- Check and ensure that the experience on the blocks checkout page exactly same as the existing feature in the short-code checkout page.


fashuvo:
**Size Change:** +338 B (0%) 

**Total Size:** 1.7 MB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `release/woocommerce-payments/dist/platform-checkout.js` | 77.2 kB | +338 B (0%) |

<details><summary>ℹ️ <strong>View Unchanged</strong></summary>

| Filename | Size |
| :--- | :---: |
| `release/woocommerce-payments/assets/css/admin.css` | 913 B |
| `release/woocommerce-payments/assets/css/success.css` | 401 B |
| `release/woocommerce-payments/dist/blocks-checkout.css` | 1.77 kB |
| `release/woocommerce-payments/dist/blocks-checkout.js` | 34 kB |
| `release/woocommerce-payments/dist/checkout.css` | 1.85 kB |
| `release/woocommerce-payments/dist/checkout.js` | 26.9 kB |
| `release/woocommerce-payments/dist/index.css` | 35.8 kB |
| `release/woocommerce-payments/dist/index.js` | 640 kB |
| `release/woocommerce-payments/dist/multi-currency-analytics.js` | 8.41 kB |
| `release/woocommerce-payments/dist/multi-currency-switcher-block.js` | 79 kB |
| `release/woocommerce-payments/dist/multi-currency.css` | 14.7 kB |
| `release/woocommerce-payments/dist/multi-currency.js` | 117 kB |
| `release/woocommerce-payments/dist/onboarding.css` | 1.12 kB |
| `release/woocommerce-payments/dist/onboarding.js` | 38.1 kB |
| `release/woocommerce-payments/dist/order.js` | 299 B |
| `release/woocommerce-payments/dist/payment-gateways.css` | 1.2 kB |
| `release/woocommerce-payments/dist/payment-gateways.js` | 91 kB |
| `release/woocommerce-payments/dist/payment-request.js` | 23 kB |
| `release/woocommerce-payments/dist/platform-checkout.css` | 3.73 kB |
| `release/woocommerce-payments/dist/settings.css` | 47.2 kB |
| `release/woocommerce-payments/dist/settings.js` | 157 kB |
| `release/woocommerce-payments/dist/subscription-edit-page.js` | 3.66 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-modal.css` | 7.18 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-modal.js` | 16.7 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-toast.js` | 428 B |
| `release/woocommerce-payments/dist/subscriptions-empty-state.css` | 298 B |
| `release/woocommerce-payments/dist/subscriptions-empty-state.js` | 49 kB |
| `release/woocommerce-payments/dist/tos.css` | 236 B |
| `release/woocommerce-payments/dist/tos.js` | 31.7 kB |
| `release/woocommerce-payments/dist/upe_checkout.css` | 1.85 kB |
| `release/woocommerce-payments/dist/upe_checkout.js` | 38 kB |
| `release/woocommerce-payments/dist/upe-blocks-checkout.css` | 1.77 kB |
| `release/woocommerce-payments/dist/upe-blocks-checkout.js` | 44.8 kB |
| `release/woocommerce-payments/includes/subscriptions/assets/css/plugin-page.css` | 633 B |
| `release/woocommerce-payments/includes/subscriptions/assets/js/plugin-page.js` | 355 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-admin-ui/src/css/jetpack-icon.css` | 224 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-assets/build/i18n-loader.js` | 2.43 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-assets/src/js/i18n-loader.js` | 1.01 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/babel.config.js` | 160 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.css` | 2.27 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.js` | 13.3 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.rtl.css` | 2.27 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-tracking/src/js/tracks-ajax.js` | 789 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-tracking/src/js/tracks-callables.js` | 925 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/about.css` | 1.2 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/admin-order-statuses.css` | 403 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/admin.css` | 3.43 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/checkout.css` | 299 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/modal.css` | 742 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/view-subscription.css` | 572 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/wcs-upgrade.css` | 411 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/admin-pointers.js` | 544 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/admin.js` | 8.97 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/jstz.js` | 6.8 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/jstz.min.js` | 3.83 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/meta-boxes-coupon.js` | 544 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/meta-boxes-subscription.js` | 2.35 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/moment.js` | 22.1 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/moment.min.js` | 11.6 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/payment-method-restrictions.js` | 1.29 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/wcs-meta-boxes-order.js` | 502 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/payment-methods.js` | 355 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/single-product.js` | 429 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/view-subscription.js` | 1.38 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/wcs-cart.js` | 387 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/modal.js` | 1.1 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/wcs-upgrade.js` | 1.27 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/build/index.css` | 392 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/build/index.js` | 3.06 kB |

</details>



<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -3242,10 +3246,7 @@ private function is_platform_payment_method( bool $is_using_saved_payment_method
 		// Make sure the payment method being charged was created in the platform.
 		if (
 			! $is_using_saved_payment_method &&
-			$this->should_use_stripe_platform_on_checkout_page() &&
-			// This flag is useful to differentiate between PRB, blocks and shortcode checkout, since this endpoint is being used for all of them.
-			! empty( $_POST['wcpay-is-platform-payment-method'] ) && // phpcs:ignore WordPress.Security.NonceVerification
-			filter_var( $_POST['wcpay-is-platform-payment-method'], FILTER_VALIDATE_BOOLEAN ) // phpcs:ignore WordPress.Security.NonceVerification,WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+			$this->should_use_stripe_platform_on_checkout_page()
 		) {
```
Mayisha, line 3023: @ricardo we are going to support platform checkout on blocks checkout page also. Do you see any issues or regression if I remove this condition from the check? Asking for your thoughts as you had added this logic initially. :eyes:

fashuvo:
**Size Change:** +355 B (0%) 

**Total Size:** 1.7 MB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `release/woocommerce-payments/dist/platform-checkout.js` | 77.2 kB | +355 B (0%) |

<details><summary>ℹ️ <strong>View Unchanged</strong></summary>

| Filename | Size |
| :--- | :---: |
| `release/woocommerce-payments/assets/css/admin.css` | 913 B |
| `release/woocommerce-payments/assets/css/success.css` | 401 B |
| `release/woocommerce-payments/dist/blocks-checkout.css` | 1.77 kB |
| `release/woocommerce-payments/dist/blocks-checkout.js` | 34 kB |
| `release/woocommerce-payments/dist/checkout.css` | 1.85 kB |
| `release/woocommerce-payments/dist/checkout.js` | 26.9 kB |
| `release/woocommerce-payments/dist/index.css` | 35.8 kB |
| `release/woocommerce-payments/dist/index.js` | 640 kB |
| `release/woocommerce-payments/dist/multi-currency-analytics.js` | 8.41 kB |
| `release/woocommerce-payments/dist/multi-currency-switcher-block.js` | 79 kB |
| `release/woocommerce-payments/dist/multi-currency.css` | 14.7 kB |
| `release/woocommerce-payments/dist/multi-currency.js` | 117 kB |
| `release/woocommerce-payments/dist/onboarding.css` | 1.12 kB |
| `release/woocommerce-payments/dist/onboarding.js` | 38.1 kB |
| `release/woocommerce-payments/dist/order.js` | 299 B |
| `release/woocommerce-payments/dist/payment-gateways.css` | 1.2 kB |
| `release/woocommerce-payments/dist/payment-gateways.js` | 91 kB |
| `release/woocommerce-payments/dist/payment-request.js` | 23 kB |
| `release/woocommerce-payments/dist/platform-checkout.css` | 3.73 kB |
| `release/woocommerce-payments/dist/settings.css` | 47.2 kB |
| `release/woocommerce-payments/dist/settings.js` | 157 kB |
| `release/woocommerce-payments/dist/subscription-edit-page.js` | 3.66 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-modal.css` | 7.18 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-modal.js` | 16.7 kB |
| `release/woocommerce-payments/dist/subscription-product-onboarding-toast.js` | 428 B |
| `release/woocommerce-payments/dist/subscriptions-empty-state.css` | 298 B |
| `release/woocommerce-payments/dist/subscriptions-empty-state.js` | 49 kB |
| `release/woocommerce-payments/dist/tos.css` | 236 B |
| `release/woocommerce-payments/dist/tos.js` | 31.7 kB |
| `release/woocommerce-payments/dist/upe_checkout.css` | 1.85 kB |
| `release/woocommerce-payments/dist/upe_checkout.js` | 38 kB |
| `release/woocommerce-payments/dist/upe-blocks-checkout.css` | 1.77 kB |
| `release/woocommerce-payments/dist/upe-blocks-checkout.js` | 44.8 kB |
| `release/woocommerce-payments/includes/subscriptions/assets/css/plugin-page.css` | 633 B |
| `release/woocommerce-payments/includes/subscriptions/assets/js/plugin-page.js` | 355 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-admin-ui/src/css/jetpack-icon.css` | 224 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-assets/build/i18n-loader.js` | 2.43 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-assets/src/js/i18n-loader.js` | 1.01 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/babel.config.js` | 160 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.css` | 2.27 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.js` | 13.3 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-identity-crisis/build/index.rtl.css` | 2.27 kB |
| `release/woocommerce-payments/vendor/automattic/jetpack-tracking/src/js/tracks-ajax.js` | 789 B |
| `release/woocommerce-payments/vendor/automattic/jetpack-tracking/src/js/tracks-callables.js` | 925 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/about.css` | 1.2 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/admin-order-statuses.css` | 403 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/admin.css` | 3.43 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/checkout.css` | 299 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/modal.css` | 742 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/view-subscription.css` | 572 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/css/wcs-upgrade.css` | 411 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/admin-pointers.js` | 544 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/admin.js` | 8.97 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/jstz.js` | 6.8 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/jstz.min.js` | 3.83 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/meta-boxes-coupon.js` | 544 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/meta-boxes-subscription.js` | 2.35 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/moment.js` | 22.1 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/moment.min.js` | 11.6 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/payment-method-restrictions.js` | 1.29 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/admin/wcs-meta-boxes-order.js` | 502 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/payment-methods.js` | 355 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/single-product.js` | 429 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/view-subscription.js` | 1.38 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/frontend/wcs-cart.js` | 387 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/modal.js` | 1.1 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/assets/js/wcs-upgrade.js` | 1.27 kB |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/build/index.css` | 392 B |
| `release/woocommerce-payments/vendor/woocommerce/subscriptions-core/build/index.js` | 3.06 kB |

</details>



<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

includes/class-wc-payment-gateway-wcpay.php
ricardo: @Mayisha From the top of my head, I remember having added this check for the reasons described in that comment 🙂. Does the PRB (payment request button) still work from the checkout page after the change?

Also, does the pay for order page still work?

Mayisha: Thanks @ricardo for mentioning the concerned features. I was confused what's the PRB is and I am not familiar with the pay for order page :sweat_smile: 
I will find out if they are working or not :eyes:

ricardo: @Mayisha props for asking and making sure 🙌.

I tested the PRB here and it didn't work. It throws an error:

> Error: No such PaymentMethod: 'pm_1Lmgk7Qq9iIkB8Vrfac04v6W'

**Steps to reproduce:**
- Enable express (Apple Pay / Google Pay) checkouts from **Payments > Settings**.
- Click to "Customize" and enable it on Checkout, Product page and Cart.
- With a public HTTPS URL (using Jurassic Tube or Ngrok), visit the product, cart or checkout page.
- Try to pay using Google Pay.

So I guess that check is still important.

In regards to the pay for order page: To get to the pay for order page, you need to create a new order from wp-admin, add a product to it, click "Create" and visit the "Customer payment page". This flow works for me.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -3242,10 +3246,7 @@ private function is_platform_payment_method( bool $is_using_saved_payment_method
 		// Make sure the payment method being charged was created in the platform.
 		if (
 			! $is_using_saved_payment_method &&
-			$this->should_use_stripe_platform_on_checkout_page() &&
-			// This flag is useful to differentiate between PRB, blocks and shortcode checkout, since this endpoint is being used for all of them.
-			! empty( $_POST['wcpay-is-platform-payment-method'] ) && // phpcs:ignore WordPress.Security.NonceVerification
-			filter_var( $_POST['wcpay-is-platform-payment-method'], FILTER_VALIDATE_BOOLEAN ) // phpcs:ignore WordPress.Security.NonceVerification,WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+			$this->should_use_stripe_platform_on_checkout_page()
```
ricardo, line 3022: Just adding this as a more formal way of review for visibility.

I didn't test or look into the other parts of the code but tested the PRB functionality (Google Pay) as requested, and it [throws an error](https://github.com/Automattic/woocommerce-payments/pull/4638#discussion_r981473141).

ricardo (changes requested):

includes/class-wc-payment-gateway-wcpay.php
ricardo: **Steps to reproduce:**
- Enable express (Apple Pay / Google Pay) checkouts from **Payments > Settings**.
- Click to "Customize" and enable it on Checkout, Product page and Cart.
- With a public HTTPS URL (using Jurassic Tube or Ngrok), visit the product, cart or checkout page.
- Try to pay using Google Pay.

Mayisha: Thanks @ricardo for testing this and the detailed steps. I will bring this back and add a new condition to support checkout blocks here. :nerd_face:

client/checkout/blocks/index.js
```diff
@@ -34,6 +34,16 @@ const api = new WCPayAPI(
...
 
+// Create an API object, which will be used for payment request buttons.
+const paymentRequestApi = new WCPayAPI(
+	{
+		publishableKey: getConfig( 'publishableKey' ),
+		accountId: getConfig( 'accountId' ),
+		locale: getConfig( 'locale' ),
+	},
+	request
+);
```
FangedParakeet: Is there any reason why we couldn't reuse the other `WCPayAPI` instance right above this? I think `WCPayAPI` is basically just a wrapper to make requests to the WCPay back-end with Stripe's elements, so I'm not sure that there's a benefit in persisting separate instances. 

The only difference between these two instances I can see here is that this one is missing the `forceNetworkSavedCards` option, whereas all the other constructor arguments are the same. Do we need to ensure that `stripePlatform` is `null` in this second instance? Let me know if I'm missing anything!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -3239,13 +3241,15 @@ public function append_stripelink_button( $field, $key, $args, $value ) {
 	 * @return boolean True if it is a platform payment method.
 	 */
 	private function is_platform_payment_method( bool $is_using_saved_payment_method ) {
+		// Return flase for express checkout method.
```
ricardo: 🤏 Nit: small typo in `false`.

client/checkout/platform-checkout/index.js
```diff
@@ -18,24 +18,45 @@ const renderSaveUserSection = () => {
...
+	);
+
+	if ( blocksCheckout.length ) {
+		const orderSummary = document.getElementsByClassName(
+			'wp-block-woocommerce-checkout-order-summary-block'
+		)?.[ 0 ];
+		const orderSummaryParent = orderSummary?.parentNode;
+
+		if ( orderSummary && orderSummaryParent ) {
+			orderSummaryParent.appendChild( checkoutPageSaveUserContainer );
```
FangedParakeet: This probably isn't my decision to make--and if it's a design decision that _has_ been made I'll gladly yield--but this seemed like an overly conspicuous position on the page to place this button. 

On the regular, shortcode checkout page, your attention is drawn vertically downward through the shipping form, then across to the payment methods, and immediately vertically downwards again to the "Place order" button. The save to WooPay button is a perfectly placed pit-stop on this path at the penultimate position in this passage.

Arrows detailing user flow on shortcode checkout page
_WooPay in the right place at the right time_

In the blocks checkout, our attention is drawn vertically downwards throughout the entire flow, with no other change in direction. However, here the save to WooPay button is found up near the top of the page underneath the order summary.

Arrows detailing userflow on blocks checkout
_Customers act like they forgot about WooPay_

In the blocks checkout, wouldn't it make more sense for this radio button to be similarly placed in the section beneath the payment methods and above the "Place order" button as well? Already knowing that this button was to be found somewhere on the page, even I had to search a little bit to discover it.

Additionally, note that when WooPay is enabled, on the shortcode checkout page the "Save payment information to my account for future purchases" is removed, leaving the save to WooPay button as the only option available to save one's current payment method. 

No Save payment information to my account for future purchases button on shortcode checkout
_No "Save payment information to my account for future purchases" button on shortcode checkout_

In the current blocks checkout, this button is still present when WooPay is enabled, which also confused me while searching for the actual Save to WooPay button. I doubt it would be clear to many users that there would be any benefit from checking both of these options and I'm sure most would ignore the WooPay button and select the other Save button that appeared more naturally and conveniently on their user-path instead.

Save payment information to my account for future purchases button present with blocks
_What are you doing here, step-button?_

I think, at the least, we should try to remove this button from the Blocks checkout, if possible, to remove this confusion and maintain consistency with the regular checkout. 

That being said, things are usually more easier said than done, so if there is some other reason why this current layout was designed or something fundamental in the nature of Blocks that appropriately blocks the implementation of the suggestions I've made above, please let me know!

client/components/platform-checkout/hooks/use-selected-payment-method.js
```diff
@@ -3,8 +3,13 @@
...
-	document.querySelector( '#payment_method_woocommerce_payments' )?.checked;
+const getWCPayRadioButtonStatus = ( isBlocksCheckout ) =>
+	isBlocksCheckout
+		? document.querySelector(
+				'#radio-control-wc-payment-method-options-woocommerce_payments'
+		  )?.checked
+		: document.querySelector( '#payment_method_woocommerce_payments' )
+				?.checked;
 
 const getNewPaymentTokenRadioButtonStatus = () =>
```
FangedParakeet: By the way, when I am using the Blocks checkout this always returns `true`, regardless of whether I have selected a saved payment method or not. Is this expected? I feel like this hook is not working exactly as expected...

...or hang on, does [the logic here](https://github.com/Automattic/woocommerce-payments/blob/task/4614-new-platform-checkout-user-from-checkout-blocks/client/components/platform-checkout/save-user/checkout-page-save-user.js#L125-L128) still work, because `isWCPayChosen` will be `false` when a saved payment method is selected in the Blocks checkout? I guess that works, but that seems like slightly dangerous unexpected, yet intended behaviour. 

Perhaps leave a comment explaining that this hook cannot be trusted in the Blocks checkout, at least, just so that no future generations accidentally tumble through this trapdoor? (I think it might still be better to fix this function if possible though... 😬)

ricardo (commented):
Great work on this 👏

I did some overall look at the code and it looks good to me, although I didn't fully test and I am not sure if storing the sign-up consent option into the session has any other implications in the checkout.

**Here's the result of my tests:**

- ✅ Payment Request Button checkout (Google Pay):
  - From checkout block.
  - From cart block.
  - From checkout shortcode.
  - From cart shortcode.
  - From product page.
- ✅ WooPay authentication + Checkout:
  - Checkout block.
  - Checkout shortcode.
- ❌ WooPay sign-up:
  - ❌ Checkout block.
  - ✅ Checkout shortcode.

To me, the sign-up didn't work from the checkout block. The customer is not created in Stripe either.

**Other minor issues:**

1. There's a small inconsistency with the shortcode checkout where the "Save payment information to my account for future purchases" checkbox is still visible. I think we intentionally replaced that with "Save my information for faster checkouts" checkbox in the shortcode checkout.

2. To me, the phone icon looks this way:
  - phone-icon
3. IMO, people will easily miss the WooPay sign-up checkbox at the top of the page, especially if it's hidden by default until you select the **WooCommerce Payments > CC** payment method.

https://user-images.githubusercontent.com/5509901/194209780-53ce58c5-3cc4-4853-b3ac-4f9ce21645e9.mov

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1342,8 +1341,11 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 			} else {
 				$save_user_in_platform_checkout = false;
 
+				$session_data = WC()->session->get( 'platform-checkout-user-data' );
 				// phpcs:ignore WordPress.Security.NonceVerification.Missing,WordPress.Security.ValidatedSanitizedInput.MissingUnslash
-				if ( ! empty( $_POST['save_user_in_platform_checkout'] ) && filter_var( $_POST['save_user_in_platform_checkout'], FILTER_VALIDATE_BOOLEAN ) ) {
+				$should_save_platform_customer = ( ! empty( $_POST['save_user_in_platform_checkout'] ) && filter_var( $_POST['save_user_in_platform_checkout'], FILTER_VALIDATE_BOOLEAN ) ) || ( isset( $session_data['save_user_in_platform_checkout'] ) && filter_var( $session_data['save_user_in_platform_checkout'], FILTER_VALIDATE_BOOLEAN ) );
+
+				if ( $should_save_platform_customer ) {
```
FangedParakeet:
```suggestion
				if ( $payment_information->should_save_payment_method_to_platform() ) {
```
Can't we just rely on the value that you set above on the `Payment_Information` instance now? This looks like exactly the same check that you perform over [here](https://github.com/Automattic/woocommerce-payments/blob/task/4614-new-platform-checkout-user-from-checkout-blocks/includes/class-wc-payment-gateway-wcpay.php#L1045), to my untrained eyes.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet:
```suggestion
		// Return false for express checkout method.
```
🔍 (Sorry.)

includes/platform-checkout-user/class-platform-checkout-extension.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class that registers a blocks extension callback for platform checkout new user.
+ */
+class Platform_Checkout_Extension {
+	const EXTENSION_NAMESPACE           = 'platform-checkout';
+	const PLATMORM_CHECKOUT_SESSION_KEY = 'platform-checkout-user-data';
```
FangedParakeet:
```suggestion
	const PLATFORM_CHECKOUT_SESSION_KEY = 'platform-checkout-user-data';
```
🔎 (Sorry again.)

includes/platform-checkout-user/class-platform-checkout-extension.php
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Checks and stores the value of 'Remember your details' checkbox and phone number for platform checkout
+	 *
+	 * @param array $data Items to update in session data array.
+	 *
+	 * @return void
+	 */
+	public function store_user_consent_for_platform_checkout( array $data ) {
+		WC()->session->set( self::PLATMORM_CHECKOUT_SESSION_KEY, $data );
```
FangedParakeet:
```suggestion
		WC()->session->set( self::PLATFORM_CHECKOUT_SESSION_KEY, $data );
```
🔎 (Forgive me.)

includes/platform-checkout-user/class-platform-checkout-save-user.php
```diff
@@ -76,4 +83,17 @@ public function maybe_add_userdata_to_metadata( $metadata, $order ) {
...
+	 * Clears if platform checkout user data is set.
+	 *
+	 * @return void
+	 */
+	public function maybe_clear_session_key() {
+		$session_data = WC()->session->get( 'platform-checkout-user-data' );
+
+		if ( ! empty( $session_data ) ) {
+			WC()->session->set( 'platform-checkout-user-data', null );
+		}
```
FangedParakeet:
```suggestion
		WC()->session->__unset( Platform_Checkout_Extension::PLATMORM_CHECKOUT_SESSION_KEY );
```
Apparently this `__unset` function exists (can see it being used here and there in the codebase): could be used here? Up to you, as I'm not sure which is the better convention.

includes/platform-checkout-user/class-platform-checkout-save-user.php
```diff
@@ -76,4 +83,17 @@ public function maybe_add_userdata_to_metadata( $metadata, $order ) {
...
 		return $metadata;
 	}
+
+	/**
+	 * Clears if platform checkout user data is set.
+	 *
+	 * @return void
+	 */
+	public function maybe_clear_session_key() {
+		$session_data = WC()->session->get( 'platform-checkout-user-data' );
```
FangedParakeet:
```suggestion
		$session_data = WC()->session->get( Platform_Checkout_Extension::PLATMORM_CHECKOUT_SESSION_KEY );
```
Why not use the lovely constant that you created over here?

includes/platform-checkout-user/class-platform-checkout-save-user.php
```diff
@@ -57,11 +58,17 @@ public function register_checkout_page_scripts() {
 	 * @return array
 	 */
 	public function maybe_add_userdata_to_metadata( $metadata, $order ) {
+		$session_data = WC()->session->get( 'platform-checkout-user-data' );
```
FangedParakeet:
```suggestion
		$session_data = WC()->session->get( Platform_Checkout_Extension::PLATMORM_CHECKOUT_SESSION_KEY );
```
Sorry, you're going to see me suggesting this a lot now that I've noticed it...but you could use your constant here!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1043,8 +1040,10 @@ protected function prepare_payment_information( $order ) {
 			$payment_information->must_save_payment_method_to_store();
 		}
 
+		$session_data = WC()->session->get( 'platform-checkout-user-data' );
```
FangedParakeet:
```suggestion
		$session_data = WC()->session->get( Platform_Checkout_Extension::PLATMORM_CHECKOUT_SESSION_KEY );
```
I can't get enough of this constant!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1342,8 +1341,11 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 			} else {
 				$save_user_in_platform_checkout = false;
 
+				$session_data = WC()->session->get( 'platform-checkout-user-data' );
```
FangedParakeet:
```suggestion
				$session_data = WC()->session->get( Platform_Checkout_Extension::PLATMORM_CHECKOUT_SESSION_KEY );
```
I'm not sure that you need this at all, due to another comment, but I Can't Believe It's Not Constant®.

includes/platform-checkout-user/class-platform-checkout-save-user.php
```diff
@@ -57,11 +58,17 @@ public function register_checkout_page_scripts() {
...
 		// If we should be creating a platform checkout customer, add the necessary customer data to the metadata.
-		$should_create_platform_customer = isset( $_POST['save_user_in_platform_checkout'] ) && filter_var( wp_unslash( $_POST['save_user_in_platform_checkout'] ), FILTER_VALIDATE_BOOLEAN ); // phpcs:ignore WordPress.Security.NonceVerification
-		if ( $should_create_platform_customer && ! empty( $_POST['platform_checkout_user_phone_field']['full'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-				$platform_checkout_phone = wc_clean( wp_unslash( $_POST['platform_checkout_user_phone_field']['full'] ) ); // phpcs:ignore WordPress.Security.NonceVerification
+		$should_create_platform_customer = ( isset( $_POST['save_user_in_platform_checkout'] ) && filter_var( wp_unslash( $_POST['save_user_in_platform_checkout'] ), FILTER_VALIDATE_BOOLEAN ) ) || ( isset( $session_data['save_user_in_platform_checkout'] ) && filter_var( $session_data['save_user_in_platform_checkout'], FILTER_VALIDATE_BOOLEAN ) ); // phpcs:ignore WordPress.Security.NonceVerification
```
FangedParakeet: Just a thought, but perhaps you could define this as a function somewhere--so we don't have to write this whole thing out each time--like perhaps in `Platform_Checkout_Utilities` maybe? Just seeing it used around quite a lot and it's a bit of a mouthful. 😬

FangedParakeet (changes requested):
I really like the approach here and the way you manipulated the WC session data to your advantage, so great job! 🙌 

That being said, I tested the regular sign-up flow through the Blocks checkout and for some reason, it did not create a new WooPay user for me. 😞 The shortcode checkout worked as expected and I could register a new WooPay user from there, so at least that's still all fine!

I've left quite a few comments, but almost all of them are just coding-style issues: I don't believe any are particularly breaking and sadly none will explain why I was unable to get the registration flow to work as expected from the Blocks checkout. 

Please let me know your thoughts on my feedback and give me a shout when you'd like me to give this another proper look. Cheers!

client/checkout/blocks/index.js
Mayisha: Here in the `blocks` when calling the `registerExpressPaymentMethod`, it was considering the `forceNetworkSavedCards`. But in [the classic checkout](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-request/index.js#L33-L38) we don't have the `forceNetworkSavedCards` when registering the payment request button. 
Because of this difference when [this](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L687) has `true` value the express checkout/ payment request button was using different stripe elements (plus payment was failing on the blocks checkout page)

I am still not 200% sure if this change will have other regression scenarios as I not not familiar with the WCPay plugin much. Can you think of any? :grimacing:

FangedParakeet: > I am still not 200% sure if this change will have other regression scenarios as I not not familiar with the WCPay plugin much. 

Cool, I have now learned a lifetime's worth of information about the `forceNetworkSavedCards` flag in a couple separate conversations today. 😅 

In conclusion, I've learned that the following can generally be assumed to be true:
 - When WooPay is disabled, [generally this value will be `false`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L954).
  - The only valid reason for this value to be `true` [is to support WooPay](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L687).
   - There is no reason for a Stripe element used to serve a payment request button to ever need to be compatible with WooPay. Moreover, since in this case the payment element will confirm the intent, if we set this value to `true` to create a platform payment method, the request will always fail, because the intent will be on the merchant account. (With WooPay, the intent is confirmed on the WCPay server, where we can clone the created platform payment method to the merchant connect account before confirming the intent.)

Basically, what I'm saying is that I'm pretty confident that this change should be fine and furthermore I can't see a better way to keep both elements functioning as expected without doing what you've done. Sorry for digging up doubt, only to answer my own question. 😹

client/components/platform-checkout/hooks/use-selected-payment-method.js
Mayisha: I totally missed this and the `Save payment information to my account for future purchases` as I was always testing with a guest user. Fixed this in the latest commits. :grimacing:

client/checkout/platform-checkout/index.js
Mayisha: Confirmed with design and I will move the checkbox below the `Payment options` sections :sunglasses:

includes/class-wc-payment-gateway-wcpay.php
Mayisha: Your untrained eyes are very powerful :open_mouth: 
I have updated this to use the new utils method you suggested in [another comment](https://github.com/Automattic/woocommerce-payments/pull/4638#discussion_r988538727).

Mayisha:
I have addressed the other feedback but the user creation works on my localhost for both blocks and shortcode checkout. Can I have another look here? :thinking:  :grimacing: 
@FangedParakeet @ricardo

client/checkout/blocks/index.js
ricardo: @Mayisha Sorry to reopen this, but I agree with @FangedParakeet's first comment. There's already a parameter inside the `WCPayAPI` to force the Stripe API to be initialized with the merchant account. I think this should work:

1. Passing a parameter `forceAccountRequest` (default to `false`) from `loadStripe` to `getStripe` [here](https://github.com/Automattic/woocommerce-payments/blob/b3d5cf9855cecb72711c6013cadd474c198f5ba8/client/checkout/api/index.js#L101).
2. Passing that parameter as `true` to `api.loadStripe()` [here](https://github.com/Automattic/woocommerce-payments/blob/b6bf28c8a85c0c5c781f2280aaa9cffa1d628dda/client/payment-request/blocks/index.js#L14) and [here](https://github.com/Automattic/woocommerce-payments/blob/b6bf28c8a85c0c5c781f2280aaa9cffa1d628dda/client/payment-request/blocks/index.js#L27).

This should force the blocks payment request button to not use the platform account even though the `forceNetworkSavedCards` is `true`.

ricardo:
@Mayisha I was testing WCPay while checked out in this branch and found the following issue on the **Appearance > Customize** page:

```
Fatal error: Uncaught Error: Call to a member function is_empty() on null
in woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php on line 778
```

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -355,6 +362,9 @@ public function __construct(
 			'link'          => 'link_payments',
 		];
 
+		// Platform checkout utilities.
+		$this->platform_checkout_util = new Platform_Checkout_Utilities();
```
FangedParakeet, line 355: Is it maybe worth injecting this class instead of instantiating it here? This way you'd just include it as a parameter to the `WC_Payment_Gateway_WCPay` constructor and the `Platform_Checkout_Utilities` instance would be created once inside `WC_Payments` and injected [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L352). The benefit of this approach is that it would now enable us to mock out all of the `Platform_Checkout_Utilities` functions if desired. However, if this isn't particularly important, perhaps this could be ignored.

Though, that being said, if mocking out the functions inside `Platform_Checkout_Utilities` isn't terribly important, I feel like that class could just contain static functions and then we wouldn't need to instantiate anything. Let me know if you feel strongly one way or the other!

FangedParakeet (commented):
Thanks for all the feedback here! I think the code is looking great now and I really have practically nothing to moan about anymore. 

> Adding an existing platform checkout user in the email field will hide the Remember your details section.

That being said, this did not ring true for me. 

> Check and ensure that the experience on the blocks checkout page exactly same as the existing feature in the short-code checkout page.

Moreover I could not load the spinner nor the OTP prompt while entering the email of a registered WooPay user, while using the Blocks checkout. Is this part supposed to be working or should this be solved by another issue? 🤔 

Let me know, if I'm missing anything! My local user creation flow is a little broken at the moment, but I don't think your code is at fault for that. I'll try to resolve that myself first before commenting on that portion of the testing here.

client/checkout/blocks/index.js
Mayisha: This worked. Thanks for the suggestion :tada:

includes/class-wc-payment-gateway-wcpay.php
Mayisha: This sounds like a good suggestion and I had already made the change just to realize that this will need me to fix 200+ tests. I am planning to do this refactor in a separate PR later to keep this PR decluttered. :grimacing:

FangedParakeet: Ooh, gotcha. That's definitely not worth the effort in this PR. 

> The benefit of this approach is that it would now enable us to mock out all of the Platform_Checkout_Utilities functions if desired. 

Dependency injection is helpful, because it gives us more control over how our dependencies are attached, specifically allowing us to mock them out relatively easily in our tests...which is only really important, if it's beneficial to mock out particular functions performed by the dependency in question. 

In this class, we only rely on the `Platform_Checkout_Utilities::should_save_platform_customer()` function in this dependency. This function basically performs a check based on the contents of the `$_POST` global and the `WC()` session data. It's up to you whether it makes more sense to simply manually fill these variables or mock out the function entirely: if you decide on the former, then there's no real benefit to injecting the dependency (for now) and--considering the exttrenuous effort--it's probably not worth the trouble.

Aaaanyways, as far as I can see right now, there aren't any present tests that test the differing outcomes of this function at all. I'll leave it up to you to decide whether there should be and in turn whether this entire refactor might bear any fruit or whether this whole discussion can be safely forgotten. 😅

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -8,10 +8,38 @@ import request from '../utils/request';
...
+
+		const checkoutBlock = document.querySelector(
+			'[data-block-name="woocommerce/checkout"]'
+		);
+		const observer = new MutationObserver( () => {
+			if ( document.querySelector( selector ) ) {
+				resolve( document.querySelector( selector ) );
+				observer.disconnect();
+			}
+		} );
```
FangedParakeet:
```suggestion
		const observer = new MutationObserver( ( mutationList, observer ) => {
			if ( document.querySelector( selector ) ) {
				resolve( document.querySelector( selector ) );
				observer.disconnect();
			}
		} );
```
Do we not need to include the `observer` as a parameter in the `MutationObserver` constructor, like in [these examples](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver)? It seems unusual to have a variable depend on itself while it's being defined. Wouldn't `observer` in the call to `observer.disconnect()` be `undefined` at the point that that callback is defined? Though, JavaScript does work in mysterious ways, so maybe I'm mistaken. 😅

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2631,7 +2638,7 @@ public function create_and_confirm_setup_intent() {
 		$should_save_in_platform_account = false;
 
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing
-		if ( ! empty( $_POST['save_payment_method_in_platform_account'] ) && filter_var( wp_unslash( $_POST['save_payment_method_in_platform_account'] ), FILTER_VALIDATE_BOOLEAN ) ) {
+		if ( $this->platform_checkout_util->should_save_platform_customer() ) {
```
FangedParakeet: I think we can get rid of the `phpcs` comment here. Also, it's not massively important, but we could get rid of this `if` condition and the `$should_save_in_platform_account` variable, by simply passing the result of `$this->platform_checkout_util->should_save_platform_customer()` to the `create_and_confirm_setup_intent()` function call below, no?

Like this below...

```php
return $this->payments_api_client->create_and_confirm_setup_intent(
	$payment_information->get_payment_method(),
	$customer_id,
	$this->platform_checkout_util->should_save_platform_customer(),
	$this->is_platform_payment_method( $payment_information->is_using_saved_payment_method() )
);
```

...Sorry! I just always find `if ( $condition ) { $var = true; }` blocks to be a little redundant. 😸

FangedParakeet (commented):
I've left a few minor comments, but the code looks good to me!

I was able to run through the regular WooPay checkout flows as expected through both the shortcode and blocks checkouts. In the blocks checkout, the email input seems to be doing exactly what it's supposed to doing.

I was still unable to confirm that the new user flow works successfully. My local environment is unreliable, but I also built a release of these changes, uploaded the package to a fresh ephemeral site, and tried to run through the new user flow, testing against the sandbox. The checkout flow was working perfectly in both blocks and the shortcode checkout. But alas, when I tried to create a new user, returning to the checkout, my freshly registered email address was not identified as a valid WooPay account. 

Let me know if there's anything different you think I can attempt here! I'll gladly step aside and grant my blessings, if anyone can give me good reason to, because the code looks pretty good to me and I frankly can't quite comprehend why I was still unable to create a new user on the sandbox. 😖

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -14,27 +14,95 @@ import AdditionalInformation from './additional-information';
...
-	} = useSelectedPaymentMethod();
+	const { isWCPayChosen, isNewPaymentTokenChosen } = useSelectedPaymentMethod(
+		isBlocksCheckout
+	);
+
+	const getPhoneFieldValue = () => {
+		let phoneFieldValue = '';
+		if ( isBlocksCheckout ) {
+			phoneFieldValue =
+				document.getElementById( 'phone' )?.value ||
```
ricardo, line 38: Because there's a field with ID `phone` in the blocks checkout, the phone icon doesn't render correctly for me:

SVG icon with mask doesn't render correctly because of duplicate ID

That's because the icon relies on a `#phone` mask, so changing the [mask ID](https://github.com/Automattic/woocommerce-payments/blob/c1162c5ad1ec20e7b7bed0d1d17b3530c700480b/client/components/platform-checkout/icons/phone.js#L16) to something like `phone-icon` fixes it.

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -14,27 +14,95 @@ import AdditionalInformation from './additional-information';
...
+			'button.wc-block-components-checkout-place-order-button'
 		);
 
 		if ( ! formSubmitButton ) {
 			return;
 		}
 
+		// send data to extension endpoint when place order button is clicked.
+		const sendRequestToExtension = () => {
+			extensionCartUpdate( {
```
ricardo, line 64: Just to confirm, does the checkout request also get triggered from the submit button `click` event?

There's no way to guarantee this is called and processed before the checkout request, right?

ricardo (changes requested):
Tests well. I was able to authenticate and sign-up from the blocks checkout. Sometimes the Stripe webhooks are taking a few minutes to arrive, causing delay in the account creation, but other than that, it appears to work as expected.

The express checkouts and classic checkout still work as expected.

Code LGTM. Just left 1 minor suggestion and question, and agreed with Samir's latest comments.

client/components/platform-checkout/save-user/checkout-page-save-user.js
Mayisha: Fixed it [b71a186](https://github.com/Automattic/woocommerce-payments/pull/4638/commits/b71a1865aa4d1d2db4a3dece958cf7097c6f192d)

includes/class-wc-payment-gateway-wcpay.php
Mayisha: Agreed and fixed [b71a186](https://github.com/Automattic/woocommerce-payments/pull/4638/commits/b71a1865aa4d1d2db4a3dece958cf7097c6f192d) :smile:

client/checkout/platform-checkout/email-input-iframe.js
Mayisha: Yup JavaScript was doing something mysterious there. :laughing:  
Decided to use the callback params to avoid confusion [b71a186](https://github.com/Automattic/woocommerce-payments/pull/4638/commits/b71a1865aa4d1d2db4a3dece958cf7097c6f192d)

client/components/platform-checkout/save-user/checkout-page-save-user.js
Mayisha: @bborman22 figured out that the order of the `extensionCartUpdate` and the main checkout request is the cause of why a new user is not created on Blocks checkout sometimes (the issue you and @FangedParakeet both faced before). 
To ensure that the extension data is always set first now the request is sent on the checkbox select/unselect event [22cd5bf](https://github.com/Automattic/woocommerce-payments/pull/4638/commits/22cd5bf8ecb7af7e1a238323b64be3e5994e12e3) :disappointed_relieved:

bborman22 (approved):
I was able to test this much more successfully now! I also did notice as @ricardo mentioned that the webhooks sometimes took quite a while, but after giving it a lot of time I found all my test submissions worked and created accounts. I also tried setting the checkbox and clicking place order as quickly as possible and never seemed to encounter an issue, so I think the order of the requests is good now.

I didn't do as deep a dive into the code as the other reviews, so it might be worth getting a second approval from either @FangedParakeet or @ricardo, but I did see that all of the requested changes were done and were fixed for me.

Approving this from my end though I primarily just did a functional test!

Mayisha:
@FangedParakeet @ricardo let me know if you have any other concerns. If not, I will merge this one tomorrow :eyes:

FangedParakeet (approved):
I still wasn't able to create a new user via the Blocks checkout on my ephemeral site...but I wasn't able to create one via the shortcode checkout either, so presumably this issue must be somewhere on my accursed side.

The code looks great to me though and all my feedback has been resolved, so I don't have any issues there nor any reason to slam the brakes on my behalf. Go on in peace. 🙌

client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -14,22 +16,89 @@ import AdditionalInformation from './additional-information';
...
+		if ( isSaveDetailsChecked ) {
+			setPhoneNumber( null );
+			if ( isBlocksCheckout ) {
+				sendExtensionData( true );
+			}
+		} else {
+			setPhoneNumber( getPhoneFieldValue() );
+		}
+		setIsSaveDetailsChecked( ( v ) => ! v );
+	};
```
ricardo: I think this would be slightly easier to parse if we checked for the current value of the checkbox instead of the old value of `isSaveDetailsChecked`.

```javascript
const handleCheckboxClick = ( ev ) => {
	const isChecked = ev.target.checked;
	if ( isChecked ) {
		setPhoneNumber( getPhoneFieldValue() );
	} else {
		setPhoneNumber( null );
		if ( isBlocksCheckout ) {
			sendExtensionData( true );
		}
	}
	setIsSaveDetailsChecked( isChecked );
};
```

Feel free to ignore this comment 🙂.

ricardo (approved):
LGTM. I was able to create the user from the blocks checkout and classic checkout.

PR #4640, opened 2022-08-25
Fixed issue with PRB and products without price

zmaglica:
Fixes #[4579](https://github.com/Automattic/woocommerce-payments/issues/4579)

When the product doesn't have price, the exception will be thrown because the function that is used for formatting the price with currency expects the number as a parameter inside request payment button handler.

This PR will prevent that by validating a price before the function that is used for formatting is called. If the product doesn't have a valid price, the function will exit without product data.

#### Testing instructions

1. Create a variable product or subscription and leave all prices empty
2. Publish product
3. Visit product and make sure that no errors are shown

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -508,7 +508,7 @@ class WC_Payments_Payment_Request_Button_Handler {
 		if ( $this->is_product() ) {
 			$is_product_supported = $this->is_product_supported();
 
-			if ( $is_product_supported ) {
+			if ( ! $is_product_supported ) {
 				Logger::log( 'Product page has unsupported product type ( Payment Request button disabled )' );
 				return false;
 			}
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -264,7 +264,7 @@ class WC_Payments_Payment_Request_Button_Handler {
 
 		$data          = [];
 		$items         = [];
-		$product_price = $this->get_product_price( $product );
+		$product_price = (float) $this->get_product_price( $product );
 
 		$items[] = [
 			'label'  => $product->get_name(),
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -517,7 +517,11 @@ class WC_Payments_Payment_Request_Button_Handler {
 		}
 
 		// Make sure to prevent displaying products when price is not set.
-		$product_price = $this->get_product_price( $this->get_product() );
+		$product = $this->get_product();
+		if ( ! is_a( $product, 'WC_Product' ) ) {
+			return false;
+		}
+		$product_price = $this->get_product_price( $product );
 
 		if ( ! is_numeric( $product_price ) ) {
 			return false;
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -516,6 +516,13 @@ class WC_Payments_Payment_Request_Button_Handler {
 			return false;
 		}
 
+		// Make sure to prevent displaying products when price is not set.
+		$product_price = $this->get_product_price( $this->get_product() );
+
+		if ( ! is_numeric( $product_price ) ) {
+			return false;
+		}
+
 		return true;
 	}
 
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -505,25 +505,25 @@ class WC_Payments_Payment_Request_Button_Handler {
 		}
 
 		// Product page, but has unsupported product type.
-		if ( $this->is_product() && ! $this->is_product_supported() ) {
-			Logger::log( 'Product page has unsupported product type ( Payment Request button disabled )' );
-			return false;
+		if ( $this->is_product() ) {
+			$is_product_supported = $this->is_product_supported();
+
+			if ( $is_product_supported ) {
+				Logger::log( 'Product page has unsupported product type ( Payment Request button disabled )' );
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -516,6 +516,13 @@ public function should_show_payment_request_button() {
 			return false;
 		}
 
+		// Make sure to prevent displaying products when price is not set.
+		$product_price = $this->get_product_price( $this->get_product() );
```
vbelolapotkov: It's not safe to call this method here. It will cause troubles on cart and checkout pages. I see there is even better method above on line 508. 

`if ( $this->is_product() && ! $this->is_product_supported() )`

So if we move the check inside is_product_supported that will make even more sense

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -505,9 +505,20 @@ public function should_show_payment_request_button() {
 		}
 
 		// Product page, but has unsupported product type.
-		if ( $this->is_product() && ! $this->is_product_supported() ) {
-			Logger::log( 'Product page has unsupported product type ( Payment Request button disabled )' );
-			return false;
+		if ( $this->is_product() ) {
+			$is_product_supported = $this->is_product_supported();
+
+			if ( $is_product_supported ) {
```
a-danae: Shouldn't we check if the product is not (!) supported here?

zmaglica: My bad. Fixed.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -264,7 +264,7 @@ public function get_product_data() {
 
 		$data          = [];
 		$items         = [];
-		$product_price = $this->get_product_price( $product );
+		$product_price = (float) $this->get_product_price( $product );
```
a-danae, line 267: I'm not sure how this casting would affect the use of `$product_price` later on. To reduce impact, why not checking whether the price is acceptable when deciding whether it's a supported product without altering this variable?

zmaglica: Hmm that might work, but how to define is price _acceptable_ ?

a-danae (commented):
@zmaglica I'm getting a fatal error on page load on the single product page when opening a simple subscription with no subscription price. Can you replicate it? This is the error I'm getting.

```
Fatal error: Uncaught TypeError: Unsupported operand types: string + string in /srv/users/userea9e5d5a/apps/userea9e5d5a/public/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-payment-request-button-handler.php:224
```

includes/class-wc-payments-payment-request-button-handler.php
a-danae: I can't think of a scenario where we'd like to display the buttons when the price is empty, so I'd say we wouldn't support products with no price. We could check that if so. I'm referring to products that have an empty `''` price, not a price of `0`. 

Glad to learn if I'm missing a scenario where we do want to support this, though.

a-danae (commented):
The error from the related issue could happen from places where we're doing operations with the product price and this is set to a string, which happens if you leave the price field empty. These are the locations that could be prone to this error:

1. `WC_Payments_Payment_Request_Button_Handler::ajax_get_selected_product_data()`, [#](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L1032)
All good here. This is triggered when updating the product quantity on the Single product page. This line can't be reached when a product doesn't have a price with the current implementation because the PRB doesn't show up.

2. `WC_Payments_Payment_Request_Button_Handler::get_product_price()`, [#](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L224)
This line is being reached and causing a fatal error. This happens when a subscription has no price, as described [here](https://github.com/Automattic/woocommerce-payments/pull/4640#pullrequestreview-1086001996).

3. `WC_Payments_Payment_Request_Button_Handler::get_product_data()`, [#](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L274), [#](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L301)
This is the source of the reported issue. The float casting introduced in the PR solves the error for these lines.

Casting the product price to a float in different locations where the price is used does work, but it seems like a band-aid to me, and it's likely that we'll continue doing operations with the product price without casting it in the future when someone that's not aware of the current issue works with this part of the code. Maybe evaluating the product as "unsupported" would prevent reaching the three scenarios mentioned above that could trigger this error. Another option could be casting the price to a float right in `get_product_price()`. But this seems to have a higher impact as I'm not sure how this would affect other areas of the code, and maybe it won't be needed if the PRB aren't loaded in the first place.

ricardo:
> Fatal error: Uncaught TypeError: Unsupported operand types: string + string in /srv/users/userea9e5d5a/apps/userea9e5d5a/public/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-payment-request-button-handler.php:224

I get the same error if I try a subscription with a sign-up fee, but without price. It seems like `WC_Subscriptions_Product::get_sign_up_fee( $product )` from line [224](https://github.com/Automattic/woocommerce-payments/blob/312c6aac7e512366fe1bfb4ef74649738e5faa7b/includes/class-wc-payments-payment-request-button-handler.php#L224) returns a string as well. But this is not the real issue.

If I remove the sign-up fee, it gets converted to `0`. The error becomes `Unsupported operand types: string + int`. So the real problem is really the first operand which is an empty string `''` in this case.

From some testing I did, the following works in PHP 7, but not in PHP 8: `'' + '0'`. There must be at least a number inside the string for the operation to work in PHP 8.

PR #4648, opened 2022-08-26
If COT enabled, use COT query to gather tracking data

### LINKED ISSUE #4148: Refactor the direct DB query which counts the number of multi-currency orders to use `wc_get_orders()`/`WC_Order_Query`
### Description

This issue is to look at the direct WP post query in [`Tracking::get_mc_order_count()`](https://github.com/Automattic/woocommerce-payments/blob/dd9d319b661427a1d22a5aeab945c23dd429f317/includes/multi-currency/Tracking.php#L117-L144) and replace it with `wc_get_orders()`/`WC_Order_Query`.

### Dev notes

There's a wiki on how to use `wc_get_orders()` with custom meta [here](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#adding-custom-parameter-support), however I don't know how that DB query will translate to using something like `wc_get_orders()`. 

### Additional context

jessy-p:
Partially fixes #4148 

The code in `get_mc_order_count` use DB query on posts and post_meta tables to gather data needed for [usage tracking](https://woocommerce.com/usage-tracking/). The code change is to check that if WC has COT and COT enabled, use a query on the order tables, or else fall back to the older query on `posts` and `post_meta`.

#### Testing instructions

1. Have orders placed with multiple currencies
2. Add the following snippet, which logs the tracking data to your `debug.log`
```
add_filter( 'woocommerce_tracker_send_override', '__return_false' );
add_filter( 'woocommerce_tracker_last_send_time', '__return_false' );

add_filter(
	'woocommerce_tracker_data',
	function( $data ) {
		error_log( __FILE__ . ':' . __LINE__ . ":\n" . '$data: ' . print_r( $data, true ) );
		return $data;
	},
	999
);
```
3. Trigger the tracking cron `wp cron event run woocommerce_tracker_send_event` ( This can also be done using WP-Crontrol )
4. Check debug.log for the output
5. The tracking data logged should be the same across various settings, such as
     -  Install older version of WC which is not COT aware
     -  COT enabled with WooCommerce Orders tables as the data store
     -  On WooCommerce ▸ Settings ▸ Advanced ▸ Custom Data Stores, set the data store to use the WordPress posts table as datastore
 6. Place an order and check that the tracking data is updated in all instances.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : **QA Testing Not Applicable**
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable. **Not Applicable**
- [x] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable. **Not Applicable**


jessepearson:
It looks like the steps to use custom order tables are here: https://developer.woocommerce.com/2022/05/16/call-for-early-testing-custom-order-table-migrations/ . Is this what you used to test with @jessy-p ?

deepakpathania (approved):
Tested in all 3 scenarios:
- Non COT WC
- COT WC with orders table
- COT WC with posts table

Tests well and changes LGTM.

jessy-p:
> It looks like the steps to use custom order tables are here: https://developer.woocommerce.com/2022/05/16/call-for-early-testing-custom-order-table-migrations/ . Is this what you used to test with @jessy-p ?

@jessepearson  I used this [COT-Upgrade-Recipe-Book#how-to-enable-custom-order-tables ](https://github.com/Automattic/woocommerce/wiki/COT-Upgrade-Recipe-Book#how-to-enable-custom-order-tables).

PR #4649, opened 2022-08-26
Payment Requests: Fix variations without price, sign-up fee taxes, and variable subscriptions with sign-up taxes

### LINKED ISSUE #4579: Product Variations without Prices Result in Critical Error on Product Page in PHP version 8.0

RadoslavGeorgiev:
Fixes #4579

This PR was meant to fix the simple fatal error, caused by incompatible addition operands in PHP 8.0, but ended up fixing a few Payment Request-related issues, as described below.

I kept those changes in a single PR, because some of them are dependent on each-other, and cannot be handled simultaneously in separate PRs, requiring a series instead. That would not be ideal for a patch release, which is warranted by the original issue.

Since it's a multi-purpose PR, testing instructions will be mixed with the steps below.

## Temporary local setup modifications

**Step 1:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/44b6b3ed578c0a76cbd78fad5676e19637a8daed

This PR includes a modified Dockerfile, which uses WordPress's PHP 8.0 version instead of 7.3. This file will be reset before merging the PR (it's in the checklist below), but in the meantime you need to use this before/after testing the PR:

```sh
❯ docker-compose down
❯ docker-compose up -d --build wordpress
```

**Step 2:** 🔍 Confirm the initial error

1. Checkout the commit from step 1, and run the commands.
2. Enable Payment Request buttons in Payments &raquo; Settings.
3. Create a variable product without any variations. I recommend using meaningful names here, because those same products will be needed later. In this case "**Variable product without variations**" makes sense.
4. ➡️ Open the product page, and you will see `Fatal error: Uncaught TypeError: Unsupported operand types: float - string`
5. Create another variable product, and add at least a couple variations to it, but leave all of them without prices. Recommended name: "**Variable product with variations**".
6. ➡️ Open the page of the new product, and observe the same error.

## Fatal errors for variations without price

This line used to cause a `TypeError: Unsupported operand types: float - string` error.
https://github.com/Automattic/woocommerce-payments/blob/44b6b3ed578c0a76cbd78fad5676e19637a8daed/includes/class-wc-payments-payment-request-button-handler.php#L274

The reason for this is that if a price is missing, an empty string would be returned, which is not a legal operation since PHP 8.0.

**Step 3:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/081101c9f2899c5d34839c112ad13758598855b6 Add a new `Invalid_Price_Exception` class, which will be used to react accordingly.

**Step 4:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/7a29d25530cef3ef2b6648c1f07a7251cfe3b86a Use the new exception:

1. Refactor `get_product_price()` a bit:
    - Switch from `$product_price` to `$base_price`, and do an `if-else` statement to indicate that the price can be loaded with/without tax.
    - Instead of immediately adding the sign-up fee for subscriptions to the base price, store it in an additional variable.
    - If any of the two variable is non-numeric, throw an exception.
    - Return the sum of both variables if everything looks good.
2. Within `get_product_data`, bail as early as possible if the exception is thrown, and log it.
3. Do the same within `ajax_get_selected_product_data`.

**Step 5:** 🔍 Confirm that the initial error has been fixed.

1. Check out https://github.com/Automattic/woocommerce-payments/pull/4649/commits/7a29d25530cef3ef2b6648c1f07a7251cfe3b86a (the commit from step 4)
2. Check both _Variable product without variations_ and _Variable product with variations_.
3. ➡️ Both products should load correctly while displaying `This product is currently out of stock and unavailable.`, which is the user-friendly way of saying that there are no prices entered.
4. Add prices to some variations. Only _Variable product with variations_ should have any variations, so use that one.
5. ➡️ Check the product page and confirm that you can only select the variations with a price.

**Step 6:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/27d732c3569e20e0f964172a20f2cd4823c25747 Toggle the visibility of the payment request button if there was an exception in `ajax_get_selected_product_data`. At this stage the step is necessary to test the last usage of the exception, but it happens to also cover an edge case.

**Step 7:** 🔍 Test that the toggling works

1. At this point you need to open a tunnel (`npm run tube:start`), and switch to using Chrome (or Safari) if you aren't already, in order to actually see the payment request button.
2. Check out the commit from step 6, https://github.com/Automattic/woocommerce-payments/pull/4649/commits/27d732c3569e20e0f964172a20f2cd4823c25747
3. Make sure that all variations of _Variable product with variations_ have prices.
4. Open the product page in a different tab on the tunneled domain. You should see the payment request button. If you don't see it, it means that your browser/local set-up is not configured.
3. In the tab, where you are editing the product, remove the price of one of the variations.
6. **Without refreshing the product page**, choose the variation without price. The button should disappear
7. Select one of the variations with a price.
8. ➡️  The button should re-appear.

## Incorrect tax calculations upon sign-up fees

While testing how `get_product_price` works with subscription sign-up fees, I realized that we're using a weird shortcut to calculate tax (product price with tax minus product price without tax) to circumvent most of the code, which is used in the cart and checkout, but this leads to incorrect calculations in some scenarios.

**Step 8:** 🔍 Confirm that taxes are calculated incorrectly

1. Create a new subscription product, and enter a sign-up fee. In my case I named it **"Subscription with a sign-up fee"**, and set both the subscription price, as well as the sign-up fee to $12.00.
2. Navigate to the product through the Jurassic Tube address.
3. ➡️ Open the payment request popup and observe the following:
Screenshot 2022-08-26 at 21 14 00

As you can see in the screenshot, we end up with a negative tax entry, and incorrect total amount. Within `get_product_price()` and `get_product_data()`, we end up with the following formulas:

```js
price = ( product_price + signup_fee );
tax  = price_including_tax - price = price_including_tax - product_price - signup_fee;
total = product_price + signup_fee + tax;
```

When you replace `tax` within the total equation, you get this:

```js
total = product_price + signup_fee + price_including_tax - product_price - signup_fee = price_including_tax;
```

In the example above, that's $12 + 15% (my generic tax) of $12 = $12.00 + $1.80 = $13.80 (as visible in the screenshot). Let's fix that.

**Step 9:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/6167d588249ae2c9fbe1bda08c466650259eb38d Calculate the tax correctly

Within this commit I decided to mimic the functionality `WC_Cart_Totals::get_item_tax_rates()` and:
1. Load the product's tax class.
2. Load all rates, associated with the given tax class.
3. Use `WC_Tax::calc_tax()` to calculate the tax for both the subscription price and signup fee combined.

❓ One thing I'm not sure about is he fact that `calc_tax()` returns a numeric one-dimensional array. In my tests, it always had a single entry, which I used in my calculations, but when would there be more entries, and how should we treat them?

**Step 10:** 🔍 Check if the taxes are correct now:

1. Check out https://github.com/Automattic/woocommerce-payments/pull/4649/commits/6167d588249ae2c9fbe1bda08c466650259eb38d
2. Refresh _Subscription with a sign-up fee_, and open the payment request dialog.
3. ➡️ Observe the expected tax and total ($24.00 * 15% = $3.60, and $27.60 in total):
Screenshot 2022-08-26 at 21 26 57

## Missing sign-up fees for variable subscriptions

After ensuring that taxes for subscriptions with sign-up fees work well, I decided to try the same with variable subscriptions, to make sure it works as well, but it didn't.

**Step 12:** #️⃣ https://github.com/Automattic/woocommerce-payments/pull/4649/commits/bbfd212d0e2e685e9370fe73fc11c066e5a881cb Include sign-up fees for subscription variations in the calculation

The only thing needed to be done here was to include `variable-subscription` in `get_product_price`'s sign-up fee-supported types.

> ⚠️ It turned out that this already works, because instead of using the `wcpay_get_selected_product_data` AJAX call, our JavaScript uses an `wcpay_add_to_cart` call, which actually modifies the cart object in order to get the correct price in this scenario. I didn't see this as first, because I was only checking the `wcpay_get_selected_data` JSON to fix the inconsistency, and only while testing without this fix, I found out that everything already works. I'm still keeping this change in the PR though, as it yields more consistent values.

**Step 13:** 🔍 Testing variable subscriptions with sign-ups

1. Create a new product with type "Variable Subscription".
2. Create three subscriptions:
    1. One without a price or a sign-up fee
    2. One with a price, but without a sign-up fee
    3. One with both a price and a sign-up fee.
3. Check the product:
    1. The first variation should not appear at all.
    2. When the second one is selected, there should be no sign-up fee.
    3. When the third variation is selected, the payment request popup should contain the correct price, signup fee, and tax.
4. ➡️ This would have already worked correctly, but if you inspect the AJAX call, you will see consistent values.

In the future we might consider removing the `add_to_cart` call.

**Currently the PR is missing additional unit tests, to be added on Monday**

-------------------

**Post merge**

- [x] Remove the configuration for PHP 8.0


client/payment-request/index.js
```diff
@@ -388,8 +388,10 @@ jQuery( ( $ ) => {
 						).then( () => {
 							wcpayPaymentRequest.unblockPaymentRequestButton();
 						} );
-					}
-				);
+					} )
+					.catch( () => {
+						wcpayPaymentRequest.hide();
```
RadoslavGeorgiev, line 393: If the server throws an exception, or returns an error after a caught one (as will be the case here), this means that we don't have enough data to display the overlay, which is why it's getting hidden here.

We need to decide if this is good UX, considering that the standard payment flow is still available, or design feedback would be required.

FWIW, this is a real edge case, for which I cannot find steps to reproduce. As indicated in the testing instructions of the PR, we need to intermittently change product settings, while the page has already been loaded, in order to even test it.

docker/wordpress_xdebug/Dockerfile
```diff
@@ -1,9 +1,9 @@
...
-	&& echo 'xdebug.remote_enable=1' >> $PHP_INI_DIR/php.ini \
-	&& echo 'xdebug.remote_port=9000' >> $PHP_INI_DIR/php.ini \
-	&& echo 'xdebug.remote_host=host.docker.internal' >> $PHP_INI_DIR/php.ini \
-	&& echo 'xdebug.remote_autostart=0' >> $PHP_INI_DIR/php.ini \
+FROM wordpress:php8.0
+RUN pecl install xdebug-3.1.5 \
+	&& echo 'xdebug.mode=debug' >> $PHP_INI_DIR/php.ini \
+	&& echo 'xdebug.client_port=9000' >> $PHP_INI_DIR/php.ini \
+	&& echo 'xdebug.client_host=host.docker.internal' >> $PHP_INI_DIR/php.ini \
+	&& echo 'xdebug.start_with_request=trigger' >> $PHP_INI_DIR/php.ini \
```
RadoslavGeorgiev: These changes will be undone before merging the PR.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -262,22 +274,38 @@ public function get_product_data() {
...
+			// No cart item, `woocommerce_cart_totals_get_item_tax_rates` can't be applied here.
+
+			// Normally there should be a single tax, but `calc_tax` returns an array, let's use it.
+			foreach ( WC_Tax::calc_tax( $price, $rates, false ) as $tax ) {
+				$total_tax += $tax;
+
+				$items[] = [
+					'label'   => __( 'Tax', 'woocommerce-payments' ),
+					'amount'  => WC_Payments_Utils::prepare_amount( $tax, $currency ),
+					'pending' => ( 0 === $tax ? true : false ),
```
zmaglica: Nitpick. This ternary operator can be replaced with` 0 === $tax`

zmaglica (approved):

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/4649/commits/21a37b3caf5c43e5feb66496892ab589170eb407

client/payment-request/index.js
deepakpathania: I feel the error in this case from the server should have been a `400` instead of a `500` and should have been handled accordingly here. The request is for an intermittent state which will be handled on refresh by that option not showing anymore so a `500` doesn't seem apt but not a blocker.

client/payment-request/index.js
```diff
@@ -436,11 +438,23 @@ jQuery( ( $ ) => {
 			} );
 		},
 
+		getElements: () => {
+			return $(
+				'#wcpay-payment-request-wrapper,#wcpay-payment-request-button-separator'
```
deepakpathania, line 443: Should we store this as constants describing the element? Would be more readable then.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -210,21 +210,33 @@ public function get_button_height() {
...
+			'subscription',
+			'subscription_variation',
+		];
+		if ( in_array( $product->get_type(), $subscription_types, true ) && class_exists( 'WC_Subscriptions_Product' ) ) {
+			// When there is no sign-up fee, `get_sign_up_fee` falls back to an int 0.
+			$sign_up_fee = WC_Subscriptions_Product::get_sign_up_fee( $product );
 		}
 
-		return $product_price;
+		if ( ! is_numeric( $base_price ) || ! is_numeric( $sign_up_fee ) ) {
```
deepakpathania, line 235: Do we need a check for the numeric `$sign_up_fee` here? We set the initial value to `0` and it also falls back to an int `0` so will there be a case where this is non numeric?

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -210,21 +210,33 @@ public function get_button_height() {
...
+			throw new Invalid_Price_Exception( __( 'Express checkout does not support products without prices! ', 'woocommerce-payments' ) );
```
deepakpathania: Do we want to log additional properties here by mentioning the product details? Would be easier to debug in case this starts showing up in logs.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -262,21 +274,29 @@ public function get_product_data() {
 			}
 		}
 
-		$data          = [];
-		$items         = [];
-		$product_price = $this->get_product_price( $product );
+		try {
+			$price = $this->get_product_price( $product );
```
deepakpathania, line 284: Any specific reason to rename it from `product_price`?

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1041,11 +1062,14 @@ public function ajax_get_selected_product_data() {
...
+			$total_tax = 0;
+			foreach ( $this->get_taxes( $product, $price ) as $tax ) {
+				$total_tax += $tax;
+
 				$items[] = [
 					'label'   => __( 'Tax', 'woocommerce-payments' ),
-					'amount'  => 0,
-					'pending' => true,
+					'amount'  => WC_Payments_Utils::prepare_amount( $tax, $currency ),
+					'pending' => 0 === $tax,
```
deepakpathania, line 1078: What does this mean? In case `tax` is non 0, we'll set this to `false`, how is that used? Asking since it was hard coded to `true` earlier.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1630,4 +1657,26 @@ public function get_login_confirmation_settings() {
...
+			return [];
+		}
+
+		// Follows the way `WC_Cart_Totals::get_item_tax_rates()` works.
+		$tax_class = $product->get_tax_class();
+		$rates     = WC_Tax::get_rates( $tax_class );
+		// No cart item, `woocommerce_cart_totals_get_item_tax_rates` can't be applied here.
+
+		// Normally there should be a single tax, but `calc_tax` returns an array, let's use it.
+		return WC_Tax::calc_tax( $price, $rates, false );
```
deepakpathania, line 1686: `calc_tax` says it returns `Array of rates + prices after tax.` Do we know in what cases we have multiple entries here?

includes/exceptions/class-invalid-price-exception.php
```diff
@@ -0,0 +1,17 @@
...
+namespace WCPay\Exceptions;
+
+use Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Exception for throwing errors in REST API controllers (e.g. issues with missing parameters in requests).
+ */
+class Invalid_Price_Exception extends Exception {}
```
deepakpathania, line 17: If we extend `Rest_Exception`, we could log additional properties as well using this.

deepakpathania (changes requested):
Tested all flows and they seem to work as intended now, left some code related comments.

client/payment-request/index.js
RadoslavGeorgiev: That was the idea behind introducing `getElements` as a new method, which is the only place those IDs are included. Based on most jQuery code I've seen, this is a fairly normal practice, rather than declaring the selector as a constant and then querying it. i can go either way here, but this change seems unnecessary.

includes/exceptions/class-invalid-price-exception.php
RadoslavGeorgiev: This is not a REST exception, and the one we have in the client (`Rest_Request_Exception`) does not include any additional properties. Considering this is client-side, do you think adding the product ID to the error message could be sufficient?

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: Not really :\ That's why I mentioned this in the PR description:

> ❓ One thing I'm not sure about is he fact that calc_tax() returns a numeric one-dimensional array. In my tests, it always had a single entry, which I used in my calculations, but when would there be more entries, and how should we treat them?

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: Here is the definition of `pending` for reference: https://stripe.com/docs/js/appendix/payment_item_object#payment_item_object-pending

In this case it was always set to `true`, because it was later calculated during payment. In this case, if we have a number, it is not likely to change, which is why `pending` is `false` if the amount is different from zero.

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: It's just a personal nitpick of mine. If there is a single price in the method, I prefer not to be overly-specific and choose the shorter version. If it was a single instance, and I didn't have changes within the method, I would have left it alone, but decided to use the opportunity :)

includes/exceptions/class-invalid-price-exception.php
deepakpathania: Ah got it confused with the server one. Yeah, adding product id would suffice.

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: My thinking here was to check both, just to be on the safe side. I saw that upon an empty string for the signup fee, `WC_Subscriptions_Product::get_sign_up_fee()` would happen to cast the empty string to false, and return the default value, which is `int(0)`, but that felt more like a lucky and unreliable side effect, rather than intentional casting.

includes/class-wc-payments-payment-request-button-handler.php
deepakpathania: Since we are looping over the array returned by this to add to the product price, might be a good idea to look into this a little, or if we can't reproduce a scenario then maybe picking the first entry from this array to always use that.

includes/exceptions/class-invalid-price-exception.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4649/commits/b973eb30a00bd00bd8e99443ba8e7b2deec98c18

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: Done here: https://github.com/Automattic/woocommerce-payments/pull/4649/commits/b973eb30a00bd00bd8e99443ba8e7b2deec98c18 (just the product ID, but that should be more than clear enough). I considered adding the product name, but I think variations don't have proper names.

includes/class-wc-payments-payment-request-button-handler.php
RadoslavGeorgiev: This was interesting to test :)

So, after digging as low as the SQL in [`WC_Tax::get_matched_tax_rates()`](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-tax.php#L350), I saw that the logic is:

1. Load the tax class of the product
2. Load a location, for this I think it is based either on the store address, or the logged in customer (not too relevant here)
3. Load all taxes, which apply to the given tax class and location
4. Pick one of those taxes per priority (this is a manually entered value, see the screenshot below)
5. Add all of the loaded taxes in ascending priority.
    - If the tax is normal, just calculate it based on product prices.
    - If the tax is compound, calculate it based on product prices, combined with all previous taxes.

In the end, the returned array can contain more than one line, but all of those should be added together in the end.

image

deepakpathania (approved):
Tests well and changes LGTM now. ✅

RadoslavGeorgiev:
I used `git rebase trunk -i` locally to drop the commit, which upgraded Docker to PHP 8.0, so checking that as done.

PR #4657, opened 2022-08-30
E2E subscription tests - Fix timeouts & optimize flow

### LINKED ISSUE #4404: Flaky E2E test - Execution context was destroyed, most likely because of a navigation
### Describe the bug

The following test fails randomly with the error ` Execution context was destroyed, most likely because of a navigation.`. Usually the error goes away when re-run.

* `subscriptions/shopper/shopper-myaccount-renew-subscription.spec.js`
* `subscriptions/shopper/shopper-subscriptions-change-payment-method.spec.js`

Reference Jobs:

* https://github.com/Automattic/woocommerce-payments/runs/7188256139?check_suite_focus=true
* https://github.com/Automattic/woocommerce-payments/runs/7168612101?check_suite_focus=true

### Actual behavior
The test fails randomly.

### Expected behavior
The test should pass consistently.

### LINKED ISSUE #4408: Flaky E2E Tests - Timeouts
### Describe the bug

The following tests fails randomly with timeouts / missing selectors.

* `subscriptions/merchant/merchant-subscriptions-purchase-no-sign-up.spec.js`
* `subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js`
* `wcpay/shopper/shopper-checkout-failures.spec.js`

Reference jobs: 

* https://github.com/Automattic/woocommerce-payments/runs/7185165856?check_suite_focus=true
* https://github.com/Automattic/woocommerce-payments/runs/7232937901?check_suite_focus=true
* https://github.com/Automattic/woocommerce-payments/runs/7174713951?check_suite_focus=true

### Actual behavior
The test fails randomly.

### Expected behavior
The test should pass consistently.

achyuthajoy:
Fixes #4404 
Fixes #4408

This PR mostly addresses flakiness present on subscription flows. Here are the major changes on this PR:

* Updates `woocommerce/e2e-utils` package from `0.1.6` to `0.2.0`.
* Adds new user account with the email `e2e-wcpay-subscriptions-customer@woocommerce.com` for running subscription tests. This prevents deleting the default customer account and causing problems with other test flows.
* Removes creating subscription products during tests. Instead imports the subscription products during E2E setup using an XML file.
* Updates flow to explicitly logout customer account after tests to prevent 500 internal server error while trying to login as admin and to prevent cascade failures.
* Deletes the flow `subscriptions/shopper/shopper-subscriptions-change-payment-method.spec.js` since the tests are already present on `subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js`.
* Deletes the flow `subscriptions/shopper/shopper-subscriptions-purchase-subscription-product.spec.js` since it's covered in other tests
* Renamed `subscriptions/shopper/shopper-subscriptions-free-trial.spec.js` to `subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js`
* Moved the following tests from `subscriptions - merchant` to `subscriptions - shopper`
  * `subscriptions/merchant/merchant-subscriptions-purchase-no-sign-up.spec.js` to `subscriptions/shopper/shopper-subscriptions-purchase-no-sign-up-fee.spec.js` 
  * `subscriptions/merchant/merchant-subscriptions-purchase-sign-up-fee.spec.js` to `subscriptions/shopper/shopper-subscriptions-purchase-sign-up-fee.spec.js`
* Updates to subscription flows to fix test flakiness 

#### Testing instructions

* On GitHub actions, trigger a new E2E test against this branch and confirm that tests against WC `latest` & `beta` are passing. Tests against other versions are failing due to other reasons which will addressed in a different PR
* On local, setup a new E2E environment with local server instance & run `E2E_GROUP='subscriptions' E2E_BRANCH='merchant' npm run test:e2e` & `E2E_GROUP='subscriptions' E2E_BRANCH='shopper' npm run test:e2e` and confirm that the tests are passing.

-------------------

**Post merge**


tests/e2e/specs/subscriptions/merchant/merchant-subscriptions-renew-subscription.spec.js
```diff
@@ -69,16 +54,17 @@ describeif( RUN_SUBSCRIPTIONS_TESTS ).skip(
...
 			);
 
-			await merchant.login();
+			await shopper.logout();
 		} );
 
 		afterAll( async () => {
 			// Delete the user created with the subscription
 			await withRestApi.deleteCustomerByEmail( customerBilling.email );
-			await merchant.logout();
```
deepakpathania, line 78: This should still be there instead of adding it to individual tests IMO, otherwise we rely on the order of execution of tests for this file so that last test passes for clean slate.

tests/e2e/specs/subscriptions/shopper/shopper-myaccount-renew-subscription.spec.js
```diff
@@ -91,6 +77,8 @@ describeif( RUN_SUBSCRIPTIONS_TESTS )(
 			await page.waitForNavigation( { waitUntil: 'networkidle0' } );
 			await shopper.placeOrder();
 			await expect( page ).toMatch( 'Order received' );
+
+			await shopper.logout();
```
deepakpathania: This can also be moved to an `afterEach` block for making it consistent if we add more tests to this file later?

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js
```diff
@@ -67,39 +54,40 @@ describeif( RUN_SUBSCRIPTIONS_TESTS )(
 			);
 
 			await shopperWCP.goToSubscriptions();
+			await page.click(
+				'.woocommerce-orders-table__cell-subscription-id > a',
```
deepakpathania: Can we move the selectors to constants?

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js
```diff
@@ -143,6 +127,8 @@ describeif( RUN_SUBSCRIPTIONS_TESTS )(
 			await expect( page ).toMatchElement( '.trial_end_date', {
 				text: renewalDateFormatted,
 			} );
+
+			await merchant.logout();
```
deepakpathania: I feel these logouts shouldn't be test specific, they can be set at block level, just review them once to see if we can move them around?

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-multiple-subscriptions.spec.js
```diff
@@ -74,20 +65,32 @@ describeif( RUN_SUBSCRIPTIONS_TESTS )(
 			await page.waitForNavigation( { waitUntil: 'networkidle0' } );
 
 			// Ensure 'Subscription totals' section lists the subscription products with the correct price.
-			const subTotalsRows = await page.$$( 'tr.order_item' );
+			const subTotalsRows = await page.$$(
+				'.shop_table order_details tbody tr.order_item'
```
deepakpathania: Moving all selectors to constants also makes sense to me to have descriptive names for them.

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-no-sign-up-fee.spec.js
```diff
@@ -2,50 +2,36 @@
...
-			// Create subscription product without signup fee
-			await merchantWCP.createSubscriptionProduct(
-				productName,
-				'month',
-				false
-			);
-
-			await merchant.logout();
+			// Delete the user, if present
+			await withRestApi.deleteCustomerByEmail( customerBilling.email );
```
deepakpathania, line 36: Why do we need to do this both before and after all?

deepakpathania (changes requested):
The run was fine for the `latest` and `beta` blocks, left some comments which might be repetitive.

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-no-sign-up-fee.spec.js
achyuthajoy: Originally the user was deleted after each test suite run. However I added it to delete the user before the test to make sure that the user account isn't present causing test failure. Though this is a rare case.

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js
achyuthajoy: I understand that the logout function should be on the teardown block to prevent cascade failures, just in case the test stopped for some reason.

However, what would be the approach where where we use both merchant & shopper logins inside the same test. We can't use `shopper.logout()` or in teardown if the shopper isn't logged in because it would throw an error since selector would be missing.

For `merchant.logout`, it doesn't fail but it won't logout a shopper account.

So I believe we should try and complete the flow. If logged in, we should logout as well. And this would be helpful to troubleshoot tests as well. 

Teardowns are useful when we do repetitive stuff, like a test where only merchant flows are in place.

Let me know your thoughts @deepakpathania

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js
achyuthajoy: All the tests we have are written in this way. What you've suggested is a good approach by defining selectors early, but I think it would be best if we update it on all flows, not just subscriptions.

tests/e2e/specs/subscriptions/shopper/shopper-myaccount-renew-subscription.spec.js
achyuthajoy: This can be done in this particular test. We only check shopper flows. But I believe we should explicitly logout and not do it inside teardown unless it's repetitive.

See https://github.com/Automattic/woocommerce-payments/pull/4657#discussion_r959863817

tests/e2e/specs/subscriptions/merchant/merchant-subscriptions-renew-subscription.spec.js
achyuthajoy: Would it be better to explicity logout rather that waiting for teardown to handle it? See https://github.com/Automattic/woocommerce-payments/pull/4657#discussion_r959863817

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-multiple-subscriptions.spec.js
achyuthajoy: See https://github.com/Automattic/woocommerce-payments/pull/4657#discussion_r959864703

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js
deepakpathania: I personally feel we should log in and logout in `beforeEach` and `afterEach` blocks for truly independent tests, I understand that's not feasible because some tests in a spec rely on the previous one but we should strive to reach as close to tests being truly isolated from each other as possible. The changes here deviate from that a little as you mentioned above so I feel we should reconsider this.

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js
deepakpathania: We can update it in places we are changing right now maybe? Otherwise updating it in all places at once becomes a chore so as a rule, we can just fix this in files we touch moving forward. WDYT?

achyuthajoy: Sure. That works

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-no-sign-up-fee.spec.js
deepakpathania: If we keep it in the `beforeAll` of all tests then logically we shouldn't need it in the `afterAll` as that becomes repetitive when the test suite runs, so we can keep it consistent imo.

achyuthajoy: Sure. Updated

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-multiple-subscriptions.spec.js
achyuthajoy: Added pre-defined selectors

tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js
achyuthajoy: Updated teardowns where possible. Let me know if more changes are required

tests/e2e/specs/subscriptions/shopper/shopper-myaccount-renew-subscription.spec.js
achyuthajoy: Updated

achyuthajoy:
Good for another review @deepakpathania

deepakpathania (approved):
Thanks for working on this, changes LGTM now. ✅

PR #4661, opened 2022-08-30
When COT enabled, use COT query for customer currencies query

### LINKED ISSUE #4148: Refactor the direct DB query which counts the number of multi-currency orders to use `wc_get_orders()`/`WC_Order_Query`

jessy-p:
Fixes #4148 

The code in `get_all_customer_currencies` uses DB query on `post_meta` tables to fetch all customer currencies used in orders. This function is used on the Analytics > Orders page. This will fail when COT is enabled and sync is off, as the currency data will be available only on the `orders` table. This code change fixes this so that when COT is enabled, we use an alternate query on the order tables.

#### Testing instructions

1. Ensure COT is enabled, WC Orders is set as the default table, sync is off
2. Add a new currency from **WooCommerce > Settings > Multicurrency** and place an order with that currency
3. Clear the  DB cache for `wcpay_multi_currency_customer_currencies` by deleting the entry with key `wcpay_multi_currency_customer_currencies` from options table. 
4. On **Analytics > Orders**, add the Advanced Filter "**Customer currency**". Check the currencies in the dropdown.
image
5. On `develop`, the new currency is not displayed in the dropdown when `orders` is the default table but will be displayed if `posts` is the default table or if sync is on.
6. On this branch, the currencies in the dropdown will be the same whether the posts or orders table is set as the authoritative table

-------------------

**Post merge** 


includes/multi-currency/MultiCurrency.php
```diff
@@ -1354,14 +1354,15 @@ public function get_all_customer_currencies() {
...
 				global $wpdb;
-
-				$currencies = $wpdb->get_col(
-					"SELECT
-						DISTINCT(meta_value)
-					FROM
-						{$wpdb->postmeta}
-					WHERE meta_key = '_order_currency'"
-				);
+				$query_currencies_posts  = "SELECT DISTINCT(meta_value) FROM {$wpdb->postmeta} WHERE meta_key = '_order_currency'";
```
jbordonado: Can you move these 2 declarations inside their respective branch ("if")? I think it's best to define variable only where they're being used

jbordonado (approved):
Apart from my previous comment, everything seems to work fine!

includes/multi-currency/MultiCurrency.php
jessy-p: On second thought, the idea of moving the queries as separate strings came from another class where the queries were really long and hard to read. It is not warranted here, so I've moved the queries back as strings within the `get_col` call. Hope that makes sense.

jbordonado (approved):

PR #4663, opened 2022-08-31
Add a fallback for retrieving legacy source objects

RadoslavGeorgiev:

> ⚠️ This PR will not work by itself, because a GET `sources/src_****` needs to be added to the server first. Currently the endpoint can be found in 2430-gh-Automattic/woocommerce-payments-server.

On WCCOM, and now some other sites, we have migrated from the Stripe gateway to WooCommerce. Those sites still have stored old-school `source` objects, which were never supported by WCPay.

We had an instance of such a case in May (p1651596922989849-slack-CGGCLBN58), and since then we have manually applied patches to WCCOM, but since they need to be applied manually, this is not sustainable. Let's fix that.

The issue appears whenever we try to retrieve a `src_***` object through the `payment_methods` API, and triggers an error like this:

> Error: A source must be attached to a customer to be used as a `payment_method`.

The only instance so far sees to be in the context of refunds, and the only property of the object we use is `type`, so I decided to compare the possible values of source objects with those of payment methods:

| [Source](https://stripe.com/docs/api/sources/object#source_object-type)              | [Payment Method](https://stripe.com/docs/api/payment_methods/object#payment_method_object-type)                                                                              |
|---------------------|---------------------------------------------------------------------------------------------|
| `ach_credit_transfer` | We have never supported it                                                                  |
| `ach_debit`           | `acss_debit`                                                                                  |
| `alipay`              | `alipay`                                                                                      |
| `bancontact`          | `bancontact`                                                                                  |
| `card`                | `card`                                                                                        |
| `card_present`        | `card_present`                                                                                |
| `eps`                 | `eps`                                                                                         |
| `giropay`             | `giropay`                                                                                     |
| `ideal`               | `ideal`                                                                                       |
| `multibanco`          | Not available in `PaymentIntents` or `PaymentMethods`                                       |
| `klarna`              | `klarna`                                                                                      |
| `p24`                 | `p24`                                                                                         |
| `sepa_debit`          | `sepa_debit`                                                                                  |
| `sofort`              | `sofort`                                                                                      |
| `three_d_secure`      | Not available anymore, but can probably be returned as `card`, as it only applied to cards. |
| `wechat`              | `wechat`                                                                                      |

Most of those types match, so I decided to go ahead, and switch to the `sources` API within `get_payment_method`, and to silently return the `source` object, which can not only be used, but should not cause any errors.

There are some exceptions:

1. `ach_credit_transfer`, which we never supported, so there should be no issues.
2. `multibanco`, which we did support in Stripe, but I do not know if it was ever used on WCCOM. In this PR I am throwing an exception in case it was encountered. If that happens, we can check if any changes are required.
3. `three_d_secure`, which is simply a wrap around `card`. I have also added a fallback for it.

#### Testing instructions

Not possible to test on a normal WCPay instance, because we cannot create a source in the first place. We'll need the patch to be applied to WCCOM, and tested through a refund, like the temporary fix: 13290-gh-Automattic/woocommerce.com.


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1556,11 +1557,34 @@ public function update_subscription_item( $wcpay_subscription_item_id, $data ) {
 	 * @throws API_Exception If payment method does not exist.
 	 */
 	public function get_payment_method( $payment_method_id ) {
-		return $this->request(
+		// A fallback for sites, which migrated from Stripe to WCPay.
```
vbelolapotkov: Let's expand a bit more on what's hapenning here, and why. E.g. "Sites migrated from Stripe to WC Payments may have old source objects in place of payment methods, so another API endpoints need to be used for fetching payment method details".

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1556,11 +1557,34 @@ public function update_subscription_item( $wcpay_subscription_item_id, $data ) {
...
+		$is_source = preg_match( '/^src_\w+$/i', $payment_method_id );
+		if ( $is_source ) {
+			$api = self::SOURCES_API;
+		} else {
+			$api = self::PAYMENT_METHODS_API;
+		}
```
vbelolapotkov: Ternary operator would be a better fit here IMO, it's short and clean.
```suggestion
		$api = $is_source ? self::SOURCES_API : self::PAYMENT_METHODS_API;
```

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1556,11 +1557,34 @@ public function update_subscription_item( $wcpay_subscription_item_id, $data ) {
...
+
+		$response = $this->request(
 			[],
-			self::PAYMENT_METHODS_API . '/' . $payment_method_id,
+			$api . '/' . $payment_method_id,
 			self::GET
 		);
+
+		if ( $is_source ) {
```
vbelolapotkov: Let's use early return to reduce nesting level.
```suggestion
		if ( ! $is_source ) {
		    return $response;
		}
```

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1556,11 +1557,34 @@ public function update_subscription_item( $wcpay_subscription_item_id, $data ) {
...
+			if ( 'three_d_secure' === $response['type'] ) {
+				// This was how 3DS was handled initially.
+				$response['type'] = 'card';
+			} elseif ( 'multibanco' === $response['type'] ) {
+				throw new API_Exception(
```
vbelolapotkov: How this exception affects flows/UI? 
Error message is also seem to be very technical, who will see the message? What's the workaround or further guidance?

vbelolapotkov (changes requested):
@RadoslavGeorgiev thanks for working on this. I only reviewed the code and left a few comments. 

Note, I didn't do any further analysis of how the change would impact other places where the method is invoked. Neither I tested it in any setup. Could you please ask somebody else from the team or maybe @dechov to continue review from here.

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: Done In https://github.com/Automattic/woocommerce-payments/pull/4663/commits/0ad203a85bf1c300d838519ab3eeec2d9e39dbd1 👍

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: Also done in https://github.com/Automattic/woocommerce-payments/pull/4663/commits/0ad203a85bf1c300d838519ab3eeec2d9e39dbd1.

FWIW my idea here was to use a full `if-else` block to attract attention, and focus on the point, but with the longer comment from the previous message, that's not needed already. Otherwise my thinking was kind of like this:

image

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4663/commits/0ad203a85bf1c300d838519ab3eeec2d9e39dbd1 👍

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: The whole PR was mostly a PoC and I did not consider where the exception could appear, thinking that it is an extreme edge-case.

AFAICT the initial error has only appeared while attempting a refund. I took a look around, and it can happen both in the edit order screen, where HEs would have contexts, but now I also see that it might happen in the front-end:

Screenshot 2022-09-02 at 10 50 46

I'll dig a bit further and post my finding here.

includes/wc-payment-api/class-wc-payments-api-client.php
vbelolapotkov: Your assumption was fair, but it also requires some good understanding of old Stripe API (sources). Many new folks, including myself could not know these details, so more verbose comment should give enough hints on what's going on there :) Thanks for fixing.

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: **The follow-up, as promised**

I checked where the API client's `get_payment_method` method is called, and I found the following instances:

- `WC_Payment_Gateway_WCPay_Subscriptions_Trait::get_specific_new_payment_method_title()` catches exceptions around `get_payment_method`, logs them, and returns the original title, without showing an error message, or preventing an action from happening.
- `WC_Payments_Token_Service::add_payment_method_to_user()` is only called when the buyer is present, and a new payment method is presented. A source cannot appear here.
- `WC_Payment_Gateway_WCPay::get_payment_method_type_for_order()` is only called during refunds to identify if the payment method type is `interac_present`, which is where we had the issue in the first place.

I'd focus on the last method from this list, `get_payment_method_type_for_order()`, which can return `"unknown"` if the payment method was not found. IMO handling the exception here, and returning the `unknown` string, should allow everything to function without confusing user-facing errors, with only an entry in the logs. I'll open a commit for that in a second.

RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4663/commits/cbb042bb6ecc416225f782fb4b1b1551136a9711 :)

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1554,11 +1555,35 @@ public function update_subscription_item( $wcpay_subscription_item_id, $data ) {
 	 * @throws API_Exception If payment method does not exist.
 	 */
 	public function get_payment_method( $payment_method_id ) {
```
zmaglica, line 1307: Can we maybe add some tests for this API requests?

zmaglica (commented):
Left one small comment. I would like to add some tests to cover new flows.

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: @zmaglica done in https://github.com/Automattic/woocommerce-payments/pull/4663/commits/8e01ddc407ddaeb4e6fbe61d80b3dcd7434d385c

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `4663` or branch name `fix/unattached-sources-error` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/unattached-sources-error) :rocket:

---

Build info: 

- Latest commit: 45a58b527c710a48ce0892d7ce237c00f065c3cc
- Build time: 2023-12-14 04:36:35 UTC

htdat:
FYI. `Sources` is going to be deprecated in 2024 https://stripe.com/docs/sources 
More details here pbUcTB-40l-p2

So we might close this PR rather than add a to-be-deprecated-soon endpoint?

naman03malhotra:
>FYI. Sources is going to be deprecated in 2024 https://stripe.com/docs/sources

Good catch @htdat, +1 on this

RadoslavGeorgiev:
Thanks for pointing this out, @htdat! Given this fact, I agree that maybe shipping this combo of PRs is not worth it. Additionally, it's been 1.5+ years since the last time this was reported to be an issue.

I'm going ahead and closing the topic.

PR #4670, opened 2022-08-31
Split UPE: Correct token behavior

### LINKED ISSUE #4530: Multiple UPE - SEPA tokens appear in list of credit card tokens
### Describe the bug

#4427 allows payment methods to be added via the Add Payment Method page. The payment methods that are currently available to be saved for future use are CC and SEPA. Prior to #4427, all stored tokens appeared under the single WC Payments gateway, as that was the only place they could be listed. Now that each UPE payment method is it's own gateway, saved SEPA payment methods should appear under the SEPA gateway rather than the CC gateway. 

### To Reproduce

1. `git checkout poc/upe-instances-multiplied`
2. Enable SEPA in the WC Payments settings
3. Go to **My Account > Payment Methods > Add Payment Method**
4. Add a SEPA payment method

### Screenshots
image

### Acceptance Criteria
- SEPA tokens should appear under the SEPA gateway on the Add Payment Method page and Checkout page
- Payments via saved SEPA methods should process without error
- changes made should be merged into `poc/upe-instances-multiplied`, not `develop`.

mdmoore:
Fixes #4530

This PR ensures that tokens use the correct gateway ID and are displayed under the correct gateway throughout WCPay. Some checks have also been added to ensure that tokens aren't duplicated per each gateway.

#### Testing instructions

* Enable SEPA in the UPE settings
* From My Account > Payment Methods > Add Payment Method, add a credit card and SEPA method
* Return to My Account > Payment Methods > Add Payment Method and refresh the page a few times to make sure tokens aren't being duplicated
* Check that SEPA tokens are appearing under SEPA Direct Debit and card tokens are appearing under Credit card / Debit card

-------------------

**Post merge**


harriswong:
👀

harriswong:
I get a processing error when I tried creating a SEPA direct debit
image

The request returned a `400` code. The request URL is `https://api.stripe.com/v1/setup_intents/xyzxyzxyzxyzxyzxyz/confirm`

The response body has this error:
```
 "error": {
    "code": "setup_intent_unexpected_state",
    "doc_url": "https://stripe.com/docs/error-codes/setup-intent-unexpected-state",
    "message": "You cannot update this SetupIntent because it has already succeeded.",
    "setup_intent": {
     ...
    }
}
```

includes/class-wc-payments-token-service.php
```diff
@@ -120,9 +122,7 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
 			$stored_tokens = [];
 
 			foreach ( $tokens as $token ) {
-				if ( WC_Payment_Gateway_WCPay::GATEWAY_ID === $token->get_gateway_id() ) {
-					$stored_tokens[ $token->get_token() ] = $token;
-				}
+				$stored_tokens[ $token->get_token() ] = $token;
```
tommyshellberg: If I'm reading this right, this line previously checked to make sure that we were only grabbing WCPay gateway tokens and now we're displaying all tokens - could WCPay potentially try to display tokens from other gateways? You could test this with saved tokens from the default Stripe extension or something. 🤔 
Perhaps we could check here if the gateway is within the array of WCPay gateways.

includes/class-wc-payments-token-service.php
```diff
@@ -169,14 +170,12 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
...
 	public function woocommerce_payment_token_deleted( $token_id, $token ) {
-		if ( WC_Payment_Gateway_WCPay::GATEWAY_ID === $token->get_gateway_id() ) {
-			try {
-				$this->payments_api_client->detach_payment_method( $token->get_token() );
-				// Clear cached payment methods.
-				$this->customer_service->clear_cached_payment_methods_for_user( $token->get_user_id() );
-			} catch ( Exception $e ) {
-				Logger::log( 'Error detaching payment method:' . $e->getMessage() );
-			}
+		try {
```
tommyshellberg: Same thing as above, this function will fire every time a payment token is deleted within WC. Will this function try to delete tokens from other gateways?

tests/unit/test-class-wc-payments-token-service.php
```diff
@@ -157,7 +157,7 @@ public function test_woocommerce_payment_token_deleted() {
 
 	public function test_woocommerce_payment_token_deleted_other_gateway() {
 		$this->mock_api_client
-			->expects( $this->never() )
+			->expects( $this->once() )
```
tommyshellberg: It sounds like this unit test was written to make sure that deleting a payment token from another gateway doesn't happen. Setting this to `once()` seems like it kinda breaks the initial intention of the test. 
I wonder if we can write perhaps a more specific test for split UPE gateways? 
For instance, if we delete a SEPA gateway we test that only a SEPA token type is deleted(and only once)? 🤔

includes/class-wc-payments-token-service.php
```diff
@@ -100,7 +100,9 @@ public function add_payment_method_to_user( $payment_method_id, $user ) {
 	 * @return array
 	 */
 	public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gateway_id ) {
-		if ( ( ! empty( $gateway_id ) && WC_Payment_Gateway_WCPay::GATEWAY_ID !== $gateway_id ) || ! is_user_logged_in() ) {
+		$gateway_ids = [ WC_Payment_Gateway_WCPay::GATEWAY_ID, WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA ];
```
tommyshellberg: Is there a utility function somewhere in WCPay to get the available gateway IDs so this doesn't have to be hard-coded? If so, we could delete this line and do the check in line 105 (this is pseudo-code):

`! in_array( $gateway_id, get_upe_gateway_ids(), true )`

tommyshellberg:
> The request returned a `400` code. 
> The response body has this error:
> 
> ```
>  "error": {
>     "code": "setup_intent_unexpected_state",
>     "doc_url": "https://stripe.com/docs/error-codes/setup-intent-unexpected-state",
>     "message": "You cannot update this SetupIntent because it has already succeeded.",
>     "setup_intent": {
>      ...
>     }
> }
> ```

I'm getting the same error after creating a _second_ SEPA saved token. The same setup_intent ID is being re-used over and over, even if you refresh the page. My guess is that the setup intent is cached, perhaps [as a result of this PR](https://github.com/Automattic/woocommerce-payments/pull/3782)?

Edit: Yeah, it looks like it's grabbing a setup_intent ID from the session:

Screen Shot 2022-09-02 at 15 23 52

Perhaps we are not [removing the UPE setup intent](https://github.com/Automattic/woocommerce-payments/blob/1af7f46206fb1771c6dde0825100605e906b75e4/includes/payment-methods/class-upe-payment-gateway.php#L1202) from the session after creating a payment method.

Hope that helps!

mdmoore:
@harriswong @tommyshellberg Nice catch! It looks like we were trying to remove the intent from the session using the `woocommerce_after_account_payment_methods` hook, but that code wasn't being reached. Thanks for finding that PR!

includes/class-wc-payments-token-service.php
mdmoore: This didn't affect display of tokens but it did cause tokens created by other gateways to be deleted when `$stored_tokens` was compared with `$payment_methods`, since those tokens weren't payment methods in the express account. Thanks for catching this! Fixed in https://github.com/Automattic/woocommerce-payments/pull/4670/commits/cc5a465171d908dc4f12e41563d39a6d9ce32a04.

includes/class-wc-payments-token-service.php
mdmoore: See the comment above.

includes/class-wc-payments-token-service.php
mdmoore: This would require a function that returns all reusable gateway IDs. Such a function doesn't currently exist unfortunately. It might be good to have at some point though. For now, this the only place it would be used.

includes/class-wc-payments-token-service.php
tommyshellberg: Verified that other tokens aren't deleted when deleting SEPA tokens. 👍

includes/class-wc-payments-token-service.php
```diff
@@ -62,7 +62,7 @@ public function add_token_to_user( $payment_method, $user ) {
 
 		if ( Payment_Method::SEPA === $payment_method['type'] ) {
 			$token = new WC_Payment_Token_WCPay_SEPA();
-			$token->set_gateway_id( CC_Payment_Gateway::GATEWAY_ID );
+			$token->set_gateway_id( 'woocommerce_payments_sepa_debit' );
```
tommyshellberg: Nitpick: It could be worth making this a constant like we do in our JS constants file:

https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/constants.js#L4

includes/class-wc-payments.php
```diff
@@ -507,16 +508,16 @@ public static function register_gateway( $gateways ) {
...
 			}
 
 			if ( is_add_payment_method_page() ) {
-				return $reusable_methods;
+				return array_merge( $gateways, $reusable_methods );
 			}
 		}
 
-		return $gateways;
+		return array_merge( $gateways, $all_upe_gateways );
```
tommyshellberg, line 520: Nitpick: you could probably do this `array_merge()` within the above if block such that you're only doing an array_merge if UPE is enabled. 

```suggestion
	public static function register_gateway( $gateways ) {
		$gateways[]       = self::$legacy_card_gateway;
		$reusable_methods = [];

		if ( WC_Payments_Features::is_upe_enabled() ) {
			$all_upe_gateways = [];
			foreach ( self::$card_gateway->get_payment_method_ids_enabled_at_checkout() as $payment_method_id ) {
				if ( 'card' === $payment_method_id ) {
					continue;
				}
				$upe_gateway        = self::get_payment_gateway_by_id( $payment_method_id );
				$upe_payment_method = self::get_payment_method_by_id( $payment_method_id );

				if ( $upe_payment_method->is_reusable() ) {
					$reusable_methods[] = $upe_gateway;
				}

				$all_upe_gateways[] = $upe_gateway;

			}

			if ( is_add_payment_method_page() ) {
				return array_merge( $gateways, $reusable_methods );
			}
			array_merge( $gateways, $all_upe_gateways );
		}

		return $gateways;
	}
```

Then, you can always just return $gateways from this function

includes/class-wc-payments.php
```diff
@@ -507,16 +508,16 @@ public static function register_gateway( $gateways ) {
...
 	}
```
tommyshellberg: Nitpick: you could probably do this `array_merge()` within the above if block such that you're only doing an array_merge if UPE is enabled. 

```suggestion
	public static function register_gateway( $gateways ) {
		$gateways[]       = self::$legacy_card_gateway;
		$reusable_methods = [];

		if ( WC_Payments_Features::is_upe_enabled() ) {
			$all_upe_gateways = [];
			foreach ( self::$card_gateway->get_payment_method_ids_enabled_at_checkout() as $payment_method_id ) {
				if ( 'card' === $payment_method_id ) {
					continue;
				}
				$upe_gateway        = self::get_payment_gateway_by_id( $payment_method_id );
				$upe_payment_method = self::get_payment_method_by_id( $payment_method_id );

				if ( $upe_payment_method->is_reusable() ) {
					$reusable_methods[] = $upe_gateway;
				}

				$all_upe_gateways[] = $upe_gateway;

			}

			if ( is_add_payment_method_page() ) {
				return array_merge( $gateways, $reusable_methods );
			}
			array_merge( $gateways, $all_upe_gateways );
		}

		return $gateways;
	}
```

Then, you can always just return $gateways from this function.
( Don't let Github commit this suggestion, I'm not sure how to do a multi-line suggestion 😆  )

tommyshellberg (approved):
This is working for me now 🥳 🎉 thanks for addressing the comments I had! I just have some really nitpicky stuff leftover but I'm approving it.
Tests are passing as well as my manual tests.

Even though it's not a huge PR in terms of lines written it deals with complex, abstract and critical code, I don't envy you for having to write this stuff 😬 
Great job on this!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1199,7 +1199,7 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 				throw new Exception( WC_Payments_Utils::get_filtered_error_message( $e ) );
 			}
 
-			$selected_payment_method_type = [ WC_Payments::get_gateway()->get_selected_stripe_payment_type_id() ];
+			$payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( null, true );
```
FangedParakeet: When the UPE is enabled, I think we get to this point essentially in two ways:

1. Selected payment method is card.
2. Selected payment method is a saved payment method.

The goal here is to return one single payment method in this array--that payment method being the one selected by the user. Using `get_payment_method_ids_enabled_at_checkout()` will give us just what we want in scenario 1.

However, in scenario 2--let's say the user has chosen to checkout with a saved SEPA token--we will fill `payment_method_types` with the full list of payment methods available at checkout, instead of just the single one we want.

Basically, in scenario 1, we want `payment_method_types` to be `[ 'card' ]`, but in scenario 2, we would like it to be `[ $this->stripe_id ]` or some other way of getting the currently selected payment method on its own in there.

includes/class-wc-payments-token-service.php
FangedParakeet:
```suggestion
			$token->set_gateway_id( UPE_Payment_Gateway::GATEWAY_ID . '_' . Sepa_Payment_Method::PAYMENT_METHOD_STRIPE_ID );
```
```suggestion
			$token->set_gateway_id( UPE_Payment_Gateway::GATEWAY_ID . '_' . Payment_Method::SEPA );
```
Alternatively, you could Frankenstein this together from some of the constants we already have? Otherwise creating a new `Enum` or something like that sounds like a good plan.

includes/class-wc-payments-token-service.php
```diff
@@ -137,12 +140,13 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
...
 		// Prevent unnecessary recursion, WC_Payment_Token::save() ends up calling 'woocommerce_get_customer_payment_tokens' in some cases.
 		remove_action( 'woocommerce_get_customer_payment_tokens', [ $this, 'woocommerce_get_customer_payment_tokens' ], 10, 3 );
+
 		foreach ( $payment_methods as $payment_method ) {
 			if ( ! isset( $payment_method['type'] ) ) {
 				continue;
 			}
 
-			if ( ! isset( $stored_tokens[ $payment_method['id'] ] ) ) {
+			if ( ! isset( $stored_tokens[ $payment_method['id'] ] ) && ( 'card' === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID === $gateway_id || 'sepa_debit' === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA === $gateway_id ) || empty( $gateway_id ) ) {
```
FangedParakeet:
```suggestion
			if ( ! isset( $stored_tokens[ $payment_method['id'] ] ) && ( Payment_Method::CARD=== $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID === $gateway_id || Payment_Method::SEPA === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA === $gateway_id ) ) {
```
Might as well use all the constants we've got. Also, consider breaking this up into a multi-line statement: this was a challenging read. 😅 

Also, do we need that last `empty( $gateway_id )` check? Presumably, that's not possible, if we've made it this far.

includes/class-wc-payments-token-service.php
```diff
@@ -187,7 +194,10 @@ public function woocommerce_payment_token_deleted( $token_id, $token ) {
 	 * @param WC_Payment_Token $token    Token object.
 	 */
 	public function woocommerce_payment_token_set_default( $token_id, $token ) {
-		if ( WC_Payment_Gateway_WCPay::GATEWAY_ID === $token->get_gateway_id() ) {
+
+		$gateway_ids = [ WC_Payment_Gateway_WCPay::GATEWAY_ID, WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA ];
```
FangedParakeet: It seems like this `$gateway_ids` array gets created and used a lot in this class. Is it perhaps worth defining this as a class constant or a private member variable in the constructor (I can't remember if PHP will let you do string concatenation when defining a class constant)?

includes/class-wc-payments.php
```diff
@@ -492,8 +492,9 @@ public static function add_plugin_links( $links ) {
...
 	 */
 	public static function register_gateway( $gateways ) {
-		$gateways[]         = self::$legacy_card_gateway;
-		$reusable_methods[] = self::$legacy_card_gateway;
+		$gateways[]       = self::$legacy_card_gateway;
+		$all_upe_gateways = [];
+		$reusable_methods = [];
 
 		if ( WC_Payments_Features::is_upe_enabled() ) {
 			foreach ( self::$card_gateway->get_payment_method_ids_enabled_at_checkout() as $payment_method_id ) {
```
```suggestion
		$gateways[]       = self::$legacy_card_gateway;

		if ( WC_Payments_Features::is_upe_enabled() ) {
	  		$all_upe_gateways = [];
    		$reusable_methods = [];
			foreach ( self::$card_gateway->get_payment_method_ids_enabled_at_checkout() as $payment_method_id ) {
```
Not that important, but maybe these should be defined inside the if UPE block.

FangedParakeet (changes requested):
I've left a few comments here, but nothing serious, for better or worse.

I was unable to save payment methods at checkout sadly. When using the card gateway or the SEPA gateway, the attempt to confirm the intent returned the following error.

```log
2022-10-07T16:12:43+00:00 ERROR `setup_future_usage` cannot be used with one or more of the values you specified in `payment_method_types`. Please remove `setup_future_usage` or remove these types from `payment_method_types`: ["eps", "giropay", "p24"]. (invalid_request_error)
```

(I have all the available payment methods enabled, by the way.)

I was a little surprised by this, but presume this was due to [this line](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payment-gateway-wcpay.php#L1202) returning all enabled payment methods instead of just the current.

Looking at that line it does make a little more sense.

```php
$payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( null, true );
```

When the UPE is enabled, `WC_Payments::get_gateway()` will return the `UPE_Payment_Gateway` instance, not the legacy card gateway instance of `WC_Payment_Gateway_WCPay`. Perhaps you might want to modify this to `$this->get_payment_method_ids_enabled_at_checkout( null, true )`; however, I've left another comment about why this might not have a desired effect when saved SEPA tokens are used at checkout. 

Actually, thinking further on this, I think using `$this` would not allow you to save a SEPA payment method at checkout either, because all of the UPE payment methods enabled at checkout would be returned. We only want to fill that parameter with a single, selected payment method. 

I was able to add both cards and SEPA payment methods from the My Account page however. In fact, doing so actually triggered a forgotten SEPA payment method to appear as a saved payment method, so something is working here. The payment tokens look normal in the database and the saved payment methods appear where they should do at checkout, so that's great news! 

Halfway there, so let me know if I can help out with anything in the meantime. Cheers!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1199,7 +1199,20 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
+				$upe_payment_method = sanitize_text_field( wp_unslash( $_POST['payment_method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification
+
+				if ( 'woocommerce_payments' !== $upe_payment_method ) {
+					$upe_payment_method = str_replace( 'woocommerce_payments_', '', $upe_payment_method );
+				} else {
+					$upe_payment_method = 'card';
+				}
+
+				$payment_methods = [ $upe_payment_method ];
+			}
```
FangedParakeet:
```suggestion

			$payment_gateway = sanitize_text_field( wp_unslash( $_POST['payment_method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification

			if ( 'woocommerce_payments' !== $payment_gateway ) {
				$payment_gateway = str_replace( 'woocommerce_payments_', '', $payment_gateway );
			} else {
				$payment_gateway = 'card';
			}

			$payment_methods = [ $payment_gateway ];

```
The logic here looks good to me...buuut unfortunately this entire `WC_Payments_Features::is_upe_enabled()` is soon going to be removed in #4919, so we won't be able to use that as a condition here. Though, wouldn't that second block work in either case, whether the gateway is a UPE gateway or not? Can we not just use that on its own?

mdmoore:
@FangedParakeet I appreciate your attention to detail! I've addressed the feedback from your previous review if you don't mind giving it another look when you have a chance.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -100,6 +100,8 @@ public function __construct(
 		add_action( "wc_ajax_wcpay_create_payment_intent_$this->stripe_id", [ $this, 'create_payment_intent_ajax' ] );
 		add_action( "wc_ajax_wcpay_update_payment_inten_ $this->stripe_id", [ $this, 'update_payment_intent_ajax' ] );
 		add_action( "wc_ajax_wcpay_init_setup_intent_$this->stripe_id", [ $this, 'init_setup_intent_ajax' ] );
+		add_action( 'woocommerce_after_account_payment_methods', [ $this, 'remove_upe_setup_intent_from_session' ], 10, 0 );
+		add_action( 'woocommerce_order_payment_status_changed', [ __CLASS__, 'remove_upe_payment_intent_from_session' ], 10, 0 );
```
FangedParakeet: Nice catch here, ensuring that this hook fires for every payment method! I don't think it's necessarily related to the changes here, but do you think we should move the `woocommerce_subscription_payment_method_updated` hook up in the main block as well, for the same reason?

FangedParakeet (commented):
All the code here looks pretty good to me and I don't have too many comments on modifications to make on your additions...That being said I did encounter some misfortunes while testing these changes.

When I was able to add payment methods, they ended up in the correct gateway, which was cool to see and I was able to checkout successfully using both saved card and SEPA payment methods. However, I faced quite a few issues while attempting to add new payment methods, mostly due to clashing caching. 

### Adding payment method from My Account page

When adding a payment method from the My Account page, after entering my payment information, submitting the form, and returning to the list of payment methods at `/my-account/payment-methods`, I found that my newly added payment method did not appear in the list. 

Checking my Stripe dashboard, I could find that the payment method had been added successfully, but by checking the debug logs, I could see that no request was being made to `GET /wcpay/payment_methods` to refresh the local database cache of payment methods. For some reason, either [this function](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payments-customer-service.php#L265) was not firing or it was not doing its job as expected. I was able to trigger the cache to clear by either reopening the Add Payment Method form (`/my-account/add-payment-method`) or visiting the checkout, where suddenly all my new saved payment methods would appear at last. 

Also, I was able to often inconsistently trigger the below ugly looking PHP warning. 

```
Warning: Illegal string offset 'data' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-customer-service.php on line 223 
```

This didn't happen consistently, but I could trigger this pretty regularly by adding, deleting, or listing payment methods from My Account dashboard. This warning refers to [this line](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payments-customer-service.php#L223) from the Customer Service, where we parse the response from the `GET /wcpay/payment_methods` request containing the user's Stripe stored payment methods. 

However, this actually originates from the Token Service: the issue is that in [this loop](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payments-token-service.php#L132-L133), we make requests to Stripe for payment methods of _all_ enabled payment method types (e.g. including Bancontact, Sofort, etc. when enabled). Stripe stored payment methods can _only_ be of the type `card` or `sepa_debit` (be they either saved SEPA payment methods or other payment methods--Bancontact, Ideal, etc.--that are converted to SEPA when saved), so we should only query these two payment method types. 

On the Stripe gateway, [we created a new `retrievable_type` property](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-payment-tokens.php#L257-L266) to gather these underlying storable payment method types. However, that might be overkill here, since for now the only reusable payment methods are `card` and `sepa_debit`, so we don't need to worry about deciphering which stored SEPA methods were actually converted from other payment methods. We _should_ still restrict the payment method types that we request saved payment methods to a maximum of just `card` and `sepa_debit`.

### Adding payment methods from checkout

The UPE saved payment method flow works by using the `setup_future_usage` flag to ensure that a payment method is stored on Stripe and clearing our local payment methods cache, so that we will, in the future, request a fresh list of customer payment methods from Stripe, when it comes time to display a customer's saved payment tokens. However, our redirection flow is not currently working as it should when SEPA is chosen as a payment gateway at checkout. 

For some reason, large sections of the UPE redirection code are not executing for SEPA. I can see in the debug logs that we reach [this point](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/payment-methods/class-upe-payment-gateway.php#L616), where we collect and examine the confirmed payment intent; however, at this point--without any errors--the trail goes cold. 

A new payment method may be stored with Stripe, but we don't clear our local cache, so a WC token is not created. The status of the order does not update and there are no notes left on the order, other than that of a payment being authorised. Worst of all, it seems like [this line](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/payment-methods/class-upe-payment-gateway.php#L665) is either ignored or ineffective: the SEPA payment intent cached to the session does not clear, so that if we begin a new checkout, when we try to place our order, we discover that the UPE is still using the already confirmed intent from the previous transaction and we are rightfully shown the following error!

```
The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.
```

I wasn't able to fully investigate these cascading catastrophes, but it's worth noting that SEPA is a ["delayed notification"](https://stripe.com/docs/payments/payment-methods#payment-notification) payment method, meaning that upon confirming the intent, the payment does not succeed immediately. Since the intent status might not be exactly what we would expect, this could cause us to mishandle these transactions. Let me know if you have any ideas here, because I think I would need to dig into this deeper to firgure what's going wrong (perhaps in a separate issue)...

### Lowercase SEPA labelling

This is a smaller issue, but is it just me or do saved SEPA tokens always appear styled like this in WooCommerce?

Sepa iban is wrong
_"Sepa iban"?_

SEPA IBAN should definitely be capitalised and I can't find anywhere where we outwardly refer to this payment method in lowercase, so I'm not fully sure where this formatting comes from. It's a small issue, but let me know if you see the same display, since we probably should figure out how to fix this.

***

Sorry, I've left you another novel to go through here, but let me know if I'm testing anything incorrectly or if you're able to casually explain away any of my erroneous encounters! If you are able to share the pain of any of these errors with me, let me know if you'd like to move any of these to separate issues. I think the checkout errors and probably also the labelling issues can definitely be moved to their own issues, if you're able to sort out everything mentioned on the Account pages.

Anyways, looking forward to hearing your responses to my exhausting feedback (sorry again!), let me know if I've managed to muddle up anything else along the way. Cheers!

mdmoore:
@FangedParakeet  - Thanks once again for the thorough review. 

### Adding payment method from My Account page

>When adding a payment method from the My Account page, after entering my payment information, submitting the form, and returning to the list of payment methods at /my-account/payment-methods, I found that my newly added payment method did not appear in the list.

It turns out this wasn't due to caching after all. When we list tokens, we also decide whether to add tokens to the user or not. To prevent tokens from being duplicated across gateways, we check that the payment method `type` matches the `gateway_id` on which the code is currently running, in the case of the Add Payment Method and Checkout pages. But, on My Account > Payment Methods, there is no gateway ID, so the token was never added to the user. I handled this previously by using [this `empty( $gateway_id )` check](https://github.com/Automattic/woocommerce-payments/pull/4670#discussion_r990160661). But when that line was brought into question, I couldn't remember or find it's purpose even though I felt like it was doing something important. Added back, newly added UPE methods immediately appear on the Payment Methods page upon redirect.

>Also, I was able to often inconsistently trigger the below ugly looking PHP warning.
>Warning: Illegal string offset 'data' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-customer-service.php on line 223 
>the issue is that in [this loop](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payments-token-service.php#L132-L133), we make requests to Stripe for payment methods of all enabled payment method types (e.g. including Bancontact, Sofort, etc. when enabled)

This is a great catch! Unfortunately, even after restricting the payment method types to `card` and `sepa_debit`, the warning still pops up at random. Sometimes it's very easy to reproduce the warning and other times it just won't happen. I was able to catch it one time while debugging the `$payment_methods` array. This array should contain 2 arrays: one for all `card` and one for all `sepa_debit` payment methods stored in Stripe. Instead, one array might be returned as a string.

```php
array(2) { [0] => string(1) " " [1] => array(3) { [0] => array(9)... }
```

This happens [here](https://github.com/Automattic/woocommerce-payments/blob/a747857083cb953b64608dbe761501feded3953a/includes/class-wc-payments-customer-service.php#L224), before we flatten the array. It feels like an error in the response.

>The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.

I encountered this before as well but thought it to be solved. I am able to reproduce it again now though. 

>Lowercase SEPA labeling

This isn't something I've encountered yet. I've always seen the all caps, **SEPA IBAN**. I'm not sure what might cause this unless it could be a browser extension perhaps?

@FangedParakeet - How do you feel about moving to merge this now that tokens are working on the My Account > Payment Methods page, then creating new issues for the intermittent warning and checkout errors?

mdmoore:
@FangedParakeet  - Thanks once again for the thorough review. 

### Adding payment method from My Account page

>When adding a payment method from the My Account page, after entering my payment information, submitting the form, and returning to the list of payment methods at /my-account/payment-methods, I found that my newly added payment method did not appear in the list.

It turns out this wasn't due to caching after all. When we list tokens, we also decide whether to add tokens to the user or not. To prevent tokens from being duplicated across gateways, we check that the payment method `type` matches the `gateway_id` on which the code is currently running, in the case of the Add Payment Method and Checkout pages. But, on My Account > Payment Methods, there is no gateway ID, so the token was never added to the user. I handled this previously by using [this `empty( $gateway_id )` check](https://github.com/Automattic/woocommerce-payments/pull/4670#discussion_r990160661). But when that line was brought into question, I couldn't remember or find it's purpose even though I felt like it was doing something important. Added back, newly added UPE methods immediately appear on the Payment Methods page upon redirect.

>Also, I was able to often inconsistently trigger the below ugly looking PHP warning.
>Warning: Illegal string offset 'data' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-customer-service.php on line 223 
>the issue is that in [this loop](https://github.com/Automattic/woocommerce-payments/blob/fix/4530-sepa-token-display/includes/class-wc-payments-token-service.php#L132-L133), we make requests to Stripe for payment methods of all enabled payment method types (e.g. including Bancontact, Sofort, etc. when enabled)

This is a great catch! Unfortunately, even after restricting the payment method types to `card` and `sepa_debit`, the warning still pops up at random. Sometimes it's very easy to reproduce the warning and other times it just won't happen. I was able to catch it one time while debugging the `$payment_methods` array. This array should contain 2 arrays: one for all `card` and one for all `sepa_debit` payment methods stored in Stripe. Instead, one array might be returned as a string.

```php
array(2) { [0] => string(1) " " [1] => array(3) { [0] => array(9)... }
```

This happens [here](https://github.com/Automattic/woocommerce-payments/blob/a747857083cb953b64608dbe761501feded3953a/includes/class-wc-payments-customer-service.php#L224), before we flatten the array. It feels like an error in the response.

>The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.

I encountered this before as well but thought it to be solved. I am able to reproduce it again now though. 

>Lowercase SEPA labeling

This isn't something I've encountered yet. I've always seen the all caps, **SEPA IBAN**. I'm not sure what might cause this unless it could be a browser extension perhaps?

@FangedParakeet - How do you feel about moving to merge this now that tokens are working on the My Account > Payment Methods page, then creating new issues for the intermittent warning and checkout errors?

includes/class-wc-payments-token-service.php
```diff
@@ -120,13 +123,14 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
...
+				if ( in_array( $token->get_gateway_id(), self::REUSABLE_GATEWAYS, true ) ) {
 					$stored_tokens[ $token->get_token() ] = $token;
 				}
 			}
 
-			$payment_methods = [ [] ];
-			foreach ( WC_Payments::get_gateway()->get_upe_enabled_payment_method_ids() as $type ) {
+			$retrievable_payment_method_types = [ Payment_Method::CARD, Payment_Method::SEPA ];
+			$payment_methods                  = [];
+			foreach ( $retrievable_payment_method_types as $type ) {
```
FangedParakeet:
```suggestion
			$retrievable_payment_method_types = [ Payment_Method::CARD ];
			if ( in_array( Payment_Method::SEPA, WC_Payments::get_gateway()->get_upe_enabled_payment_method_ids(), true ) {
				$retrievable_payment_method_types[] = Payment_Method::SEPA;
			}
			$payment_methods                  = [];
			foreach ( $retrievable_payment_method_types as $type ) {
```
Just a curt caveat that we only to request the SEPA payment methods when SEPA is actually enabled at checkout. Otherwise even when SEPA is disabled, I believe we will still show users their saved SEPA tokens on the Accounts page.

FangedParakeet:
> But, on My Account > Payment Methods, there is no gateway ID, so the token was never added to the user. I handled this previously by using https://github.com/Automattic/woocommerce-payments/pull/4670#discussion_r990160661.

Whoops, sorry for making you remove that line! That makes perfect sense, even though it is a little irritating that the `$gateway_id` parameter is always missing on the Accounts page. 

> Unfortunately, even after restricting the payment method types to card and sepa_debit, the warning still pops up at random.

😢 

> This happens [here](https://github.com/Automattic/woocommerce-payments/blob/a747857083cb953b64608dbe761501feded3953a/includes/class-wc-payments-customer-service.php#L224), before we flatten the array. It feels like an error in the response.

Yep, agreed. I think the error is in the response handling and should be handled in the WPay Server API client, before it reaches the token service. 

> How do you feel about moving to merge this now that tokens are working on the My Account > Payment Methods page, then creating new issues for the intermittent warning and checkout errors?

This sounds like a good plan. Right now my local environment is all twisted up for some reason and not returning any WCPay gateways at checkout on this branch, but I'm going to try to fix that and retest tomorrow morning. Here are a few things that I think we can probably handle in separate issues, so let me know if you agree.

 - Saved payments functionality from shortcode checkout
 - Saved payments functionality from blocks checkout
 - API error handling for payment method requests

includes/class-wc-payments-token-service.php
```diff
@@ -120,13 +123,14 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
...
 			foreach ( $tokens as $token ) {
-				if ( WC_Payment_Gateway_WCPay::GATEWAY_ID === $token->get_gateway_id() ) {
+				if ( in_array( $token->get_gateway_id(), self::REUSABLE_GATEWAYS, true ) ) {
 					$stored_tokens[ $token->get_token() ] = $token;
 				}
 			}
 
-			$payment_methods = [ [] ];
-			foreach ( WC_Payments::get_gateway()->get_upe_enabled_payment_method_ids() as $type ) {
+			$retrievable_payment_method_types = [ Payment_Method::CARD, Payment_Method::SEPA ];
```
FangedParakeet: Collecting all of the saved tokens should be fine for now, since we filter out irrelevant payment method types later in this function. Alternatively, if `$gateway_id` is empty, you could loop through all available retrievable payment method types, whereas if the `$gateway_id` is present, we could simply request tokens from the relevant payment method type first, to simplify our requests and avoid filtering the list manually later. 

I think we still need to check whether all retrievable payment method types are currently enabled at checkout though. I was able to create a slightly unintended effect by saving some SEPA payment methods and then disabling SEPA as a payment method. SEPA was removed from checkout and the Add payment method pages, but my saved SEPA payment method still appeared on the My Accounts > Payment methods page.

Moreover since the code in this function creates a new WC Token for new payment methods found on Stripe, since this SEPA WC Token was no longer present in the tokens list, with SEPA disabled as a payment method, when Stripe returned this saved payment method our code created a new WC Token for it. When I re-enabled SEPA as a payment method, I now was shown more SEPA tokens than I had originally stored due to the additional duplicates. 

I'm fine to create a new issue for this, however, since it is a little edge-casey.

includes/class-wc-payments-token-service.php
mdmoore: Thanks for catching that. Makes total sense. -> https://github.com/Automattic/woocommerce-payments/pull/4670/commits/d3e9a3dc08c21b61185b0e8fff20f513629830f5

mdmoore:
>Here are a few things that I think we can probably handle in separate issues, so let me know if you agree.

>Saved payments functionality from shortcode checkout
    Saved payments functionality from blocks checkout
    API error handling for payment method requests

Yep, these are all good issues to pull from the discovery in this PR. I'll get these added.

FangedParakeet (approved):
I've left a few comments, but this is generally working pretty well for me from the My Accounts page now!

On one occasion, I was met with this ugly error message.

WTF
_Wtf?_

But checking the error log, I think my local WCPay Server was just throwing a tantrum, because the setup intent was created successfully, a new payment method was added by Stripe, but for some reason the response came back empty, causing this nonsense. I think we need better error handling for these such cases, but we've discussed that already and this can definitely be move to a new issue.

Also, I had to downgrade npm to ~v6 to get build the packages via `npm install`. Dunno why that happened or whether that could have caused anything above. Anyways, let's just forget about that for now.

TL;DR: I'm going to give this PR my blessings!...But first, let me describe the things that worked, the things that didn't, and the things we can move to new issues. 

### Things I tested

- Listing saved payment methods on My Accounts > Payment methods ✅ 
- Listing saved payment methods on Add payment methods page ✅ 
- Listing saved payment methods on shortcode checkout ✅ 
- Listing saved payment methods on blocks checkout ✅ 
- Adding saved payment methods (UPE & non-UPE) from My Accounts ✅ 
- Adding saved payment methods (UPE & non-UPE) from shortcode checkout ✅ 
- Adding saved payment methods (UPE and non-UPE) from blocks checkout ❌ 
- Completing checkout using non-UPE saved payment method from shortcode checkout ✅ 
- Completing checkout using UPE saved payment method from shortcode checkout ✅ 
- Completing checkout using non-UPE saved payment method from shortcode checkout ✅ 
- Completing checkout using UPE saved payment method from blocks checkout ✅ 
- Completing checkout using non-UPE saved payment method from blocks checkout ❌ 

### Things that didn't work quite right

#### Blocks checkout

This is generally quite broken, but #4681 has just been merged into the feature branch and not yet updated here, so I'm optimistic that some of these issues will be resolved automatically by merging...though, after also reviewing that PR, I think a few of them might remain as well. 😅 

Anyways, no matter for now. We'll solve this elsewhere.

#### SEPA Tokens

I left a comment about how to make duplicate SEPA tokens unintentionally and get them to appear at on the Accounts page when SEPA is disabled as a payment method...but I think you solved it right as I've been writing this review, so I'll count that resolved. 😁 

#### SEPA Token labelling

I found that on the form pages (i.e. add payment method, shortcode checkout) the labelling was capitalised correctly...

SEPA IBAN
_SEPA IBAN: excellent_

...But on the main payment methods list page, where all payment methods are shown together and on the blocks checkout page, the payment method title is in lowercase.

Sepa iban on payment methods list page
_Sepa iban??? My eyes!_

Sepa iban on blocks checkout list
_Sepa iban again??? Someone please make it stop!_

I honestly have no idea why this is happening and hope it isn't a problem that needs to be solved in WC Core. Anyways, this can definitely be moved to a new issue. 

#### WCPay Server API Errors

From time to time these random hideous PHP errors about missing `data` properties fill up the entire screen, usually solvable by simply refreshing the page. This is basically happening because for some reason the response from the WCPay Server is empty. I hope this happens more frequently locally, due to some Docker temperamentality, but either way, I think we could perhaps handle this better by swallowing the message or retrying such requests. 

### Things that can be solved in new issues

- Blocks checkout compatibility
- SEPA labelling
- API error handling

### Conclusion

But in general, despite all of that, this PR is working really well in pretty much all use-cases! Thanks for running this marathon with me and responding to all of my cluster-bombed feedback along the way. 

I think this is a huge step forward and definitely now ready to join the rest of the party!

:shipit:

includes/class-wc-payments-token-service.php
FangedParakeet: Pretty sure this is resolved in [`d3e9a3d` (#4670)](https://github.com/Automattic/woocommerce-payments/pull/4670/commits/d3e9a3dc08c21b61185b0e8fff20f513629830f5). At the least, I can't readily repeat this issue anymore. 👍

mdmoore:
@FangedParakeet Thanks a ton for sticking with me on this. Your comments and reviews have been invaluable. The SEPA token labeling problem is a strange one that I haven't encountered anywhere that the label is used. This is what I see on the **My Account > Payment Methods** page: 

image

It'll be good to have some deeper investigation on this. I could see it being due to some difference in version among WooCommerce, WooCommerce Blocks, or WCPay.

After merging `poc/upe-instances-multiplied`, I notice a couple new issues when adding new payment methods. 

1. Adding a new card is failing now that card has been removed from the UPE payment method config [here](https://github.com/Automattic/woocommerce-payments/blob/644443233bb99343d903a613ad33d1a3c4b71c65/includes/payment-methods/class-upe-payment-gateway.php#L1095-L1097)
2. When adding a SEPA payment method, success notices are now duplicated.

With that, our list of new issues to be created will be: 

1. Blocks checkout compatibility
2. SEPA labeling
3. API error handling
4. Saving card payment methods
5. Duplicate notices when saving SEPA methods

I'll add these ASAP 👍

Issue #4506, opened 2022-07-29
[Authorizations] Capture payment from payment details screen


mgascam:
### Description

As part of an effort to improve the capture payment workflow, we want to inform merchants when a payment needs to be captured in the payment details screen. We will also add a `Capture charge` button to allow them to capture it without having to navigate to the Order details screen.

### Acceptance criteria

* Show a new section in the Payment details screen when a payment needs to be captured.
* The new section contains:
  * Text informing the user of the capture deadline: `You need to capture this charge before...`
  * A `Capture charge` button
* Clicking the button will capture the charge.
* Expose new endpoint that the front end will use to capture the charge.

### Designs

<img width="1278" alt="Screen Shot 2022-07-29 at 14 22 24" src="https://user-images.githubusercontent.com/1553182/181757473-a7a42c7b-7071-4f73-8cdc-b997e33f79fe.png">

### Dev notes

Check if we can reuse [WC_Payment_Gateway_WCPay::capture_charge](https://href.li/?https://github.com/Automattic/woocommerce-payments/blob/147a4fc73a0237f79128656e5cd5e07abc6ae5a5/includes/class-wc-payment-gateway-wcpay.php#L2110)

PR #4676, opened 2022-09-01
Capture authorizations from table

mgascam:
Fixes #

This PR adds a Capture button to each row of the new authorizations list table. This button will allow merchants to capture an authorization from the `Uncaptured` tab in the `Transactions` page in WP Admin. 

It uses mocked data, since the server endpoints are not available yet. For this reason, it is not yet possible to manually test the complete flow. Only the loading and error states can be tested manually at this moment.

The scope of changes on this PR includes:

- UI components: adds a new `CaptureAuthorizationButton` to the list of non captured authorizations
- State management boilerplate in the front end
- Refactor `capture_terminal_payment` endpoint. The functionality needed to capture a terminal payment and to capture an authorization are very similar. To avoid code duplication, I refactored the endpoint to make it work for both types of payments, passing a new `$is_terminal_payment` parameter.

https://user-images.githubusercontent.com/1553182/188485604-d5de638e-8af2-491e-964b-8416a8ac8a3b.mp4

#### Testing instructions

* Set `displayAuthorizations = true ` [here](https://github.com/Automattic/woocommerce-payments/blob/e9fec5a7a492ecbe59171991867e4f831b9d1296/client/transactions/index.tsx#L24). Don't forget to rebuild statics (i.e. `npm start`)
* Navigate to `WP Admin > Transactions > Uncaptured` tab. You should see a `Capture` button in each row of the table.
* Click on any `Capture` button
* Check that the button transitions to loading state
* Check that it errors eventually and displays an error notice.

-------------------

**Post merge**


client/data/authorizations/resolvers.ts
```diff
@@ -64,6 +72,49 @@ export function* getAuthorizations( query: Query ): any {
...
+	const randomAmount = () => {
+		return 1400 + Math.floor( Math.random() * 5000 );
+	};
+
+	const randomDate = dateI18n(
+		'M j, Y / g:iA',
+		moment.utc( new Date() ).local().toISOString()
+	);
+
+	const randomCaptureDate = dateI18n(
```
anu-rock: Is the time part necessary here? The design screenshot in #4506 doesn't have it.

Also, when the data is eventually wired up with server, would there be a benefit to have this string in `ISO-8601` format?

kalessil (changes requested):
@mgascam, could you please update the branch and look into CI failures?

client/data/authorizations/actions.ts
```diff
@@ -39,3 +51,59 @@ export function updateAuthorizationsSummary(
...
+
+		authorization = {
+			authorization_id: id,
+			captured: true,
+		};
+
+		yield updateAuthorization( authorization as Authorization );
+
+		// Need to invalidate the resolution so that the components will render again.
+		yield dispatch(
```
anu-rock, line 100: I wonder if there's a side effect of invalidating resolutions for selectors on a page where they are not used? Eg. `getAuthorizations` is used on Transactions page but not on Payment Details page. I have not noticed anything, but wanted to confirm.

client/data/authorizations/resolvers.ts
anu-rock: Although mocked data here will eventually be removed, using static (not random) dates would help in writing predictable tests.

client/data/authorizations/selectors.ts
```diff
@@ -18,24 +28,49 @@ import {
  *
  * @return {Object} The list of authorizations for the given query.
  */
-const getAuthorizationsForQuery = ( state: Record< string, any > ) => {
-	return state.authorizations;
+const getAuthorizationsForQuery = (
+	state: Record< string, any >,
+	query: Query
+) => {
+	const index = getResourceId( query );
```
anu-rock, line 35: Thinking out loud - using a stringified query string as index makes it difficult to visualize how the state is maintained. Would having a simpler index be an oversimplification of the underlying logic?

client/data/authorizations/test/selectors.test.ts
```diff
@@ -0,0 +1,146 @@
...
+		authorizations: {
+			byId: {},
+			[ getResourceId( mockQuery ) ]: {
+				data: mockAuthorizations,
+			},
+			summary: {},
+		},
+	};
+
+	test( 'Returns empty disputes list when disputes list is empty', () => {
```
anu-rock: We may have a typo here - `disputes` in place of `authorizations`.

client/transactions/uncaptured/capture-authorization-button.tsx
```diff
@@ -0,0 +1,42 @@
+/** @format **/
```
anu-rock: I think this button can be moved to another location for its reuse potential.

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -137,21 +173,26 @@ public function capture_terminal_payment( $request ) {
 			}
 
 			// Update the order: set the payment method and attach intent attributes.
-			$order->set_payment_method( WC_Payment_Gateway_WCPay::GATEWAY_ID );
-			$order->set_payment_method_title( __( 'WooCommerce In-Person Payments', 'woocommerce-payments' ) );
+			if ( $is_terminal_payment ) {
+				$order->set_payment_method( WC_Payment_Gateway_WCPay::GATEWAY_ID );
+				$order->set_payment_method_title( __( 'WooCommerce In-Person Payments', 'woocommerce-payments' ) );
+			}
 			$intent_id     = $intent->get_id();
```
anu-rock: We can move the declaration for these four variables above and merge the below condition with the one above.

client/data/authorizations/test/selectors.test.ts
mgascam: Oops thanks. Fixed in [d4431d4](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/d4431d4d6c8ae4d367f5ab2a69125d17bdcd4eb3)

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -168,7 +209,7 @@ public function capture_terminal_payment( $request ) {
 				'id'     => $intent->get_id(),
 			];
 
-			$result = $is_intent_captured ? $result_for_captured_intent : $this->gateway->capture_charge( $order, false );
+			$result = ( $is_intent_captured && $is_terminal_payment ) ? $result_for_captured_intent : $this->gateway->capture_charge( $order, false );
```
anu-rock: The call to `capture_charge` is now further hidden down the ternary chain. An opportunity to refactor this ternary into an `if-else`?

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -534,6 +534,399 @@ public function test_capture_terminal_payment_not_found() {
...
+		$this->assertEquals(
+			[
+				'status' => 'succeeded',
+				'id'     => $this->mock_intent_id,
+			],
+			$response_data
+		);
+
+		$result_order = wc_get_order( $order->get_id() );
+		$this->assertEquals( 'woocommerce_payments', $result_order->get_payment_method() );
```
anu-rock: On behalf of @kalessil - `assertSame` :)

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -534,6 +534,399 @@ public function test_capture_terminal_payment_not_found() {
...
+		$this->assertEquals( 'WooCommerce Payments', $result_order->get_payment_method_title() );
+		$this->assertEquals( 'processing', $result_order->get_status() );
+		$url = '/wc/v3/' . ( $this->is_wpcom() ? 'sites/3/' : '' ) . 'payments/readers/receipts/' . $this->mock_intent_id;
+		$this->assertStringEndsWith( $url, $result_order->get_meta( 'receipt_url' ) );
```
anu-rock: Is it necessary to assert reader receipt here?

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -534,6 +534,399 @@ public function test_capture_terminal_payment_not_found() {
...
+	}
+
+	public function test_capture_authorization_succeeded_intent() {
```
anu-rock, line 594: Most of the tests for `capture_authorization` are similar to ones for `capture_terminal_payment`. Perhaps we can remove tests with exact same assertions and keep the ones with some difference?

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -103,7 +117,29 @@ public function register_routes() {
...
+	}
+
+	/**
+	 * Handles payment capture.
+	 *
+	 * @param  WP_REST_Request $request Full data about the request.
+	 * @param  bool            $is_terminal_payment Marks if the request is for a Terminal payment (In Person Payment or Interac at the moment).
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	private function handle_payment_capture( WP_REST_Request $request, bool $is_terminal_payment ) {
```
kalessil: This doesn't look right: `capture_terminal_payment` had a clear focus and goal.
Now it's making even more what is violating the single responsibility we tried to follow there.

How can we use the built-in capture functionality from order actions for the new endpoint?

kalessil (changes requested):

includes/admin/class-wc-rest-payments-orders-controller.php
mgascam: Absolutely, thanks!

includes/admin/class-wc-rest-payments-orders-controller.php
mgascam: Good point thanks!

includes/admin/class-wc-rest-payments-orders-controller.php
mgascam: Thanks for the review @kalessil! 

I [pushed a change](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/c352f2ee9e8f35d16c5bfb5a697aae51ceeff8b1) to have dedicated endpoints for `capture_terminal_payment` and `capture_authorization`.

For context, it turns out that the capturing of a terminal payment and an authorization share the step to prepare the order object before calling the API to capture the charge. To avoid code duplication I tried to factor out those common parts into a new `prepare_order_for_capture`. It accepts a `$is_terminal_payment parameter` since there is one additional step to do for Terminal payments. Please let me know what you think when you have a moment.

>How can we use the built-in capture functionality from order actions for the new endpoint?

The capture of the order from the order actions is triggered by an action hook: `woocommerce_order_action_capture_charge `. This action is handled in [WC_Payment_Gateway_WCPay::capture_charge](https://github.com/Automattic/woocommerce-payments/blob/c352f2ee9e8f35d16c5bfb5a697aae51ceeff8b1/includes/class-wc-payment-gateway-wcpay.php#L374). This is the same method that both [`capture_terminal_payment`](https://github.com/Automattic/woocommerce-payments/blob/c352f2ee9e8f35d16c5bfb5a697aae51ceeff8b1/includes/admin/class-wc-rest-payments-orders-controller.php#L149) and [`capture_authorization` ](https://github.com/Automattic/woocommerce-payments/blob/c352f2ee9e8f35d16c5bfb5a697aae51ceeff8b1/includes/admin/class-wc-rest-payments-orders-controller.php#L209) will call eventually. Does this answer your question or am I missing something?

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
mgascam: Thanks! Done in [c352f2e](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/c352f2ee9e8f35d16c5bfb5a697aae51ceeff8b1)

client/data/authorizations/resolvers.ts
mgascam: I removed the mocked data in favor of using fixtures in [eaa5fba](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/eaa5fbab64503b208bfa08223c547e90091d6d25) so I guess this comment no longer applies, please let me know otherwise @anu-rock

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
mgascam: You are right, it is not needed since the receipt_url is intended for Terminal payments. Removed in [85d34d2](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/85d34d2bf9646943845321a37e726feb0182b1d9)

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
mgascam: I agree, they are very similar but I could not find any tests that have the exact same assertions.

client/transactions/uncaptured/capture-authorization-button.tsx
mgascam: Indeed, thanks. Done in [63d289f](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/63d289fe6e9e4c89e21d197cf31ad4a1057524ef)

client/data/authorizations/actions.ts
mgascam: That's a good question, to answer it I dig the[ available docs](https://unfoldingneurons.com/2020/wordpress-data-store-properties-resolvers) for the @wordpres/data package. This is my understanding of how this will work, apologies if the explanation is too verbose.

The `invalidateResolutionForStoreSelector` is a 'special' type of action used by the @wordpres/data to signal the resolution state of a selector/resolver pair. It has its own internal reducer to deal with this type of actions. In this case, the `getAuthorizations` selector will be marked as unresolved in the internal state (aka `metadada`). This means that next time this selector is called, @wordpres/data will know that it needs to resolve it first (fetch from API). Please note that, even if the `getAuthorizations` selector is not used in the Payments Details page, it exists in the store's metadata. In the case it was already resolved (i.e the merchant navigated to `Uncaptured` before) it will be marked as unresolved and refreshed next time is accessed. In the case that it was never resolved, it will be marked as such.

For above reasons, I don't think there will be any undesired side effects. I hope this explanation makes sense.

client/data/authorizations/selectors.ts
mgascam: This will be used when we have pagination in place. If you check the transactions state for instance, you'll see that the state has this shape:

Screen Shot 2022-09-08 at 18 53 41

We will have something similar for authorizations when pagination is implemented. I'd prefer to keep it as it is since we will use it eventually, but I am also happy to change it if you think we should. I wonder what would be a better index in this case?

anu-rock: Sounds good, thanks for explaining :) We can keep it like this.

client/data/authorizations/resolvers.ts
anu-rock: It looks quite better now 👍

Now that the resolver returns empty array/object, what would be the best way to capture button functionality with some data?

mgascam: I'd suggest to import the mocked data from the authorizations fixture. In `client/data/authorizations/resolvers.ts`:

```
+import authorizationsFixture from '../../data/authorizations/test/authorizations.fixture.json';
 
export function* getAuthorizations( query: Query ): Generator< unknown > {
-       yield updateAuthorizations( query, [] );
+       yield updateAuthorizations( query, authorizationsFixture );
}
```

includes/admin/class-wc-rest-payments-orders-controller.php
kalessil: > Does this answer your question or am I missing something?

Yes.

> Please let me know what you think when you have a moment.

My main concern remains same: IPP capturing logic (incl. preparing the order for capture) accumulated many edge cases and corrections along the way. IMO the new endpoint should be implemented from scratch and use `\WC_Payment_Gateway_WCPay::capture_charge`.

If hooks allow properly capture payment via admin, the simplest approach for the new endpoint would be have basic validation followed by `\WC_Payment_Gateway_WCPay::capture_charge` call.

Did you try this approach and evaluate if the order meta gets properly populated with this approach?

mgascam: > Did you try this approach and evaluate if the order meta gets properly populated with this approach?

Thanks for the clarification. I'll try your suggestion and will follow up with my findings.

mgascam: @kalessil turns out that you were right. I reverted the changes to the `capture_terminal_payment` endpoint in [373f9d0](https://github.com/Automattic/woocommerce-payments/pull/4676/commits/373f9d0dabcd4a27363808174c40e2d6677b35e3).

> If hooks allow properly capture payment via admin, the simplest approach for the new endpoint would be have basic validation followed by \WC_Payment_Gateway_WCPay::capture_charge call.

I also followed your suggestion in the `capture_authorization` endpoint. I kept the same validations that the `capture_terminal_payment` uses, but omitted the call to `attach_intent_info_to_order` because is not needed according to my tests. Finally, I do call `mark_payment_capture_completed`, instead of `update_order_status_from_intent` to update the status of the order.

client/types/authorizations.d.ts
```diff
@@ -5,12 +5,14 @@ export interface Authorization {
 	authorization_id: string;
 	authorized_on: string;
 	capture_by: string;
+	captured: boolean;
 	order: OrderDetails;
-	risk_level: RiskLevel;
+	risk_level: RiskLevel | string;
```
kalessil, line 9: Looks excessive, what is the reason for this change?

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -199,6 +213,78 @@ public function capture_terminal_payment( $request ) {
...
+				return new WP_Error( 'wcpay_intent_order_mismatch', __( 'The payment cannot be captured', 'woocommerce-payments' ), [ 'status' => 409 ] );
+			}
+			if ( ! in_array( $intent->get_status(), [ 'processing', 'requires_capture', 'succeeded' ], true ) ) {
+				return new WP_Error( 'wcpay_payment_uncapturable', __( 'The payment cannot be captured', 'woocommerce-payments' ), [ 'status' => 409 ] );
+			}
+
+			$result = $this->gateway->capture_charge( $order, false );
+
+			if ( 'succeeded' !== $result['status'] ) {
+				$http_code = $result['http_code'] ?? 502;
```
kalessil: Can we inline the variables' value?

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -199,6 +213,78 @@ public function capture_terminal_payment( $request ) {
...
+						__( 'Payment capture failed to complete with the following message: %s', 'woocommerce-payments' ),
+						$result['message'] ?? __( 'Unknown error', 'woocommerce-payments' )
+					),
+					[ 'status' => $http_code ]
+				);
+			}
+
+			// Actualize order status.
+			$charge    = $intent->get_charge();
+			$charge_id = $charge ? $charge->get_id() : null;
```
kalessil: Hm, I believe we should have a charge available at that point. 
Also looking into `mark_payment_capture_completed`, it's expected `string`.

Can you explore a little bit deeper this, please?

client/transactions/uncaptured/index.tsx
```diff
@@ -112,7 +120,7 @@ export const AuthorizationsList = (): JSX.Element => {
 
 	const { authorizations, isLoading } = useAuthorizations( getQuery() );
 
-	const getRiskColor = ( risk: RiskLevel ) => {
+	const getRiskColor = ( risk: RiskLevel | string ) => {
```
kalessil: Also looks excessive.

client/data/authorizations/actions.ts
```diff
@@ -39,3 +54,58 @@ export function updateAuthorizationsSummary(
...
+
+		let authorization = yield apiFetch( {
+			path: `/wc/v3/payments/orders/${ orderId }/capture_authorization`,
+			method: 'post',
+			data: {
+				payment_intent_id: paymentIntentId,
+			},
+		} );
+
+		authorization = {
```
kalessil: We should add a note to remove this once we have live data.

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -76,6 +76,20 @@ public function register_routes() {
 				],
 			]
 		);
+		register_rest_route(
+			$this->namespace,
+			$this->rest_base . '/(?P<order_id>\w+)/capture_authorization',
```
kalessil, line 81: Also, it'd be great to have it documented under https://github.com/Automattic/woocommerce-payments/blob/develop/docs/rest-api/source/includes/wp-api-v3/order.md

mgascam:
Closing this PR because it grew too big and we decided to split it. Please check [here](https://github.com/Automattic/woocommerce-payments/pull/4962) for more details.

PR #4677, opened 2022-09-01
Refactor the SQL query modifications in Analytics

### LINKED ISSUE #4632: Refactor the queries in Analytics that use post meta queries
### Description

This issue is to look at the direct postmeta queries in `Analytics.php`  and replace it with `wc_get_orders()`/`WC_Order_Query`.

#### Dev notes
There's a wiki on how to use wc_get_orders() with custom meta [here](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#adding-custom-parameter-support), however I don't know how that DB query will translate to using something like wc_get_orders().

#### Additional context
Project thread: pdjTHR-Xp-p2
Exploration p2: pdjTHR-YX-p2

RadoslavGeorgiev:
Fixes #4632

Rework `WCPay\MultiCurrency\Analytics` in order to support Custom Order Tables.

### What does the class do?

@dmallory42 described what the class does in https://github.com/Automattic/woocommerce-payments/issues/4632#issuecomment-1232681417. Here is the summary:

- The class modifies the queries, made by WooCommerce Analytics to adjust for order's currencies.
- WC Analytics has custom queries, so neither `get_posts` nor `wc_get_orders` are used or modified here.
- The additional data, needed to adjust the queries, is stored as order meta.

### Analytics and Custom Order Tables

I didn't know whether Analytics should work with COT, so I asked in a couple of places in Slack without getting a clear answer:

1. p1662026947873699-slack-C03TN23TZB5
2. p1662026407137989-slack-C9QDV4TL4

All orders (custom order table and not) on my local instance were already deleted so I could have a clean slate, and begin fixing the issue with order data contained only in custom order tables, to make sure that Analytics are not relying on missing data. This could have been a bad decision.

After creating a few orders, and experimenting with [Analytics' settings](http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Fsettings), mainly `Import Historical Data`, I could not get those orders to show up in Analytics.

My conclusion is that the only way to test this (PR) right now is to:

1. Enable custom order tables, and choose to use WooCommerce Orders tables, and to keep synchronization on in [custom data stores section](http://localhost:8082/wp-admin/admin.php?page=wc-settings&tab=advanced&section=custom_data_stores).
2. Create a few orders, which will store the data both in custom order tables, as well as posts.
3. Modify the code to support reading from both datastores, **while ensuring that in COT mode we are not accessing `wp_posts` and `wp_postmeta` in any way**, and vice-versa. See the next point.

### The changes

In _What does the class do?_ I already mentioned that all of the required additional data is stored as order meta. After looking at the class, I realized that it's job is primarily to `LEFT JOIN` the `wp_postmeta` table a few times, and to modify the rest of the query accordingly. <-- @dmallory42 could you validate this assumption?

Within this PR I use `OrderUtil::custom_orders_table_usage_is_enabled()`, and if enabled, I switch from joining `wp_postmeta` to joining `wp_wc_orders_meta` while using `order_id` as a key, instead of `post_id`. Both meta tables have the same structure, except for the name of the foreign key column, so this work.

As an additional modification, I changed the aliases of those tables within the query to just use `meta` instead of `postmeta` for clarity.

## Testing instructions

1. Set up COT to use the orders table, and keep sync on.
2. Make sure the posts and orders tables are synchronized (WooCommerce > Status > Action Scheduler > Pending > Run all jobs).
3. Make sure you have Multi-Currency set up, and at least one order, which was paid in a non-default currency.
4. Open an Analytics page, any of them should work, but I prefer the overview.
5. Clone the tab (in order to keep this one for reference).
6. **Only then check out this branch**
7. Refresh the cloned tab and **ensure that nothing changes**.
8. Inspect the performed SQL query/queries, and make sure that all of them join the `wp_wc_orders_meta` table, and `wp_postmeta` is nowehere in sight.

### How to inspect the performed SQL queries?

This one is a bit tricky:

- I like using [Query Monitor](https://wordpress.org/plugins/query-monitor/), but because all of the data is loaded through AJAX calls, those queries do not appear there.
- My initial choice was to use XDebug, and pause somewhere in the code to inspect the SQL, but somehow it got truncated all of the time...

I ended up modifying `Automattic\WooCommerce\Admin\API\Reports\Orders\Stats\DataStore::get_data()` (source [here](https://github.com/woocommerce/woocommerce/blob/cot-main/plugins/woocommerce/src/Admin/API/Reports/Orders/Stats/DataStore.php)). I replaced this piece of code:

https://github.com/woocommerce/woocommerce/blob/9b3aee30f3185455af09db7c172eabf7ac49b644/plugins/woocommerce/src/Admin/API/Reports/Orders/Stats/DataStore.php#L330-L333

with:

```php
$query_statement = $this->total_query->get_query_statement();
$file = fopen( '/var/www/html/analytics-query.sql', 'a' );
fwrite( $file, $query_statement . "\n\n" );
fclose( $file );
$totals = $wpdb->get_results(
	$query_statement,
	ARRAY_A
); // phpcs:ignore cache ok, DB call ok, unprepared SQL ok.
```

Then I reloaded the page, and checked out `docker/wordpress/analytics-query.sql` in my editor. It might also help to empty the file before reloading the page.

-------------------


dmallory42:
Thanks for putting together this PR! I should be able to review it by the end of this week.

> I didn't know whether Analytics should work with COT, so I asked in a couple of places in Slack without getting a clear answer:

I'm a bit concerned about this, I know it's not directly related to the PR, but analytics compatibility seems like an important consideration for COT. I see we have a GlobalStep issue [here](https://github.com/woocommerce/woocommerce/issues/34287) - I was wondering if you think I should raise this on the project P2 thread, since I can't see a mention of analytics there already?

>  I realized that it's job is primarily to LEFT JOIN the wp_postmeta table a few times, and to modify the rest of the query accordingly. <-- @dmallory42 could you validate this assumption?

Yes, that's pretty much correct. The other thing it does primarily is that in the `set_sql_replacements` function it will replace the default analytics `SELECT` statement with a set of `CASE WHEN` statements which will either use a converted rate, or the default rate, depending on whether the order is MultiCurrency or not.

RadoslavGeorgiev:
> I was wondering if you think I should raise this on the project P2 thread, since I can't see a mention of analytics there already?

I'll give it a few more hours, and if there is no response in Slack by tomorrow morning, I'll raise my concern.

> Yes, that's pretty much correct. The other thing it does primarily is that in the set_sql_replacements function it will replace the default analytics SELECT statement with a set of CASE WHEN statements which will either use a converted rate, or the default rate, depending on whether the order is MultiCurrency or not.

The second part (`CASE WHEN`) did not need any modifications outside of renaming the aliases, which is why I missed it, but that's exactly how I understood it.

includes/multi-currency/Analytics.php
```diff
@@ -263,10 +264,10 @@ public function filter_select_clauses( array $clauses, $context ): array {
 		}
 
 		if ( $this->is_supported_context( $context ) && ( in_array( $context_page, self::SUPPORTED_CONTEXTS, true ) || $this->is_order_stats_table_used_in_clauses( $clauses ) ) ) {
-			$new_clauses[] = ', wcpay_multicurrency_currency_postmeta.meta_value AS order_currency';
-			$new_clauses[] = ', wcpay_multicurrency_default_currency_postmeta.meta_value AS order_default_currency';
-			$new_clauses[] = ', wcpay_multicurrency_exchange_rate_postmeta.meta_value AS exchange_rate';
-			$new_clauses[] = ', wcpay_multicurrency_stripe_exchange_rate_postmeta.meta_value AS stripe_exchange_rate';
+			$new_clauses[] = ', wcpay_multicurrency_currency_meta.meta_value AS order_currency';
```
deepakpathania: Will this vary based on the COT check as well like we did the `meta_table` below?

deepakpathania (commented):
Left one clarifying comment. Tests well and changes look good otherwise.

includes/multi-currency/Analytics.php
dmallory42: Unless I've misunderstood the code, I don't think it would need to, because these are just alises we assign to the tables in `filter_join_clauses`. 

These would be the table aliases regardless of whether COT is enabled or not.

RadoslavGeorgiev: Exactly. All meta tables in WordPress follow the same structure, the only difference being their names and the name of the foreign key column (ex. `post_id`, `comment_id`, `order_id`, etc.), so that they can be used with [`WP_Meta_Query`](https://developer.wordpress.org/reference/classes/wp_meta_query/) afterwards.

In this class we're joining the tables based on the foreign key column (`post_id` or `order_id`), and giving them aliases within `filter_join_clauses`. Afterwards all other queries utilize the same set of fields (`meta_key` and `meta_value`), so there is no need for changes there.

deepakpathania (approved):
Changes LGTM ✅

includes/multi-currency/Analytics.php
```diff
@@ -291,17 +292,26 @@ public function filter_join_clauses( array $clauses, $context ): array {
...
-			$clauses[] = "LEFT JOIN {$wpdb->postmeta} {$currency_tbl} ON {$wpdb->prefix}wc_order_stats.order_id = {$currency_tbl}.post_id AND {$currency_tbl}.meta_key = '_order_currency'";
-			$clauses[] = "LEFT JOIN {$wpdb->postmeta} {$default_currency_tbl} ON {$wpdb->prefix}wc_order_stats.order_id = {$default_currency_tbl}.post_id AND ${default_currency_tbl}.meta_key = '_wcpay_multi_currency_order_default_currency'";
-			$clauses[] = "LEFT JOIN {$wpdb->postmeta} {$exchange_rate_tbl} ON {$wpdb->prefix}wc_order_stats.order_id = {$exchange_rate_tbl}.post_id AND ${exchange_rate_tbl}.meta_key = '_wcpay_multi_currency_order_exchange_rate'";
-			$clauses[] = "LEFT JOIN {$wpdb->postmeta} {$stripe_exchange_rate_tbl} ON {$wpdb->prefix}wc_order_stats.order_id = {$stripe_exchange_rate_tbl}.post_id AND ${stripe_exchange_rate_tbl}.meta_key = '_wcpay_multi_currency_stripe_exchange_rate'";
+			$clauses[] = "LEFT JOIN {$meta_table} {$currency_tbl} ON {$wpdb->prefix}wc_order_stats.order_id = {$currency_tbl}.{$id_field} AND {$currency_tbl}.meta_key = '_order_currency'";
```
dmallory42: This clause will need to be updated because we no longer store `_order_currency` as a meta key in the new COT meta tables. Instead, it is stored in the `wp_wc_orders` table under the key `currency`.

I think we should do something like this:

- If COT is not enabled, use the query as it is currently
- If using COT, instead of joining on `currency_tbl`, do a `LEFT JOIN` on `wp_wc_orders` (joining on the order ID), then update `filter_select_clauses` function to get the `currency` value from this table instead of getting it from the meta table as it does currently.

This value populates the `Customer Currency` field which you can see in `Analytics > Orders`

dmallory42 (changes requested):
Implementation LGTM, when testing I found one issue that we'll need to resolve before :shipping: but other than that it seems to be working well, both when Custom Order Tables are enabled and disabled.

includes/multi-currency/Analytics.php
RadoslavGeorgiev: Great catch, @dmallory42! 🙌 

I changed the column in https://github.com/Automattic/woocommerce-payments/pull/4677/commits/d59a0d83790da30a5160d4b80609609d1550684f to address this. In https://github.com/Automattic/woocommerce-payments/pull/4677/commits/7c6294501882e827a7e28538a81c824bad7ba105 I renamed the table when COT is enabled, because I realized aliasing it as "meta" is not very elegant.

As an additional step, I deleted a couple of orders from `wp_posts` and `wp_postmeta` to make sure that they still appear in Analytics, sourced from the orders table. It worked great! 🎊

includes/multi-currency/Analytics.php
```diff
@@ -320,7 +341,7 @@ public function filter_where_clauses( array $clauses ): array {
 		}
 
 		$prefix       = 'wcpay_multicurrency_';
-		$currency_tbl = $prefix . 'currency_postmeta';
+		$currency_tbl = $prefix . 'currency_meta';
```
dmallory42: One small thing, we need to also do the check to see which table/column to use for the WHERE clause when filtering by currency. I tested this code locally and it will achieve this (replaces lines 344-355). 
```suggestion
		$prefix       = 'wcpay_multicurrency_';
		$currency_tbl = $this->is_cot_enabled()
			? $prefix . 'order_currency'
			: $prefix . 'currency_meta';

		$currency_args = $this->get_customer_currency_args_from_request();
		if ( ! empty( $currency_args['currency_is'] ) ) {
			$currency_is = sprintf( "'%s'", implode( "', '", $currency_args['currency_is'] ) );
			if ( $this->is_cot_enabled() ) {
				$clauses[] = "AND {$currency_tbl}.currency IN ({$currency_is})";
			} else {
				$clauses[] = "AND {$currency_tbl}.meta_value IN ({$currency_is})";
			}

		}

		if ( ! empty( $currency_args['currency_is_not'] ) ) {
			$currency_is_not = sprintf( "'%s'", implode( "', '", $currency_args['currency_is_not'] ) );
			if ( $this->is_cot_enabled() ) {
				$clauses[] = "AND {$currency_tbl}.currency NOT IN ({$currency_is_not})";
			} else {
				$clauses[] = "AND {$currency_tbl}.meta_value NOT IN ({$currency_is_not})";
			}
		}
```

includes/multi-currency/Analytics.php
```diff
@@ -528,4 +551,13 @@ private function set_sql_replacements() {
 			],
 		];
 	}
+
+	/**
+	 * Checks whether Custom Order Tables are enabled.
+	 *
+	 * @return bool
+	 */
+	private function is_cot_enabled() {
```
dmallory42: Nitpick: can we add a return type hint on this function?

dmallory42 (approved):
LGTM! Thanks for working on this!

I tested by making a few orders and syncing them across to the COT, then tried accessing the analytics with COT on/off. The figures looked correct and the extra fields/filters we added all work as expected. 

I added one comment which needs resolving to fix the advanced filters, but I'm going to pre-approve this!

includes/multi-currency/Analytics.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4677/commits/ed2d295ddd4a19369f85f8896e05f09464071527

includes/multi-currency/Analytics.php
RadoslavGeorgiev: Another great catch!

I implemented this change in https://github.com/Automattic/woocommerce-payments/pull/4677/commits/203339e06589c79cb61fed3afe708b02900839c7, but by pre-loading the field instead of adding `if-else` statements.

PR #2429, opened 2021-07-14
UPE Block Checkout

bborman22:
Fixes #2162 

#### Changes proposed in this Pull Request

This PR adds the UPE to the block checkout experience. When UPE is enabled and the store is using the block checkout, UPE will now be loaded into the payment section. The block checkout makes use of the Stripe's React PaymentElement and so needed to update the `@stripe/react-stripe-js` package to the latest (1.4.1) this update also forced an update to the `@stripe/stripe-js` package to avoid any linting issues.

I kept the UPE block registration flow as independent as possible from the CC block registration flow in case there is ever a need to customize it without adding a lot of conditional checks. There is also the new UPE fields component to then use the PaymentElement. 

Some of the more notable details of the UPE fields component flow:
- The first was just that we need to create a payment intention before mounting the PaymentElement. I commented on the code here for any input on a better way to handle this just in case the approach I took isn't optimal. 
- Once mounted, there is an onChange event listener to check for the UPE form being completed.
- Once submitted, the onPaymentProcessing event will fire and check for the completed status and throw an error if not complete.
- In the classic checkout flow, we simulated a confirmPayment call to trigger the field errors. In this implementation I was unable to call the confirmPayment with an empty return_url as it caused it's own error. So the individual field errors will not be displayed unless the user has focused and blurred on them.
- Finally, the onCheckoutAfterProcessingWithSuccess is called after the server is updated. This part was tricky as we need to update the intent in case there was any price changes (different shipping method, coupon applied, etc). To update the intent required the order ID which was not available until this point. This leads to a slightly longer period between submitting the checkout and the final redirection from Stripe.

While working on this, we also got the feature from Stripe to pass in the billing details as needed and so that has been implemented here as part of the PR.

#### Testing instructions

- Start with a store with UPE enabled and Checkout blocks [enabled](https://docs.woocommerce.com/document/using-the-new-block-based-checkout/).
- Have 2 shipping methods, one flat rate (I used $10) and one free shipping.
- Add an item to the cart and select the flat rate shipping option from the cart.
- Confirm checkout loads properly and displays the UPE in the payments area.
- Try to submit without providing any payment details and confirm error is displayed.
- Provide only partial payment information, submit, and confirm error is displayed.
- Change the shipping method to the free shipping option, confirm the checkout total updates to reflect the change.
- Provide full payment information and submit
- Confirm that the checkout is successful 
- Confirm that the order was for the correct total (using free shipping)
[...]

Issue #4672, opened 2022-08-31
Multiple UPE - Explore single payment intent with multiple instances of UPE


harriswong:
### Description
This came up in one of the team chats. Currently, we have a PR (https://github.com/Automattic/woocommerce-payments/pull/3968) to split UPE into multiple UPE. Each split payment method is its own UPE instance.

A question came up: "Have we tried to reuse same payment intent with multiple instances of UPE?"

We would like to investigate/confirm the feasibility of this. Here are the proposed steps to confirm:

1. Create a single payment intent with all payment types allowed on it ( [payment_method_types](https://stripe.com/docs/api/payment_intents/create#create_payment_intent-payment_method_types) arg).
2. Create an instance of UPE with that intent, but pick only a single PM to display on the instance (hope it's possible via config options).
3. Create another instance of UPE with the same intent, but other config options (for another PM)

### Acceptance criteria
Answer the question - "Have we tried to reuse the same payment intent with multiple instances of UPE?"

### Additional context
Check p1661964644907429/1660685029.312559-slack-CGGCLBN58 for more context.
Most likely not possible paJDYF-5dK-p2#comment-13799

FangedParakeet:
> Create an instance of UPE with that intent, but pick only a single PM to display on the instance (hope it's possible via config options).

So it turns out that this is indeed possible, though not exactly in the way that I'd imagined!

By first tweaking the #3968 branch, so that all payment elements would share the single intent--all using the same payment method--I confirmed that it is possible to have multiple payment elements on the page all using the same payment/setup intent. I tried to update the payment intent after mounting, but this had no effect on the mounted payment elements...which proved that the payment element *only presents the payment methods included in the `payment_method_types` field and is unaffected by any changes to the intent after the initial mount*.

This means if we update the value of this field to contain a different, single payment method *before we mount each new payment element*, this will permit us to present us multiple payment elements present on the same checkout, all using the same payment or setup intent, but all displaying different payment methods at the same time! The actual `payment_method_types` value will always refer to the most recently mounted/updated payment method and not necessarily the actual payment method currently selected or used in the final transaction; however, this is not a problem at all, because in every UPE checkout we always update this value to the correct, chosen payment method immediately prior to confirming the payment, so this will always be corrected at the last available opportunity and it appears Stripe's API has no outward offense with us doing this.

[...]

PR #4681, opened 2022-09-01
Integrate split UPE payments with WooCommerce Blocks

### LINKED ISSUE #4603: Multiple UPE - Payment methods in blocks
### Description
This issue is one of the TODOs in the "splitting into multiple UPE" project https://github.com/Automattic/woocommerce-payments/pull/3968

We should check and make sure enabling UPE doesn't break the payment method in blocks (`includes/class-wc-payments-blocks-payment-method.php`).

### Acceptance criteria
WooCommerce Blocks should work with WCPay payment methods.

### Testing instructions
1. Pull this branch https://github.com/Automattic/woocommerce-payments/pull/3968
2. Go to wp-admin > Pages > Create a new page with Block checkout
3. Confirm that the block checkout loads and can render.
image
4. Able to checkout after publishing the page. ie .`localhost/block-checkout`.

harriswong:
Fixes https://github.com/Automattic/woocommerce-payments/issues/4603

Block checkout registers UPE as a payment method (Introduced by https://github.com/Automattic/woocommerce-payments/pull/2429). What this PR attempts to do is to register multiple payment methods instead.

#### Difference in payment methods management in classic checkout vs Blocks 🧐
Classic checkout renders and manages individual payment methods via WC Core and [`woocommerce_payment_gateways` filter](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L365).

WooCommerce Blocks provide the `PaymentMethodRegistry` for the method to be registered. During the payment method registration, we [rollout js scripts and styles](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-blocks-payment-method.php#L17), and this time [the side responsible for the methods rendering is the front-end](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe.js#L43) where blocks give us an assist in the form of [registerPaymentMethod](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/assets/js/blocks-registry/payment-methods/registry.ts#L34-L53).

#### Problem solution 👨🏽‍💻
Having in mind the above, I took the opportunity and used the [`paymentMethodsConfig` field](https://github.com/Automattic/woocommerce-payments/blob/53e37c973707d8b084153655c4d393439c1ec091/includes/payment-methods/class-upe-payment-gateway.php#L717), which comes from the back-end and is exposed in JavaScript. This field represents all the enabled payment methods from WCPay settings. On the front end, [I take all the enabled payment methods and register a payment method for each of them](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-f6bacead8106193d743286ebb7f58e7bad5b05526103306750625a01943cebb5R65). For each payment method, [its name is being provided to the component](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-f6bacead8106193d743286ebb7f58e7bad5b05526103306750625a01943cebb5R70) in order to be used later on in the [effect hook to create a payment intent](https://github.com/Automattic/woocommerce-payments/blob/53e37c973707d8b084153655c4d393439c1ec091/client/checkout/blocks/upe-fields.js#L370-L408), right after opening the payment method on the UI.  

I also [provide testingInstructions](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-f6bacead8106193d743286ebb7f58e7bad5b05526103306750625a01943cebb5R72) for each particular payment method to be displayed on the UI, if we are in the Test mode.

#### How the basic Card Element replaced the card Stripe payment element 
The idea of replacing the card payment method with the legacy card is pretty straightforward - [ignore the card payment method while attaching all the enabled UPE payment methods](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-b541c0a287f40f82a1d0e43c52c7c7a2b67ae233f261655cbb437726fec83feeR1071-R1073) and attach the legacy card separately. The same idea is applied in the classic split UPE checkout. My first thought was to update the [UPE implementation](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe.js) to somehow be able to inject a legacy card there. But then I thought - we already have [the fully working separate component with the working legacy card](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/index.js), which is used when UPE aren't enabled.

Finally, what I did is I rolled out two components [here](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-c5f750f16fa665d48bb6ccb2b37c5d3f10d80e18158ce7419555aa3ffd1cf0e6R17-R29) - the legacy card and UPE payment elements without the card. This approach helped us to reduce the amount of global refactoring and allowed us to roll back the solution very quickly in case we will ever need to take a new path.

#### Testing instructions
1. Go to the `wp-admin` console and create a new page. Search for the checkout block and add it to the page. 
image
2. Go add some items to your cart.
3. Go to your new block checkout page (ie. http://localhost:8082/block-checkout/)
4. You should see all the enabled UPE payment methods render.

image

5. Go to `Payments` -> `Settings` and disable some of the payment methods from the list. Then, go to the blocks checkout page and confirm, that the rendered methods list reflects only the enabled payment methods. 
6. Confirm, that there are test messages appearing for Card and SEPA payment.
7. Make sure the confirmed order is on the Orders list in your WooCommerce console in the admin account. Check transaction details.

image
image

8. Perform regression tests with non-UPE checkout and confirm, that card payment works as expected.

-------------------

**Post merge**


harriswong:
**Update**

- Nonce for these individual UPE payment method may need work https://github.com/Automattic/woocommerce-payments/blob/5b0c660fb911c9f2539fd4c934039bce030ff55c/includes/payment-methods/class-upe-payment-gateway.php#L237. Otherwise, you may see this error: 
image

- The payment method title (screenshot from description) needs to reflect the actual payment method.
- I only created 4 payment methods to test the idea (card, bancontact, SEPA, Giropay). If creating classes to `extends WC_Payments_UPE_Blocks_Payment_Method` is the right approach, then we will need to create all the payment method classes.
- Test broke. The UPE block checkout E2E tests rely on 1 UPE instance. We'll need to fix that since we now have multiple UPE instances.
- If SEPA is disabled from UPE payment settings, then it shouldn't show here. Right now, I loop and display all of them.

timur27:
@harriswong I noticed one more thing - currently, with blocks, we are rendering the UPE credit card (at the bottom in the first screenshot) instead of the legacy credit card that uses Stripe Elements (the top one):

image
image


Whereas during the classic checkout with split UPE turned on, we're rendering the legacy credit card. There is some background of why it looks like that on the classic checkout in [the description of this PR](https://github.com/Automattic/woocommerce-payments/pull/3968). My first thought was that we should align them and make blocks checkout behave the same as classic checkout. Do you agree with this? If yes, I'll cover this within this issue as well.

harriswong:
@timur27 Thanks for the note! Yes, I think making both behaviours the same would be good. This way, both will use the classic checkout for cards, and UPE elements for other payment methods. That said, the card classic checkout doesn't seem to work with Link, so [whatever fix we decided upon](https://github.com/Automattic/woocommerce-payments/issues/4720) will have to be applied to both in the future.

Please check with @FangedParakeet though.

FangedParakeet:
@timur27, yeah, the Blocks element should use the legacy payment element for the card payment method, but the UPE element for all other payment methods. This should align with how the classic checkout functions in the split-UPE sanctum.

The purpose for doing this is because the UPE is currently incompatible with WooPay, so this split-UPE approach will allow us to hurdle this obstacle by using the compatible legacy element for the WooPay card element alone, while using the UPE for all other payment methods. 

I am honestly not sure how Link will eventually play into this (sorry lol), but there's more discussion on this on #4720.

timur27:
@FangedParakeet today, I found one inconsistency between classic UPE checkout vs. the blocks one, and with this message, I'd like to clarify the performance expectations and if there is a need for alignment between these two. 

### How the classic split UPE checkout creates payment intents
As we know, the classic split UPE checkout is implemented in a way that `wcpay_create_payment_intent_{payment_method}` async AJAX calls are sent all at once during the page load when mounting UPE elements, providing our users a seamless experience when using payment methods since each method is already loaded and opens immediately whenever used. 

### How the blocks split UPE checkout creates payment intents
With woocommerce-blocks, things are implemented a bit differently, and I noticed a dissimilarity in the strategy of sending the requests to create a payment intent. We have [the effect hook](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe-fields.js#L368-L400), which is responsible for creating a payment intent **after rendering the component** (unless `paymentIntentId, hasRequestedIntent, API, errorMessage, appearance` fields didn't change, of course). 

At the same time, the [registerPaymentMethod](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe.js#L43) function from woocommerce-blocks creates a payment method config within blocks **but does not render the payment method yet**. The rendering occurs when users open a particular payment method. Effectively, a payment intent is created only for a specific form of payment when its block component is chosen from the list.

### What I'd like to clarify
While there should be a way for Blocks implementation to create payment intents on the page load, I wanted to confirm the expectations. Should each payment method with blocks open immediately like in the classic checkout, or the loading time due to the intent creation API call for each method is acceptable in Blocks? Alternatively, maybe at this point, the most important thing for us is to make the Blocks checkout work, and we will get back to the performance later. These are the questions I don't have an answer to yet. 😊

Thank you for clarifying all this!

FangedParakeet:
> Should each payment method with blocks open immediately like in the classic checkout, or the loading time due to the intent creation API call for each method is acceptable in Blocks?

This is actually quite a dense philosophical answer and one I don't entirely have an answer for. You are correct in that the classic checkout creates payment intents for _all_ payment method types immediately on page load. This is "seamless", such that the end user will not have to wait for a new element to load while changing payment methods at checkout; however, this isn't tremendously efficient. We spend background resources to make these requests and moreover we create lots of additional perhaps superfluous requests to Stripe. 

We create a new payment intent for _every_ enabled payment method, irregardless of whether every payment method is even selected at checkout. An end user might never select a payment method and never see its payment element, but we will still trouble Stripe to create an intent and the front-end resources required to support this unseen element. These additional requests and unused payment intents might look a little suspicious from Stripe's perspective and is usually a symptom of some spammy sickness on a site's checkout.

It is designed this way not by some enlightened decision making, but simply because I understandably chose the simplest path when writing the initial PoC. Personally, I think the latency to mount a new payment element (create intent, create element using intent, and mount it to DOM) is generally a relatively negligible, constant-time operation, whereas the negatives of avoiding this scenario by preloading all payment elements is rather noticeable (multiplying our requests to Stripe, multiplying unused payment intents, etc.). So I would lean towards refactoring our existing code to only prepare and mount UPE elements when they are required, as opposed to loading them all on page load. 

> Alternatively, maybe at this point, the most important thing for us is to make the Blocks checkout work, and we will get back to the performance later. 

That being said, I'd still hold this goal in higher regard, so for now I think following the shortest path is the most sensible option. It's worth remaining aware that once Stripe finally drops their new UPE implementation, this entire debate will be nullified, since we will no longer require a payment intent in order to present a UPE payment element to our customers. So I feel that perhaps expending too much effort on the most optimal solution in this case will be somewhat in vain, since this will not be a long-term concern for us.

Anyways, I'd vote for now on finding the simplest solution for Blocks right now and since that also happens to involve delaying payment intent creation until the moment is unavoidable, I'm even happier with choosing this path forwards. @mdmoore, we've chatted about this a few times here and there: let me know if you have any contradictory thoughts on this subject--or even better, if you happen to agree with me! 😁

mdmoore:
>I'd vote for now on finding the simplest solution for Blocks right now and since that also happens to involve delaying payment intent creation until the moment is unavoidable, I'm even happier with choosing this path forwards.

I agree completely. It might be worth sharing your findings in [#4672](https://github.com/Automattic/woocommerce-payments/issues/4672#issuecomment-1266273803) here. Hopefully we'll ultimately move to the same approach in classic checkout as well and keep the intent history tidy. If blocks checkout is giving us an assist on that, I'd say go with it!

client/checkout/blocks/upe.js
```diff
@@ -13,53 +13,89 @@ import {
...
+	sofort: PAYMENT_METHOD_NAME_SOFORT,
+};
 
-const paymentMethodsConfig = getConfig( 'paymentMethodsConfig' );
+const enabledPaymentMethodsConfig = getUPEConfig( 'paymentMethodsConfig' );
 const isStripeLinkEnabled =
-	paymentMethodsConfig.link !== undefined &&
-	paymentMethodsConfig.card !== undefined;
+	enabledPaymentMethodsConfig.link !== undefined &&
+	enabledPaymentMethodsConfig.card !== undefined;
```
FangedParakeet, line 50: Currently, this value of `isStripeLinkEnabled` will always be false, no? Since inside the `paymentMethodsConfig` the `card` payment method is skipped for the UPE config, which means that it will always be `undefined`. 

I think this makes sense though, since Link should only be enabled on the UPE card payment method, which will no longer exist (for now). This situation will probably be resolved somehow in #4720, so I wouldn't mind if you left it this way for now, as opposed to explicitly disabling Stripe Link (setting `isStripeLinkEnabled` to `false`).

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -99,7 +99,7 @@ public function __construct(
 		$this->icon               = $payment_method->get_icon();
 
 		add_action( "wc_ajax_wcpay_create_payment_intent_$this->stripe_id", [ $this, 'create_payment_intent_ajax' ] );
-		add_action( "wc_ajax_wcpay_update_payment_inten_ $this->stripe_id", [ $this, 'update_payment_intent_ajax' ] );
```
FangedParakeet, line 102: 🤦 🔍

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1069,12 +1069,23 @@ private function get_enabled_payment_method_config() {
...
+			}
 			$payment_method                 = $this->wc_payments_get_payment_method_by_id( $payment_method_id );
 			$settings[ $payment_method_id ] = [
 				'isReusable'           => $payment_method->is_reusable(),
 				'title'                => $payment_method->get_title(),
 				'upePaymentIntentData' => $this->get_payment_intent_data_from_session( $payment_method_id ),
 				'upeSetupIntentData'   => $this->get_setup_intent_data_from_session( $payment_method_id ),
+				'testingInstructions'  => WC_Payments_Utils::esc_interpolated_html(
+					/* translators: link to Stripe testing page */
+					$payment_method->get_testing_instructions(),
```
FangedParakeet, line 1106: Should we maybe return an empty string `''` instead of `false` for payment methods that do not have testing instructions? It appears as if this is working _fine_, but this function definitely seems to be expecting a string. Not a particularly big deal, so I'll leave it up to you!

client/checkout/blocks/upe-fields.js
```diff
@@ -381,10 +383,9 @@ const ConsumableWCPayFields = ( { api, ...props } ) => {
 		if ( paymentIntentId || hasRequestedIntent ) {
 			return;
 		}
-
-		async function createIntent() {
+		async function createIntent( paymentMethodId ) {
 			try {
-				const response = await api.createIntent();
+				const response = await api.createIntent( paymentMethodId );
```
FangedParakeet, line 394: In the classic shortcode checkout, before creating the intent, we first check to see if [there is a cached payment intent on the current session](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L231-L233). There should be a unique cache key for each payment method type at checkout, to mitigate the multiplications of payment intent requests that we make, ensuring that we will create at most a single payment intent at checkout for each UPE payment gateway.

In this blocks implementation, we don't appear to be doing this, which is exacerbated by the fact that every time we select a new radio button gateway at checkout, the _entire_ gateway re-renders, recreating and remounting a new payment element, _and_ creating a brand new payment intent as well! I found that if I simply switched back and forth between UPE gateways at checkout, watching my network tab, I could see that we created a brand new payment intent each time the element re-rendered.

I believe you should be able to pull the cached intent from the same UPE `paymentMethodsConfig` object that you're using here ([here's](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/includes/payment-methods/class-upe-payment-gateway.php#L1079-L1080) where that cached data gets stored in the config). Let me know if you think that'd be an easy update or if you tried it already and something unfortunate and troubling happened instead! 😬

FangedParakeet (commented):
This issue is probably the largest one we have remaining and the one I've been keeping the most distance from, so awesome work, @timur27, for tackling it without breaking a sweat!

Generally everything here is working even better than expected! I was able to do pretty muche everything, except for completing a checkout with a saved UPE payment method. However, I think this may be related to #4530, so I'm happy to ignore this issue.

Here are a few oddities that I discovered.

### Invisible notices 

I'm not sure why, but every time I managed to coax an error out of the checkout, I found a blank notice at the top of the page.

Mysterious invisible error notice
_Mysterious invisible error notice example_

By using the Redux debug tools, I could find each missing error notice stored correctly in the state, but for some reason this is not being communicated to the front-end. Let me know if you're able to replicate this as well!

### Payment gateway re-renders

I left another comment related to this issue, but essentially, whenever we select a new UPE payment gateway at checkout, the entire element re-renders and we remount a new payment element. I see that we discussed this earlier, but I didn't realise that the entire element still re-renders even if the select payment gateway has been selected before. 

This is a little irritating, but not a goliath gaffe. However, I would maybe suggest that we might want to block the checkout--or at least the Place Order button--while the element is rendering, if possible. Several of the redirection UPE payment gateways do not require any additional payment information and when one of these gateways is selected, it isn't necessarily clear that anything is loading. 

If you submit the form while the payment element is mounting, you are greeted with a rather strange error message. I had to sift through the Redux state to find it, due to the former issue, but I believe the error message states, "Invalid payment method. Please input a new card number" when the currently selected payment gateway is not the card gateway. I'm not entirely sure what's going on here, but if it's possible to block the "Place Order" button during this loading time, I think that would negate this unkindness. 

Those are the only issues I could consistently replicate with these changes, so top work once more! Let me know your thoughts, as I wouldn't be opposed to opening new issues to address these challenges, if you'd like to move on from this current issue, as I do think you have accomplished the core heart of this task successfully.

timur27:
> I'm not sure why, but every time I managed to coax an error out of the checkout, I found a blank notice at the top of the page.

@FangedParakeet, this is a valid observation, thank you for bringing this up! 🙌🏼 I spent some time trying to find the root cause since I could reproduce it on the non-UPE blocks checkout and on the `develop` branch. 

Finally, I discovered that the `woocommerce-blocks` plugin unexpectedly handles the error, probably [in this area](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/assets/js/base/context/providers/cart-checkout/checkout-processor.js#L234-L281). I [reached out Team Rubik for a confidence check](https://a8c.slack.com/archives/C8X6Q7XQU/p1667307161395229), if my assumptions are correct or not.  It turned out that there was [the issue](https://github.com/woocommerce/woocommerce-blocks/pull/7231) resolved some time ago, which fixes the mysterious invisible error notice starting from version `8.7.0` of Woocommerce Blocks.

After updating the plugin on my local WCPay environment, I confirmed that the issue is not appearing anymore.

FangedParakeet:
> It turned out that there was https://github.com/woocommerce/woocommerce-blocks/pull/7231 resolved some time ago, which fixes the mysterious invisible error notice starting from version 8.7.0 of Woocommerce Blocks.

Nice one! Can confirm that upgrading the Blocks plugin has now resolved this issue for me. 💪 Ma b.

client/checkout/blocks/upe-fields.js
timur27: @FangedParakeet I did some minor research to find answers to our questions, which is why this comment came out a bit later. There is a slight difference in Stripe elements management between the classic and blocks checkout.

### Classic checkout
Stripe elements mounting [on the ](https://github.com/Automattic/woocommerce-payments/blob/88be4a965c11235ab78709577eba83967a48f09d/client/checkout/classic/upe.js#L352-L376)classic checkout page occur only once on the updated_checkout[ event](https://github.com/Automattic/woocommerce-payments/blob/88be4a965c11235ab78709577eba83967a48f09d/client/checkout/classic/upe.js#L352-L376). At this stage, the state is examined to confirm whether the UPE element was already mounted. If it was mounted, we ensure that the UI focuses on the element. If it wasn’t, we would utilize the mountUPEElement function. 

The mountUPEElement function, among others, [checks if there is no payment intent data related to the particular method in the session storage](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L231-L233), as you have pointed out before. If there is nothing in the session storage, we contribute to creating a new payment intent with the help of the createIntent Stripe API method. 

The key point here to pay attention to is that [the attempt to fetch the payment intent data from the session](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L231-L233) is not executed when users are switching back and forth between UPE gateways at the checkout page. Instead, it is executed when the page is refreshed. 

In other words, assuming that our session storage is empty (site cookies removed) when opening the classic checkout page for the first time with seven different UPE payment methods turned on, we will send seven requests to create payment intents since there is nothing in the session storage yet. During the `create_payment_intent_ajax()`  execution, the storage will be supplemented with intents for each payment method, respectively, allowing for the session storage to be successfully used when users refresh the page or return to the checkout page after traveling to other corners of our product. 

But, the lack of UPE elements loading while switching between radio buttons of different payment methods is explained by the fact that the component is already mounted to the DOM.

### Blocks checkout
Based on the previous research of the classic checkout implementation, we can now bring more light on how switching between different payment methods on the Blocks checkout page works. By leveraging the session storage, we will cut down the amount of new unnecessary payment intents pretty decently since, on each page refresh, no further payment intent will be created, since the existing intent will be taken from the storage. However, the use case of switching back and forth between methods while staying on the page cannot be improved with the help of this mechanism. Well, after the refresh, when we will leverage the session storage, users can switch between methods seamlessly, but only after the page refresh. 

Due to the nature of `registerPaymentMethod`, we treat each Stripe element as a separate component and render them each time we click on them.

Having regard to the small study from above, I made the following amendment:
* Leverage the session storage to check if there is a payment intent already which can be used. This improves the performance after the page refresh, but only. (d81847205bcfc1e92dd54750559a527fdd368690).

The idea of dealing with the problem of switching elements on the page came to my mind just now, which is based on invoking registerPaymentMethod only once as opposed to [doing it in a loop](https://github.com/Automattic/woocommerce-payments/pull/4681/files#diff-f6bacead8106193d743286ebb7f58e7bad5b05526103306750625a01943cebb5R65-R66), providing the enabledPaymentMethodsConfig to it, and setting up all the enabled payment methods within the one component. 

However, my little hunch here is that we can face integration problems with WooCommerce blocks, both in terms of displaying multiple payment methods as one and supporting the solution in the future without a big headache. In terms of the implementation efforts, I wouldn’t say it is a thing for 8 story points, but the whole `upe-fields.js` file needs to be reviewed, and some adjustments need to be applied. Let me know what you think about all this! 🙌🏼

includes/payment-methods/class-upe-payment-gateway.php
timur27: Fully agree with you! Fixed in https://github.com/Automattic/woocommerce-payments/pull/4681/commits/7dd6bfe783f9b8556710533e1df893c0fd2dc8d0.

client/checkout/blocks/upe.js
timur27: Yes, leaving it as it is for now since it will be impossible to bypass this part of the code while making Link work.

client/checkout/blocks/upe-fields.js
FangedParakeet: > The mountUPEElement function, among others, [checks if there is no payment intent data related to the particular method in the session storage](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js?rgh-link-date=2022-11-02T13%3A42%3A34Z#L231-L233), as you have pointed out before. If there is nothing in the session storage, we contribute to creating a new payment intent with the help of the createIntent Stripe API method.

Just FYI, the lines that you linked to [here](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js?rgh-link-date=2022-11-02T13%3A42%3A34Z#L231-L233) aren't related to the intent session storage, but actually the [Pay for Order checkout page](https://woocommerce.com/document/managing-orders/#paying-for-an-order). When a customer pays for an order via this page, there is no cart or checkout session, but the order ID is present as a GET parameter, so we have to use this to look up the order and collect all billable fields from there. 

[This bit](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L235-L240) is where we check to see if there is any session data present and only create a new payment intent [if there is no intent data found](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L245)...but I'm pretty sure you knew this already, because it looks like that's basically how you implemented this--so I'll give you the benefit of the doubt and presume that you linked the order ID bit accidentally. Though, this likely redundant explanation is here just in case it was no accident. 😅 

> But, the lack of UPE elements loading while switching between radio buttons of different payment methods is explained by the fact that the component is already mounted to the DOM.

👍 

> However, the use case of switching back and forth between methods while staying on the page cannot be improved with the help of this mechanism.

If there were some way to persist the payment intent after it is created either on the state or by making a request to the backend to store it in the transient cache on the backend, that'd be grand. I don't mind (well, I don't mind so much) that the UPE is remounted every time we switch between payment methods at checkout, but in my testing, it appears that currently a new payment intent is created every time this switch is made! 

I did notice that this is no longer the case, after a page refresh, as you described. However, it would still be really cool if we could figure out a way to update the transient after a payment intent is created (this could be a quick async AJAX call, for example).

> However, my little hunch here is that we can face integration problems with WooCommerce blocks, both in terms of displaying multiple payment methods as one and supporting the solution in the future without a big headache.

🎉 

> In terms of the implementation efforts, I wouldn’t say it is a thing for 8 story points, but the whole upe-fields.js file needs to be reviewed, and some adjustments need to be applied.

Hah, these points are a little arbitrary and that 8 point estimate includes all the work that has already been expended into this issue. That being said, I had a quick scan through your updates and a very quick test running through the checkout flow...and everything looks pretty good to me right now!

I'd rate the basic split-UPE blocks flow as functioning pretty successfully here and would be happy solving how we can be more economical with our intents in a separate issue. Let me know how that sounds or if you have another plan in mind! I think I'd be ready to approve this PR, if you're happy with it, at the earliest convenience (which will probably be Monday, because I'm AFK tomorrow (soz 😞 )). LMK how U feel pls. Cheers!

timur27: @FangedParakeet a simple one-liner (https://github.com/Automattic/woocommerce-payments/pull/4681/commits/f208c9edd9a9c081e1e682a2ae1d00d3251e8c55) did the trick, and we now create a payment intent only once. Please let me know if it works for you! 

It seems that under the influence of the many elements and mechanisms that make up the split UPE, I was so fascinated that I forgot about this simple solution.

changelog/fix-4603-block-split-upe
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Integrate split UPE payments with WooCommerce Blocks
```
FangedParakeet: By the way, I'm not sure if we need to include a changelog entry while we are still just merging to a feature branch. Presumably we'd wait until we merge the whole thing into `develop` before making an addition to the changelog. 

The [doc on contributing to WCPay](https://github.com/Automattic/woocommerce-payments/blob/develop/CONTRIBUTING.md) describes when changelog are not required: "A special case is functionality hidden behind feature flags: such features should be mentioned when released." Split-UPE isn't behind a feature flag, but it has not yet been released, so right now this changelog is a little less meaningful...but let me know if you disagree with that interpretation!

client/checkout/utils/upe.js
```diff
@@ -36,3 +36,31 @@ export const isWCPayChosen = function () {
...
+/**
+ * Returns the cached payment intent for the current cart state.
+ *
+ * @param {Object} paymentMethodsConfig Array of configs for payment methods.
+ * @param {string} paymentMethodType Type of the payment method.
+ * @return {Object} The intent id and client secret required for mounting the UPE element.
+ */
+export const getPaymentIntentFromSession = (
+	paymentMethodsConfig,
+	paymentMethodType
```
FangedParakeet, line 49: Since you've just updated this function signature, can you please pass the `paymentMethodsConfig` parameter through when this function is called in the classic checkout over [here](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L706)? I think this currently breaks the classic checkout, but should be an easy fix!

FangedParakeet (changes requested):
Thanks for responding to all of my feedback from the last time! Don't worry, I only have a little more this time. 😅 

Basically, everything is working as expected! The regular Blocks checkout flow, excluding saved payment methods, is working happily. Moreover the loading time while switching between payment methods at checkout is much healthier now that we persist the payment intent for the entire duration of a checkout. Great work!

However, sadly I discovered that the regular shortcode UPE checkout flow is now unfortunately broken and returns the following error consistently.

```
paymentMethodsConfig[paymentMethodType] is undefined
```

However, I've left another comment that explains why I'm very confident this is happening (you need to fix the parameters used on [this function call](https://github.com/Automattic/woocommerce-payments/blob/fix/4603-block-split-upe/client/checkout/classic/upe.js#L706)), which should hopefully be a very easy amend. Please ping me again once you've fixed that and I don't think I'll have any other reason not to approve this PR. Cheers!

changelog/fix-4603-block-split-upe
timur27: This absolutely makes sense! I [removed the changelog file](https://github.com/Automattic/woocommerce-payments/pull/4681/commits/e3ceb21160673b1145a513fc0b5ed071b352c2e0).

client/checkout/utils/upe.js
timur27: Oh, good catch! Thanks for noticing and sorry for not checking the classic checkout flow after all the changes. [It's fixed now](https://github.com/Automattic/woocommerce-payments/pull/4681/commits/c16cfec392a41b1328280bb1b6d8159c323c22c2) and should work as expected.

timur27:
@FangedParakeet today, I found one more thing to align - `paymentIntent`s weren't removed from the session after an order gets processed successfully. Because of this, a user was getting an error while trying to make two purchases with one payment method, one after the other. 

The reason for that was that the payment gateway is considered separate when using Blocks, and [adding actions in the `else` statement](https://github.com/Automattic/woocommerce-payments/blob/88be4a965c11235ab78709577eba83967a48f09d/includes/payment-methods/class-upe-payment-gateway.php#L108-L121) in `class-upe-payment-gateway` was never done because none of the payment methods a user chooses is a cart.

Somehow this worked with the classic UPE checkout, and I guess that hooks' callbacks from `$card_gateway` were somehow re-used in other payment methods. 

I now fixed this behavior to work everywhere, and the problem disappeared. I tend to think that [the solution is not ideal](https://github.com/Automattic/woocommerce-payments/pull/4681/commits/f17d4a478ef180248d1ec0ec9f45b30c80b8223f), but it seems to be bulletproof within our area of interest.

FangedParakeet (approved):
Thank you so much for outlasting me and my endless feedback in this marathon PR. Code looks great and works like a dream. 

> I found one more thing to align - paymentIntents weren't removed from the session after an order gets processed successfully. 

Fantastic find here as well! I agree that the solution looks a little ugly (sorry 😁 ), but I can't think of a better way of ensuring that these actions definitely fire when the UPE is enabled, but only fire once. In fact, in the long term, I think this is a much more robust solution that allows us to potentially disable the `card` payment method entirely, without affecting these hooks.

I ran through a few regular checkout flows, using both the Blocks and shortcode checkout, and everything appears to be working as expected! I hereby offer the entirety of my grattitude and blessings. 

L. G. T. M. :shipit:

PR #4687, opened 2022-09-02
Display analytics orders in customer currency

### LINKED ISSUE #4446: MC Analytics: Support display analytics in customer currency
### Description

Follow on from #2535.

In #4441 we implemented support for filtering by customer currency.

We also need to implement the the ability for customers to choose in which currency they want the analytics to display. For example, if GBP is selected as the customer currency, the user should be able to choose whether to see the results in GBP or in the store default currency (the standard behaviour).

This will require us to perform conversions within our select clauses, if a customer currency is selected.

### Additional context

[Design thread](https://wcpay.wordpress.com/2021/04/16/multi-currency-analytics/)

ismaeldcom:
Fixes #4446 

- Add a **Customer Currency** filter in **Analytics → Orders** with all the enabled ones plus `All currencies` as default value. Hooking into `woocommerce_admin_orders_report_filters`.
- Use `woocommerce_admin_report_currency` to hook into report currency format and display amounts using the selected currency symbol.
- Use `currency` as argument for this new filter. As it doesn't collide with existing `currency_is` and `currency_is_not`. Adding it as another `WHERE` clause to only display orders in the selected currency.
- Filter `woocommerce_analytics_clauses_select_orders_subquery` and `woocommerce_analytics_clauses_select_orders_stats_total` when there is a currency selected, and it isn't the default one.
- Replace all occurrences of `wc_order_stats.net_total` on those queries with:
  - `CASE` to check if we have `wcpay_multicurrency_stripe_exchange_rate_meta` or need to fall back to `wcpay_multicurrency_exchange_rate_meta`.
  - `ROUND` of the arithmetic operation using `wc_get_price_decimals()`.
  - The actual multiply or divide operation based on the available rate.

#### Testing instructions
You'll need a store with more than one currency added and orders in at least 2 different currencies.

- Go to **Analytics → Orders**.
- The new **Customer Currency** filter should be there with value `All currencies`.
- Select a **Date range** where you have orders in more than one currency.
- Select your store currency in the **Customer Currency** filter.
  - Only orders in your store currency should be visible.
- Select a different currency in the filter.
  - Only orders in this currency should be visible.
  - All amounts should be converted and use the correct currency symbol and store format.
  - Ensure all the conversions are good including **Net sales** and **Average order value**, and there is no rounding issues between the order amounts and totals.
- Disable

#### Screenshots
| Currency options | All currencies | EUR | ISK |
|-|-|-|-|
|4446-currency-options|4446-all-currencies|4446-eur|4446-isk|

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : [Testing instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-4.9.0#display-analytics-orders-in-customer-currency)

### DIFF
--- a/client/multi-currency-analytics/index.js
+++ b/client/multi-currency-analytics/index.js
@@ -122,3 +122,41 @@ addFilter(
 		return tableData;
 	}
 );
+
+addFilter(
+	'woocommerce_admin_orders_report_filters',
+	'woocommerce-payments',
+	( filters ) => [
+		{
+			label: __( 'Currency', 'woocommerce-payments' ),
+			param: 'currency',
+			staticParams: [],
+			showFilters: () => true,
+			defaultValue: wcSettings.currency.code,
+			filters: getCustomerCurrencies(),
+		},
+		...filters,
+	]
+);
+
+const currencySymbols = Object.values( wcpaySettings.currencyData ).reduce(
+	( symbols, currency ) => {
+		symbols[ currency.code ] = currency.symbol;
+		return symbols;
+	},
+	{}
+);
+
+addFilter(
+	'woocommerce_admin_report_currency',
+	'woocommerce-payments',
+	( config, { currency } ) => {
+		if ( currency && currencySymbols[ currency ] ) {
+			return {
+				...config,
+				symbol: currencySymbols[ currency ],
+			};
+		}
+		return config;
+	}
+);
--- a/client/multi-currency-analytics/index.js
+++ b/client/multi-currency-analytics/index.js
@@ -139,21 +139,20 @@ addFilter(
 	]
 );
 
+const currencySymbols = Object.values( wcpaySettings.currencyData ).reduce(
+	( symbols, { code, symbol } ) => {
+		symbols[ code ] = symbol;
+		return symbols;
+	},
+	{}
+);
+
 // Filter report currency formatter to display the selected currency symbol.
 addFilter(
 	'woocommerce_admin_report_currency',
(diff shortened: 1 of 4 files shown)


client/multi-currency-analytics/index.js
```diff
@@ -122,3 +122,44 @@ addFilter(
 		return tableData;
 	}
 );
+
+addFilter(
+	'woocommerce_admin_orders_report_filters',
+	'woocommerce-payments',
+	( filters ) => [
+		{
+			label: __( 'Currency', 'woocommerce-payments' ),
```
dmallory42: What do you think about labelling this `Customer currency` just to make it clear that it refers to the currency the order was placed in?

dmallory42 (approved):
Tested and works really well. The solution is really nice as well, easy to understand and all the logic is in one place. I made one small suggestion but happy to approve this! :shipit:

client/multi-currency-analytics/index.js
ismaeldcom: 👍 I used `Currency` following the design, but it can be confusing as we're actually filtering and converting the order currency. Addressed in 881d882

dmallory42: Thanks for addressing!

PR #4700, opened 2022-09-06
Removing unnecessary check against server response

### LINKED ISSUE #4699: Internal E2E tests failing due to unneeded key existence check
### Describe the bug

Internal end to end tests are failing due to a newly updated check against `deposits.status` in the account status. Previously this check referenced `deposits_status`. 

https://github.com/Automattic/woocommerce-payments/blob/3ec6d887b5ed77e8f764fe30c316228a61c61410/includes/class-wc-payments-account.php#L185

https://github.com/Automattic/woocommerce-payments/blob/7360a79851af3f93034ea989c1846462f2ffef54/includes/class-wc-payments-account.php#L185

Reviewing the server response shows that if `payments_enabled` is returned, the full expected response is also returned making this check unnecessary.

### Additional context
p1662388714071439-slack-CGGCLBN58

brucealdridge:
Fixes #4699

Removed unnecessary check for `deposit.status` / `deposit_status` on response from server. As this key always exists when the other conditions are met.

Whenever the `payments_enabled` key is returned from the server, the deposit status is also returned.

#### Testing instructions

* Check that when `payments_enabled` is returned from the server that `deposits.status` or `deposits_status` is returned.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -192,7 +192,7 @@ class WC_Payments_Account {
 			'country'             => $account['country'] ?? 'US',
 			'status'              => $account['status'],
 			'paymentsEnabled'     => $account['payments_enabled'],
-			'deposits'            => $account['deposits'],
+			'deposits'            => $account['deposits'] ?? [],
 			'currentDeadline'     => isset( $account['current_deadline'] ) ? $account['current_deadline'] : false,
 			'pastDue'             => isset( $account['has_overdue_requirements'] ) ? $account['has_overdue_requirements'] : false,
 			'accountLink'         => $this->get_login_url(),
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -180,9 +180,7 @@ class WC_Payments_Account {
 			];
 		}
 
-		if ( ! isset( $account['status'] )
-			|| ! isset( $account['payments_enabled'] )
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments-account.php
```diff
@@ -180,9 +180,7 @@ public function get_account_status_data() {
 			];
 		}
 
-		if ( ! isset( $account['status'] )
-			|| ! isset( $account['payments_enabled'] )
-			|| ! isset( $account['deposits']['status'] ) ) {
+		if ( ! isset( $account['status'], $account['payments_enabled'] ) ) {
```
mattallan, line 183: This is probably pretty minor but using `! isset( $a, b )` is changing our original logic.
This is line is now equivalent to `! isset( $account['status'] ) && ! isset( $account['payments_enabled'] )` but previously we had `||`

mattallan: Based on this test code (props @brucealdridge), my understanding was incorrect and these are equivalent.

mattallan (approved):
Thanks @brucealdridge this PR LGTM.

includes/class-wc-payments-account.php
brucealdridge: Just to close the loop on this, the logic was [confirmed](https://3v4l.org/r6i6h) to behave as expected.

PR #4705, opened 2022-09-06
Hide Link payment method for non-us accounts

### LINKED ISSUE #4668: Stripe Link is available for non-US accounts when UPE is enabled
### Describe the bug
Stripe Link is only available for US accounts. However, when UPE is enabled, non-US accounts can enable Stripe Link.

### To Reproduce

1. Spin up a new JN site.
2. Enable dev mode.
3. Create a new account in a country based outside the US.
4. Navigate to **Payments > Settings**.
5. Note that Stripe Link is not an option.
6. Enable UPE.
7. Note that Stripe Link is now available.
8. Disable UPE.
9. Stripe Link is still an option?
10. Try to enable Stripe Link and save the settings; an error will be displayed.
11. Reload the page—Stripe Link is no longer available.

### Actual behavior
Stripe Link can be enabled on sites that aren't based in the U.S.; some users report that this can lead to errors on checkout such as:

```
An error was encountered when preparing the payment form. Please try again later.
```

I cannot reproduce these checkout errors.

### Screenshots

US account:

Markup on 2022-08-31 at 11:08:42

CA account:

Markup on 2022-08-31 at 11:10:23

Markup on 2022-08-31 at 11:09:55

### Expected behavior
Stripe Link should only be an option for U.S. accounts—regardless of whether or not UPE is enabled.

### Additional context

5504970-zen

dpaun1985:
Fixes #4668 

Stripe Link is available only in US. 
This PR updates the available payment methods based, in order to add Link only for US country and if the capability exists for this payment method.

#### Testing instructions

* With an US account, enable Link payment method in admin
* Create an order with Link payment method. 
* No errors should occur
* Test also with Link capability disabled. Link option should not be displayed in Express checkout section
* Link a non US account.
* Go to available payment methods. Link should not be visible. Note: if you previously enabled Link, save the form, so the Link will be removed from the enabled payment methods
* Create an order. No errors should occur

-------------------

**Post merge**


includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -984,7 +984,17 @@ public function get_upe_available_payment_methods() {
 		$methods[] = Sofort_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
 		$methods[] = Sepa_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
 		$methods[] = P24_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
-		$methods[] = Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
+
```
kalessil: So it was the capability ignored. Can we slightly refactor this to be more universal?

```php
                $capabilities = $this->get_payment_method_capability_key_map();
		if ( isset( $capabilities[ Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID ] ) )  {
			$payment_methods_statuses = $this->get_upe_enabled_payment_method_statuses();
			$link_status              = $payment_methods_statuses[ $capabilities[ Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID ] ]['status'];
			if ( $link_status === 'active' ) {
				$methods[] = Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
			}
		}
```

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -993,11 +1003,7 @@ public function get_upe_available_payment_methods() {
 			)
 		);
 
-		$methods_with_fees = array_values( array_intersect( $methods, array_keys( $fees ) ) );
-
-		$methods_with_fees[] = Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID;
-
-		return $methods_with_fees;
+		return array_values( array_intersect( $methods, array_keys( $fees ) ) );
```
kalessil: I believe it's connected to the previous change, but I'm not following it. Can you enlighten me, please?

kalessil (changes requested):
Looks good overall! I left some comments. Can you take a look into them, please?

includes/payment-methods/class-upe-payment-gateway.php
dpaun1985: `$link_status === 'active'` - I would keep `$link_status !== 'inactive'`, because I noticed that the status can be also `unrequested` . In this case we should show the Link payment method, so the merchant can enable it.
What do you think @kalessil ?

kalessil: In that case, it still not be available for use in payments. That's why I suggested using `$link_status === 'active'`.

kalessil: Update: with new input from Dan we agreed to proceed with `$link_status !== 'inactive'`

kalessil: @dpaun1985 : can you also update PR testing instructions please, to test the US account with the enabled/disabled link capabilities?

kalessil: @frosso, @ismaeldcom: I'm not sure how much you are remember about this logic, but why don't we rely on account capabilities for building the list initially (and later cross-checking it against available fees)?

includes/payment-methods/class-upe-payment-gateway.php
dpaun1985: `$methods_with_fees[] = Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID;` was added before adding fees for Link payment method. So, after fees were added, 'link' was added two times.

includes/payment-methods/class-upe-payment-gateway.php
frosso: Sorry @kalessil , I'm not familiar with the reasoning behind it 😬 maybe because the cached account data might not be available when we build the list (and we might want to display the payment methods on the settings page anyways)?

ismaeldcom: AFAIK we don't use account capabilities because this list is used to show available payment methods to be enabled/disabled and their capability requested. If we build it based on current capabilities, they will not see any additional payment method to enable on the list (in settings).

I understand it like a list of available UPE payment methods on our platform and merchant country. Not tied to merchant account capabilities.

dpaun1985: >I understand it like a list of available UPE payment methods on our platform and merchant country. Not tied to merchant account capabilities.

For some countries, not all payment methods can be enabled. In this case, in Stripe response, that capability will not be present. Based on Stripe response I think we can hide/show available payment methods in the admin -> if a capability is not present ( no matter the status ), that means it is not available for that merchant ( it's not in an eligible country... or any other reason )

ismaeldcom: That's what we're doing with the intersection with fees. It allows us to better control which payment methods to show. Because Stripe might support it but we don't. And as we manually add fees for every country and payment method, that will match our available payment methods 🙂

mgascam (approved):
LGTM (did code review only, I'll run some manual tests and let you know if I find any issues)

mgascam (approved):
I did some manual tests and works as expected:
✅ Order successfully created with US account with Link enabled & disabled
✅ Link payment method not visible for non US account
✅ Order successfully created with non US account

kalessil (approved):
LGTM!

kalessil:
@dpaun1985, we are good to merge this PR.

PR #4495, opened 2022-07-28
Make updating existing customer details during checkout async

deepakpathania:
Fixes #4490 

#### Changes proposed in this Pull Request
- Make updating existing customer details during checkout async for performance benefits.

#### Testing instructions

- Follow the normal checkout flow(both UPE and non UPE) for an existing customer.
- Ensure there is a scheduled action created as `wcpay_update_customer_with_order_data` in Scheduled Actions (Tools > Scheduled Actions)
- Run the action and ensure there are no errors, additionally check logs to ensure a POST `wcpay/customers/{{customer_id}}` call was made (check WP-Admin > WooCommerce > Status > Logs and there should be a log for WooCommerce Payments with today's date)
- Ensure any UPE and subscription flows are not affected by this change. 
  - Context for subscriptions: p1658831357623559-slack-C02BW3Z8SHK

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4443, opened 2022-07-15
Make updating saved payment method async during checkout 

deepakpathania:
Fixes #4440 

#### Changes proposed in this Pull Request
- Make updating saved payment method during checkout async for performance benefits.

#### Testing instructions

* Follow the normal checkout flow using a saved payment method.
* Ensure there is a scheduled action created as `wcpay_update_saved_payment_method` in Scheduled Actions (Tools > Scheduled Actions)
* Run the action and ensure there are no errors, additionally check logs to ensure a `POST wcpay/payment_methods/{{payment_method}}` call was made. 

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : [Testing instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-4.5.0#make-updating-saved-payment-methods-async)
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4713, opened 2022-09-07
Retain test mode context in CRON jobs queued up while checking out

### LINKED ISSUE #1257: Support non-USD currencies in fees on the settings screen
Ensure fees with non-USD fixed rates are displayed correctly on the WCPay settings page (including discount fee structures).

lovo-h:
* Fixes N/A

* Before, the test mode context would be lost when the CRON job was dequeued. Now, the live mode context is passed along and applied in the CRON job.

#### Testing instructions

**Checkout through WooPay**
* Ensure adding fee details to order notes async applies correctly in test/live mode (#4429).
* Ensure updating saved payment method async applies correctly in test/live mode (#4443).
* Ensure updating existing customer details applies correctly in test/live mode (#4495).

**Checkout through the merchant site using a saved payment method**
* Ensure adding fee details to order notes async applies correctly in test/live mode (#4429).
* Ensure updating saved payment method async applies correctly in test/live mode (#4443).
* Ensure updating existing customer details applies correctly in test/live mode (#4495).

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1229,7 +1229,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		$metadata = $this->get_metadata_from_order( $order, $payment_information->get_payment_type() );
 
 		$customer_details_options   = [
-			'is_woopay' => $metadata['paid_on_woopay'] ?? false,
+			'is_woopay' => filter_var( $metadata['paid_on_woopay'] ?? false, FILTER_VALIDATE_BOOLEAN ),
 		];
 		list( $user, $customer_id ) = $this->manage_customer_details_for_order( $order, $customer_details_options );
 
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -386,8 +386,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		// Update the current request logged_in cookie after a guest user is created to avoid nonce inconsistencies.
 		add_action( 'set_logged_in_cookie', [ $this, 'set_cookie_on_current_request' ] );
 
-		add_action( self::UPDATE_CUSTOMER_WITH_ORDER_DATA, [ $this, 'update_customer_with_order_data' ], 10, 2 );
 		add_action( self::UPDATE_SAVED_PAYMENT_METHOD, [ $this, 'update_saved_payment_method' ], 10, 3 );
+		add_action( self::UPDATE_CUSTOMER_WITH_ORDER_DATA, [ $this, 'update_customer_with_order_data' ], 10, 3 );
 
 		// Update the email field position.
 		add_filter( 'woocommerce_billing_fields', [ $this, 'checkout_update_email_field_priority' ], 50 );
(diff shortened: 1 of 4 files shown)


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1103,18 +1103,39 @@ protected function prepare_payment_information( $order ) {
...
-	 * Update the customer details with the incoming order data.
+	 * Update the customer details with the incoming order data, in a CRON job.
 	 *
 	 * @param int    $order_id       WC order id.
 	 * @param string $customer_id    The customer id to update details for.
-	 */
-	public function update_customer_with_order_data( $order_id, $customer_id ) {
+	 * @param array  $options        Additional options to apply.
+	 */
+	public function update_customer_with_order_data( $order_id, $customer_id, $options = [] ) {
```
deepakpathania: Nitpick but it might be better to pass the exact params to the async functions instead of an options array here for clear reference in the function signature.

deepakpathania (commented):
I tested only the merchant checkout flow and that seems to be working fine, left a small nitpick comment.

includes/class-wc-payment-gateway-wcpay.php
lovo-h: Since there are only 2 additional params, this makes sense. Fixed in [9785250](https://github.com/Automattic/woocommerce-payments/pull/4713/commits/9785250eb8f8ebd642102582195d21f78c4619f7).

deepakpathania: Don't we have to update the accepted params in the `add_action` call for this method for the extra param to be reflected? 🤔  @lovo-h

lovo-h: Good catch. Fixed in [f8365a0](https://github.com/Automattic/woocommerce-payments/pull/4713/commits/f8365a0085d6c97ef3f78a1c608605d5a57ceb4c). Also, added a test in [6facffb](https://github.com/Automattic/woocommerce-payments/pull/4713/commits/6facffb68171625ca3aa896d1081ff0eaf69e901), to avoid a similar issue in the future.

EvanDotPro (approved):
Tested with WooPay and seems to be working :+1:.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1194,7 +1228,10 @@ public function process_payment_for_order( $cart, $payment_information, $additio
 		$amount   = $order->get_total();
 		$metadata = $this->get_metadata_from_order( $order, $payment_information->get_payment_type() );
 
-		list( $user, $customer_id ) = $this->manage_customer_details_for_order( $order );
+		$customer_details_options   = [
+			'is_woopay' => $metadata['paid_on_woopay'] ?? false,
```
reykjalin: I think the metadata may be serialized into a `'1'` in some cases so it's probably worth it to `filter_var( $metadata['paid_on_woopay'], FILTER_VALIDATE_BOOLEAN ) ?? false;` here 🤔

reykjalin (changes requested):
I had one minor comment but otherwise this is looking good 🎉

deepakpathania (approved):
Thanks for addressing the comment, changes LGTM now. (tested only merchant site)  ✅

includes/class-wc-payment-gateway-wcpay.php
lovo-h: Good call. Fixed in [1a88ed4](https://github.com/Automattic/woocommerce-payments/pull/4713/commits/1a88ed48cc8e663c331bf475c79cb5b72b3b8d9d).

reykjalin (approved):
LGTM 🚀

PR #4033, opened 2022-03-25
Update wordpres/components to fix VisuallyHidden styles not loading

oscarssanchez:
Fixes #4015

#### Changes proposed in this Pull Request

Update wordpres/components to fix VisuallyHidden styles not loading.

This PR updates the @wordpress/components dependency to `19.7.0`. It seems the old dependency has a bug where the `VisuallyHidden` component styles are not loaded by `ExternalLink`

#### Testing instructions

1.- Navigate to Payments → Settings.
2.- Scroll down to the "Transactions and deposits" section.
3.- See the text `opens in a new tab`  is now hiding correctly.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) : _Add link here / 'QA Testing Not Applicable'_

Issue #4015, opened 2022-03-23
Visually hidden text is not visually hidden


aprea:
### To Reproduce

1. As a merchant, navigate to **Payments → Settings**.
2. Scroll down to the "Transactions and deposits" section.
3. Observe the text:
> Manage and update your deposit account information to receive payments and payouts. [Manage in Stripe(opens in a new tab)](https://alleged-okapi.jurassic.ninja/wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments&wcpay-login=1&_wpnonce=1a2a83408e)

The "(opens in a new tab)" text should not be visible. It's present in the DOM for screen readers but shouldn't actually be visible. I noticed the span element uses the `components-visually-hidden` class but I do not see any CSS styles attached to that class.

### Actual behavior
Visually hidden text is not visually hidden.

### Screenshots
<img width="627" alt="Screen Shot 2022-03-23 at 11 18 38 am" src="https://user-images.githubusercontent.com/1527164/159597714-224b175e-bf3b-40af-8cc0-2c8c7a4d9116.png">

### Expected behavior
Visually hidden text is visually hidden.

### Desktop (please complete the following information):

* Device: MBP
* OS: macOS 11.2.3 (20D91)
* Browser: Google Chrome Version 98.0.4758.109 (Official Build) (x86_64)

oscarssanchez:
Hi @aprea ,

It seems the `<ExternalLink> `component uses `VisuallyHidden` as a dependency. The current version of `@wordpress/components` used by WC Payments seems to have a bug where the styles from `VisuallyHidden` are not being properly loaded.

In my opinion there's two possible solutions:

1.- Update the `@wordpress/components` dependency to the latest version (`19.7.0`) and re-build the assets. This will get rid of similar scenarios happening like these:

<img width="1356" alt="Screen Shot 2022-03-23 at 23 04 14" src="https://user-images.githubusercontent.com/31049169/159846671-2834a62b-5608-46ed-8e53-4b17b87ce981.png">

Also, this line could probably be deleted as well: https://github.com/Automattic/woocommerce-payments/blob/develop/client/components/payment-methods-checkboxes/payment-method-checkbox.scss#L77

Finally it will also ensure future compatibility.

2.- Create the css class (hacky), but it will solve the problem without needing to update the dependency.

I see the plugin has some tests in place, but am not sure how big of a lift it is to test updating the dependency. I'm happy to craft a PR based on your preference.

[...]

PR #4715, opened 2022-09-08
Adding manual option to deposit schedule selection

### LINKED ISSUE #4696: [Manual deposits] Deposit Scheduling: Merchants can select manual schedule.
### Description

#4464 added the ability for merchants to choose an automatic schedule for deposits (daily, weekly, monthly) but to meet the WCPay 4.7 release window the ability for a merchant to select a manual schedule was descoped.

Add a set of controls to the Payments → Settings page which allows the merchant to change their deposit schedule from an automatic schedule to manual.

Deposits section

### Acceptance criteria

- When `Automatic` is selected, the automatic schedule options are available (Daily, Weekly, Monthly)
- When `Manual` is selected the automatic schedule options are disabled (but still visible).
- The "Deposit schedule" controls are implemented as per the designs W8FHz6thdADIqhlPlBo4Mf-fi-1718%3A77943
- When the merchant saves their settings via the "Save changes" button, their deposit schedule is updated in Stripe via WCPay Server as requested.
- The "Learn more" links that appear in the helper text link to https://woocommerce.com/document/payments/faq/change-deposit-schedule/

**Post-release:** Add user docs for this feature to https://woocommerce.com/document/payments/

### Designs

W8FHz6thdADIqhlPlBo4Mf-fi-1718%3A77943

### Testing instructions

TBC

### Additional context

Project thread: pdjTHR-1c8-p2
PRD: pdjTHR-1bY-p2

brucealdridge:
Fixes #4696

Introduce top-level radio selections, `Automatic` and `Manual`, for deposits schedule.

#### Testing instructions

* Run `npm run build:client`.
* Go to wp-admin > Payments > Settings.
* Make sure `Automatic` and `Manual` radios appear below `Deposit schedule`.
* Make sure they are functional, eg selecting `Manual` and clicking the `Save changes` button will set the Stripe account payout schedule to `Manual` (check the Stripe dashboard to confirm this).
* The deposit settings match the designs → paJDYF-6a4-p2

-------------------

**Post merge**


shendy-a8c:
@brucealdridge mentioned p1662608177560469/1662603893.523189-slack-C03KT8NLL13 that currently, the UI forgets previous selection when switch to manual, that is:
- Select `Automatic deposits`.
- Select `Weekly`.
- Then select `Manual deposits`.
- Then select `Automatic deposits`.
- Notice that the dropdown is now set to `Daily` instead of `Weekly`.

IMO that is fine. To make it *not* forget, we would need to save previous state / unsaved state and I think that is too over-engineering the problem.

Let me know if making the UI remember is a must have.

Jinksi:
> IMO that is fine. To make it _not_ forget, we would need to save previous state / unsaved state and I think that is too over-engineering the problem.

I agree that it is probably not important, since the user is most likely going to select Manual or Automatic, and not flick between like we are in testing.

The Weekly/Monthly anchor persists when toggling automatic/manual and interval options. This makes it slightly less painful if the user accidentally selected "Manual" and can't remember which day of the month they had previously selected.

Jinksi:
Should we be disabling the form/inputs while saving is in progress?
Currently, the form inputs are enabled and can be changed while the settings request is in progress. This could be confusing for users, especially if the request takes a while.

Note that this applies to the entire Payments Settings for, so it is probably out-of-scope for this PR.

https://user-images.githubusercontent.com/3147296/189325927-9d2c3522-8ca2-4334-a726-eb6ff95dea94.mov

shendy-a8c:
That is a good point, @Jinksi! I think it's important to address and it's out of scope for this PR. I quickly searched if there is existing issue to address this problem (disabling the entire settings page when saving) but found none. Could you please create a new issue?

Jinksi:
> Could you please create a new issue?

Thanks Shendy, #4740 created (I couldn't find an existing issue either).

aprea:
p.s. @shendy-a8c, this PR is ready for review.

client/settings/deposits/index.js
```diff
@@ -72,6 +75,17 @@ const monthlyAnchors = [
 ];
 
 const CustomizeDepositSchedule = () => {
+	const {
+		accountStatus: {
+			deposits: { minimum_deposit_amounts: minimumDepositAmounts },
+		},
+		storeCurrencies,
+	} = useContext( WCPaySettingsContext );
```
shendy-a8c, line 83: If for some reason (maybe unexpected future change from server), `accountStatus.deposits` isn't defined, this line will throw an error, right?

Maybe we can be more defensive here?

shendy-a8c (changes requested):
When I choose Manual, save, and reload the page, the selection becomes Automatic daily. That is not expected right?

https://user-images.githubusercontent.com/73803630/196812104-162ef7fc-bafe-493d-9d91-4bd1b1efbfaf.mov

shendy-a8c:
Also, `ExternalLink` looks weird. It adds `(opens in a new tab)`. No space prefix makes it look weird but it's out of the scope of this PR.

Screen Shot 2022-10-20 at 05 00 50

aprea:
> Also, `ExternalLink` looks weird. It adds `(opens in a new tab)`. No space prefix makes it look weird but it's out of the scope of this PR.

This issue will be handled in https://github.com/Automattic/woocommerce-payments/issues/4015.

client/settings/deposits/index.js
aprea: I don't think we can be too defensive about the server not returning the expected response. If we need to eventually move/change the payload then it should be done in a backwards-compatible way.

aprea:
> When I choose Manual, save, and reload the page, the selection becomes Automatic daily. That is not expected right?

Good catch! This is because WCPay Server returns the `daily` schedule even if the actual schedule in Stripe is set to `manual`. This happens when we determine that the merchant **has not** completed the new account waiting period. Related code → https://github.com/Automattic/woocommerce-payments-server/blob/5534876318a40f980f1e8d3521b48cc81a726f9b/server/wp-content/rest-api-plugins/endpoints/wcpay/stripe/class-stripe-account.php#L299-L324

I have a PR that is waiting to be reviewed which should fix this issue, would you be able to take a look at it? https://github.com/Automattic/woocommerce-payments-server/pull/2495

shendy-a8c:
Thank you for the explanation.

After testing with https://github.com/Automattic/woocommerce-payments-server/pull/2495, every thing works as expected, , so approving this PR.

I checked:
- Deposit schedule changes from WooCommerce Payments settings page are reflected to the deposit schedule on Stripe dashboard.
- Helper text looks good.
- The deposits settings UI aligns with the latest design update from paJDYF-6a4-p2, ie the dropdowns are disabled instead of hidden.
- Code diff looks good too.

shendy-a8c (approved):

shendy-a8c:
PR of https://github.com/Automattic/woocommerce-payments/issues/4015, https://github.com/Automattic/woocommerce-payments/pull/4033, seems stale even though it looks ready to go.

aheckler:
@aprea @shendy-a8c The docs URL used by `const depositSettingsDocsUrl` is not ideal. It should be:

`https://woocommerce.com/document/payments/faq/change-deposit-schedule/`

aprea:
Thanks for raising this @aheckler, I'll raise an issue/PR to update that URL.

PR #4551, opened 2022-08-09
Fix errors for apple pay registration

dpaun1985:
Fixes #4511

#### Changes proposed in this Pull Request
- when merchant is not in the Apple Pay [supported country list](https://support.apple.com/en-us/HT207957
), or `apple-developer-merchantid-domain-association` file has different permissions than 644, the apple domain registration is not called anymore
- for the conditions above, a notice is shown in the WooCommerce Payments admin panel

Screenshot 2022-08-09 at 14 37 20

#### Testing instructions
1. Set the merchant country other than the ones listed in https://support.apple.com/en-us/HT207957 .
2. Enable Apple Pay
3. `Apple Pay: You are not in the countries and regions that support Apple Pay. You can find a full list of countries and regions at https://support.apple.com/en-us/HT207957` should be displayed in Payment methods panel
4. Rename `docker/wordpress/.well-known/apple-developer-merchantid-domain-association` file
5. `Apple Pay: Invalid domain association file permissions. /var/www/html/.well-known needs to have 755 permissions and /var/www/html/.well-known/apple-developer-merchantid-domain-association 644. Set the proper permissions and disable/enable Apple Pay.`  should be displayed in Payment methods panel

*

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4726, opened 2022-09-09
Fix Apple Pay verify domain waring notice showing up constantly

### LINKED ISSUE #4712: Apple Pay "We were not able to verify your domain" admin notice showing up
### Describe the bug

An Apple Pay warning admin notice is displayed on the admin, regardless of whether the steps provided by the notice are actually needed. This is happening since 4.7.0.

The admin notice on the WCPay settings page reads "Apple Pay: We were not able to verify your domain. [This help documentation](https://woocommerce.com/document/payments/apple-pay/#apple-pay-button-does-not-appear) will walk you through the process to verify with Apple that you control your domain". 

Payments made with Apple Pay should continue to work as long as the configuration file Apple Pay looks for exists, regardless of the presence of the admin notice. More details about this are below in the "Expected behavior" section.

### To Reproduce

1. Go to Payments -> Settings
2. Under Express Checkouts, enable "Apple Pay / Google Pay", if not enabled already. Save and reload the page
3. Notice the notice at the top of the page

### Actual behavior

Apple Pay's warning shows up when it's not needed. That is when either the configuration file exists or the option under Admin > Settings > Permalinks has anything but "Plain" selected.

### Screenshots

image

### Expected behavior
We use custom rewrites to generate the file Apple Pay requires when this option in Admin > Settings > Permalinks is anything but "Plain". The notice shouldn't show up in this case.
When the option under Admin > Settings > Permalinks has "Plain" selected and no configuration file exists, the notice should show up. 

### Additional context
This issue is high priority because following the steps provided by the warning could lead merchants to mess up their Apple Pay flow.

Introduced in https://github.com/Automattic/woocommerce-payments/pull/4551

mordeth:
Fixes #4712

The Apple Pay admin domain verify notice introduced on #4551 displays the false positive **error** notice without checking the site permalinks structure asking for not required steps that could confuse merchants.

This PR proposes adding a  **warning** notice that informs merchants we are not able to access their domain verification file when the pretty permalinks structure is enabled with the following message:

`We were not able to access your domain verification file. WooCommerce Payments will keep on working, but we recommend using this documentation to walk you through the process to verify with Apple that you control your domain.`

#### Testing instructions
- Navigate to `Payments` -> `Settings`
- Under Express Checkouts, enable "Apple Pay / Google Pay", if not enabled already. Save and reload the page.

**Scenario 1**
- Navigate to `Settings` -> `Permalinks` and set the permalinks to `Post name`
- Rename `docker/wordpress/.well-known/apple-developer-merchantid-domain-association`
- Navigate to `Payments` -> `Settings`
- Verify a warning message is displayed like the one below:
Screenshot 2022-09-09 at 14 17 59

**Scenario 2**
- Navigate to `Settings` -> `Permalinks` and set the permalinks to `Plain`
- Rename `docker/wordpress/.well-known/apple-developer-merchantid-domain-association`
- Navigate to `Payments` -> `Settings`
- Verify a warning message is displayed like the one below:
Screenshot 2022-09-09 at 14 19 22

**Scenario 3**
- Rename `docker/wordpress/.well-known/apple-developer-merchantid-domain-association` to original file.
- Navigate to `Payments` -> `Settings`
- Verify no error or warning message is displayed.

-------------------

### DIFF
--- a/includes/class-wc-payments-apple-pay-registration.php
+++ b/includes/class-wc-payments-apple-pay-registration.php
@@ -386,7 +386,7 @@ class WC_Payments_Apple_Pay_Registration {
 		$well_known_dir = untrailingslashit( ABSPATH ) . '/' . self::DOMAIN_ASSOCIATION_FILE_DIR;
 		$full_path      = $well_known_dir . '/' . self::DOMAIN_ASSOCIATION_FILE_NAME;
 
-		return is_dir( $well_known_dir ) && is_writable( $well_known_dir ) && file_exists( $full_path ) && is_readable( $full_path );
+		return is_dir( $well_known_dir ) && is_writable( $well_known_dir ) && file_exists( $full_path ) && is_writable( $full_path );
 	}
 
 	/**
--- a/includes/class-wc-payments-apple-pay-registration.php
+++ b/includes/class-wc-payments-apple-pay-registration.php
@@ -545,7 +545,7 @@ class WC_Payments_Apple_Pay_Registration {
 						<?php
 							echo sprintf(
 								/* translators: 1: apple pay support */
-								__( 'We were not able to access your domain verification file. WooCommerce Payments will keep on working, but we recommend using <a href="%1$s" target="_blank">this documentation</a> to walk you through the process to verify with Apple that you control your domain.', 'woocommerce-payments' ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+								__( 'We are not able to access your Apple Pay domain verification file. WooCommerce Payments will continue to work, but we suggest using <a href="%1$s" target="_blank">this document</a> to help you verify with Apple that you control this domain.', 'woocommerce-payments' ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 								'https://woocommerce.com/document/payments/apple-pay/#apple-pay-button-does-not-appear'
 							);
 						?>
--- a/includes/class-wc-payments-apple-pay-registration.php
+++ b/includes/class-wc-payments-apple-pay-registration.php
@@ -386,7 +386,7 @@ class WC_Payments_Apple_Pay_Registration {
 		$well_known_dir = untrailingslashit( ABSPATH ) . '/' . self::DOMAIN_ASSOCIATION_FILE_DIR;
 		$full_path      = $well_known_dir . '/' . self::DOMAIN_ASSOCIATION_FILE_NAME;
 
-		return is_dir( $well_known_dir ) && is_writable( $well_known_dir ) && file_exists( $full_path ) && is_writable( $full_path );
+		return ( file_exists( $full_path ) && is_writable( $full_path ) ) || ( is_dir( $well_known_dir ) && is_writable( $well_known_dir ) );
 	}
 
 	/**
--- a/includes/class-wc-payments-apple-pay-registration.php
+++ b/includes/class-wc-payments-apple-pay-registration.php
@@ -89,6 +89,7 @@ class WC_Payments_Apple_Pay_Registration {
 
 		add_action( 'woocommerce_woocommerce_payments_admin_notices', [ $this, 'display_error_notice' ] );
 		add_action( 'woocommerce_woocommerce_payments_admin_notices', [ $this, 'display_live_account_notice' ] );
+		add_action( 'woocommerce_woocommerce_payments_admin_applepay_notice', [ $this, 'display_not_supported_apple_pay' ] );
 		add_action( 'add_option_woocommerce_woocommerce_payments_settings', [ $this, 'verify_domain_on_new_settings' ], 10, 2 );
 		add_action( 'update_option_woocommerce_woocommerce_payments_settings', [ $this, 'verify_domain_on_updated_settings' ], 10, 2 );
 	}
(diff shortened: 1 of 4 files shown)


includes/class-wc-payments-apple-pay-registration.php
```diff
@@ -351,6 +352,54 @@ public function verify_domain_on_updated_settings( $prev_settings, $settings ) {
...
+	/**
+	 * Check if domain association file has the proper permissions.
+	 *
+	 * @return bool
+	 */
+	public function has_domain_association_file_permissions(): bool {
+		$well_known_dir = untrailingslashit( ABSPATH ) . '/' . self::DOMAIN_ASSOCIATION_FILE_DIR;
+		$full_path      = $well_known_dir . '/' . self::DOMAIN_ASSOCIATION_FILE_NAME;
+
+		return is_dir( $well_known_dir ) && is_writable( $well_known_dir ) && file_exists( $full_path ) && is_readable( $full_path );
```
RadoslavGeorgiev: This line does too many things at once, and I'm not sure if all/any of them are correct. Please take a look at [`copy_and_overwrite_domain_association_file()`](https://github.com/Automattic/woocommerce-payments/blob/fix/4712-apple-pay-incorrect-admin-notice/includes/class-wc-payments-apple-pay-registration.php#L165-L176) before proceeding with this comment:

- `copy_and_overwrite_domain_association_file()` requires neither the file nor directory to exist in order to work, but here we require both.
- Why do we check if the file `is_readable()`? We need to either check if it's writable, or verify that it already has content, while comparing the content to what we have in the gateway.

Also, during my testing, I went ahead and created an empty, and writable file. The gateway did nothing to populate it, which is not ideal either.

IMO whenever we detect that the file does not exist, we should try creating it. If that fails, we should look for an exception or an error before displaying this warning. WDYT?

includes/class-wc-payments-apple-pay-registration.php
```diff
@@ -434,4 +483,76 @@ public function display_error_notice() {
...
+				<?php
+			} else {
+				?>
+				<div id="wcpay-applepay-warning" class="notice notice-warning">
+					<p>
+						<b><?php esc_html_e( 'Apple Pay: ', 'woocommerce-payments' ); ?></b>
+						<?php
+							echo sprintf(
+								/* translators: 1: apple pay support */
+								__( 'We were not able to access your domain verification file. WooCommerce Payments will keep on working, but we recommend using <a href="%1$s" target="_blank">this documentation</a> to walk you through the process to verify with Apple that you control your domain.', 'woocommerce-payments' ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
RadoslavGeorgiev: This string feels weird. Could a native English speaker take a look?

RadoslavGeorgiev (changes requested):

includes/class-wc-payments-apple-pay-registration.php
kalessil: > Also, during my testing, I went ahead and created an empty, and writable file. The gateway did nothing to populate it, which is not ideal either.

That sounds like a potential bug, but not related to this PR. Mind filing a new bug-report please, so we could look into the bug in the next sprint?

> `copy_and_overwrite_domain_association_file()` requires neither the file nor directory to exist to work, but here we require both.

Good point! This condition is not matching to the function logic.
If I follow it correctly, it should be either:
- file exists and is writable
- well known directory exists and is writable
- parent directory of well-known one exists and writable

@RadoslavGeorgiev, @mordeth  can you verify my understanding, please?

RadoslavGeorgiev: The logic is correct, yes. 

> That sounds like a potential bug, but not related to this PR. Mind filing a new bug-report please, so we could look into the bug in the next sprint?

As far as I'm concerned, this PR already fixes the immediate issue and can be shipped as is. Seems like we need to properly consider the whole logic around registration, so a more complete solution in a separate PR makes more sense to me.

includes/class-wc-payments-apple-pay-registration.php
aheckler: Maybe something like this?

> We are not able to access your Apple Pay domain verification file. WooCommerce Payments will continue to work, but we suggest using this document to help you verify with Apple that you control this domain.

includes/class-wc-payments-apple-pay-registration.php
mordeth: In my opinion, in the case of checking if the file exists to display the error message, I think the `is_hosted_domain_association_file_up_to_date()` method should be enough. It checks if the file exists and if the content is up to date. However, the check for if the directory or the parent directory exists and is writeable should be part of the `copy_and_overwrite_domain_association_file()` when the file is created or updated. What do you think?

includes/class-wc-payments-apple-pay-registration.php
kalessil: Updated in e7382294083d0c9f1dd4f3ada333e00d34024a02

includes/class-wc-payments-apple-pay-registration.php
kalessil: @RadoslavGeorgiev, @mordeth : can you create a new issue to revisit the logic around the registration file (also to cover the condition)? Thanks!

kalessil: Meanwhile, I've updated the file permission check in 0ddcf4b7027a54c61fa83b20079b1949fbd8b48b and b0a405963523a8a81198c74c75e0054864f611b7

dpaun1985 (approved):
✅ LGTM!
Tested and works as expected!

kalessil (approved):
LGTM!

For retrospective: the fix itself is `'' === get_option( 'permalink_structure', '' )`, but also some improvements were made (a message wording, association file permission check and code relocation into Apple Pay specialize component).

kalessil:
@RadoslavGeorgiev : the PR requires one more review round from you.

RadoslavGeorgiev (approved):
The changes look good as a fix for the immediate issue. :shipit: 

As mentioned in [this conversation](https://github.com/Automattic/woocommerce-payments/pull/4726#discussion_r967011655), we should find a way to robustly check and update the file, but that can be handled as a separate issue in the future. For now this PR will stop showing an error, even though payment requests are still working.

RadoslavGeorgiev:
If we are about to release a patch with this PR, it should be rebased on top of `trunk` before merging.

PR #4782, opened 2022-09-21
Cloak client secret 

### LINKED ISSUE #4774: Try to hide the `client_secret` of Stripe from the card testing automation scripts
### Description
We need to prevent the user from confirming the payment intent using Stripe’s REST API, or creating their own Stripe Element using the client_secret we expose on the browser. This is more like an experimentation of trying multiple things to see if it can be done.

tpaksu:
Fixes #4774

This PR includes a PoC about hiding the exposed `client_secret` variable on the  `create_payment_intent` and `create_setup_intent` responses from automated scripting for card testing attacks on the client. 

Basically, it encrypts the `client_secret` variable with AES-128-CBC encryption just before pouring it to the browser, and decrypts it just before the calls to the Stripe API via the Stripe JS library. 

This is a measure to prevent the attacker reaching to the `client_secret` variable, which lets them call Stripe REST API directly to confirm intents, without using our server endpoints.

Also it can be enabled/disabled by clicking the "**Enable public key encryption**" checkbox in **Payments > Settings** page, hidden under the `Advanced` section.

image

#### Testing instructions

- Checkout this branch, run `npm run build:client`
- Open **Payments > Settings** and expand the **Advanced** section, see that there's a new toggle, labeled with **Enable UPE public key encryption**, and it should be **off** as default.
- Test if the toggle persists its value to `wp_options` table on the client database, with the option name `_wcpay_feature_client_secret_encryption`.
- Toggle the feature on.

 **Client Secret Encryption** | **Enabled** | **Enabled** | **Disabled** | **Disabled** 
------------------------------|-------------|-------------|--------------|--------------
 **UPE**                      | Enabled     | Disabled    | Enabled      | Disabled     
 **Regular sale**.            |             |             |              |              
 **Subscription renewal**     |             |             |              |
 **Free item sale**           |             |             |              |              
 **Save cards to account**    |             |             |              |              

And see all those work. Also while doing those; 
- check if the `wc-ajax=create_payment_intent` and `wc-ajax=create_setup_intent` endpoints return a encrypted version of the `client_secret` (not starting with the payment intent ID) when the encryption is enabled, 
- and if they return the plain `client_secret` when the encryption is disabled.

> **Note**: Free item sale is done like this: 
>
> Create a free trial subscription product first;
> - Product data > Simple Subscription
> - Subscription price > 10$ every month (any price)
> - Free trial > 30 days (any number of days)
> 
> Add the product to the cart, sell it.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.1.0#client-secret-encryption-toggle


client/settings/advanced-settings/client-secret-encryption-toggle.js
```diff
@@ -0,0 +1,47 @@
...
+	};
+
+	return (
+		<CheckboxControl
+			label={ __(
+				'Enable UPE Public Key Encryption',
+				'woocommerce-payments'
+			) }
+			help={ __(
+				'Encrypt the public keys used in the checkout form to harden card testing on your store.',
```
allie500: > Encrypt the public keys used in the checkout form to harden card testing on your store.

What do you think about changing the wording to this:

> Encrypt the public keys used in the checkout form to harden your store against card testing attacks.

includes/admin/class-wc-rest-payments-settings-controller.php
```diff
@@ -556,6 +558,23 @@ private function update_is_multi_currency_enabled( WP_REST_Request $request ) {
 		update_option( '_wcpay_feature_customer_multi_currency', $is_multi_currency_enabled ? '1' : '0' );
 	}
 
+	/**
+	 * Updates customer Multi-Currency feature status.
```
allie500: Looks like the method description needs to be updated. What do you think about this for the description:

> Updates the '_wcpay_feature_client_secret_encryption' option.

allie500:
Thanks for working on this feature, Taha! It looks awesome. I left a couple inline comments for two things that jumped out to me when I was looking over the code changes. Nothing major. I haven't run through the tests yet, but I will do so on Monday morning.

tpaksu:
## Bundle size changes by package: 

- Trimmed down CryptoJS, only with AES, UTF and PKCS libraries, 
Size Change: +30 kB (+3%)
Total Size: 1.05 MB

- node:crypto package 
Size Change: +935 kB (+92%) 
Total Size: 1.96 MB

- aes-js package https://www.npmjs.com/package/aes-js
Size Change: +94.2 kB (+9%)
Total Size: 1.11 MB

client/settings/advanced-settings/client-secret-encryption-toggle.js
```diff
@@ -0,0 +1,47 @@
...
+	}, [] );
+
+	const handleClientSecretEncryptionStatusChange = ( value ) => {
+		updateIsClientSecretEncryptionEnabled( value );
+	};
+
+	return (
+		<CheckboxControl
+			label={ __(
+				'Enable UPE Public Key Encryption',
```
allie500: Since this feature is available on non-UPE checkout we should probably remove 'UPE' from the label string. What do you think?

allie500: As we discussed briefly earlier, since the client secret is only exposed to the client when UPE is enabled, we should  hide this checkbox feature on the Payments > Settings page if UPE is not enabled.

allie500 (commented):
I ran through your tests yesterday afternoon and they all passed. After each test, I checked that the correct `_wcpay_feature_client_secret_encryption` option value persisted. All unit tests passed as well.

| Client Secret Encryption | Enabled | Enabled | Disabled | Disabled |
| ------------------------ | ------- | ------- | -------- | -------- |
| UPE | Enabled | Disabled | Enabled | Disabled |
| Regular sale | Pass | Pass | Pass | Pass |
| Free item sale | Pass | Pass | Pass | Pass |
| Save cards to account | Pass | Pass | Pass | Pass |

Once you've got the changes in place to hide the feature checkbox from the **Payments > Settings** page when UPE is not enabled, I'll be happy to run through these tests for you again.

includes/class-wc-payments-utils.php
```diff
@@ -805,4 +805,25 @@ public static function format_explicit_currency(
...
+	 * Encrypts client secret of intents created on Stripe.
+	 *
+	 * @param   string $stripe_account_id Stripe account ID.
+	 * @param   string $client_secret     Client secret string.
+	 *
+	 * @return  string                 Encrypted value.
+	 */
+	public static function encrypt_client_secret( string $stripe_account_id, string $client_secret ): string {
+		if ( \WC_Payments_Features::is_client_secret_encryption_enabled() ) {
+			return openssl_encrypt(
```
jrodger, line 819: Is `openssl_encrypt` available in PHP without extensions? If not, what happens if it isn't available?

changelog/add-4774-cloak-client-secret
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Adds encryption to the exposed client_secret to harden card testing attacks
```
jrodger: `to harden against card testing attacks` would be clearer

client/checkout/api/index.js
```diff
@@ -9,6 +9,7 @@ import {
 	getPaymentRequestAjaxURL,
 	buildAjaxURL,
 } from '../../payment-request/utils';
+import { decryptClientSecret } from '../utils/upe';
```
jrodger: Since the changes cover non-UPE payments as well (if I've understood correctly?) then does it make sense for this function to be in the UPE namespace here?

jrodger (commented):
Some comments inline.

Am I correct in thinking that this doesn't make it impossible for an attacker to decrypt? With a small amount of knowledge about the code it seems trivial to decrypt the secret and make the same post request to Stripe? Is the idea here just to make it slightly harder?

What does the rollout plan look like? Would we advise certain merchants to enable it if they're seeing sustained card testing issues?

The testing plan looks good. Has it also been tested with:

* WooPay?
* TumblrPay? I don't anticipate them turning the feature on (although maybe they might need to), but we have seen settings enabled by mistake in the past that caused unexpected issues.
* There's mention of free items in the test plan, but has some testing been carried out with "normal" subscriptions? What happens if a subscription is created with this setting off and then renews?

changelog/add-4774-cloak-client-secret
tpaksu: Fixed in [76aae9f](https://github.com/Automattic/woocommerce-payments/pull/4782/commits/76aae9f0b2d3559463403b3ad23a69f20af52f15).

includes/class-wc-payments-utils.php
tpaksu: Thanks for pointing that out! After committing [4cd98a9](https://github.com/Automattic/woocommerce-payments/pull/4782/commits/4cd98a95bb03708d267e393e5de4bf1a1383e2f3), I'm now hiding the feature from people that don't have the PHP openssl extension installed/activated, and not having the `openssl_encrypt` method available in their servers. Even if it's enabled, it'll revert back to disabled if those conditions are not met. And it doesn't show the toggle on the advanced settings section.

client/checkout/api/index.js
tpaksu: Moved the function to `../utils/encryption` namespace on [b12bbfe](https://github.com/Automattic/woocommerce-payments/pull/4782/commits/b12bbfef1a3cac1688bce2359833dfe05957a23d), so it doesn't get confused when used out of the UPE context. And it can be extended and improved on that file now.

tpaksu:
Thanks @jrodger for the initial review,

> Am I correct in thinking that this doesn't make it impossible for an attacker to decrypt? With a small amount of knowledge about the code it seems trivial to decrypt the secret and make the same post request to Stripe? Is the idea here just to make it slightly harder?

Yes, it's an initial attempt of encrypting the client_secret exposed to the browser, and decrypting just before passing it to the Stripe.js methods. And it affects the responses of our `?wc-ajax=create_payment_intent` and `?wc-ajax=create_setup_intent` calls, the content of the stored session variable, and the redirection URL's containing the `client_secret`'s. It'll be improved in the next versions.

> What does the rollout plan look like? Would we advise certain merchants to enable it if they're seeing sustained card testing issues?

It's designed as disabled by default, and yes, we can offer that feature to people experiencing card testing attacks, and we can track them to see how effective it is, and whether the attackers found a way to overcome the encryption in their scripts. 

> The testing plan looks good. Has it also been tested with:
>
> WooPay?

I discussed it on p1667913927586549-slack-C022WMN88KG with the WooPay people, that there's only one instance of `client_secret` there, and if the feature stays off, nothing will be changed. I ran the E2E tests of WooPay and they passed with this branch being the merchant site.

> TumblrPay? I don't anticipate them turning the feature on (although maybe they might need to), but we have seen settings enabled by mistake in the past that caused unexpected issues.

Nope, I didn't test that actually. And I don't know how to test that using this branch of WCPay in a TumblrPay environment. I can create a test payment with TumblrPay, but I don't know how to change the WCPay instance that TumblrPay uses under the hood to this branch on my sandbox. 

> There's mention of free items in the test plan, but has some testing been carried out with "normal" subscriptions? What happens if a subscription is created with this setting off and then renews?

Both free trial and normal subscriptions have been tested with the E2E tests in `tests/e2e/specs/subscriptions/shopper` folder, and they all passed. But I'm not sure about the tests cover the renewal charges. I'll check it again.

client/settings/advanced-settings/client-secret-encryption-toggle.js
tpaksu: Instead hiding the checkbox, I removed the `UPE` word from the label on [4cd98a9](https://github.com/Automattic/woocommerce-payments/pull/4782/commits/4cd98a95bb03708d267e393e5de4bf1a1383e2f3).

includes/class-wc-payments-utils.php
jrodger: Wonderful, thanks. The extra defensiveness here feels worthwhile since `openssl_encrypt` being missing would cause fatal errors.

client/checkout/api/index.js
jrodger: Great 😄 This should help communicate clearly where the code is used (i.e. everywhere!)

jrodger (approved):
I've approved it from an approach point of view. Can I leave the final testing and review to you @allie500?

allie500 (approved):
Simple subscription renewal testing. No free trials.

| Client Secret Encryption | Enabled | Enabled | Disabled | Disabled |
| ------------------------ | ------- | ------- | -------- | -------- |
| UPE | Enabled | Disabled | Enabled | Disabled |
| Initial subscription order | Pass | Pass | Pass | Pass |
| Subscription renewal order | Pass | Pass | Pass | Pass |

Ran `npm run wp option get _wcpay_feature_client_secret_encryption` command after each order to ensure that the correct option value persisted. 

JS and PHP unit tests all passed on my local.

allie500:
Re-running the 13 failed tests. Looks like some temporary issue when they ran. All of them had the following as the last error when they ran:

```
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

allie500:
Looks like something else is going on with the tests...

```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function _wp_theme_has_theme_json_clean_cache_upon_upgrading_active_theme(), 1 passed in /tmp/wordpress/wp-includes/class-wp-hook.php on line 310 and exactly 2 expected in /tmp/wordpress/wp-content/plugins/gutenberg/lib/compat/wordpress-6.2/get-global-styles-and-settings.php:67
```

PR #4802, opened 2022-09-26
Release branch for 4.8.0

cesarcosta99:
```
= 4.8.0 - 2022-xx-xx =
* Add - Add bundle size check for PR's.
* Add - Allow subscription processing via WooPay.
* Add - Auto redirect logged in platform checkout users.
* Add - Remove deprecated beta headers from Stripe requests.
* Add - Send a few extra pieces of data when checking if a WooPay user exists.
* Add - StripeLink - prefill first and last names on checkout.
* Add - Timezone formatting for transaction filters.
* Fix - Ask for login when guest mode is disabled while checking out with WooPay.
* Fix - Change 'zero-cost-fee' to 'empty-order', for product_code while sending level 3 data for shipping-only orders, without products.
* Fix - Correct empty email error when StripeLink is active on checkout page.
* Fix - Fix off by one error.
* Fix - Fix the rate calculation when using Table Rate Shipping and per item or per line item calculation type.
* Fix - Fix trial subscription checkout without WooPay signing up.
* Fix - Hide button below email field at checkout, when StripeLink is disabled.
* Fix - Hide Link payment method for non-us accounts.
* Fix - Limit level 3 product code within 12 digits.
* Fix - Prevent circumstance where WCPay Subscriptions customers could be double charged when the WC Subscriptions extension is active.
* Fix - Retain test mode context in CRON jobs queued up while checking out.
* Fix - The feature flag for the task list in Payments > Overview page was not passed correctly. We now see the business details and reconnect wpcom user task when appropriate.
* Update - Add timezone formatting only in case client provides user timezone.
* Dev - Bump minimum required version of WooCommerce from 6.4 to 6.6.
* Dev - Removes gutenberg plugin installation from E2E environment setup.
* Dev - Update node to v16 and npm to v8.
* Dev - Update to add E2E tests for the Multi-Currency functionality.
```


cesarcosta99:
Regarding the failing checks:

- The `E2E Tests - Pull Request / WC - latest | blocks` check [is failing](https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-pull-request.yml) since yesterday, but it doesn't seem because of this branch. When it was first created, [the same check passed](https://github.com/Automattic/woocommerce-payments/actions/runs/3131349932).
- I don't know why the `Compressed Size / Bundle size check` check is failing, but I don't think it's a blocker for the release. It's a check to analyze the bundle size of the PR, and I didn't notice any abrupt change in the bundle size during the testing cycles.

ricardo (approved):
LGTM

PR #4810, opened 2022-09-28
Disable WooPay when cart does not need payment

### LINKED ISSUE #4635: 'wc is not defined' error in the browser console on checkout when buying a free product
### Describe the bug
There's an error in the browser console on the checkout page when purchasing a free product.

### To Reproduce

1. As a merchant, create a product with a $0 price
2. As a merchant, confirm WooPay is enabled
3. As a shopper, add this free product to your cart and proceed to check out
4. On the checkout page, open the browser console
5. Notice there's an error with the message `Uncaught ReferenceError: wc is not defined`

### Actual behavior
There's an error in the console when purchasing a free product.

### Expected behavior
I wouldn't expect having errors in the console in this scenario.

alefesouza:
Fixes #4635

* Disable WooPay when the cart does not need payment, as currently WCPay is not loaded but it loads WooPay scripts.

#### Testing instructions

* As a merchant, create a product with a $0 price
* As a merchant, confirm WooPay is enabled
* As a shopper, add this free product to your cart and proceed to check out
* On the checkout page, open the browser console
* No error message should be shown
* Add a trial subscription or another product that needs payment
* WCPay and WooPay should be loaded as expected

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -283,8 +283,10 @@ class WC_Payments {
 			include_once __DIR__ . '/multi-currency/wc-payments-multi-currency.php';
 		}
 
-		// // Load platform checkout save user section if feature is enabled.
-		add_action( 'woocommerce_cart_loaded_from_session', [ __CLASS__, 'init_platform_checkout' ] );
+		// Load platform checkout save user section if feature is enabled.
+		// It needs to be after calculate totals due to subscriptions with trial
+		// that can change the WC()->cart->needs_payment() value.
+		add_action( 'woocommerce_after_calculate_totals', [ __CLASS__, 'init_platform_checkout' ] );
 
 		// Init the email template for In Person payment receipt email. We need to do it before passing the mailer to the service.
 		add_filter( 'woocommerce_email_classes', [ __CLASS__, 'add_ipp_emails' ], 10 );
--- a/tests/unit/helpers/class-wc-helper-product.php
+++ b/tests/unit/helpers/class-wc-helper-product.php
@@ -57,9 +57,8 @@ class WC_Helper_Product {
 	}
 
 	/**
-	 * Create simple product.
+	 * Create free product.
 	 *
-	 * @since 2.3
 	 * @param bool $save Save or return object.
 	 * @return WC_Product_Simple
 	 */
--- a/tests/unit/helpers/class-wc-helper-product.php
+++ b/tests/unit/helpers/class-wc-helper-product.php
@@ -57,6 +57,37 @@ class WC_Helper_Product {
 	}
 
 	/**
+	 * Create simple product.
+	 *
+	 * @since 2.3
+	 * @param bool $save Save or return object.
+	 * @return WC_Product_Simple
+	 */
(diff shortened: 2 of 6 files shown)


tests/unit/helpers/class-wc-helper-product.php
```diff
@@ -56,6 +56,37 @@ public static function create_simple_product( $save = true ) {
 		}
 	}
 
+	/**
+	 * Create simple product.
+	 *
+	 * @since 2.3
```
brianyu28: What's the `2.3` here in reference too? If it's meant to be a WCPay version, this PR would likely end up in `4.9`?

brianyu28 (changes requested):
Thanks @alefesouza! This fixes the error for me for normal products, but I believe this is breaking WooPay free trial subscription checkout — when I add a subscription with a trial to my cart and go to the checkout page, WooPay isn't loaded because `WC()->cart->needs_payment()` is returning `false`. Are you able to replicate?

I'm wondering if perhaps the Subscriptions Core hook onto `woocommerce_cart_needs_payment` hasn't been registered yet at the time that we're making the call to `should_enable_platform_checkout`?

tests/unit/helpers/class-wc-helper-product.php
alefesouza: I was sure it work before 🤔 the trial subscriptions problem was fixed on https://github.com/Automattic/woocommerce-payments/pull/4810/commits/672fc7dfedc27ca20f7b126e5237852132c23af7 , also the `2.3` was copied from the function above, I've removed as we are not using the `@since` anymore.

includes/class-wc-payments.php
```diff
@@ -283,8 +283,10 @@ public static function init() {
 			include_once __DIR__ . '/multi-currency/wc-payments-multi-currency.php';
 		}
 
-		// // Load platform checkout save user section if feature is enabled.
-		add_action( 'woocommerce_cart_loaded_from_session', [ __CLASS__, 'init_platform_checkout' ] );
+		// Load platform checkout save user section if feature is enabled.
+		// It needs to be after calculate totals due to subscriptions with trial
+		// that can change the WC()->cart->needs_payment() value.
+		add_action( 'woocommerce_after_calculate_totals', [ __CLASS__, 'init_platform_checkout' ] );
```
brianyu28, line 289: I'm a little uncertain about potential consequences of this approach for a couple of reasons. While `woocommerce_cart_loaded_from_session` happens exactly once as a hook on `wp_loaded`, `woocommerce_after_calculate_totals` happens every time the cart's `calculate_totals` is called. As a result:
- Might it be possible that `calculate_totals` gets called multiple times in a single request, and as a result we end up taking actions like enqueuing scripts / setting up tracking multiple times?
- Since `calculate_totals` happens later in a request than `wp_loaded`, can we think of anything that might require `init_platform_checkout` to have been called earlier that would break now that it's later?

I think this will require some additional investigation/testing to verify.

alefesouza:
The problem doesn't occur anymore, apparently fixed by #4638.

PR #4820, opened 2022-09-28
Update default platform checkout host to pay.woo.com

hsingyuc:
Fixes [1146-gh-Automattic/woopay](https://github.com/Automattic/woopay/issues/1146)

Update default platform checkout host from testing site `host.docker.internal:8090` to `pay.woo.com`

#### Screenshots
// Before
before-merchant
before-woopay

// After
Screen Shot 2022-09-28 at 12 09 52 PM
Screen Shot 2022-09-28 at 12 10 00 PM

#### Testing instructions
1. Make a purchase via WooPay on your ephemeral site
1. Go to your merchant site -> Orders -> Item section -> Refund button -> Select Qty -> See refunded as image
1. Go to https://pay.woo.com/account/orders 
1. See order status shown Refunded

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -360,7 +360,7 @@ class WC_Payments {
 		add_action( 'woocommerce_woocommerce_payments_updated', [ new Allowed_Payment_Request_Button_Types_Update( self::get_gateway() ), 'maybe_migrate' ] );
 		add_action( 'woocommerce_woocommerce_payments_updated', [ new \WCPay\Migrations\Update_Service_Data_From_Server( self::get_account_service() ), 'maybe_migrate' ] );
 		add_action( 'woocommerce_woocommerce_payments_updated', [ '\WCPay\Migrations\Track_Upe_Status', 'maybe_track' ] );
-		add_action( 'woocommerce_woocommerce_payments_updated', [ '\WCPay\Migrations\Delete_Active_Webhook', 'maybe_delete' ] );
+		add_action( 'woocommerce_woocommerce_payments_updated', [ '\WCPay\Migrations\Delete_Active_Platform_Checkout_Webhook', 'maybe_delete' ] );
 
 		include_once WCPAY_ABSPATH . '/includes/class-wc-payments-explicit-price-formatter.php';
 		WC_Payments_Explicit_Price_Formatter::init();
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -356,9 +356,11 @@ class WC_Payments {
 		require_once __DIR__ . '/migrations/class-allowed-payment-request-button-types-update.php';
 		require_once __DIR__ . '/migrations/class-update-service-data-from-server.php';
 		require_once __DIR__ . '/migrations/class-track-upe-status.php';
+		require_once __DIR__ . '/migrations/class-delete-active-webhook.php';
 		add_action( 'woocommerce_woocommerce_payments_updated', [ new Allowed_Payment_Request_Button_Types_Update( self::get_gateway() ), 'maybe_migrate' ] );
 		add_action( 'woocommerce_woocommerce_payments_updated', [ new \WCPay\Migrations\Update_Service_Data_From_Server( self::get_account_service() ), 'maybe_migrate' ] );
 		add_action( 'woocommerce_woocommerce_payments_updated', [ '\WCPay\Migrations\Track_Upe_Status', 'maybe_track' ] );
+		add_action( 'woocommerce_woocommerce_payments_updated', [ '\WCPay\Migrations\Delete_Active_Webhook', 'maybe_delete' ] );
 
 		include_once WCPAY_ABSPATH . '/includes/class-wc-payments-explicit-price-formatter.php';
 		WC_Payments_Explicit_Price_Formatter::init();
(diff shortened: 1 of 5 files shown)


rafaelzaleski (approved):
It worked on my Sandbox. Thanks for the fix!

brianyu28 (commented):
Thanks @hsingyuc! This does work for me on a fresh site, though on an existing site I already had a "WCPay platform checkout order status sync" webhook for the wrong host, so webhook with the correct host was never created. I had to manually disable/re-enable WooPay in order to get the new webhook created — did you have to do the same thing for an existing site? And if so, should we also add some logic to ensure that sites that have the incorrect webhook delivery URL get their webhook updated to use the correct URL?

hsingyuc:
> on an existing site I already had a "WCPay platform checkout order status sync" webhook for the wrong host, so webhook with the correct host was never created. I had to manually disable/re-enable WooPay in order to get the new webhook created — did you have to do the same thing for an existing site? And if so, should we also add some logic to ensure that sites that have the incorrect webhook delivery URL get their webhook updated to use the correct URL?

@brianyu28, I made the change and below is the test instruction:
1. Go to WC -> settings -> Advanced -> Webhooks
1. Find the `Active` webhook and remember the ID number
1. Click edit to update the Delivery URL to `http://host.docker.internal:8090/wp-json/platform-checkout/v1/merchant-notification`
1. Go to option and update `woocommerce_woocommerce_payments_version`'s value to any number that is less than the WCPay version you are on.
1. Go back to WC -> settings -> Advanced -> Webhooks and refresh the page
1. Check both ID number and Delivery URL are updated

### Screenshots
//Before
before

//After
after

includes/migrations/class-delete-active-webhook.php
```diff
@@ -0,0 +1,37 @@
...
+ * Fires an event on plugin upgrade to delete the avtive webhook.
+ * Runs only once. We want to know whether existing install had it
+ * enabled before the current version.
+ */
+class Delete_Active_Webhook {
+	/**
+	 * Checks whether we should trigger the event.
+	 */
+	public static function maybe_delete() {
+		if ( version_compare( get_option( 'woocommerce_woocommerce_payments_version' ), '4.8.0', '>' ) ) {
```
brianyu28: Should this be `4.9.0` instead, since that's the next scheduled release?

includes/migrations/class-delete-active-webhook.php
```diff
@@ -0,0 +1,37 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Delete_Active_Webhook
+ *
+ * Fires an event on plugin upgrade to delete the avtive webhook.
+ * Runs only once. We want to know whether existing install had it
+ * enabled before the current version.
+ */
+class Delete_Active_Webhook {
```
brianyu28: Not a blocker, but wondering if this class should include mention of WooPay/platform checkout in its name (e.g. `Delete_Active_Platform_Checkout_Webhook`) just for clarity.

includes/migrations/class-delete-active-webhook.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace WCPay\Migrations;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Delete_Active_Webhook
+ *
+ * Fires an event on plugin upgrade to delete the avtive webhook.
```
brianyu28: Nit: typo `active`

brianyu28 (changes requested):
Thanks @hsingyuc! This worked for me (tested both with WooPay enabled and disabled, to be safe), left a few comments but otherwise looks good.

includes/migrations/class-delete-active-webhook.php
hsingyuc: Good call on the next version!  I think we should update it to >= 4.9.0 because the logic here is to compare 4.9.0 with the version they are on.

includes/migrations/class-delete-active-webhook.php
hsingyuc: Good call on the sanitized naming!

includes/migrations/class-delete-active-webhook.php
hsingyuc: Thank you! 😆

includes/class-wc-payments.php
```diff
@@ -356,9 +356,11 @@ public static function init() {
 		require_once __DIR__ . '/migrations/class-allowed-payment-request-button-types-update.php';
 		require_once __DIR__ . '/migrations/class-update-service-data-from-server.php';
 		require_once __DIR__ . '/migrations/class-track-upe-status.php';
+		require_once __DIR__ . '/migrations/class-delete-active-webhook.php';
```
brianyu28: This filename will need to be updated too, I'm currently getting a fatal error because of it.

brianyu28 (changes requested):

includes/class-wc-payments.php
hsingyuc: Thank you!  I just pushed the change.

brianyu28 (approved):
Works well, thanks @hsingyuc!

hsingyuc:
Thank you for the feedback, @brianyu28 and @rafaelzaleski!

PR #4823, opened 2022-09-29
Handle rate limit errors during the checkout process

### LINKED ISSUE #4801: Handle 429 responses with lock_timeout error
### Description
Sometimes Stripe responds with the code `429` when `confirmPayment()` is called. That's related to Stripe's internal issues, which do not have a timeline (of fix) for now.
They have suggested implementing next while they're working on a solution:
> if you receive a lock timeout, when the merchant code is calling `confirmPayment()` function and it error's out (in cases of `lock_timeout`), you can manually call `retrievePaymentIntent()` from Stripe.js to re-fetch the `PaymentIntent` yourself

### Additional context
Slack threads: 
- more info about Stripe response: p1664188787481779/1663167748.635789-slack-CGGCLBN58
- history of the issue: p1663721387669259-slack-C9976E5MJ

zmaglica:
Fixes #4801 

During the checkout process, sometimes Stripe API return an error with code that says that request is rate limited. In most cases, the payment is already processed and the only step that needs to be done is to redirect the customer to the order received page.

To address this issue, additional checks have been added that will monitor error code and if the error code is the one that tells us that rate limit is in place, an additional request will be made that will get payment intent. If payment intent is succeeded, the user will be redirected. If that's not a case, the original error message will be shown

#### Testing instructions

This flow will be difficult to test because it will be difficult to replicate stripe response with specific error code. One way to test it is to manually edit the code with a specific error code that is mentioned in the original issue. Make sure that payment intent succeeded and make sure that part of code that calls additional payment intent API request is called by manually setting error variable.

-------------------

**Post merge**

### DIFF
--- a/client/checkout/api/index.js
+++ b/client/checkout/api/index.js
@@ -400,7 +400,7 @@ export default class WCPayAPI {
 	}
 
 	/**
-	 * Updates a payment intent with data from order: customer, level3 data and and maybe sets the payment for future use.
+	 * Updates a payment intent with data from order: customer, level3 data and maybe sets the payment for future use.
 	 *
 	 * @param {string} paymentIntentId The id of the payment intent.
 	 * @param {int} orderId The id of the order.
@@ -442,6 +442,45 @@ export default class WCPayAPI {
 					throw new Error( error.statusText );
 				}
 			} );
+	}
+
+	/**
+	 * Confirm Stripe payment with fallback for rate limit error.
+	 *
+	 * @param {Object} confirmParams Confirm payment request parameters.
+	 * @param {Object|StripeElements} elements Stripe elements.
+	 * @param {string|null} paymentIntentSecret Payment intent secret used to validate payment on rate limit error
+	 *
+	 * @return {Promise} The final promise for the request to the server.
+	 */
+	async handlePaymentConfirmation(
+		confirmParams,
+		elements,
+		paymentIntentSecret
+	) {
+		const stripe = this.getStripe();
+		const confirmPaymentResult = await stripe.confirmPayment( {
+			elements,
+			confirmParams,
+		} );
+		if (
+			paymentIntentSecret &&
+			confirmPaymentResult.error &&
+			'lock_timeout' === confirmPaymentResult.error.code
+		) {
+			const paymentIntentResult = await stripe.retrievePaymentIntent(
+				paymentIntentSecret
+			);
+			if (
+				! paymentIntentResult.error &&
+				'succeeded' === paymentIntentResult.paymentIntent.status
+			) {
+				window.location.href = confirmParams.redirect_url;
+				return paymentIntentResult; //To prevent returning an error during the redirection.
+			}
+		}
+
+		return confirmPaymentResult;
 	}
 
 	/**
--- a/client/checkout/api/index.js
+++ b/client/checkout/api/index.js
@@ -447,15 +447,15 @@ export default class WCPayAPI {
 	/**
 	 * Confirm Stripe payment with fallback for rate limit error.
 	 *
-	 * @param {Object} confirmParams Confirm payment request parameters.
 	 * @param {Object|StripeElements} elements Stripe elements.
+	 * @param {Object} confirmParams Confirm payment request parameters.
 	 * @param {string|null} paymentIntentSecret Payment intent secret used to validate payment on rate limit error
 	 *
-	 * @return {Promise} The final promise for the request to the server.
+	 * @return {Promise} The payment confirmation promise.
 	 */
 	async handlePaymentConfirmation(
-		confirmParams,
 		elements,
+		confirmParams,
 		paymentIntentSecret
 	) {
 		const stripe = this.getStripe();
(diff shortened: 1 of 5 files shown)


client/checkout/api/index.js
```diff
@@ -444,6 +444,45 @@ export default class WCPayAPI {
...
 	}
 
+	/**
+	 * Confirm Stripe payment with fallback for rate limit error.
+	 *
+	 * @param {Object} confirmParams Confirm payment request parameters.
+	 * @param {Object|StripeElements} elements Stripe elements.
+	 * @param {string|null} paymentIntentSecret Payment intent secret used to validate payment on rate limit error
+	 *
+	 * @return {Promise} The final promise for the request to the server.
```
jbordonado: It's not used by the server afterward, it's just `The payment confirmation promise` (that is resolved only in case of an error, which would then be thrown right after and shown on the checkout page, with no interaction with the server)

client/checkout/classic/upe.js
```diff
@@ -379,12 +381,13 @@ jQuery( function ( $ ) {
...
-			const { error } = await api.getStripe().confirmPayment( {
-				elements,
-				confirmParams: {
+			const { error } = await api.handlePaymentConfirmation(
+				{
 					return_url: returnUrl,
 				},
-			} );
+				elements,
+				null
```
jbordonado, line 389: Should it be `paymentIntentClientSecret` here?

client/checkout/api/index.js
```diff
@@ -444,6 +444,45 @@ export default class WCPayAPI {
...
+	 */
+	async handlePaymentConfirmation(
+		confirmParams,
```
jbordonado: I would put `elements` as the first parameter, and then `confirmParams` to keep it aligned with the param order of `stripe.confirmPayment` method

client/checkout/api/index.js
zmaglica: I forgot to update it. I will fix it.

client/checkout/classic/upe.js
zmaglica: Nope. The error is already known and this line only triggers validation errors like it was previously mentioned in the comment above this line: `If UPE fields are not filled, confirm payment to trigger validation errors`. I could pass `paymentIntentClientSecret`, but it is not needed anyway so I decided to skip it to prevent any strange cases.

client/checkout/api/index.js
zmaglica: Noted. I will fix it.

client/checkout/api/index.js
```diff
@@ -444,6 +444,45 @@ export default class WCPayAPI {
...
+			paymentIntentSecret &&
+			confirmPaymentResult.error &&
+			'lock_timeout' === confirmPaymentResult.error.code
+		) {
+			const paymentIntentResult = await stripe.retrievePaymentIntent(
+				paymentIntentSecret
+			);
+			if (
+				! paymentIntentResult.error &&
+				'succeeded' === paymentIntentResult.paymentIntent.status
```
jbordonado, line 476: Do we also need to check against `requires_capture` and `processing`?

See in our gateway code: https://github.com/Automattic/woocommerce-payments/blob/ca4325592ba30bf2d15bd09c5d4e407e2a747fc2/includes/class-wc-payment-gateway-wcpay.php#L65

client/checkout/classic/upe.js
jbordonado: Tried it myself, and you are correct, we are expecting the payment confirmation to fail in that case (thus triggering the error messages below the incorrect payment fields), so the client secret is not needed

jbordonado (commented):
I left a few comments but the overall logic seems correct to me (we might need to check the intent against 2 other statuses, to be double-checked)

client/checkout/api/index.js
zmaglica: Based on the conversation with Stripe, their suggestion was to redirect when succeeded status is set.

client/checkout/api/index.js
zmaglica: Fixed.

client/checkout/api/index.js
zmaglica: Fixed.

client/checkout/api/index.js
jbordonado: Okay, but is that because we told them we wanted to do that in our message (see 5574871-zd-woothemes) or is this because this is truly the only status to check? I wouldn't be so sure.

Why wouldn't we want to redirect if the payment intent's status is `requires_capture`? There's no sense of payment confirmed in that case since the merchant explicitly wants to capture the payment manually later,  it's tricky (checking `succeeded` ensures everything seems correct on Stripe's end so we can proceed to the next page manually, but does `requires_capture` provide the same guarantees? Do we know every action required is done if we get that status? I don't know)

For the `processing` status, I'm not sure too but it looks like it's a matter of how much time does it take to process the payment, and apparently we know with that status that every required action is done too (see https://stripe.com/docs/payments/intents). So, to me, it would also make sense to redirect to the "Order received" page in that case.

What does @Automattic/transact-architecture think of this?

zmaglica: I will merge this PR and open a new issue to address this topic.

jbordonado (approved):
Let's merge that already and make a note to later check if we need to take into account `processing` and `requires_capture` in this new fallback logic

PR #4830, opened 2022-10-01
Support client-side intent confirmation via filter

dechov:

Adds a filter `wcpay_server_side_intent_confirmation'` allowing the `confirm` param to be omitted on intent creation for on-session payments. This applies to the non-UPE flow only (since UPE already/currently uses client-side confirmation), reuses the same code path that was built for the SCA flow (i.e. hash redirect), and enables Stripe.js to perform additional pre-confirmation actions if necessary.

The filter will be used in the same way as the existing filter `wcpay_force_network_saved_cards`, but I'm definitely open to suggestions on better ways to enable this customization (and on whether the filter should live at a higher level than the API client, for that matter) – cc @zmaglica et al.

I believe we _could_ enable client-side confirmation in all cases, but I wasn't prepared to make that change without further consideration, in part because there is a performance hit related to making the extra confirmation request (in many cases if not all).

(Can use an automated test, based on [this one](https://github.com/Automattic/woocommerce-payments/blob/ca4325592ba30bf2d15bd09c5d4e407e2a747fc2/tests/unit/wc-payment-api/test-class-wc-payments-api-client.php#L68) – or if it's moved to the caller code, perhaps a test based on [this one](https://github.com/Automattic/woocommerce-payments/blob/ca4325592ba30bf2d15bd09c5d4e407e2a747fc2/tests/unit/test-class-wc-payment-gateway-wcpay-process-payment.php#L698).)

#### Testing instructions

Add this somewhere (like [here](https://github.com/Automattic/woocommerce-payments/blob/ca4325592ba30bf2d15bd09c5d4e407e2a747fc2/includes/class-wc-payments.php#L390), or as a snippet):
```php
add_filter( 'wcpay_server_side_intent_confirmation', '__return_false' );
```
With "new checkout experience" disabled, place order on checkout and verify that the payment goes through. (Also, logs related to the intent should show an extra `/confirm` request.)

Can also test a renewal order and verify that it goes through as well (despite lack of client-side confirmation).

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1426,7 +1426,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				}
 			}
 
-			if ( 'requires_action' === $status ) {
+			if ( 'requires_action' === $status || 'requires_confirmation' === $status ) {
 				if ( isset( $next_action['type'] ) && 'redirect_to_url' === $next_action['type'] && ! empty( $next_action['redirect_to_url']['url'] ) ) {
 					$response = [
 						'result'   => 'success',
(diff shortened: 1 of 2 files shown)


dmvrtx:
Closing as not planned/implemented in another manner.

PR #4833, opened 2022-10-03
Bump WC versions for 4.9.0 release

zmaglica:

- Minimum required WooCommerce version bumped to 6.8.2 (L2 policy will be integrated with this release and WooCommerce 7 will be released in 8 days).
- WC tested up to bumped to 7.0.0
- Verify that E2E tests are passing. More info [here](https://github.com/Automattic/woocommerce-payments/actions/runs/3173926608/jobs/5170090762).

#### Testing instructions
- Verify that version numbers line up with the [WooCommerce release version numbers](https://developer.woocommerce.com/releases/).

### DIFF
--- a/woocommerce-payments.php
+++ b/woocommerce-payments.php
@@ -11,7 +11,7 @@
  * WC requires at least: 6.8
  * WC tested up to: 7.0.0
  * Requires at least: 5.8
- * Requires PHP: 7.2
+ * Requires PHP: 7.0
  * Version: 4.8.0
  *
  * @package WooCommerce\Payments
--- a/woocommerce-payments.php
+++ b/woocommerce-payments.php
@@ -8,10 +8,10 @@
  * Woo: 5278104:bf3cf30871604e15eec560c962593c1f
  * Text Domain: woocommerce-payments
  * Domain Path: /languages
- * WC requires at least: 6.6
- * WC tested up to: 6.9.0
+ * WC requires at least: 6.8
+ * WC tested up to: 7.0.0
  * Requires at least: 5.8
- * Requires PHP: 7.0
+ * Requires PHP: 7.2
  * Version: 4.8.0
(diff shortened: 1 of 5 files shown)


woocommerce-payments.php
```diff
@@ -8,10 +8,10 @@
  * Woo: 5278104:bf3cf30871604e15eec560c962593c1f
  * Text Domain: woocommerce-payments
  * Domain Path: /languages
- * WC requires at least: 6.6
- * WC tested up to: 6.9.0
+ * WC requires at least: 6.8
+ * WC tested up to: 7.0.0
  * Requires at least: 5.8
- * Requires PHP: 7.0
+ * Requires PHP: 7.2
```
achyuthajoy: @zmaglica We haven't updated the PHP version in the past releases. Is there a discussion somewhere?

achyuthajoy (commented):
Changes look good to me @zmaglica. I've left a comment regarding min PHP version.

woocommerce-payments.php
zmaglica: There isn't discussion, but I bumped it because of the latest supported PHP version in WooCommerce. I can revert it if you think that this is mistake.

achyuthajoy: I don't think it's a mistake @zmaglica, but this would force a merchant to upgrade PHP version to `7.2` if they're currently running `7.0` and might cause some problems if we bump this without prior notice.

Right now, anyone running PHP 7.0 or above can run WC Payments. So let's have a discussion before bumping this one. WDYT?

zmaglica: Make sense. I will revert PHP version change and add it to the retro post.

achyuthajoy (approved):
LGTM 🚢

achyuthajoy (approved):

PR #4840, opened 2022-10-04
Hide upload buttons, amend placeholders in submitted evidence form

### LINKED ISSUE #495: Update submitted evidence form
Submitted evidence form looks weird with file upload buttons. IMO following updates woulb be nice to have:

1. Hide upload buttons.
2. Show only file name if it was uploaded during submit.
3. Show some placeholder if the file was not uploaded for the field. E.g. `Evidence file was not uploaded.`
4. Don't show any placeholder for additional details, just hide the button.

@LevinMedia what do you think?

Screenshot 2020-03-03 at 13 44 56

nagpai:
Fixes #495 

When dispute details are submitted in the evidence form : 
* Hide upload buttons
* Show file name, if there was an evidence file uploaded
* If evidence file was not uploaded show a placeholder text - `Evidence file was not uploaded`
* Hide helper / placeholder texts in each section and field
* Show a notice mentioning details are read-only on the submitted form

### Screenshots

<details><summary>Before</summary>

image

</details>

<details><summary>After</summary>

image

image

</details>

#### Testing instructions

1. Do a test purchase and create a dispute transaction using[ Stripe test card for dispute](https://stripe.com/docs/testing#disputes) case - `4000 0000 0000 1976` 
2. Browse to `WP Admin > Payments > Disputes and click on the dispute that appears there
3. Click on `Challenge Dispute` button
4. On the page that appears next, select a Product Type in the dropdown. 
5. Make sure all fields appear correctly in the resulting form, with placeholder texts wherever applicable. Fields should be editable at this stage
6. After this step, try out three scenarios: 

##### a. Submit with all fields filled
i. Submit the evidence with all fields filled in the form. Make sure you also upload any small test file for each of the uploads. 
ii. After submission, browse to `Payments > Disputes` and select `All disputes` in dropdown. Select the recently submitted evidence. Click `View submitted evidence` button.
iv. You should  see a notice on top indicating the fields are all read only, similar to the screenshot: 
image
v. Text fields and date should show what you entered during submission, and it should not be editable
vi. There should be no upload buttons visible for files, instead, you should see file names of uploaded evidence for each field
vii. There should be no placeholder / help text indicating what a field / section is meant for. ( no longer helpful since fields are already submitted ) 

##### a. Submit with no fields filled

i. Do not fill up any field, other than those prepopulated ( User name , address etc ) . Do not upload any proof documents either.
ii. Submit the evidence, browse to `Payments > Disputes` and select `All disputes` in dropdown. Select the recently submitted evidence. Click `View submitted evidence` button.
iii. Text fields where there was no information submitted should show text - `No information submitted`
iv. File upload sections should not show any upload buttons and should show - `Evidence file was not uploaded`

-------------------

**Post merge**


nagpai:
Other than review of code changes, I request guidance on whether this will require any additional tests. Thanks.

kalessil:
In general, when we touch JS files in Pulsar, we evaluate if the file can be converted into TS.
Did we consider migrating `client/disputes/evidence/index.js` to TS? How much effort do we need for this?

client/disputes/evidence/index.js
```diff
@@ -109,7 +110,7 @@ export const DisputeEvidenceForm = ( props ) => {
 			onChange( field.key, value );
 		},
 		disabled: readOnly,
-		help: expandHelp( field.description ),
+		help: readOnly ? '' : expandHelp( field.description ),
```
kalessil: IMO, help tooltips should remain even if we are in read-only mode.

client/disputes/evidence/index.js
```diff
@@ -133,28 +134,53 @@ export const DisputeEvidenceForm = ( props ) => {
 			isLoading,
 			isDone,
 			error,
-			help: expandHelp( field.description ),
+			help: readOnly ? '' : expandHelp( field.description ),
```
kalessil: IMO, help tooltips should remain even if we are in read-only mode.

client/disputes/evidence/index.js
```diff
@@ -133,28 +134,53 @@ export const DisputeEvidenceForm = ( props ) => {
...
 		};
 	};
 
 	const composeFieldControl = ( field ) => {
 		switch ( field.type ) {
 			case 'file':
-				return (
+				return ! readOnly ? (
```
kalessil: This condition is inconsistent with `readOnly && ! evidence[ field.key ]`, why so (I'm probably lacking context here)?

kalessil (changes requested):
I left some comments, but also not sure which design was used for introducing read-only variations of the controls and the test used in there. Can you share a link to design, please?

client/types/disputes.d.ts
```diff
@@ -82,7 +82,7 @@ interface UploadFieldObject {
 	label: string;
 }
 
-export interface DisputeFileUpload {
+export interface FileUploadControlProps {
```
kalessil, line 85: Not sure if this is a good idea. @dmvrtx, @anu-rock, @KarlisJ WDYT?

client/disputes/evidence/index.js
nagpai: My intention was to keep the submitted dispute details page cleaner, by removing all helper text. By helper text, I mean this :

image

Since the merchant would've already read the instruction in that and submitted details, I thought there is not much utility for that text after submission.  Do you think it will be still helpful to have it, @kalessil ?

kalessil: This might be helpful, thanks for sharing your reasoning. Let's check with @rogermattic (who is our UI/UX guru) on keeping/hiding help text for disabled evidence upload.

client/disputes/evidence/index.js
nagpai: All other cases are text-based fields - text, text area and date. Here if the form is submitted ( indicated by `readOnly`) , and if there is no text information submitted ( indicated by `! evidence[ field.key ]` , leading to text submitted ), we show a static text mentioning `No information submitted` . 

In case of `file` upload field though, things are slightly different. 

We are testing only for `readOnly`. If the form is `readOnly` then we show a modified file upload component - [UploadedReadOnly](https://github.com/Automattic/woocommerce-payments/blob/54cfd573047adf349e074264ca2d967e4b3ade50/client/components/file-upload/index.tsx#L139). It is similar to the original `FileUploadControl` component, with just that all sections such as - dropzone, upload button, upload error handling, are stripped out. Only the file preview, showing file name is shown . If there is no file, then a static message is shown indicating evidence file was not uploaded.  

The `UploadedReadOnly` component [has the logic](https://github.com/Automattic/woocommerce-payments/blob/54cfd573047adf349e074264ca2d967e4b3ade50/client/components/file-upload/index.tsx#L151) to detect presence or absence of an uploaded file, and show message accordingly.

client/types/disputes.d.ts
nagpai: Renaming of this interface came up during a slack discussion - p1664438225993279-slack-C02C00HCJ8J - where it was suggested to rename this interface to make it consistent with the [`FileUploadControl`](https://github.com/Automattic/woocommerce-payments/blob/54cfd573047adf349e074264ca2d967e4b3ade50/client/components/file-upload/index.tsx#L22) component, that uses this interface. 

I took care to check all references of `DisputeFileUpload` and rename it to `FileUploadControlProps`

client/disputes/evidence/index.js
anu-rock: I agree with @kalessil's point here about keeping the hint texts visible even on submitted forms for (a) consistency, and (b) later reference. But we'll wait for @rogermattic's final call.

client/disputes/evidence/index.js
anu-rock: @nagpai While I agree with your reasoning for excluding the `! evidence[ field.key ]` check, I think the `readOnly` condition should be reversed here to make it consistent with the rest.

```javascript
return readOnly ? <UploadedReadOnly /> : <FileUploadControl />;
```

client/disputes/evidence/index.js
```diff
@@ -133,28 +134,53 @@ export const DisputeEvidenceForm = ( props ) => {
...
 					/>
+				) : (
+					<UploadedReadOnly
+						key={ field.key }
+						{ ...composeFileUploadProps( field ) }
+					/>
 				);
 			case 'text':
-				return (
+				return readOnly && ! evidence[ field.key ] ? (
```
anu-rock: I think there's scope to simplify this. While there's no problem with the code, I generally find reading conditions (especially the ternary operator ones) in JSX harder and verbose. We could instead use conditional props. For example:

```javascript
const displayAsReadOnly = readOnly && ! evidence[ field.key ];

<TextControl
	key={ field.key }
	label={ field.label }
	value={  displayAsReadOnly ? __(
		'No information submitted',
		'woocommerce-payments'
	) : null }
	disabled={ displayAsReadOnly }
    { ...( displayAsReadOnly ? {} : composeDefaultControlProps( field ) ) }
/>
```

anu-rock: Reference - https://www.geeksforgeeks.org/how-to-conditionally-add-attributes-to-react-components/

client/disputes/evidence/index.js
```diff
@@ -178,7 +206,11 @@ export const DisputeEvidenceForm = ( props ) => {
 			<Card size="large" key={ section.key }>
 				<CardHeader>{ section.title }</CardHeader>
 				<CardBody>
-					{ section.description && <p>{ section.description }</p> }
+					{ readOnly
```
anu-rock: This can be simplified as:

```javascript
{ readOnly && section.description && ( <p>{ section.description }</p> ) }
```

client/disputes/evidence/index.js
```diff
@@ -279,6 +324,7 @@ export const DisputeEvidencePage = ( props ) => {
 	return (
 		<Page isNarrow className="wcpay-dispute-evidence">
 			{ testModeNotice }
+			{ readOnly && ! isLoading ? readOnlyNotice : '' }
```
anu-rock:
```suggestion
			{ readOnly && ! isLoading && readOnlyNotice }
```

client/types/disputes.d.ts
anu-rock: Renaming the interface was my suggestion. I still think it greatly improves readability by making the name explicit.

anu-rock (changes requested):
I like the overall approach but I've left a few minor comments which will help improve readability.

client/disputes/evidence/index.js
dmvrtx: I'm not a JS master, but I don't think suggested shortcut follows the same logic: show `section.description` if it is available and if it is not in the `readOnly` mode.

client/disputes/evidence/index.js
nagpai: Maybe removing the verbs like `Provide` will help in the read only version. Although users would obviously be seeing the prominent message on top mentioning that the form is read-only. Will await the call from UI/ UX colleague(s) :) .

client/disputes/evidence/index.js
anu-rock: @dmvrtx Good catch! I missed this in a hurry. It should be:

```javascript
{ ! readOnly && section.description && ( <p>{ section.description }</p> ) }
```

client/disputes/evidence/index.js
nagpai: Thanks for catching that, @anu-rock! I have changed[ it here](https://github.com/Automattic/woocommerce-payments/pull/4840/files#diff-6730096ed91d184adb48f33188e02b51841e49c6472b39b89b69ba2e112124cbR144).

client/disputes/evidence/index.js
nagpai: This is a brilliant learn for me! It makes the code look better and reduces repetition too!  I have changed this [here](https://github.com/Automattic/woocommerce-payments/pull/4840/files#diff-6730096ed91d184adb48f33188e02b51841e49c6472b39b89b69ba2e112124cbR144) and[ here](https://github.com/Automattic/woocommerce-payments/pull/4840/files#diff-6730096ed91d184adb48f33188e02b51841e49c6472b39b89b69ba2e112124cbR179)

client/disputes/evidence/index.js
nagpai:
```
{ ! readOnly && section.description && ( <p>{ section.description }</p>) }
```
I think this works! Updated in the code.

client/disputes/evidence/index.js
nagpai: Updated this!  Also made changes based on this[ here](https://github.com/Automattic/woocommerce-payments/blob/6b9d9b34d64b808071a8e8aa0ee901a4de92bebf/client/disputes/evidence/index.js#L136-L137) and [here](https://github.com/Automattic/woocommerce-payments/blob/6b9d9b34d64b808071a8e8aa0ee901a4de92bebf/client/disputes/evidence/index.js#L112)

nagpai:
> In general, when we touch JS files in Pulsar, we evaluate if the file can be converted into TS.
> Did we consider migrating `client/disputes/evidence/index.js` to TS? How much effort do we need for this?

Thanks for pointing this out! I studied the file for a few days and found that the effort needed for the migration would be larger. Although the JS files are only two, there seems to be quite a few complex components that may need help from someone with more experience in TS. I am skipping TS migration for this PR for now.

client/disputes/evidence/index.js
nagpai: Reply from [UI / UX team](https://github.com/Automattic/woocommerce-payments/issues/495#issuecomment-1284366514) . Linking here for posterity.

tl;dr - They seem okay with either showing or hiding helper text, but leaning towards hiding to keep it clean for submitted forms.

client/types/disputes.d.ts
KarlisJ: The link to slack discussion is broken for me. But after spending some time understanding the changes, I'm ok with the decision to rename the interface.

KarlisJ (approved):
LGTM!

I've review recent changes that follow as improvements after longer discussions with other reviewers. The changes looks good and overall PR in good shape. 

Hats off to Nagesh for this PR, this isn't your good first issue PR anymore. A lot of though and details has gone into it.

client/disputes/evidence/index.js
rogermattic: @nagpai @kalessil  sorry for the late reply! I'm just looking at it now. I agree with what  @bradyv mentioned in [this other thread](https://github.com/Automattic/woocommerce-payments/issues/495#issuecomment-1284366514) 
and wondering if we could have the helper text visible in case the form is in the edit mode and then hidden, for simplicity when it's all submitted? Or would that be a too complex solution to implement?

nagpai: @rogermattic - No worries :). The PR in its current form, does hide the helper text once submitted. The helper text remains visible before submission, when editing is possible. Thanks to Brad and you for confirming this 👍

nagpai:
>Hats off to Nagesh for this PR, this isn't your good first issue PR anymore. A lot of though and details has gone into it. 

This means a lot! Thank you so much, @KarlisJ . I loved the conversations and learning through each comment on this PR too  🙌

client/disputes/evidence/index.js
rogermattic: Awesome:) thanks!

anu-rock (approved):
LGTM after recent changes 🎉 

I'd also like to echo what @KarlisJ said. Thanks for putting in a lot of thought and following up persistently.

client/disputes/evidence/index.js
nagpai: I am marking this conversation as resolved, since we have received recommendations from our colleagues in the UI / UX team ( Magda and Brad ) . Thanks for the guidance, @kalessil  and @anu-rock 🙌

client/types/disputes.d.ts
nagpai: Thanks, Karlis! I realized that if I copy links of threaded replies in slack, it does not resolve. I have replaced it with the link to the main message of the thread ( for future reference )

client/disputes/evidence/index.js
nagpai: Marking this as resolved, since it is along the same lines as the [conversation above](https://github.com/Automattic/woocommerce-payments/pull/4840#discussion_r990986152). Thanks for your review and guidance, Vlad!

client/types/disputes.d.ts
nagpai: I am marking this as resolved for now. Happy to revisit if needed.

nagpai: Marking this as resolved for now. Happy to revisit if needed.

Issue #823, opened 2020-08-17
Refactor the Gateway class


marcinbot:
Right now, the `WC_Payment_Gateway_WCPay` class is responsible for multiple things, among others:
* Rendering and handling the saving of the gateway settings
* Rendering the payment and card save form
* Processing the payments, both regular and 3ds, synchronous and AJAX
* Processing the refunds, charge captures etc.

While this happens mostly because it's an interface for the Woo core gateway, we could still split the above functionalities up, move them to their own services, and use this class only as a way to hook into Woo core when calling those services.

zmaglica:
It will be handled as part of the reengineering payment process project.

vbelolapotkov:
@deepakpathania routing this one to you based on the latest comment. I'm not 100% sure it's a good fit though. What do you think?

deepakpathania:
Not the best fit for us imo, could be handled by Architects to look at it from a central PoV and then create smaller issues for individual focus areas for different teams including us.

Removing from our board for now to request re-triage.

vbelolapotkov:
Ok, I'll keep it as misc for now.

PR #2530, opened 2021-07-22
Dynamic gateway title by available payment methods

FangedParakeet:
Fixes #2433 

#### Changes proposed in this Pull Request

# Dynamic Gateway Title and Some Extra Goodies

The focus of this PR is to update the title of the UPE payment gateway at checkout to make the title dependent on the number of payment methods enabled. If there is only one payment method enabled, the title of the gateway will just be the name of that payment method.

Gateway title single method
_Gateway title with only a single method enabled_

If there are two or more payment methods currently enabled, the title of the gateway will be named "Popular payment methods"

Gateway title multiple methods
_Gateway title with multiple methods enabled_

Also, note that there are now new tab selected styles: with multiple payment methods, when a payment method is selected, the tab will have a border/outline similar to the selected input states, instead of being shaded dark matching the pay button. Moreover fonts and styles for all text elements within the UPE should now be consistent with the existing theme. These changes should be reflected on all pages where the UPE is present, so this includes the following.
 - Class checkout page
 - Blocks checkout page
 - Pay for order page
 - Add payment method page
 - Change payment method page
 
Note that the UPE on the Blocks checkout in this PR is able to read theme styling from the stored transient, if this information has been saved earlier, but it cannot currently compute these styles on its own. This means that if you visit the regular checkout page before using the Blocks checkout, the Blocks checkout should be able to copy the stored themes; however, if you visit the Blocks checkout without any transient information available or with transient information for a different theme, the UPE will not reflect the current theme. This will be further explored and hopefully resolved in #2603.

In addition to all of the above, there are few changes here for the main WooCommerce payments settings screen (**WooCommerce > Settings > Payments > All payment methods**). The title of the payment gateway is now simply "WooCommerce Payments", with no subheading; the payment icons have been removed from this page; and the description has been updated to the following:
```
Payments made simple, with no monthly fees - designed exclusively for WooCommerce stores. Accept credit cards, debit cards, and other popular payment methods.
```

New payments settings screen
_New general payments settings screen_

#### Testing instructions

1. Please pull the changes in this PR, the latest version of the Payments Server, and enable the UPE using the WCPay Dev Tools.
2. Visit the regular checkout with a single payment method and confirm that the gateway title is the name of that payment method.
3. Visit the regular checkout with more than one payment enabled and confirm that the gateway title is now "Popular payment methods".
[...]

PR #4850, opened 2022-10-05
Refactor WC_Payment_Gateway_WCPay Part 1

leonardola:
Related to #823 
This PR is the first part of a series of refactors of the WC_Payment_Gateway_WCPay class.

#### Testing instructions
- With test mode active
- Go to WooCommerce > Payments
- Make sure the notice `Test mode active: All transactions are simulated. Customers can't make real purchases through WooCommerce Payments.` shows up

- Go to plugins
- Make sure the Settings link shows up and works

- Make [is_available_for_current_currency](https://github.com/Automattic/woocommerce-payments/blob/3d81b3467672c989c9143a16673adb4c4b0d72b1/includes/class-wc-payment-gateway-wcpay.php#L504) return false
- Make sure the `Unsupported currency: The selected currency is not available for the country set in your WooCommerce Payments account.` notice shows up on wp-admin

- As a shopper add a product to the cart
- Make sure you can see the card inputs
- Test again using UPE

- As a shopper buy any product
- As a shopper buy any product using WooPay

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-823-refactor-gateway-class
+++ b/changelog/fix-823-refactor-gateway-class
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Refactor WC_Payment_Gateway_WCPay part 1
--- a/includes/admin/class-wc-payments-admin-settings.php
+++ b/includes/admin/class-wc-payments-admin-settings.php
@@ -18,6 +18,17 @@ class WC_Payments_Admin_Settings {
 	private $gateway;
 
 	/**
+	 * Set of parameters to build the URL to the gateway's settings page.
+	 *
+	 * @var string[]
+	 */
+	private static $settings_url_params = [
+		'page'    => 'wc-settings',
+		'tab'     => 'checkout',
+		'section' => WC_Payment_Gateway_WCPay::GATEWAY_ID,
+	];
+
+	/**
 	 * Initialize class actions.
 	 *
 	 * @param WC_Payment_Gateway_WCPay $gateway Payment Gateway.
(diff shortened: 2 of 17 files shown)


malithsen (changes requested):
Nice work so far! 
This is not a full review, but I noticed some issues in my initial testing and wanted to get them out earlier. 

UPE checkout is broken. There is a reference to `parent::get_payment_fields_js_config` in `class-upe-payment-gateway` which we need to update to point to the correct class. This is breaking the UPE checkout for me at the moment.

While I did not manually test this, judging from the failing unit tests, it looks like wcpay onboarding is broken. It seems like we just need to update some broken references there as well.
```
WC_Payments_Account_Test::test_maybe_redirect_to_onboarding_account_error
Failed asserting that exception of type "Error" matches expected exception "Exception". Message was: "Class 'WC_Payments_Admin_Settings' not found" 
```

 Both above issues look straightforward to fix, and again great work in splitting the wcpay gateway class 👍

leonardola:
@malithsen the errors seen on the e2e tests here started to happen after the last merge from develop. Discussion at p1665665849068839-slack-C026J0AL8NS. I've ran the tests before merging and they were all passing. So I think this is good to be reviewed

malithsen (commented):
@leonardola New changes look good, and UPE checkout and onboarding are now functional. I noticed that wcpay-dev-tools is throwing an error as it relies on `WC_Payment_Gateway_WCPay::get_settings_url()`. Would there be other plugins that will break because of these changes? I am not sure of the best way to handle this, one option might be to keep the original public methods, but mark them deprecated and call the new functions internally.

leonardola:
@malithsen yesterday I created this [PR](74-gh-Automattic/woocommerce-payments-dev-tools) to fix WC Pay dev tools and forgot to mention it here.

I've re-added the public functions. 

I haven't re-added `add_all_payment_methods_logos_column` and `add_all_payment_methods_icon_logos` because they were only being used by hooks and they were removed on #2530

I didn't re-add `display_test_mode_notice` and `display_not_supported_currency_notice` because they were also only being used by hooks. 

All those functions don't seem like functions that would be used by other plugins but I am ok to re-add them again in case you think that's better.

malithsen (approved):
Looks great now. I didn't notice any more regressions. Let's bump the versions in function deprecation notices to the correct version before the merge.

PR #4854, opened 2022-10-06
Update webhook processing to override existing meta data

### LINKED ISSUE #4771: Unable to refund a successful charge after a blocked payment
### Describe the bug

If a payment is blocked by a Radar rule (such as a postal code check failure), but there's a subsequent successful charge, that successful charge may not be able to be refunded.

The event says `A request to create a refund failed` with the body:

> "message": "This PaymentIntent (pi_abcdefghijklmnopqrstuvwxyz) does not have a successful charge to refund."
"type": "invalid_request_error"

Reported in 5436424-zen, 5303793-zen, 5564572-zen

p1655764075000849-slack-C7U3Y3VMY

### To Reproduce

Difficult so far to reproduce. Blocking with a Radar rule and subsequently successfully charging the `pi_` has proven to work in testing. Merchants continue to encounter this, though.

### Actual behavior

Order shows as refunded in WooCommerce, but the refund has failed in Stripe.

### Screenshots

image

### Expected behavior

Order is refunded correctly.

deepakpathania:
Fixes #4771 

- Currently, we had a check for `! $order->get_meta( $key )` during webhook processing for updating metadata, which meant we only update it in case the key isn't present already for that order. This led to an issue where we might have stale data stored for order and an incoming webhook for that order wouldn't update it since the key already exists.
- This PR removes that check and thus overrides the metadata whenever there is an incoming webhook.

#### Testing instructions

* Create an order like you normally would.
* Update the metadata associated with the order, for example charge_id.
* Rerun the webhook for the `payment_intent.succeeded` event using the Stripe CLI
```
stripe events resend<EVENT_ID> --account=<ACCOUNT_ID>   
```
* Ensure the charge_id value is updated to the original value.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-webhook-processing-service.php
+++ b/includes/class-wc-payments-webhook-processing-service.php
@@ -385,17 +385,14 @@ class WC_Payments_Webhook_Processing_Service {
 			WC_Payments_Utils::ORDER_INTENT_CURRENCY_META_KEY => $currency,
 		];
 
-		$order_changed = false;
 		foreach ( $meta_data_to_update as $key => $value ) {
 			// Override existing meta data with incoming values, if present.
 			if ( $value ) {
-				$order_changed = true;
 				$order->update_meta_data( $key, $value );
 			}
 		}
-		if ( true === $order_changed ) {
-			$order->save();
-		}
+		// Save the order after updating the meta data values.
+		$order->save();
 
 		$this->order_service->mark_payment_completed( $order, $intent_id, $intent_status, $charge_id );
 
--- a/includes/class-wc-payments-webhook-processing-service.php
+++ b/includes/class-wc-payments-webhook-processing-service.php
@@ -387,7 +387,8 @@ class WC_Payments_Webhook_Processing_Service {
 
 		$order_changed = false;
 		foreach ( $meta_data_to_update as $key => $value ) {
-			if ( $value && ! $order->get_meta( $key ) ) {
+			// Override existing meta data with incoming values, if present.
+			if ( $value ) {
 				$order_changed = true;
 				$order->update_meta_data( $key, $value );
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -387,7 +387,8 @@ private function process_webhook_payment_intent_succeeded( $event_body ) {
 
 		$order_changed = false;
 		foreach ( $meta_data_to_update as $key => $value ) {
-			if ( $value && ! $order->get_meta( $key ) ) {
+			// Override existing meta data with incoming values, if present.
+			if ( $value ) {
 				$order_changed = true;
```
achyuthajoy: @deepakpathania Is `$order_changed` required since we explicitly update the incoming metadata? Also, we're not comparing whether the value has changed.

achyuthajoy (changes requested):
@deepakpathania I haven't tested the changes, but left a comment on the code change.

includes/class-wc-payments-webhook-processing-service.php
deepakpathania: It is unlikely, but if we get a webhook event for this where the `value` isn't present for these keys, then we will skip the metadata update in the loop and in that case the order save call can be avoided.
We could remove this as well and always save the order after this loop, WDYT?

achyuthajoy: Let's remove it @deepakpathania. Like you said it's highly unlikely that all the keys would be empty and we do have a value check to prevent clearing the existing value.

achyuthajoy (approved):
Tested changes and working as expected. Thanks for working on this @deepakpathania 

LGTM 🚢

PR #4890, opened 2022-10-10
Prevent auto redirection for different customer

Mayisha:
If email field is pre-populated, sending the email when checking platform checkout login session. If the email does not match the logged in user, then we won't auto redirect.
Also for pre-populated emails, showing the spinner while the login session check is in-progress.

#### Testing instructions
Steps are mentioned in the other PR.


harriswong (approved):
Tested this following the steps following the other PR. LGTM

PR #4919, opened 2022-10-12
Remove "new checkout experience" setting + Enable using WooPay and UPE at the same time

### LINKED ISSUE #4724: Multiple UPE - Remove "new checkout experience" setting if it is no longer needed
### Description

The "new checkout experience" setting is used because UPE consolidates all payment methods into 1. But the multiple UPE branch splits payment methods back to individual ones so the setting is no longer needed. Remove this setting. 

### Acceptance criteria
- Remove the "new checkout experience"

### Additional context
Check paJDYF-5dK-p2#comment-13768 for additional context.

gpressutto5:
Fixes #4724

Remove "new checkout experience" setting

This PR removes the need to enable UPE and any banner, button, or message related to disabling it or referencing "Early access". This was done because now UPE methods are separate methods on checkout, and the credit card element being used is the legacy one, which works with WooPay, so there's no need to ever disable UPE entirely, as disabling each method individually would achieve the same goal.

#### Testing instructions

1. Before checking out this branch, disable UPE for testing.
1. Go to WooCommerce Payments settings and notice there's no UPE banner, and all payment methods are displayed.
1. Enable some of these UPE methods and WooPay.
1. Change the store currency to EUR.
1. Make a purchase using a UPE method and ensure it works.
1. Make a purchase using WooPay and ensure it works.

-------------------

**Post merge**


includes/notes/class-wc-payments-notes-additional-payment-methods.php
```diff
@@ -1,139 +0,0 @@
-<?php
-/**
- * Display a notice to merchants that have WCPay installed (and don't have UPE enabled)
- * to inform them that they can enable additional payment methods.
```
Mayisha, line 4: In this implementation the methods are listed in the settings from the beginning but in disabled state. Shouldn't we still let the merchant's know that they can enable some additional payment methods?

tests/unit/admin/test-class-wc-rest-payments-settings-controller.php
```diff
@@ -252,7 +252,7 @@ public function test_get_settings_fails_if_user_cannot_manage_woocommerce() {
 	public function test_update_settings_request_returns_status_code_200() {
 		$request = new WP_REST_Request( 'POST', self::$settings_route );
 		$request->set_param( 'is_wcpay_enabled', true );
-		$request->set_param( 'enabled_payment_method_ids', [ 'card' ] );
+		$request->set_param( 'enabled_payment_method_ids', [] );
```
Mayisha, line 257: Just curious why it was made empty.

client/payment-methods/index.js
```diff
@@ -26,92 +18,32 @@ import {
...
 };
 
-const UpeSetupBanner = () => {
-	const [ , setIsUpeEnabled ] = useIsUpeEnabled();
-
-	const handleEnableUpeClick = () => {
-		setIsUpeEnabled( true ).then( () => {
-			window.location.href = getAdminUrl( {
-				page: 'wc-admin',
-				path: '/payments/additional-payment-methods',
```
Mayisha, line 73: After removing this, there is no other place in the codebase that can take user to this `/payments/additional-payment-methods` page. Though I can directly access the page with the URL which looks a little off without `step 1`

additional

If there is no button or any option to redirect user to that page, I think everything related to that page can also be deleted safely.

Mayisha (changes requested):
I am noting down observations though some of them are not related to this specific PRs or can be handled in a follow up ticket.

- If I disable the credit card/debit card method from the settings page, it still appears on the checkout page. This does not happen on `develop` branch.
- On the shortcode checkout everything looks fine but on the blocks checkout the payment methods does not load for me.
blocks error

- There was a `Learn More` link on the settings page and the additional methods page which would take the user to https://woocommerce.com/document/payments/additional-payment-methods/. Should we have this link somewhere on the settings for self guided user onboarding?
learn more 1
learn more 2

- We need to update the docs https://woocommerce.com/document/payments/additional-payment-methods/ (details and screenshots there) once these changes make their way to the production.

- On the `payment/additiona-payment-methods` page there [used to be a message](https://woocommerce.com/wp-content/uploads/2021/07/wcpay-additional-payment-methods-onboarding.gif) saying we are adding additional currencies to the merchant store to support the selected payment method. I think we somewhere should inform the merchants about this kind of changes.

I am hitting request changes in case you want to address some of them in this PR.

includes/notes/class-wc-payments-notes-additional-payment-methods.php
gpressutto5: Well thought. I think we can use the old task we had to enable additional methods, but instead of enabling UPE, it will be used only to review the enabled methods in the wizard.

tests/unit/admin/test-class-wc-rest-payments-settings-controller.php
gpressutto5: We used to use [this method](https://github.com/Automattic/woocommerce-payments/blob/346c3d5881241c9d87c938c386ebfaccc488106a/includes/class-wc-payment-gateway-wcpay.php#L3121-L3131), which returns `[ 'card' ]` with no additional logic.
Now we use [this one](https://github.com/Automattic/woocommerce-payments/blob/task/remove-unused-option/includes/payment-methods/class-upe-payment-gateway.php#L986-L1016), which will filter to return only methods that have a fee set in Stripe, and we can't mock them because we are testing an endpoint.

client/payment-methods/index.js
gpressutto5: Elizabeth asked us not to remove it so we can improve it and reuse it later. She also mentioned the tasks that I removed that redirect to this page. We can reuse them to solve the issue you raised above.

gpressutto5:
This feature has been postponed indefinitely. See this comment: https://github.com/Automattic/woocommerce-payments/issues/4724#issuecomment-1344986490

PR #4925, opened 2022-10-13
Add new GitHub workflows to manage pre-release and release packages

### LINKED ISSUE #4684: [Release automation] Package creation
### Description
As per the [release train model](https://wcpay.wordpress.com/dev-resources/release-train-model/), we need to create a test package each Monday, and there is a code freeze set to Sunday EOD of Week 2 (meaning a production package should be created right after the official code freeze we announce).

We could create a GitHub workflow to automate the package creation ([with arguments](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) to specify whether this is a prerelease package or a production package).

### Acceptance criteria

* We have a way to quickly create a test package
* Release branch, tag and version updates across multiple files are automatically handled every Sunday EOD of Week 2, and a PR is raised to merge changes in trunk
* When the release PR is merged to trunk, a workflow to build the release zip file and push it to the GH release/tag is triggered
* etc.

### Dev notes
For the release package, hereafter, an example of code freeze workflow from WooCommerce: https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/release-code-freeze.yml

jbordonado:
Fixes #4684 

- A new workflow `create-pre-release.yml` (see [here](https://github.com/Automattic/woocommerce-payments/actions/workflows/create-pre-release.yml)); inspired by the existing workflow `pre-release-publish-on-tag.yml`
  - To be triggered manually (with a required input `tagName` (e.g. `x.y.z-test-n`))
  - Create a branch `testing/x.y.z-test-n` and a tag `x.y.z-test-n`
  - Generate the changelog corresponding to version `x.y.z`
  - Bump the version header in `woocommerce-payments.php`
  - Build the plugin (with a zip file named woocommerce-payments.zip)
  - Create a GitHub release flagged as "prerelease" (i.e. not a stable release), associated with the tag `x.y.z-test-n` and using the generated changelog notes and zip file

- A new workflow `release-code-freeze.yml` (see [here](https://github.com/Automattic/woocommerce-payments/actions/workflows/release-code-freeze.yml))
  - Scheduled to run every Sunday at 1200 UTC and can also be triggered manually
  - Checks if we need a code freeze
     - Find out the current version to figure out the next version to be released
     - Assess whether a code freeze is needed by checking if a stable tag (`x.y.z`) or a test tag for week 2 (`x.y.z-test-2`) for the next release already exists. If we already have a test tag for week 2 and no stable tag, then we need a code freeze
  - If code freeze is needed:
    -  A release branch and tag are created
    - The changelog is generated with a release date based on the next Wednesday by default
    - Multiple files (package.json, woocommerce-payments.php, readme.txt) are updated with the next release version and changelog content 
    - A PR to merge changes from the new release branch to trunk is created

- A new workflow `create-release.yml` (see [here](https://github.com/Automattic/woocommerce-payments/actions/workflows/create-release.yml))
  - Triggered when a PR targeting trunk is merged
  - Get the release version from package.json
  - Build the plugin
  - Create a **draft** GitHub release with the zip file and the data from the pull request targeting trunk that triggered the workflow

Note that I also included a new workflow `build-zip-file.yml` (see [here](https://github.com/Automattic/woocommerce-payments/actions/workflows/build-zip-file.yml)) that is not directly relevant to the release automation project but seemed important to add because:
- Sometimes, we need to test the ongoing development and we don't want to wait for the next pre-release package (`x.y.z-test-n`) to be published on Mondays
- it only took me like 5 minutes to set it up and test that it works correctly because I just had to reuse the composite actions I created (setup-repo, build)

### Impact on the release process

The release process is described [here](https://wcpay.wordpress.com/release-train-model/).
Once merged, we can leave the new workflows disabled and progressively enable them so they can be used and tested for real for the next release (so not 4.9.0)

#### Package preparation for testing

We will just need to trigger manually the `create-pre-release.yml` workflow and pass the relevant tag name (e.g. `4.9.0-test-1` if we want to test the version 4.9.0 during Week 1 of the release cycle). The zip file to be used for testing will be available in the published GitHub pre-release.

#### Code freeze / release package preparation

The new `release-code-freeze.yml` workflow will run every Sunday 1200 UTC and, if it is the right timing (week 3), create a new release branch and a PR targeting trunk. We should receive a Slack notification to make us all aware that this happened. Note that the created PR will also contain a new commit updating the version in some files and copying the changelog in the README, which is something that we did manually at some point or that the `woorelease` tool used to do for us on the day of the release when using the `release` command

When the PR is ready to be merged, use the suggested command in the PR description to merge it instead of the GitHub UI (like described [here](https://fieldguide.automattic.com/developing-woocommerce-payments/client-deployment/)).  It will trigger the new workflow `create-release.yml`, which will publish a new release on GitHub with the zip file to be used for the release. Note that the GitHub release will be a DRAFT, because the new version is not yet released at this moment so we don't want to make it look like it is.

#### Release day

We used to let the `woorelease` tool handle everything for us:
- build the zip
- bump the version in the relevant files
- deploy to wp.org
- trigger the translations

With our new workflows, we now have the zip file to use for the release (so with version updated across multiple files) as soon as the release PR targeting trunk is merged. So not necessarily on the day of the release, but most likely a day before? This leaves us time to smoke test it without pressure.

On the day of the release, we will now use directly the `woorelease wporg:release` command instead of the `woorelease release` command.
You'll need to execute: 
1. `woorelease wporg:release --product_version=x.y.z --zip_file=<absolute_path_to_the_zip_file> https://github.com/Automattic/woocommerce-payments/tree/trunk` (simulate first, as usual; then add `--release=true` option when you are ready to release)
2. `curl --request POST https://translate.wordpress.com/api/import-new-release/woocommerce/woocommerce-payments/x.y.z`

#### Testing instructions

SInce these workflows are brand-new, they don't exist on `develop` and so I have been using the GitHub CLI to trigger them. I don't think it's possible to test the manual ones otherwise.
**Note**: I have been using `7.2.6` and `7.2.6-test-3` versions for my tests, to avoid confusion in case someone sees it before I clean it up

* Install GitHub CLI: `brew install gh`
* Checkout this branch
* Trigger a workflow from this branch and with inputs: `gh workflow run workflow.yml --ref add/4684-package-creation-workflow -f param=value`
* Trigger a workflow from this branch without any inputs: `gh workflow run workflow.yml --ref add/4684-package-creation-workflow`

If you need to delete a branch: `git push origin --delete release/x.y.z`
If you need to delete a tag: `git push origin --delete x.y.z`
If you need to delete a GitHub release (which is necessarily bound to a tag): `gh release delete x.y.z`

Of course, you can do all of this from the GitHub UI too.

-------------------

- [ ] Update the Slack webhook for the code freeze to post to the right channel (I set myself as the recipient for testing purpose)
- [x] Remove the commented out code in `create-pre-release.yml`, `release-code-freeze.yml` and `release-pr.yml` to have the action run on the right branch instead of this branch by default

**Post merge**


.github/workflows/create-pre-release.yml
```diff
@@ -0,0 +1,86 @@
...
+  create-release:
+    name: Create the pre-release
+    runs-on: ubuntu-20.04
+    env:
+      TAG_NAME: ${{ inputs.tagName }}
+      
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+#        with:
```
achyuthajoy: Can we remove these commented lines if not using them?

.github/workflows/release-code-freeze.yml
```diff
@@ -0,0 +1,145 @@
...
+    name: "Check that today is the day of the code freeze"
+    runs-on: ubuntu-20.04
+    outputs:
+      freeze: ${{ steps.check-freeze.outputs.FREEZE }}
+      releaseVersion: ${{ steps.next-version.outputs.RELEASE_VERSION }}
+      releaseDate: ${{ steps.next-version.outputs.RELEASE_PLANNED_DATE }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+#        with:
```
achyuthajoy: Can we remove these commented lines if not using them?

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,104 @@
...
+    outputs: 
+      branch: ${{ steps.create_branch_tag.outputs.branch-name }}
+    env:
+      TAG_NAME: ${{ inputs.tagName }}
+      RELEASE_DATE_STRING: ${{ inputs.releaseDateString }}    
+    
+    steps:
+      - name: Checkout WCPay repository
+        uses: actions/checkout@v3
+#        with:
```
achyuthajoy: Can we remove these commented lines if not using them?

.github/actions/create-branch-tag/action.yml
```diff
@@ -0,0 +1,46 @@
+name: Create a branch and a tag
+description: Create a branch and a tag from the version provided
+
+inputs:
+  tag-name:
```
achyuthajoy, line 5: Should any formatting be done on the tag name to keep it uniform? Like tag name shouldn't contain spaces.

.github/actions/generate-changelog/action.yml
```diff
@@ -0,0 +1,40 @@
+name: Generate the changelog for WooCommerce Payments
+description: Generate the changelog from the version provided
+
+inputs:
+  release-version:
+    description: "The release version for which the action should generate the changelog"
+    required: true
+  release-date:
```
achyuthajoy, line 8: What's the release date format if it's not `unreleased`?

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,104 @@
+name: "Release - Create PR to trunk"
+
+# This action will run inside the release code freeze workflow or when it is triggered manually
+on:
+  workflow_call:
+    inputs:
+      tagName:
+        description: 'The tag name (e.g. 4.5.0)'
```
achyuthajoy, line 8: As mentioned above, do we follow a naming convention for tag names?

.github/workflows/create-pre-release.yml
```diff
@@ -0,0 +1,86 @@
+---
+name: "Create a pre-release"
+
+# This action will run when it is triggered manually
+on:
+  workflow_dispatch:
+    inputs:
+      tagName:
```
achyuthajoy, line 8: As mentioned above, do we follow a strict naming convention for tag names?

achyuthajoy (changes requested):
Awesome work on automating the release process @jbordonado 🥳. I've left some minor comments. Let me know your thoughts.

I haven't ran the new workflows, but have verified the results of your test runs.

.github/workflows/create-pre-release.yml
jbordonado: Sure, the plan is to uncomment it once the PR is good to go.
I was forced to comment this bit because the new workflows only exist on this branch, so the only way to test some of them (which reuse other new workflows) is to checkout this branch `add/4684-...`.
Once the PR is merged, all the workflows will live in `develop` so we won't need to worry about that anymore.

I already added an action item for this at the bottom of the description:

> Remove the commented out code in create-pre-release.yml, release-code-freeze.yml and release-pr.yml to have the action run on the right branch instead of this branch by default

.github/workflows/release-code-freeze.yml
jbordonado: Yes, but rather uncomment it (see previous comment [here](https://github.com/Automattic/woocommerce-payments/pull/4925#discussion_r997854804))

.github/workflows/release-pr.yml
jbordonado: Yes, but rather uncomment it (see previous comment [here](https://github.com/Automattic/woocommerce-payments/pull/4925#discussion_r997854804))

.github/workflows/create-pre-release.yml
jbordonado: I uncommented this part. We want the pre-release workflow to create a testing/ branch from `develop`.

.github/workflows/release-code-freeze.yml
jbordonado: I uncommented this part. We want the code freeze workflow to execute from the `develop` branch

.github/workflows/release-pr.yml
jbordonado: I removed this part. We want the release-pr workflow to either work from the `develop` branch (the default one) if running inside the code freeze workflow or from the branch selected when manually triggered (I can imagine cases where we might want to create the release branch from a previous test branch if we're not comfortable including risky changes merged in week 3 for instance... Not too sure about this, but it doesn't hurt to leave that possibility).

.github/actions/create-branch-tag/action.yml
jbordonado: I added a step to check the format of the tag input.
Instead of letting this composite action figure out whether we are in a pre-release context (by checking if "test" is present in the tag), it is now up to the parent actions to pass that info. That way, we can check if the tag respects the expected format (release or test) here.
We trim the leading and ending whitespace before checking, and the trimmed value is set as an output to be reused bythe next steps of the parent actions.

If the format is not OK, we stop here, and the workflow will fail.

.github/actions/generate-changelog/action.yml
jbordonado: I added the expected format (i.e. YYYY-MM-DD) in the description
But there is no margin for error here, as the parent action will format it and pass it as input, and this workflow cannot be triggered manually (so with a manual input which could have been wrong)

.github/workflows/release-pr.yml
jbordonado: `x.y.z` for release tags and `x.y.z-test-n` for test tags

As mentioned [here](https://github.com/Automattic/woocommerce-payments/pull/4925#discussion_r998010554), I added a check for the format

.github/workflows/create-pre-release.yml
jbordonado: Yes, see [my reply above](https://github.com/Automattic/woocommerce-payments/pull/4925#discussion_r998013653)

.github/workflows/release-code-freeze.yml
```diff
@@ -0,0 +1,145 @@
...
+          if ( empty( $latest_version_with_release ) ) {
+            echo "::error::Unable to get the latest released version";
+            exit( 1 );
+          }
+          
+          // Because we go from 5.9 to 6.0, we can get the next major_minor by adding 0.1 and formatting appropriately.
+          $latest_version_as_float = (float) $latest_version_with_release;
+          $branch_major_minor      = number_format( $latest_version_as_float + 0.1, 1 );
+          $next_version            = $branch_major_minor.0;
+          echo "::set-output name=RELEASE_VERSION::$next_version" . PHP_EOL;
```
achyuthajoy, line 65: `set-output` command is being repaced with new environment files and this command will stop working from June 2023. It be best if we can replace them to the new format - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/release-code-freeze.yml
```diff
@@ -0,0 +1,145 @@
...
+
+          $next_wednesday = date('Y-n-d', strtotime('next Wednesday'));
+          echo "::set-output name=RELEASE_PLANNED_DATE::$next_wednesday" . PHP_EOL;
```
achyuthajoy, line 68: `set-output` command is being repaced with new environment files and this command will stop working from June 2023. It be best if we can replace them to the new format - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

achyuthajoy (changes requested):
@jbordonado I've requested 2 more changes to replace `set-output` command with the new one. 

Apart from that everything looks good to me.

.github/workflows/release-code-freeze.yml
achyuthajoy: After discussing with @jbordonado this will be handled as a separate PR, since the implementation is inside PHP.

.github/workflows/release-code-freeze.yml
achyuthajoy: After discussing with @jbordonado this will be handled as a separate PR, since the implementation is inside PHP.

achyuthajoy (approved):
LGTM 🚢 

Thanks for your awesome work on this @jbordonado 🥳

jbordonado:
I created #4946 to replace the two remaining instances of the `set-output` command.

After merging, I disabled the `release-code-freeze` and the `create-release.yml` workflows to avoid interference with the current release process until everything is in place and covered.
The rest of the new workflows can only be triggered manually, so we can leave them active.

jbordonado:
The release doc was updated w.r.t to test package creation: https://wcpay.wordpress.com/dev-resources/release-train-model/#prepare-package-for-testing

We can now use the new `create-pre-release.yml` workflow.
The old workflow `pre-release-publish-on-tag.yaml` that was triggered when a test tag was pushed to the repo is now disabled (and can be deleted in a few weeks once we made sure everything's smooth with the new workflow)

PR #4785, opened 2022-09-22
Add a request signature to user/exists platform checkout endpoint.

brettshumaker:
#### Changes proposed in this Pull Request

This change adds a few necessary pieces of information when checking if a platform checkout customer exists.

#### Testing instructions

* Apply this patch.
* Fill out the email address field on the shortcode checkout.
* You should see the two new items in the request payload.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4927, opened 2022-10-13
Add platform checkout signature endpoint

bborman22:

Add an endpoint to get platform checkout signature

After #4785 added a signature to the user exists endpoint, a follow up issue was created to create an endpoint to get this signature so that it could be retrieved when needed as the signature is time sensitive. The issue is that the signature can become stale if the user is on the checkout page for too long, so this solution is to add a way to retrieve the signature when it is needed so that there isn't any risk to it being stale.

There were two options to approach this, one was to retrieve the signature on an interval and the other was to retrieve it once it was needed. I opted to retrieve it once it was needed as a way to eliminate (possibly many) unnecessary requests to the merchant store as well as avoid some complexity with retrieving it at an interval. The downside to this approach is a slightly longer time to check the `users/exists` endpoint on the platform checkout. This seemed like a reasonable trade off to me.

The changes are straightforward and use the WC Ajax so that a nonce can be verified in the right context.

#### Testing instructions

- Checkout this branch locally with a working WooPay setup.
- Set up cart with an item and go to checkout.
- Fill out the email address field on the shortcode checkout.
- Confirm that there was a request to the merchant store at: `?wc-ajax=wcpay_get_platform_checkout_signature`.
- Confirm the request included an `_ajax_nonce`.
- Confirm the response was successful and contained a signature.
- Confirm that signature was then passed as a query parameter to the WooPay `user/exists` endpoint.
- Confirm that the user exists look up responded successfully (whether true or false).

**Test some failing requests**

- Modify line https://github.com/Automattic/woocommerce-payments/compare/add/1313-platform-checkout-signature-endpoint?expand=1#diff-d65a9401ef14963cb2c9f8d930e98c69f81e5832f50fe794934471c4fd1a8c4fR340 to have an incorrect AJAX URL (for example, I changed it to: `get_platform_checkout_signatureAAA`) and refresh the checkout page.
- Confirm that the request does not return successfully.
- Confirm that the `WooPay is unavailable...` error message is displayed below the email field.
- Confirm the spinner was removed.
- Confirm that there was NO request to the WooPay `user/exists` endpoint.
- Fix the previous typo in the AJAX action.
- Append some characters to the `_ajax_nonce` to simulate it failing and refresh the checkout page.
- Confirm that the request returns a 403.
- Confirm that the `WooPay is unavailable...` error message is displayed below the email field.
- Confirm the spinner was removed.
- Confirm that there was NO request to the WooPay `user/exists` endpoint.

-------------------

**Post merge**


mdmoore (approved):
This tests well. Nice work! 

It looks like E2E tests are failing and some conflicts have come up. I'll go ahead and approve, but let me know if it needs any further testing.

PR #4949, opened 2022-10-20
Authorizations: Replace mocked data in reducer/store with actual fetching + add tests

elazzabi:
This is a PR part of a list of PRs to work on the Authorizations listing in Transactions.

**These changes are not customer facing, the authorizations screen is still behind a flag.**

This PR replaces the mocked data we have in the authorizations reducer/store with actual fetching now that we have those API endpoints ready. It also adds testing suites to those changes.

Questions for PR author:
- How can this code break? This code isn't used yet, as the authorizations screen is behind a flag
- What are we doing to make sure this code doesn't break? .

#### Testing instructions

No manual testing needed, it's already covered with automated tests.

-------------------

**Post merge**


client/data/authorizations/actions.ts
```diff
@@ -25,6 +25,24 @@ export function updateAuthorizations(
 	};
 }
 
+export function updateErrorForAuthorizations(
+	query: Query,
+	data: Authorization[],
+	error: Error
```
mgascam, line 30: Do you know where the interface/type `Error` is defined? I am asking because I could not find it i. I had to implement this in [other PR](https://github.com/Automattic/woocommerce-payments/pull/4961/files#diff-a4dc3e95060d774aa48b453ce22912c42d9526e21285deeceaacf25c1675815dR36) but used ApiError interface.

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,57 @@ import {
...
-	useSelect( ( select ) => {
-		const {
-			getAuthorizations,
-			getAuthorizationsError,
-			isResolving,
-		} = select( STORE_NAME );
+export const useAuthorizations = ( {
+	paged,
+	// eslint-disable-next-line @typescript-eslint/naming-convention
+	per_page,
```
mgascam, line 20: I'd suggest to use an [alias](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to make eslint happy.
```suggestion
	per_page: perPage,
```

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,57 @@ import {
...
+				getAuthorizationsError,
+				isResolving,
+			} = select( STORE_NAME );
 
-		const query = {};
+			const query = {
+				paged: Number.isNaN( parseInt( paged ?? '', 10 ) )
+					? '1'
+					: paged,
+				per_page: Number.isNaN( parseInt( per_page ?? '', 10 ) )
```
mgascam:
```suggestion
				perPage: Number.isNaN( parseInt( per_page ?? '', 10 ) )
```

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,57 @@ import {
...
+		( select ) => {
+			const {
+				getAuthorizations,
+				getAuthorizationsError,
+				isResolving,
+			} = select( STORE_NAME );
 
-		const query = {};
+			const query = {
+				paged: Number.isNaN( parseInt( paged ?? '', 10 ) )
```
mgascam: I'd suggest to extract the default value into a constant. There are other default values that could be extracted into constants below.

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,57 @@ import {
...
-		const {
-			getAuthorizations,
-			getAuthorizationsError,
-			isResolving,
-		} = select( STORE_NAME );
+export const useAuthorizations = ( {
+	paged,
+	// eslint-disable-next-line @typescript-eslint/naming-convention
+	per_page,
+	orderby,
```
mgascam, line 21: This is something minor: I'd rather have orderBy

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,57 @@ import {
...
 } =>
 	useSelect( ( select ) => {
-		const { getAuthorizationsSummary, isResolving } = select( STORE_NAME );
+		const {
+			getAuthorizationsSummary,
+			getAuthorizationsSummaryError,
+			isResolving,
+		} = select( STORE_NAME );
 
 		const query = {};
```
mgascam: Shouldn't we use the passed query parameter? I mean, instead of redefining it here?

client/data/authorizations/reducer.ts
```diff
@@ -18,10 +21,10 @@ const receiveAuthorizations = (
 		data = [],
 		error,
 	}: {
-		type: string;
-		query: Query;
-		data: Authorization[];
-		error: string;
+		type?: string;
```
mgascam: I am not convinced of marking `type`, `query`, `data` and `error` as optional. I went for a different approach, feel free to have a look [here](https://github.com/Automattic/woocommerce-payments/pull/4961/files#diff-fe9d6377e0f1c7629a6c3c5a165531a3f3eca75514570fe6a9e174f3b2298845R23)

mgascam (commented):
Nice!

I finished first review round and left a couple of minor-ish comments, I'll review again after the weekend with a fresh mind :)

client/data/authorizations/actions.ts
```diff
@@ -25,6 +25,24 @@ export function updateAuthorizations(
 	};
 }
 
+export function updateErrorForAuthorizations(
+	query: Query,
+	data: Authorization[],
```
mgascam: AFAIU `data` is not used, do you think we can remove it?

client/data/authorizations/actions.ts
```diff
@@ -39,3 +57,21 @@ export function updateAuthorizationsSummary(
 		query,
 	};
 }
+
+export function updateErrorForAuthorizationsSummary(
+	query: Query,
+	data: AuthorizationsSummary,
```
mgascam: Same as comment above, can we remove `data`?

mgascam (commented):
New review round completed.

client/data/authorizations/hooks.ts
elazzabi: The reason why I kept this and used the `eslint-disable` instead is mainly the typing.

The `Query` type (`client/transactions/declarations.d.ts`) has a `per_page` parameter, and returning `perPage` will raise typing errors  in many places that needs to be also silenced/worked on. I think fixing typing errors and introducing workarounds is not worth the effort compared to a one `eslint-disable` here.

What do you think?

client/data/authorizations/actions.ts
elazzabi: I personally added an Error and VS Code automatically added one. After seeing which one is included, it looks like a default one that ships with TypeScript:

```
interface Error {
    name: string;
    message: string;
    stack?: string;
}
```
Not sure if we can keep this or if I should move to another one like you used ^

client/data/authorizations/hooks.ts
elazzabi: Same as ^, the `orderby` is defined in the `Query` type `client/transactions/declarations.d.ts`, so I think changing it would require changing it everywhere. Unless we want to rename it in the deconstruction `orderby: orderBy` but I don't think that's going to add any value?

client/data/authorizations/hooks.ts
elazzabi: Oops my bad, thank you for catching it! Fixed in [2ca14f6](https://github.com/Automattic/woocommerce-payments/pull/4949/commits/2ca14f65272d5e7659519430df5d05dbaaa17d7c)

client/data/authorizations/reducer.ts
elazzabi: Hmm interesting. At first glance, it looks like this will change the signature of the reducer (ie moving away from `reducer( state, { type, error, data } )`) but I'll give it a fresh look tomorrow just to be sure :)

client/data/authorizations/actions.ts
elazzabi: I agree, this should be removed. Done in 70c0d227

client/data/authorizations/actions.ts
elazzabi: Done part of 70c0d227

client/data/authorizations/actions.ts
mgascam: I am not TS expert, but I'd expect to have the interface defined somewhere, or at least explicitly imported? It seems magic to me. I am concerned that maybe this interface is available in our local environment, but not in other environments. But it does not fail in the CI so 🤷‍♂️ it seems we are good.

client/data/authorizations/hooks.ts
mgascam: It's okay to stick with the `eslint-disable`, but please note that using  an alias (`per_page: perPage`) allows you to treat the var as perPage within the scope of the function without modifying it to 'outside world'. What I am trying to say is that you can work with the perPage alias within the function, making eslint happy, while in the TS type it remains as per_page, or whatever you need.

client/data/authorizations/hooks.ts
mgascam: Yep, feel free to skip.

mgascam (approved):
LGTM!

client/data/authorizations/hooks.ts
elazzabi: I only raised the TS issue because I've seen your suggestion to change the `per_page` in the `query` object too. Which will raise an error given it will be different than the `Query` type.

```js
// This will raise errors as the query object should be casted to Query type
perPage: Number.isNaN( parseInt( per_page ?? '', 10 ) )
```

But I agree with you that changing it inside the function only should be fine.

client/data/authorizations/hooks.ts
elazzabi: Agree! Changed in [bc2145f](https://github.com/Automattic/woocommerce-payments/pull/4949/commits/bc2145f990e31e4abcd4ef94e8fb4139d9a04604)

client/data/authorizations/hooks.ts
```diff
@@ -14,34 +14,56 @@ import {
...
+	useSelect(
+		( select ) => {
+			const {
+				getAuthorizations,
+				getAuthorizationsError,
+				isResolving,
+			} = select( STORE_NAME );
 
-		const query = {};
+			const pagedQuery = Number.isNaN( parseInt( paged ?? '', 10 ) );
```
anu-rock, line 32: A first glance, it's a bit hard to understand. IMO, it's uncommon to use `isNaN` like this. Also, is it necessary to specify the base in `parseInt`?

client/data/authorizations/reducer.ts
anu-rock: I like what Miguel has done in that PR. Typing for `action` is more explicit and readable.

client/data/authorizations/resolvers.ts
```diff
@@ -3,76 +3,66 @@
...
+	updateErrorForAuthorizationsSummary,
+} from './actions';
 import { Query } from '@woocommerce/navigation';
 
 export function* getAuthorizations( query: Query ): any {
-	const getMockedRows = () => {
-		const randomAmount = () => {
-			return 1400 + Math.floor( Math.random() * 5000 );
-		};
+	const path = addQueryArgs( `${ NAMESPACE }/authorizations`, {
```
anu-rock, line 23: TIL `addQueryArgs`

client/data/authorizations/resolvers.ts
```diff
@@ -3,76 +3,66 @@
...
+		direction: query.order,
+		page: query.paged,
+	} );
 
-		const randomDate = dateI18n(
-			'M j, Y / g:iA',
-			moment.utc( new Date() ).local().toISOString()
+	try {
+		const results = yield apiFetch( { path } );
+		yield updateAuthorizations( query, results.data || [] );
```
anu-rock: I'm pretty sure we can use the null coalescing operator `??` here as in an earlier file.

anu-rock: Same's true for other places below.

I see `??` as an improvement over `||` since it specifically checks for null and undefined. So, if we aren't expecting other falsy values here ('', 0, etc.), `??` should be used. It reduces the uncertainty in what's expected in response.

What do you think?

anu-rock (commented):
The changes are looking quite good. I've left a few minor remarks.

client/data/authorizations/hooks.ts
dmvrtx: I had the same question, went to MSDN and learned, that:

 - Passing base in `parseInt` helps against awkwardly passed arguments, e.g. `025`. Without base it would be treated as octal and become `32`, with base it will stay `25`.
 - There is no way to have a "default" with `parseInt`, and `Number.isNaN` helps with it further.

dpaun1985 (commented):
LGTM. Only code review.

client/data/authorizations/resolvers.ts
elazzabi: I agree. Should be fixed with cd906a56

client/data/authorizations/hooks.ts
anu-rock: Right. I checked MDN and got the answer to my second question. But my first observation still remains. It's a lot of number crunching to arrive at whether there's pagination in the URL. However, I'm not sure of a better way and don't have strong feelings about the current approach. So, feel free to skip.

client/data/authorizations/reducer.ts
elazzabi: Thank you! Fixed [b652639](https://github.com/Automattic/woocommerce-payments/pull/4949/commits/b65263975e9abcfba30aad048fa278d423107636)

client/data/authorizations/hooks.ts
elazzabi: I used this bit from transactions and it looks like it was added in [this commit](https://github.com/Automattic/woocommerce-payments/commit/02074c791acb2f72dda14c5a196d18ba84ce8d1d) as a strict fallback for query params. So I believe there were cases where this was breaking and the strict callback was introduced.

I also vote to keep it even if it doesn't look "pretty" 😅

anu-rock (approved):
LGTM after the new changes 🎉

PR #4688, opened 2022-09-02
Add Capture Charge button on payment details page

anu-rock:
Fixes 

#### Changes proposed in this Pull Request

* Add a button on Payment Details page to capture an authorized payment that is yet to be finalized.

A glimpse at the capture flow: (pending final changes; see TODO below)  

https://user-images.githubusercontent.com/1288616/188133926-de5b1309-ef35-4243-b6cc-5d21adfaea1e.mp4

#### Testing instructions

* Go to WP Admin → Payments → Settings and make sure the checkbox "Issue an authorization on checkout, and capture later" is checked.
* Create a new order using WC frontend. (eg. buy a Beanie using a [test card](https://woocommerce.com/document/payments/testing/#test-card-numbers))
* Open the order details page from WP Admin → WooCommerce → Orders.
* Click the payment intent link to visit the order's payment details page.
* Verify that you see a notice "You need to capture this charge before" along with a capture button.
* Click the Capture button.
* Verify that the payment is successfully captured. (the notice disappears, a success message appears, timeline is refreshed)
* Refresh the page and verify that the capture notice doesn't appear.

#### TODO

- [x] Notice background color as per design.
- [x] Remove mockings.
- [x] Get authorization id from query params.
- [x] Streamline `submitCaptureAuthorization` to support both "Uncaptured transactions" and "Payment details" pages.
- [x] Fix broken tests
- [x] Add supporting tests

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #4960, opened 2022-10-20
[Authorizations][Capture from table] Add CaptureAuthorizationButton component

### LINKED ISSUE #4569: [Authorizations] Capture authorization from table
### Description

As part of an effort to improve the capture payment workflow, we will add a new 'Authorizations' tab to the transactions page. This new tab will contain a table with pending authorizations. . We will also add a Capture charge button to each row of the table to allow merchants to capture it without having to navigate to the Order details screen.
### Acceptance criteria

* Add an 'Action' column to the authorizations table.
* Add a 'Capture' button to each row of the table.
* Payment is captured when the CTA is clicked. It should show a 'spinner' while working on the request
* Show a notification to inform the merchant about the outcome of the request: success or error.

### Designs

Screen Shot 2022-08-11 at 16 40 36

Screen Shot 2022-08-11 at 16 42 40

### Testing instructions

TBD

mgascam:
Fixes #4569

_These changes come from a [bigger PR](https://github.com/Automattic/woocommerce-payments/pull/4676) that we decided to split into smaller chunks. I created an [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962) where we will merge all split PRs, including this one._

This PR adds a new CaptureAuthorizationButton React component. It will be used to allow merchants to capture a payment from the authorizations list and from the payment details screen.

#### Testing instructions

No manual testing needed, changes are covered by unit tests. We will test manually the [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962), once all partial PRs are merged.

-------------------


client/data/authorizations/action-types.ts
```diff
@@ -2,6 +2,7 @@
 
 export default {
 	SET_AUTHORIZATIONS: 'SET_AUTHORIZATIONS',
+	SET_AUTHORIZATION: 'SET_AUTHORIZATION',
```
elazzabi, line 5: This looks completely fine, just wanted to raise an idea for our future selfs: Do we want to make the difference between this and the plural version a little more obvious? The current difference is only the (S), which may make things difficult when reading quickly 😅 . An example of something a little more obvious would be `SET_INDIVIDUAL_AUTHORIZATION`

elazzabi (approved):
Looks good for me but left a question.

Another question given the reducer isn't worked on here. What's the use case of adding individual authorizations to the store? Is this would be used afterwards to update the `authorizations[]` array in the store?

client/components/capture-authorization-button/test/index.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+			authorization: {} as Authorization,
+		} );
+
+		renderCaptureAuthorizationButton( 42, 'paymentIntentId' );
+
+		expect( doCaptureAuthorizationMock.mock.calls.length ).toBe( 0 );
+
+		await user.click( screen.getByRole( 'button' ) );
+
+		expect( doCaptureAuthorizationMock.mock.calls.length ).toBe( 1 );
```
elazzabi, line 94: One extra thing here: Should we also test if the state of the button switches to `busy` after we click on it + after the capture function is called?

client/data/authorizations/action-types.ts
mgascam: I like your suggestion 👍

client/components/capture-authorization-button/test/index.test.tsx
mgascam: I think it is a good idea. I made an attempt to test it in this commit [4adfb2f](https://github.com/Automattic/woocommerce-payments/pull/4960/commits/4adfb2f668e2dc3b52d181a1340ef99076006724). I am not fully convinced of the test though. The problem is that this component relies on the `useAuthorization` hook to populate the `isBusy`  prop of the button (via isLoading). So I need to change manually the `isLoading` prop and then rerender the component.  Please have a look and let me know what you think.

mgascam:
> Looks good for me but left a question.
> 
> Another question given the reducer isn't worked on here. What's the use case of adding individual authorizations to the store? Is this would be used afterwards to update the `authorizations[]` array in the store?

Good question @elazzabi. This is going to be used in the [payment details screen](https://github.com/Automattic/woocommerce-payments/pull/4688). We are adding a Capture button and a notice to the merchant if the payment needs to be captured. So we need a way to access the store to retrieve a single authorization and also commit it to the store when it is captured. Hope this answer your question, let me know otherwise.

client/components/capture-authorization-button/test/index.test.tsx
elazzabi: I think the mock function you used is clear and communicates what we are trying to test. I think it's good to keep 👍

client/data/authorizations/action-types.ts
anu-rock: While I agree with @elazzabi's suggestion, I'm worried that would make our action names inconsistent with existing ones.

anu-rock (approved):
LGTM 👍

And as I've seen it integrated into the main flow, I don't have anything to add here.

PR #4957, opened 2022-10-20
[Authorizations][Capture from table] Add capture_authorization endpoint

mgascam:
Fixes #4569 

_These changes come from a [bigger PR](https://github.com/Automattic/woocommerce-payments/pull/4676) that we decided to split into smaller chunks. I created an [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962) where we will merge all split PRs, including this one._

#### Changes proposed in this Pull Request

This PR adds a new `capture_authorization` endpoint. It will be used to allow merchants to capture a payment from the authorizations list.

#### Testing instructions

No manual testing needed, changes are covered by unit tests. We will test manually the [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962), once all partial PRs are merged.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

Issue #123, opened 2019-07-01
Associate each transaction with a Deposit


dechov:
Depends on https://github.com/Automattic/woocommerce-payments-server/pull/140

The designs call for a link from each row in the Transactions table to the Deposit it is associated with. As discussed in paJDYF-5c-p2 (subthread #comment-531):

> @dechov: we will have an available_on timestamp [[doc](https://stripe.com/docs/api/balance/balance_transaction#balance_transaction_object-available_on)], but not a reference / link to the Payout as far as I can tell.

> @DanReyLop: I found [this](https://stripe.com/docs/api/accounts/object#account_object-settings-payouts-schedule) and [this](https://stripe.com/docs/connect/payouts#payout-schedule) on the API docs. So, for a given account, we can know which “Payout schedule” it has. Considering that the payouts use UTC times to calculate “days”, we can calculate which payout is a given transaction going to be in, and which transactions are going to be in a given payout.

Note: this is related to another task of creating a Deposit view for sets of transactions that don't exist in Stripe as a Payout object yet, if we want to do that (see p1560792286063300-slack-wcpay).

PR #4961, opened 2022-10-20
[Authorizations][Capture from table] Data layer and UI components

### LINKED ISSUE #4569: [Authorizations] Capture authorization from table

mgascam:
Fixes #4569 

_These changes come from a [bigger PR](https://github.com/Automattic/woocommerce-payments/pull/4676) that we decided to split into smaller chunks. I created an [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962) where we will merge all split PRs, including this one._

This PR implements the front end data layer (aka Redux store) for authorizations. It also adds a 'Capture' cta button to each row of the authorizations list.

To facilitate reviews, I tried to organize the functionality into the following commits:

1. Implementation of Redux store in [a127e95](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/a127e95933ffbbfb29ef83179c95c4350ecb8fc7)
2. Add Capture button to authorizations list in [2430ff1](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/2430ff142384251c8b44225e488b564c20274d78)
3. Add CaptureAuthorizationButton component (please note that there is a [dedicated PR for this change](https://github.com/Automattic/woocommerce-payments/pull/4960))

#### Testing instructions

No manual testing needed, changes are covered by unit tests. We will test manually the [integration branch](https://github.com/Automattic/woocommerce-payments/pull/4962), once all partial PRs are merged.

-------------------

**Post merge**


changelog/add-4569-capture-from-table
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Add data layer and UI components
```
elazzabi, line 4: Given this shows in the public changelog, I think we should make it descriptive

client/data/authorizations/actions.ts
```diff
@@ -1,30 +1,62 @@
...
 	Authorization,
+	AuthorizationsList,
 } from 'wcpay/types/authorizations';
+import { STORE_NAME } from '../constants';
+import { ApiError } from 'wcpay/types/errors';
 
 export function updateAuthorizations(
 	query: Query,
-	data: Array< Authorization >
+	authorizationsList: AuthorizationsList
```
elazzabi: This will create some merge conflicts with the other PRs, but let's see how it goes 😄

client/data/authorizations/actions.ts
```diff
@@ -39,3 +71,58 @@ export function updateAuthorizationsSummary(
...
+		const result = yield apiFetch( {
+			path: `/wc/v3/payments/orders/${ orderId }/capture_authorization`,
+			method: 'post',
+			data: {
+				payment_intent_id: paymentIntentId,
+			},
+		} );
+
+		const authorization = {
+			payment_intent_id: ( result as any ).id,
```
elazzabi: can we use a type other than `any` here? Or I think `apiFetch`'s return value isn't typed?

client/data/authorizations/actions.ts
```diff
@@ -39,3 +71,58 @@ export function updateAuthorizationsSummary(
...
+			'invalidateResolutionForStoreSelector',
+			'getAuthorizationsSummary'
+		);
+		// Create success notice.
+		yield dispatch(
+			'core/notices',
+			'createSuccessNotice',
+			__( 'You have captured the payment.', 'woocommerce-payments' )
+		);
+	} catch ( error ) {
```
elazzabi, line 126: The catched `error` isn't used here and thus not exposed to the user. Should we include it part of the message/notice to give the user some context?

client/data/authorizations/resolvers.ts
```diff
@@ -3,76 +3,55 @@
...
+	yield updateAuthorization( {
+		payment_intent_id: paymentIntentId,
+	} as Authorization );
+}
 
-	yield updateAuthorizationsSummary( query, data );
+export function* getAuthorizationsSummary(
+	query: Query
+): Generator< unknown > {
+	yield updateAuthorizationsSummary( query, {} );
```
elazzabi, line 65: I think we are missing an api fetch here + error handling

client/data/authorizations/resolvers.ts
```diff
@@ -3,76 +3,55 @@
...
-		total: 100,
-		currency: 'USD',
-		store_currencies: [ 'USD' ],
-		customer_currencies: [ 'USD', 'EUR' ],
-		totalAmount: 126790,
-	};
+export function* getAuthorization(
+	paymentIntentId: string
+): Generator< unknown > {
+	yield updateAuthorization( {
```
elazzabi, line 56: I think we are missing an api fetch here + error handling?

client/data/authorizations/actions.ts
```diff
@@ -39,3 +71,58 @@ export function updateAuthorizationsSummary(
...
+		yield dispatch(
+			STORE_NAME,
+			'invalidateResolutionForStoreSelector',
+			'getAuthorizationsSummary'
+		);
+		// Create success notice.
+		yield dispatch(
+			'core/notices',
+			'createSuccessNotice',
+			__( 'You have captured the payment.', 'woocommerce-payments' )
```
elazzabi: I imagine customers clicking on multiple `Capture` buttons at the same time but getting the same generic success message which makes it hard to know which payment is captured. Should we make this a little more descriptive by including the order id for example? Something like: `Payment for order #123 captured successfully`

client/data/authorizations/actions.ts
```diff
@@ -39,3 +71,58 @@ export function updateAuthorizationsSummary(
...
+		);
+	} catch ( error ) {
+		const message = __(
+			'There has been an error capturing the payment. Please try again later.',
```
elazzabi: Same as ^, I think including the order id here will help better communicate the error

client/data/authorizations/test/authorizations.fixture.json
```diff
@@ -0,0 +1,122 @@
+[
+  {
+      "created": "Sep 7, 2022 / 11:00AM",
+      "captured": false,
+      "order_id": 254,
+      "risk_level": "elevated",
```
elazzabi: fyi, I've seen that the `risk_level` is marked as integer on the DB. I made some adjustments on my PR to address that, hopefully we can do the same here easily 🤞

client/transactions/uncaptured/test/__snapshots__/index.tsx.snap
```diff
@@ -243,77 +243,167 @@ exports[`Authorizations list renders correctly 1`] = `
...
+                  <a
+                    class="woocommerce-table__clickable-cell"
+                    data-link-type="wc-admin"
+                    href="admin.php?page=wc-admin&path=%2Fpayments%2Ftransactions%2Fdetails&id=pi_4242"
+                    tabindex="-1"
                   >
-                    High
-                  </span>
+                    <span>
+                      N/A
```
elazzabi: This is related to my comment about `risk_level` being an integer. Switching it to an integer will correctly fix this issue (showing a correct risk level -`Normal / High / Elevated` - instead of `N/A`)

tsconfig.json
```diff
@@ -17,10 +17,11 @@
       "iti/utils": [ "../node_modules/intl-tel-input/build/js/utils" ]
     },
     "types": [
-      "node", 
+      "node",
       "jest",
       "@testing-library/jest-dom",
-    ]
+    ],
+    "resolveJsonModule": true,
```
elazzabi, line 24: Is this used for the fixtures to work?

elazzabi (changes requested):
Looks good! I left a few comments for ameliorations. Let me know what do you think!

changelog/add-4569-capture-from-table
mgascam: My intention was to add a proper changelog in the integration PR. This one is only here to make CI happy. But I can change it if it makes things easier.

client/data/authorizations/actions.ts
mgascam: Indeed, fingers crossed XD

client/data/authorizations/resolvers.ts
mgascam: Yes, I took it from [here](https://www.typescriptlang.org/tsconfig#resolveJsonModule)

tsconfig.json
mgascam: Yes, I took it from [here](https://www.typescriptlang.org/tsconfig#resolveJsonModule)

client/data/authorizations/actions.ts
mgascam: Yes, I struggled with this one because the return type of apiFetch is `unknown`. So I ended using `any`. I went for a different approach in this commit [4c42d1e](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/4c42d1efc8b2b2300b8c3d0e83c0ed7cdeec6ced).

client/data/authorizations/actions.ts
mgascam: I think this is a good point, but I checked other places where we use the createErrorNotice, and we usually add a generic message, without using the caught error message. Maybe this is on purpose, to avoid exposing too much information. I don't know what to do here to be honest.

client/data/authorizations/resolvers.ts
mgascam: Indeed. I'll add a TODO comment here, saying that this is implemented in this other [PR](https://github.com/Automattic/woocommerce-payments/pull/4982).

mgascam: Done here [93e59f2](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/93e59f279f2017b38550cbb8995889774c13f04f)

client/data/authorizations/actions.ts
mgascam: This is a nice improvement thanks!
Done here [47cf00f](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/47cf00fb6fece9a6a2af45a222cc90bf5b228716)

client/data/authorizations/actions.ts
mgascam: Yes, added it here [47cf00f](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/47cf00fb6fece9a6a2af45a222cc90bf5b228716) thanks!

client/data/authorizations/resolvers.ts
mgascam: Yes, I added a TODO comment, this will be implemented in other PR

client/data/authorizations/test/authorizations.fixture.json
mgascam: Good catch! I fixed it here [2028ea7](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/2028ea7057dfd02686fe73ab2f3839e2914a80f0)

client/transactions/uncaptured/test/__snapshots__/index.tsx.snap
mgascam: Indeed, thanks! [2028ea7](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/2028ea7057dfd02686fe73ab2f3839e2914a80f0)

client/data/authorizations/actions.ts
elazzabi: I agree. I'm just thinking of catching the known errors already documented here #4957 

```
 -   `wcpay_missing_order` - Order not found
 -   `wcpay_refunded_order_uncapturable` -  Payment cannot be captured for partially or fully refunded orders
 -   `wcpay_payment_uncapturable` - The payment cannot be captured if intent status is not one of 'processing', 'requires_capture', or 'succeeded'
 -   `wcpay_intent_order_mismatch` - Payment cannot be captured because the order id does not match
 -   `wcpay_capture_error` - Unknown error
 ```

I think as a customer I would love to see `You can not capture this payment as it is partially/fully refunded` instead of `There has been an error capturing the payment. Please try again later.`. Because I'll keep trying later multiple times, but it will never work 😅 

I think having generic errors will just flood support with multiple requests that can be avoided

**Review by elazzabi — APPROVED** [review]
Thank you for working on my feedback! The [comment left here](https://github.com/Automattic/woocommerce-payments/pull/4961#discussion_r1006628495) is not a blocker for this PR, but I think worth working on in the future.

File: client/data/authorizations/actions.ts (hunk 4 above)
Comment (anu-rock) [reply client/data/authorizations/actions.ts:126]: +1 to @elazzabi's point. I had a similar concern in my error handling code for get (single) authorization:

https://github.com/Automattic/woocommerce-payments/pull/4688/files#diff-20dbcc4a43fe9db63759c8f939bed0803e2cbecdd715f5f6badc5d27189c09b1R56

The get auth endpoint is pretty simple and doesn't return a host of error codes, so there was not much need for the error object there. Here is a different story. I realize that would be additional work, but it'd definitely set a good example for others to follow in future.

File: client/data/authorizations/test/authorizations-summary.fixture.json [hunk 12]
```diff
@@ -0,0 +1,8 @@
+{
```
anu-rock, line 1: Does it need to be updated based on recent changes (currencies data) in the summary endpoint?

client/data/authorizations/actions.ts
mgascam: Thanks for the suggestion @anu-rock @elazzabi. I think this is a valid point and I will look into it. I might create a separate issue to tackle it after we solve the review bottleneck but before going live. Would that work for you if I finally create it?

client/data/authorizations/types.d.ts
```diff
@@ -0,0 +1,41 @@
+/** @format */
```
anu-rock: We also have `client/types/authorizations.d.ts` and I wonder if both files can be merged?

client/types/authorizations.d.ts
```diff
@@ -27,3 +28,19 @@ export interface AuthorizationsSummary {
...
+
+export interface AuthorizationsState {
+	summary: {
+		[ x: string ]: { data?: AuthorizationsSummary; error?: string };
+	};
+	byId: Record< string, Authorization >;
+}
+
+interface AuthorizationsList {
+	data: Authorization[];
```
anu-rock, line 45: If only we didn't have the `data` field, we could get rid of this type in favor of `Authorization[]`.

tsconfig.json
anu-rock: Nice addition 👍

anu-rock (commented):
The changes look good overall. I have left some minor comments, but things look solid already. Great job with the plumbing 🙌

client/data/authorizations/test/authorizations-summary.fixture.json
mgascam: Indeed good catch. Updated it in this commit [d24263c](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/d24263ca3db10dcf26eeea9d3f53b43848b07f29)

client/data/authorizations/types.d.ts
mgascam: Yes, I agree. I merged types from both files in this commit [be18446](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/be18446d9725f72d0426b8c7fbfff5ecf7192770) Thanks!

client/types/authorizations.d.ts
mgascam: Thanks @anu-rock that's a great point. I admit I introduced this type because I was struggling to type the `getAuthorizations` generator in `resolvers`. Since generator function is typed as `unknown`, TS was complaining when I accessed `result.data` I forgot that later I found a different approach that I liked better, but forgot to apply here. 

I believe we were missing a type for the response of the endpoint, which I added in this commit [40b7dc3](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/40b7dc33abd12aa53541be3aceccf936d6ee3ae9). Now TS is happy, and so am I.

client/data/authorizations/actions.ts
mgascam: I finally decided to create an issue to improve this https://github.com/Automattic/woocommerce-payments/issues/5002

client/data/authorizations/test/authorizations-summary.fixture.json
anu-rock: Thanks for updating. I thought the field was called `authorization_currencies` rather than `customer_currencies`. But may be I'm confusing it with something else?

client/data/authorizations/types.d.ts
anu-rock: Awesome, thank you ✋

client/types/authorizations.d.ts
anu-rock: That looks like a good middle ground to me :)

anu-rock (approved):
LGTM after the latest changes 🎉 

Left one last minor comment. Pre-approving the PR.

client/data/authorizations/test/authorizations-summary.fixture.json
mgascam: No, you are totally right, I checked the endpoint, but looked at an outdated version. I renamed it to 'all_currencies' following this [conversation](https://github.com/Automattic/woocommerce-payments-server/pull/2622#discussion_r1005690778). Commit here [f855ed8](https://github.com/Automattic/woocommerce-payments/pull/4961/commits/f855ed8c4273648235d4212d4f6765bc00c758e0)

client/data/authorizations/actions.ts
anu-rock: @mgascam Thank you! A separate issue is even better :)

client/data/authorizations/test/authorizations-summary.fixture.json
anu-rock: Yay ❤️

PR #4986, opened 2022-10-25
Add a class, which controls WCPay's mode

### LINKED ISSUE #4950: WCPay Platform: Create an API namespace
### Description

Create a new `WCPay\API` namespace, containing a static class for utility methods (`is_in_test_mode`, `is_in_dev_mode`, etc.) for access from consumers. Add setters for those methods instead of hooks, which modify them, in order to decrease the amount of calls to the same hook.

Over time this namespace should evolve to contain everything, needed by consumers, which on the other hand can be isolated into separate namespaces.

### Acceptance criteria

1. There is a new `WCPay\API` namespace in a separate directory within the client.
2. The new namespace contains a `Mode` class, which has:
    - `is_test() : bool` and `is_dev() : bool` methods.
    - `enter_dev_mode()`, `enter_test_mode()`, and `reset()` methods.
3. Those methods are fully tested.
4. Keep the gateway's `is_in_dev_mode` and `is_in_test_mode` backwards-compatible. To mediate this, we should also create WooPay/TumblrPay issues to remove existing calls to the old methods.
5. Create an issue to remove backwards compatibility when consumers stop requiring it.

### Additional context

RadoslavGeorgiev:
Fixes #4950
Replaces #4983

This PR introduces a new `WCPay\Core` namespace, containing a `Mode` class, which holds WCPay's working mode.

The class is instantiated in `WC_Payments` because of it's dependency on the gateway (for the test mode setting), and accessible through `WC_Payments::mode()`. This brings a couple of changes:

1. The logic for controlling the mode is no longer contained within the gateway class, as we're trying to make it as light as possible, and separate concerns.
2. Instead of checking environment and applying filters everytime `is_in_test_mode()` and `is_in_dev_mode()` are executed, now this is done during the initialization of the class, and the working mode is stored.

With the class implemented, all instances within the gateway have been altered to use the new functionality. During the development of the PR, the existing methods (gateway's `is_in_test_mode()` and `is_in_dev_mode()`) were removed, and only restored at the end in https://github.com/Automattic/woocommerce-payments/pull/4986/commits/4c5fd8374b0127c0f4a7388f42383739e7743b39.

The PR has been tested with WooPay as well. I'll open a PR for using the new methods there, and link it here.

- [x] The gateway methods should be restored for compatibility.
- [ ] `Update WCPay\Core\Mode::AVAILABLE_SINCE` with the right version before merging.
- [ ] Update docs

#### Testing instructions

Unit tests already work, so a smoke test should be sufficient.

-------------------

**Post merge**

This PR introduces no changes in critical flows, no testing instructions needed here.


includes/core/class-mode.php
```diff
@@ -0,0 +1,172 @@
...
+	 * Only recommended when testing, use filters otherwise.
+	 *
+	 * @param string $name The name of the mode to enter.
+	 * @param array  $args All args for the method (not used).
+	 * @return void
+	 */
+	public function __call( $name, $args ) {
+		// Only simulate the possible modes.
+		if ( ! in_array( $name, [ 'live', 'test', 'dev' ], true ) ) {
+			return;
```
zmaglica: I would throw `BadMethodCallException` here to catch issues with methods that we don't support here to prevent possible issues with calling random functions that do not exist

includes/core/class-mode.php
```diff
@@ -0,0 +1,172 @@
...
+	/**
+	 * Checks whether WCPay is in the given mode.
+	 *
+	 * @param  string $property The mode to check for. Either `live`, `test`, or `dev`.
+	 * @return bool If the propery is toggled.
+	 *
+	 * @throws Exception In case the class has not been initialized yet.
+	 */
+	public function __get( $property ) {
+		if ( ! isset( $this->dev_mode ) || ! isset( $this->test_mode ) ) {
```
zmaglica: If this statement is false, nothing will be returned? Is there a change for this statement to be false?

includes/core/class-mode.php
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Forces a switch of the current mode.
+	 * Only recommended when testing, use filters otherwise.
+	 *
+	 * @param string $name The name of the mode to enter.
+	 * @param array  $args All args for the method (not used).
+	 * @return void
+	 */
+	public function __call( $name, $args ) {
```
zmaglica: Is there a reason for creating __call magic method for only three functions that might never change in the future? Can we simply create` live()`, `test()`, `dev()` functions as its own standalone function and add logic there?

includes/core/class-mode.php
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Checks whether WCPay is in the given mode.
+	 *
+	 * @param  string $property The mode to check for. Either `live`, `test`, or `dev`.
+	 * @return bool If the propery is toggled.
+	 *
+	 * @throws Exception In case the class has not been initialized yet.
+	 */
+	public function __get( $property ) {
```
zmaglica: All properties are private and this is only used inside the function and for sake of simplicity, could we create functions like is_live, is_dev, is_SOMETHING and put the logic there.

I understand this approach for an abstract classed or classes that shares common logic to other classes, but making abstractions like this could be an overkill. Yet again, maybe I am wrong, but putting a couple of functions and not relaying on magic methods makes things easier to understand and read.

zmaglica (commented):
Left some comments. Another thing that came to my mind is that we could use `WC_Payments::mode()`  inside base controller class and expose private function or property and use that and replace all `WC_Payments::mode()->test` calls with a single property call from extended class.

includes/core/class-mode.php
RadoslavGeorgiev: Not applicable after the switch to normal methods in https://github.com/Automattic/woocommerce-payments/pull/4986/commits/896eaa3f9aac54d505f17ac81b0d71d48a9b9534

includes/core/class-mode.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4986/commits/896eaa3f9aac54d505f17ac81b0d71d48a9b9534

includes/core/class-mode.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/4986/commits/896eaa3f9aac54d505f17ac81b0d71d48a9b9534, I agree getters are not helping enough here.

includes/core/class-mode.php
RadoslavGeorgiev: The statement is now in a different place, and inverted, but yes: If the statement is was false, the class will attempt to initialize itself, or throw an exception if dependencies (one in this case, the gateway) are not ready yet.

zmaglica (approved):
LGTM

PR #4987, opened 2022-10-25
Refund Confirmation modal on order status change

### LINKED ISSUE #2814: Manual refund is created when changing the order status to "Refunded"
### Describe the bug

When users change the order status to "Refunded" it creates an odd issue where their customer doesn't get refunded from WooCommerce Payments and the user loses the option for a refund on WooCommerce. 

It even shows the following core message on the order notes creating even more confusion for the users:

> Order status set to refunded. To return funds to the customer you will need to issue a refund through your payment gateway.

### To Reproduce

1. Go to an order paid using WooCommerce Payments
2. Change the status of the order to "Refunded"
3. Try creating a refund using WooCommerce Payments (you shouldn't be able to do so)

### Actual behavior

A manual refund is created on WooCommerce, so the user can't refund using WooCommerce Payments.

### Screenshots

Annotation on 2021-08-31 at 04-24-23

### Expected behavior

Some options:
* Remove the manual refund from the status change
* Remove the manual refund, but add a screen/option asking the user if they would like to proceed with a refund on WooCommerce Payments
* Add automatic WooCommerce Payments refund when changing the status

### Additional context

p6q7sZ-9aw-p2

jessy-p:
Fixes #2814

Changing the status from the dropdown on the Edit Order screen and updating the order, would directly change the Order status to "Refunded" without triggering a WCPay refund. This causes confusion and if merchants accidentally used this workflow, they would have to contact support to get the charges properly refunded to the customer. 

These changes add a Refund Confirmation modal. From the Refund Confirmation Modal, the merchant can confirm refunding the full order amount via  WCPay, correctly refunding the money back to the customer before changing the status to "Refunded". 

Note: "Refund" button on the "Order Edit" screen can still be used for a full or partial refund of the order, like before.

**Screenshots**
image

#### Testing instructions

**Successful Refund**
- Place an order and navigate to the 'Edit Order' screen. 
- Change the status to 'Refunded' in the Status: dropdown. 
- The Refund Confirmation Modal is displayed. 
- 'Cancel' will dismiss the modal and return the order to the previous state.
- 'Refund' refunds the order successfully on Stripe and WCPay. 

**Error cases**
- If order is already refunded, do not show the Refund Confirmation modal.
- If the order has no `_charge_id` metadata, it cannot be refunded, do not show the Refund Confirmation modal.
- If refund failed on Stripe/server, should display an error and change the status dropdown to the previous state (so that refunded state change does not get changed).

**Other tests**
- Partial refunds
- Non-US and Zero-decimal currencies

-------------------

**Post merge**
- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.0.0#refund-confirmation-modal
- [x] Not Applicable - Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.


client/order/refund-confirm-modal/index.js
```diff
@@ -0,0 +1,122 @@
...
+									__( 'Refund %s', 'woocommerce-payments' ),
+									formattedRefundAmount
+								) }
+							</Button>
+						</>
+					}
+				>
+					<p>
+						{ __(
+							"Issue a full refund back to your customer's credit card using WooCommerce Payments." +
```
jbordonado, line 111: You're missing a space after `WooCommerce Payments.` and before `This action can not...`

client/order/refund-confirm-modal/index.js
```diff
@@ -0,0 +1,122 @@
...
+								'This action can not be undone. To issue a partial refund, click "Cancel", and use ' +
```
jbordonado, line 111: You're missing a space after `WooCommerce Payments.` and before `This action can not...`

client/order/refund-confirm-modal/index.js
```diff
@@ -0,0 +1,122 @@
...
+		jQuery( '#order_status' ).val( orderStatus ).change();
+	};
+
+	const handleRefundCancel = () => {
+		resetOrderStatus();
+		closeModal();
+	};
+
+	const handleRefundConfirm = () => {
+		blockUI();
```
jbordonado: That one seems redundant given the `blockUi()` l.67?

includes/admin/class-wc-payments-admin.php
```diff
@@ -556,10 +556,16 @@ public function register_payments_scripts() {
 		wp_register_script(
 			'WCPAY_ADMIN_ORDER_ACTIONS',
 			plugins_url( 'dist/order.js', WCPAY_PLUGIN_FILE ),
-			[ 'jquery-tiptip' ],
```
jbordonado, line 558: Can you explain why we need the dependencies? I tried to remove it and refund using the new modal, and I didn't stumble upon any issues, but maybe I missed something?

jbordonado (commented):
I tested the successful and error scenarios, and they worked well.
I didn't test partial refunds and Zero-decimal currencies yet.

Still, I'm leaving a few comments here and there now because I won't be able to have another look before Wednesday. If someone else would like to review it before that, please do so :)

includes/admin/class-wc-payments-admin.php
jessy-p: Without the dependencies, you should get an undefined error on the React render. As this is just a normal WC Admin page, not one of our WCPayments pages, we need to include the dependencies on the page. Can you check if your page was rebuilt and refreshed properly?

jbordonado: I can confirm I don't see any undefined error without the dependencies. Can you check again on your side?
I started the client container and built the app without the dependencies, and used an incognito page to try the scenario.

jbordonado:
While testing the rest, I noticed the styles are wrong on the new modal:

image

Not sure why I didn't see this last week (and I see all you did in the meantime is pulling develop twice), but it looks like the `margin-top` rule for the content is what is messing with the design.
If I remove it, it looks OK.

Can you double-check?

jbordonado:
Besides my last comment, I can confirm partial refunds and zero-decimal currencies (e.g. JPY) scenarios still work fine

client/order/refund-confirm-modal/index.js
jessy-p: Thanks for spotting that! Fixed.

client/order/refund-confirm-modal/index.js
jessy-p: Fixed

jessy-p:
>Not sure why I didn't see this last week (and I see all you did in the meantime is pulling develop twice)

I think some CSS changes in develop broke the modal this week, fixed it now.

includes/admin/class-wc-payments-admin.php
jessy-p: When I remove that, and then click on Refunded, I get the below error in console and the modal will not get rendered 
image

jbordonado: Maybe an issue with the way I start my local setup. Nevermind then!

jbordonado (approved):
Made one last simple test after the fixes and it still works as expected (including the modal styles).

LGTM 🚢

aheckler:
@jessy-p After this change, is it still possible to set the order status to Refunded without any other action? I can imagine a couple cases where that might be necessary. Example:

1. Merchant takes an order.
2. Merchant's site goes offline for some reason.
3. Merchant gets a customer request for a refund.
4. Since they cannot refund via the site, they have support refund via Stripe.
5. They bring the site back online and want to change the order status to match up with the fact that we refunded the customer payment.

@csmcneill probably has some more scenarios like this which may occur from time to time.

jessy-p:
After this change, it is not possible to change the status to refunded in the scenario mentioned (offline refund)
Depending on customer reports, we may need to go back to design to provide for that scenario.

csmcneill:
> @csmcneill probably has some more scenarios like this which may occur from time to time.

This is the only instance I can think of where this functionality may be necessary.

One thing I'm thinking would be a way to offer this manual refund option but in a way that would prevent a user from performing that action accidentally while also having some CYA components, such as (pardon my crude Droplr editing skillz):

1. Add a third option that is left-aligned that reads `Advanced Options` (or similar):
Markup on 2022-11-15 at 14:28:27
2. Selecting this would expand the modal and allow a customer to mark the order as refunded _without_ processing a refund.
3. Marking an order as refunded in this manner would require an additional confirmation, like the confirmation that occurs when going through the standard refund process:
Markup on 2022-11-15 at 14:32:41

PR #5011, opened 2022-10-28
Fix currency switcher block padding

### LINKED ISSUE #4945: Bad currency switcher block style
### Describe the bug
Just noticed that the currency switcher block has render issues while seeing it on **WP Admin → Appearance**. Looks like after WC 6.6.0 `.wp-core-ui select` overrides the old style and cause that issue. It seems to be introduced in https://github.com/woocommerce/woocommerce/pull/32729

It's a WP Admin only thing, not happening on the store. But we should be careful with the fix to not alter the front style.

### To Reproduce
1. Go to **WP Admin → Appearance**.
2. Add the `Currency Switcher Block`.
4. See how it renders wrong.

### Actual behavior
The current fixed `padding` applied on the `<select>` conflicts with the style applied by `.wp-core-ui select`.

### Screenshots
Screenshot 2022-10-19 at 12 16 03

### Expected behavior
It should like similar to the store one.

Screenshot 2022-10-19 at 12 15 26

ismaeldcom:
Fixes #4945 

With WC 6.6.0 `.wp-core-ui select` style gets applied while a block is edited. Causing a UI issue with our fixed `padding` on the currency switcher. Removing the `padding` from the block JS definition is enough to fix it. Ensuring it's kept when it renders on the frontend.

#### Screenshot
| Block type | Rendered on the frontend |
|-|-|
|Screenshot 2022-10-28 at 11 43 12|Screenshot 2022-10-28 at 11 43 33|

#### Testing instructions
- Add the **Currency Switcher Block** to a page or as a widget.
- While editing the block properties it should render correctly, as int the screenshot.
- It should also look ok on the store frontend.

-------------------

**Post merge**


dmallory42 (approved):
LGTM! Tested and it renders correctly in both the widget section and on the frontend.

Issue #4954, opened 2022-10-20
WCPay Platform: API request classes


RadoslavGeorgiev:
### Description

Create a `WCPay\API\Request` class, which exposes the right hooks at the right time. An example would be to allow the addition of extra data, while preventing extensions with tampering with the rest of the request.

Currently consumers alter the arguments for API requests in various places. While extracting their code elsewhere, we’d need to allow them to still alter the necessary arguments.

The preferred approach here is to only allow values to be added, while specifying which existing values can be modified, and add validation to the modified values.

We should create a set of Data Transfer Objects under a new namespace (ex. `WCPay\API\Request`), where we would carefully control which setters are private/final, and which are protected. The API client methods will expect a specific class of request as an argument, but we would allow the class to be extended by consumers, and the child class to be used/filtered.

Those new classes should also live within the `WCPay\API` namespace, as they can be extended by consumers.

### Acceptance criteria

1. There is a base request class, which can be instantiated by consumers with and without specifying the HTTP client class.
2. The base request class is be usable within the API client class for making requests.
3. The base class is extendable for specific requests, which would allow marking properties as private/final.
4. The mechanism should prevent extended classes from overwriting parent properties.
5. We should allow hooks to be used in such a way, that consumers can extend those classes, and replace the used one, while still having parent properties protected.

### Additional context

See the epic for links to the project thread and PRD.

PR #5012, opened 2022-10-28
WCPay base API Request class

### LINKED ISSUE #5056: WCPay Platform: Base API request class
### Description
Create a base API request class that will be used as a foundation for other API requests. The main purpose is to create classes and features that will be extended or used when API requests are executed to WcPay server. All requests need to be created like a builder where you could easily build a request using methods and these methods will validate data that is passed.

For example, the API request class should look something like this:

```
$create_charge = new Create_Charge();
		$create_charge
			->set_amount( 100 )
			->set_source_id( 'source_id' );
$result = $this->payments_api_client->send_wcpay_request( $create_charge );
```

- If you pass an incorrect amount, the exception will be thrown and an error message will tell what's wrong with it.
- If you pass empty source id, the exception will be thrown and an error message will tell what's wrong with it.

Since this is the base class. Make sure that you create base classes that will share common features and methods. One example of this is getter and setters methods. Since the validation probably will be performed inside setter methods, make sure that you use magic methods to share a common logic and prevent duplication of methods if needed.

The good example of setters and getter methods could be achieved with something like this:
```

	public function __call( $name, $arguments ) {

		/**
		 * If get or set method is not defined, this call function will simply add them and use basic get and set functionality.
		 */

		$method_data = explode( '_', $name );
		$prefix      = $method_data[0] ?? null;
		$property    = $method_data[1] ?? null;
		if ( $property && property_exists( $this, $property ) ) {
			if ( 'set' === $prefix && 1 === count( $arguments ) ) {
				$this->{$property} = $arguments[0];
				return $this;
			}
			if ( 'get' === $prefix ) {
				return $this->{$property};
			}
		}
	}
```
If you look at the example above, setter and getters methods are automatically added and all you need to do is to add dockblock for methods.

Another thing that we could add is to create Response DTO that will hold common functionalities and where response will be treated as object. The same response DTO could be used to create DTO responses if we decide to take this approach. Instead of relaying on arrays, the DTO Response object could be used and common methods like getting body, checking is there any errors and what not could be added to this DTO. Make sure that you create base DTO class that will hold logic for all others DTOs to prevent code duplication.

### Acceptance criteria

* Create base request class
* Create base response class
* Create one API request using above, classed as an example
* Create future issues or PRs to add all requests we agree to implement.

### Testing instructions

1. Cover all classes with tests
2. Make sure that requests are working and prevent modifying request class that could damage something. One example is to prevent modifying of certain properties in classes.
3. Make sure that everything is inside the same namespace (the core namespace)
4. Make sure that all responses are under same DTO or interface for easier replacement/refactor if needed.

### Dev notes

- Use DTOs
- Use ValueObjects
- If data is mutable, make sure that data is validated securely

zmaglica:
Fixes #5056.

The purpose of this PR is to add the foundation class (and all methods) where you could simply execute WCPay API endpoint   using the class like a builder where you could chain function calls and where you could build fully working API requests to WCPay API server. The approach that will be used is combination of DTOs and Value Objects where data could be setter in a way that will provide argument validation. 

This PR will add all necessary classed and functionality where you could create request and all necessary checks will be performed. The main purpose is to create a functionality that will look something like this:

```
$create_charge = new Create_Charge();
		$create_charge
			->set_amount( 100 )
			->set_source_id( 'source_id' );
$result = $this->payments_api_client->send_wcpay_request( $create_charge );
```
		
With this approach, the type-safety will be ensured and building API requests will be way easier. 

Instead of relaying on arrays when response is returned from server, a new DTO have been added that extends request and extend it by functions that could be used somewhere else. The main purpose of this DTO is that we could extend request class and include DTOs for API responses, where we will ensure safety of the parameters that we got from the API response.

This PR is only one PR where others will be created once we conclude on how request class should look like. To have a better understanding, take a look at Create_Charge class where basic request have been created.

Another thing that is added are interfaces. The main purpose of them is to create a set of rules that we could use if we decide to change the request (or response) class sometime in the future. There are abstract or base classed that defines core functionality that will be used from all requests, DTOs, and value objects. For every request class, there is a generic `__set` method that is inherited from the base class to every class that extends it. the `__set` function will try to find an internal `set_SOMETHING` function that will hold validation data (if any) and if a method doesn't exist, the normal way of setting a variable will be used.

This PR has been edited, and some code have been removed. The reason behind this is that there is too much code that solves too many problems, and separating or creating multiple PRs will be easier to track and work on.

#### Testing instructions

1. Create new create charge API request using added class ( `new Create_Charge()` )
2. Make sure that exception will be thrown when charge amount is not number
3. Make sure that exception will be thrown when amount is below zero
4. Make sure that exception will be thrown if source id is empty
5. Fill class with correct data
6. Execute WCPay request by calling `$this->payments_api_client->send_wcpay_request()` and pass created class from step 1
7. Make sure that return of this function is Response DTO that was added in this PR

-------------------

**Post merge**


includes/core/DataTransferObjects/class-data_transfer_object.php
```diff
@@ -0,0 +1,47 @@
...
+	/**
+	 * Return array of all properties.
+	 *
+	 * @return array
+	 */
+	public function to_array() {
+		$class = new ReflectionClass($this);
+		$data = [];
+		foreach ($class->getProperties() as $reflectionProperty) {
+			$data[$reflectionProperty->getName()] = $reflectionProperty->getValue();
```
KarlisJ: I suggest being more explicit about what kind of properties we are working on here. Which is private non-static in this case.

includes/core/DataTransferObjects/class-data_transfer_object.php
```diff
@@ -0,0 +1,47 @@
...
+		return $data;
+	}
+
+	/**
+	 * Change DTO property value if property is not private.
+	 * @param ...$values
+	 * @return self Cloned object with changed properties.
+	 * @throws ReflectionException
+	 */
+	public function with(...$values)
```
KarlisJ: What would be the use case for this method?

includes/core/Enums/class-wc-pay-endpoints.php
```diff
@@ -0,0 +1,44 @@
+<?php
+namespace WCPay\Core\Enums;
+
+final class Wc_Pay_Endpoints
```
RadoslavGeorgiev: Let's just use `Endpoints` as th class name, and importantly, try to either stick to `WCPay` (for the core namespace) or `WC_Payments` (the existing one for standard includes), but not add `Wc_Pay` or `WC_Pay`.

includes/core/DataTransferObjects/class-data_transfer_object.php
RadoslavGeorgiev: It might be a good idea to focus on properties with specific access (public or private).

includes/core/DataTransferObjects/class-data_transfer_object.php
```diff
@@ -0,0 +1,47 @@
...
+use ReflectionException;
+
+abstract class Data_Transfer_Object
+{
+	/**
+	 * Return array of all properties.
+	 *
+	 * @return array
+	 */
+	public function to_array() {
```
RadoslavGeorgiev: I'll probably come back to this, but why do we need a `to_array()` method? Can't we implement `ArrayAccess` here?

includes/core/DataTransferObjects/class-data_transfer_object.php
```diff
@@ -0,0 +1,47 @@
...
+		$data = [];
+		foreach ($class->getProperties() as $reflectionProperty) {
+			$data[$reflectionProperty->getName()] = $reflectionProperty->getValue();
+		}
+
+		return $data;
+	}
+
+	/**
+	 * Change DTO property value if property is not private.
```
RadoslavGeorgiev: This description could make a lot more sense, it seems like this changes a single property.

includes/core/DataTransferObjects/class-data_transfer_object.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @throws ReflectionException
+	 */
+	public function with(...$values)
+	{
+		$class = new ReflectionClass($this);
+		$clone = $class->newInstanceWithoutConstructor();
+		foreach ($class->getProperties() as $reflectionProperty) {
+			$propertyValue = $reflectionProperty->getValue();
+			$propertyName = $reflectionProperty->getName();
+			if ( !$reflectionProperty->isPrivate() && array_key_exists($propertyName, $values)) {
```
RadoslavGeorgiev: `array_key_exists` would not really work here. `$values` has numeric indexes, we wouldn't know which key to use.

includes/core/Contracts/API/Response/BaseResponse.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace WCPay\Core\Contracts\API\Response;
+
+interface BaseResponse {
```
RadoslavGeorgiev: Right now this seems like it is describing a request, let's see if it get clearer....

includes/core/DataTransferObjects/API/class-address.php
```diff
@@ -0,0 +1,142 @@
...
+		return $data;
+	}
+
+	/**
+	 * Get formatted address.
+	 *
+	 * @return string
+	 */
+	public function get_formatted_address() {
+		return WC()->countries->get_formatted_address( $this->get_address_as_array() );
```
RadoslavGeorgiev: Aren't DTOs supposed to be business-logic-free?

includes/core/DataTransferObjects/API/Request/class-create-charge.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace WCPay\Core\DataTransferObjects\API\Request;
+
+use WCPay\Core\DataTransferObjects\Data_Transfer_Object;
+
+final class Create_Charge extends Data_Transfer_Object
```
RadoslavGeorgiev: Isn't this supposed to be a request?

includes/core/DataTransferObjects/API/Request/class-request.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+namespace WCPay\Core\DataTransferObjects\API\Request;
+
+use WCPay\Core\Contracts\API\Request\BaseRequest;
+use WCPay\Core\DataTransferObjects\Data_Transfer_Object;
+
+final class Request extends Data_Transfer_Object implements BaseRequest
```
RadoslavGeorgiev: Is this supposed to be a generic request? How is it used?

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,191 @@
+<?php
+
+namespace WCPay\Core\DataTransferObjects;
+
+use WP_Http_Cookie;
+
+final class Response extends Data_Transfer_Object
```
RadoslavGeorgiev: How will this be used?

includes/core/API/class-wc_pay_request.php
```diff
@@ -0,0 +1,263 @@
...
+use WCPay\Core\DataTransferObjects\Response;
+use WCPay\Core\Enums\Http_Methods;
+use WCPay\Core\Enums\Wc_Pay_Endpoints;
+use WCPay\Core\Factories\Create_Charge_Dto;
+use WCPay\Exceptions\Amount_Too_Small_Exception;
+use WCPay\Exceptions\API_Exception;
+use WCPay\Fraud_Prevention\Fraud_Prevention_Service;
+use WCPay\Logger;
+
+class Wc_Pay_Request
```
RadoslavGeorgiev: How is this used?

naman03malhotra:
@zmaglica @RadoslavGeorgiev this PR is linked to which issue? Also, @zmaglica can you please update the PR description?

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: This is base method. If something needs to be removed, the parent class could override a method and create a custom to_array function.

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: For simplicity, I used `to_array()`. We could also use `ArrayAccess`. Please let me know what you think that it would be best. If you ask me, I would leave `to_array` since it can be easily edited based on the needs.

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: Fixed.

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: The main question is: What properties we are allowing to change? The properties should be set to private by default. We could prevent changing some key properties if needed based on the business logic. LMK what properties should be overwritten with this method, and we can come with the conclusion.

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: Something like this:

```
class Dto {
    private $id;
    private $amount;
	__construct($id, $amount) {
		$this->id = $id;
		$this->amount = $amount;
	}
}

$dto = new Dto(1, 100);
$dto->with(['id' => 2]);

echo $dto->id // will echo 2;
```

With this method, you could simply change any value in DTO.

includes/core/DataTransferObjects/API/class-address.php
zmaglica: This DTO have been removed, but you can use methods in DTO if they are not updating an original state (I could be wrong on this one).

includes/core/DataTransferObjects/API/Request/class-create-charge.php
zmaglica: It is updated now. It is request now.

includes/core/API/class-wc_pay_request.php
zmaglica: I removed this code for a sake of simplicity and used original wc pay http api class. We could add our own implementation if needed.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: Response DTO will be a base class for all responses or response object that will be parsed from response we got from server. Instead of relaying on arrays, you can wrap the response in this DTO or even create a brand-new response from this DTO and inherit all core functions.

includes/core/class-loader.php
```diff
@@ -0,0 +1,27 @@
...
+/** Load enums. */
+
+require_once __DIR__ . '/Enums/class-http-methods.php';
+require_once __DIR__ . '/Enums/class-wc-pay-endpoints.php';
+
+/** Load value objects */
+
+require_once __DIR__ . '/ValueObjects/class-value-object.php';
+require_once __DIR__ . '/ValueObjects/API/Request/class-base-request.php';
+require_once __DIR__ . '/ValueObjects/API/Request/class-create-charge.php';
```
RadoslavGeorgiev: Oh no, the autoloader is gone! 😭 

Anyway, let's include those files in a meaningful place for now. We should discuss how (difficult it would be) to start using an autoloader. Maybe doing it for the `core` directory isn't a bad place to start with PSR-4, I just want to focus on the issue at hand first.

includes/core/Enums/class-http-methods.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\Enums;
+
+/**
+ * HTTP methods enum class.
+ */
+final class Http_Methods {
```
RadoslavGeorgiev: What do you think about using WordPress's buit-in constants? https://github.com/WordPress/WordPress/blob/master/wp-includes/class-requests.php#L22-L77

includes/core/Enums/class-wc-pay-endpoints.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Endpoints enum class.
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\Enums;
```
RadoslavGeorgiev: I'm wondering whether we could have a constants directory instead for now, in order to make it clearer when we switch to actual enums someday in the future when our minimum supported version is 8.1?

Note that I haven't used enums yet, and I'm not sure if that's needed. If we switch from `class` to `enum` someday, will we be able to keep on using something like `Wc_Pay_Endpoints::APPLE_PAY_API` the same way?

includes/core/Enums/class-wc-pay-endpoints.php
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Endpoints enum class.
+ */
+final class Wc_Pay_Endpoints {
```
RadoslavGeorgiev: Nitpick: We're in the `WCPay` namespace already, let's make this just `Endpoints`.

includes/core/ValueObjects/class-value-object.php
```diff
@@ -0,0 +1,86 @@
...
+	 * @param array  $arguments Method arguments.
+	 * @return self|mixed
+	 */
+	public function __call( $name, $arguments ) {
+
+		/**
+		 * If get or set method is not defined, this call function will simply add them and use basic get and set functionality.
+		 */
+
+		$method_data = explode( '_', $name );
```
RadoslavGeorgiev: `explode` is not sufficient here:

```php
$name = 'set_origin_country';

$method_data = explode( '_', $name );
$prefix      = $method_data[0] ?? null;
$property    = $method_data[1] ?? null;
var_dump( compact( 'prefix', 'property' ) );

// array(2) {
//   ["prefix"]=>
//   string(3) "set"
//   ["property"]=>
//   string(6) "origin"
// }

if ( preg_match( '~^(set|get)_(.+)$~', $name, $matches ) ) {
	[ , $prefix, $property ] = $matches;
	var_dump( compact( 'prefix', 'property' ) );
}

// array(2) {
//   ["prefix"]=>
//   string(3) "set"
//   ["property"]=>
//   string(14) "origin_country"
// }
```

includes/core/Contracts/API/Request/interface-base-request.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\Contracts\API\Request;
+
+/**
+ * Base request interface.
+ */
+interface Base_Request {
```
RadoslavGeorgiev: Can you think of scenario where someone would implement this contract instead of using the base class (`WCPay\Core\ValueObjects\API\Request\Base_Request`)?

includes/core/Contracts/API/Response/interface-base-response.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace WCPay\Core\Contracts\API\Response;
+
+use WCPay\Core\DataTransferObjects\Response;
+use WP_Http_Cookie;
+
+/**
+ * Base response interface.
+ */
+interface Base_Response {
```
RadoslavGeorgiev: Similar question to the base request contract: How can this practically help?

Those methods (`get_body`, `get_code`, `get_message`, etc.) seem to explicitly define the WCPay server responses. If it was more generic (ex. `get_data()`, `has_error()`, `get_error()`, etc.), I would understand the usefulness of such a contract, but in this case I do not see who would implement it, and how they would implement it.

includes/core/Enums/class-wc-pay-endpoints.php
zmaglica: Noted.

includes/core/ValueObjects/class-value-object.php
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	/**
+	 * Custom value object setter.
+	 *
+	 * @param string $name Name of variable.
+	 * @param mixed  $value Value of variable.
+	 * @return $this
+	 */
+	public function __set( $name, $value ) {
```
RadoslavGeorgiev: Let's pick an approach:

1. `__get()` and `__set()` for `echo $object->amount` and `$object->amount = 200`
2. Just `__call` for  `echo $object->get_amount()` and `$object->set_amount( 200 )`
3. `__get()` and `__call()` for `echo $object->amount` and `$object->amount( 200 )`

Right now we there are two ways to write values, through `__set` and `__call( 'set_*' )`.

includes/core/ValueObjects/class-value-object.php
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Fill array from data.
+	 *
+	 * @param array $data Data to fill.
+	 * @return void
+	 */
+	public function fill_from_array( $data ) {
+		foreach ( $data as $key => $value ) {
+			if ( property_exists( $this, $key ) ) {
+				$this->$key = $value;
```
RadoslavGeorgiev: This should use the same logic as `__set()`.

includes/core/ValueObjects/class-value-object.php
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	/**
+	 * Return array of all properties.
+	 *
+	 * @return array
+	 */
+	public function to_array() {
+		$data = [];
+		foreach ( get_object_vars( $this ) as $key => $value ) {
```
RadoslavGeorgiev: `get_object_vars` returns an array already.

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+		$this->response_data = json_decode( $body, true );
+	}
+
+	/**
+	 * Returns true if response is JSON.
+	 *
+	 * @return bool
+	 */
+	public function is_json_response() {
+		return 'application/json' === substr( $this->headers['content-type'] ?? '', 0, strlen( 'application/json' ) );
```
RadoslavGeorgiev: Isn't this the same as:

```suggestion
		return 0 === strpos( $this->headers['content-type'] ?? '', 'application/json' );
```

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+
+namespace WCPay\Core\DataTransferObjects;
+
+use WCPay\Core\Contracts\API\Response\Base_Response;
+use WP_Http_Cookie;
+
+/**
+ * DTO for parsing response from WCPay server API.
+ */
+final class Response extends Data_Transfer_Object implements Base_Response {
```
KarlisJ:
```php
final class ... implements Base_Response
```
What is the gain of having interface for final class if it's not used anywhere else? Do we plan to support other types of responses (like we have multiple Requests in this PR)?

KarlisJ:
```php
extends Data_Transfer_Object
```

I feel that this class is doing a bit too much to be categorized as DTO. To me, it is a strict object representing a response from the wcpay server. I think this is a core class that will be involved in many important business logic representations. Calling it `DTO` is devaluating it and we might have problems in future where the "DTO" label will influence our decision of not adding specific methods to it when needed and looking for other solutions. 

The focus of this PR is making the interaction with the server more structured (`Request` and `Response` classes). M worry is that by introducing the term "DTO" we'll bypass the necessary discussions of how we want the DTO to be implemented in our codebase. IMHO, what's currently in `Data_Transfer_Object` is not fully versatile for mass adoption so I would be worried that we'll fill the terminology slot and make things harder for us later.

With that in mind, I would suggest moving the methods from `Data_Transfer_Object` to the `Response`.

KarlisJ:
```php
... Response ... Base_Response
```

Is this something we plan to use for other external systems? If this is to be solely catered to the communication with the server, then shouldn't that we reflected in the class names?

includes/core/Contracts/API/Response/interface-base-response.php
```diff
@@ -0,0 +1,75 @@
...
+	 */
+	public function has_error_in_response_data();
+
+	/**
+	 * Create DTO from WCPay response.
+	 *
+	 * @param array $response Server response.
+	 * @return Response
+	 */
+	public static function create_from_wc_pay_response( $response);
```
KarlisJ: Already been mentioned in other comments, but I would like to see more explicit naming that contains "server". `wc_pay_response` is quite unclear within the context. It's WCpay plugin, so reading this, it feels as if we're doing something with the response we send to the frontend.

includes/core/DataTransferObjects/class-data-transfer-object.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace WCPay\Core\DataTransferObjects;
+
+use ReflectionClass;
+use ReflectionException;
+
+/**
+ * Abstract DTO class.
+ */
+abstract class Data_Transfer_Object {
```
KarlisJ: As I mentioned in another comment, I'd suggest focusing on solidifying how we interact with the server and leaving the introduction of DTO to another issue. 

We can just move these methods to the `Response` class which so far in this PR is the only class that extends `Data_Transfer_Object` and focus on the API for server interactions.

If needs be, we can always refactor these methods out into DTO base class later.

includes/core/ValueObjects/API/Request/class-base-request.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace WCPay\Core\ValueObjects\API\Request;
+
+use WCPay\Core\Contracts\API\Request\Base_Request as Base_Request_Contract;
+use WCPay\Core\DataTransferObjects\Value_Object;
+
+/**
+ * Base request value object.
+ */
+class Base_Request extends Value_Object implements Base_Request_Contract {
```
KarlisJ: Similar to other comments, I'd suggest `Base_Server_Request`

includes/core/ValueObjects/class-value-object.php
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\DataTransferObjects;
+
+/**
+ * Base Value Object class.
+ */
+abstract class Value_Object {
```
KarlisJ: What's the difference between DTO and Value object? 
1. I genuinely want to raise discussion about the usage of these terms in this PR
2. Doc block doesn't give much away to help me

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -2079,6 +2081,29 @@ public function test_authorizations_summary_success() {
...
+				'currency'    => 'usd',
+			]
+		);
+
+		$create_charge = new Create_Charge();
+		$create_charge
+			->set_amount( $expected_amount )
+			->set_source_id( 'source_id' );
+		$result = $this->payments_api_client->send_wcpay_request( $create_charge );
+		$this->assertInstanceOf( Response::class, $result );
```
KarlisJ, line 2104: Kudos for providing this usage example! I love the direction the API is going!

I'd suggest taking bit more control over creation
```php
$create_charge = Create_Charge::new/make()
    ->chained_method()
    // ...
```

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1986,6 +1988,17 @@ public function is_in_test_mode() {
 		return WC_Payments::get_gateway()->is_in_test_mode();
 	}
 
+	/**
+	 * Send WC pay request using value objects.
+	 *
+	 * @param WC_Pay_Request $request - Request object.
+	 * @return Response
+	 */
+	public function send_wcpay_request( WC_Pay_Request $request ) {
```
KarlisJ: In one comment I asked if we plan to use Base_Request for other API than the server. Here it's quite evident that we care only about the server so `send_wcpay_server_request` would be more descriptive name.

includes/core/ValueObjects/API/Request/class-create-charge.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Get source id
+	 *
+	 * @return string
+	 */
+	public function get_source_id() {
+		return $this->source_id;
+	}
```
KarlisJ: Do we ever need getters for a Request class? 

We'll set the amount and source id and then send the request. In what case would we want to get something from the Request?

includes/core/ValueObjects/API/Request/class-create-charge.php
```diff
@@ -0,0 +1,111 @@
...
+	 * @return self
+	 * @throws \InvalidArgumentException
+	 */
+	public function set_amount( $amount ) {
+		if ( ! is_numeric( $amount ) ) {
+			throw new \InvalidArgumentException( 'Amount is not numeric value' );
+		}
+		$this->amount = $amount;
+		return $this;
+	}
```
KarlisJ: I like the general approach that implementation classes can use setters to validate the fields.

`Base_Request` could provide a validator or validation methods to make things easier and more consistent. Something like 
```php
	public function set_amount( $amount ) {
		$this->validator->number($amount);
		$this->amount = $amount;
		return $this;
	}
```

KarlisJ (commented):
Overall, I like the direction of this PR! Huge improvement in how we interact with server.

I've left some comments and questions. My main message I'd like to send with this review is to focus on improving the interaction with server. Currently, this PR tries to create reusable base classes for multi-purpose usage. Such focus could negatively impact the API and implementation of main focus point - interaction with server API.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: Yeah, I will remove final since this class could be potentially extended.

zmaglica: Yes, the focus is on interaction between client and server, but I wanted to make a foundation for other parts of the plugin where you could potentially use DTO and create a class that could share common logic for all DTOs somewhere in the future.

But as you mentioned, the purpose of this PR is server interaction and I will rename all the DTOs, ValueObjects, and other things to represent client server interaction. DTOs will be replaced with Requests/Responses

zmaglica: Maybe we could rename a class or put it inside an additional folder to display that we are actually referring to wcpay server. Our whole communication is done from the server and as it seems, the only communication will be from the server, so that's why I put naming like this. But putting classes into separate folder or change naming could also improve readability and extensibility if we decide to have more 'server' providers.

includes/core/Contracts/API/Response/interface-base-response.php
zmaglica: I will update the function name to reflect from where the response is created.

includes/core/DataTransferObjects/class-data-transfer-object.php
zmaglica: Make sense. I will remove all DTO abstractions into parent Request/Response class. If needed, we could easily extract logic to separate class if we need it in the future.

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
zmaglica: Thank you for the suggestion. I will add it.

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: As mentioned in the previous comment, I will either change naming or put it into a separate folder. I am more in favor in renaming a class, but I will also consider a folder option too.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: It cost me like 10 seconds with my IDE to generate getters, and if you need them, you don't have to rely on magic methods and doc block to have a type hinting. They won't do any harm, and sometimes you might need them. The main reason is that I spend like seconds to add them, and it is a bit more clear when you see inside class that you have getters.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: I wanted to do it like this, but was afraid that it will be too broad for this use case. One approach would be to use validator class somewhere from composer package or built our own validation methods. For now, I will add those into parent Request class, and maybe we could expand it with more validators.

includes/core/ValueObjects/API/Request/class-create-charge.php
```diff
@@ -0,0 +1,111 @@
...
+		}
+		throw new \InvalidArgumentException( 'Source ID is invalid! Source needs to be string type and not empty' );
+	}
+
+	/**
+	 * Get parameters.
+	 *
+	 * @return array
+	 */
+	public function get_parameters() {
```
RadoslavGeorgiev: This feels like a good place to enforce validation. Right now both parameters could not be set, and this method would still not throw any errors or exceptions.

includes/core/ValueObjects/API/Request/class-create-charge.php
```diff
@@ -0,0 +1,111 @@
...
+
+namespace WCPay\Core\ValueObjects\API\Request;
+
+use WCPay\Core\Enums\Http_Methods;
+use WCPay\Core\Enums\Wc_Pay_Endpoints;
+
+/**
+ * Create charge request.
+ */
+final class Create_Charge extends Base_Request {
```
RadoslavGeorgiev: Do we want to make it final, or allow it to be extended?

includes/core/ValueObjects/API/Request/class-base-request.php
```diff
@@ -0,0 +1,97 @@
...
+		return $this->to_array();
+	}
+
+	/**
+	 * Get request method.
+	 *
+	 * @return string
+	 */
+	public function get_method() {
+		return '';
```
RadoslavGeorgiev: Shouldn't we make this an abstract method instead of not returning anything?

includes/core/ValueObjects/API/Request/class-base-request.php
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	/**
+	 * Get request route.
+	 *
+	 * @return string
+	 */
+	public function get_route() {
+		return '';
```
RadoslavGeorgiev: Same as `get_method()`.

includes/core/ValueObjects/API/Request/class-base-request.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+class Base_Request extends Value_Object implements Base_Request_Contract {
+
+
+	/**
+	 * Request headers.
+	 *
+	 * @var array $headers
+	 */
+	private $headers;
```
RadoslavGeorgiev: AFAICT we don't have any requests, which have specific headers, and those are not retrieved in `send_wcpay_request`. Should we maybe remove them?

includes/core/DataTransferObjects/class-data-transfer-object.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Return array of all properties.
+	 *
+	 * @return array
+	 */
+	public function to_array() {
+		$class = new ReflectionClass( $this );
+		$data  = [];
+		foreach ( $class->getProperties() as $reflectionProperty ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
RadoslavGeorgiev: Let's just change the variable names to snake case, and remove all of those PHPCS suppressions. 40% of this file is PHPCS comments at the moment :\

includes/core/DataTransferObjects/class-data-transfer-object.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * The purpose of this function is to DTOs data if passed array have a key where property exists in class . Since all DTOs are immutable, new DTO will be created with the data that was passed.
+	 *
+	 * @param mixed ...$values Values to change.
+	 * @return self Cloned object with changed properties.
+	 * @throws ReflectionException
+	 */
+	public function with( ...$values ) {
```
RadoslavGeorgiev: Considering we have setters now, does this method still make sense? The objects are not immutable.

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+		$this->cookies       = $cookies;
+		$this->response_data = json_decode( $body, true );
+	}
+
+	/**
+	 * Returns true if response is JSON.
+	 *
+	 * @return bool
+	 */
+	public function is_json_response() {
```
RadoslavGeorgiev: How/where is this method used?

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+		return 'application/json' === substr( $this->headers['content-type'] ?? '', 0, strlen( 'application/json' ) );
+	}
+
+	/**
+	 * Get body.
+	 *
+	 * @return string
+	 */
+	public function get_body() {
```
RadoslavGeorgiev: Does this method make any sense alongside `get_response_data`?

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+	public function get_cookies() {
+		return $this->cookies;
+	}
+
+	/**
+	 * Get response body as array.
+	 *
+	 * @return array|mixed
+	 */
+	public function get_response_data() {
```
RadoslavGeorgiev: Maybe just `get_data()`?

includes/core/DataTransferObjects/API/class-response.php
```diff
@@ -0,0 +1,256 @@
...
+	public function get_error_message() {
+		return $this->response_data['error']['message'] ?? null;
+	}
+
+	/**
+	 * Determinate does code exists in response data.
+	 *
+	 * @return bool
+	 */
+	public function has_code_in_response_data() {
```
RadoslavGeorgiev: Considering how errors are currently handled in the API client class, how do you imagine proceeding with this?

RadoslavGeorgiev (commented):
Are properties protected by any means?Why is PR addressing a new issue instead of https://github.com/automattic/woocommerce-payments/issues/4954?

---

I have left some comments. While addressing them, let's see how all of those new classes would tie in with the existing client. AFAICT we never create charges directly, this might be left-over functionality from earlier stages of the gateway. Let's pick one of the more common existing API methods, and implement it as a request class, while also using it within the codebase. I appreciate the test, but that's still very isolated.

Also, a reminder/question: The focus here is consumers. How is this PR helping them?

includes/core/DataTransferObjects/API/class-response.php
zmaglica: I can easily remove all "syntax sugar" functions from the Response class, since at this phase and usage they will not be used. If we decide to step away from the current function that executes HTTP requests to WCPay server, the functions like these could help to become an easy way to get certain data somewhere in the code without doing manual checks.

If we would like to keep Response class clean and without too much code, I will remove all unused and unnecessary methods we don't use at this moment.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: `get_response_data()` name looked more detailed since you are aware that it is response data, but since it is a response class, I will rename it to `get_data()`.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: `$body` and `get_body()` represents raw response from the server, and `get_response_data` is `json_decoded` version of `get_body`. I had an idea that you might need raw data from the server sometimes, and that's why I kept it.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: All validation and checks from current function that sends HTTP requests to wcpay server API have been implemented into Response DTO because we could create or own request method and use all validations, checks we might need. I can remove this function if needed and return it when we decide that we will use this approach.

includes/core/DataTransferObjects/class-data-transfer-object.php
zmaglica: The original idea was to make easier to create a new DTO where data is protected. I will remove it.

includes/core/ValueObjects/class-value-object.php
zmaglica: Good catch. Will fix it.

includes/core/class-loader.php
zmaglica: The problem with autoloader is that it was called for every loaded class after it's called, and if statement with namespace check had to be added to prevent any issues with loading. With all that in mind, I remove it to prevent any issues with performance. I haven't faced any performance issues, but I wasn't 100% certain, and that's why I removed it.

includes/core/Enums/class-http-methods.php
zmaglica: I will use the ones you linked here.

zmaglica:
> Are properties protected by any means?Why is PR addressing a new issue instead of #4954?
I separate it to for a batter planning and easily monitoring and handling. We could easily create another issue and create Request/Response for certain WCPay server endpoints without pushing too much code into original PR.  

> I have left some comments. While addressing them, let's see how all of those new classes would tie in with the existing client. AFAICT we never create charges directly, this might be left-over functionality from earlier stages of the gateway. Let's pick one of the more common existing API methods, and implement it as a request class, while also using it within the codebase. I appreciate the test, but that's still very isolated.
> 
The `create_charge` was the first request function in WC Pay HTTP client, and it is easier to understand the concept when you have a request with only two parameters. I will create add `Create_Charge` request with all tests and examples.
> Also, a reminder/question: The focus here is consumers. How is this PR helping them?

The majority of wcpay server requests are tightly coupled with the WooCommerce Payments server plugin. With that in mind, I would create a request class for Transactions or for Intentations.

includes/core/DataTransferObjects/class-data_transfer_object.php
KarlisJ: Would `fill` or `fill_with` be more descriptive method name?

includes/core/ValueObjects/API/Request/class-create-charge.php
KarlisJ: My point is that the presence of such getter indicates that there's a need for it. I'd say there's no need for getter(or any method) if we don't need it. If we'd be discussion parent class that provides functionality that be a different case but here I think the presence of these methods is misleading.

includes/core/Enums/class-wc-pay-endpoints.php
zmaglica: Yes. We would have to change class to enum, but the syntax should remain the same.

includes/core/DataTransferObjects/class-data_transfer_object.php
zmaglica: I remove this function since we allowed people to change data that needs to be immutable.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: DTO class have been removed and all logic is presented in parent class.

zmaglica: Since we are dealing with Requests and Responses, this class have been refactored and DTO abstract class is removed.

zmaglica: New structure and namespace have been used in new commit.

includes/core/Contracts/API/Response/interface-base-response.php
zmaglica: Fixed.

includes/core/DataTransferObjects/class-data-transfer-object.php
zmaglica: Fixed.

includes/core/ValueObjects/class-value-object.php
zmaglica: This value object is moved to Request class for better understanding.

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
zmaglica: Added.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: Make sense. I remove them.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: Done.

includes/core/ValueObjects/class-value-object.php
zmaglica: Instead of using regex, which is slow, I added a third argument to the explode function that limits the number of array elements.

includes/core/DataTransferObjects/class-data-transfer-object.php
zmaglica: The function was removed.

includes/core/DataTransferObjects/API/class-response.php
zmaglica: Since we don't need it, I removed it.

includes/core/class-loader.php
RadoslavGeorgiev: We chatted about this: Let's convert the core files, and add autoloading when the time comes. In the meantime, let's:

1. Make the directory names follow WPCS (lowercase in this case).
2. Include the files instead of having a `class-autoloader.php`, which doesn't autoload anything.

includes/core/ValueObjects/API/Request/class-create-charge.php
RadoslavGeorgiev: This one still applies. What if some of the setters was never called, and there's an incomplete object?

includes/core/DataTransferObjects/class-data-transfer-object.php
RadoslavGeorgiev: This file is gone, but I won't resolve the comment, as the same suppression appears in `includes/core/Server/Response/class-response.php`.

includes/core/DataTransferObjects/API/class-response.php
RadoslavGeorgiev: I don't think we have any cases where the raw response is currently being used. Let's remove this method in order to enforce proper implementation for edge cases. What do you think? I don't have any realistic scenarios in my mind, but in case it is needed, maybe it will be better to keep whatever it is needed for in the response class 🤔

includes/core/DataTransferObjects/API/class-response.php
RadoslavGeorgiev: > I can easily remove all "syntax sugar" functions from the Response class

I'm not sure what you mean by "syntax sugar" here, I only see a method, which is never used, and similar code in the API client class. If I saw this code being used instead, we could have worked with it, even though there would be similar code in the API client class. Otherwise we're left with two confusingly similar pieces of code, only one of which is ever used.

includes/core/class-loader.php
zmaglica: Maybe we can autoload/preload files with `spl_autoload_register` like I did in commits before, that will support preloading of core directory by using namespace to find a path to file. What do you think?

includes/core/DataTransferObjects/API/class-response.php
zmaglica: I will remove get body. We can easily return it if needed.

includes/core/class-loader.php
zmaglica: All folders have been replaced with WPCS standard from the last commit and autoloader class have been deleted and all requires have been placed in `class-wc-payments.php` file where majority of classes that are required are presented.

includes/core/Contracts/API/Request/interface-base-request.php
zmaglica: The purpose is to place some kind of skeleton interface in order if you would like to replace `Request` (or `Response`) class with something else in the future. If you follow the interface, the refactoring or adding a new Request class should be easy, and it won't break something. This concept is usually adopted in frameworks.

If you think that this is a bit overkill, we could simply delete it.

includes/core/Contracts/API/Response/interface-base-response.php
zmaglica: Make sense. I will implement that.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: Good catch. Fixed in https://github.com/Automattic/woocommerce-payments/pull/5012/commits/86f728c774f5e241fbe192dfeb927f04da905604

includes/core/ValueObjects/class-value-object.php
zmaglica: FIxed.

includes/core/Contracts/API/Response/interface-base-response.php
zmaglica: Added important functions as part of contract.

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: Create_Charge should cover all possible charge requests towards WC pay server. Allowing extending the class allows overwriting some core functionalities like validation and request preparing. In this case, it is better to create a brand-new request using base class as foundation.

Do you see benefit in making it non-final? I just wanted to make strict rules in building requests towards wcpay client.

includes/core/ValueObjects/API/Request/class-base-request.php
zmaglica: Addressed in https://github.com/Automattic/woocommerce-payments/pull/5012/commits/d0a3fabf120775bc5dc23d7a36addab47042eee3

includes/core/ValueObjects/API/Request/class-base-request.php
zmaglica: Addressed in https://github.com/Automattic/woocommerce-payments/pull/5012/commits/d0a3fabf120775bc5dc23d7a36addab47042eee3

includes/core/DataTransferObjects/class-data-transfer-object.php
zmaglica: Replaced Reflection class with get object vars here: https://github.com/Automattic/woocommerce-payments/pull/5012/commits/4d1571a8cd0cbb3d9214e7f6545ad5487c0a5a0a

includes/core/DataTransferObjects/API/class-response.php
zmaglica: This was just an easy way to get some data from the response object. Currently, we don't need functions like these, and I remove them.

includes/core/ValueObjects/API/Request/class-base-request.php
zmaglica: Make sense. The code is removed here: https://github.com/Automattic/woocommerce-payments/pull/5012/commits/b2681db0b2d7e6096b50a672e78de60a6dc7f15a

includes/core/server/request/class-create-charge.php
```diff
@@ -0,0 +1,97 @@
...
+		return Endpoints::INTENTIONS_API;
+	}
+
+	/**
+	 * Make sure that properties are filled.
+	 *
+	 * @return bool
+	 */
+	public function is_request_data_valid() {
+		return property_exists( $this, 'amount' ) && property_exists( $this, 'source_id' );
```
KarlisJ: I think this will always return true no matter if setters have been called or not.

includes/core/ValueObjects/API/Request/class-create-charge.php
RadoslavGeorgiev: The `is_request_data_valid()` implementations need some attention, but right now both of those calls are valid:

```php
$request->get_parameters();
$request->get_request_data();
```

I think we need to modify the access to `get_parameters` to limit it within the class.

Also, let's make `get_request_data` just `get_data`. We know we're working with the request already :)

includes/core/ValueObjects/API/Request/class-create-charge.php
RadoslavGeorgiev: On one hand I like the idea of `final` classes, on the other hand I'd like to see how this would play in real life.

[This line](https://github.com/Automattic/woocommerce-payments/blob/dev/wc-pay-request-class/includes/class-wc-payment-gateway-wcpay.php#L975) is a prime example of consumer code within WCPay, which should be isolated, and not live where it lives right now. Also, there's the `save_payment_method_to_platform` paramter for `create_and_confirm_intention`, which seems like something additional. With this class marked as `final`, I'm not sure how WooPay is expected to add the `woopay_has_subscription` paramter? 

> In this case, it is better to create a brand-new request using base class as foundation.

Are they expected to copy-paste this whole class within their codebase, rename it, and add the necessary parameters?

The main objective of the issue this PR resolves is such cases, and it seems we've left those behind. While most requests are likely better to implement in other PRs/issues, I'd like this PR to include:

1. An example of standard requests, used by WCPay, as it already does. However, I want to see those requests in action, not just in theory/standalone directory. You have already started the implementation of `create_intention`, but `create_and_confirm_intention` is used much more heavily through the plugin. Let's focus on that request, and do it in a way, where you can go to the checkout page, and complete a purchase through those new methods. Then we'll be able to discuss stuff [like this](https://github.com/Automattic/woocommerce-payments/blob/b2681db0b2d7e6096b50a672e78de60a6dc7f15a/includes/wc-payment-api/class-wc-payments-api-client.php#L2003-L2014), and decide what to do with that early (ex. a trait).
2. An example of a WooPay-specific parameter(s), isolated outside of the current `includes` folder, but still working. This means that somewhere in our code there should be a filter/action, which allows them to add those parameters safely, and this PR should include an example of how that is done. The example can then be converted to docs right before merging.

includes/core/server/request/class-create-charge.php
zmaglica: Good catch. Fixed.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -7,6 +7,8 @@
 
 defined( 'ABSPATH' ) || exit;
 
+use WCPay\Core\Contracts\Server\Request\Base_Request_Interface as WC_Pay_Request;
```
RadoslavGeorgiev: Do we need to alias the class here? Why not just use `Base_Request_Interface`?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1986,6 +1988,35 @@ public function is_in_test_mode() {
...
+	 * @param WC_Pay_Request $request - Request object.
+	 *
+	 * @return Response
+	 * @throws InvalidArgumentException|API_Exception
+	 */
+	public function send_wcpay_request( WC_Pay_Request $request ) {
+		$params = $request->get_request_data();
+
+		// If level3 data doesn't contain any items, add a zero priced fee to meet Stripe's requirement.
+		if ( ! isset( $params['level3']['line_items'] ) || ! is_array( $params['level3']['line_items'] ) || 0 === count( $params['level3']['line_items'] ) ) {
```
RadoslavGeorgiev: Is there a better place for this code? A generic request method should not contain such specificity. Maybe a small trait, or at least moving it to `Request->get_request_data()`?

includes/wc-payment-api/class-wc-payments-api-client.php
RadoslavGeorgiev: This class is solely responsible for contacting the server. Why would we need `wcpay` or `server` in the name? Not that I'm suggesting it (as it's already there), but I would prefer renaming the whole class instead of this method.

Why not just `send_request()`?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1986,6 +1988,35 @@ public function is_in_test_mode() {
...
+					'product_description' => 'The order is empty',
+					'quantity'            => 1,
+					'tax_amount'          => 0,
+					'unit_cost'           => 0,
+				],
+			];
+		}
+
+		$response = $this->request( $params, $request->get_route(), $request->get_method(), $request->is_site_specific(), $request->use_user_token(), true );
+		return Response::create_from_wc_pay_server_response( $response );
```
RadoslavGeorgiev, line 2001: Could we replace this with $request->create_response( $response ) in order to allow the request class to decide which `create_from_wc_pay_server_response` to use whenever we start implementing specific responses?

includes/core/contracts/server/response/class-base-response-interface.php
```diff
@@ -0,0 +1,40 @@
...
+	 */
+	public function to_array();
+
+	/**
+	 * Create DTO from WCPay response.
+	 *
+	 * @param array $response Server response.
+	 * @return Response
+	 */
+	public static function create_from_wc_pay_server_response( $response );
```
RadoslavGeorgiev, line 38: Something feels wrong here. The idea of `WCPay\Core\Contracts\API\Response` being a contract, is to make it more universal, right? Here we're binding it to the WCPay server by declaring this method. If we are to keep this contract (I'm still not convinced it would help at any point), let's move this method to the base response.

includes/core/contracts/server/response/class-base-response-interface.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace WCPay\Core\Contracts\API\Response;
+
+use WCPay\Core\Server\Response;
+use WP_Http_Cookie;
+
+/**
+ * Base response interface.
+ */
+interface Base_Response_Interface {
```
RadoslavGeorgiev, line 16: Let's remove `Base_` here? Then we'll have `Response_Interface` here, which has more than enough meeting already, and the base response class can be just `Response` without causing any conflicts.

includes/core/contracts/server/request/class-base-request-interface.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\Contracts\Server\Request;
+
+/**
+ * Base request interface.
+ */
+interface Base_Request_Interface {
```
RadoslavGeorgiev, line 13: Similar to the response, why not just have `Request_Interface` here, and `Request` instead of `Base_Request`?

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
+<?php
+/**
+ * Base request value object.
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Core\Server\Request;
+
+use \WCPay\Core\Contracts\Server\Request\Base_Request_Interface;
```
RadoslavGeorgiev, line 10: The first backslash is redundant.

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * Validate and get request data.
+	 *
+	 * @return array
+	 * @throws \InvalidArgumentException
+	 */
+	final public function get_request_data() {
+		if ( ! $this->is_request_data_valid() ) {
+			throw new \InvalidArgumentException( 'Request data is not valid' );
```
RadoslavGeorgiev, line 81: Should we leave the exception to `is_request_data_valid`, which could also be `validate_request_data`? Then we can be more precise than this.

`Request data is not valid` is not valid (pun intended). It is rather incomplete. If the validation method handles this, it could also describe which properties are missing.

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+		}
+		if ( ! in_array( $this->get_method(), [ \Requests::GET, \Requests::HEAD, \Requests::OPTIONS, \Requests::POST, \Requests::PUT, \Requests::PATCH, \Requests::DELETE ], true ) ) {
```
RadoslavGeorgiev, line 83: What about `if ( ! defined( 'Requests::' . $this->get_method() ) )`?

FWIW I see the value of true enums, we could just have them as the return type of the abstract `get_method()`, and this condition would be not required.

RadoslavGeorgiev: Also, this is not something that feels in place within `get_request_data()`, as it's a separate concern. What do you think about this?

```php
final public function send( WC_Payments_API_Client $api_client ) {
	if ( ! $this->is_request_data_valid() ) {
		throw new \InvalidArgumentException( 'Request data incomplete' );
	}

	if ( ! defined( 'Requests::' . $this->get_method() ) ) {
		throw new \InvalidArgumentException( 'Request method is not valid' );
	}

	return $this->create_response(
		$this->request(
			$this->get_parameters(),
			$this->get_route(),
			$this->get_method(),
			$this->is_site_specific(),
			$this->use_user_token(),
			true
		)
	);
}
```

Maybe this doesn't play best with the concept of requiring a specific class for a specific request, but I'm curious what you think?

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+	 *
+	 * @param string $value Value to validate.
+	 * @return bool
+	 * @throws \InvalidArgumentException
+	 */
+	public function is_string_and_not_empty( $value ) {
+		if ( is_string( $value ) && $value ) {
+			return true;
+		}
+		throw new \InvalidArgumentException( "The property $value is invalid! It needs to be string type and not empty" );
```
RadoslavGeorgiev, line 116: I'm not sure if this helper method really helps. Imagine the charge class receiving an empty parameter:

```
The property  is invalid! It needs to be string type and not empty
```

If the charge class generated this, it could properly state

```
The property `source_id` cannot be an empty string.
```

Alternatively, this method could receive the name of the property as a second parameter.

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Create new request class without using new Class().
+	 *
+	 * @param mixed $arguments Constructor arguments (if any).
+	 * @param bool  $splat_arguments Split arguments if needed. Used to pass arguments as an array where elements will be used as a arguments for constructor.
+	 * @return Base_Server_Request
+	 */
+	public static function new( $arguments = null, $splat_arguments = false ) {
+
+		$class = get_called_class();
```
RadoslavGeorgiev, line 146: You can just use the `static` keyword for the name of the called class.

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	/**
+	 * Create new request class without using new Class().
+	 *
+	 * @param mixed $arguments Constructor arguments (if any).
+	 * @param bool  $splat_arguments Split arguments if needed. Used to pass arguments as an array where elements will be used as a arguments for constructor.
+	 * @return Base_Server_Request
+	 */
+	public static function new( $arguments = null, $splat_arguments = false ) {
```
RadoslavGeorgiev, line 144: How is this method used? I don't see any examples, and it seems kind of useless, given there are no constructors for requests.

includes/core/server/request/class-base-server-request.php
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	/**
+	 * Dynamic getters and setters for value objects without them.
+	 *
+	 * @param string $name Method name.
+	 * @param array  $arguments Method arguments.
+	 * @return self|mixed
+	 */
+	public function __call( $name, $arguments ) {
```
RadoslavGeorgiev, line 165: There are a couple of `if`s within the method, but no exception for unknown methods. Could you please throw one?

includes/core/server/request/class-create-intention.php
```diff
@@ -0,0 +1,394 @@
...
+		$this->order_number = $order_number;
+		return $this;
+	}
+
+	/**
+	 * Confirm intention.
+	 *
+	 * @return $this
+	 */
+	public function confirm_intention() {
```
RadoslavGeorgiev, line 293: This is super misleading. Maybe change it to `should_confirm_intention()`?

includes/core/server/request/class-create-intention.php
```diff
@@ -0,0 +1,394 @@
...
+		if ( $this->payment_method_id ) {
+			$request['payment_method'] = $this->payment_method_id;
+		}
+
+		if ( $this->payment_methods ) {
+			$request['payment_method_types'] = $this->payment_methods;
+		}
+
+		if ( $this->save_payment_method_to_store ) {
+			$request['setup_future_usage'] = 'off_session';
```
RadoslavGeorgiev, line 355: Why not store those defaults when defining the class properties?

RadoslavGeorgiev (changes requested):

includes/core/ValueObjects/API/Request/class-create-charge.php
zmaglica: We could add these extra fields or custom logic could be part of the request builders via functions. The good example of this function inside Create_Intention is `confirm_intention()` (probably I would have to rename this function since it could be a misleading)

For example:

This line of code could be replaced and implemented inside the Request class like this:

  Current code:
  
```
if ( '1' === $order->get_meta( '_woopay_has_subscription' ) ) {
         $additional_api_parameters['woopay_has_subscription'] = 'true';
}
```

Proposed approach using Request class

// Inside Create_Intention class, we could create additional methods to add small functionalities like this. One of the examples is to create something like this:

```
public function maybe_has_woopay_subscription( $order ){
   if ( '1' === $order->get_meta( '_woopay_has_subscription' ) ) {
         $this->additional_parameters['woopay_has_subscription'] = 'true';
    }
}
$intent = new Create_Intention();
$intent->maybe_has_woopay_subscription( $order )
```

Or something that could be useful by setting `WP_Order` in `Create_Intention`

```
//Inside Request class.

private $wp_order;

public function set_wp_order( $order ) {
    // ... will validate is order object ...
    $this->wp_order = $order;
}
public function maybe_has_woopay_subscription( ){
   if ( $this->wp_order && '1' === $order->get_meta( '_woopay_has_subscription' ) ) {
         $this->additional_parameters['woopay_has_subscription'] = 'true';
    }
}
$intent = new Create_Intention();
$intent->set_wp_order ( $order );
$intent->maybe_has_woopay_subscription() //Maybe add this automaticlly ?
```

But adding a bunch of these cases could be problematic because you don't know all of them, so a practical approach would be to remove final mark and allow other teams to extend the class, or they could add their own methods via PRs in the future.

I can fully understand your point now. Let me come up with some of the ideas. I will commit them with examples for easier understanding with a code examples I could find in our codebase.

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: We don't need it. Fixed in https://github.com/Automattic/woocommerce-payments/pull/5012/commits/b2681db0b2d7e6096b50a672e78de60a6dc7f15a

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: I will keep it inside `get_request_data()` since keeping it in trait is great if this code will be used somewhere else, since we are not planning to use this logic somewhere else in core (for now).

zmaglica: Addressed in https://github.com/Automattic/woocommerce-payments/pull/5012/commits/26096e9af2b9689a60d1d7b71ae9fac53c441b97

includes/wc-payment-api/class-wc-payments-api-client.php
zmaglica: The reason behind naming like this is that I wanted to be explicit. Requests and responses can have many methods that could be added later on, and making them explicit like this helps in understanding the matter, and you really know that you are creating from wc pay server response.

With your approach, we should refactor it by removing constructor injection and replace it with setters. This allows us to validate data at some point.

I will refactor it.

PR #5034, opened 2022-11-01
Fix error that would happen when visiting the plugins page

leonardola:

This PR tries to fix an error that @brettshumaker encountered when updating WC Payments in his ephemeral site.

#### Testing instructions
Checkout this branch and visit the plugins page. Make sure the Settings link shows up in the WC Payments plugin and no error is registered in your logs or shows up

-------------------


brettshumaker (approved):
Thanks for the fix! Tests well for me.

PR #5037, opened 2022-11-02
Send merchant's setting for `ship_to_billing_address_only` to platform checkout

Mayisha:
Sending the merchant's selected settings for `ship_to_billing_address_only` to platform checkout with platform checkout session request.

#### Testing instructions
Needs to be tested together with the platform checkout PR.


bborman22 (approved):
I see some tests are failing for unrelated reasons, but might still need a changelog entry even just as a patch? Once we have a changelog this update LGTM!

PR #5039, opened 2022-11-02
Refactor WC_Payment_Gateway_WCPay Part 2

leonardola:
Related to #823 
This PR is the second part of a series of refactors of the WC_Payment_Gateway_WCPay class.

#### Testing instructions
- As a shopper add a product to the cart
- Make sure you can see the card inputs
- Buy the product
- Test again using UPE
- As a shopper buy any product using WooPay

- Change and save WC Pay settings via wp-admin and make sure they are saved correctly and reflect in the shopper flow

- Execute other [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) that you think are necessary

-------------------

**Post merge**


includes/class-wc-payments-gateway-wcpay-settings.php
```diff
@@ -0,0 +1,572 @@
...
+use WCPay\Logger;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Gateway settings class for WooCommerce Payments.
+ */
+class WC_Payments_Gateway_WCPay_Settings {
```
RadoslavGeorgiev: Would you consider using namespaces here and simply calling the class `Settings` within the `WCPay` namespace?

includes/class-wc-payments-gateway-wcpay-settings.php
```diff
@@ -0,0 +1,572 @@
...
+
+	const OPTIONS = [
```
RadoslavGeorgiev: Let's name this settings, as options are still something slightly different.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1587,48 +1398,18 @@ public function generate_account_statement_descriptor_html( $key, $data ) {
...
-				return $this->get_deposit_schedule_weekly_anchor();
-			case 'deposit_schedule_monthly_anchor':
-				return $this->get_deposit_schedule_monthly_anchor();
-			case 'deposit_delay_days':
-				return $this->get_deposit_delay_days();
-			case 'deposit_status':
-				return $this->get_deposit_status();
-			case 'deposit_completed_waiting_period':
-				return $this->get_deposit_completed_waiting_period();
+		if ( in_array( $key, WC_Payments_Gateway_WCPay_Settings::OPTIONS, true ) ) {
```
RadoslavGeorgiev: Do you think a more drastic approach where **all** options are accessed through `WC_Payments_Gateway_WCPay_Settings` might make sense? Here's what I mean:

1. `WC_Payments_Gateway_WCPay_Settings` is constructed with a dependency on the gateway in order to be able to access `$this->gateway->get_option()` for settings/as a fallback.
2. All usages of `WC_Payments()->get_gateway()->get_option()` are replaced with proper getters within the settings class, ex. `WC_Payments()->get_settings()->is_enabled()`.
3. This method (`get_option` here) is completely removed.

I know this will involve a lot of file changes (ex. there are 36 changed files in https://github.com/Automattic/woocommerce-payments/pull/4986 just for `is_in_test_mode()` and `is_in_dev_mode()`), but I see a few benefits:

- ~It would not matter~ The gateway would not need to know where the option comes from (cached account data vs gateway settings).
- All options will be clearly accessible from a single place.
- We'd be able to utilize proper types/casting.

Expanding a bit further, we could also use the field types from `WC_Payments_Gateway_WCPay_Settings::get_form_fields()` and parse the value, ending the annoying `'yes' === $gateway->get_option( 'something' )`, which also premises bugs (imagine assuming it's a boolean, then even 'no' will be true). This can be done either through magic methods (`__get` for `$settings->enabled`) and PHPDoc `@property` hints (see [this example](https://github.com/Automattic/woocommerce-payments/pull/4986/files#diff-1f0e10aa8fe69406fae8d3d26368dd8b46e67c541485b5a84669086bb0150f2fR16)), or explicit methods like this:

```php
/**
 * @property-read bool $enabled Indicates if the gateway is enabled.
 */
class {
	private function get_option( $name ) {
		$value = $this->gateway->get_option( $name );

		// Of course, this formatting would change.
		$type = $this->get_form_fields()[ $name ]['type'];
		switch ( $type ) {
			case 'checkbox':
				$value = ( 'yes' === $value );
				break;
		}

		return $value;
	}

	public function is_enabled() : bool {
		return $this->get_option( 'enabled' );
	}

	public function __get( $name ) {
		if ( isset( $this->get_form_fields()[ $name ] ) ) {
			return $this->get_option( $name );
		} else {
			// throw exception
		}
	}
}
```

This example combines both approaches, but we should stick to only one of them.

RadoslavGeorgiev (commented):
Nice work! I have a couple of suggestions :)

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -130,6 +130,13 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 */
 	protected $payment_method_capability_key_map;
 
+	/**
+	 * WC_Payments_Gateway_WCPay_Settings instance.
+	 *
+	 * @var WC_Payments_Gateway_WCPay_Settings
+	 */
+	protected $wc_pay_settings;
```
brianyu28: As a naming convention, I believe we generally use either `wcpay` (no underscore) or `wc_payments` (with underscore). For consistency, what do you think about `wcpay_settings` here instead of `wc_pay_settings`?

includes/class-wc-payments-order-service.php
```diff
@@ -814,4 +836,42 @@ private function update_order_status( $order, $order_status, $intent_id = '' ) {
...
+
+	/**
+	 * Manages customer details held on WCPay server for WordPress user associated with an order.
+	 *
+	 * @param WC_Order $order   WC Order object.
+	 * @param array    $options Additional options to apply.
+	 *
+	 * @return array First element is the new or updated WordPress user, the second element is the WCPay customer ID.
+	 */
+	public function manage_customer_details_for_order( $order, $options = [] ) {
```
brianyu28, line 840: Given that we need test mode information every time this function is called, I'd suggest making it its own parameter for this function instead of including it in `options`.

brianyu28 (commented):
Thanks @leonardola, overall this looks good to me! A couple of small comments in addition to what's been mentioned in the previous review. Though per paJDYF-6lv-p2, should we consider waiting to merge these changes?

includes/class-wc-payments-gateway-wcpay-settings.php
leonardola: Fixed by e3a64238dcd23f440e9fed479e94c18d3d53783c

includes/class-wc-payments-gateway-wcpay-settings.php
leonardola: Fixed by 58e0a75b6c0f63f485ade4841c4fb4d40185016a

includes/class-wc-payment-gateway-wcpay.php
leonardola: `WC_Payment_Gateway_WCPay` already depend on `Settings` if we make `Settings` depend on `WC_Payment_Gateway_WCPay` then we would have a circular reference. 

This could be avoided by using `WC_Payments::get_gateway_settings` but that's not the most elegant way to do this. Though we already have code like that in other places. So I am not totally against it but not a big fan either. 

What do you think?

leonardola:
@brianyu28 
> Though per paJDYF-6lv-p2, should we consider waiting to merge these changes?

For sure. I'll probably only merge this next year after things get back to normal. This is not a urgent change

includes/class-wc-payment-gateway-wcpay.php
leonardola: Fixed by 3f55c6ceb8bc74a66093e63f286b9967c025a36b

includes/class-wc-payments-order-service.php
leonardola: Fixed by 650d2cbb61b6507909a4d643ddc54d2b3ae21a61

includes/class-wc-payment-gateway-wcpay.php
leonardola: Fixed by 3f55c6ceb8bc74a66093e63f286b9967c025a36b

leonardola:
I added a blocked label so we don't merge this before January

leonardola:
@alefesouza this PR is now unblocked and is ready to be reviewed

alefesouza (approved):
LGTM :shipit:

leonardola:
Blocked by pdpOw2-Pn-p2 till version 5.5 is launched

PR #5060, opened 2022-11-07
Fetch authorization data only if the payment is marked for manual capture

### LINKED ISSUE #5061: [Authorizations][Payment details] Avoid fetching authorization data when payment intent is not set to manual capture
### Describe the bug

After we merged this [PR](https://github.com/Automattic/woocommerce-payments/pull/5006), we will display a notice in the Payment details page to inform the user when the payment needs to be captured. This is a feature behind a feature flag. 
The problem is we are not checking if the payment is marked to be manually captured before fetching the authorization data. This request can be spared if the capture method is set to automatic, since there will not be any authorization data in that case

### To Reproduce

1. Go to WP Admin > Orders
2. Enter the details of an order that was paid with a payment method not set to manual capture.
3. Navigate to the payment details page from there

### Actual behavior

A GET request to `/payments/authorizations/<payment_intent_id>` and it will return an error

### Screenshots

Screen Shot 2022-11-07 at 12 53 16

### Expected behavior

A GET request to `/payments/authorizations/<payment_intent_id>` **should not** be triggered

mgascam:
Fixes #5061 

This change implements the logic to avoid fetching authorization data from the server if the payment is not marked as manual capture. In this case it makes no sense to retrieve the data from the server, since there will be no authorization linked to the payment

* Modifies useAuthorization hook to accept a new parameter: `requiresCapture`
* Inside the hook, use the flag to call the selector/resolver only if `requiresCapture` is true. 

#### Testing instructions

_Payment with automatic capture_
* Navigate to the details page of a payment set to be captured automatically (this is the default behavior, so any payment from transactions details should work)
* Check that there is no GET request to `/payments/authorizations/<payment_intent_id>`

_Payment with manual capture_

- Enable `Issue an authorization on checkout and capture later` option in WP Admin > Payments > Settings
- Set `displayAuthorizations` and `displayCaptureAuthorizationSection` to true
- Make sure WCPay client is connected to local server, and that the server is listening for webhooks events from Stripe
- Purchase an order
- Navigate to WP Admin > Payments > Transactions > Uncaptured tab
- Click on the row for the order you just created
- A GET request to `/payments/authorizations/<payment_intent_id>` should be fired
- The Capture section should be displayed. It should contain the deadline date and the Capture CTA button

-------------------

**Post merge**

### DIFF
--- a/client/payment-details/summary/index.tsx
+++ b/client/payment-details/summary/index.tsx
@@ -107,9 +107,17 @@ const PaymentDetailsSummary = ( {
 		: placeholderValues;
 	const renderStorePrice =
 		charge.currency && balance.currency !== charge.currency;
+
+	// We should only fetch the authorization data if the payment is marked for manual capture
+	const shouldFetchAuthorization =
+		charge.amount !== charge.amount_captured &&
+		charge.amount_refunded === 0 &&
+		displayCaptureAuthorizationSection;
+
 	const { authorization } = useAuthorization(
 		charge.payment_intent as string,
-		charge.order?.number as number
+		charge.order?.number as number,
+		shouldFetchAuthorization
 	);
 
 	return (
--- a/changelog/fix-fetch-auth
+++ b/changelog/fix-fetch-auth
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fetch authorization data in payment details page only when the payment needs manual capture
--- a/client/data/authorizations/hooks.ts
+++ b/client/data/authorizations/hooks.ts
@@ -71,16 +71,19 @@ export const useAuthorizationsSummary = (
 
 export const useAuthorization = (
 	paymentIntentId: string,
-	orderId: number
+	orderId: number,
+	requiresCapture = true
 ): {
 	isLoading: boolean;
 	doCaptureAuthorization: () => void;
-	authorization: Authorization;
+	authorization?: Authorization;
 } => {
 	const { authorization, isLoading } = useSelect( ( select ) => {
 		const { getAuthorization, isResolving } = select( STORE_NAME );
 		return {
-			authorization: getAuthorization( paymentIntentId ),
+			authorization: requiresCapture
+				? getAuthorization( paymentIntentId )
+				: null,
 			isLoading: isResolving( 'getAuthorization', [ paymentIntentId ] ),
 		};
 	} );


client/payment-details/summary/index.tsx
```diff
@@ -107,9 +107,17 @@ const PaymentDetailsSummary = ( {
...
+	const shouldFetchAuthorization =
+		charge.amount !== charge.amount_captured &&
+		charge.amount_refunded === 0 &&
+		displayCaptureAuthorizationSection;
+
 	const { authorization } = useAuthorization(
 		charge.payment_intent as string,
-		charge.order?.number as number
+		charge.order?.number as number,
+		shouldFetchAuthorization
```
anu-rock, line 120: Nitpick: Perhaps we can use `shouldFetchAuthorization` outside rather than inside the hook? That'll make it the responsibility of `PaymentDetailsSummary` to decide whether to call the hook. IMO, that will make the decision clear and upfront.

But I also understand what you've done, which is better from a stylistic standpoint.

anu-rock (approved):
I did some manual testing. The fix works!

Scenarios tested

✅ Loading an instantly captured payment did not trigger a call to the get authorization endpoint.
✅ Loading a manually captured payment with missing entry in charges cache did not trigger a call to the get authorization endpoint.

Pre-approving this PR with a nitpick.

client/payment-details/summary/index.tsx
mgascam: Yes that was my intention but React demands that hooks cannot be called [conditionally](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level).
For this reason, I decided to modify the hook to call the selector/resolver depending on the `shouldFetchAuthorization` flag.
The React docs mention the possibility of using a custom hook, but I'm not too familiar with that. Maybe we can explore it as an improvement?

client/payment-details/summary/index.tsx
```diff
@@ -107,9 +107,17 @@ const PaymentDetailsSummary = ( {
...
 	const renderStorePrice =
 		charge.currency && balance.currency !== charge.currency;
+
+	// We should only fetch the authorization data if the payment is marked for manual capture
+	const shouldFetchAuthorization =
+		charge.amount !== charge.amount_captured &&
+		charge.amount_refunded === 0 &&
+		displayCaptureAuthorizationSection;
+
 	const { authorization } = useAuthorization(
```
kalessil, line 117: Instead of dispatching `shouldFetchAuthorization` into `useAuthorization`, can we initialize with null instead? Something like a ternary operator?

I don't see the point to modify `useAuthorization`,  as it violates the single responsibility approach.

kalessil (changes requested):

client/payment-details/summary/index.tsx
elazzabi: I think this is related to Vlad's comment above ^, and I also agree to keep it outside `useAuthorization`. One solution to this, and to get past the "Hook can not be called conditionally" error, is to call the conditional inside `useEffect`:

```
useEffect( () => {
  if( shouldFetchAuthorization ) {
    // Call useAuthorization here
  } else {
    // Do something else here
  }
}, [ charge ] );
```

And to make the variables as arguments to `useEffect`, this way we'll re-run the `useEffect` whenever there is a change (ie, we got data back from the server)

mgascam: Thanks @elazzabi I'll have a look to your suggestion, but it seems to me that we'd still be calling useAuthorization conditionally?

elazzabi: Actually, I just tried this locally and it worked:

```js
let authorization = null;

if (shouldFetchAuthorization) {
  const result = useAuthorization(
    charge.payment_intent as string,
    charge.order?.number as number
  );

  authorization = result.authorization;
}
```
I'm not getting the conditionally error 🤔

kalessil:
@elazzabi @mgascam, if https://github.com/Automattic/woocommerce-payments/pull/5060#discussion_r1015462533 is working, please make adjustments, Otherwise, let's merge the fix and make followup refactoring on Tuesday.

kalessil (approved):
With reference to my last comment, preapproving this PR.

client/payment-details/summary/index.tsx
mgascam: I tried but get this error when the linter runs before commit:

```
✖ eslint:

/Users/miguelgasca/code/wc-pay/client/payment-details/summary/index.tsx
  120:18  error  React Hook "useAuthorization" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks

✖ 1 problem (1 error, 0 warnings)
```

elazzabi: Aaah it was the lint, I was expecting the console to throw... Sorry for the false hope 🙇

kalessil: Ok, then let's add a changelog and merge the PR @mgascam .

mgascam: For reference, I will follow up with a different approach in a separate PR. My idea is to create a wrapper component that will be rendered conditionally. It will be responsible for fetching the authorization data. This way we will call the hook always. 
`PaymentDetailsSummary` component will no longer use the `useAuthorization` hook. It will be responsible only of rendering the new component or not.

PR #4781, opened 2022-09-20
Save tokens for Link payment method

dpaun1985:
Fixes #4703 

#### Changes proposed in this Pull Request
For subscriptions, the payment method token should be saved in order to be re-used for subscription renewal. 
This PR updates Link payment method to save the token.

#### Testing instructions

* **Scenario 1**
* Create a new subscription order and pay it using Link
* Payment should be completed
* The payment of renewal for the subscription should be processed without errors. To check this, use a 1 day subscription renewal.
* Subscription fee should be applied
* **Scenario 2**
* Create a new subscription with 1 day trial version order and pay it using Link
* Payment should be completed
* The payment for the subscription should be processed without errors after the trial ends.
* Subscription fee should be applied
* **Scenario 3**
* Install [Woocommerce Subscriptions](https://woocommerce.com/products/woocommerce-subscriptions/)
* When WCS is enabled it will take over the subscription functionality and will be managed by WCS.
* In order to trigger subscription renewals, you can follow the steps [ [here](https://woocommerce.com/document/testing-subscription-renewal-payments/) ]
* Repeat the tests above. 
* Subscription fee should NOT be applied.

Notes: 
1. In order to test the built-in subscription, you have to create a 1day subscription and wait for Stripe to renew the subscription. You cannot test it with Stripe's Subscription Clocks, because this will create a duplicate customer, and a new payment method ( credit card payment method ). 
2. If the subscription is not renewed after 1 day on client side, you can check the events on Stripe side for that subscription because you might have the listen webhook closed on your local when Stripe renewed the subscription.
3. [Subscriptions plugin](https://woocommerce.com/products/woocommerce-subscriptions/) overlaps the built in subscriptions.
4. In order to test the subscriptions plugin, please use the steps from [here](https://woocommerce.com/document/testing-subscription-renewal-payments/). 
-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
[...]

Issue #4720, opened 2022-09-08
Multiple UPE - Stripe Link compatibility


harriswong:
### Describe the issue
This is part of the Multiple UPE project (https://github.com/Automattic/woocommerce-payments/pull/3968). 

The multiple UPE POC uses classic element for credit card checkout, but Stripe Link requires UPE. Does this mean Stripe Link no longer work with multiple UPE? If so, how should we handle this? 

### Accepted criteria
- Find out if the multiple POC branch works with Stripe Link 
- If not, please spearheads a conversation/solution with the team(s).

Mayisha:
[Stripe Link](https://woocommerce.com/document/payments/woocommerce-payments-stripe-link/) seems very similar to WooPay in general. When Stripe Link is enabled-

- On the checkout page the email field is rendered on top of Billing Address (same as WooPay)
- When a user inputs an email address that does not exist in Stripe Link, a checkbox appears below the credit card option. (very similar experience as WooPay)
![new user](https://user-images.githubusercontent.com/33387139/190491060-4dc77380-bb68-4310-bc9e-12e4df534d49.png)

- On place order, a Link user is created (same experience as WooPay)
- When a user inputs an existing Link email address, an OTP Modal appears. (same experience as WooPay)
![otp](https://user-images.githubusercontent.com/33387139/190491307-531c7b4d-5ca0-4616-ac79-0e7cf12d1234.png)

- On successful OTP verification, another modal appears with the account data and a button to autofill the current merchant checkout page with the selected Link account data (this step is different from WooPay as the user always stays on the merchant's checkout page)
![account](https://user-images.githubusercontent.com/33387139/190491353-eb6d201b-564a-422b-93d3-eb0cfa9ba371.png)

- The OTP modal appears when the user types in an existing email. Otherwise, for prefilled email field the input box shows a button that initiates the OTP modal. (WooPay does not have this extra button)
![link button](https://user-images.githubusercontent.com/33387139/190491730-0db1637c-e3ad-4ae5-abc8-bd3af366c518.png)

harriswong:
Thanks @Mayisha for the screenshot and notes!

🤔  If we are using the same email address, how are we going to switch between WooPay or Link?

Mayisha:
> Find out if the multiple POC branch works with Stripe Link

### Existing Link customer flow 
✅ When a user inputs an existing Link email address, an OTP Modal appears as expected.
[...]

PR #5082, opened 2022-11-09
Merge develop into poc/upe instances multiplied

timur27:
### Description in progress
Fixes https://github.com/Automattic/woocommerce-payments/issues/5109

Merge the latest changes from `develop` and ensure, that `poc/upe-instances-multiplied` with all the changes on it are stable. 

https://github.com/Automattic/woocommerce-payments/pull/4850 is the PR that affects our split UPE implementation. With this refactor, we now have two new classes, which represent the checkout - `WC_Payments_Checkout` and `WC_Payments_UPE_Checkout`. These classes are encapsulating the UI field rendering.

### Following commits are introduced besides the regular merging and conflict resolving:
[Minor fixes without any logic change](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/bcb1196aa3ba3406a8483fa8600bd7486c87639e)

[Minor](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/bcb1196aa3ba3406a8483fa8600bd7486c87639e) | [fixes](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/c38ddf2c1af6fd3a155c73fd6cb5f48b04dc9f0a) | [to avoid regression](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/850f2d49e1322c00cddc6a30880041f3300f0247)

[Fixing](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/672adb925f2ccbb7c7dad03c2a2324744868be91) | [tests](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/c84178eaacc28ed2986a0c7d25a56d44d68964c4) | [suite](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/4a4d03c2bf3ea6205b4f21684b097706f0064b1d)

[Resolving conflicts and ensuring tests](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/381323514ac2a63bfa137b6ec3736f0838fdf5f7) pass after both [token behavior was corrected](https://github.com/Automattic/woocommerce-payments/pull/4670) **and**  [the logic for LINK tokens was updated](https://github.com/Automattic/woocommerce-payments/pull/4781). ⚠️ There is an additional logic introduced, basically a copy-paste from [the SEPA tokes behavior](https://github.com/Automattic/woocommerce-payments/pull/4670/files#diff-ad7970a483b7b3582e959f6236b3c69739d80349cb2236536c7db538726ffe2f). We need to make sure, that tokens for LINK are still working as expected. I think we can do it after https://github.com/automattic/woocommerce-payments/issues/4720. 

[Initializing payment checkouts in `class-wc-payments`](https://github.com/Automattic/woocommerce-payments/pull/5082/commits/c2d67f1c8beb7b81ce98b88e181f881c1f57bd00) to reflect the latest changes from `develop` which introduce `WC_Payments_Checkout` and `WC_Payments_UPE_Checkout` classes.

https://github.com/Automattic/woocommerce-payments/pull/5082/commits/15f08c75ea7ede68ccfc2dba8fc836db17a8db9a to let the `WC_Payments_UPE_Checkout` store the list of payment gateways to reflect the split UPE work and be able to render the UI for each split payment method.

-------------------
`WC_Payments_Checkout` and `WC_Payments_UPE_Checkout` are now responsible for rendering the UI for a gateway. Since the refactor was done before the UPE split, it takes a single method into the account even for `WC_Payments_UPE_Checkout`. 

The current logic of rendering gateways on `develop` (after the refactor, but before this PR) [keeps the gateway in the checkout object](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L68). So, when UPE is turned on, `WC_Payments_UPE_Checkout` with the `$card_gateway` is being initialized and [when the time comes](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L409-L411), the `$card_gateway` is being successfully rendered by `WC_Payments_UPE_Checkout` as a single payment element with all the bits and pieces inside of it. 

Our mission was to use this brand new checkout class in order to render [all the registered payment methods](https://github.com/Automattic/woocommerce-payments/pull/3968/files#diff-57ee486c4a3ba4bae478eb31255c9a968b5b4eccd04b27f512e4003440d6f821R498-R524), which are now represented by a group of split payments, instead of only `$card_gateway`.

There were multiple options to achieve this. First of all, we could create a `WC_Payments_UPE_Checkout` object for each gateway. In practice, [with ten methods available](https://github.com/Automattic/woocommerce-payments/pull/3968/files#diff-57ee486c4a3ba4bae478eb31255c9a968b5b4eccd04b27f512e4003440d6f821R337-R346), doing this seems not optimal. I think that the reason for the refactor was to abstract the payment fields UI rendering part to a separate class, and with the approach I describe, we would multiply the class with the same purpose ten times and use the memory without any meaningful reason. 

Another option was to ensure, that only one instance of the `WC_Payments_UPE_Checkout` is created, and each $gateway provides itself as a callback parameter of the action hook [in this place](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L409-L411). Then, we could set the `$gateway` in the checkout class and use it while rendering. However, this approach most probably leads to the violation of Liskov substitution principle, since `payment_fields()` method's definition will be updated by adding a new parameter (`$gateway`). We then would need to `set_gateway` in the `payment_fields()` method and thus change the state of the class. All this seemed a bit "unusual" to me.

Finally, I went with the approach, which allows us to [store all the payment gateways](https://github.com/Automattic/woocommerce-payments/blob/7115db427f1ef0f14f9bcdd802dcdf452486e212/includes/class-wc-payments-upe-checkout.php#L80) in the `WC_Payments_UPE_Checkout` class. When rendering the payment fields, [we invoke the callback from a new `wc_payments_set_gateway` hook I added](https://github.com/Automattic/woocommerce-payments/blob/4a4d03c2bf3ea6205b4f21684b097706f0064b1d/includes/payment-methods/class-upe-payment-gateway.php#L447-L449) and then proceed with the regular flow. This allowed us to keep the purpose of the `WC_Payments_UPE_Checkout` class by creating only one instance of it, and at the same time, made the operation of setting the gateway loosely coupled with the object itself (if `$gateway` won't be set, then `$card_gateway` Stripe element will be used).

Another thing worth mentioning is, that the non-UPE checkout should be initialized even when we are dealing with the UPE setup, since it [defines the hook](https://github.com/Automattic/woocommerce-payments/blob/4a4d03c2bf3ea6205b4f21684b097706f0064b1d/includes/class-wc-payments-checkout.php#L76) which is crucial for the legacy card to be rendered properly. This is why [it's initialized](https://github.com/Automattic/woocommerce-payments/blob/7115db427f1ef0f14f9bcdd802dcdf452486e212/includes/class-wc-payments.php#L358) regardless of wether users have UPE turned on or not.

> Note: Please keep in mind, that the legacy card gateway is not working with the Blocks checkout after [this change](https://github.com/Automattic/woocommerce-payments/pull/4670/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1254-R1255). I assume that it was a part of discussion [here](https://github.com/Automattic/woocommerce-payments/pull/4670#pullrequestreview-1174793520) and it will be resolved in [this](https://github.com/Automattic/woocommerce-payments/issues/5111) issue.

#### Testing instructions

1. Ensure that the classic UPE checkout works as expected with both the legacy card and UPE elements.
2. Ensure that the non-UPE checkout works as expected (turn off UPE in the admin panel).
3. Ensure that blocks checkout works for split UPE elements. (please see the note right above the testing instructions).
4. Ensure that [SEPA tokens](https://github.com/Automattic/woocommerce-payments/pull/4670) are managed as expected

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. (was not needed, since the PR will be merged to the feature branch).
- [x] Covered with tests (or have a good reason not to test in description ☝️) - existing tests were updated.

**Post merge**


includes/class-wc-payments-upe-checkout.php
```diff
@@ -0,0 +1,299 @@
...
+
+	/**
+	 * Set gateway
+	 *
+	 * @param string $gateway_id UPE gateway ID.
+	 * @return void
+	 */
+	public function set_gateway( $gateway_id ) {
+		if ( null !== $gateway_id ) {
+			$this->gateway = $this->gateway_map[ $gateway_id ];
```
```suggestion
			$this->gateway = WC_Payments::get_payment_gateway_by_id( $gateway_id );
```
Could be safer to call this function here, yeah?

includes/class-wc-payments-upe-checkout.php
```diff
@@ -0,0 +1,299 @@
...
+	 * @return array
+	 */
+	public function get_payment_fields_js_config() {
+
+		$payment_fields                             = parent::get_payment_fields_js_config();
+		$payment_fields['accountDescriptor']        = $this->gateway->get_account_statement_descriptor();
+		$payment_fields['addPaymentReturnURL']      = wc_get_account_endpoint_url( 'payment-methods' );
+		$payment_fields['gatewayId']                = UPE_Payment_Gateway::GATEWAY_ID;
+		$payment_fields['isCheckout']               = is_checkout();
+		$payment_fields['checkoutTitle']            = $this->gateway->title;
```
```suggestion
```
~I was wondering how this would work, because surely this would simply return the card gateway's title every time, instead of the correct gateway title for each specific gateway~ (Edit: actually my bad, misread, got it now)...but then when searching for the key `checkoutTitle`, I couldn't actually find this used anywhere in the code anymore. 

I'm not sure when this property got disappeared, but I think we can safely remove it, since it's both irrelevant and now inaccurate anyways.

includes/class-wc-payments-upe-checkout.php
FangedParakeet:
```suggestion
			$this->gateway = $this->gateway->wc_payments_get_payment_method_by_id( $gateway_id );
```
There's this one as well, if you'd prefer using the private gateway, instead of the static function; though this function is merely a wrapper around the same static function anyways.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -0,0 +1,299 @@
...
+				?>
+			</div>
+			<?php
+		}
+	}
+
+	/**
+	 * Set gateway
+	 *
+	 * @param string $gateway_id UPE gateway ID.
```
FangedParakeet: This is a pretty minor point, but this parameter isn't a _gateway_ ID--where the gateway IDs are all prefixed with `woocommerce_payments`--but rather a payment method (type) ID.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -0,0 +1,299 @@
...
+			wp_enqueue_style(
+				'wcpay-upe-checkout',
+				plugins_url( 'dist/checkout.css', WCPAY_PLUGIN_FILE ),
+				[],
+				WC_Payments::get_file_version( 'dist/checkout.css' )
+			);
+
+			// Output the form HTML.
+			?>
+			<?php if ( ! empty( $this->gateway->get_description() ) ) : ?>
```
FangedParakeet, line 211: I _think_ `$this->gateway->get_description()` will always be empty, because [we leave this property blank](https://github.com/Automattic/woocommerce-payments/blob/merge-develop-into-poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L96) and I can't see it getting set anywhere else. 🤔

includes/class-wc-payments-upe-checkout.php
```diff
@@ -0,0 +1,299 @@
...
+		WC_Payments_Customer_Service $customer_service,
+		array $gateway_map
+	) {
+		$this->gateway                = $gateway;
+		$this->platform_checkout_util = $platform_checkout_util;
+		$this->account                = $account;
+		$this->customer_service       = $customer_service;
+		$this->gateway_map            = $gateway_map;
+
+		add_action( 'wc_payments_set_gateway', [ $this, 'set_gateway' ] );
```
FangedParakeet, line 72: I was wondering whether we even need to do this at all. The only gateway-specific function that might get called in this class appears to be [this one](https://github.com/Automattic/woocommerce-payments/blob/merge-develop-into-poc/upe-instances-multiplied/includes/class-wc-payments-upe-checkout.php#L123). This is an edgy-casey function that is only called on the Order Pay page.

I'm not sure if there's a better way to handle this, but I suppose we can leave it this way for now and resolve it in another refactor sometime later.

includes/compat/subscriptions/trait-wc-payment-gateway-wcpay-subscriptions.php
```diff
@@ -211,7 +215,7 @@ public function prepare_intent_for_order_pay_page(): bool {
 			return false;
 		}
 
-		$js_config                     = $this->get_payment_fields_js_config();
+		$js_config                     = WC_Payments::get_wc_payments_checkout()->get_payment_fields_js_config();
```
FangedParakeet, line 218: Could something go wrong here? I _think_ it should be fine, because I believe the only gateway-specific function is triggered after a redirection and not on initial Order Pay page load...but probably something to keep an eye on nonetheless.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -59,7 +60,7 @@ class UPE_Payment_Gateway extends WC_Payment_Gateway_WCPay {
 	 *
 	 * @var UPE_Payment_Method
 	 */
-	protected $payment_method;
+	public $payment_method;
```
FangedParakeet: This comment has nothing to do with this line, but do you think it might be useful to move all the AJAX actions in the constructor of this class ([these ones](https://github.com/Automattic/woocommerce-payments/blob/merge-develop-into-poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L111-L145)) to the new `WC_Payments_UPE_Checkout` class instead? This way we can ensure they only fire once, without having to check `has_action` before adding each hook.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -438,6 +439,16 @@ public function parent_process_payment( $order_id ) {
 		return parent::process_payment( $order_id );
 	}
 
+	/**
+	 * Renders the credit card input fields needed to get the user's payment information on the checkout page.
+	 *
+	 * We also add the JavaScript which drives the UI.
+	 */
+	public function payment_fields() {
+		do_action( 'wc_payments_set_gateway', $this->get_selected_stripe_payment_type_id() );
```
FangedParakeet, line 448: Just a note that I found it worth mentioning that `UPE_Payment_Gateway::get_selected_stripe_payment_type_id()` returns a string, whereas [its parent equivalent `WC_Payment_Gateway_WCPay::get_selected_stripe_payment_type_id()`](https://github.com/Automattic/woocommerce-payments/blob/merge-develop-into-poc/upe-instances-multiplied/includes/class-wc-payment-gateway-wcpay.php#L2824) returns an array. 

Ultimately, I think this currently has no ill issues, because the function in the parent is never actually called anywhere (the function is only called inside `UPE_Payment_Gateway` and `WC_Payments_UPE_Checkout`), but there is still a small potential conflict here. Anyways, this could be slightly hazardously ignored for now...

FangedParakeet (approved):
There were about ten trillion files changed here, so I tried my best to focus on the most important ones. 😅 This was truly a rhinocerous-sized refactor, so nice one resolving all these changes as best as you could! 

In general, I think everything is fine here. I've left a few comments, but almost all of them are pretty inconsequential. I did encounter a few new errors, but I'll describe them below and I think almost all can be moved to new issues.

Let me just get into it, by stepping through your testing instructions one-by-one. 

### 1. Ensure that the classic UPE checkout works as expected with both the legacy card and UPE elements.

Using the shortcode checkout, I tested the following with both the non-UPE card gateway and other UPE gateways.
- Standard checkout
- Failed checkout
- Checkout while saving payment method
- Checkout with saved payment method
- Checkout with card requiring 3DS authentication

...And all of the above, appeared to be working normally for me! Huzzah.

### 2. Ensure that the non-UPE checkout works as expected (turn off UPE in the admin panel).

Since the non-UPE card gateway was working properly in the previous tests, I wasn't expecting there to be much difference with the UPE disabled. Nonetheless, without testing as many flows as in the previous step, I did test the basic regular shortcode and blocks flows and discovered that the shortcode checkout appeared to be working as expected (🎉), while the blocks checkout was not working at all, which leads neatly into the next section. 

### 3. Ensure that blocks checkout works for split UPE elements. (please see the note right above the testing instructions).

The legacy non-UPE gateway appeared broken for me. I was able to checkout with a saved payment method, but could not complete a blocks checkout using a new credit card. I don't think this is related to the issues I raised [here](https://github.com/Automattic/woocommerce-payments/pull/4670#pullrequestreview-1174793520), as you might have opined.

Digging a little further I found that the following error was returned to the checkout.

> Error: You passed an empty string for 'payment_method_types[0]'. We assume empty values are an attempt to unset a parameter; however 'payment_method_types[0]' cannot be unset. You should remove 'payment_method_types[0]' from your request or supply a non-empty value.

I believe this error refers to [this parameter](https://github.com/Automattic/woocommerce-payments/blob/merge-develop-into-poc/upe-instances-multiplied/includes/class-wc-payment-gateway-wcpay.php#L1029), which @mdmoore also grappled with while resolving the saved payment method functionality. I'm not entirely sure how this occurred, but checking my error logs, this `payment_method_types` parameter was filled with a single empty string in our request to Stripe.

```log
2022-11-15T01:30:42+00:00 INFO REQUEST POST https://public-api.wordpress.com/wpcom/v2/sites/%s/wcpay/intentions
2022-11-15T01:30:42+00:00 INFO HEADERS: array (
  'Content-Type' => 'application/json; charset=utf-8',
  'User-Agent' => 'WooCommerce Payments/5.0.1',
  'Idempotency-Key' => '3afc322e-7bad-4b22-89b8-1c1c8e33866c',
)
2022-11-15T01:30:42+00:00 INFO BODY: array (
  'test_mode' => true,
  'amount' => 2779,
  'currency' => 'eur',
  'payment_method' => 'pm_1M4Dz12H1rYvbvjIUQiWe4Us',
  'customer' => 'cus_LXS3WAVT5O8VZT',
  // ...
  'description' => 'Online Payment for Order #244 for fangedparakeet.jurassic.tube blog_id 191029001',
  'payment_method_types' => 
  array (
    0 => '',
  ),
  'is_platform_payment_method' => false,
)
```

I'm pretty sure this is a new issue, but I'm happy to resolve this in a separate issue, because it does seem that something uniquely unusual is afoot here.

For the rest, I believe things were working similarly to my experiences noted [here](https://github.com/Automattic/woocommerce-payments/pull/4670#pullrequestreview-1174793520). I was able to complete a checkout with a new UPE payment method and a saved non-UPE payment method, but not a saved UPE payment method. There is already an issue to resolve this limitation, so I'm happy to ignore that here.

### 4. Ensure that SEPA tokens are managed as expected

Everything here seemed to be in proper working order! I wasn't really expecting anything here to go up in flames, but from the shortcode checkout and the My Account pages, managing SEPA tokens seemed to be doing it was supposed to.

#

Those are my thoughts, let me know if you have any immediate comments! I am still approving this PR, because I think the main refactor is fine! The only real issue is using cards in the blocks checkout, but, as stated before, more than happy to resolve that elsewhere.

I think it could be beneficial to also add some more comprehensive confidence checking to checkouts from the Order Pay page, Subscriptions--especially free-trial subscriptions, that rely on setup intents rather than payment intents--and combinations of these two checkout flows, since there are a few uncertainties surrounding these areas that I've raised in separate comments. That being said, I think these examinations can be done another day and I'm eager to catch up with the elusive `develop` branch in the meantime. 

Great work, @timur27! For the most part, LGTM. 👍

timur27:
Thanks for the review, @FangedParakeet! 

> I don't think this is related to the issues I raised https://github.com/Automattic/woocommerce-payments/pull/4670#pullrequestreview-1174793520, as you might have opined.

To be more specific, I was referring to [this paragraph](https://github.com/Automattic/woocommerce-payments/pull/4670#:~:text=Things%20that%20didn%27t%20work,%F0%9F%98%85). 

Before making this PR ready for the review, I was able to make the use case with a new credit card work by rolling back the change for [this line of code](https://github.com/Automattic/woocommerce-payments/pull/4670/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1254-R1255). So, when I provide 

```
$selected_payment_method_type = [ WC_Payments::get_gateway()->get_selected_stripe_payment_type_id() ];
```

instead of 
```
$upe_payment_method = sanitize_text_field( wp_unslash( $_POST['payment_method'] ?? '' ) );
```

To the `payments_api_client->create_and_confirm_intention`, Blocks checkout with a legacy card works like a charm. However, I was skeptical about fixing this here and in this way because I felt this might break tokens for SEPA again. At the same time, I noticed [the issue](https://github.com/Automattic/woocommerce-payments/issues/5111) with the acceptance criteria `While UPE is enabled, completing checkout from blocks checkout with a credit card should process without error.` I thought then that the issue could be fixed as part of it and thus linked it in the description.

includes/class-wc-payments-upe-checkout.php
timur27: Currently, when not setting the gateway and using the card gateway, payment gateways are not rendered on the UI for the classic UPE checkout. I agree that we could resolve it in another refactor to reduce the number of things updated simultaneously, which increases the probability of unintentionally breaking other areas.

includes/class-wc-payments-upe-checkout.php
timur27: Great idea! I also removed the `$gateway_map` from the checkout class. Thanks!

includes/class-wc-payments-upe-checkout.php
timur27: Right, it looks like `checkoutTitle` has been with us for a more extended period. I've removed the field and can confirm that nothing is broken afterward.

includes/class-wc-payments-upe-checkout.php
timur27: I am on the same page as you. The same implementation is also in `poc` branch. Having that said, do you think we can safely remove it? How about the case, if some description will appear in the feature and we'll want to have it displayed?

includes/payment-methods/class-upe-payment-gateway.php
timur27: That'd be great! After a quick try, I can say that it doesn't work. We need to make this work, as the checkout class seems to be the ideal place for these hooks definitions.

includes/class-wc-payments-upe-checkout.php
timur27: Thanks a lot! I updated this.

includes/class-wc-payments-upe-checkout.php
FangedParakeet: Ooh, that's a point. I'm generally loathe to include unused things just in case they might one day be useful, especially since there currently isn't a way to set the description on a per payment method basis. 

That being said, this inclusion is entirely harmless and isn't the only obsolete obstacle observed (for example, the test mode instructions just below are also irrelevant for the UPE, since the card element is currently disabled), so perhaps it's best to do as little possible harm now and save for this for a more appropriate refactor down the line instead.

includes/class-wc-payments-upe-checkout.php
FangedParakeet: > Currently, when not setting the gateway and using the card gateway, payment gateways are not rendered on the UI for the classic UPE checkout.

Cool, let's leave it and not break everything for now then. 😁 

> and using the card gateway

That's weird though. The card gateway shouldn't be using this code at all. Anyways, no matter now: let's just keep things working and moving forward.

includes/payment-methods/class-upe-payment-gateway.php
FangedParakeet: > After a quick try, I can say that it doesn't work. 

Alright, forget I mentioned it then, haha. 😂

FangedParakeet:
> Before making this PR ready for the review, I was able to make the use case with a new credit card work by rolling back the change for [this line of code](https://github.com/Automattic/woocommerce-payments/pull/4670/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1254-R1255). So, when I provide
>
> $selected_payment_method_type = [ WC_Payments::get_gateway()->get_selected_stripe_payment_type_id() ];

You would first have to change the return value of `WC_Payment_Gateway_WCPay::get_selected_stripe_payment_type_id()` to a single string (`'card'`), but I _think_ this could work. I actually think this would work in all situations, but there are definitely a few more edge cases we'd need to test out first, so--as you offered--I think it's best to resolve this in a new issue.

Also, I just noticed one more thing! (Sorry.)

In this merge branch, for some reason, the WCPay gateway seems to have disappeared from the main WC Payments Settings screen (**WooCommerce > Settings > Payments > All payment methods**).

WCPay missing from WC Settings
_O brother where art thou?_

It's still there on both the `develop` branch and the `poc/upe-instances-multiplied` branch.

Normal WC Settings screen
_Regular correct settings page_

If you think there's a quick fix for this, let me know: otherwise I'll create a new issue to resolve this missing method, once you've completed this merge.

Also, we had some discussion above about the gateway description field: I just noticed that these description values are actually incorrect in our PoC branch (see above screenshot). I'll create a new issue to fix this as well shortly...

timur27:
Thanks for another round of review, @FangedParakeet ! I fixed the failing test (https://github.com/Automattic/woocommerce-payments/pull/5082/commits/a4222e7e54a893d5a88320765a5700a910f79c61) and the hidden card gateway in the admin panel (https://github.com/Automattic/woocommerce-payments/pull/5082/commits/1bebfee63a32f68790909bcaa159ed6b1f562b78).

FangedParakeet (approved):
LGTM! I think pretty much everything pending from here has an open issue already ready to capture further remaining work--if not, please feel free to pop any open, if I missed anything! 

Otherwise let's get this merged and finally get our branch up to date. :shipit:

Issue #4951, opened 2022-10-20
WCPay Platform: Meta Data


RadoslavGeorgiev:
### Description

Prevent accessing meta data directly: paJDYF-50y-p2#woopay-meta-data.

Within this issue the focus is on `_intent_id`and `_payment_method_id`.

### Acceptance criteria

1. There is a new `Order` helper in the `WCPay\Core` namespace.
2. The helper has methods to access and change an order's intent and payment method IDs.
3. The gateway class's `attach_intent_info_to_order` functionality lives within the new class.
4. All instances where the gateway access those order meta values use the getters and setters from the new class.
5. The new class triggers an action whenever those values change, in order to let consumers consume them.

### Additional context

See the epic for links to the project thread and PRD.

kalessil:
@Automattic/gamma , would it be possible to start documenting the order metadata under https://github.com/Automattic/woocommerce-payments-server/blob/trunk/docs/Metadata.md (as you are fresh into context)?

RadoslavGeorgiev:
Hey Vlad, @jessy-p's working on documenting this issue, however it's just related to (post) meta data on the client, not intent meta data. I don't think we've touched that topic.

(issue was closed because of me moving it to the closed column in ZenHub, I did not know that automatically closes it)

kalessil:
I meant to add a new chapter for Post metadata specific to WCPay, this issue is clearly not related to intents.

RadoslavGeorgiev:
Our plan is to add a readme to the client, do you think this server file could be more suitable?

kalessil:
I don't have a strong opinion on this, though if you make the docs public I'd really appreciate adding a link in the doc I referenced originally.

PR #5094, opened 2022-11-10
Provide Services used by consumers through the WCPay\Core namespace

### LINKED ISSUE #4953: WCPay Platform: Use services
### Description

At one point all services, which are used by consumers, should live in or be accessible through the `WCPay\Core` namespace. For this issue, we should focus on `WC_Payments_Customer_Service`, which is the only used service at the moment.

💡 We cannot guarantee backwards compatibility this way. Instead we should include shortcuts to those services and/or specific methods within the `WCPay\Core` namespace. Doing so will ensure that all methods and their arguments remain backwards-compatible. At first those calls can be absolutely identical to the ones to existing services, serving as a facade.

### Acceptance criteria

1. There is a `WCPay\Core\Customer_Service` class, which contains all needed methods for WooPay.
2. The class has proper test coverage.
3. Any other services, which are used in a similar way, should be also accessible there. Currently only the customer service seems to be used, but once this issue is closed and the outcome observable, we should create a phase 2 issue for the other services.

### Additional context

jessy-p:
Fixes #4953

- Added a class `includes/core/class-wc-payments-customer-service-api.php`
- Initialised this class in `includes/class-wc-payments.php`
- The class can be accessed using `WC_Payments::get_customer_service_api()`
- At present this class offers all the same methods as `class-wc-payments-customer-service.php` 
- The intention is the external consumers e.g. WooPay should use this integration point provided in the `WCPay\Core` namespace, and should not use any methods from other classes, which will be considered internal to WCPay
- This provides the ability for WCPay to refactor internally while maintaining backward compatibility on this integration point.
- In future PRs, additional services will also be added to `WCPay\Core `namespace.
- The tests provided in `tests/unit/core/test-class-wc-payments-customer-service-api.php` are integration tests for the  system, mocking only `WC_Payments_Http`. 

#### Testing instructions
- Tests pass
- Able to write code that uses `WC_Payments_Customer_Service_API` in place of `WC_Payments_Customer_Service`
-------------------

- [x] Does not apply - Tested on mobile (or does not apply)

**Post merge**

- [x] Not applicable - Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [x] Not applicable - Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [x] Not applicable - Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.


tests/unit/core/test-class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,142 @@
...
+	/**
+	 * Pre-test setup
+	 */
+	public function set_up() {
+		$this->mock_http_client = $this
+			->getMockBuilder( 'WC_Payments_Http' )
+			->disableOriginalConstructor()
+			->setMethods( [ 'get_blog_id', 'is_connected', 'remote_request' ] )
+			->getMock();
+		add_filter(
```
RadoslavGeorgiev: Let's not forget to remove this filter in `tear_down()`.

tests/unit/core/test-class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,142 @@
...
+			'wc_payments_http',
+			function() {
+				return $this->mock_http_client;
+			}
+		);
+		$this->customer_service     = new WC_Payments_Customer_Service( WC_Payments::create_api_client(), WC_Payments::get_account_service(), WC_Payments::get_database_cache() );
```
RadoslavGeorgiev: I didn't realize `create_api_client` is public, and available to use this way for mocks, without replacing the main API client. I'll keep this approach in mind for future work. Thank you!

RadoslavGeorgiev (commented):
I like this testing approach. Let's apply it to the other methods, I think it will do the job just fine 👌

tests/unit/core/test-class-wc-payments-customer-service-api.php
jessy-p: Done

includes/core/class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,183 @@
...
+	/**
+	 * Given a WC_Order or WC_Customer, returns an array representing a Stripe customer object.
+	 * At least one parameter has to not be null.
+	 *
+	 * @param WC_Order    $wc_order    The Woo order to parse.
+	 * @param WC_Customer $wc_customer The Woo customer to parse.
+	 *
+	 * @return array Customer data.
+	 */
+	public static function map_customer_data( WC_Order $wc_order = null, WC_Customer $wc_customer = null ): array {
```
zmaglica: I really like type-hinting, but the mos of the codebase doesn't have it. Shall we maybe consider removing function type-hinting?

includes/core/class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,183 @@
+<?php
```
zmaglica: Shall we put this file or service into a dedicated folder? I would like to use this approach to prevent putting plenty of classes into root core directory. I would put it inside service folder? What do you think about this approach?

includes/core/class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,183 @@
...
+use WCPay\Database_Cache;
+use WCPay\Exceptions\API_Exception;
+use WCPay\Logger;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class handling any customer functionality
+ */
+class WC_Payments_Customer_Service_API {
```
zmaglica: Shall we put this class into Core dedicated namespace?

includes/core/class-wc-payments-customer-service-api.php
```diff
@@ -0,0 +1,183 @@
+<?php
+/**
+ * Class file for WCPay\Core\WC_Payments_Customer_Service_API.
+ *
+ * @package WooCommerce Payments
+ */
+
+use WCPay\Database_Cache;
+use WCPay\Exceptions\API_Exception;
+use WCPay\Logger;
```
zmaglica: I am not sure is this Logger class used somewhere. Do we need to remove it?

zmaglica (commented):
Left some comments.

includes/core/class-wc-payments-customer-service-api.php
jessy-p: Removed

includes/core/class-wc-payments-customer-service-api.php
jessy-p: Fixed

includes/core/class-wc-payments-customer-service-api.php
jessy-p: Its present in `Customer_Service` class so I think it would be better to retain it here.

includes/core/class-wc-payments-customer-service-api.php
jessy-p: We can do that, but didn't want to over-optimize too soon as I think we are currently planning for only two classes used by external consumers - Account_Service and Customer_Service. I think we can move it later, as the need arises?

includes/core/class-wc-payments-customer-service-api.php
zmaglica: Oh, that's nice. We can leave it. I was afraid that you could get other variation of order (overridden one we have in our  codebase), but if it's already there, we can leave it.

includes/core/class-wc-payments-customer-service-api.php
zmaglica: The way we structure now, someone will follow it and use it in the future. We could remove it in the future to a specific folder like you mentioned, but I would like to keep it organized and defined from the start. I would put it in the specific folder, but as you suggested, we can do that later on. If you would like to keep it like this, I am perfectly fine, and if you want to put it in a specific folder, I would also be perfectly fine (finer than in the first case 😄 )

jessy-p: Moved class into services folder within core as I started working on #4951

PR #5098, opened 2022-11-10
Check if user can `manage_woocommerce` in maybe_redirect_to_onboarding()

### LINKED ISSUE #5096: Non-admin users cannot access the dashboard
### Describe the bug

Users without the `manage_woocommerce` capability are facing fatal errors when trying to access WP Admin:

```
Fatal error: Uncaught Error: Class 'WC_Payments_Admin_Settings' not found in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-account.php on line 572

Error: Class 'WC_Payments_Admin_Settings' not found in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-account.php on line 572

Call Stack:
    0.0001     402408   1. {main}() /var/www/html/wp-admin/index.php:0
    0.0001     403656   2. require_once('/var/www/html/wp-admin/admin.php') /var/www/html/wp-admin/index.php:10
    1.1864   16878384   3. do_action() /var/www/html/wp-admin/admin.php:175
    1.1864   16878760   4. WP_Hook->do_action() /var/www/html/wp-includes/plugin.php:517
    1.1864   16878760   5. WP_Hook->apply_filters() /var/www/html/wp-includes/class-wp-hook.php:332
    1.2094   17123008   6. WC_Payments_Account->maybe_redirect_to_onboarding() /var/www/html/wp-includes/class-wp-hook.php:308
```

This is caused by this line:

https://github.com/Automattic/woocommerce-payments/blob/055775044ca09aaa3e1cf85f44b183c33dc75535/includes/class-wc-payments-account.php#L572

it tries to access `WC_Payments_Admin_Settings`, but the class is not present, because of a condition, which only includes it whenever the user can `manage_woocommerce`:

https://github.com/Automattic/woocommerce-payments/blob/055775044ca09aaa3e1cf85f44b183c33dc75535/includes/class-wc-payments.php#L392-L397

### To Reproduce

1. Create a new user without the `manage_woocommerce` capability. This includes all default non-admin roles.
2. Log in with that user's credentials.
3. Navigate to any admin page, including just `/wp-admin`.

### Actual behavior

The error described above is thrown.

### Expected behavior

`wp-admin` should load without issues.

shendy-a8c:
Fixe #5096

As explained in https://github.com/Automattic/woocommerce-payments/issues/5096, the fatal error is because `WC_Payments_Account::maybe_redirect_to_onboarding()` calls `WC_Payments_Admin_Settings` but `WC_Payments_Admin_Settings` is only defined if page is admin and user has `manage_woocommerce` capability. The fatal error should be experienced only by users who accesses admin pages but they don't have `manage_woocommerce` capability.

1. Always load `WC_Payments_Admin` (`includes/admin/class-wc-payments-admin-settings.php`) and `WC_Payments_Admin_Settings` (`includes/admin/class-wc-payments-admin-settings.php`) if page is admin. There is no risk of doing that because there is only class definition and no side-effect from including those files. I don't see any logic in the code base that checks for those class existence too, so I don't expect any side-effect just by having those classes defined.

2. Check if user can `manage_woocommerce` in [maybe_redirect_to_onboarding()](https://github.com/Automattic/woocommerce-payments/blob/ab029f5facb1c48495df4ac00a538bd2a35b4b83/includes/class-wc-payments-account.php#L567-L572). I also don't see the risk of adding this check because only admins with `manage_woocommerce` capability that should be able to onboard. 

3. Updating unit tests that calls `maybe_redirect_to_onboarding()` because in those unit tests, the test user does not have `manage_woocommerce` capability. My change sets it so test user has `manage_woocommerce` capability to make those unit tests pass.

4. Add a new unit test to make sure `maybe_redirect_to_onboarding()` does not do anything if user does not have `manage_woocommerce` capabiltiy.

#### Testing instructions

1. Create a new user without the `manage_woocommerce` capability. This includes all default non-admin roles, for example: `Editor` role.
5. Log in with that user's credentials.
6. Navigate to any admin page, including just /wp-admin.
7. Admin pages should load just fine without fatal error. With this base branch, you will see a fatal error.
```
Fatal error: Uncaught Error: Class 'WC_Payments_Admin_Settings' not found in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-account.php on line 572
```

-------------------

**Post merge**

### DIFF
--- a/tests/unit/test-class-wc-payments-account.php
+++ b/tests/unit/test-class-wc-payments-account.php
@@ -73,11 +73,27 @@ class WC_Payments_Account_Test extends WCPAY_UnitTestCase {
 		parent::tear_down();
 	}
 
+	/**
+	 * @param bool $can_manage_woocommerce
+	 *
+	 * @return Closure
+	 */
+	private function create_can_manage_woocommerce_cap_override( bool $can_manage_woocommerce ) {
+		return function ( $allcaps ) use ( $can_manage_woocommerce ) {
+			$allcaps['manage_woocommerce'] = $can_manage_woocommerce;
+
+			return $allcaps;
+		};
+	}
+
 	public function test_maybe_redirect_to_onboarding_stripe_disconnected_redirects() {
 		// Simulate the situation where the redirect has not happened yet.
 		update_option( 'wcpay_should_redirect_to_onboarding', true );
 
 		$this->mock_empty_cache();
+
+		$cb = $this->create_can_manage_woocommerce_cap_override( true );
+		add_filter( 'user_has_cap', $cb );
 
 		$this->mock_api_client->expects( $this->once() )->method( 'get_account_data' )->will(
 			$this->throwException( new API_Exception( 'test', 'wcpay_account_not_found', 401 ) )
@@ -94,6 +110,9 @@ class WC_Payments_Account_Test extends WCPAY_UnitTestCase {
 		update_option( 'wcpay_should_redirect_to_onboarding', true );
 
 		$this->mock_empty_cache();
+
+		$cb = $this->create_can_manage_woocommerce_cap_override( true );
+		add_filter( 'user_has_cap', $cb );
 
 		$this->mock_api_client->expects( $this->once() )->method( 'get_account_data' )->will(
 			$this->throwException(
@@ -117,6 +136,9 @@ class WC_Payments_Account_Test extends WCPAY_UnitTestCase {
 
 		$this->mock_empty_cache();
 
+		$cb = $this->create_can_manage_woocommerce_cap_override( true );
+		add_filter( 'user_has_cap', $cb );
+
 		$this->mock_api_client->expects( $this->once() )->method( 'get_account_data' )->will(
 			$this->throwException( new Exception() )
 		);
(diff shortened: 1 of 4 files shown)


leonardola (approved):
This looks and tests good. I am not a big fan of using abbreviated variable names. But that does not make a difference in the behavior so you can skip it in favor of deploying a fix faster.

tests/unit/test-class-wc-payments-account.php
```diff
@@ -73,12 +73,28 @@ public function tear_down() {
...
+		};
+	}
+
 	public function test_maybe_redirect_to_onboarding_stripe_disconnected_redirects() {
 		// Simulate the situation where the redirect has not happened yet.
 		update_option( 'wcpay_should_redirect_to_onboarding', true );
 
 		$this->mock_empty_cache();
 
+		$cb = $this->create_can_manage_woocommerce_cap_override( true );
```
shendy-a8c, line 95: Are you referring to this `$cb`? I copy pasted from `tests/unit/admin/test-class-wc-rest-payments-settings-controller.php`. However, that is a good call that it's not a good practice to use abbreviation as variable name. Also I agree that it can be fixed later so this can be shipped sooner.

If you think it's worth fixing, please create a follow up issue / PR.

PR #5123, opened 2022-11-14
Release 5.0.2.

shendy-a8c:
Release 5.0.2.


leonardola (dismissed):
don't we need to change the versions in the packages.json and woocommerce-payments.php files ?

leonardola (approved):
Apparently we don't need to manually update versions in the packages.json and woocommerce-payments.php files anymore

PR #5124, opened 2022-11-14
WCPay as a Platform: Add base request classes

### LINKED ISSUE #4954: WCPay Platform: API request classes

RadoslavGeorgiev:
Alternative approach/Fixes #4954 and #5056.

Adding a base set of request (and a very basic response as a placeholder) classes.

> **This PR includes examples!**
> Check out [includes/core/server/class-temp-request-examples.php](https://github.com/Automattic/woocommerce-payments/pull/5124/files#diff-c94d03fbf684a3218266ae652189248b0b7b032f9cde0643e0b0a5d1de3b918f) to see them.
> The front-end will show the output. Use http://localhost:8082/?show_basic_examples to see it.

There is a new `WCPay\Core\Server\Request` class:
- It is abstract, but can be:
    - Used through `WCPay\Core\Server\Request\Generic`, allowing for existing requests to be wrapped and moved to the `send_request` method of the API client. This class does not have any validation, or additional structure right now.
    - Extended by specific requests, ex. `WCPay\Core\Server\Request\Create_Intent`.
- Uses an internal `$props` array, to privately store all needed values.
- Has abstract/default methods for APIs, HTTP methods, and generic request flags (is site specific, etc.)
- Contains the necessary logic in `get_params()` to ensure that all required parameters are set. Those can be defined as a `IMMUTABLE_PARAMS` constant in child classes, and the constant is retrieved from each individual class to ensure that child classes are not overwriting protected values.
- A `format_response` method, which in the future can be overwritten to return specific responses in child classes.
- A protected `set_param` method, which can be used by inheritants to modify the internal `$params` array.
- Introduces protected mode, which makes it impossible to alter existing protected parameters. This is useful when the class is process through hooks.
- A static `extend` method, which can be used through extended classes to create a new instance, which already contains the same values as the parent object. It is only available when using the `apply_filters` method (see below).

Examples of how the class can be extended can be found in the `Create_Intent` and `WooPay_Create_intent` classes.

**Here is the important part:**

Some details about the `apply_filters` and `extend` methods. `extend` can only be called through `apply_filters`, here is how it cal be called:

```php
<?php
// Imagine you are somewhere in the gateway's `process_payment` method:
$request = WC_Payments::create_request( WCPay\Core\Server\Request\Create_Intent::class )
    ->set_amount( 100 )
    ->set_currency( 'eur' );

// Once everything is prepared, allow modifications. This will also check if all required fields are present. Note that just like the normal `apply_filters`, the method supports an arbitrary number of parameters:
$some_additional_field = 'xyz';
$response = $request->send( 'wcpay_create_intent_request', $some_additional_field ); 
```
 
Once the request is created, this allows it to be safely modified and extended. Here is an example:

```php
add_filter( 'wcpay_create_intent_request', 'woopay_modify_create_intent', 10, 2 );
function woopay_modify_create_intent( WCPay\Core\Server\Request\Create_Intent $base_request, $some_additional_field ): WCPay\Core\Server\Request\WooPay_Create_Intent {
    echo $some_additional_field; // Don't do that, the point is it's available.

    $request = WCPay\Core\Server\Request\WooPay_Create_Intent::extend( $base_request );
    $request->set_save_payment_method_to_platform( true );
    $request->set_amount( 300 ); // This will throw an exception, as the amount is protected.
    return $request;
}
```

Required implementation for request-specific classes includes (except the class skeleton):
1. An `IMMUTABLE_PARAMS` constant, if required.
2. A `REQUIRED_PARAMS` constant for defining required fields.
3. One setter for each prop.

A nice side-effect of the way requests can be overwritten by consumers (example above) is that we do not need a specific method for each request within the API client class, as there would be no possibility to overwrite the type of request before it reaches the `send_request` method.

### Closed Tasks:

- [x] [Make sure all parameters are set](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/c9e14b9d588bd28f540766c38a48ac2e2ab8e16c).
- [x] ~Should any parameters be overwritable?~ Mutable properties can just be overwritten.
- [x] Make sure the generic class can be easily used a replacement of what's currently there: Tried it within a couple of basic requests, it works, but is not commited.
- [x] ~Check if the generic class can be extended similarly to specific classes.~ It does not need to be extended, all params can be changed already.
- [x] [Decide how to instantiate requests.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] [Handle boolean variables properly](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/615dfec60c3afddae3eb177c8abfed10e78dd603).
- [x] [Handle level3 data in the right place](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/a977affe7a9733a45f3f3a476611a70de78f22dd).
- [x] [Update the docs of `apply_filters`.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/91d10f3939d6e6f1a867f84572ab3fccf27b1997)
- [x] [Restore the unique keys.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] [Add proper dependencies](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] Revisit the `Intention` trait: The code is pretty clear, and does not seem to require changes.
- [x] [Rethink the create_and_confirm intention calls while adding validation and helpers](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/ef10ff85f66961312a7c1ba7bdfd93dc46dd7668?diff=split&w=0).
- [x] [Require a key within the `send_request` method for the filter to avoid doing it manually](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/e2933240c2c1b0b62deb39c124e3b8260c04b41d).
- [x] [Remove dependencies on the HTTP client, and make `WC_payments::get_wc_payments_http()` private again.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)

### Remaining tasks

Here is a list of open tasks for the PR:

#### Responses

Currently theis is a basic [`WCPay\Core\Server\Response` class](https://github.com/Automattic/woocommerce-payments/blob/spike/rado-request-classes/includes/core/server/class-response.php), which simply wraps a successfull response.

**This PR should not focus on responses (it's way too huge already)**, but while creating a base to rewrite request methods, we should make sure that we also have a solid base to build upon for responses later.

Main point: Currently we're still using a lot of functionality from the API client class, but that seems to have attracted a lot of quick & kind-of-dirty solutions like response formatting (examine charges and intentions). This functionality includes error handling (converting errors to exceptions), which should remain in the API client class until we remove all request methods from the class, but we need to make sure that it would be easy for us to move it into the `Response` class later.

This is not something we want to do yet, but currently intents and charges are being "deserialized" (turned into DTOs) within the API client class. Consider what dependencies/actions/filters will be required in order to extract this functionality into custom `format_response` request methods.

In the end, we should have the right infrastructure in place in the PR, then create another PR/issue to implement everything described here.

#### Extract WooPay code to the right place

There is already a working [`WCPay\Core\Server\Request\WooPay_Create_And_Confirm_Intention` request class](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/6e1bf4b23ca55cd7bc342b39467b1bc6577ba85d) in this PR, but the WooPay code, which prepares the parameters for the request [still lives within the gateway class](https://github.com/Automattic/woocommerce-payments/blob/spike/rado-request-classes/includes/class-wc-payment-gateway-wcpay.php#L989-L990).

We need to figure out where to place that code, and how to link it. This could be a follow-up PR, based on this branch as well.

#### Use the right exceptions

Right now the PR is using a generic `\Exception` everywhere, this should be replaced with appropriate exception classes, and translatable messages.

#### Use ENUMs wherever possible

 I'm not sure where else/how we could use enums here:

 - It's hard to validate APIs, as we're receiving the raw API, which we cannot correlate back to a constant.
 - Request methods are validated already, and using ENUM.

Let's take a final look to make sure there are no missed opportunities.

#### Add unit tests and docs

Biggest task so far, yet pretty self-explanatory.

## Testing instructions

Two options right now:

- Check examples
- Perform a standard payment to see the `Create_And_Confirm_Intention` and `WooPay_Create_And_Confirm_Intention` requests in action.

-------------------

### DIFF
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -10,6 +10,60 @@ namespace WCPay\Core\Server;
 /**
  * Base for requests to the WCPay server.
  */
-class Request {
+abstract class Request {
+	/**
+	 * Holds the parameters of the request.
+	 *
+	 * @var mixed[]
+	 */
+	protected $params = [];
 
+	/**
+	 * Prevents the class from being constructed directly.
+	 */
+	protected function __construct() {
+		// Nothing to do here yet.
+	}
+
+	/**
+	 * Returns the needed API.
+	 *
+	 * @return string Check WCPay\Core\Server\APIs.
+	 */
+	abstract public function get_api(): string;
+
+	/**
+	 * Returns the method of the request.
+	 *
+	 * @return string See the constants in WordPress's `Requests` class.
+	 */
+	abstract public function get_method(): string;
+
+	/**
+	 * This is a legacy method, and is the same throughout the codebase.
+	 * Might be worth removing while refactoring to use the Core\Server API.
+	 *
+	 * @return bool
+	 */
+	public function is_site_specific(): bool {
+		return false;
+	}
+
+	/**
+	 * If true, the request will be signed with the user token rather than blog token. Defaults to false.
+	 *
+	 * @return bool
+	 */
+	public function use_user_token(): bool {
+		return false;
+	}
+
+	/**
+	 * Indicates if the raw response should be returned.
+	 *
+	 * @return bool
+	 */
+	public function should_return_raw_response(): bool {
+		return false;
+	}
 }
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -42,6 +42,15 @@ abstract class Request {
 	private $protected_mode = false;
 
 	/**
+	 * Creates a new instance of the class.
+	 *
+	 * @return static
+	 */
+	public static function create() {
+		return new static();
+	}
+
+	/**
 	 * Prevents the class from being constructed directly.
 	 */
 	protected function __construct() {
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -105,7 +105,7 @@ abstract class Request {
 	 * @return array
 	 * @throws \Exception If the request has not been initialized yet.
 	 */
-	public function get_params() {
+	final public function get_params() {
 		$missing_params = [];
 		foreach ( $this->get_required_params() as $name ) {
 			if ( ! isset( $this->params[ $name ] ) ) {
@@ -145,7 +145,7 @@ abstract class Request {
 	 * @param string $key   The name of the parameter.
 	 * @param mixed  $value And the value to set.
 	 */
-	protected function set_param( string $key, $value ) {
+	final protected function set_param( string $key, $value ) {
 		if ( $this->protected_mode && in_array( $key, $this->get_immutable_params(), true ) ) {
 			$this->throw_immutable_exception( $key );
 		}
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -23,6 +23,20 @@ abstract class Request {
 	const IMMUTABLE_PARAMS = [];
 
 	/**
+	 * Indicates which parameters are required (keys only).
+	 *
+	 * @var string[]
+	 */
+	const REQUIRED_PARAMS = [];
+
+	/**
+	 * Contains default values for parameters, which are not set automatically.
+	 *
+	 * @var string[]
+	 */
+	const DEFAULT_PARAMS = [];
+
+	/**
 	 * Holds the parameters of the request.
 	 *
 	 * @var mixed[]
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -145,7 +145,7 @@ abstract class Request {
 	 * @throws \Exception     In case a class does not exists, or immutable properties are modified.
 	 * @todo                  Add proper exceptions here.
 	 */
-	final public function announce( $hook, ...$args ) {
+	final public function apply_filters( $hook, ...$args ) {
 		$cloned = clone $this; // Work with a clone to avoid mutations of parameters.
 
 		/**
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -320,7 +320,7 @@ abstract class Request {
 	 * @return string[] The names of those params.
 	 */
 	private function get_immutable_params() {
-		return $this->traverse_class_constants( 'IMMUTABLE_PARAMS' );
+		return $this->traverse_class_constants( 'IMMUTABLE_PARAMS', true );
 	}
 
 	/**
@@ -329,7 +329,7 @@ abstract class Request {
 	 * @return string[] The names of those params.
 	 */
 	private function get_required_params() {
-		return $this->traverse_class_constants( 'REQUIRED_PARAMS' );
+		return $this->traverse_class_constants( 'REQUIRED_PARAMS', true );
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


includes/core/server/class-request.php
```diff
@@ -0,0 +1,306 @@
...
+	 * If true, the request will be signed with the user token rather than blog token. Defaults to false.
+	 *
+	 * @return bool
+	 */
+	public function should_use_user_token(): bool {
+		return false;
+	}
+
+	/**
+	 * Indicates if the raw response should be returned.
```
RadoslavGeorgiev, line 120: This is only used for files, and once/if we move response processing out of the API client, this can be removed. That's why it's not available in the generic request class.

zmaglica (commented):
Generally, I like your approach. Using filters feels like a WordPress way of manipulating certain parts of flow.

The only thing that looks a bit of worry about me is this immutability and parent chain of immutable contsants. The mutable properties could be protected by setters in a way that a property could be set only if it not setted before. It would be really hard to see what class is mutable if you chain them (extend) multiple times.

Something like this will do a trick:

```
private string $a;

public function set_a(string $a) {
      if (is_string($a))
          throw new \Exception ('A already set');
          ...
}
```

The other thing that puzzles me is protected mode. The only time when it is set to true is when you are applying filters. I am not sure do we need it there since nothing will change class properties since the function is still executing and PHP by default will wait for the function to finish. 
The reason why I avoided filters is that I wanted to have well constructed classes for every request and make them in a way where you built it using builder pattern. Your approach is great one, more like combination of OOP way and usage of WordPress built-in functions.

So, my final conclusion is that I would use a different approach regarding the immutability, since it looks a bit complex where you loop over extended classes and looking for a immutably constant. The rest of the approach is good. Me personally, I wouldn't use filters. I would create a dedicated request classes for each API request. The approach is not too much different from the one that I made, so I am perfectly good with it. The only part that I don't like is, like I said, the immutability approach. The extend function looks really nice because it allows us to create custom scenarios for each request where we could use filters to achieve that. The same approach could be used to create a custom function in the Request class. Something like `create_wc_pay_charge_request()` inside Create_Charge and cover everything there.

RadoslavGeorgiev:
Closing in favor of the non-spike https://github.com/Automattic/woocommerce-payments/pull/5136

PR #5135, opened 2022-11-15
Prevent the double-zip issue when downloading from zip file workflow

jbordonado:

Downloading the artifact from the [build-zip-file workflow](https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/build-zip-file.yml) results in a double-zip, which prevents us from easily using that downloaded zip in a test site (error: `The package could not be installed. No valid plugins were found. Plugin installation failed`).

Why?
- We upload the built zip using the `actions/upload-artifact@v3` action
- When clicking on the GitHub UI to download the artifact, a zip file is automatically generated with the content that was uploaded

This is explained here: https://github.com/actions/upload-artifact#zipped-artifact-downloads

Unzipping the built zip to a temporary folder called `zipfile` and uploading that folder instead solves this issue.
The only downside with that approach is that we will see a large size for the artifact (like 12MB), which can scare someone that is used to seeing something around 5MB for the zip file. This is also why I added a comment in the summary to explain this.

See the latest run to confirm you can upload directly what you are downloading from the artifacts: https://github.com/Automattic/woocommerce-payments/actions/runs/3470087644

-------------------

**Post merge**

### DIFF
--- a/.github/workflows/build-zip-file.yml
+++ b/.github/workflows/build-zip-file.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: "Unzip the file (prevents double zip issue)"
         run: |
-          echo ":information_source: Unzipping the zip file to a folder. This is to prevent a double-zip issue ([known limitation](https://github.com/actions/upload-artifact#zipped-artifact-downloads)). The increase of the artefact's size is expected since we're uploading the folder instead of a zip." >> $GITHUB_STEP_SUMMARY
+          echo ":information_source: Unzipping to a folder to prevent double-zip issue ([known limitation](https://github.com/actions/upload-artifact#zipped-artifact-downloads)). Increase in artifact size is expected since we're uploading the folder." >> $GITHUB_STEP_SUMMARY
           unzip ${{ steps.build_plugin.outputs.release-filename }} -d zipfile
       
       - name: "Upload the zip file as an artifact"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-zip-file.yml
```diff
@@ -18,12 +18,17 @@ jobs:
       - name: "Build the plugin"
         id: build_plugin
         uses: ./.github/actions/build
+
+      - name: "Unzip the file (prevents double zip issue)"
+        run: |
+          echo ":information_source: Unzipping the zip file to a folder. This is to prevent a double-zip issue ([known limitation](https://github.com/actions/upload-artifact#zipped-artifact-downloads)). The increase of the artefact's size is expected since we're uploading the folder instead of a zip." >> $GITHUB_STEP_SUMMARY
```
achyuthajoy: Can we re-write `Unzipping the zip file to a folder. This is to prevent a double-zip issue` as `Unzipping to a folder to prevent double-zip issue`. Also `The increase of the artefact's size is expected` as `Increase in artifact size is expected`

Also fix the spelling - `artifact` instead of `artefact`

deepakpathania (approved):
Works as expected and changes LGTM. ✅

achyuthajoy (changes requested):
@jbordonado Left a minor comment. :)

.github/workflows/build-zip-file.yml
jbordonado: Fixed

achyuthajoy (approved):
LGTM 🚢

PR #5136, opened 2022-11-15
WCPay as a Platform: Add base request classes

### LINKED ISSUE #4954: WCPay Platform: API request classes

zmaglica:
Alternative approach/Fixes #4954 and #5056.

### WooCommerce Payments Server Request class

The current way to execute WooCommerce Payments Server API class is performed by using [WC_Payments_API_Client](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/wc-payment-api/class-wc-payments-api-client.php) class. The class functionality and scope is well-defined, but during the development, a lot of code have been added on the fly to accommodate need of extended functionality that was generated over the time. The concept of this class is to get the data, built a request with that data and send to WooCommerce Payments Server. At the moment of writing, this class does a considerable amount of other things that "should" not be there. Adding more features to this class makes things hared to refactor and maintain. We were piling up so much extra code there to accommodate our needs, and sometimes we didn't consider factors that are addressed in a couple of sentences before. Making requests were harder because single function all API request was doing so much outside "work".

This PR address this issue by creating a request classes using combination of setters, filters, DTO-ish approach and validation. Sending the request and building it is faster because heavy lifting was already implemented with the request classes. Need to build request? With one line of code (`$request = WC_Payments::create_request( WCPay\Core\Server\Request\Create_Intent::class )`) you will start a process of building it. Setting parameters are performed by setters where each setter is valided if needed. Current validators validates various data like  stripe ids (customer id, payment metod id, ...), amount, ... and to execute request, the final function checks have you entered all required parameters  with their own validation. 

### Code examples:

Let's create an instance of request class.

`$request = Create_And_Confirm_Intention::create();
`
Now we will request with the data

```
$request->set_amount( 500 );
$request->set_currency_code( 'usd' );
$request->set_payment_method( $payment_information->get_payment_method() );
$request->set_customer( 'cus_1' );
$request->set_capture_method( $payment_information->is_using_manual_capture() );
$request->set_metadata( $metadata );
$request->set_level3( $this->get_level3_data_from_order( $order ) );
$request->set_off_session( $payment_information->is_merchant_initiated() );
if ( $save_payment_method_to_store ) {
         $request->setup_future_usage();
}
```

If you send invalid data, the validation exception will be thrown.

Here are the examples:

```
$request->set_amount( 'foo' ); // Invalid_Request_Parameter_Exception with message will be thrown if value is below 0.
$request->set_currency_code( 'foo' ); // Invalid_Request_Parameter_Exception with message will be thrown if currency is not supported.
$request->set_payment_method( 'payment_id' ) // Invalid_Request_Parameter_Exception will be thrown that displays that stripe id is incorrect.
$request->set_customer( '1' ); // Same as previous method. Stripe id will be validated, and an exception will be thrown.
```
Forgot to add required metadata values? No problem, the request validates metadata values for specific request. 
Adding a base set of request (and a very basic response as a placeholder) classes.

### Building your own request class

Want to build your own request? With added classes, adding or creating a new request is easy.
To accomplish that, create a new class inside core/server/request folder (any folder will do, but we would like to keep all requests within this folder) and extend [Core Server Abstract class](https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/core/server/class-request.php).
After that, make sure to fill [required params constant (if needed)](https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/core/server/class-request.php#L35) and [immutable param constant](https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/core/server/class-request.php#L28). These constants are optional, and it used to validate immutability and requirement of some parameters. The immutability is added to prevent changing some parameters when someone wants to extend the request. The good example is amount or customer_id. These parameters need to be immutable, and they only need to be changed only when they are setted for the first time.

### Extending request classed that are already added

Sometimes you wish to add some functionality or alter some functionality of specific request class. Changing or refactoring an existing class or API request could potently break some flows and sometimes it is hard to add new things inside API request calls. To overcome this issue, the [send](https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/core/server/class-request.php#L203) have been added that accepts filer name as a first argument and second argument is a list of arguments for that specific filter. The approach here is to build a filter function that will create a brand-new Request class from the one that is received via arguments. The example of this approach could be found inside [WooPay checkout service](https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/woopay/services/class-checkout-service.php) where new request have been created from the current and some specific set of parameters have been added.

### Testing

Mocking is done automatically and all you have to do is to define what return of base request function that format request will return. Here are the example that will cover 99% of your need.

The request mock have been added to core test class and mocking request is way easier now.

In the most cases, you will need something like this:

```
$request = $this->mock_wcpay_request( Create_And_Confirm_Intention::class );

		$request->expects( $this->once() )
			->method( 'format_response' )
			->willReturn( $this->payment_intent );
```
If you don't need to test request parameters, this small piece of code will mock your requests effortlessly and efficiently. 

If you need more detailed or advanced request test where you need to test request parameters, the following example could be used.
```
$request = $this->mock_wcpay_request( Create_And_Confirm_Intention::class );

		$request->expects( $this->once() )
			->method( 'set_customer' )
			->with( self::CUSTOMER_ID );

		$request->expects( $this->once() )
			->method( 'set_payment_method' )
			->with( self::PAYMENT_METHOD_ID );

		$request->expects( $this->once() )
			->method( 'set_cvc_confirmation' )
			->with( null );

		$request->expects( $this->once() )
			->method( 'set_amount' )
			->with( 5000 )
			->willReturn( $request );

		$request->expects( $this->once() )
			->method( 'set_currency_code' )
			->with( 'usd' )
			->willReturn( $request );

		$request->expects( $this->once() )
			->method( 'setup_future_usage' );

		$request->expects( $this->once() )
			->method( 'set_capture_method' )
			->with( false );

		$request->expects( $this->once() )
			->method( 'set_off_session' )
			->with( false );

		$request->expects( $this->once() )
			->method( 'set_capture_method' )
			->with( false )
			->willReturn( $request );

		$request->expects( $this->once() )
			->method( 'set_metadata' )
			->with(
				$this->callback(
					function( $metadata ) {
						$required_keys = [ 'customer_name', 'customer_email', 'site_url', 'order_id', 'order_number', 'order_key', 'payment_type' ];
						foreach ( $required_keys as $key ) {
							if ( ! array_key_exists( $key, $metadata ) ) {
								return false;
							}
						}
						return true;
					}
				)
			);
		$request->expects( $this->once() )
			->method( 'format_response' )
			->willReturn( $this->payment_intent );
```

All you have to do is to mock methods, what they will receive, and certain function will be called for request class. Forget about checking will API request class will get some boolean and what not. In this case, you validate setters and request methods and its usages, which is quite easier to fix and maintain.

### Old PR text. Keeping it for reference. We will remove it once we decide to merge PR.

> **This PR includes examples!**
> Check out [includes/core/server/class-temp-request-examples.php](https://github.com/Automattic/woocommerce-payments/pull/5124/files#diff-c94d03fbf684a3218266ae652189248b0b7b032f9cde0643e0b0a5d1de3b918f) to see them.
> The front-end will show the output. Use http://localhost:8082/?show_basic_examples to see it.

There is a new `WCPay\Core\Server\Request` class:
- It is abstract, but can be:
    - Used through `WCPay\Core\Server\Request\Generic`, allowing for existing requests to be wrapped and moved to the `send_request` method of the API client. This class does not have any validation, or additional structure right now.
    - Extended by specific requests, ex. `WCPay\Core\Server\Request\Create_Intent`.
- Uses an internal `$props` array, to privately store all needed values.
- Has abstract/default methods for APIs, HTTP methods, and generic request flags (is site specific, etc.)
- Contains the necessary logic in `get_params()` to ensure that all required parameters are set. Those can be defined as a `IMMUTABLE_PARAMS` constant in child classes, and the constant is retrieved from each individual class to ensure that child classes are not overwriting protected values.
- A `format_response` method, which in the future can be overwritten to return specific responses in child classes.
- A protected `set_param` method, which can be used by inheritants to modify the internal `$params` array.
- Introduces protected mode, which makes it impossible to alter existing protected parameters. This is useful when the class is process through hooks.
- A static `extend` method, which can be used through extended classes to create a new instance, which already contains the same values as the parent object. It is only available when using the `apply_filters` method (see below).

Examples of how the class can be extended can be found in the `Create_Intent` and `WooPay_Create_intent` classes.

**Here is the important part:**

Some details about the `apply_filters` and `extend` methods. `extend` can only be called through `apply_filters`, here is how it cal be called:

```php
<?php
// Imagine you are somewhere in the gateway's `process_payment` method:
$request = WC_Payments::create_request( WCPay\Core\Server\Request\Create_Intent::class )
    ->set_amount( 100 )
    ->set_currency( 'eur' );

// Once everything is prepared, allow modifications. This will also check if all required fields are present. Note that just like the normal `apply_filters`, the method supports an arbitrary number of parameters:
$some_additional_field = 'xyz';
$response = $request->send( 'wcpay_create_intent_request', $some_additional_field ); 
```
 
Once the request is created, this allows it to be safely modified and extended. Here is an example:

```php
add_filter( 'wcpay_create_intent_request', 'woopay_modify_create_intent', 10, 2 );
function woopay_modify_create_intent( WCPay\Core\Server\Request\Create_Intent $base_request, $some_additional_field ): WCPay\Core\Server\Request\WooPay_Create_Intent {
    echo $some_additional_field; // Don't do that, the point is it's available.

    $request = WCPay\Core\Server\Request\WooPay_Create_Intent::extend( $base_request );
    $request->set_save_payment_method_to_platform( true );
    $request->set_amount( 300 ); // This will throw an exception, as the amount is protected.
    return $request;
}
```

Required implementation for request-specific classes includes (except the class skeleton):
1. An `IMMUTABLE_PARAMS` constant, if required.
2. A `REQUIRED_PARAMS` constant for defining required fields.
3. One setter for each prop.

A nice side-effect of the way requests can be overwritten by consumers (example above) is that we do not need a specific method for each request within the API client class, as there would be no possibility to overwrite the type of request before it reaches the `send_request` method.

### Closed Tasks:

- [x] [Make sure all parameters are set](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/c9e14b9d588bd28f540766c38a48ac2e2ab8e16c).
- [x] ~Should any parameters be overwritable?~ Mutable properties can just be overwritten.
- [x] Make sure the generic class can be easily used a replacement of what's currently there: Tried it within a couple of basic requests, it works, but is not commited.
- [x] ~Check if the generic class can be extended similarly to specific classes.~ It does not need to be extended, all params can be changed already.
- [x] [Decide how to instantiate requests.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] [Handle boolean variables properly](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/615dfec60c3afddae3eb177c8abfed10e78dd603).
- [x] [Handle level3 data in the right place](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/a977affe7a9733a45f3f3a476611a70de78f22dd).
- [x] [Update the docs of `apply_filters`.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/91d10f3939d6e6f1a867f84572ab3fccf27b1997)
- [x] [Restore the unique keys.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] [Add proper dependencies](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)
- [x] Revisit the `Intention` trait: The code is pretty clear, and does not seem to require changes.
- [x] [Rethink the create_and_confirm intention calls while adding validation and helpers](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/ef10ff85f66961312a7c1ba7bdfd93dc46dd7668?diff=split&w=0).
- [x] [Require a key within the `send_request` method for the filter to avoid doing it manually](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/e2933240c2c1b0b62deb39c124e3b8260c04b41d).
- [x] [Remove dependencies on the HTTP client, and make `WC_payments::get_wc_payments_http()` private again.](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce)

### Remaining tasks

Here is a list of open tasks for the PR:

#### Responses

Currently theis is a basic [`WCPay\Core\Server\Response` class](https://github.com/Automattic/woocommerce-payments/blob/spike/rado-request-classes/includes/core/server/class-response.php), which simply wraps a successfull response.

**This PR should not focus on responses (it's way too huge already)**, but while creating a base to rewrite request methods, we should make sure that we also have a solid base to build upon for responses later.

Main point: Currently we're still using a lot of functionality from the API client class, but that seems to have attracted a lot of quick & kind-of-dirty solutions like response formatting (examine charges and intentions). This functionality includes error handling (converting errors to exceptions), which should remain in the API client class until we remove all request methods from the class, but we need to make sure that it would be easy for us to move it into the `Response` class later.

This is not something we want to do yet, but currently intents and charges are being "deserialized" (turned into DTOs) within the API client class. Consider what dependencies/actions/filters will be required in order to extract this functionality into custom `format_response` request methods.

In the end, we should have the right infrastructure in place in the PR, then create another PR/issue to implement everything described here.

#### Extract WooPay code to the right place

There is already a working [`WCPay\Core\Server\Request\WooPay_Create_And_Confirm_Intention` request class](https://github.com/Automattic/woocommerce-payments/pull/5124/commits/6e1bf4b23ca55cd7bc342b39467b1bc6577ba85d) in this PR, but the WooPay code, which prepares the parameters for the request [still lives within the gateway class](https://github.com/Automattic/woocommerce-payments/blob/spike/rado-request-classes/includes/class-wc-payment-gateway-wcpay.php#L989-L990).

We need to figure out where to place that code, and how to link it. This could be a follow-up PR, based on this branch as well.

#### Use the right exceptions

Right now the PR is using a generic `\Exception` everywhere, this should be replaced with appropriate exception classes, and translatable messages.

#### Use ENUMs wherever possible

 I'm not sure where else/how we could use enums here:

 - It's hard to validate APIs, as we're receiving the raw API, which we cannot correlate back to a constant.
 - Request methods are validated already, and using ENUM.

Let's take a final look to make sure there are no missed opportunities.

#### Add unit tests and docs

Biggest task so far, yet pretty self-explanatory.

## Testing instructions

Two options right now:

- Check examples
- Perform a standard payment to see the `Create_And_Confirm_Intention` and `WooPay_Create_And_Confirm_Intention` requests in action.

-------------------


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -978,22 +971,23 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
-					$save_payment_method_to_store,
-					$payment_information->should_save_payment_method_to_platform(),
-					$metadata,
-					$this->get_level3_data_from_order( $order ),
-					$payment_information->is_merchant_initiated(),
-					$additional_api_parameters,
-					$payment_methods,
-					$payment_information->get_cvc_confirmation()
-				);
+				add_filter( 'woopay_create_and_confirm_intention_request_from_base_intent_request', [ Checkout_Service::class, 'create_woopay_intention_request' ], 10, 3 );
```
RadoslavGeorgiev: I'm fine leaving this line here within this PR, but in the long term it should be moved within `WooPay\Checkout_Service` as well.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -978,22 +971,23 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
+					->set_metadata( $metadata )
+					->set_level3( $this->get_level3_data_from_order( $order ) )
+					->set_off_session( $payment_information->is_merchant_initiated() )
+					->set_payment_methods( $payment_methods )
+					->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
+
+				if ( $save_payment_method_to_store ) {
+					$request->setup_future_usage();
+				}
+				$intent = $request->send( 'create_woopay_intention_request', $order, $this->is_platform_payment_method( $payment_information->is_using_saved_payment_method() ) );
```
RadoslavGeorgiev: `$this->is_platform_method()` is a WooPay method, and should not remain in the gateway. As a follow-up, and partially fixing https://github.com/automattic/woocommerce-payments/issues/4953, we should move out all WooPay-related method from this class, to the new service. This includes not only `is_platform_payment_method`, but also ones like `should_use_stripe_platform_on_checkout_page()`, and whatever they are modifying.

I'd like the one example we have of `$request->send()` and its filter to be a proper one, which would not need to change in a version or two of the gateway, when `should_use_stripe_platform_on_checkout_page`-like methods get extracted from this class.

I'm OK with making `is_platform_payment_method` public for now, and using it within `WCPay\WooPay\Service\Checkout_Service` instead of moving it, but this line should not contain anything woopay-related.

Actually, why did you change the filter to `create_woopay_intention_request`?! This is not a WooPay create intention request, it's a **WCPay** create intention request?

- The idea is to have WooPay use the same set of functionality, available for all other extensions.
- Other developers might think that this is a WooPay-specific filter and restrain from using it, even though it is not.
- By the way, this is supposed to be `wcpay_create_and_confirm_intention_request`, what happened to the `_and_confirm` part? There are other requests for `create_intention`, and it's still something that is and will be used.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Also, where does `woopay_create_and_confirm_intention_request_from_base_intent_request` from? Did you test this in any way? It's not being applied within manual tests (such a filter is not called anywhere), and I don't see it in tests.

RadoslavGeorgiev (changes requested):
Thanks for moving this forward, I left a couple of comments.

I see a few tasks open:
- Existing tests are broken (seeing unexecutable [tests in GitHub](https://github.com/Automattic/woocommerce-payments/actions/runs/3492455358/jobs/5846252638), and 30+ failures/errors locally. Let's fix those.
- New tests: I haven't had the opportunity to look into them, and it seems like I won't be able to do it this week either.
- Responses: Anything to change within this PR?
- The mechanism needs to be documented, preferably with a markdown file within the repository. Let's not forget that before diving into particular requests.

includes/class-wc-payments.php
```diff
@@ -244,6 +244,26 @@ public static function init() {
...
 		include_once __DIR__ . '/exceptions/class-api-exception.php';
 		include_once __DIR__ . '/exceptions/class-connection-exception.php';
+		include_once __DIR__ . '/core/class-mode.php';
+
+		// Include core exceptions.
+		include_once __DIR__ . '/core/exceptions/class-invalid-request-parameter-exception.php';
+		include_once __DIR__ . '/core/exceptions/class-immutable-parameter-exception.php';
+		include_once __DIR__ . '/core/exceptions/class-extend-request-exception.php';
+		include_once __DIR__ . '/core/exceptions/class-server-response-exception.php';
+		include_once __DIR__ . '/core/exceptions/class-server-request-exception.php';
```
KarlisJ: Should we maybe namespace the Exceptions with `Server` same as for `core/server` classes? 

This namespace + class name combination 
```php
/core/exceptions/class-server-response-exception.php
```

explain the scope very well, however this 
```php
/core/exceptions/class-extend-request-exception.php
```

isn't. The name indicates that it can be used globally but implementation is meant only for specific `Server/` namespace use.

includes/class-wc-payments.php
```diff
@@ -1301,4 +1323,28 @@ public static function enqueue_dev_runtime_scripts() {
...
 	}
+
+	/**
+	 * Creates a new request object for a server call.
+	 *
+	 * @param  string $class_name The name of the request class. Must extend WCPay\Core\Server\Request.
+	 * @return Request
+	 * @throws Exception          If the request class is not really a request.
+	 */
+	public static function create_request( $class_name ) {
```
KarlisJ: Why do we need the factory method here? Is there a strong reason not to put it in the `Request` class? 

We're losing valuable typehints in IDE with this move.

RadoslavGeorgiev: > Why do we need the factory method here?

Dependencies. I think it was added in the context of level5 data, but overall this allows the request to be created with some dependencies, which are otherwise private to the `WC_Payments` class.

On the other hand, I would not mind the request constructor/static method to do call something like `WC_Payments::inject_server_request_dependencies( $this )` in there. This way we could still only provide those dependencies to the request without making them truly public.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+use WCPay\Core\Exceptions\Invalid_Request_Parameter_Exception;
+
+/**
+ * Base for requests to the WCPay server.
+ */
+abstract class Request {
+	/**
+	 * Contains a set of params, which the class considers immutable by others.
+	 *
+	 * Overwrite this in your class for individual properties.
```
KarlisJ, line 27: Small nitpick but can we add the same description lines to the 3 constants below?

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+	 * @param WC_Payments_API_Client     $api_client     The API client to use to send requests.
+	 * @param WC_Payments_Http_Interface $http_interface The HTTP interface for the server.
+	 */
+	public function __construct( WC_Payments_API_Client $api_client, WC_Payments_Http_Interface $http_interface ) {
+		$this->api_client     = $api_client;
+		$this->http_interface = $http_interface;
+	}
+
+	/**
+	 * Returns the needed API.
```
KarlisJ, line 121: From the example, it's clear what the "API" is meant but overall the description it's not clear how the "API" should look.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+					get_class( $this ),
+					implode( ', ', $missing_params )
+				),
+				'wcpay_core_invalid_request_parameter_missing_parameters'
+			);
+		}
+
+		foreach ( $params as $key => $value ) {
+			if ( true === $value ) {
+				// The WCPay server requires the string 'true'.
```
KarlisJ, line 192: Shouldn't we also handle `false` boolean?

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+
+	/**
+	 * Validates Stripe identifiers.
+	 *
+	 * @param  string     $id        The identifier to validate.
+	 * @param  mixed|null $prefixes  A prefix or an array of them (Optional).
+	 * @throws Invalid_Request_Parameter_Exception An exception if the format is not matched.
+	 * @return void
+	 */
+	protected function validate_stripe_id( $id, $prefixes = null ) {
```
KarlisJ, line 488: This looks more fit for dedicated support/helper class that is used within this class.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+
+	/**
+	 * Validate is number larger than passed compared number.
+	 *
+	 * @param  float $value_to_validate Value to validate.
+	 * @param  float $value_to_compare Value to compare.
+	 * @throws Invalid_Request_Parameter_Exception An exception if the format is not matched.
+	 * @return void
+	 */
+	protected function validate_is_larger_then( float $value_to_validate, float $value_to_compare ) {
```
KarlisJ, line 536: This looks more fit for dedicated support/helper class that is used within this class.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+
+	/**
+	 * Currency code validator.
+	 *
+	 * @param string $currency_code Currency code.
+	 *
+	 * @return void
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function validate_currency_code( string $currency_code ) {
```
KarlisJ, line 555: This looks more fit for dedicated support/helper class that is used within this class.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+	}
+
+	/**
+	 * Allows the request to be modified, and then sends it.
+	 *
+	 * @param string $hook    The filter to use.
+	 * @param mixed  ...$args Other parameters for the hook.
+	 * @return mixed          Either the response array, or the correct object.
+	 */
+	final public function send( $hook, ...$args ) {
```
KarlisJ, line 226: I don't see an example of this method in this PR. The sending of requests is not performed using this method.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+		$request = $this->apply_filters( $hook, ...$args );
```
KarlisJ: Shouldn't we account for cases when multiple filters are used? What if the extended request has different filters than base request?

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+
+	/**
+	 * Get request param by key.
+	 *
+	 * @param string $key Key to get.
+	 *
+	 * @return mixed
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	final public function get_param( $key ) {
```
KarlisJ, line 208: What would be the use case for this method?

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+	 *
+	 * Use this method within child classes in order to allow
+	 * those properties to be protected by overwriting.
+	 *
+	 * @param  string $key   The name of the parameter.
+	 * @param  mixed  $value And the value to set.
+	 * @return static        The instance of the class, ready for method chaining.
+	 */
+	final protected function set_param( string $key, $value ) {
+		if ( $this->protected_mode && in_array( $key, $this->get_immutable_params(), true ) ) {
```
KarlisJ, line 279: Could we split this check in two so we'd have different exception messages for each case?

includes/core/server/class-request.php
```diff
@@ -0,0 +1,523 @@
...
+	 */
+	public function should_return_raw_response(): bool {
+		return false;
+	}
+
+	/**
+	 * Returns all of the parameters for the request.
+	 *
+	 * @return array
+	 * @throws Invalid_Request_Parameter_Exception If the request has not been initialized yet.
```
KarlisJ, line 166:
```suggestion
* @throws Invalid_Request_Parameter_Exception If the request has not been initialized properly.
```

It's not clear what "initialization" means in this context

includes/core/server/request/class-create-and-confirm-intention.php
```diff
@@ -0,0 +1,217 @@
...
+		return $this->set_param( 'currency', $currency_code );
+	}
+
+	/**
+	 * Payment method setter.
+	 *
+	 * @param  string $payment_method_id ID of payment method to process charge with.
+	 * @return static                    Instance of the class for method chaining.
+	 */
+	public function set_payment_method( $payment_method_id ) {
```
KarlisJ: missing typehint

includes/core/server/request/class-create-and-confirm-intention.php
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	/**
+	 * Formats the response from the server.
+	 *
+	 * @param  mixed $intention_array The response from `WC_Payments_API_Client::request`.
+	 * @return mixed                  Either the same response, or the correct object.
+	 */
+	public function format_response( $intention_array ) {
+		return WC_Payments::get_payments_api_client()->deserialize_intention_object_from_array( $intention_array );
```
KarlisJ: For consistency and future-proof 
```suggestion
return new Response( WC_Payments::get_payments_api_client()->deserialize_intention_object_from_array( $intention_array ) );
```

includes/core/server/request/class-generic.php
```diff
@@ -0,0 +1,126 @@
...
+	private $should_use_user_token = false;
+
+	/**
+	 * Creates a new instance of the class.
+	 *
+	 * @throws Server_Request_Exception
+	 */
+	public static function create() {
+		throw new Server_Request_Exception( 'You cannot create request this way.', 'wcpay_core_server_request_invalid_method_call' );
+	}
```
KarlisJ: I don't understand this method. The parent class doesn't have such a method to override.

includes/core/server/request/class-generic.php
```diff
@@ -0,0 +1,126 @@
...
+	/**
+	 * Generic setter for parameters.
+	 *
+	 * @param  string $key   Key of the parameter.
+	 * @param  mixed  $value Value of the parameter.
+	 * @return Generic       Instance of the class for method chaining.
+	 */
+	public function set( $key, $value ) {
+		// Use the `Request` setter here.
+		$this->set_param( $key, $value );
```
KarlisJ, line 87: Why do we need this double-abstraction? Couldn't we simply call `set_param` from the constructor?

includes/core/server/request/class-generic.php
```diff
@@ -0,0 +1,126 @@
...
+use WCPay\Core\Server\Request;
+use Requests;
+
+/**
+ * Generic WCPay Server Request.
+ *
+ * This class is not extendable, if you need something specific,
+ * it will not be generic anymore, create your own.
+ */
+final class Generic extends Request {
```
KarlisJ, line 20: It's great that we have this class for quick requests to the server. Though, I fail to see (haven't run the code) how this class is locked to extending.

includes/core/server/request/class-create-and-confirm-intention.php
```diff
@@ -0,0 +1,217 @@
...
+		return $this->set_param( 'customer', $customer_id );
+	}
+
+	/**
+	 * Capture method setter.
+	 *
+	 * @param  bool $manual_capture Whether to capture funds via manual action.
+	 * @return static               Instance of the class for method chaining.
+	 */
+	public function set_capture_method( bool $manual_capture = false ) {
```
KarlisJ: The default value of `false` feel very counter-intuitive

KarlisJ (commented):
Hats off, this is great improvement from previous iteration and I love the direction! 

Left several questions and comments. 

It took rather some time to understand "extending" inside-out. At first I wanted to suggest some changes but after deep dive realised that current implementation does a great job handling all my concerns. Bottom line - we'll need proper documentation for these changes.

includes/core/server/class-request.php
RadoslavGeorgiev: I prefer not to pollute the class with validator methods as well, but some of them (this one is a good example) need the dependencies, present here. Maybe we could isolate them within a single-use trait for better readability?

includes/core/server/request/class-generic.php
RadoslavGeorgiev: The parent class used to have it until https://github.com/Automattic/woocommerce-payments/pull/5136/commits/49170a83303cf7a2e5d26d8e70104972f9e51dce#diff-8f66861195fea9a9c2c7473a1ab6691d71e2cff20214347eca77c728fd4ea65b, this is a left-over, which should be removed.

includes/core/server/request/class-generic.php
RadoslavGeorgiev: Generic requests will still allow the `->send`/`->apply_filters` method, meaning that upon a filter, others should be able to use `->set` without having to replace the object.

includes/core/server/request/class-create-and-confirm-intention.php
RadoslavGeorgiev: Good point, it just used to be the default in the API client method, and this was thoughtlessly carried over here.

includes/core/server/request/class-generic.php
RadoslavGeorgiev: If you can use `->set`, then you probably don't need to replace the whole method. The point here is to:

1. Allow a quick way to develop new features without having to define the request from the start. We'll recommend not merging the PR with such an instance in the docs.
2. If we don't have time to create classes for all existing requests within the current project, we can easily wrap the request within a `Request` class to allow future compatibility with filters/etc. I haven't thought about it thoroughly though, it could go away.

includes/core/server/class-request.php
RadoslavGeorgiev: Maybe you are missing the point: The parameter is only immutable while the object is in protected mode (being extended or in a filter).

includes/core/server/class-request.php
RadoslavGeorgiev: I did not add it in the beginning, because within the gateway it should not have a case, but consumers might benefit from checking the existing parameters of the request, so I prefer to keep it.

includes/core/server/class-request.php
RadoslavGeorgiev: It is now :)

The only example is https://github.com/Automattic/woocommerce-payments/blob/dev/core-requests-class/includes/class-wc-payment-gateway-wcpay.php#L990, but it's a showcase how this both sends the request, and allows it to be modified. For me this is a practice worth enforcing.

includes/core/server/class-request.php
RadoslavGeorgiev: Do you mean that it requires more context (additional args for the filter)? Otherwise this is a `final` method exactly to prevent child classes from messing with the main request.

Now, the question might be "What if two separate pieces of code try to apply their own `Custom_Request::extend()` method?" We probably need to cover this scenario as well actually, but I think it's doable. @zmaglica could you tinker with this idea?

includes/core/server/class-request.php
KarlisJ: Yes, I was missing the point.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: This is addressed in this commit: https://github.com/Automattic/woocommerce-payments/pull/5136/commits/6df27962755b9e06dcc5325a13d9ded7a211fc09 . The functions are moved to WooPay Checkout service. If we agree on this approach, I will replace/fix  all duplicate occurrences

includes/class-wc-payments.php
zmaglica: I agree with you on this topic. Will refactor.

includes/core/server/class-request.php
zmaglica: I am not sure do we handle 'false' on server, but as you suggested, I will add both true and false.

includes/class-wc-payments.php
KarlisJ: Couldn't we use `wc_get_container()` for this?

includes/core/server/class-request.php
KarlisJ: Hm, I don't see how this method have dependencies that restricts moving it. 

```
WC_Payments::get_account_service()->get_cached_account_data()
```

can be called from anywhere and the level of complexity and/or testability wouldn't change.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: This is resolved with new commits. The filter naming is returned and some of the functions are for now extracted to specific WooPay service. To maintain backward compatibility, there could be a duplicate code, but we can easily move it once we migrate all API requests to request classes.

The filter part is also extracted to a different (custom) WooPay service with brand new folder and namespace.

includes/core/server/class-request.php
zmaglica: We kept them here for now, but we will migrate them to a separate Validator class once we add them more.

includes/core/server/class-request.php
```diff
@@ -0,0 +1,531 @@
...
+		}
+		throw new Invalid_Request_Parameter_Exception( 'The passed key does not exist in Request class', 'wcpay_core_invalid_request_parameter_uninitialized_param' );
+	}
+
+	/**
+	 * Allows the request to be executed and formatted.
+	 *
+	 * @return mixed          Either the response array, or the correct object.
+	 */
+	final public function send() {
```
RadoslavGeorgiev: I do not understand why the `apply_filters` parameters and functionality were removed [in this commit](https://github.com/Automattic/woocommerce-payments/commit/aef3bb360e583a8a518b96b499341376474a3765#diff-8f66861195fea9a9c2c7473a1ab6691d71e2cff20214347eca77c728fd4ea65b). The idea behind having them here was that it would enforce requests to go through filters, and allow them to be more "consumable". Could you please restore that?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -978,22 +972,23 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
+				$request->set_metadata( $metadata );
+				$request->set_level3( $this->get_level3_data_from_order( $order ) );
+				$request->set_off_session( $payment_information->is_merchant_initiated() );
+				$request->set_payment_methods( $payment_methods );
+				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
+				if ( $save_payment_method_to_store ) {
+					$request->setup_future_usage();
+				}
+
+				$request = $request->apply_filters( 'create_woopay_intention_request', $order, $payment_information->is_using_saved_payment_method() );
```
RadoslavGeorgiev: Why did you change this from a `wcpay_` filter to a `woopay_` filter? At this point the request does not contain anything WooPay-related, so the name doesn't even make sense.

RadoslavGeorgiev: And why not provide all of the `$payment_information` object? This would give the filter a lot more versatility compared to just one boolean.

includes/core/exceptions/class-immutable-parameter-exception.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+namespace WCPay\Core\Exceptions;
+
+use WCPay\Exceptions\Base_Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Exception for throwing an error when someone tries to change immutable param.
```
RadoslavGeorgiev:
```suggestion
 * Exception for throwing an error when someone tries to change an immutable param.
```

includes/core/exceptions/class-extend-request-exception.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+namespace WCPay\Core\Exceptions;
+
+use WCPay\Exceptions\Base_Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Exception for throwing an error when request class is extended in wrong way.
```
RadoslavGeorgiev:
```suggestion
 * Exception for throwing an error when a request object is extended in wrong way.
```

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -859,7 +862,7 @@ public function update_saved_payment_method( $payment_method, $order_id, $is_tes
 	 * @return array|null                      An array with result of payment and redirect URL, or nothing.
 	 * @throws API_Exception                   Error processing the payment.
 	 * @throws Add_Payment_Method_Exception    When $0 order processing failed.
-	 * @throws Intent_Authentication_Exception When the payment intent could not be authenticated.
+	 * @throws Intent_Authentication_Exception|WC_Data_Exception When the payment intent could not be authenticated.
```
RadoslavGeorgiev: Where did `WC_Data_Exception` come from? I'm not sure it should remain, but if it does, let's add another `@throws` tag on a new line here, and describe when to expect it.

includes/class-wc-payments.php
RadoslavGeorgiev: FWIW I'd move all request-related (all except the response exception) within `core/exceptions/server/request/` and `WCPay\Core\Exceptions\Server\Request`.

includes/class-wc-payments.php
```diff
@@ -310,6 +330,8 @@ public static function init() {
 			include_once __DIR__ . '/multi-currency/wc-payments-multi-currency.php';
 		}
 
+		\WCPay\WooPay\Service\Checkout_Service::init();
```
RadoslavGeorgiev: Let's move this a bit lower in the file, next to the instantiation of most services, and store a local handle to it, ex:

```php
self::$woopay_checkout_service = new Checkout_Service();
self::$woopay_checkout_service->init();
```

The `->init()` part is not something we normally encounter within this file, but we should, because it would mean that a lot of hooks were removed from constructors.

includes/class-wc-payments.php
RadoslavGeorgiev: Similarly to autoloading, this is a topic we're really interested in, but is out of the scope of this work.

includes/core/exceptions/class-invalid-request-parameter-exception.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+namespace WCPay\Core\Exceptions;
+
+use WCPay\Exceptions\Base_Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Exception for throwing an error when payment method can not be identified or used.
```
RadoslavGeorgiev:
```suggestion
 * Exception for throwing an error when an invalid parameter has been provided to a request.
```

The current message seems like a left-over.

includes/core/exceptions/class-server-request-exception.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Class Server_Response_Exception
```
RadoslavGeorgiev:
```suggestion
 * Class Server_Request_Exception
```

Request or response? :P This applies to the class description below as well.

includes/core/server/request/class-create-intent.php
```diff
@@ -0,0 +1,58 @@
...
+namespace WCPay\Core\Server\Request;
+
+use WCPay\Core\Exceptions\Invalid_Request_Parameter_Exception;
+use WCPay\Core\Server\Request;
+use WC_Payments_API_Client;
+
+/**
+ * Request class for creating intents.
+ */
+class Create_Intent extends Request {
```
RadoslavGeorgiev: Nothing to resolve here, just noting that this class was used as a dummy during development and is currently not representing the actual request. I would have suggested deleting it if there were no tests already written for it.

Let's keep that in mind when we start replacing requests throughout the gateway.

includes/core/server/request/trait-level3.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * Trait file for WCPay\Core\Server\Request\Intention.
```
RadoslavGeorgiev:
```suggestion
 * Trait file for WCPay\Core\Server\Request\Level3.
```

includes/woopay/services/class-checkout-service.php
```diff
@@ -0,0 +1,100 @@
...
+		return false;
+	}
+
+	/**
+	 * Whether we should use the platform account to initialize Stripe on the checkout page.
+	 *
+	 * @return bool
+	 */
+	public static function should_use_stripe_platform_on_checkout_page() {
+			// TODO: Add support for blocks checkout.
```
RadoslavGeorgiev: The indentation within this method is very inconsistent.

includes/woopay/services/class-checkout-service.php
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	/**
+	 * Determine if current payment method is a platform payment method.
+	 *
+	 * @param boolean $is_using_saved_payment_method If it is using saved payment method.
+	 *
+	 * @return boolean True if it is a platform payment method.
+	 */
+	public static function is_platform_payment_method( bool $is_using_saved_payment_method ) {
```
RadoslavGeorgiev: There is no point of having this static, let's instantiate the service.

includes/woopay/services/class-checkout-service.php
```diff
@@ -0,0 +1,100 @@
...
+	 * @param Request   $base_request Base request.
+	 * @param \WC_Order $order Order.
+	 * @param bool      $using_saved_payment_method Using saved payment method.
+	 *
+	 * @return WooPay_Create_And_Confirm_Intention
+	 * @throws Invalid_Request_Parameter_Exception
+	 * @throws \WCPay\Core\Exceptions\Extend_Request_Exception
+	 */
+	public function create_woopay_intention_request( $base_request, $order, $using_saved_payment_method ) {
+		if ( ! $order ) {
```
RadoslavGeorgiev: As we control the code within the gateway, I'd trust it not to changes those filters. As a solution in the middle, let's define types for the parameters of the method, and PHP will deal with it.

RadoslavGeorgiev (changes requested):
Checked the `includes` tree, I'll proceed with `tests` next.

includes/core/server/class-request.php
zmaglica: I was thinking that sometimes you won't need any filters to pass, and you will not be able to execute the request. I will return the code how it was and make `$hook` parameter default to null.

RadoslavGeorgiev: > sometimes you won't need any filters to pass, and you will not be able to execute the request.

Wait, you do realize filters (and hooks in general, incl actions) are optional, right? By providing the hook, you allow somebody (incl. 3rd-party devs) to use it if/whenever it is needed. The more hooks a plugin has, the easier it is to adjust its behavior to suit your needs without having to change its code. That's why I am trying to enforce every request to go through a filter like this, and making it optional, renders the point moot.

includes/class-wc-payments.php
```diff
@@ -310,6 +331,8 @@ public static function init() {
 			include_once __DIR__ . '/multi-currency/wc-payments-multi-currency.php';
 		}
 
+		( new Checkout_Service() )->init();
```
RadoslavGeorgiev: Let's store the checkout service as a static property of the class here. Instantiating a class with hooks without storing it in a variable is one of the worst things you can do to a WordPress developer 😛

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -2069,6 +2115,30 @@ public function test_is_in_test_mode() {
 		$this->assertFalse( $this->wcpay_gateway->is_in_test_mode() );
 	}
 
+	public function test_request_class_example() {
```
RadoslavGeorgiev: Let's add a to-do to remove this method before merging.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions.php
```diff
@@ -226,10 +227,10 @@ function ( $id ) use ( $mock_subscription ) {
 			->with( self::USER_ID )
 			->willReturn( self::CUSTOMER_ID );
 
-		$this->mock_api_client
-			->expects( $this->once() )
-			->method( 'create_and_confirm_intention' )
-			->with( $this->anything(), $this->anything(), self::PAYMENT_METHOD_ID, self::CUSTOMER_ID, $this->anything(), false, false, $this->anything(), $this->anything(), true )
+		$request = $this->mock_wcpay_request( Create_And_Confirm_Intention::class );
```
RadoslavGeorgiev, line 232: I love how simple it was to replace this API client mock with a request class!

However, please notice that there is a check for the payment method and customer IDs, as well as some boolean flags. I think we need to carry them over.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
```diff
@@ -173,10 +174,9 @@ public function test_new_card_subscription() {
 		$subscriptions = [ WC_Helper_Order::create_order( self::USER_ID ) ];
 		$this->mock_wcs_get_subscriptions_for_order( $subscriptions );
 
-		$this->mock_api_client
-			->expects( $this->once() )
-			->method( 'create_and_confirm_intention' )
-			->with( $this->anything(), $this->anything(), self::PAYMENT_METHOD_ID, self::CUSTOMER_ID, $this->anything(), true, false, $this->anything(), $this->anything(), false )
+		$request = $this->mock_wcpay_request( Create_And_Confirm_Intention::class );
```
RadoslavGeorgiev, line 184: We have missing parameters here.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
```diff
@@ -282,10 +282,9 @@ public function test_saved_card_subscription() {
 		$subscriptions = [ WC_Helper_Order::create_order( self::USER_ID ) ];
 		$this->mock_wcs_get_subscriptions_for_order( $subscriptions );
 
-		$this->mock_api_client
-			->expects( $this->once() )
-			->method( 'create_and_confirm_intention' )
-			->with( $this->anything(), $this->anything(), self::PAYMENT_METHOD_ID, self::CUSTOMER_ID, $this->anything(), false, false, $this->anything(), $this->anything(), false )
+		$request = $this->mock_wcpay_request( Create_And_Confirm_Intention::class );
```
RadoslavGeorgiev, line 351: Missing parameters here as well. Notice how the 7th parameter (`$save_payment_method_to_store`) is not the same as the previous test.

tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
```diff
@@ -0,0 +1,214 @@
...
+		$request->set_capture_method( true );
+		$request->setup_future_usage();
+		$request->set_metadata( [ 'order_number' => 1 ] );
+		$request->set_level3( [ 'level3' => 'level3' ] );
+		$request->set_off_session();
+		$request->set_payment_methods( [ 'pm' => '1' ] );
+		$request->set_cvc_confirmation( 'cvc' );
+		$request->set_is_platform_payment_method();
+		$request->set_has_woopay_subscription();
+		$this->assertInstanceOf( WooPay_Create_And_Confirm_Intention::class, $request );
```
RadoslavGeorgiev: I guess initially this started in a different way, but right now you are using the `new` keyword above, so this test is redundant.

RadoslavGeorgiev (changes requested):
The tests look pretty well, I only have a couple of minor notes.

includes/core/server/class-request.php
zmaglica: I am aware that filters are optional, but the way you describe their usages is a way where you enforce it. Adding a filter for a sake of adding it doesn't help too much either. What if you need only request without any filters? With the approach, you mentioned here and with the required parameter in `send()` function make things a bit complicated. With this approach, you are using getters and setters to build the request and after that, you create a filter that will do something and in the cases where you need plain request class, the filter is not needed here.

I agree on your topic here regarding the extensibility with filters, but enforcing it is something we should avoid IMHO. That's why I refactored the send function to look something like this:

```
 	final public function send( $hook = null, ...$args ) {


		$request = $this;
		// In case where you don't want to send request.
		if ( $hook ) {
			$request = $this->apply_filters( $hook, ...$args );
		}

		return $this->format_response(
			$this->api_client->send_request( $request )
		);
	}
```

With this approach, if you need a filter, you can pass it and if you don't need it, simple send() will do the trick.

RadoslavGeorgiev: > I am aware that filters are optional, but the way you describe their usages is a way where you enforce it.

Yes, enforcing them is the point.

> Adding a filter for a sake of adding it doesn't help too much either.

It might not help too much in most cases, but it could be gold in others, in any case it never hurts. The only downside I could think of is that during tests you’ll need to provide a dummy hook name there.

> What if you need only request without any filters? With the approach, you mentioned here and with the required parameter in send() function make things a bit complicated.

How do you know that you need a request without filters? What does this even mean? Can you give me an example of such a filter, and explain it? With filters, you don’t know who will use them, and when they will use them. You just give them the option.

> you create a filter that will do something and in the cases where you need plain request class, the filter is not needed here.

What is a plain request class? When do you need it?

RadoslavGeorgiev: Keep in mind that the `wcpay_api_request_params` (currently in the API’s `send_request` method) defeats all of the protection and structure, which we’re introducing in this PR. We should aim to remove that filter once all (incl. in other open PRs) requests have been converted to classes, and at that point, we’d need to add a generic filter for requests classes somewhere anyway, so all requests will need to go through a filter anyway. Having a more specific filter when calling the `send()` method, as well as the seemingly significant parameters for that filter, will make WCPay a much better platform.

zmaglica: I just wanted to make optional this hook parameter, but as you suggested, we can leave it as a required. If the filter is not added or missing, nothing will be damaged.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: Done. Addressed in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd

includes/core/exceptions/class-immutable-parameter-exception.php
zmaglica: Done. Addressed in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd

includes/class-wc-payment-gateway-wcpay.php
zmaglica: I think it way typo Fixed.

includes/core/server/request/class-create-intent.php
zmaglica: Noted.

includes/core/server/request/trait-level3.php
zmaglica: Done. Addressed in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd

includes/woopay/services/class-checkout-service.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd

includes/woopay/services/class-checkout-service.php
zmaglica: Fixed. These methods are not static anymore.

includes/woopay/services/class-checkout-service.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd . The payment information object is passed that holds the order object.

includes/class-wc-payments.php
zmaglica: Valid point. Added static variable in https://github.com/Automattic/woocommerce-payments/pull/5136/commits/8f1aa8fd35a4f80de54b317d8892d0aae95753cd

tests/unit/test-class-wc-payment-gateway-wcpay.php
zmaglica: Added to-do remainder.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
zmaglica: Setters for filling this data have been added to test that will cover this case

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions.php
zmaglica: Setters for crucial data like customer id, payment method id and some booleans have been added to test

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
RadoslavGeorgiev: I don't see the changes. Have you pushed everything?

zmaglica: I am sorry. The linter prevented commit, and I didn't notice it until now. It is pushed.

tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
zmaglica: I agree. Before we had on each set function return statement of self and this self could change class at some point (even it will occur in extremely rare cases). I will remove this line.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
zmaglica: Missing parameters have been added.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions.php
RadoslavGeorgiev: AFAICT the most crucial parameter this test is checking for is `off_session => true` (the last boolean true) in the call, as well as making sure `setup_future_usage` and `save_payment_method_to_platform` are not set: the payment method should be set up and saved already in this scenario.

So I'd keep only the tests for those parameters, as well as maybe the customer and payment method IDs. The rest don't matter.

The `save_payment_method_to_platform` parameter is now in the WooPay request class, so ideally that would warrant a whole separate test for that part, but I think we're okay to skip it.

tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
RadoslavGeorgiev: I guess that's 👍 for removing the method chaining :)

However, the line is still there :p

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
RadoslavGeorgiev: Are you sure?
image

includes/core/server/class-request.php
RadoslavGeorgiev: Thank you! In that case let's remove the `if ( $hook ) {` block from the method (it must always be present).

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Almost there! Let's prefix it by `wcpay_` (`wcpay_create_intention_request`) to make sure we have a prefix.

includes/class-wc-payments.php
zmaglica: This is done.  all request exceptions are moved to its own directory.

RadoslavGeorgiev (changes requested):
Most of the changes in the code look good, just a couple of small details left.

I started looking at tests, but I have the feeling the lint process has kind of got rid of most of the changes. I recommend checking your `git stash`, sometimes they end up there.

includes/core/server/class-request.php
zmaglica: Fixed.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: Fixed.

tests/unit/test-class-wc-payment-gateway-wcpay-subscriptions-process-payment.php
zmaglica: The new asseerts have been added.

RadoslavGeorgiev (approved):

PR #5138, opened 2022-11-16
Add payment flows documentation

RadoslavGeorgiev:
TBD


RadoslavGeorgiev:
The diagram in this PR was used as a part of our duplicate charges investigation, but was never polished enough for merging, and is already outdated. Now that we're working on https://github.com/Automattic/woocommerce-payments/issues/6673, we will be adding proper documentation there.

PR #5149, opened 2022-11-17
Add idempotency header for create intent call for non UPE checkout

deepakpathania:
Fixes #

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/changelog/poc-idempotency-headers
+++ b/changelog/poc-idempotency-headers
@@ -1,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Add idempotency key to the create intent call
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1005,6 +1005,9 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				}
 			}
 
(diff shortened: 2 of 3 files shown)


jrodger:
I love the approach here (which is why I'm drive by reviewing it), after all, this is exactly what this Stripe API feature is for!

However, I've always struggled with how to implement it exactly. For example, what happens if a customer enters their card details, it fails, they use a different card for the same cart of goods? I appreciate this is a draft, but I think in this case we'd send the same key and the new request would fail?

I was wondering if we might find some inspiration from the [Stripe plugin](https://github.com/woocommerce/woocommerce-gateway-stripe/search?q=idempotency)?

deepakpathania:
>  what happens if a customer enters their card details, it fails, they use a different card for the same cart of goods?

I considered using the payment method also for the hash but wasn't sure if that would be available in all scenarios, will check that again once.

The Stripe plugin uses this [retry approach](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/541) for updating the idempotency key by using the retry count as a param in the key hash, they check if the `source_object` has the status `chargeable`, which triggers this but I am not sure if we have something similar to base this trigger condition on, will explore further, let me know if you have any ideas on the same. @jrodger

htdat:
One thing to consider for this PR is how we relay error messages from Stripe to shoppers/customers on the checkout page.

htdat:
I am going to close this PoC PR in favor of this comment paJDYF-6wC-p2#comment-15189

PR #5152, opened 2022-11-18
Use existing intents/order status before creating (and confirming)  new intents.

RadoslavGeorgiev:
TBD

In the meantime, commits have thorough messages, please check them one by one.


htdat:
I think this was a part of the duplicate charge project but it's no longer relevant so I am closing it.

Issue #5118, opened 2022-11-12
WooPay scripts are loaded when the WCPay gateway isn't available


a-danae:
### Describe the bug
When WCPay is excluded from the available payment methods, WooPay scripts are loaded on the checkout page even though they're not used.

### To Reproduce

1. Enable WooPay
2. Enable [Conditional Shipping and Payments](https://woocommerce.com/products/conditional-shipping-and-payments/)
3. Create a condition to disable WCPay on the checkout page
4. As a shopper, add a product to your cart and go to the checkout page. Make sure the condition you created to disable WCPay is met
5. On the checkout page, notice that WCPay isn't displayed in the available payment gateways
6. Open the browser developer tools -> Network -> search for checkout.js
7. Notice that WooPay scripts (platform-checkout.js) and styles (platform-checkout.css) are loaded

### Actual behavior
WooPay scripts and styles are loaded on the checkout page even though WCPay isn't available.

### Screenshots

WooPay scripts are being loaded.

<img width="1487" alt="image" src="https://user-images.githubusercontent.com/41606954/201481744-46ed5fa0-02d7-40ab-b15d-0ca7214a7e9c.png">

### Expected behavior
WooPay scripts and styles shouldn't be loaded if WCPay isn't available.

brettshumaker:
Noting that, while working on another PR, I noticed that `platform-checkout.js` and `platform-checkout.css` are both loaded on all pages - not just checkout.

dbrabyn:
I'm getting the express checkouts (Apple Pay for sure and Google Pay probably) buttons displaying when the WooCommerce Payments payment gateway is disabled via WC's Conditional Shipping & Payments plugin. Same issue?

I'm using the latest version of both: WooCommerce Payments (5.1.2) and WooCommerce Conditional Shipping and Payments (1.14.4).

PR #5153, opened 2022-11-18
Add/1427 style woopay express checkout button

brettshumaker:

This PR adds styling for the WooPay Express Checkout button as well as converting it to a React component. This was done because we will need a React component for blocks checkout and we wanted only **one** instance of the button that needs to be maintained (as opposed to having a static HTML button in addition to the React component).

#### Testing instructions

* Install this PR.
* Make sure the latest dev tools plugin is installed.
* Go to the dev tools page and click the Enable the WooPay Express Checkout button. checkbox and save the settings.
* Go to the WooCommerce Payments settings screen and click "Customize" next to WooPay under "Express Checkouts."
* Make changes to the button's settings.
* View the button on the front end of the site (product, cart, and shortcode checkout pages) and make sure that the button appears styled according to the designs in Figma.

-------------------

**Post merge**


client/checkout/platform-checkout/express-button/index.js
```diff
@@ -0,0 +1,35 @@
...
+	renderPlatformCheckoutExpressButton();
+} );
+
+// mount component again if parent fragment if re-rendered after ajax request by woocommerce core
+// https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/legacy/js/frontend/checkout.js#L372
+jQuery( function ( $ ) {
+	$( document ).ajaxComplete( function () {
+		renderPlatformCheckoutExpressButton();
+	} );
+} );
```
Mayisha: I think this part is not needed. In the [save-user checkbox section](https://github.com/Automattic/woocommerce-payments/blob/add/1427-style-woopay-express-checkout-button/client/checkout/platform-checkout/index.js#L66) we had to add this extra check because the save user section is pushed to the document from frontend. And after this ajax call it would disappear from the screen without this extra jquery check.
As we are preparing the placeholder `div` for the express button from the backend, this check might be unnecessary.

For me, the button loads perfectly if I remove these codes too. But you may want to verify it on your own as the ajax request order could be different for us.

client/checkout/platform-checkout/express-button/index.js
```diff
@@ -0,0 +1,35 @@
...
+ * Internal dependencies
+ */
+import { WoopayExpressCheckoutButton } from './woopay-express-checkout-button';
+
+const renderPlatformCheckoutExpressButton = () => {
+	const platformCheckoutContainer = document.getElementById(
+		'wcpay-platform-checkout-button'
+	);
+
+	ReactDOM.render(
```
Mayisha: How about checking if `platformCheckoutContainer` exists before trying to render the react component?

includes/class-wc-payments-platform-checkout-button-handler.php
```diff
@@ -94,14 +94,30 @@ public function scripts() {
...
+		wp_localize_script( 'WCPAY_PLATFORM_CHECKOUT_EXPRESS_BUTTON', 'wcpayWooPayExpressParams', $platform_express_button_params );
+
+		wp_set_script_translations( 'WCPAY_PLATFORM_CHECKOUT_EXPRESS_BUTTON', 'woocommerce-payments' );
+
+		wp_enqueue_script( 'WCPAY_PLATFORM_CHECKOUT_EXPRESS_BUTTON' );
+
 		wp_register_style(
-			'wcpay-platform-checkout-express-button-styles',
-			plugins_url( 'includes/platform-checkout/assets/css/platform-checkout-express-button.css', WCPAY_PLUGIN_FILE ),
+			'WCPAY_PLATFORM_CHECKOUT',
```
Mayisha, line 114: Is it required to register the styles as this CSS file is already registered [here](https://github.com/Automattic/woocommerce-payments/blob/add/1427-style-woopay-express-checkout-button/includes/platform-checkout-user/class-platform-checkout-save-user.php#L44-L49)? :thinking:

Mayisha (dismissed):
Nice work :tada: :tada: 

I noticed a little visual unalignment in the button text on the settings page vs the checkout page. On the settings page the text position looks a little to the bottom side whereas on the checkout page it looks a bit on top.

Button on the settings page-
settings button

Button on the checkout page-
checkout button

I have left a couple of other tiny pieces of feedback. Apart from that, I would like to suggest removing [this function](https://github.com/Automattic/woocommerce-payments/blob/add/1427-style-woopay-express-checkout-button/includes/class-wc-payments-platform-checkout-button-handler.php#L271) to get rid of the `OR` between this WooPay express checkout button and Gpay/Apple Pay.

includes/class-wc-payments-platform-checkout-button-handler.php
brettshumaker: Yes, I think so...eventually. Let me explain. Currently, the `Platform_Checkout_Save_User` is always loaded on every page and the scripts/styles are always enqueued. I haven't fully checked into it yet, but I don't _think_ that should happen - it should only be loaded when we expect the email field and the save user box to be rendered. So, if that gets fixed and is only loaded when appropriate, this code will still need to register/enqueue that CSS file unless we put it in its own file in `/dist`. I briefly tried to get webpack to build a new CSS file but couldn't get it to work.

client/checkout/platform-checkout/express-button/index.js
brettshumaker: Resolved in fe01b1ee

client/checkout/platform-checkout/express-button/index.js
brettshumaker: Resolved in ef6cb212

brettshumaker:
> I would like to suggest removing [this function](https://github.com/Automattic/woocommerce-payments/blob/add/1427-style-woopay-express-checkout-button/includes/class-wc-payments-platform-checkout-button-handler.php#L271) to get rid of the OR between this WooPay express checkout button and Gpay/Apple Pay

I looked into this a bit and it's going to be more complicated than just not hooking into that location and displaying the OR. If we just remove it, when there is more than one express payment option on the page, it would be fine. However, if WooPay is the only express checkout button present, there will be no divider between it and the content below.

Do you think the reworking of the "OR" separators should be spun out into its own issue since it will likely require touching the payment request button code or handled in this one?

brettshumaker:
Noting that it looks like something is up with the tests. All the failing ones have this error:

```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function _wp_theme_has_theme_json_clean_cache_upon_upgrading_active_theme(), 1 passed in /tmp/wordpress/wp-includes/class-wp-hook.php on line 310 and exactly 2 expected in /tmp/wordpress/wp-content/plugins/gutenberg/lib/compat/wordpress-6.2/get-global-styles-and-settings.php:67
```

includes/class-wc-payments-platform-checkout-button-handler.php
bborman22: I understand why you're putting this in in anticipation of the stylesheet not always being present. Do we have an issue to look in to that?

brettshumaker: > Do we have an issue to look in to that?

Not yet. I can make one, though.

bborman22 (approved):
This all tested well for me! I think it's very helpful to have everything centered in the one component, though I have some slight concerns about performance we can cross that bridge if it even comes up as a problem.

I think spinning the `OR` off into it's own issue is a good move here as you mentioned it would impact other payment request methods as well, so would be nice to have a broad solution.

LGTM!

includes/class-wc-payments-platform-checkout-button-handler.php
brettshumaker: Actually, it looks like this could be fixed along with #5118. I'll leave a comment there noting that the script/style files are being included on all pages.

PR #5154, opened 2022-11-18
Check for valid deadline when setting Update Business Details task

### LINKED ISSUE #4846: Update by January 1970 to avoid a disruption in deposits.
### Describe the bug
It is possible that a user may have pending requirements for the account that do *not* have a deadline, which can be stored within `future_requirements` of the account object.
This sets the account's status to `restricted_soon` and [prompts a task to complete their account information](https://github.com/Automattic/woocommerce-payments/blob/8cb0e930add4d4371d6e32fd7224f3d967a8d397/client/overview/task-list/tasks.js#L30). However, the deadline will be `false` in this case (because Stripe does not provide a deadline) instead of a timestamp and we don't check if the deadline is actually a timestamp [before trying to convert it](https://github.com/Automattic/woocommerce-payments/blob/8cb0e930add4d4371d6e32fd7224f3d967a8d397/client/overview/task-list/tasks.js#L39).

Thus, we see January 1970 as the date.

Slack discussion:
p1664971487465379-slack-C03KTTK2YMA

### Expected behavior
If the user has future requirements but no deadline, we could show a simpler message like `Update to avoid a disruption in deposits`.

dmallory42:
Fixes #4846

* Check whether we have a valid `currentDeadline` when setting the update business details task. 
* Don't show the task if we don't have a `currentDeadline` (since there's no action for the user to take right now).
* Also fixed a bug where we were not checking the proper key when checking for pastDue requirements, so the other task that would appear should work properly now (previously, it would never have been displayed).

#### Testing instructions

**Prerequisite**

There's a few tests here for showing the task based on various conditions. To test properly, you'll need to do the following before every test:

* Update your local server code to return `account_status`, `current_deadline` and `has_overdue_requirements` to the values specified in the test.
* Refresh the account cache using the dev tools.

**Tests**

Testing with a `restricted_soon` account with a current deadline.

* Update `account_status` to `restricted_soon`, and `current_deadline` to a date in the future.
* Check the `Payments > Overview` page. The task should appear and look like this:

Screenshot 2022-11-29 at 11 22 33

Testing with a `restricted_soon` account with no current deadline.

* Update your local server code to return the `account_status` as `restricted_soon` and `current_deadline` as `null` when fetching the basic account info.
* Check the `Payments > Overview` page. You should see no task related to updating account details.

Testing with a `restricted` account when `past_due` is false.

* Update your local server code to return `account_status` as `restricted` and `has_overdue_requirements` as `false`.
* Check the `Payments > Overview` page. You should see no task relating to updating account details.

Testing with a `restricted` account when `past_due` is true.

* Update your local server code to return `account_status` as `restricted` and `has_overdue_requirements` as `true`.
* Check the `Payments > Overview` page. You should see a task saying `Payments and deposits are disabled for this account until missing business information is updated.`

-------------------

**Post merge**


ismaeldcom:
>Usually, it seems when there is no currentDeadline then there is no actual requirement for the user to do anything. Would it be better to not show this message at all in this case? I haven't looked deeply into the possible cases this could occur.

According to our current findings if the returned `currentDeadline` is `null` there is no actual request for new details from Stripe. They could be required in the future. And we're not even sure if they could be submitted before they're actually requested.

So I'd vote for hiding the message, as it might sound like something mandatory. Aiming to mitigate the issue in the next release until we do a proper research on the server one.

WDYT?

dmallory42:
Thanks @ismaeldcom! I agree. If there's no requirement for the user to do anything right now it's better to hide the message altogether IMO.

ismaeldcom (approved):
Thanks for taking care of it, and nice catch with `has_overdue_requirements` being used instead of `pastDue`!

Everything looks good to me and tests well. So I'm preapproving it 🙂 

I just wonder if we could also handle those cases where the merchant already have the note with `Jan 1, 1970`. If I'm not wrong, they will keep seeing it until they manually remove  `wcpay_show_update_business_details_task` option.

I just made a quick test modifying `add_update_business_details_task` to not return early if the option value is `yes` and update it with `yes|no` based on the current properties. Causing it to show/hide the task when the cache is cleared.

Although, I'm not sure about how it can affect the expected behavior in other cases. What do you think?

dmallory42:
> I just wonder if we could also handle those cases where the merchant already have the note with Jan 1, 1970. If I'm not wrong, they will keep seeing it until they manually remove wcpay_show_update_business_details_task option.

That's a great point. I think they wouldn't see that message anymore, because we added a condition in the task list JS, but they would see a similar message that they don't actually need to see (just without the `1 Jan 1970` bit).

> I just made a quick test modifying add_update_business_details_task to not return early if the option value is yes and update it with yes|no based on the current properties. Causing it to show/hide the task when the cache is cleared.

I think this would fix it. The only issue with this is that it is a bit less efficient, but in this case I think it's worth it since we shouldn't be showing this option in a lot of cases.

I wonder if perhaps we could instead clear the option value when we clear the cache, or expire it after say 2 hours?

ismaeldcom:
As we discussed in Slack, much better now, there is no need to persist it as an option every time. And the evaluation is cheap enough to not use a transient either. Thanks for the internal discussion and update!

With this new change I have a nitpick 🙂  What do you think about replacing the `yes|no` value with a boolean? We will need to update `task-list/tasks.js` and `task-list/test/tasks.js` too, but seems like a straightforward change.

dmallory42:
Good catch, I think that makes the code simpler!

Updated in e1009156334ea244caba5790a515c5fb263d1bba and 9a3c0bda557d22bb4d286c0e9b5155242e8cc9d6.

ismaeldcom (approved):
Thanks for addressing it! Ready to be merged 🚀

PR #5156, opened 2022-11-18
fixed bug that would not allow a customer to add new payment methods in WooPay

### LINKED ISSUE #5155: Can't add new payment method in WooPay
### Describe the bug
After the changes in #4638 it's not possible for a customer to add a new payment method in WooPay because the `save_user_in_platform_checkout` POST key is not set making WC Payments try to add a payment method to a user in the merchant account which does not exist. 

The key `save_payment_method_in_platform_account` is available. I believe we ended up misreading the key name and refactored to use the incorrect function.

### To Reproduce
- As a shopper checkout using WooPay or go to /account page of WooPay
- Try to add a new payment method
- You'll see an error telling you that the customer was not found

leonardola:
Fixes #5155

This PR fixes a bug introduced in #4638 which made it not possible for a customer to add a new payment method in WooPay because the `save_user_in_platform_checkout` POST key is not set making WC Payments try to add a payment method to a user in the merchant account which does not exist. 

The key `save_payment_method_in_platform_account` is available. I believe we ended up misreading the key name and refactored to use the incorrect function.

#### Testing instructions
- As a shopper checkout using WooPay or go to /account page of WooPay
- Try to add a new payment method
- No error should show up
- As a shopper add a new payment method in the merchant store
- No error should show up

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-5155-cant-add-payment-method
+++ b/changelog/fix-5155-cant-add-payment-method
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+fixed bug that would not allow customers to add new payment methods to WooPay
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2635,7 +2635,12 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 */
 	public function create_and_confirm_setup_intent() {
 		$payment_information             = Payment_Information::from_payment_request( $_POST ); // phpcs:ignore WordPress.Security.NonceVerification
-		$should_save_in_platform_account = $this->platform_checkout_util->should_save_platform_customer();
+		$should_save_in_platform_account = false;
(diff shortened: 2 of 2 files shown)


leonardola:
@Mayisha I added you as the PR reviewer because a PR you worked on introduced the bug. So you might have more context to tell if that should be fixed in another way. Thanks in advance.

malithsen (approved):
Thanks for the fix. This looks good to me but let's wait for @Mayisha's input as well. 

For some additional context, `should_save_in_platform_account` was introduced in https://github.com/Automattic/woocommerce-payments/commit/23a3b08ba18f5dc855798d233267dd44baf1ac3b to differentiate the `setup_intent` usage in payment method creation and the use for trial subscriptions.

Mayisha (approved):
I actually confused `save_payment_method_in_platform_account` with `save_user_in_platform_checkout` :facepalm:  

Thanks for the fix @leonardola :rocket:

PR #5159, opened 2022-11-21
Fix undefined element error on Cart block for WooPay enabled site.

Mayisha:
Added a check before attaching `MutationObserver` on the Checkout Block element.

#### How to recreate the bug
- Install `woocommerce-gutenberg-product-blocks`.
- Create `Cart` page with `Cart Block`.
- Add some product, go to the `Cart` page and open the browser console.
- Notice that there is an error saying `Uncaught (in promise) TypeError: MutationObserver.observe: Argument 1 is not an object.`

#### Testing instructions
- Install `woocommerce-gutenberg-product-blocks`.
- Create `Cart` page with `Cart Block`.
- Add some product, go to the `Cart` page and open the browser console.
- Ensure that there is no error message in the console.
- Go to `Checkout` page and ensure that nothing has changed there. Providing an email should trigger the platform checkout as before.

### DIFF
--- a/changelog/fix-woopay-cart-block-console-error
+++ b/changelog/fix-woopay-cart-block-console-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix undefined element error on Cart block for WooPay enabled site.
--- a/client/checkout/blocks/index.js
+++ b/client/checkout/blocks/index.js
@@ -52,7 +52,11 @@ registerPaymentMethod( {
 
 registerExpressPaymentMethod( paymentRequestPaymentMethod( api ) );
 
-if ( getConfig( 'isPlatformCheckoutEnabled' ) ) {
+// If platform checkout is enabled and this is the checkout page.
+if (
+	getConfig( 'isPlatformCheckoutEnabled' ) &&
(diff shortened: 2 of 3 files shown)


bborman22 (approved):
This fixes the error and is simple enough! I was curious though why the `isBlocksCheckout` parameter is true here (or really why this is getting loaded on the cart page?) Is it related to this issue/comment: https://github.com/Automattic/woocommerce-payments/issues/5118#issuecomment-1320441454 This just a question for my own curiousity though, the PR looks good!

LGTM!

Mayisha:
@bborman22 I found that [this file](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/index.js) loads for both the cart and checkout block. Seeing how the payment methods and express methods are registered here, I think this is intentional that it loads for both cart and checkout. That's all my assumption though :grimacing: 

`isBlocksCheckout` is `true` here because of this hard-coded `true` [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/index.js#L56). On second thought I can actually check if this is the cart page or checkout page before calling this too. I think I am actually gonna do that :thinking:

PR #5162, opened 2022-11-22
Introduce WCPay_Order

### LINKED ISSUE #4951: WCPay Platform: Meta Data

jessy-p:
Fixes #4951 

- methods for accessing and updating order metadata were introduced into `WC_Payments_Order_Service`
- refactored code that directly sets and gets order metadata to use the new methods.

#### Testing instructions
- Tests pass
- Functionality of refactored code works as expected.

-------------------

**Post merge**


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2950,6 +2951,17 @@ public function append_stripelink_button( $field, $key, $args, $value ) {
...
 	}
 
+
+	/**
+	 * Get the WC_Payments_Order from order id.
+	 *
+	 * @param  mixed $order the order_id or order object.
+	 * @return WC_Payments_Order order object with additional methods.
+	 */
+	public function wcpay_get_order( $order ) {
```
RadoslavGeorgiev: I might see the answer to my question somewhere else in the PR, but why is this a method of the gateway class? It makes sense as a stand-alone function, or as something more easily accessible, ex. `WC_Payments::get_order()`. I prefer either just `get_order()`, or if we want to indicate it's a WCPay one, let's make it `get_wcpay_order()`.

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+use \WC_Order;
+use WC_Payments_Utils;
+
+/**
+ * WC_Payments_Order
+ */
+class WC_Payments_Order {
```
RadoslavGeorgiev: Why not `WC_Pay\Core\Order`?

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+ * Class file for WCPay\Core\WC_Payments_Order
+ *
+ * @package WooCommerce Payments
+ */
+
+namespace WCPay\Core;
+
+defined( 'ABSPATH' ) || exit;
+
+use \WC_Order;
```
RadoslavGeorgiev: The backslash is not required here, `use` addresses the global namespace.

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	private $order;
+
+	/**
+	 * Constructor when order_id is provided.
+	 *
+	 * @param WC_Order|int $order The order Id.
+	 */
+	public function __construct( $order ) {
+		$this->order = wc_get_order( $order );
```
RadoslavGeorgiev: `wc_get_order` always constructs a new object, and I'm not sure we want that.

- On one hand it's good, because there will probably be no side effects.
- On the other hand, imagine this:

```php
$order = 'A WC_Order object, which we already have access to.';
$wcpay_order = new WC_Payments_Order( $order );
$wcpay_order->set_payment_intent_id( 'pi_XXX' ); // this is a public method, this can easily be done.
// Since there is no `save()` method of `WC_Payments_Order`, the only option is to save the original.
$order->save();
```

But this would accomplish nothing, since `WC_Payments_Order` only modified its internal instance, which was not saved. This example looks ugly though, as we're creating a second object, and saving the first one.

IMO we should require an `int $order_id` here, to make it obvious, that the received `$order` object will not be modified at any stage. If someone wants a shortcut, we could create a static `WC_Payments_Order::from_wc_order()` method.

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/**
+	 * Get the payment metadata for key _intent_id.
+	 *
+	 * @return string
+	 */
+	public function get_payment_intent_id() : string {
+		return $this->order->get_meta( '_intent_id', true );
```
RadoslavGeorgiev: This could be a great opportunity to introduce stronger types. I'd check if there is a non-empty string and return `string`, otherwise `null`.

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+
+
+	/**
+	 * Set the payment metadata for key _intent_id.
+	 *
+	 * @param  string $intent_id
+	 */
+	public function set_payment_intent_id( $intent_id ) {
+		$this->order->update_meta_data( '_intent_id', $intent_id );
+		 do_action( 'wcpay_payment_intent_id_updated' );
```
RadoslavGeorgiev: Let's [document actions](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters) to provide description.

Also, maybe including `order` in the action would make more sense, otherwise it looks like the payment intent is updated. So that would make it `wcpay_order_payment_intent_id_updated`. Combined with the method name change, it would turn to `wcpay_order_intent_id_updated`.

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,106 @@
...
+		return $this->order->get_meta( '_intent_id', true );
+	}
+
+
+	/**
+	 * Set the payment metadata for key _intent_id.
+	 *
+	 * @param  string $intent_id
+	 */
+	public function set_payment_intent_id( $intent_id ) {
```
RadoslavGeorgiev: This method is not used just for `PaymentIntent` objects, but also for `SetupIntent`, which is why the original method is called `attach_intent_info_to_order` instead of `attach_payment_intent_info_to_order`. With that in mind, let's use `set_intent_id` here.

tests/unit/core/test-class-wc-payments-order.php
```diff
@@ -0,0 +1,61 @@
...
+		$wcpay_order->set_payment_intent_id( $intent_id );
+		$this->assertEquals( $wcpay_order->get_wc_order()->get_meta( '_intent_id', true ), $intent_id );
+		$this->assertSame( 1, did_action( 'wcpay_payment_intent_id_updated' ) );
+		$this->assertSame( 0, did_action( 'wcpay_payment_method_id_updated' ) );
+	}
+
+	public function test_get_payment_intent_id() {
+		$intent_id   = 'pi_mock';
+		$wcpay_order = new WC_Payments_Order( WC_Helper_Order::create_order() );
+		$wcpay_order->get_wc_order()->update_meta_data( '_intent_id', $intent_id );
```
RadoslavGeorgiev: Minor difference, but let's call `update_meta_data` on `WC_Helper_Order::create_order()` before passing it to the constructor.

RadoslavGeorgiev (changes requested):
The approach looks good so far, let's see how it works in practice. I'm mostly curious how we'll deal with mocking the return of `wcpay_get_order` (or `get_wcpay_order` :p) in tests.

includes/core/class-wc-payments-order.php
jessy-p: Changed to int `$order_id`

includes/core/class-wc-payments-order.php
jessy-p: Fixed

tests/unit/core/test-class-wc-payments-order.php
jessy-p: Fixed

includes/core/class-wc-payments-order.php
jessy-p: Fixed

includes/core/class-wc-payments-order.php
jessy-p: Fixed

includes/core/class-wc-payments-order.php
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	/**
+	 * Set the payment metadata for key _payment_method_id.
+	 *
+	 * @param  string $payment_method_id The value to be set.
+	 */
+	public function set_payment_method_id( $payment_method_id ) {
+		$this->order->update_meta_data( '_payment_method_id', $payment_method_id );
+		$this->order->save_meta_data();
```
RadoslavGeorgiev: Let's see where this will be used, and if the order is saved in that context. This might uncover a bigger issue, which would be that the order object here gets modified, but the original order does not, or vice versa. Is that a problem?

includes/class-wc-payment-gateway-wcpay.php
jessy-p: Removed.

includes/core/class-wc-payments-order.php
jessy-p: Fixed

PR #5163, opened 2022-11-22
Prevent duplicate orders: Add intent key

RadoslavGeorgiev:
This PR introduces the concept of an "intent key", which is a uniquely generated ID, meant to be:

- associated with orders.
- associated with payment/setup intents.
- stored within the customer's session.

Some code paths in the plugin (incl. payment success, and UPE) have a tight connection between orders and intents, as the intent contains the order ID, and the order contains the intent ID, both as meta data. This connection is only fully established once the intent ID reaches the client, but errors and exceptions might sometimes prevent that from happening.

The purpose of the intent key is to have it generated before the API is ever contacted with a request. This will allow us to retrieve both intents and orders based on their intent key:

- Querying orders by key will allow us to check if there is an identical order with the same key during checkout, marked as paid by a webhook, based on a key stored in the customer's session. **This is the case this PR addresses**.
- Querying intents by key on the server will allow us to retrieve the object, which will be useful while on checkout on a site, which did not manage to receive a webhook for payment confirmation. Once/if implemented on the server (this would require an additional column), another PR here might build on top of this one.

#### How it works

- [  ] Switch from cookie to (WC?) session.
- [  ] This PR checks the cart, make it work for the pay for order page as well.
- [  ] Check for compatibility with UPE.

---

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -76,6 +76,13 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	const USER_FORMATTED_TOKENS_LIMIT = 100;
 
 	/**
+	 * The cookie name for storing intent keys.
+	 *
+	 * @var string
+	 */
+	const INTENT_COOKIE = 'wcpay_intent_key';
+
+	/**
 	 * Client for making requests to the WooCommerce Payments API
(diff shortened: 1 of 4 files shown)


htdat:
Closing in favor of https://github.com/Automattic/woocommerce-payments/pull/5247

PR #5139, opened 2022-11-16
WooPay express checkout button on blocks checkout

Mayisha:
#### Changes proposed in this Pull Request
Used `registerExpressPaymentMethod` to render the WooPay express checkout button on the blocks checkout and cart page. This PR only takes care of displaying the button. No functionality has been added yet.

#### Testing instructions
- Run `npm start`
- Install the latest [dev-tools](https://github.com/Automattic/woocommerce-payments-dev-tools). Go to the dev tools page and click the `Enable the WooPay Express Checkout button` checkbox and save the settings.
- Go to the WooCommerce Payments settings screen and click the `Customize` button of WooPay.
- Select `Cart` and `Checkout` as the button location and save the settings.
- Add `Checkout Block` on the checkout page and `Cart Block` on the cart page.
- Ensure that the WooPay button shows up on the blocks checkout screen as per the saved settings.
- Ensure that the WooPay button shows up on the blocks cart screen as per the saved settings.

PR #5164, opened 2022-11-22
Add WooPay Express Checkout functionality to shortcode and cart

bborman22:

As part of the effort to build a WooPay Express Checkout button, we have added the display of the button and the styling of the button. This PR is to add functionality to the buttons on the shortcode checkout and the cart pages. This does not impact the button block checkout or product page checkout. In short this PR implements the following changes:
- Display the WooPay iframe in a more generic centered position when triggered by the Express Checkout button.
- Allow the iframe to be triggered by a click on the React button compoenent.
- Pass the user's email address from the OTP verification to the session initialization.
- Redirect the user upon successful verification and session initialization.

This PR should be tested with https://github.com/Automattic/woopay/pull/1470 if it has not already been merged.

Some of the main implementation details:
- The iframe functionality was lifted and simplified from the existing `email-input-iframe.js` code. This simplification involved removing the `LoginSession` iframe flow, removing the popover arrow, and aligning the iframe with the email input field.
- The passing of the JS parameters proved to be quite tricky. There was already an existing `wcpayWooPayExpressParams` namespace that I ultimately ended up using. But what I found was that I needed a lot of the values that come from the [existing checkout fields](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-checkout.php#L95). The issue then was coming from the `getConfig` function pulling from the `wcpay_config` namespace. On the cart page I could define this JS namespace as `wcpay_config` and `getConfig` worked as expected. But on the checkout page `wcpay_config` was overwritten by the existing fields. So I ended up keeping the `wcpayWooPayExpressParams` namespace and then implementing a `getWooPayExpressData` function to be used in place of `getConfig` when necessary. This is copied from the `paymentRequest` model which I debated adding on to, but opted to create a parallel functionality instead.
- Passing of the email field from the OTP verification to the session initialization call rather than using the email input field from checkout (since we won't always have access to it from other pages).

**Known Issue**
I noticed that the `wcpayTracks` events were failing due to the same `getConfig` issue mentioned above. I'm breaking this out in to it's own [issue](https://github.com/Automattic/woopay/issues/1472).

#### Testing instructions

1. Be sure to be running WooPay with the https://github.com/Automattic/woopay/pull/1470 branch if it hasn't been merged yet.
2. Pull down this PR and compile JS.
3. Make sure the Express Checkout feature flag is enabled from the latest Dev Tools.
4. Make sure the Express Checkout button is set to display on Cart and Checkout page from Payments > Settings > WooPay Customize > Show express checkouts on.
5. Add a product to the cart.
6. Visit the cart, confirm display of WooPay button.
7. Click on WooPay button and confirm iframe loads with email input.
8. Enter email and verify OTP.
9. Confirm that session gets initialized and shopper is redirected to WooPay checkout.
10. Complete a full checkout and confirm successfully returned to merchant store with successful checkout.
11. Return to merchant store and add a product to the cart.
12. Visit the checkout page and confirm the display of the WooPay button.
13. Click on WooPay button and confirm iframe loads with email input.
14. Enter email and verify OTP.
15. Confirm that session gets initialized and shopper is redirected to WooPay checkout.
16. Complete a full checkout and confirm successfully returned to merchant store with successful checkout.

Would be good to also test a few other things:
- Confirm this is no regression on the existing WooPay checkout flow as these two paths should work in parallel while in development.
- Try out a few errors or failed login in the iframe and confirm all is handled properly on the merchant store.
- Try opening and closing the iframe with the button and make sure the DOM isn't getting cluttered and events are properly being unbound.

-------------------

**Post merge**


bborman22:
Changing the approach in this PR based on our roundtable discussion. Setting as draft until ready.

bborman22:
I have updated this PR to now conform to the updates in https://github.com/Automattic/woopay/pull/1470.

client/checkout/platform-checkout/express-button/utils.js
```diff
@@ -0,0 +1,16 @@
...
+ * Retrieves a configuration value.
+ *
+ * @param {string} name The name of the config parameter.
+ * @return {*}         The value of the parameter of null.
+ */
+export const getWooPayExpressData = ( name ) => {
+	// Config for the Edit Order screen.
+	const config =
+		wcpayWooPayExpressParams ??
+		wc.wcSettings.getSetting( 'woocommerce_payments_data' );
```
gpressutto5: Should we try to use `wcpay_config` before using `wc.wcSettings.getSetting` to keep the same logic we have in `getConfig`? Or maybe we could default directly to `getConfig` when wcpayWooPayExpressParams is falsy.

gpressutto5 (changes requested):
I know it was not plan, but do you think we can easily make this work for blocks checkout and cart page too? It fails with the error bellow because #5139 just got merged and it didn't include the api prop in the button:

https://github.com/Automattic/woocommerce-payments/blob/30a2e9e67bb710ffa04458e2fea329e179b791c7/client/checkout/platform-checkout/express-button/woopay-express-checkout-payment-method.js#L12-L14

image

client/checkout/platform-checkout/express-button/utils.js
Mayisha: I have a quick question here. From the description, I assume this function is inspired from [this](https://github.com/Automattic/woocommerce-payments/blob/add/1401-woopay-express-checkout-functionality/client/utils/checkout.js#L9-L17) or [this](https://github.com/Automattic/woocommerce-payments/blob/add/1401-woopay-express-checkout-functionality/client/utils/order.js#L9-L16).

My understanding is the `wcpay_config` or `wcpay_order_config` is present for classic shortcode cart/checkout and the `wc.wcSettings.getSetting( 'woocommerce_payments_data' )` is present on the Blocks cart/checkout.

In our case shouldn't this `wcpayWooPayExpressParams` be present on both shortcode/blocks pages? Do we need this fallback? Or does this fallback contain the correct data? :thinking: 

In case we need the fallback, I agree with @gpressutto5 that `wcpay_config` should be there too.

client/checkout/platform-checkout/express-button/express-checkout-iframe.js
```diff
@@ -0,0 +1,226 @@
...
+		parentDiv.insertBefore( iframeWrapper, null );
+
+		setPopoverPosition();
+
+		// Focus the iframe.
+		iframe.focus();
+	};
+
+	const showErrorMessage = () => {
+		parentDiv.insertBefore( errorMessage );
```
Mayisha, line 150: Definitely not as part of this PR but we should have a separate issue and consult with the design about the placement of the error message. Previously the OTP modal was triggered with the email field so it made sense to show the error message below the email field. But with the express checkout button which will even work with an empty email field, that place does not feel suitable anymore.

bborman22: Good call, this actually errors out for me in console. Created an issue to track this [here](https://github.com/Automattic/woopay/issues/1490).

bborman22:
> didn't include the api prop in the button

Fixed this with [a24ca4b](https://github.com/Automattic/woocommerce-payments/pull/5164/commits/a24ca4b64a716f365922b1596cfdeda7f4152883) and everything seems to work with blocks cart / checkout as well! So should be able to close https://github.com/Automattic/woopay/issues/1411 with this PR.

client/checkout/platform-checkout/express-button/utils.js
bborman22: I removed the fallback altogether from this function, since as @Mayisha mentioned that variable should be available in any scenario where the button is being loaded with functionality anyway. However, I did have a hard time with getting these JS params passed through in a way that made sense to me, so I'm very much open to feedback on better approaches for handling this rather than the myriad of fallbacks and different utility functions, even if it gets handled in a separate issue?

gpressutto5 (approved):
It's working great!

PR #5169, opened 2022-11-24
Warn merchant when deactivating WCPay plugin that WCPay Subscriptions (Stripe Billing )will continue renewing

### LINKED ISSUE #5090: Warn merchant when deactivating WCPay plugin that WCPay Subscriptions will continue renewing (offboarding)
### Description
WCPay Subscriptions use [Stripe Billing](https://stripe.com/docs/connect/subscriptions) to handle renewals. 

This means that the WooCommerce site delegates the renewals to the WCPay back end (i.e. Stripe infrastructure). If a merchant disables WCPay on their site, they currently need to manually pause or migrate subscriptions to prevent them from continuing to renew (charge the customer) via Stripe.

Note that this is documented here: https://woocommerce.com/document/payments/subscriptions/deactivate/#existing-subscriptions

However there may be more we can do to ensure merchants don't unintentionally leave WCPay Subscriptions active and accidentally double or unnecessarily charge their customers.

If merchants disable WCPay on their site, WCPay Subscriptions should stop renewing by default. 

There are a few scenarios we may want to consider:

- Merchant deactivating WCPay, e.g. switching to another payment gateway.
- Merchant shutting down site or business (possibly without deactivating plugin - worst case, just shut down WooCommerce service).

### Designs
TBD - this may require design for how the pause is handled. 

### Additional context
Further internal discussion, including support tickets from an affected merchant: pbUcTB-1DU-p2

haszari:
Fixes #5090

This PR adds a new modal if merchant clicks deactivate link for WCPay on plugins screen. If the store has active WCPay subscriptions, the modal will be shown. 

The purpose of the modal is to warn merchants that Stripe billing (off site) subscriptions will continue renewing automatically. 

### Screenshots
This modal is displayed if merchant clicks `Deactivate` for WooCommerce Payments plugin, with active WCPay built-in subscriptions.

Screenshot 2022-12-02 at 3 41 50 PM

Full wording for easy review:

```
Your store has active subscriptions using the built-in WooCommerce Payments functionality. Due to the [off-site billing engine](https://woocommerce.com/document/payments/subscriptions/comparison/#billing-engine) these subscriptions use, they will continue to renew even after you deactivate WooCommerce Payments. [Learn more](https://woocommerce.com/document/woocommerce-payments/built-in-subscriptions/deactivate/#existing-subscriptions).

If you do not want these subscriptions to continue to be billed, you should [cancel all subscriptions](https://woocommerce.com/document/subscriptions/store-manager-guide/#cancel-or-suspend-subscription) prior to deactivating WooCommerce Payments.
Are you sure you want to deactivate WooCommerce Payments?
```

## Testing instructions

**Basic setup**

0. Ensure your WooCommerce site is set up and the WooCommerce Subscriptions plugin is not installed or not active. 
1. Ensure WooCommerce Payments is onboarded and set up. You will need to use USA store location to enable WCPay Subscriptions. For full details refer to this flow: [Set up store for WCPay Subscriptions](https://github.com/Automattic/woocommerce-payments/wiki/WC-Pay-Subscriptions-critical-flows-testing-instructions#base)
2. [Create a WCPay subscription product](https://github.com/Automattic/woocommerce-payments/wiki/WC-Pay-Subscriptions-critical-flows-testing-instructions#create-a-subscription-product)
3. [Purchase WCPay subscription as shopper](https://github.com/Automattic/woocommerce-payments/wiki/WC-Pay-Subscriptions-critical-flows-testing-instructions#purchase-a-subscription-product)

You should now have an active WCPay Subscription.

**Deactivate WCPay Plugin** to see the warning modal & test deactivate/cancel flow

1. Go to `Dashboard > Plugins`.
2. Scroll down to `WooCommerce Payments`.
3. Click `Deactivate`. 
   - You should see the modal above.
4. Click `Cancel`
   - WCPay should still be active, zero impact on anything.
5. Repeat step 3 - click `Deactivate`.
6. Click `Yes, deactivate WooCommerce Payments` - you accept that subscriptions will continue renewing, you have a plan for how to handle that (e.g. reactivate plugin, this is temporary).
   - Plugin should deactivate as normal.

**Repeat test with no WCPay Subscriptions active**

Reactivate WCPay & ensure you have an active subscription (`Basic setup` above).

1. Go to `Dashboard > WooCommerce > Subscriptions`. 
2. Find active subscriptions - should have green `Active` status.
3. In status column, click `Suspend` link.
4. Repeat **`Deactivate WCPay Plugin`** test steps. The modal should not be displayed.

--- 

I followed the test steps above. I encourage reviewers to test other flow combinations, for example:

- Test with WooCommerce Payments active. The modal should display if there are any active WCPay subscriptions.
- Experiment with other WCPay Subscriptions states (statuses) that might be relevant. 

-------------------

- [x] Tested on mobile (or does not apply). I tested on Chrome responsive mode, to ensure the modal is readable and functions. Screenshot:

Screenshot 2022-11-28 at 9 58 28 AM

## Post merge

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions): _[**Testing instructions**](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.2.0#when-deactivating-wcpay-plugin-warn-merchant-that-active-wcpay-subscriptions-will-continue-collecting-payment---pr5169)_
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable. _n/a - this is not a critical flow_
- [x] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable. _Added to a new `What's New in 5.2` draft on WCPay P2._


includes/subscriptions/class-wc-payments-subscription-service.php
```diff
@@ -1004,4 +1004,26 @@ private function validate_subscription_data( $subscription_data ) {
...
+	 * @return bool True if store has active WCPay subscriptions, otherwise false.
+	 */
+	public static function store_has_active_wcpay_subscriptions() {
+		$results = wcs_get_subscriptions(
+			[
+				'subscriptions_per_page' => 1,
+				'fields'                 => 'ids',
+				'subscription_status'    => 'active',
+				// Ignoring phpcs warning, we need to search meta.
+				'meta_query'             => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
```
haszari, line 1019: Is there a way to avoid the `meta_query` here?

includes/subscriptions/templates/html-wcpay-deactivate-warning.php
```diff
@@ -0,0 +1,54 @@
...
+					<div class="inner">
+						<button id="wcpay-plugin-deactivate-modal-submit" class="button button-primary button-large"><?php esc_html_e( 'Yes, deactivate WooCommerce Payments', 'woocommerce-payments' ); ?></button>
+						<button class="modal-close button button-secondary button-large"><?php esc_html_e( 'Cancel', 'woocommerce-payments' ); ?></button>
+					</div>
+				</footer>
+			</section>
+		</div>
+	</div>
+	<div class="wc-backbone-modal-backdrop modal-close"></div>
+</script>
```
haszari, line 56: This template is a copy of the other warning modal. I thought that was simpler / more robust than parameterising so we have a single modal template.

includes/subscriptions/class-wc-payments-subscription-service.php
```diff
@@ -1004,4 +1004,26 @@ private function validate_subscription_data( $subscription_data ) {
...
+					self::SUBSCRIPTION_ID_META_KEY => 'EXISTS',
+				],
+			]
+		);
```
haszari, line 1026: Generally keen for feedback on this query and edge cases it might need to detect. For the modal it doesn't need to be 100% perfect; false positives are ok. But if this method gets used elsewhere later it's probably a good idea to be super clear about what this searches for, and any potential edge cases.

includes/subscriptions/assets/js/plugin-page.js
```diff
@@ -3,27 +3,76 @@ jQuery( function ( $ ) {
...
-			// Display modal
 			$( this ).WCBackboneModal( {
 				template: 'wcpay-subscriptions-plugin-warning',
 			} );
 
 			return false;
 		},
-		deactivate_subscriptions: function ( event ) {
+		// Trigger deactivate flow for WC Subscriptions.
+		redirect_deactivate_wc_subscriptions: function ( event ) {
```
haszari, line 75: Renaming these as I've taken a "duplicate the existing code" approach, rather than parameterise these functions (which IMO would be more complicated without much payoff - these functions are all one-liners).

haszari:
Requesting a review from @Automattic/helix FYI @mattallan 

Questions for reviewer:

- Should this be covered by e2e tests (or other automated tests)? I'm not sure if this is practical / worthwhile, keen for advice.
- Keen for advice on other ways to test this or edge cases to explore, especially around subscription statuses. I've tested with admin manual `Suspend` only.

Also FYI:

-  @aheckler - this links out to documentation, keen for your feedback on modal wording
- Sourav - FYI since you are aware of this issue/helped prioritise this fix

haszari:
@elizaan36 - FYI adding a new modal when disabling WCPay, to warn merchant that WCPay (built in) subscriptions will continue renewing. Let me know if you have any feedback on the modal or how this fits into the overall UI. Thanks!

aheckler:
@haszari Can we adjust the first two paragraphs of the modal to be the following?

```
Your store has active subscriptions using the built-in WooCommerce Payments functionality. Due to the [off-site billing engine](https://woocommerce.com/document/payments/subscriptions-basics/comparison/#billing-engine) these subscriptions use, **they will continue to renew even after you deactivate WooCommerce Payments.**

If you do not want these subscriptions to continue to be billed, you should [cancel all subscriptions](https://woocommerce.com/document/subscriptions/store-manager-guide/#cancel-or-suspend-subscription) prior to deactivating WooCommerce Payments.
```

elizaan36:
Thanks for the ping, @haszari !

Just a quick note on the modal component styles and CTA. 
- We usually have the primary CTA positioned on the right of the modals. 
- And on mobile, the primary CTA would be below the Cancel CTA. We also need to maintain the same padding between the buttons horizontally and vertically.

Last thing is the modal title ("Are you sure?") looks like the right size but Bold weight. This is the `subtitle` WordPress component and should be semi-bold.

Here's a quick sketch to illustrate the notes above.
Screen Shot 2022-11-28 at 10 31 45 AM

includes/subscriptions/class-wc-payments-subscription-service.php
mattallan: I don't think there's a way to avoid using meta_query here.

That said, this isn't the correct way to use the `'meta_query'` param.
It needs to be a nested inside an array for it to work.

For example:

```
$results = wcs_get_subscriptions(
	[
		'subscriptions_per_page' => 1,
		'subscription_status'    => 'active',
		'meta_query'             => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
			[
				'key'     => '_wcpay_subscription_id',
				'compare' => 'EXISTS',
			]
		],
	]
);
```

includes/subscriptions/class-wc-payments-subscription-service.php
```diff
@@ -1004,4 +1004,26 @@ private function validate_subscription_data( $subscription_data ) {
...
+	/**
+	 * Determines if the store has any active WCPay subscriptions.
+	 *
+	 * @return bool True if store has active WCPay subscriptions, otherwise false.
+	 */
+	public static function store_has_active_wcpay_subscriptions() {
+		$results = wcs_get_subscriptions(
+			[
+				'subscriptions_per_page' => 1,
+				'fields'                 => 'ids',
```
mattallan: `'fields'` isn't a valid param for `wcs_get_subscriptions()` so this can be removed. This function always returns the object.

includes/subscriptions/class-wc-payments-subscription-service.php
mattallan: I don't have any feedback on this. I think this approach will catch 99.99999% which I'm fine with :D There will always be edge cases but the nice thing about having this code in a separate function is if we do discover edge cases we can update this query or add more queries

mattallan (commented):
I've done some testing on this PR, mainly focusing on the new `store_has_active_wcpay_subscriptions()` function.

From my testing, this function was always returning `true` even when I didn't have a single subscription with `_wcpay_subscription_id` meta.

I've provided some feedback on how to fix this as part of my review (slight issue with how the meta query param was used)

With my suggested changes, I've tested this function will work in HPOS and CPT environments and also with older versions of WC Subscriptions and the latest version of subscriptions bundled in WC Payments.

includes/subscriptions/class-wc-payments-subscription-service.php
haszari: Thanks for explaining, I had no idea how this was supposed to work (I did read the docs for wcs_get_subscriptions!).

Once I've fixed up the PR I'll add something to the docblock (meta_query is not mentioned).

haszari: Just retested with this logic in place. I see now the flaw in my previous testing – I needed to test with WCS / tokenised subscriptions active. Since the `meta_query` wasn't doing anything I was getting false positives – warning shown when merchant only has active tokenised subs. Good catch 🏀

haszari:
Addressed reviewer feedback. @mattallan thanks for the tip about `meta_query`, I'm keen to document that while we're here (tbc).

@mattallan I did some more testing with WCS active and inactive, and a couple of subscriptions (one tokenised, one stripe billing). If you could retest with a couple of interesting scenarios I'd appreciate it :) 

Here's the current state of the modal, incorporating new wording (@aheckler) and revised default button position (on right) (@elizaan36).

Screenshot 2022-12-02 at 11 29 09 AM

---

Note - I haven't fixed the other styling issues picked up by @elizaan36 - title weight, button layout on mobile. 

These styles come from WooCommerce core, and as far as I can tell, our markup is correct. So if we want to improve/fix the title or button layout issues, I think that should be logged as a WC core issue. I checked a core modal and see that it also has a bold title.

Screenshot 2022-12-02 at 11 26 29 AM

includes/subscriptions/class-wc-payments-subscription-service.php
haszari: I saw that eventually `wc_get_orders` was called, so assumed the args would carry through. Fixed in 14ef0a4974d1e21ff79202c30fd58bd03f7bff07

aheckler:
@haszari Two minor changes:

1. Can we move the `Learn more` link to be after the first paragraph, rather than the second one? I think that's a more natural place for it.
2. In the `Learn more` link, the period/full stop is also part of the link. Usually, punctuation is not included in links like that. Can we move the period/full stop outside the link?

Thanks!

haszari:
Thanks @aheckler  - good spotting! I've fixed that up now:

Screenshot 2022-12-02 at 3 41 50 PM


```
Your store has active subscriptions using the built-in WooCommerce Payments functionality. Due to the [off-site billing engine](https://woocommerce.com/document/payments/subscriptions/comparison/#billing-engine) these subscriptions use, they will continue to renew even after you deactivate WooCommerce Payments. [Learn more](https://woocommerce.com/document/woocommerce-payments/built-in-subscriptions/deactivate/#existing-subscriptions).

If you do not want these subscriptions to continue to be billed, you should [cancel all subscriptions](https://woocommerce.com/document/subscriptions/store-manager-guide/#cancel-or-suspend-subscription) prior to deactivating WooCommerce Payments.
Are you sure you want to deactivate WooCommerce Payments?
```

haszari:
Ready for a final review @mattallan / @brucealdridge  🙇

mattallan (approved):
Really nice additions coming in this PR @haszari !!

I have tested the following cases against your latest changes:
 - [x] Store with one or many active WC Pay Subscriptions and no other payment methods. Correctly shows the message on deactivation ✅ 
 - [x] Store with one or many canceled WC Pay Subscriptions. No notice is shown. ✅ 
 - [x] Store with only active WC Payments (tokenized) and no stripe billing subscriptions. No notice is shown. ✅ 
 - [x] Store with some active WC Payments (tokenized) and some active stripe billing subscriptions. Notice is correctly shown. ✅ 
 - [x]  Store with some active WC Payments (tokenized) and some cancelled stripe billing subscriptions. No notice is shown. ✅ 
 - [x] Tested clicking "Yes, Deactivate WooCommerce Payments" actually does deactivate the plugin ✅ 

The notice/warning also looks good to me (I double checked the links as well):

image

If there are no more design changes/feedback this is good to merge :) Marking as approved.

haszari:
Thanks for the thorough testing @mattallan 👌

includes/subscriptions/class-wc-payments-subscription-service.php
haszari: I added a follow up issue to document `meta_query`: 

- https://github.com/Automattic/woocommerce-subscriptions-core/issues/337

PR #5172, opened 2022-11-24
New workflow to amend release changelog

### LINKED ISSUE #5126: [Release automation] Workflow to amend a changelog
### Description
The [code freeze workflow](https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/release-code-freeze.yml) takes care of preparing the release on Sunday Week 2:
- a release branch `release/x.y.z` is created, with a commit for the required changes (bump versions)
- a PR targeting trunk is created

In case we detect an issue, or we want to include a critical feature after the code freeze, we will need to update the changelog too.

We could do it by hand:
- Execute `./vendor/bin/changelogger write --use-version="x.y.z" --release-date="YYYYMMDD" --amend` on the release branch to amend the changelog that was generated by the code freeze workflow in `changelog.txt`
- Copy/paste the new entries in `readme.txt`

But we could also have a GitHub workflow to do that and save us the trouble.

jbordonado:
Fixes #5126

- Add a new workflow `release-changelog.yml`
- Update the composite action `generate-changelog` to be able to either generate or amend the changelog entries, depending on a new input

The new workflow allows us to generate or amend the changelog entries and push the changes to a branch. I expect that this will only be used to amend the entries, but I still let the possibility to generate them as it didn't complicate things at all.
This can be used whenever one or several commits are added to the release branch after the official release code freeze, or during a patch release if needed.

#### Testing instructions

- Create and push a dummy branch to the repo, using that branch (so that it contains the new workflow, otherwise you won't be able to test it)
- Run the workflow to generate the changelog: `gh workflow run release-changelog.yml --ref <DUMMY_BRANCH> -f release-version="5.1.0" -f action-type="generate"`
- You should see a new commit on the dummy branch with the following changes:
  - All changelog entries under the `changelog` folder were inserted into `changelog.txt` under a new block for the release version provided, and those new entries were deleted from the folder too
  - The same changelog entries were pasted in `readme.txt`
  - The release version and release date are correct in both files w.r.t the inputs passed to the workflow (release date is set to "today" per default)

- Now, create one or several new changelog entries on the dummy branch with `npm run changelog`
- Run the workflow to amend the last changelog: `gh workflow run release-changelog.yml --ref >DUMMY_BRANCH> -f release-version="5.1.0" -f release-date="tomorrow" -f action-type="amend"`
- You should see a new commit on the dummy branch with the following changes:
  - All changelog entries under the `changelog` folder were inserted into the last changelog that was generated earlier, and those new entries were deleted from the folder too
  - The same new changelog entries were pasted in `readme.txt` under the last changelog that was generated earlier
  - The release version and release date are correct in both files w.r.t the inputs passed to the workflow (especially the date, which is intentionally different from the one we used to generate)

-------------------

**Post merge**


.github/workflows/release-changelog.yml
```diff
@@ -0,0 +1,81 @@
...
+            sed -i "s|$OLD_CHANGELOG|$CHANGELOG|" readme.txt
+          else
+            sed -ri "s|(== Changelog ==)|\1\n\n= $RELEASE_VERSION - $RELEASE_DATE =\n$CHANGELOG|" readme.txt
+          fi
+                    
+      - name: "Commit and push the changes"
+        run: |
+          git config user.name "${{ github.actor }}"
+          git config user.email "${{ github.actor }}@users.noreply.github.com"
+          git commit -am "Process changelog entries for release $RELEASE_VERSION"
```
deepakpathania, line 80: Could we have different commit messages for `generate` and `amend` here?

deepakpathania: Could we have different commit messages for `generate` and `amend` here?

.github/workflows/release-pr.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       - name: "Generate the changelog"
         id: generate_changelog
-        uses: ./.github/actions/generate-changelog
+        uses: ./.github/actions/process-changelog
```
deepakpathania, line 64: Should we have a retry for failure for these? I did get an email from GH for the failure but I am trying to understand what's the ideal way for the release lead to know something isn't right in case of a failure and then acoid proceeding without noticing this.

deepakpathania (changes requested):
Left minor comments. Apart from the already discussed missing README update in case of amending, things LGTM.

.github/workflows/release-changelog.yml
jbordonado: Done

.github/workflows/release-pr.yml
jbordonado: I don't think we need to have that.

`release-pr` should run once every 3 weeks just before the final week during which we release WCPay. I don't see any reason for it to fail at this step unless there isn't any changelog to generate (very unlikely after 3 weeks without releasing). 
Moreover, the release lead has to ask for a review for the PR from the release branch targeting trunk after that, and I'm confident it would be caught if neither the changelog nor the version (the step after that) is bumped across multiple files. The [doc](https://href.li/?https://fieldguide.automattic.com/developing-woocommerce-payments/client-deployment#prepare-release) (point 4) explicitly says to check for this when looking at the PR (which, by the way, would not be created automatically by the workflow in this case since this is the last step, so this alone should also make the release lead aware that something's weird).

jbordonado:
I fixed the amend flow and tweaked the commit message depending on the type of flow we're executing.

Testing on `test/amend-workflow`:

- First run to generate the changelog for 5.2.0 for tomorrow: https://github.com/Automattic/woocommerce-payments/actions/runs/3548252223
- Commit created from this: https://github.com/Automattic/woocommerce-payments/commit/e3f4ef5d2e6604e0e85b613ebaec54d1d275bf60

- Second run to amend the 5.2.0 changelog for today: https://github.com/Automattic/woocommerce-payments/actions/runs/3548262228
- Commit created from this: https://github.com/Automattic/woocommerce-payments/commit/cc9614bf974aa8f8f3773c02606d4b8666b98209

.github/workflows/release-changelog.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          release-version: ${{ env.RELEASE_VERSION }}
+          release-date: ${{ steps.format_date.outputs.RELEASE_DATE }}
+          action-type: ${{ env.CHANGELOG_ACTION }}
+        
+      - name: "Update readme.txt"
+        env:
+          RELEASE_DATE: ${{ steps.format_date.outputs.RELEASE_DATE }}
+          CHANGELOG: ${{ steps.process_changelog.outputs.changelog }}
+          OLD_CHANGELOG: ${{ steps.get_old_changelog.outputs.OLD_CHANGELOG }}
```
deepakpathania: Is this still needed?

jbordonado: Nice catch, I removed it

deepakpathania (approved):
Thanks for working on this, LGTM. ✅

PR #5191, opened 2022-11-25
Hide remember details checkbox on returning from platform checkout.

Mayisha:
When a user returns to the merchant's checkout page by signout or guest checkout or back button, the `Remember your details` checkbox is visible there though this is a registered WooPay user. This PR fixes the issue by hiding the checkbox on returning from WooPay.

#### Testing instructions
- Add some products to the cart, and go to the checkout page.
- Use an email having a corresponding WooPay account.
- When OTP modal appears notice that the `Remember your details` checkbox is hidden.
- Complete OTP verification and go to platform checkout.
- From the `Contact` section click `Sign out` or `Checkout as guest` to return to the merchant checkout page.
- Ensure that the `Remember your details` checkbox is hidden.


bborman22 (approved):
This all worked great for me on blocks checkout. I wasn't sure if this PR should account for the shortcode checkout or not? But for blocks checkout LGTM!

Mayisha:
> I wasn't sure if this PR should account for the shortcode checkout or not

@bborman22 it is supposed to work on both shortcode and blocks checkout. Is it not working on the shortcode for you? :eyes:

bborman22:
Yeah it wasn't working on the shortcode checkout for me. Just to give the steps I followed to be sure. I visit checkout, see the remember details checkbox, enter email, verify OTP, visit WooPay, click Back (and I see the URL param is being set), and still see the remember details checkbox. For blocks checkout, the checkbox gets removed. 

I noticed that `isRegisteredUser` goes from false to true to false. So I think the order of the events dispatching is messing it up. My containers have been running quite slow the past two days, so I'm wondering if that is making the difference since the user exist call is taking longer for me and so dispatching after the back button dispatch?

Mayisha:
Thanks @bborman22 

I will check this again and see if I can recreate the issue on the shortcode. For me, it was working for both blocks and shortcode. But I also noticed my docker being really really slow for the past couple of days :thinking: 

> so I'm wondering if that is making the difference since the user exist call is taking longer for me and so dispatching after the back button dispatch.

This I don't think should be the reason as we don't make any `user-exist` call if the user came here via the back button. I will try to figure out the cause here :eyes:

Mayisha:
I have checked again and it is working in my local for shortcode checkout. I have increased the timeout in [27d1128](https://github.com/Automattic/woocommerce-payments/pull/5191/commits/27d1128df5de32da1e5be91eaf212a7af3247aa0) to be safe though. I will merge this now. :nerd_face:

Issue #4907, opened 2022-10-10
Multiple UPE - Resolve failing tests and checks


FangedParakeet:
### Description
The unit tests on the current working branch have been failing for quite a long time: I didn't update or create any new unit tests when I first created the POC branch and I'm not too sure that any have been updated since. Nonetheless the split-UPE fundamentally changes the checkout layout and the ways in which the payment gateway classes are utilised.

We need to update these tests and ensure that are no failures before we can eventually merge this branch back into `develop`. If we could be more confident in our testing environment, I think we could all sleep a little sounder in the knowledge that most user-flows and edge-cases have some form of acceptable test coverage.

### Acceptance criteria

* Ensure no more test failures on GH.

N.B. The `_wcpay_feature_upe` feature flag will soon be retired, since in the split-UPE approach there will not need to be a distinction between legacy and "new" checkout experiences (see #4724). Currently, existing tests are generally split into separate test-cases for when the UPE is enabled and disabled. We will need to resolve and consolidate these diverging tests into compatible tests for our combined single checkout experience.

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/4907?planning-poker) with Zenhub @mdmoore @timur27

timur27:
Thanks for creating the issue, @FangedParakeet! 🚀 
Are we talking about e2e tests failing ([1](https://github.com/Automattic/woocommerce-payments/actions/runs/3242683626/jobs/5316304675), [2](https://github.com/Automattic/woocommerce-payments/actions/runs/3242683626/jobs/5316305260), [3](https://github.com/Automattic/woocommerce-payments/actions/runs/3242683626/jobs/5316305120), [4](https://github.com/Automattic/woocommerce-payments/actions/runs/3242683626/jobs/5316304990) and [5](https://github.com/Automattic/woocommerce-payments/actions/runs/3242683626/jobs/5316304816)), or there are unit tests which need to be aligned? I didn't find anything besides the E2E tests failing on [the branch](https://github.com/Automattic/woocommerce-payments/pull/3968), which is why I wanted to double-check. 

Thank you!

FangedParakeet:
[...]

Issue #5116, opened 2022-11-11
Multiple UPE - Cannot add card payment method


mdmoore:
### Describe the bug
When adding a credit card as a payment method at **My Account > Add Payment Method**, the following error occurs: 

`A WooCommerce Payments payment method was not provided`

The console sheds some light on what's gone wrong: 

```
Uncaught TypeError: gatewayUPEComponents[_paymentMethodType6] is undefined
    <anonymous> upe.js:744
    jQuery 2
upe.js:744:4
    <anonymous> upe.js:744
    jQuery 2
```

Since we'll now always use the legacy card gateway,  `'card'` is no longer included as a UPE payment method [here in the `payment_methods_config`](https://github.com/Automattic/woocommerce-payments/blob/644443233bb99343d903a613ad33d1a3c4b71c65/includes/payment-methods/class-upe-payment-gateway.php#L1095-L1097). 

### To Reproduce

1. Enable UPE
2. Navigate to **My Account > Payment Methods > Add Payment Method**
3. Attempt to add a credit card

### Acceptance Criteria

- Credit cards can be added without any errors
- Changes made should be merged into `poc/upe-instances-multiplied`, not `develop`.

mdmoore:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5116?planning-poker) with Zenhub @FangedParakeet @timur27

FangedParakeet:
It looks like we've diagnosed the origin of the disease here and this is an area (saved payment methods) that we've worked on pretty closely in recent issues, which is why I voted for a mere `2` points. I can see how perhaps if there is a deeper unseen cause at play, the work here could bloat. 

@mdmoore @timur27 can either of you be budged to round up to `3` or drop down to `2` on the rating here?

timur27:
We now have a `2` SP estimation [here](https://github.com/Automattic/woocommerce-payments/issues/5117#issuecomment-1322435701) - and there is much more information on the solution there, in my opinion. Hence, comparing these two in terms of _knowns_ and _unknowns_, I would round up my estimation here to `3` SP.

PR #5205, opened 2022-11-28
Increase the test coverage and fix failing tests

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/4907

* [Fix the if-statement](https://github.com/Automattic/woocommerce-payments/pull/5205/files#diff-ad7970a483b7b3582e959f6236b3c69739d80349cb2236536c7db538726ffe2fR172) that was the result of adding token twice for the non-gateway specific requests + simplify the statement. This fixed subscriptions tests.
* Introduce e2e [tests for Blocks checkout](https://github.com/Automattic/woocommerce-payments/pull/5205/files#diff-b49b9c008428ea968ee3257af76dcab68b13408a25fc2389e206bb3d8ed03114R29) with UPE methods turned on
* [Align the shortcode checkout tests](https://github.com/Automattic/woocommerce-payments/pull/5205/files#diff-0bcfb6ae8e7247b36260ac8358cb9a0e84a992343372f0f329d7d05974186755R26) to be executed with UPE enabled
* [Test adding payment methods on My Account page](https://github.com/Automattic/woocommerce-payments/pull/5205/files#diff-5db1dc5474f9c3e4365bf9d1f5298c30c4f97744a0d10aacca234894dcaf2318R21) with turned on UPE
* Remove a not used and duplicated [shortcode UPE test](https://github.com/Automattic/woocommerce-payments/pull/5205/files#diff-6044e75da9acf8b9aa190af4b6e96a71c959400ced7d96cba537736cc7e64bd6)
* Cover token management and payment method(s) parameters handling areas with unit tests

> 💡 Note: After I added some new logic to the `shopper-myaccount-payment-methods-add-delete-success.spec.js` test, it's now failing because of https://github.com/Automattic/woocommerce-payments/issues/5116. After it will be fixed, the test will pass.

> ❕ Note: After https://github.com/Automattic/woocommerce-payments/issues/4724 is done, only `await merchantWCP.activateUpe();` lines should be deleted as this function enables UPE, which won't be the case anymore. E2E should not be blocked after we remove the `activateUpe()` invocation.

#### Testing instructions
* Run `npm run test:e2e` and confirm all tests, but `shopper-myaccount-payment-methods-add-delete-success.spec.js` (see the note above) are green.

-------------------

**Post merge**


includes/class-wc-payments-token-service.php
```diff
@@ -166,13 +169,7 @@ public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gat
...
 				continue;
 			}
-
-			if ( ! isset( $stored_tokens[ $payment_method['id'] ] ) && (
-					( Payment_Method::CARD === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID === $gateway_id ) ||
-					( Payment_Method::SEPA === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA === $gateway_id ) ||
-					( Payment_Method::LINK === $payment_method['type'] && WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::LINK === $gateway_id ) ) ||
-					empty( $gateway_id )
-				) {
+			if ( ! isset( $stored_tokens[ $payment_method['id'] ] ) && ( self::REUSABLE_GATEWAYS_BY_PAYMENT_METHOD[ $payment_method['type'] ] === $gateway_id || empty( $gateway_id ) ) ) {
```
FangedParakeet, line 172: Nice clean refactor here! I'm trying to think of a situation where the `$payment_method['type']` might not be set in this array, but I'm pretty sure there isn't one. There is actually an ugly, little edge case where the `type` property on the payment method isn't necessarily the original payment method type. 

For example, an iDEAL payment method can be saved, though it will be stored as a SEPA payment method and have the `type` value of `sepa_debit`. However, We would like to associate these tokens in WC with the iDEAL gateway instead of the SEPA gateway. There is a deeper, nested value that will reveal the true origins of the converted payment method...

...But none of those convertible payment methods are currently enabled as reusable payment methods, so for now, this issue will never arise! Although, it is something to remain aware of should we one day enable that feature for all reusable payment methods.

tests/e2e/specs/wcpay/shopper/shopper-checkout-save-card-and-purchase.spec.js
```diff
@@ -8,31 +8,45 @@ const { shopper } = require( '@woocommerce/e2e-utils' );
...
+
+	afterAll( async () => {
+		await shopperWCP.changeAccountCurrencyTo( 'USD' );
+		await shopperWCP.logout();
+		await merchant.login();
+		//disable SEPA
+		await merchantWCP.disablePaymentMethod( sepaPaymentMethod );
+		await merchantWCP.deactivateUpe();
+		await merchant.logout();
+	} );
```
FangedParakeet: This is a comment more than a suggestion, but I suppose in these E2E tests, where saved payments are involved, the tests begin by enabling the UPE; whereas in the other traditional checkout tests we do not enable the UPE before each test. I find this fine, because the saved payments flows are the ones where I might foresee the most frictions between the legacy and UPE code. 

However, might it be beneficial to have additional mirrored copies of these tests, so that we could run all the checkout and saved payment flow E2E tests with the UPE both disabled and enabled? Let me know if there is some limitation or greater wisdom that led you separate the tests as such.

FangedParakeet (approved):
Top job here, mate. These tests took me an age to even set up and run locally, so I can only imagine that tweaking the tests and ensuring their success must have cost an exponential amount of additional strain. I've left a few comments, which are generally just philosophical ponderings I have on a few premises, but everything looks fine to me and appears to be running on the GH action happily.

All in all, LGTM. I give this my blessings and am a considerably comforted to now have a little more reinforcing test coverage for our UPE use flows. 🚢

tests/e2e/specs/blocks/shopper/shopper-wc-blocks-saved-card-checkout-and-usage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+} from '../../../utils';
+
+const billingDetails = config.get( 'addresses.customer.billing' );
+const productName = config.get( 'products.simple.name' );
+const card = config.get( 'cards.basic' );
+
+import { fillCardDetailsWCB } from '../../../utils/payments';
+
+describeif( RUN_WC_BLOCKS_TESTS )(
+	'WooCommerce Blocks > Saved cards with enabled UPE methods',
```
FangedParakeet: Are these tests still for the enabled UPE methods? Tests look fine, but does this little description need a quick update?

FangedParakeet (approved):
Wonderful work here and cheers for taking the extra time to separate out the UPE tests at the end. I couldn't get the all the green lights to go off consistently locally, but such is life with these E2E tests. Most importantly, they appear to be running happily here on GitHub, so that's what counts the most.

I think these tests will really go a long way towards building confidence in releasing our feature branch, so nice one. I bless this with all my heart. LGTM. 🚢

PR #5223, opened 2022-12-01
Fix fatal error when hidden Blocks input added from CRON

FangedParakeet:
Please see p1669899593524019-slack-CGGCLBN58 for more details. 

#### Changes proposed in this Pull Request

# Fix fatal error when hidden Blocks input added from CRON

Ensure that hidden Blocks input is only added when `the_content` filter is fired from front-end and WC Session is available. 

#### Testing instructions

See @jessepearson's [instructions below](https://github.com/Automattic/woocommerce-payments/pull/5223#issuecomment-1334480617) to simulate the issue and test this patch. 

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #5228, opened 2022-12-02
Fix fatal error when hidden Blocks input added from CRON (#5223)

jbordonado:
Patch to fix fatal error that occurs when a scheduled action calls `the_content` filter, creating a fatal error, because this inserted Blocks hidden input relies on absent WC session data.

It was merged into develop (see https://github.com/Automattic/woocommerce-payments/pull/5223) but we need this fix into trunk to start the release PR creation


htdat (approved):
LGTM!

PR #5235, opened 2022-12-02
Implementation of Intent request classes

### LINKED ISSUE #5057: WCPay Platform: Base API request Intentation and Charge request
### Description
Add Intentation and Charge endpoints using WcPay Request class.

### Acceptance criteria
Before you start with issue, make sure to check this [issue](https://github.com/Automattic/woocommerce-payments/issues/5056) before. When this issue is merged, use the new classes that were added by this issue and create `Intentation` and `charge` requests using the base classes as foundations. 

### Testing instructions

Cover all requests with unit tests.

### Dev notes

1. Make sure that every request have getters and setters
2. Make sure that every argument or property is validated
3. Make sure immutability of properties that doesn't need to be changed.
4. Cover everything with tests.

zmaglica:
Fixes #5057

The following requests have been added using request class:

- Create_Intent
- Update_Intent
- Get_Intent

The examples can be found below.

To start with this request, create a new Create Intent request and fill the data.

Example:

**Create intent**

```php
$currency                 = strtolower( $order->get_currency() );
$$customer_id             = $request->get_param( 'customer_id' );
$metadata                 = $request->get_param( 'metadata' ) ?? [];
$metadata['order_number'] = $order->get_order_number();

$wcpay_server_request = Create_Intent::create();
$wcpay_server_request->set_currency_code( $currency );
$wcpay_server_request->set_amount( WC_Payments_Utils::prepare_amount( $order->get_total(), $currency ) );
if ( $customer_id ) {
	$wcpay_server_request->set_customer( $customer_id );
}
$wcpay_server_request->set_metadata( $metadata );
$wcpay_server_request->set_payment_method_types( $this->get_terminal_intent_payment_method( $request ) );
$wcpay_server_request->set_capture_method( 'manual' === $this->get_terminal_intent_capture_method( $request ) );
$intent = $wcpay_server_request->send( 'create_wcpay_terminal_intent_request', $order );
```
The `create_wcpay_terminal_intent_request` filter have been added where someone could use it according to their needs.

**Update intent**

```php
$request = Update_Intention::create();
$request->set_intent_id( $payment_intent_id );
$request->set_currency_code( strtolower( $currency ) );
$request->set_amount( WC_Payments_Utils::prepare_amount( $amount, $currency ) );
$request->set_metadata( $this->get_metadata_from_order( $order, $payment_type ) );
$request->set_level3( $this->get_level3_data_from_order( $order ) );
$request->set_selected_upe_payment_method_type( (string) $selected_upe_payment_type, $this->get_payment_method_ids_enabled_at_checkout( null, true ) );
if ( $payment_country ) {
	$request->set_payment_country( $payment_country );
}
if ( true === $save_payment_method ) {
	$request->setup_future_usage();
}
if ( $customer_id ) {
	$request->set_customer( $customer_id );
}
$updated_payment_intent = $request->send( 'wcpay_update_intention_request', $order, $payment_intent_id );
```

The `wcpay_update_intention_request` filter have been added where someone could use it according to their needs. In this example, order and `intent_id` have been passed.

**Get intent**

```php
$request = Get_Intent::create();
$request->set_intent_id( $intent_id );
$intent = $request->send( 'wcpay_get_intent_request', $order );
```

The `wcpay_get_intent_request` filter have been added where someone could use it according to their needs. In this example, order have been passed.

For get and update intents requests, the intent id can only be set once to prevent possible issues with data tempering.

#### Testing instructions

Make sure that terminal create intent endpoint works correctly.
-->

*

-------------------

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -367,7 +367,7 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			$wcpay_server_request->set_metadata( $metadata );
 			$wcpay_server_request->set_payment_method_types( $this->get_terminal_intent_payment_method( $request ) );
 			$wcpay_server_request->set_capture_method( 'manual' === $this->get_terminal_intent_capture_method( $request ) );
-			$intent = $wcpay_server_request->send( 'create_wcpay_terminal_intent_request', $order );
+			$intent = $wcpay_server_request->send( 'create_wcpay_intent_request', $order );
 
 			return rest_ensure_response(
 				[
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -10,7 +10,6 @@ defined( 'ABSPATH' ) || exit;
 use WCPay\Constants\Payment_Method;
 use WCPay\Core\Server\Request\Create_Intention;
 use WCPay\Core\Server\Request\Get_Intention;
-use WCPay\Exceptions\API_Exception;
 use WCPay\Logger;
 
 /**
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -8,6 +8,8 @@
 defined( 'ABSPATH' ) || exit;
 
 use WCPay\Constants\Payment_Method;
+use WCPay\Core\Server\Request\Create_Intent;
+use WCPay\Exceptions\API_Exception;
 use WCPay\Logger;
 
 /**
@@ -350,16 +352,28 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 		if ( false === $order ) {
 			return new WP_Error( 'wcpay_missing_order', __( 'Order not found', 'woocommerce-payments' ), [ 'status' => 404 ] );
 		}
-
 		try {
-			$result = $this->gateway->create_intent(
-				$order,
-				$this->get_terminal_intent_payment_method( $request ),
-				$this->get_terminal_intent_capture_method( $request ),
-				$request->get_param( 'metadata' ) ?? [],
-				$request->get_param( 'customer_id' )
+			$currency                 = strtolower( $order->get_currency() );
+			$customer_id              = $request->get_param( 'customer_id' );
+			$metadata                 = $request->get_param( 'metadata' ) ?? [];
+			$metadata['order_number'] = $order->get_order_number();
+
+			$wcpay_server_request = Create_Intent::create();
+			$wcpay_server_request->set_currency_code( $currency );
+			$wcpay_server_request->set_amount( WC_Payments_Utils::prepare_amount( $order->get_total(), $currency ) );
+			if ( $customer_id ) {
+				$wcpay_server_request->set_customer( $customer_id );
+			}
+			$wcpay_server_request->set_metadata( $metadata );
+			$wcpay_server_request->set_payment_method_types( $this->get_terminal_intent_payment_method( $request ) );
+			$wcpay_server_request->set_capture_method( 'manual' === $this->get_terminal_intent_capture_method( $request ) );
+			$intent = $wcpay_server_request->send( 'create_wcpay_terminal_intent_request', $order );
+
+			return rest_ensure_response(
+				[
+					'id' => ! empty( $intent ) ? $intent->get_id() : null,
+				]
 			);
-			return rest_ensure_response( $result );
 		} catch ( \Throwable $e ) {
 			Logger::error( 'Failed to create an intention via REST API: ' . $e );
 			return new WP_Error( 'wcpay_server_error', __( 'Unexpected server error', 'woocommerce-payments' ), [ 'status' => 500 ] );
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -141,9 +141,9 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$request = Get_Intention::create();
-			$request->set_intent_id( $intent_id );
-			$intent                   = $request->send( 'wcpay_get_intent_request', $order );
+			$request = Get_Intention::create( $intent_id );
+			$intent  = $request->send( 'wcpay_get_intent_request', $order );
+
 			$intent_metadata          = is_array( $intent->get_metadata() ) ? $intent->get_metadata() : [];
 			$intent_meta_order_id_raw = $intent_metadata['order_id'] ?? '';
 			$intent_meta_order_id     = is_numeric( $intent_meta_order_id_raw ) ? intval( $intent_meta_order_id_raw ) : 0;
@@ -246,9 +246,8 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$request = Get_Intention::create();
-			$request->set_intent_id( $intent_id );
-			$intent = $request->send( 'wcpay_get_intent_request', $order );
+			$request = Get_Intention::create( $intent_id );
+			$intent  = $request->send( 'wcpay_get_intent_request', $order );
 
 			$intent_metadata          = is_array( $intent->get_metadata() ) ? $intent->get_metadata() : [];
 			$intent_meta_order_id_raw = $intent_metadata['order_id'] ?? '';
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -9,6 +9,7 @@ defined( 'ABSPATH' ) || exit;
 
 use WCPay\Constants\Payment_Method;
 use WCPay\Core\Server\Request\Create_Intent;
+use WCPay\Core\Server\Request\Get_Intent;
 use WCPay\Exceptions\API_Exception;
 use WCPay\Logger;
 
@@ -140,7 +141,9 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$intent                   = $this->api_client->get_intent( $intent_id );
+			$request = Get_Intent::create();
+			$request->set_intent_id( $intent_id );
+			$intent                   = $request->send( 'wcpay_get_intent_request', $order );
 			$intent_metadata          = is_array( $intent->get_metadata() ) ? $intent->get_metadata() : [];
 			$intent_meta_order_id_raw = $intent_metadata['order_id'] ?? '';
 			$intent_meta_order_id     = is_numeric( $intent_meta_order_id_raw ) ? intval( $intent_meta_order_id_raw ) : 0;
@@ -243,7 +246,10 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$intent                   = $this->api_client->get_intent( $intent_id );
+			$request = Get_Intent::create();
+			$request->set_intent_id( $intent_id );
+			$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 			$intent_metadata          = is_array( $intent->get_metadata() ) ? $intent->get_metadata() : [];
 			$intent_meta_order_id_raw = $intent_metadata['order_id'] ?? '';
 			$intent_meta_order_id     = is_numeric( $intent_meta_order_id_raw ) ? intval( $intent_meta_order_id_raw ) : 0;
--- a/includes/admin/class-wc-rest-payments-orders-controller.php
+++ b/includes/admin/class-wc-rest-payments-orders-controller.php
@@ -8,8 +8,8 @@
 defined( 'ABSPATH' ) || exit;
 
 use WCPay\Constants\Payment_Method;
-use WCPay\Core\Server\Request\Create_Intent;
-use WCPay\Core\Server\Request\Get_Intent;
+use WCPay\Core\Server\Request\Create_Intention;
+use WCPay\Core\Server\Request\Get_Intention;
 use WCPay\Exceptions\API_Exception;
 use WCPay\Logger;
 
@@ -141,7 +141,7 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$request = Get_Intent::create();
+			$request = Get_Intention::create();
 			$request->set_intent_id( $intent_id );
 			$intent                   = $request->send( 'wcpay_get_intent_request', $order );
 			$intent_metadata          = is_array( $intent->get_metadata() ) ? $intent->get_metadata() : [];
@@ -246,7 +246,7 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			}
 
 			// Do not process intents that can't be captured.
-			$request = Get_Intent::create();
+			$request = Get_Intention::create();
 			$request->set_intent_id( $intent_id );
 			$intent = $request->send( 'wcpay_get_intent_request', $order );
 
@@ -364,7 +364,7 @@ class WC_REST_Payments_Orders_Controller extends WC_Payments_REST_Controller {
 			$metadata                 = $request->get_param( 'metadata' ) ?? [];
 			$metadata['order_number'] = $order->get_order_number();
 
-			$wcpay_server_request = Create_Intent::create();
+			$wcpay_server_request = Create_Intention::create();
 			$wcpay_server_request->set_currency_code( $currency );
 			$wcpay_server_request->set_amount( WC_Payments_Utils::prepare_amount( $order->get_total(), $currency ) );
 			if ( $customer_id ) {
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -13,6 +13,7 @@ use WCPay\Core\Mode;
 use WCPay\Exceptions\{ Add_Payment_Method_Exception, Amount_Too_Small_Exception, Process_Payment_Exception, Intent_Authentication_Exception, API_Exception };
 use WCPay\Core\Server\Request\Create_And_Confirm_Intention;
 use WCPay\Core\Server\Request\Create_Intent;
+use WCPay\Core\Server\Request\Get_Intent;
 use WCPay\Fraud_Prevention\Fraud_Prevention_Service;
 use WCPay\Logger;
 use WCPay\Payment_Information;
@@ -960,7 +961,10 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			$intent = null;
 			if ( ! empty( $platform_checkout_intent_id ) ) {
 				// If the intent is included in the request use that intent.
-				$intent                   = $this->payments_api_client->get_intent( $platform_checkout_intent_id );
+				$request = Get_Intent::create();
+				$request->set_intent_id( $platform_checkout_intent_id );
+				$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 				$intent_meta_order_id_raw = $intent->get_metadata()['order_id'] ?? '';
 				$intent_meta_order_id     = is_numeric( $intent_meta_order_id_raw ) ? intval( $intent_meta_order_id_raw ) : 0;
 
@@ -1502,8 +1506,11 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			$payment_method_id      = $order->get_meta( '_payment_method_id', true );
 			$payment_method_details = $this->payments_api_client->get_payment_method( $payment_method_id );
 		} elseif ( $order->meta_exists( '_intent_id' ) ) {
-			$payment_intent_id      = $order->get_meta( '_intent_id', true );
-			$payment_intent         = $this->payments_api_client->get_intent( $payment_intent_id );
+			$payment_intent_id = $order->get_meta( '_intent_id', true );
+			$request           = Get_Intent::create();
+			$request->set_intent_id( $payment_intent_id );
+			$payment_intent = $request->send( 'wcpay_get_intent_request', $order );
+
 			$charge                 = $payment_intent ? $payment_intent->get_charge() : null;
 			$payment_method_details = $charge ? $charge->get_payment_method_details() : [];
 		}
@@ -2080,8 +2087,12 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		$currency                 = WC_Payments_Utils::get_order_intent_currency( $order );
 
 		try {
-			$intent_id    = $order->get_transaction_id();
-			$intent       = $this->payments_api_client->get_intent( $intent_id );
+			$intent_id = $order->get_transaction_id();
+
+			$request = Get_Intent::create();
+			$request->set_intent_id( $intent_id );
+			$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 			$payment_type = $this->is_payment_recurring( $order->get_id() ) ? Payment_Type::RECURRING() : Payment_Type::SINGLE();
 
 			$metadata_from_intent = $intent->get_metadata(); // mobile app may have set metadata.
@@ -2108,7 +2119,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				$http_code     = $e->get_http_code();
 
 				// Fetch the Intent to check if it's already expired and the site missed the "charge.expired" webhook.
-				$intent = $this->payments_api_client->get_intent( $order->get_transaction_id() );
+				$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 				if ( 'canceled' === $intent->get_status() ) {
 					$is_authorization_expired = true;
 				}
@@ -2166,7 +2178,10 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		} catch ( API_Exception $e ) {
 			try {
 				// Fetch the Intent to check if it's already expired and the site missed the "charge.expired" webhook.
-				$intent = $this->payments_api_client->get_intent( $order->get_transaction_id() );
+				$request = Get_Intent::create();
+				$request->set_intent_id( $order->get_transaction_id() );
+				$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 				$status = $intent->get_status();
 				if ( 'canceled' !== $status ) {
 					$error_message = $e->getMessage();
@@ -2363,7 +2378,10 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 
 			if ( $amount > 0 ) {
 				// An exception is thrown if an intent can't be found for the given intent ID.
-				$intent    = $this->payments_api_client->get_intent( $intent_id );
+				$request = Get_Intent::create();
+				$request->set_intent_id( $intent_id );
+				$intent = $request->send( 'wcpay_get_intent_request', $order );
+
 				$status    = $intent->get_status();
 				$charge    = $intent->get_charge();
 				$charge_id = ! empty( $charge ) ? $charge->get_id() : null;
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1256,9 +1256,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		if ( $currency_order !== $currency_account ) {
 			// We check that the currency used in the order is different than the one set in the WC Payments account
 			// to avoid requesting the charge if not needed.
-			$request = Get_Charge::create();
-			$request->set_charge_id( $charge_id );
-			$charge = $request->send( 'wcpay_get_charge_request', $charge_id );
+			$request = Get_Charge::create( $charge_id );
+			$charge  = $request->send( 'wcpay_get_charge_request', $charge_id );
 
 			$exchange_rate = $charge['balance_transaction']['exchange_rate'] ?? null;
 			if ( isset( $exchange_rate ) ) {
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -962,9 +962,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			$intent = null;
 			if ( ! empty( $platform_checkout_intent_id ) ) {
 				// If the intent is included in the request use that intent.
-				$request = Get_Intention::create();
-				$request->set_intent_id( $platform_checkout_intent_id );
-				$intent = $request->send( 'wcpay_get_intent_request', $order );
+				$request = Get_Intention::create( $platform_checkout_intent_id );
+				$intent  = $request->send( 'wcpay_get_intent_request', $order );
 
 				$intent_meta_order_id_raw = $intent->get_metadata()['order_id'] ?? '';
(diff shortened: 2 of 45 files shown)


includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -138,7 +141,9 @@ public function capture_terminal_payment( WP_REST_Request $request ) {
 			}
 
 			// Do not process intents that can't be captured.
-			$intent                   = $this->api_client->get_intent( $intent_id );
+			$request = Get_Intent::create();
+			$request->set_intent_id( $intent_id );
```
RadoslavGeorgiev: I'm starting to think that for update/get request, we could maybe include the ID in the constructor. Do you think that makes sense? This is pretty descrptive as well though, no strong opinions here.

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	/**
+	 * Returns the request's API.
+	 *
+	 * @return string
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function get_api(): string {
+		if ( null === $this->intent_id ) {
```
RadoslavGeorgiev: Is this check being performed here because `get_params` is not called for GET requests?

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,196 @@
...
+			throw new Invalid_Request_Parameter_Exception( __( 'Intent ID is not set.', 'woocommerce-payments' ), 'wcpay_core_request_intent_not_set' );
+		}
+		return WC_Payments_API_Client::INTENTIONS_API . '/' . $this->intent_id;
```
RadoslavGeorgiev: At this point we're not returning an API here, but a URL instad. Should we rename the method altogether?

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,196 @@
...
+	 *
+	 * @return void
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function set_intent_id( string $intent_id ) {
+		$this->validate_stripe_id( $intent_id, [ 'pi' ] );
+
+		// Prevent mutation of intent id. It can be only set once.
+		if ( null === $this->intent_id ) {
+			$this->intent_id = $intent_id;
```
RadoslavGeorgiev: This would silently ignore the new ID if one is present, an exception would be better.

Why not just set it as an immutable param though?

includes/core/server/request/class-create-intent.php
```diff
@@ -15,6 +16,9 @@
  * Request class for creating intents.
  */
 class Create_Intent extends Request {
```
RadoslavGeorgiev: `Create_Intention` :)

RadoslavGeorgiev (changes requested):

includes/admin/class-wc-rest-payments-orders-controller.php
zmaglica: The reason why I choose this approach is that we have dependencies in abstract class constructor and create method is static, and it needs to be compatible with parent class. I was thinking to add constructor injection for this situation, but we used setters for filling data into request classes, so I decided to go with the same approach. If you think that constructor parameter is also a valid option, I will gladly refactor it. For me, it is 50-50 and that's why I chose setter since we already using it.

includes/core/server/request/class-update-intention.php
zmaglica: Yes. This intent ID is not part of request parameters. It is part of endpoint and for this reason, I validate it here.

includes/core/server/request/class-update-intention.php
zmaglica: Yeah, we could. Maybe we can rename it to `get_url` or `get_endpoint`, `get_uri`, or `get_path` (or something similar) ?

includes/core/server/request/class-update-intention.php
zmaglica: Immutable param is part of params property and this is part of endpoint. Maybe I can throw immutable exception here or stick to constructor injection to prevent confusion between params, setters and IDs in endpoints.

includes/core/server/request/class-create-intent.php
zmaglica: Will refactor.

includes/core/server/request/class-update-intention.php
RadoslavGeorgiev: Ah, let's just leave it like this for now. I'm looking at the API client's `request` method, and it already uses `$api` there, maybe it's not worth changing.

tests/unit/core/server/request/test-class-core-update-intention-request.php
```diff
@@ -58,7 +58,7 @@ public function test_exception_will_throw_if_intent_is_invalid() {
 	}
 
 	public function test_payment_intent_api_url_when_action_is_set() {
```
RadoslavGeorgiev: I think this test is redundant now?

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -380,13 +385,13 @@ public function create_setup_intent() {
...
-			array_values( $enabled_payment_methods )
-		);
+		$request = Create_Setup_Intention::create();
+		$request->set_customer( $customer_id );
+		$request->set_payment_method_types( array_values( $enabled_payment_methods ) );
+		$setup_intent = $request->send( 'wcpay_create_setup_intention_request' );
 		return [
-			'id'            => $setup_intent['id'],
-			'client_secret' => $setup_intent['client_secret'],
+			'id'            => $setup_intent->offsetGet( 'id' ),
```
RadoslavGeorgiev: The idea of implementing `ArrayAccess` was to allow this line to just be `$setup_intent['id']`.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1045,14 +1053,12 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
-					false,
-					$this->is_platform_payment_method( $payment_information->is_using_saved_payment_method() ),
-					$save_user_in_platform_checkout,
-					$metadata
-				);
+				$create_and_confirm_intent_request = Create_And_Confirm_Setup_Intention::create();
+				$create_and_confirm_intent_request->set_customer( $customer_id );
+				$create_and_confirm_intent_request->set_payment_method( $payment_information->get_payment_method() );
+				$create_and_confirm_intent_request->set_metadata( $metadata );
+				$intent = $create_and_confirm_intent_request->send( 'wcpay_create_and_confirm_setup_intention_request', $payment_information, false, $save_user_in_platform_checkout );
```
RadoslavGeorgiev: This is really a nitpick, but could we just use `$request` here? Having half of the line occupied by the variable name feels very wasteful.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2617,12 +2589,12 @@ public function create_and_confirm_setup_intent() {
 			$customer_id   = $this->customer_service->create_customer_for_user( $user, $customer_data );
 		}
 
-		return $this->payments_api_client->create_and_confirm_setup_intent(
-			$payment_information->get_payment_method(),
-			$customer_id,
-			$should_save_in_platform_account,
-			$this->is_platform_payment_method( $payment_information->is_using_saved_payment_method() )
-		);
+		$create_and_confirm_intent_request = Create_And_Confirm_Setup_Intention::create();
```
RadoslavGeorgiev: Same note about long variable names here.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2617,12 +2589,12 @@ public function create_and_confirm_setup_intent() {
...
+		$create_and_confirm_intent_request->set_customer( $customer_id );
+		$create_and_confirm_intent_request->set_payment_method( $payment_information->get_payment_method() );
+		$intent = $create_and_confirm_intent_request->send( 'wcpay_create_and_confirm_setup_intention_request', $payment_information, $should_save_in_platform_account, false );
+
+		return $intent->to_array();
```
RadoslavGeorgiev: Is there something, which requires the intent to be an actual array? Unless there is a strong `array` typehint in a function, this conversion should not be needed.

includes/core/server/request/class-create-and-confirm-setup-intention.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Set metadata.
+	 *
+	 * @param array $metadata Metadata to set.
+	 *
+	 * @return void
+	 */
+	public function set_metadata( array $metadata ) {
+		$this->set_param( 'metadata', $metadata );
```
RadoslavGeorgiev, line 70: I'm wondering whether we should add similar processing here to `Create_Intention`'s `set_metadata` method, namely the intent description. I know it was like this before, but IMO that's a missed opportunity rather than done intentionally.

RadoslavGeorgiev (changes requested):
Looking good, almost there :)

Left a few small comments regarding the code, nothing drastic IMO. I'll start reviewing tests now, but I wanted to unblock you in the meantime.

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -6,6 +6,10 @@
  */
 
 use PHPUnit\Framework\MockObject\MockObject;
+use WCPay\Core\Server\Request\Create_Intention;
+use WCPay\Core\Server\Request\Get_Intention;
+use WCPay\Core\Server\Request\Update_Intention;
```
RadoslavGeorgiev: This `use` statement is never used.

tests/unit/admin/test-class-wc-rest-payments-orders-controller.php
```diff
@@ -6,6 +6,10 @@
...
+use WCPay\Exceptions\API_Exception;
 use WCPay\Exceptions\Rest_Request_Exception;
```
RadoslavGeorgiev: And this one seems not to be in use as well.

includes/payment-methods/class-upe-payment-gateway.php
zmaglica: Yeah... I totally missed it. It is fixed.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: Fixed.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: Fixed.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: I returned default Intent or Request object now as you suggested. I still used `->to_array()` when rest response is returned. The reason behind this is that I want to be sure that correct data or array will be used.

includes/core/server/request/class-create-and-confirm-setup-intention.php
zmaglica: I am sorry, but I am not 100% certain what do you mean here. Do you want to use same metadata setter with description and fingerprint here from `Create_Intention` class or do you want to use simple metadata setter where you don't add description and fingerprinting to `Create_Intention` request class?

Adding description is easy, and we could include it easily. Please let me know what you had in mind and I can easily implement it.

RadoslavGeorgiev: Thinking about it a bit more, I think it makes sense to add the description here, without messing with the fingerprint metadata, as that could turn out to be problematic in the future, while the description just makes sense.


The only part I would maybe do differently would be to only set the description whenever an order ID is available, instead of throwing an exception if it was missing.

```php
// Don't do this
if ( ! isset( $metadata['order_number'] ) ) {
	throw new Exception();
}
$description = $this->get_intent_description( $metadata['order_number'] ?? 0 );
$this->set_param( 'description', $description );

// Do this
if ( isset( $metadata['order_number'] ) ) {
	$description = $this->get_intent_description( $metadata['order_number'] ?? 0 );
	$this->set_param( 'description', $description );
}
```

tests/WCPAY_UnitTestCase.php
```diff
@@ -28,21 +28,25 @@ protected function is_wpcom() {
...
-	 * @param  WC_Payments_API_Client $api_client_mock Specific API client mock if necessary.
-	 * @param  WC_Payments_Http       $http_mock       Specific HTTP mock if necessary.
-	 * @return Request                                 The mocked request.
+	 * @param  string                 $request_class                The class of the mocked request.
+	 * @param  int                    $total_api_calls              Number of same api calls that will be executed. Used when you want to send multiple request, using the same instance of class, i.e. retry mechanism.
+	 * @param  string|null            $request_class_constructor_id Used when constructor class gets ID (like intent id or charge id) and passes it as a constructor dependency in mocked request class.
+	 * @param  mixed                  $response                     The expected response.
+	 * @param  WC_Payments_API_Client $api_client_mock              Specific API client mock if necessary.
+	 * @param  WC_Payments_Http       $http_mock                    Specific HTTP mock if necessary.
+
```
RadoslavGeorgiev: Let's remove this empty line.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -9,30 +9,24 @@
...
-use WC_Customer;
 use WC_Helper_Order;
 use WC_Helper_Intention;
 use WC_Helper_Token;
-use WC_Payments_Utils;
-use WC_Subscriptions;
-use WC_Subscriptions_Cart;
 use WCPay\WC_Payments_UPE_Checkout;
 use WCPAY_UnitTestCase;
 use WP_User;
```
RadoslavGeorgiev: This one is not used either.

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,186 @@
...
+
+	/**
+	 * Set selected UPE payment method type.
+	 *
+	 * @param string $selected_upe_payment_type Selected UPE payment method.
+	 * @param array  $enabled_payment_methods Enabled payment methods.
+	 *
+	 * @return void
+	 */
+	public function set_selected_upe_payment_method_type( string $selected_upe_payment_type, array $enabled_payment_methods ) {
```
RadoslavGeorgiev: I agree we already have plenty of logic within supposedly simple setters (intents, level3 data, etc.), but I have the feeling that the contents of this methods are way too specific to live within the request class. 

After checking where this method gets called (exclusively within `UPE_Payment_Gateway`), with the same method signature (`set_selected_upe_payment_method_type( (string) $selected_upe_payment_type, $this->get_payment_method_ids_enabled_at_checkout( null, true ) )`), I think we could move this logic out of the request, and within the gateway. WDYT?

I landed here now, after noticing that within all requests, this is the only setter, which has more than a single parameter.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -483,28 +555,57 @@ public function test_update_payment_intent_with_payment_country() {
...
+		$request->expects( $this->once() )
+			->method( 'set_payment_method_types' )
+			->with(
+				$this->callback(
+					function( $argument ) {
+						return is_array( $argument ) && ! empty( $argument );
+					}
+				)
+			);
+
```
RadoslavGeorgiev, line 587: I'm not sure that mocking `set_capture_method`, `set_metadata`, and `set_payment_method_types` (or at least this specific one's current form) are contributing anything to the test. We can remove them.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -515,17 +616,43 @@ public function test_create_payment_intent_defaults_to_automatic_capture() {
...
+		$request->expects( $this->once() )
+			->method( 'set_payment_method_types' )
+			->with(
+				$this->callback(
+					function( $argument ) {
+						return is_array( $argument ) && ! empty( $argument );
+					}
+				)
+			);
```
RadoslavGeorgiev: The `set_metadata` and `set_payment_method_types` calls could be removed here as well.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -536,17 +663,43 @@ public function test_create_payment_intent_with_automatic_capture() {
...
+
+		$request->expects( $this->once() )
+			->method( 'set_currency_code' )
+			->with( strtolower( $intent->get_currency() ) );
+
+		$request->expects( $this->once() )
+			->method( 'set_capture_method' )
+			->with( false );
+
+		$request->expects( $this->once() )
```
RadoslavGeorgiev: Redundant tests here as well.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -557,17 +710,44 @@ public function test_create_payment_intent_with_manual_capture() {
...
+			->method( 'set_currency_code' )
+			->with( strtolower( $intent->get_currency() ) );
+
+		$request->expects( $this->once() )
+			->method( 'set_capture_method' )
+			->with( true );
+
+		$request->expects( $this->once() )
```
RadoslavGeorgiev, line 662: Maybe remove this one as well?

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -585,15 +765,30 @@ public function test_create_setup_intent_existing_customer() {
...
-		$this->mock_api_client
-			->expects( $this->once() )
-			->method( 'create_setup_intention' )
-			->with( 'cus_mock', [ 'card' ] )
+		$request = $this->mock_wcpay_request( Create_Setup_Intention::class );
+		$request->expects( $this->once() )
+			->method( 'set_customer' )
+			->with( 'cus_mock' );
+
+		$request->expects( $this->once() )
```
RadoslavGeorgiev, line 688: This mock can be removed as well, `[ 'card' ]` is not of significance in this case.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -1267,11 +1465,43 @@ public function test_create_payment_intent_uses_cached_minimum_amount() {
...
+		$request = $this->mock_wcpay_request( Create_Intention::class );
+		$request->expects( $this->once() )
+			->method( 'set_amount' )
+			->with( $intent->get_amount() );
+
+		$request->expects( $this->once() )
+			->method( 'set_currency_code' )
+			->with( strtolower( $intent->get_currency() ) );
+
+		$request->expects( $this->once() )
```
RadoslavGeorgiev: Let's clean the insignificant mocks from this test as well.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -1290,13 +1520,42 @@ public function test_create_payment_intent_creates_new_intent_with_minimum_amoun
...
+			->method( 'set_capture_method' )
+			->with( false );
+
+		$request->expects( $this->once() )
```
RadoslavGeorgiev: Some more un-needed mocks.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -1290,13 +1520,42 @@ public function test_create_payment_intent_creates_new_intent_with_minimum_amoun
...
-			->method( 'create_intention' )
-			->withConsecutive(
-				[ 45, 'usd', [ 'card' ] ],
-				[ 50, 'usd', [ 'card' ] ]
-			)
+		$request = $this->mock_wcpay_request( Create_Intention::class, 2 );
+
+		$request->expects( $this->exactly( 2 ) )
+			->method( 'set_amount' )
+			->withConsecutive( [ 45 ], [ $intent->get_amount() ] );
```
RadoslavGeorgiev, line 1403: Correct me if I'm wrong, but I find it very beautiful how the same request can be modified, resent, and we can test only for the modification setters 😍

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2079,24 +2089,25 @@ public function capture_charge( $order, $include_level3 = true ) {
...
 
 			$metadata_from_intent = $intent->get_metadata(); // mobile app may have set metadata.
 			$metadata_from_order  = $this->get_metadata_from_order( $order, $payment_type );
 			$merged_metadata      = array_merge( (array) $metadata_from_order, (array) $metadata_from_intent ); // prioritize metadata from mobile app.
 
-			$this->payments_api_client->prepare_intention_for_capture(
-				$intent_id,
-				$merged_metadata
-			);
+			$wcpay_request = Update_Intention::create( $intent_id );
```
RadoslavGeorgiev, line 2103: The old `prepare_intention_for_capture` method, which was called here, did not just update the metadata, but also set the `receipt_email` to be an empty string:

```
// 'receipt_email' is set to prevent Stripe from sending receipts (when intent is created outside WCPay).
```

Based on the comment in there, I'd say that's a pretty important detail, which we could easily miss. Because of that, I'd recomment extending the `Update_Intention` class with another `Prepare_Intent_for_Capture` one, and adding an empty `receipt_email` as a default value there, nothing more:

```php
class Prepare_Intention_for_Capture extends Update_Intention {
	const DEFAULT_PARAMS = [
		'receipt_email' => '',
	];
}
```

☝️ something like this should suffice.

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -776,15 +777,30 @@ public function test_capture_charge_success() {
...
-		);
+		$request = $this->mock_wcpay_request( Get_Intention::class, 1, $intent_id );
+
+		$request->expects( $this->once() )
+			->method( 'format_response' )
+			->willReturn( $mock_intent );
+
+		$update_intent_request = $this->mock_wcpay_request( Update_Intention::class, 1, $intent_id );
+		$update_intent_request->expects( $this->once() )
+			->method( 'set_metadata' )
```
RadoslavGeorgiev, line 790: Checking metadata here is also not needed.

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -835,15 +851,34 @@ public function test_capture_charge_success_non_usd() {
...
+			->with(
+				$this->callback(
+					function( $argument ) {
+						return is_array( $argument ) && ! empty( $argument );
```
RadoslavGeorgiev: I'm leaving a lot of similar comments, let's just look for `return is_array( $argument ) && ! empty( $argument );` within all tests, and remove it in all but a single place. It's included in too many tests without any obvious use of it.

tests/unit/core/server/request/test-class-core-update-intention-request.php
```diff
@@ -0,0 +1,93 @@
...
+		$this->expectException( Invalid_Request_Parameter_Exception::class );
+		$request->set_customer( '1' );
+	}
+
+	public function test_exception_will_throw_if_intent_is_invalid() {
+		$this->expectException( Invalid_Request_Parameter_Exception::class );
+		$request = new Update_Intention( $this->mock_api_client, $this->mock_wc_payments_http_client, '1' );
+	}
+
+	public function test_create_intent_request_will_be_created() {
```
RadoslavGeorgiev: Let's change the name of this test.

includes/admin/class-wc-rest-payments-orders-controller.php
```diff
@@ -8,6 +8,9 @@
 defined( 'ABSPATH' ) || exit;
 
 use WCPay\Constants\Payment_Method;
+use WCPay\Core\Server\Request\Create_Intention;
+use WCPay\Core\Server\Request\Get_Intention;
+use WCPay\Exceptions\API_Exception;
```
RadoslavGeorgiev: Unused line.

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Class file for WCPay\Core\Server\Request\Update_Intent.
+ *
+ * @package WooCommerce Payments
+ */
+
+namespace WCPay\Core\Server\Request;
+
+use WC_Payments;
+use WC_Payments_Http_Interface;
```
RadoslavGeorgiev: Unused import.

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Request class for creating intents.
+ */
+class Update_Intention extends Request {
+	use Intention;
+	use Level3;
+
+	const IMMUTABLE_PARAMS = [ 'amount' ];
+	const REQUIRED_PARAMS  = []; // This use to be amount and currency, but since it is not required on server API, I will leave it blank.
```
RadoslavGeorgiev: > This use to be amount and currency, but since it is not required on server API, I will leave it blank.

Let's just remove it then.

includes/core/server/request/class-update-intention.php
```diff
@@ -0,0 +1,186 @@
...
+	 * @param string $id Sets the intent ID, which will be used in the request URL.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	protected function set_id( string $id ) {
+		$this->validate_stripe_id( $id );
+		$this->id = $id;
+	}
+
+
```
RadoslavGeorgiev: There are a lot of cases of unnecessary empty lines in the PR, please go through them ana minimze them (1 in this case).

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2079,24 +2089,25 @@ public function capture_charge( $order, $include_level3 = true ) {
...
+			$wcpay_request->send( 'wcpay_prepare_intention_for_capture', $order );
 
-			$intent = $this->payments_api_client->capture_intention(
-				$intent_id,
-				WC_Payments_Utils::prepare_amount( $amount, $order->get_currency() ),
-				$include_level3 ? $this->get_level3_data_from_order( $order ) : []
-			);
+			$capture_intention_request = Capture_Intention::create( $intent_id );
+			$capture_intention_request->set_amount_to_capture( WC_Payments_Utils::prepare_amount( $amount, $order->get_currency() ) );
+			$capture_intention_request->set_level3( $include_level3 ? $this->get_level3_data_from_order( $order ) : [] );
```
RadoslavGeorgiev: Small tweak, but let's not call `set_level3` if `$include_level3` is false.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: I'm also not completely inline with the name here. Considering `capture_charge` includes plenty of requests, let's have all of them assigned to a distinctive variable, ex. `$get_request` and `$update_request` instead of `$request` and `$wcpay_request`.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -11,6 +11,11 @@
 use WC_Payment_Token_WCPay_SEPA;
 use WC_Payments_Explicit_Price_Formatter;
```
RadoslavGeorgiev: Let's remove this one here.

RadoslavGeorgiev (changes requested):
Left a bit more comments than expected, but half of them are just for removing redundant parts of tests.

Let's not forget to add the necessary tests to make everything light up in green, and we should be done.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: The reason why I mocked setters for each of these classes is that I wanted to be sure that the correct setter method will be called with correct parameters.  With that approach, we are confident that the request will be built in a way that was supposed to be for certain parts of code. Not mocking the whole construction of class of object could lead to incomplete tests and false positive tests.

If you ask me, I would leave them, but I understand your point of view. I removed the "duplicate" ones.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: Yeah, I love it. Make things(testing and mocking) way easier 🥳

includes/class-wc-payment-gateway-wcpay.php
zmaglica: I honestly don't like the idea of creating a new class and change one line for the sake of covering something that is hard to sport... I mean, where is the line? Shall we create a new class, require it manually (because we don't have autoloader) and write tests for that class? We are already covering scenarios like these and famous `receipt_email` parameter is by default set to correct value. 

If we decide to use this approach, we are setting examples that it is fine to create a bunch of classes to mimic specific use case and after some time you will have a bunch of them and good luck navigating them and onboard someone new into this where you have multiple classes for same requests where only one parameter is changed.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: Fixed. All duplicate/unnecessary asserts have been removed.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: Removed.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: I not sure why to remove it? We want to be sure that set capture method is called with desired param. I did remove a bunch of them from other tests, but here it is mandatory IMHO.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
zmaglica: Is there a reason to remove it? This method is called with this parameter, and there is noting bad in mocking that. I removed that from other tests, but it would be nice to have it at least somewhere to validate is it called or setter correctly.

tests/unit/test-class-wc-payment-gateway-wcpay.php
zmaglica: Fixed.

tests/unit/core/server/request/test-class-core-update-intention-request.php
zmaglica: Fixed.

includes/admin/class-wc-rest-payments-orders-controller.php
zmaglica: Fixed.

includes/core/server/request/class-update-intention.php
zmaglica: Fixed.

includes/core/server/request/class-update-intention.php
zmaglica: Fixed.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: Fixed.

includes/payment-methods/class-upe-payment-gateway.php
zmaglica: Fixed.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: It makes sense to include the empty parameter with the defaults of `Update_Intention`, the way it is right now.

However, I had not noticed the [default value](https://github.com/Automattic/woocommerce-payments/blob/b255f2d9cc6f23bc6b89a0c85880bc069959f545/includes/core/server/request/class-update-intention.php#L24) until now, even though it's pretty important. Let's add a comment there, and I think we can close this discussion.

tests/unit/payment-methods/test-class-upe-payment-gateway.php
RadoslavGeorgiev: My point was that we don't need to test the format 7 times in a single test suite, but now that there are no duplicates, it looks good :)

tests/unit/payment-methods/test-class-upe-payment-gateway.php
RadoslavGeorgiev: Same here, having it once here and there it's fine, there was just no need to test it everywhere.

RadoslavGeorgiev (changes requested):
Everything looks good, only one small comment required on my side (check the open conversation) 🎉 

And of course, let's not forget the missing/needing update tests.

RadoslavGeorgiev (approved):

PR #5236, opened 2022-12-02
Fix platform checkout tracks for the express payment button.

brettshumaker:

Once we added the WooPay Express Checkout button functionality, `wcpayTracks` requests were failing due to the use of `getConfig` and how we were making the necessary parameters available when showing the express checkout button. This PR addresses this by adding a filter to the value returned by `WC_Payments_Checkout->get_payment_fields_js_config()` method. We're also attaching that config to our express payment button js file (it's normally only attached to the `checkout.js` file). This ensures that we can use `getConfig( 'value' )` for one of our custom config values anywhere we're displaying the WooPay express checkout button.

This PR also adds two new track events: `platform_checkout_express_button_offered` and `platform_checkout_express_button_clicked`. These events will also include a `context` event property that says which screen they originated from - `product`, `cart`, or `checkout`.

#### Testing instructions

1. Apply this PR.
2. Make sure the Express Checkout feature flag is enabled from the latest Dev Tools.
3. Make sure the Express Checkout button is set to display on the Product, Cart, and Checkout pages from Payments > Settings > WooPay Customize > Show express checkouts on.
4. Visit a single product page.
5. In your browser's `Network` tab, ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_offered
tracksEventProp: {"context":"product"}
```
6. Click the WooPay express payment button and ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_clicked
tracksEventProp: {"context":"product"}
```
7. Add the product to your cart and visit the cart page.
8. In your browser's `Network` tab, ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_offered
tracksEventProp: {"context":"cart"}
```
9. Click the WooPay express payment button and ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_clicked
tracksEventProp: {"context":"cart"}
```
10. Visit the checkout page.
11. In your browser's `Network` tab, ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_offered
tracksEventProp: {"context":"checkout"}
```
12. Click the WooPay express payment button and ensure that a call was made to `admin-ajax.php` with this in the `Payload`:
```
...
tracksEventName: platform_checkout_express_button_clicked
tracksEventProp: {"context":"checkout"}
```

-------------------

**Post merge**


includes/class-wc-payments-checkout.php
```diff
@@ -126,6 +126,8 @@ public function get_payment_fields_js_config() {
 			'platformCheckoutSignatureNonce' => wp_create_nonce( 'platform_checkout_signature_nonce' ),
 			'platformCheckoutMerchantId'     => Jetpack_Options::get_option( 'id' ),
 		];
+
+		return apply_filters( 'wcpay_payment_fields_js_config', $js_config );
```
brettshumaker, line 135: @jrodger Do you see any issue with adding a filter to this return value? My experience with WC Pay doesn't go back too far so I want to make sure that adding a filter here doesn't have any unintended consequences.

jrodger: The only real consequence I can think of is exactly what you added it for, other code can alter the array. That's not necessarily a big problem, but it does make it easy for other plugins to alter the JS config as well. Any malicious plugin will find a way to do nasty things, but we don't necessarily want to make it easy for them.

Looking at where `wcpay_payment_fields_js_config` is used, is there any reason you need to use a filter here as opposed to adding the config line directly in this method?

cc @marcinbot, I think we've discussed similar things around filters in the past?

brettshumaker: > Looking at where wcpay_payment_fields_js_config is used, is there any reason you need to use a filter here as opposed to adding the config line directly in this method?

I should have been more clear in _why_ I chose to do it this way instead of adding it directly. The issue we ran into is that we need our express payment button to be able to be displayed on the product, cart, cart block, checkout, and block checkout pages. `wcpay_config` isn't normally available anywhere except the checkout page so I thought that if I'm going to be editing the content of `wcpay_config`, I need to do it everywhere it's printed to the screen. I'll say that we are printing the modified version of `wcpay_config` (which we need because a dependency of the button uses `getConfig()`) on every page that we need it, but due to script dependencies/order of output on the screen, we're unable to print our version of `wcpay_config` _after_ the default version so the default version always overwrites our custom one.

Let me know if that makes sense.

marcinbot: > cc @marcinbot, I think we've discussed similar things around filters in the past?

Security is a valid concern, but with all the filters that WP provides, if someone had the malicious plugin on their site, it doesn't need a filter here to mess with things.

However, I feel like adding filters that are then hooked into inside the same codebase is a bit of an "antipattern". Is it possible for the `WC_Payments_Checkout` to be aware of the `WC_Payments_Platform_Checkout_Button_Handler` and then conditionally attach the `platformCheckoutButton` param here?

The filters would make sense if we created them for extensions. Whereas here everything is in one place, owned by us, and there should be a way to do things in a way that's easier to understand and debug in the future.

Alternatively, it could be argued that the platform checkout is kind of an extension, so my feelings about this are not as strong as they would be in other cases :) If it is too difficult to add a conditional, we could leave it as is.

brettshumaker: > I feel like adding filters that are then hooked into inside the same codebase is a bit of an "antipattern".

Agreed, but it's not entirely unheard of as there are a few examples inside of WooCommerce Payments already.

> Is it possible for the `WC_Payments_Checkout` to be aware of the `WC_Payments_Platform_Checkout_Button_Handler` and then conditionally attach the `platformCheckoutButton` param here?

Possibly. Some changes would have to be made to the `WC_Payments_Platform_Checkout_Button_Handler` class, though. As of right now, the class is loaded but the `init` method returns early if the feature is not enabled. So the class always exists but we'd need a way for external code to see if it's going to do anything on that request.

> Whereas here everything is in one place, owned by us, and there should be a way to do things in a way that's easier to understand and debug in the future.

Would adding a comment above the `apply_filters()` line stating where it's being used within the codebase help? Sort of how core Woo documents where it's using its own filters/actions?

> Alternatively, it could be argued that the platform checkout is kind of an extension

These are my feelings as well. It felt "cleaner" to me to have the `WC_Payments_Platform_Checkout_Button_Handler` class contain its own logic where possible and not sprinkle it in where needed in the rest of the codebase.

bborman22: One question to this whole effort, that I haven't been able to nail down and was curious other's opinions on, is whether it would be preferable to keep all of our JS params under one umbrella (like wcpay_config) or is it preferable to break them down in to their own components as needed (like wcpay_config, wcpayPaymentRequestParams, and wcpayWooPayExpressParams)? The existing arrangement seemed to make sense for the time when there were only two (wcpay_config and wcpayPaymentRequestParams), but after I added a third I started to question it. Is there a preferred pattern to this?

client/checkout/platform-checkout/express-button/woopay-express-checkout-button.js
```diff
@@ -23,6 +43,13 @@ export const WoopayExpressCheckoutButton = ( {
 			return; // eslint-disable-line no-useless-return
 		}
 
+		wcpayTracks.recordUserEvent(
+			'platform_checkout_express_button_clicked',
```
bborman22: Should we use the event constants here instead?

client/checkout/platform-checkout/express-button/woopay-express-checkout-button.js
```diff
@@ -2,19 +2,39 @@
...
+		height,
+		size,
+		theme,
+		context,
+	} = buttonSettings;
+
+	useEffect( () => {
+		if ( ! isPreview ) {
+			wcpayTracks.recordUserEvent(
+				'platform_checkout_express_button_offered',
```
bborman22: I'm not super familiar with these track events, would this "offered" event cause too much noise? I noticed the `PLATFORM_CHECKOUT_OFFERED` is only recorded after a user exists check, while this one would be recorded I believe on every page load of a product, cart, or checkout (settings dependent of course). This may be a great thing to track, but it was just a thought I had when I saw it.

bborman22 (approved):
This all worked for me! I just had a few small notes around the track events that were minor. I'll also defer to @jrodger and @marcinbot on the larger question of the hook and approach overall as we've discussed ourselves quite a bit and it seems acceptable to me, but still doesn't feel like a slam dunk solution.

client/checkout/platform-checkout/express-button/woopay-express-checkout-button.js
brettshumaker: Yeah, I agree that there could be a lot of noise, but I think we would need to track it so we can see what the conversion rate is (button clicks/button offered).

client/checkout/platform-checkout/express-button/woopay-express-checkout-button.js
brettshumaker: Absolutely. Good catch! Resolved in 1279fc30

includes/class-wc-payments-checkout.php
marcinbot: > Possibly. Some changes would have to be made to the WC_Payments_Platform_Checkout_Button_Handler class, though. As of right now, the class is loaded but the init method returns early if the feature is not enabled. So the class always exists but we'd need a way for external code to see if it's going to do anything on that request.

@brettshumaker 

Just an idea: add a flag or something inside `init` and then conditionally modify (or not) the config object. But maybe that's not the path we necessarily need to go down.

> Would adding a comment above the apply_filters() line stating where it's being used within the codebase help? Sort of how core Woo documents where it's using its own filters/actions?
> ...
> These are my feelings as well. It felt "cleaner" to me to have the WC_Payments_Platform_Checkout_Button_Handler class contain its own logic where possible and not sprinkle it in where needed in the rest of the codebase.

I think I can concede here if we do add comments and if it helps to keep things clean :) 

> One question to this whole effort, that I haven't been able to nail down and was curious other's opinions on, is whether it would be preferable to keep all of our JS params under one umbrella (like wcpay_config) or is it preferable to break them down in to their own components as needed (like wcpay_config, wcpayPaymentRequestParams, and wcpayWooPayExpressParams)? The existing arrangement seemed to make sense for the time when there were only two (wcpay_config and wcpayPaymentRequestParams), but after I added a third I started to question it. Is there a preferred pattern to this?

@bborman22 

There isn't a preferred pattern, but separating them would be good IMO. It would help us avoid the question of filters altogether in situations like here. And duplicating some data shouldn't be a huge issue (especially if we think of these areas as extensions of WCPay)

PR #5238, opened 2022-12-02
Import fingerprint package directly instead of lazy loading

eduardoumpierre:
Fixes #5224 

#### Changes proposed in this Pull Request

- Remove lazy loading as it may be causing compatibility issues with script concatenators plugins.
- Improve the error message in case anything fails.

#### Testing instructions

* Unit tests should be passing
* Apply the following patch
```diff
diff --git a/client/checkout/utils/fingerprint.js b/client/checkout/utils/fingerprint.js
index 101a177d..354aa604 100644
--- a/client/checkout/utils/fingerprint.js
+++ b/client/checkout/utils/fingerprint.js
@@ -10,9 +10,9 @@ export const FINGERPRINT_GENERIC_ERROR =
 export const getFingerprint = async () => {
 	const agent = await FingerprintJS.load( { monitoring: false } );
 
-	if ( ! agent ) {
+	// if ( ! agent ) {
 		throw new Error( FINGERPRINT_GENERIC_ERROR );
-	}
+	// }
 
 	return await agent.get();
 };
```
* Enable UPE in **WCPay Dev**
* Go to your shop, add a product to your cart and open the checkout page
* It should show the `An error was encountered when preparing the payment form. Please try again later.` error message
* Also test the same flow for
  * UPE checkout - Block checkout
  * Non-UPE checkout - Classic checkout

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #5240, opened 2022-12-03
Revert "Import fingerprint package directly instead of lazy loading"

dwainm:
Reverts Automattic/woocommerce-payments#5238


alefesouza (approved):
LGTM

Issue #5117, opened 2022-11-11
Multiple UPE - Duplicate notices when saving SEPA methods


mdmoore:
### Describe the bug
When adding SEPA payment methods via **My Account > Payment Methods > Add Payment Method** success notices are duplicated several times. 

<img src="https://user-images.githubusercontent.com/3473953/201438232-7229b820-e8da-46a0-83b7-a38e8a8ad49d.png" width="500" />

### To Reproduce

1. On the `poc/upe-instances-multiplied` branch, enable SEPA Direct Debit
2. Navigate to **My Account > Payment Methods > Add Payment Method**
3. Add a SEPA payment method

### Acceptance Criteria

- Adding a SEPA payment method results in only a single success notice
- Changes made should be merged into `poc/upe-instances-multiplied`, not develop.
- [Callback functions](https://github.com/Automattic/woocommerce-payments/blob/0731300c7f9a04e7f92f9a031c80f7e9589d8518/includes/payment-methods/class-upe-payment-gateway.php#L111-L145) are moved to the `WC_Payments_UPE_Checkout` class and are added to the hook only once.

### Technical refinement
Please see the discussion below starting from [this](https://github.com/Automattic/woocommerce-payments/issues/5117#issuecomment-1318654165) comment, in order to get an idea of how to achieve the goal of this task.

mdmoore:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5117?planning-poker) with Zenhub @FangedParakeet @timur27

FangedParakeet:
I might have a few theories here...

@mdmoore, does this occur when adding the same, standard SEPA testing account (`AT611904300234573201`) repeatedly or does this still occur if you try any of the other payment methods listed [here](https://stripe.com/docs/testing#non-card-payments)?

mdmoore:
@FangedParakeet Your theories are all welcome!

>does this occur when adding the same, standard SEPA testing account (AT611904300234573201) repeatedly or does this still occur if you try any of the other payment methods listed [here](https://stripe.com/docs/testing#non-card-payments)?

The same is occurring with other test accounts as well. I've checked with `AT321904300235473204` and `AT861904300235473202`. I'm getting 10 notices each time, regardless of the number of payment methods I have.

timur27:
[...]

PR #5237, opened 2022-12-02
Allow cards to be saved via My Account

mdmoore:
Fixes #5116 

#### Changes proposed in this Pull Request

When the UPE is enabled, `upe.js` is enqueued which attempts to take over submissions of the Add Payment Method form. Because we're using the legacy card element, an error is thrown when the card payment method isn't found in the `gatewayUPEComponents` array. Because the normal, non-UPE, `checkout.js` is still enqueued, we can continue relying on it to handle payment method creation for cards from the My Account page.

#### Testing instructions

1. Enable the UPE in the WooCommerce Payments settings
2. Navigate to My Account > Payment Methods > Add Payment Method
3. Add a new credit card and verify that it was successful(no errors on page.)
4. Add a new UPE method(SEPA) and verify that it was successful(no errors on page.)

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #5246, opened 2022-12-05
Assign callbacks to gateway-specific actions in the abstract checkout class to avoid unnecessary duplication

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5117

This PR: 
1. Fixes the duplicated notices bug while saving the SEPA payment method (https://github.com/Automattic/woocommerce-payments/issues/5117)
2. Adds callbacks to hooks in the checkout class instead of UPE gateway class which is now initialized multiple times per each payment method. This allows us to avoid the situation when one callback is added to a hook multiple times, whereas only one addition is needed
3. Aligns the setup/payment intents removal methods to remove all the payment intents since we now are invoking them from the checkout class

##### :one: Test hook invocation, responsible for removing setup intent from the session. 

> 💡  Setup intent is removed [when loading the payment methods page](https://github.com/Automattic/woocommerce.com/blob/trunk/themes/woo/woocommerce/myaccount/payment-methods.php#L91) in _My account_. 

1. Go to _My account -> Payment methods_ and add a new SEPA payment method.
5. After adding and redirected back to the tokens list, go and add another SEPA payment method and ensure that there are no failures while adding the payment method. When there are problems with removing the setup intent, errors with the message `You cannot update this SetupIntent because it has already succeeded.` usually appear. In this step, we'd like to confirm that we don't see such errors. 
6. Repeat steps 1-2 with the legacy card.

##### :two: Test payment intent removal callback invocation
Confirm there are no errors in the two following use cases (does not matter if you save the payment method or not):
1. Use shortcode checkout to buy any product with a card twice in a row.
2. Use shortcode checkout to buy any product with the SEPA payment method twice in a row.

##### :three: Test fixed duplicated notices when saving SEPA
1. Go to _My account -> Payment methods_ and add a new SEPA payment method.
2. Ensure that there is only one `Payment method successfully added.` notice after being redirected to the payment method list.

##### :four: Test if appearance callbacks are invoked
To check if `save_upe_appearance_ajax` and `clear_upe_appearance_transient` callbacks are invoked: 
1. Set debug breakpoints in [save_upe_appearance_ajax()](https://github.com/Automattic/woocommerce-payments/blob/27426eaeeacba7002333f44830579afab0e6c729/includes/payment-methods/class-upe-payment-gateway.php#L891) and [clear_upe_appearance_transient()](https://github.com/Automattic/woocommerce-payments/blob/27426eaeeacba7002333f44830579afab0e6c729/includes/payment-methods/class-upe-payment-gateway.php#L924) methods.
2. Change the theme in the admin panel under _Appearance_ and ensure that the `clear_upe_appearance_transient` method is invoked and the debugger stops at the breakpoint. In my case, I switched from Storefront to Twenty Twenty-Two.
3. Go to the store, add a product and proceed to the checkout page.
4. On the shortcode checkout page loading, confirm that the `save_upe_appearance_ajax()` method is invoked and debugger stops at the breakpoint.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-upe-checkout.php
+++ b/includes/class-wc-payments-upe-checkout.php
@@ -71,6 +71,8 @@ class WC_Payments_UPE_Checkout extends WC_Payments_Checkout {
 
 		add_action( 'wc_payments_set_gateway', [ $this, 'set_gateway' ] );
 		add_action( 'wc_payments_add_upe_payment_fields', [ $this, 'payment_fields' ] );
+		add_action( 'woocommerce_after_account_payment_methods', [ $this->gateway, 'remove_upe_setup_intent_from_session' ], 10, 0 );
+		add_action( 'woocommerce_subscription_payment_method_updated', [ $this->gateway, 'remove_upe_setup_intent_from_session' ], 10, 0 );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


mdmoore (approved):
Everything is testing well except adding a card from **My Account > Payment Methods** in test 1.3, _Repeat steps 1-2 with the legacy card._. But I think that's solved in #5237. So everything it looking great here! :shipit:

PR #5247, opened 2022-12-05
Fix duplicate order: Store order ID to session, check it before update/create_confirm intents

### LINKED ISSUE #5190: Fix Duplicate Charge: Store intent keys in session, check before processing payments
### Description

Explore the possibility of an `wcpay_intent_key`, which is generated before the call to create intent. 

### Acceptance criteria

- We could store `wcpay_intent_key` in the session, and check if there is a recently paid order with the same key (while eventually comparing order contents with the cart as well). This would prevent both new orders, and new intents from being created.

- As an intermittent remedy, we could try forcing the checkout process to create a new order if we detect that the existing one has an `wcpay_intent_key`, but does not have an `intent_id`. Then at least the new order will have a separate intent, meaning that both orders will be refundable. But again, this applies to cases where:

1.  There is a communication error
2.  A webhook does not reach the client site timely
3.  UPE is not used

### Dev notes

Rado's PoC here: https://github.com/Automattic/woocommerce-payments/pull/5163. It needs to be thoroughly polished, but shows how we could prevent duplicate orders (which normally result) in duplicate payments as well, by checking if there is an existing order with the same randomly generated order key.

### For future: 

In the future, we might add this order key as a column on the server, and allow us to query for intents, based on this key. This will fix the only gap we see as remaining after the current set of fixes, which is when:

1.  there is a communication error somewhere between the client and Stripe during `create_and_confirm_intent`
2.  webhooks webhooks are not delivered (this is different from processed) on time

In that case the intent ID would have never been able to reach us before a merchant retries payment, resulting in duplicate intents, one of them orphaned.

### Additional context

paJDYF-6wC-p2#comment-15189

htdat:
Fixes #5190

- Save the current order ID to session data when processing payments.
- Clear it when the payment is successful. 
- In the next process_payment call, check this order (called session order) with the current processing order.
    - If two orders have the same cart hash, and the session order was paid, redirect customers to the order-received page of the session order.
    - Otherwise, let the payment process continue. 
- Add a notice like this in the redirected thank-you/order-received page (as it has the flag `wcpay_paid_for_previous_order`: 
Markup on 2022-12-24 at 17:55:49

### This approach is different from PoC #5163 in these aspects:

- Cookie vs session: 
    - The biggest advantage of using sessions is that it's attached to each customer rather than separate devices. With the cookie approach, if customers log in different devices, and then do checkout, we lose the key/info to check.  
    - We also do not need to work through the process to update our docs about the purpose of this 

- Only `order_id` vs a hash including `order_id + uuid`. I felt like using the later is pretty excessive, especially when the purpose of the hash is to... search for the order_id while it has already been there.

- Comparing `cart_hash` with the `$order_id` in `process_payment` rather than `WC()->cart`. 
    - In the `checkout` page, this `$order_id`'s `cart_hash` is exactly the `cart_hash` of `WC()->cart`. See `WC_Checkout::process_checkout` ([ref](https://github.com/woocommerce/woocommerce/blob/baa1c120bce6df9e06b80ecb876ffda8121ffd0d/plugins/woocommerce/includes/class-wc-checkout.php#L1256)) then `WC_Checkout::create_order` ([ref](https://github.com/woocommerce/woocommerce/blob/baa1c120bce6df9e06b80ecb876ffda8121ffd0d/plugins/woocommerce/includes/class-wc-checkout.php#L386-L401)).
    - This `$order_id` works with the order-pay page (i.e http://localhost/checkout/order-pay/123/?pay_for_order=true&key=wc_order_abcdefzy).
 

### Prep 

- Enable webhook in server. In your local server, it's usually `npm run listen`.
- Run `npm run watch` to build JS files.
- Feel free to test in multiple flows: UPE and non-UPE, classic and block checkout pages.

### Start 

- Add this snippet to your site https://gist.github.com/htdat/8971b603ea0ab4fe07474a7a7b30000c. You can add it with plugin `Code Snippets` or directly to the bottom of file `woocommerce-payments.php` in the root directory of this project.
- Add some products, visit to the checkout page.
- Visit http://localhost:8082/?dump_session - confirm that `Nothing stored`.
- Try to make an payment in the checkout page.
- See an error due to our snippet (trying to replicate potential issues that real customers can get): 
     - For non-UPE: See an error in the checkout page.
     - For UPE: see that the maybe_process_upe_redirect process was interrupted.

- Visit http://localhost:8082/?dump_session - it should display the progressing order_id, e.g. `xyz`. 
- Check in wp-admin -> Orders, this progressing order_id is precisely the newest order `xyz`. After a while, this order `xyz` should be changed from Failed to Processing/Paid as the `payment_intent.succedded` webhook arrived. 
- Go to the checkout page, try to process again.
- Get redirected to `order-received` page of order `xyz` with this parameter `wcpay_paid_for_previous_order=yes`, e.g.: http://localhost:8082/checkout/order-received/zyz/?key=wc_order_1b9oTbcXtWH0l&wcpay_paid_for_previous_order=yes
- Check this order `xyz` in WP Admin, see a note like this: `WooCommerce Payments: detected and deleted order ID abc, which has duplicate cart content with this order.` Note: `abc` would be a bigger number than `xyz`.
- Visit http://localhost:8082/?dump_session - confirm that `Nothing stored`.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.3.0#store-order-id-to-session-check-it-before-updatecreate_confirm-intents-pr-5247
- [x] <strike>Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable</strike>.


RadoslavGeorgiev:
@htdat this looks good so far, but as in my previous tries, the session (as it's related to the cart) seems to be a bit unpredictable. See this video as an example:
https://user-images.githubusercontent.com/5311119/207834860-3125aa31-e2cf-4732-a891-586b436783cc.mp4

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -73,7 +73,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 * @type int
 	 */
 
-	const USER_FORMATTED_TOKENS_LIMIT = 100;
+	const USER_FORMATTED_TOKENS_LIMIT  = 100;
+	const SESSION_KEY_PROCESSING_ORDER = 'wcpay_processing_order';
```
dmvrtx, line 85: Let's either add a comment to describe the constant or at least a whitespace, since constant on the line above belongs to a different context.

htdat: 👍 Addressed https://github.com/Automattic/woocommerce-payments/pull/5247/commits/4fd5e8b46c3e2571a26fdf4689ab178187afb0d3

htdat:
FYI. This E2E run failed consistently when testing a 3DS card: [WC - latest | wcpay - shopper](https://github.com/Automattic/woocommerce-payments/actions/runs/3762573488/jobs/6395387213).

> FAIL tests/e2e/specs/wcpay/shopper/shopper-checkout-save-card-and-purchase.spec.js (161.631s)
>   Saved cards 
>     when using a basic card added on checkout
>       ✓ should save the card (13667ms)
>       ✓ should process a payment with the saved card (9022ms)
>       ✓ should delete the card (3385ms)
>     when using a 3DS card added on checkout
>       ✓ should save the card (22232ms)
>       ❌  should process a payment with the saved card (100015ms)
>       ✓ should delete the card (3270ms)

I've pushed this change b77deb045c89c5d16304045f7879323cdbe4d382 to fix it. 

---- 

### The cause: 

For a 3DS card, it goes through this `requires_action` [route](https://github.com/Automattic/woocommerce-payments/blob/1fb31918265813803879d00b7793146794e3d7e1/includes/class-wc-payment-gateway-wcpay.php##L1183-L1206), and customers need to finalize the payment on the front-end, and then get redirected to the order-received page on the front-end too.

That means it does not go through [update_order_status_from_intent](https://github.com/Automattic/woocommerce-payments/blob/1fb31918265813803879d00b7793146794e3d7e1/includes/class-wc-payment-gateway-wcpay.php#L1210), which includes `remove_session_processing_order` for successful intents/payments. 

In the test flow above, after the first 3DS payment, i.e. `should save the card`, the order of this test is saved in WC Session and does not get removed. When the next payment comes with the same cart content, it redirects to the order paid in the first payment.

### The fix: 

To fix the issue, I am simply removing the order ID from the session when customers can reach out to order-received page, i.e. the payment is successful. This approach is pretty similar to the Woo core to clear its cart (based on WC Session) when arriving this order-received page - [Woo core code ref](https://github.com/woocommerce/woocommerce/blob/52640de58ae0b0be5fe0dfe5f9cffee94d33821b/plugins/woocommerce/includes/wc-cart-functions.php#L165-L196).

client/checkout/classic/upe.js
```diff
@@ -453,14 +453,18 @@ jQuery( function ( $ ) {
...
+			const updateResponse = await api.updateIntent(
 				paymentIntentId,
 				orderId,
 				savePaymentMethod,
 				$( '#wcpay_selected_upe_payment_type' ).val(),
 				$( '#wcpay_payment_country' ).val()
 			);
 
+			if ( updateResponse.data ) {
+				await api.handlePreviousOrderPaid( updateResponse.data );
```
RadoslavGeorgiev: No need for `await` here, `handlePreviousOrderPaid` is not asynchronous.

However, it is also worth returning something from `handlePreviousOrderPaid`, checking it here and bailing if an order has been handled. Otherwise you risk execution continuing after `window.location = ...`. In this case there's just `api.handlePaymentConfirmation`, and that would likely be able to continue, even though it won't accomplish anything. There's still the chance of quirky behavior though.

client/checkout/blocks/upe-fields.js
```diff
@@ -272,6 +272,8 @@ const WCPayUPEFields = ( {
 			onCheckoutAfterProcessingWithSuccess(
 				( { orderId, processingResponse: { paymentDetails } } ) => {
 					async function updateIntent() {
+						await api.handlePreviousOrderPaid( paymentDetails );
```
RadoslavGeorgiev: Similarly to the previous comment, no need for `await` here, but let's check if an order was handled before letting this method continue execution.

includes/class-wc-payments-order-success-page.php
```diff
@@ -68,6 +69,23 @@ public function show_woopay_payment_method_name( $order_id ) {
...
+	 *
+	 * @param string $text  the default thank you text.
+	 *
+	 * @return string
+	 */
+	public function add_notice_previous_paid_order( $text ) {
+		if ( isset( $_GET[ WC_Payment_Gateway_WCPay::FLAG_PREVIOUS_ORDER_PAID ] ) ) { // phpcs:disable WordPress.Security.NonceVerification.Recommended
+			$text .= sprintf(
+				'<div class="woocommerce-info">%s</div>',
+				esc_attr__( 'This already paid order has the same content with the one you have just tried to make payment. Please checkout again if you want to have another order with the same content.', 'woocommerce-payments' )
```
RadoslavGeorgiev: I'd change the wording a bit here. `Please checkout again` feels a bit aggressive somehow, check number 9 on [this page](https://idioms.thefreedictionary.com/check+out) :D

How about something like this:

> We detected and prevented an attempt to pay for a duplicate order. If this was a mistake and you wish to try again, please create a new order.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1880,6 +1905,109 @@ protected function get_account_branding_icon( $default_value = '' ): string {
...
+
+	/**
+	 * Action to remove the order ID when customers reach its order-received page.
+	 *
+	 * @return void
+	 */
+	public function clear_session_processing_order_after_landing_order_received_page() {
+		global $wp;
+
+		if ( is_order_received_page() && isset( $wp->query_vars['order-received'] ) ) {
```
RadoslavGeorgiev, line 2005: Instead of using a global here, could you try `get_query_var()` instead?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1880,6 +1905,109 @@ protected function get_account_branding_icon( $default_value = '' ): string {
...
+			$order_id = absint( $wp->query_vars['order-received'] );
```
RadoslavGeorgiev, line 2006: A small note, nothing to fix: I was wondering what happens if the `woocommerce_order_number` filter is used to change the order number, and if that might break here, but the URL still includes the order ID instead of number.

RadoslavGeorgiev (approved):
Changes look good!

I have a couple of nitpicks, nothing significant. Pre-approving the PR on my side, as I will be AFK next week. Keep in mind that I don't have the opportunity for thorough manual testing right now.

includes/class-wc-payment-gateway-wcpay.php
htdat: [get_query_var](https://github.com/WordPress/WordPress/blob/bf344bb3bf7f2540355c6e7d26ff09c29fdd9cb3/wp-includes/query.php#L27-L27) actually fetches the value from `global $wp_query`. That is different from `global $wp`, which is to [set up WP environment](https://developer.wordpress.org/reference/classes/wp/).

Also, I used `$wp->query_vars['order-received']` as it's used in Woo core function [is_order_received_page](https://github.com/woocommerce/woocommerce/blob/52640de58ae0b0be5fe0dfe5f9cffee94d33821b/plugins/woocommerce/includes/wc-conditional-functions.php#L218-L224).

includes/class-wc-payment-gateway-wcpay.php
htdat: Good point. Marking as resolved.

includes/class-wc-payments-order-success-page.php
htdat: 😬 Very good point! Addressed here a8709cf44ffd29e36b730f658424c5a98e857ddb

client/checkout/classic/upe.js
htdat: > No need for await here, handlePreviousOrderPaid is not asynchronous.

I used `await` to avoid the risk you mentioned. That is, when the redirection has not finished yet, the code continues to confirm payment, and yell an error like the following.

Markup on 2022-12-23 at 17:57:47

But yeah, it does not sound the correct way to handle this. 

> However, it is also worth returning something from handlePreviousOrderPaid, checking it here and bailing if an order has been handled.

This is the right way. Addressed here a3aa50bf648abbf99c38e99e4f424f2a8468f0a5

htdat: Marking this conversation as resolved.

client/checkout/blocks/upe-fields.js
htdat: See https://github.com/Automattic/woocommerce-payments/pull/5247#discussion_r1056265961

htdat:
@RadoslavGeorgiev - addressed your feedback. It's great if you can have another look. Otherwise, I will merge it tomorrow (Saturday) so that it can be tested by GlobalStep.

PR #5252, opened 2022-12-06
Latest develop merge to POC branch

timur27:
Another PR to resolve conflicts between `develop` and `poc/upe-instances-multiplied`.

#### Conflicts resolved in:
client/checkout/blocks/fields.js
client/checkout/blocks/upe-fields.js

Conflicts were not difficult to resolve, and nothing should modify the behavior we are working on.

#### Testing instructions

1. Buy a product using the Blocks checkout page, and the UPE turned off 
2. Buy a product using the Blocks checkout page, and the UPE turned on  

[E2E tests are green](https://github.com/Automattic/woocommerce-payments/actions/runs/3629931353).

-------------------

**Post merge**


FangedParakeet (commented):
The additions here look inoffensive to me. I just finished running through a bunch of usual checkout flows on the shortcode and blocks checkouts and almost everything worked exactly as you'd expect it to. I did encounter one issue however, but I'm not sure if this already an identified issue or not. 

If I try to complete a checkout using the blocks checkout with a saved SEPA payment method, the checkout fails and shows the error message "No payment method provided."

Saved SEPA payment method selected in blocks checkout
_Choose a saved SEPA payment method..._

No payment method provided error message
_Nani?!_

Server error response
_Error response from request to `/wc/store/v1/checkout`_

I remember seeing this before, but I can't remember whether we resolved this or not. @mdmoore, do you happen to know if this is still an identified bug and whether we have an open issue to capture it already? @timur27, would you mind quickly double-checking my homework to confirm whether you can replicate this issue on this branch? Cheers!

timur27:
@FangedParakeet good catch! 

I confirm that I observe the same behavior. I can replicate it on both `merge/develop-to-poc` and `poc/upe-instances-multiplied` branches.

timur27:
I think I found something else that's worth taking a closer look at.

@FangedParakeet could you share with us how did you save the SEPA payment method which you used afterward to make a purchase? Would I guess this payment method was added in _My account -> Payment methods_? 

I think I found another issue, which is reproducible on our feature branch and is not a regression - I took the latest `poc/upe-instances-multiplied` and a random commit from November - both allow me to reproduce the issue. Basically, SEPA payment methods are not saved from Blocks checkout. More details are below. 

#### How to reproduce
1. Buy a product using Blocks checkout and SEPA. Save SEPA during the process.
2. Open the Blocks checkout page again and check if you see the SEPA method saved previously. I expect it to appear, but it doesn’t. 🔴 
3. Open the Shortcode checkout page and check if you see the SEPA method saved previously. I expect it to appear, but it doesn’t. 🔴
4. Go to My account -> Payment methods and check if you see the SEPA method on the list. It’s not displayed there, and the list is empty. 🔴 
5. Add a new SEPA payment method in _My account -> Payment methods_ OR by saving the new SEPA method on the shortcode checkout page. After being added, it should appear on the list. Also, the previously added SEPA method starts to appear on the list after this action.

#### Where the issue happens? 
When invoking `get_payment_methods_for_customer`, the value of `$cache_payment_methods` is always true [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L212). Because of that, [we are always trying to get SEPA payment methods from the database cache](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L215-L220) and it returns `0` entries. 

At the same time, while debugging, I executed [the fetch from payments api client](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L223), as if the `$cache_payment_methods` flag would be false, and I received the SEPA methods, which are not visible after adding. 

As mentioned in step 5 above, adding another payment method from within _My account -> Payment methods_ OR shortcode checkout helps. In practice, when opening Blocks checkout then, the result of [this part of the code](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L207-L220) remains the same (we still have `cache_payment_methods` flag set to `true` and we still get no payment methods [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L216)), but the difference is, that `$tokens` array already has all the saved SEPA methods provided [when entering the woocommerce_get_customer_payment_tokens method](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L110).

#### Next steps?
I want to confirm if we don't have this bug covered in any existing issues. If we don't, and you confirm the same behavior in your environment, we need to create a ticket for this. Besides fixing the bug, I'd say it's worth adding a new e2e test to cover UPE payment method savings. It might seem like too much, but given that we are working actively with the token implementation, I think this can save us some time in the future.

mdmoore:
@FangedParakeet 
>do you happen to know if this is still an identified bug and whether we have an open issue to capture it already?

Checking out with saved SEPA methods was working in earlier tests, but I'm also able to reproduce this now. There's not an open issue currently. 

>I took the latest poc/upe-instances-multiplied and a random commit from November - both allow me to reproduce the issue. Basically, SEPA payment methods are not saved from Blocks checkout. More details are below.

I'm not able to reproduce this for some reason. I first chose a couple commits from early November and found that SEPA payment methods saved via a Blocks checkout were indeed appearing when returning to Blocks checkout immediately after completing a checkout. I thought it might be a regression but upon testing on the current `poc/upe-instances-multiplied`, SEPA payment methods are stored and appear on checkout correctly. I'm not sure how to explain this 🤔.

FangedParakeet:
> I think I found another issue, which is reproducible on our feature branch and is not a regression - I took the latest poc/upe-instances-multiplied and a random commit from November - both allow me to reproduce the issue. Basically, SEPA payment methods are not saved from Blocks checkout. 

Cheers, @timur27, I was able to replicate this error following your steps. I was indeed unable to save SEPA payment methods from the Blocks checkout. I have a sneaking suspicion why this might have occurred, but let me first walk you through my deductions.

> When invoking get_payment_methods_for_customer, the value of $cache_payment_methods is always true [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php?rgh-link-date=2022-12-07T14%3A03%3A35Z#L212). Because of that, [we are always trying to get SEPA payment methods from the database cache](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php?rgh-link-date=2022-12-07T14%3A03%3A35Z#L215-L220) and it returns 0 entries.

This is true--`$cache_payment_methods` is always `true`, since we want to use the cache as much as possible. In order to make sure that we do collect new payment methods from Stripe, whenever we delete a payment method, set a new default payment method, or add a new payment method we must ensure that we clear the cache (see [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments-token-service.php#L205), [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments-token-service.php#L225), and [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments-token-service.php#L67) respectively), so that the next time we try to check the cache, [we will find it empty](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments-customer-service.php#L218) and be forced to use the API client to request fresh payment methods from Stripe instead.

What is currently happening is that we are not clearing the cache after adding a new payment method from the Blocks checkout. You can confirm this by adding a new SEPA payment method from the Blocks checkout and then proceeding immediately to your customer Account dashboard. You should be disappointed to see that the payment method you earlier saved is not listed as an available payment token. However, delete one of your existing WC payment tokens or set a new default payment token--in order to trigger the cache refresh--and like magic your stored payment tokens will now appear.

In the UPE checkout flow, this cache refresh is eventually triggered by this [block of code](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L664-L667), when the `save_payment_method` parameter is present on the return, received order URL with a value of `"yes"`. If you complete a blocks checkout, while selecting that you would like to save a SEPA payment method, you will be shocked to discover that this parameter is in fact set to `"no"`.

Return URL parameter save_payment_method=no
_Stripe return URL, j'accuse!_

This parameter should be added to the return URL sent to Stripe [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L532); however, I believe this is not occurring because we are setting this value incorrectly over [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L448).

```php
$save_payment_method = $payment_type->equals( Payment_Type::RECURRING() ) || ! empty( $_POST[ 'wc-' . self::GATEWAY_ID . '_' . $this->stripe_id . '-new-payment-method' ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```

I'm pretty sure the parameter `wc-woocomerce_payments_sepa_debit-new-payment-method` will only be present on the shortcode checkout, which means that this is always set to `false` on the Blocks checkout. 

Comparing save payment method checkbox on shortcode and blocks checkout
_Shortcode checkout on the left, Blocks on the right--note the differing name on the input_

We basically need to fix this (presumably [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L448)) to ensure that we pass this parameter correctly and clear the cache upon adding a new payment method.

**But wait there's more!**

The original issue that I was complaining about was that I was unable to get even this far when trying to complete a checkout with a SEPA payment method that I had _already_ saved as a payment token and I'm _still_ unable to do this--forever foiled by the error message "No payment method provided".

I'm not sure that this is related to the above issue, but it could be. I also don't see an obvious place in these merge changes where this second error might have been introduced. I think we should create one--or possibly two--new issues to track these resolutions and in the meantime, I will gladly approve this PR.

Let me know if that sounds like a good plan, @timur27. Also, @mdmoore let me know if you have any thoughts on how to resolve either of these issues, because I feel like I remember you tackling something similar before when dealing with our initial saved payment method compatibility.

timur27:
Brilliant, @FangedParakeet, thanks for your time investigating this! 

@mdmoore @FangedParakeet I will go and create two issues.

UPDATE: issues created: https://github.com/Automattic/woocommerce-payments/issues/5260, https://github.com/Automattic/woocommerce-payments/issues/5261

FangedParakeet (approved):
LGTM! Two problems discovered and explained in depth above, with issues forthcoming, but for now I'll do my part by approving this PR, while feeling like the gladdest man alive. :shipit:

mdmoore:
>I was able to replicate this error following your steps. I was indeed unable to save SEPA payment methods from the Blocks checkout. I have a sneaking suspicion why this might have occurred, but let me first walk you through my deductions.

So I was still unable to reproduce this when testing further today, but I realized you both might be testing with no saved payment methods present. Because I had saved tokens, the cache is being cleared when getting those tokens and the saved SEPA method I had just added was appearing, as if everything was working as expected. 

>We basically need to fix this (presumably [here](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/payment-methods/class-upe-payment-gateway.php#L448)) to ensure that we pass this parameter correctly and clear the cache upon adding a new payment method.

I agree that this is the issue and is that same problem I ran into in the shortcode checkout. Thanks for creating those issues @timur27!

PR #5254, opened 2022-12-06
Return `_payment_method_id` for subscription payment meta.

### LINKED ISSUE #128: Active Subscriptions not updated when changing default payment method
## Describe the bug

Updating the default payment method and choosing `Yes` to update active subscriptions to the new payment method does not update the subscription payment method as expected.

## To Reproduce

1. Create a new subscription order.
2. As a customer, add a new payment method.
3. Make the new payment method the default.
4. When prompted, select `Yes` to update subscriptions to the new payment method.
5. See payment method for the subscription is not updated.

## Product impact

- [x] Does this issue affect WooCommerce Payments? Yes https://github.com/Automattic/woocommerce-payments/issues/3729 
- [x] Does this issue affect WooCommerce Subscriptions? No

## Screenshots
Clicking `Make Default` on the Amex card ending in 0005. Presents the prompt.
Screen Shot 2022-03-23 at 12 59 15 PM
Clicking `Yes` does not update the payment method for the subscription
Screen Shot 2022-03-23 at 12 59 35 PM

shendy-a8c:
Fixes https://github.com/Automattic/woocommerce-subscriptions-core/issues/128.

#### Testing instructions

* Create a new subscription order.
* As a customer, add a new payment method.
* Make the new payment method the default.
* When prompted "Would you like to update your subscriptions to use this new payment method", select Yes to update subscriptions to the new payment method.
* Go to My Account > Subscriptions > See payment method for all the subscriptions are updated as expected.

note:
- I notice a bug (or maybe by design) and plan to create a separate issue for it that settings a default payment method only applies to subscriptions with payment token of the same gateway. I see that it's because gateway id is passed to `WCS_Payment_Tokens::get_customer_tokens()` [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wcs-my-account-payment-methods.php#L172).
- Also, I notice a bug and plan to create a separate issue for it that when all payment methods are deleted, then adding a new payment won't update the existing subscriptions' payment token. I think it's because we look for subscriptions based on existing tokens [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wcs-my-account-payment-methods.php#L222-L232). When payments are deleted and new tokens are added, this logic fails because it assumes subscriptions' payment tokens are part of the existing tokens.
- I am also still working on adding e2e test for this PR change.

-------------------

**Post merge**


mattallan:
Hey @shendy-a8c, thanks for drafting this PR!

I just left a comment on https://github.com/Automattic/woocommerce-subscriptions-core/pull/342 with some more details on how I think we should address this issue in WC Payments, but in short, I don't think we should update our `get_payment_meta()` function as it's used in other subscription payment method related flows.

What I'm thinking we do is:

1. Hook onto `woocommerce_subscriptions_update_subscription_token`
2. Check if we're updating a WC Payments token
3. If we are, run:

```
$subscription->update_meta_data( '_payment_method_id', $new_token->get_token() );
$subscription->add_payment_token( new_token );
$subscription->save();
```

4. Return true if updated.

This approach makes sure the proper meta data that WC Payments expects is added to the subscription as well as the token is saved and doesn't interfere with other payment gateways.

Thoughts?

---

Example code:

```
add_filter( 'woocommerce_subscriptions_update_subscription_token', 'update_subscription_token', 10, 4 );

function update_subscription_token( $updated, $subscription, $new_token, $old_token ) {
	if ( $this->id !== $subscription->get_payment_method() ) {
		return false;
	}

	$subscription->update_meta_data( '_payment_method_id', $new_token->get_token() );
	$subscription->add_payment_token( $new_token );
	$subscription->save();

	return true;
}
```

mattallan (approved):
@shendy-a8c I've tested this and I've confirmed it fixes the issues!
I'm happy for this to be merged.

Note: The new WC Pay 5.2 release hasn't been tagged yet (there are issues with the workflow), so we might need to hold off merging this one until after that is fixed. See internal slack thread here for discussion: p1671408686150429-slack-CGGCLBN58

If we do want to get in 5.2, we can merge it now. I'll leave that up to you to decide 😄

mattallan:
While testing this PR in HPOS environments, i noticed the payment methods still weren't updating. I've pushed up a fix for this in https://github.com/Automattic/woocommerce-subscriptions-core/pull/349

shendy-a8c:
> While testing this PR in HPOS environments, i noticed the payment methods still weren't updating. I've pushed up a fix for this in https://github.com/Automattic/woocommerce-subscriptions-core/pull/349

Woah. Thank you for the quick review and fix for HPOS.

> If we do want to get in 5.2, we can merge it now. I'll leave that up to you to decide 😄

Let's do it. I'm merging it now. I can do the e2e test in a separate PR.

PR #5257, opened 2022-12-08
Allow i18n for UPE checkout blocks error message

### LINKED ISSUE #4612: Hard-coded "Your payment information is incomplete" cannot be translated.
### Description
Any user wanting to translate the field "Your payment information is incomplete", cannot do so with a translation plugin, and would have to hard-code this for every WooCommerce Payments update.

image

### Acceptance criteria
Make this string translation friendly.

### Designs
None required.

### Testing instructions
1. Install any Translation extension.
1. Try and find the string under WooCommerce, WooCommerce Payments or WooCommerce Blocks.
1. Fail in your search.
1. Resort to hard-coding a solution.

### Dev notes
For reference, the core JS files here are minified in the live version of the plugin:
image

### Additional context
Internal ticket `#5497343`

vladolaru:
Fixes #4612 

**Title:** Allow i18n for UPE checkout blocks error message

**Description:** 

The "Your payment information is incomplete." message string displayed by the UPE checkout blocks could not be translated via the `.pot` file (either through using an editor or a dedicated translation plugin).

I wrapped the string in a `__()` call so it is picked up by the `.pot` generation task and included in the `languages/woocommerce-payments.pot` file.

#### Testing instructions

1. Install any translation plugin (e.g. [Loco Translate](https://wordpress.org/plugins/loco-translate/)).
2. Add a new language to your site via the installed translation plugin.
3. Search the string `Your payment information is incomplete.` under the **WooCommerce Payments** plugin or the `woocommerce-payments` text domain (depending on how your chosen translation plugin works).
4. You should be able to add your translation of it.

Here is how it looks in Loco Translate:
LocoTranslate-4612-issue

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is insignificant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests - no need for tests
- [x] Tested on mobile - does not apply

**Post merge**

### DIFF
--- a/changelog/fix-4612-checkout-blocks-upe-error-message-i18n
+++ b/changelog/fix-4612-checkout-blocks-upe-error-message-i18n
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Allow the "Your payment information is incomplete." UPE checkout block message to be translated.
--- a/client/checkout/blocks/upe-fields.js
+++ b/client/checkout/blocks/upe-fields.js
@@ -15,6 +15,7 @@ import {
 } from '@woocommerce/blocks-registry';
 import { useEffect, useState } from '@wordpress/element';
 import { useDispatch, useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
 
 /**
  * Internal dependencies
@@ -210,7 +211,10 @@ const WCPayUPEFields = ( {
 				if ( ! isUPEComplete ) {
 					return {
 						type: 'error',
-						message: 'Your payment information is incomplete.',
+						message: __(
+							'Your payment information is incomplete.',
+							'woocommerce-payments'
+						),
 					};
 				}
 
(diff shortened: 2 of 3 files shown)


ismaeldcom (commented):
Thanks for working on it, the change looks good! Although the change is only applied to UPE blocks checkout. To make it consistent, I think we should also apply the same change to the classic one here:
https://github.com/Automattic/woocommerce-payments/blob/de290e8795bbcfa8dd9fe989323800b875681158/client/checkout/classic/upe.js#L398

And going a bit further, now that we allow to translate that error message. What about the other ones? For example `This payment method can not be saved for future use`. I think we should aim to make all the checkout strings translatable.

It's something out of the scope of the issue, but what do you think about creating another issue for it?

vladolaru:
Good catch about the classic checkout @ismaeldcom. I pushed a commit with the fix for that.

Regarding the rest of the strings that are not translatable around the checkout functionality, yes, I noticed those also. I will open a more general issue to fix i18n issues around checkout and get cranking on it.

ismaeldcom (approved):
Good! Let's merge it 🙂

PR #5276, opened 2022-12-12
Filter Stripe Link payment method availability check by country

### LINKED ISSUE #5278: Stripe Link is enabled for merchants outside the US
### Describe the bug
[Stripe Link](https://woocommerce.com/document/woocommerce-payments/payment-methods/link-by-stripe/) is supposed to be available to US merchants only. But we have received reports from Happiness where the Link payment method was enabled for non-US merchants. This has led to issues like the No base fee applied for account error and double fees deduction.

### Additional context
paJDYF-6BT-p2

mordeth:
Fixes #5278

Update Stripe Link availability checks by available Link fees to prevent displaying Stripe Link as an available payment method in non-US countries.

#### Testing instructions
- Set Stripe Link capability to `available`.
- Use a non-US connected account.
- Verify the Stripe Link is not available on the Checkout page.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2832,6 +2832,19 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	}
 
 	/**
+	 * Returns the list of enabled payment method types that will function with the current checkout filtered by fees.
+	 *
+	 * @param string $order_id optional Order ID.
+	 * @param bool   $force_currency_check optional Whether the currency check is required even if is_admin().
+	 * @return string[]
+	 */
+	public function get_payment_method_ids_enabled_at_checkout_filtered_by_fees( $order_id = null, $force_currency_check = false ) {
+		return [
+			'card',
+		];
+	}
+
+	/**
 	 * Returns the list of available payment method types for UPE.
 	 * See https://stripe.com/docs/stripe-js/payment-element#web-create-payment-intent for a complete list.
 	 *
@@ -2922,7 +2935,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	public function checkout_update_email_field_priority( $fields ) {
 		$is_link_enabled = in_array(
 			Link_Payment_Method::PAYMENT_METHOD_STRIPE_ID,
-			\WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( null, true ),
+			\WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout_filtered_by_fees( null, true ),
 			true
 		);
 
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2839,9 +2839,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 * @return string[]
 	 */
 	public function get_payment_method_ids_enabled_at_checkout_filtered_by_fees( $order_id = null, $force_currency_check = false ) {
-		return [
-			'card',
-		];
+		return $this->get_payment_method_ids_enabled_at_checkout( $order_id, $force_currency_check );
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2831,6 +2831,19 @@ public function get_payment_method_ids_enabled_at_checkout( $order_id = null, $f
...
 
+	/**
+	 * Returns the list of enabled payment method types that will function with the current checkout filtered by fees.
+	 *
+	 * @param string $order_id optional Order ID.
+	 * @param bool   $force_currency_check optional Whether the currency check is required even if is_admin().
+	 * @return string[]
+	 */
+	public function get_payment_method_ids_enabled_at_checkout_filtered_by_fees( $order_id = null, $force_currency_check = false ) {
+		return [
```
kalessil: Can we replace it with `return $this->get_payment_method_ids_enabled_at_checkout( $order_id, $force_currency_check )`? Tthis way, we maintain a relation to the source of payment methods.

kalessil (approved):
LGTM! I did a code review, though - it looks aligned with Fusion's decision in the corresponding P2. 

Grumpy mode: Though Stripe can, in the future, enable capabilities for PMs where the server is not ready yet, we'll have the same issue - the current solution aims for Link only.

nagpai:
If there is a ticket/ issue associated with this PR, we may want to add the number . Else we can remove the first line in the description.

>Fixes #

includes/class-wc-payment-gateway-wcpay.php
anu-rock: Agree with the above comment.

anu-rock: Do we have this method to maintain a 1:1 mapping with the UPE gateway class?

anu-rock: Never mind. I see that the UPE gateway inherits from the WCPay gateway class.

dpaun1985 (approved):
✅ LGTM.
✅ Tested and works as expected. 
Also, I agree with the above suggestion from @kalessil.

anu-rock:
I did a code-only review yesterday and found everything to be all right. There's Vlad's comment that we should probably address. Apart from that, I am still looking for a way to successfully test the changes manually.

includes/class-wc-payment-gateway-wcpay.php
anu-rock: Vlad's comment was addressed in dde270a.

anu-rock:
I did some thorough testing. It seems the fix works only partially. Here's what I did.

#### Store setup

Woo store location: US
Store currency: USD
Link checkbox enabled on Payments → Settings page *

\* It seems to be a prerequisite. In recent plugin versions, non-US accounts cannot see the Link checkbox. So, I assume a merchant affected by the issue had this checkbox enabled while using a version before #4705.

#### Testing steps

1. While my site was linked to a US-based, I enabled the Link checkbox on the Payments → Settings page.
2. I set my site to use a non-US account with Link capability enabled in the Stripe dashboard.
3. On `develop` - the email field shows on top of the checkout form, and the Link button/form also appears.
4. On `fix/stripe-link-country-check` - the email field **no longer** shows on top of the checkout form, but the Link button/form **still appears**. See the screen recording below.

https://user-images.githubusercontent.com/1288616/207593727-135f13f7-b179-43b0-a9f1-2701a16d6a47.mov

anu-rock:
@mordeth - Thanks for baking in 304f7a8. I can confirm that I no longer see the Link sign-up form with the new changes.

anu-rock (approved):
✅ Code changes LGTM
✅ Tested successfully after the latest changes

Also noting that we should create a separate issue to deal with @kalessil's following suggestion:

> Grumpy mode: Though Stripe can, in the future, enable capabilities for PMs where the server is not ready yet, we'll have the same issue - the current solution aims for Link only.

mordeth:
> Grumpy mode: Though Stripe can, in the future, enable capabilities for PMs where the server is not ready yet, we'll have the same issue - the current solution aims for Link only.

 #5297 issue was created as a follow-up regarding the problem.

PR #5281, opened 2022-12-13
Dev/paginated request class

### LINKED ISSUE #5058: WCPay Platform: Base API request Dispute and Transaction request
### Description
Add Dispute and Transaction endpoints using WcPay Request class.

### Acceptance criteria
Before you start with issue, make sure to check this [issue](https://github.com/Automattic/woocommerce-payments/issues/5056) before. When this issue is merged, use the new classes that were added by this issue and create `Dispute` and `Dispute` requests using the base classes as foundations. 

### Testing instructions

Cover all requests with unit tests.

### Dev notes

1. Make sure that every request have getters and setters
2. Make sure that every argument or property is validated
3. Make sure immutability of properties that doesn't need to be changed.
4. Cover everything with tests.

zmaglica:
Fixes #5058 

#### Add Paginated request classes

New paginated request class have been created. Using this class, you can easily execute paginated API call to wc pay server.

Here is the example of paginated request:

```
 $page      = 2;
$page_size = 50;
$direction = 'asc';
$sort      = 'created';
$filters   = [ 'key'  => 'value' ];
$uri       = 'uri';

$request = Paginated::create();
$request->set_page( $page );
$request->set_page_size( $page_size );
$request->set_sort_direction( $direction );
$request->set_sort_by( $sort );
$request->set_filters( $filters );
$request->set_uri( $uri );
```
This class will return the same response from wc pay server. For simple requests where you need results from the server without formatting or modeling response data, use this paginated request. Since there could be plenty of similar requests where URI or URL is the only difference, the `set_uri` method provide an easy way to define API URI.

There are `List_Disputes` and `List_Transactions` request classes that are similar like paginated request class, but the only difference is that these classes format responses.

#### Testing instructions

* Make sure that all tests are passing
* Test new request classes on your environment and make sure that everything is working as expected.

-------------------

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-payments-authorizations-controller.php
+++ b/includes/admin/class-wc-rest-payments-authorizations-controller.php
@@ -5,7 +5,7 @@
  * @package WooCommerce\Payments\Admin
  */
 
-use WCPay\Core\Server\Request\Paginated as Paginated_Request;
+use WCPay\Core\Server\Request\List_Authorizations;
 
 defined( 'ABSPATH' ) || exit;
 
@@ -60,12 +60,7 @@ class WC_REST_Payments_Authorizations_Controller extends WC_Payments_REST_Contro
 	 * @param WP_REST_Request $request Full data about the request.
 	 */
 	public function get_authorizations( WP_REST_Request $request ) {
-		$wcpay_request = Paginated_Request::create();
-		$wcpay_request->set_uri( WC_Payments_API_Client::AUTHORIZATIONS_API );
-		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
-		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
-		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
-		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
+		$wcpay_request = List_Authorizations::from_rest_request( $request );
 
 		return $wcpay_request->handle_rest_request( 'wcpay_list_authorizations_request' );
 	}
--- a/includes/admin/class-wc-rest-payments-authorizations-controller.php
+++ b/includes/admin/class-wc-rest-payments-authorizations-controller.php
@@ -5,6 +5,8 @@
  * @package WooCommerce\Payments\Admin
  */
 
+use WCPay\Core\Server\Request\Paginated as Paginated_Request;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -58,11 +60,14 @@ class WC_REST_Payments_Authorizations_Controller extends WC_Payments_REST_Contro
 	 * @param WP_REST_Request $request Full data about the request.
 	 */
 	public function get_authorizations( WP_REST_Request $request ) {
-		$page      = (int) $request->get_param( 'page' );
-		$page_size = (int) $request->get_param( 'pagesize' );
-		$sort      = $request->get_param( 'sort' );
-		$direction = $request->get_param( 'direction' );
-		return $this->forward_request( 'list_authorizations', [ $page, $page_size, $sort, $direction ] );
+		$wcpay_request = Paginated_Request::create();
+		$wcpay_request->set_uri( WC_Payments_API_Client::AUTHORIZATIONS_API );
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
+
+		return $wcpay_request->handle_request( 'wcpay_list_authorizations_request' );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-payments-authorizations-controller.php
+++ b/includes/admin/class-wc-rest-payments-authorizations-controller.php
@@ -67,7 +67,7 @@ class WC_REST_Payments_Authorizations_Controller extends WC_Payments_REST_Contro
 		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
 		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
 
-		return $wcpay_request->handle_request( 'wcpay_list_authorizations_request' );
+		return $wcpay_request->handle_rest_request( 'wcpay_list_authorizations_request' );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-payments-charges-controller.php
+++ b/includes/admin/class-wc-rest-payments-charges-controller.php
@@ -5,6 +5,7 @@
  * @package WooCommerce\Payments\Admin
  */
 
+use WCPay\Core\Server\Request\Get_Charge;
 use WCPay\Exceptions\API_Exception;
 
 defined( 'ABSPATH' ) || exit;
@@ -45,7 +46,9 @@ class WC_REST_Payments_Charges_Controller extends WC_Payments_REST_Controller {
 		$charge_id = $request->get_param( 'charge_id' );
 
 		try {
-			$charge = $this->api_client->get_charge( $charge_id );
+			$wcpay_request = Get_Charge::create();
+			$wcpay_request->set_charge_id( $charge_id );
+			$charge = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
 		} catch ( API_Exception $e ) {
 			return rest_ensure_response( new WP_Error( 'wcpay_get_charge', $e->getMessage() ) );
 		}
--- a/includes/admin/class-wc-rest-payments-charges-controller.php
+++ b/includes/admin/class-wc-rest-payments-charges-controller.php
@@ -46,9 +46,8 @@ class WC_REST_Payments_Charges_Controller extends WC_Payments_REST_Controller {
 		$charge_id = $request->get_param( 'charge_id' );
 
 		try {
-			$wcpay_request = Get_Charge::create();
-			$wcpay_request->set_charge_id( $charge_id );
-			$charge = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
+			$wcpay_request = Get_Charge::create( $charge_id );
+			$charge        = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
 		} catch ( API_Exception $e ) {
 			return rest_ensure_response( new WP_Error( 'wcpay_get_charge', $e->getMessage() ) );
 		}
--- a/includes/admin/class-wc-rest-payments-charges-controller.php
+++ b/includes/admin/class-wc-rest-payments-charges-controller.php
@@ -47,7 +47,7 @@ class WC_REST_Payments_Charges_Controller extends WC_Payments_REST_Controller {
 
 		try {
 			$wcpay_request = Get_Charge::create( $charge_id );
-			$charge        = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
+			$charge        = $wcpay_request->send( 'wcpay_get_charge_request' );
 		} catch ( API_Exception $e ) {
 			return rest_ensure_response( new WP_Error( 'wcpay_get_charge', $e->getMessage() ) );
 		}
--- a/includes/admin/class-wc-rest-payments-disputes-controller.php
+++ b/includes/admin/class-wc-rest-payments-disputes-controller.php
@@ -94,7 +94,7 @@ class WC_REST_Payments_Disputes_Controller extends WC_Payments_REST_Controller {
 		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) ?? 'desc' );
 		$wcpay_request->set_filters( $this->get_disputes_filters( $request ) );
 
-		return $wcpay_request->handle_request( 'wcpay_list_disputes_request', $request );
+		return $wcpay_request->handle_rest_request( 'wcpay_list_disputes_request', $request );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-payments-disputes-controller.php
+++ b/includes/admin/class-wc-rest-payments-disputes-controller.php
@@ -94,7 +94,7 @@ class WC_REST_Payments_Disputes_Controller extends WC_Payments_REST_Controller {
 		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) ?? 'desc' );
 		$wcpay_request->set_filters( $this->get_disputes_filters( $request ) );
 
-		return $wcpay_request->handle_request( 'wcpay_list_disputes_request' );
+		return $wcpay_request->handle_request( 'wcpay_list_disputes_request', $request );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-payments-disputes-controller.php
+++ b/includes/admin/class-wc-rest-payments-disputes-controller.php
@@ -87,12 +87,7 @@ class WC_REST_Payments_Disputes_Controller extends WC_Payments_REST_Controller {
 	 * @param WP_REST_Request $request Full data about the request.
 	 */
 	public function get_disputes( WP_REST_Request $request ) {
-		$wcpay_request = List_Disputes::create();
-		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
-		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
-		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) ?? 'created' );
-		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) ?? 'desc' );
-		$wcpay_request->set_filters( $this->get_disputes_filters( $request ) );
+		$wcpay_request = List_Disputes::from_rest_request( $request );
 
 		return $wcpay_request->handle_rest_request( 'wcpay_list_disputes_request', $request );
 	}
--- a/includes/admin/class-wc-rest-payments-disputes-controller.php
+++ b/includes/admin/class-wc-rest-payments-disputes-controller.php
@@ -5,6 +5,8 @@
  * @package WooCommerce\Payments\Admin
  */
 
+use WCPay\Core\Server\Request\List_Disputes;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -85,13 +87,14 @@ class WC_REST_Payments_Disputes_Controller extends WC_Payments_REST_Controller {
 	 * @param WP_REST_Request $request Full data about the request.
 	 */
 	public function get_disputes( WP_REST_Request $request ) {
-		$page      = (int) $request->get_param( 'page' );
-		$page_size = (int) $request->get_param( 'pagesize' );
-		$sort      = $request->get_param( 'sort' ) ?? 'created';
-		$direction = $request->get_param( 'direction' ) ?? 'desc';
-		$filters   = $this->get_disputes_filters( $request );
+		$wcpay_request = List_Disputes::create();
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) ?? 'created' );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) ?? 'desc' );
+		$wcpay_request->set_filters( $this->get_disputes_filters( $request ) );
 
-		return $this->forward_request( 'list_disputes', [ $page, $page_size, $sort, $direction, $filters ] );
+		return $wcpay_request->handle_request( 'wcpay_list_disputes_request' );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-payments-reader-controller.php
+++ b/includes/admin/class-wc-rest-payments-reader-controller.php
@@ -279,9 +279,8 @@ class WC_REST_Payments_Reader_Controller extends WC_Payments_REST_Controller {
 
 			$charge         = $payment_intent->get_charge();
 			$charge_id      = $charge ? $charge->get_id() : null;
-			$charge_request = Get_Charge::create();
-			$charge_request->set_charge_id( $charge_id );
-			$charge_array = $charge_request->send( 'wcpay_get_charge_request', $charge_id );
+			$charge_request = Get_Charge::create( $charge_id );
+			$charge_array   = $charge_request->send( 'wcpay_get_charge_request', $charge_id );
 
 			/* Collect receipt data, stored on the store side. */
 			$order = wc_get_order( $charge_array['order']['number'] );
--- a/includes/admin/class-wc-rest-payments-reader-controller.php
+++ b/includes/admin/class-wc-rest-payments-reader-controller.php
@@ -280,7 +280,7 @@ class WC_REST_Payments_Reader_Controller extends WC_Payments_REST_Controller {
 			$charge         = $payment_intent->get_charge();
 			$charge_id      = $charge ? $charge->get_id() : null;
 			$charge_request = Get_Charge::create( $charge_id );
-			$charge_array   = $charge_request->send( 'wcpay_get_charge_request', $charge_id );
+			$charge_array   = $charge_request->send( 'wcpay_get_charge_request' );
 
 			/* Collect receipt data, stored on the store side. */
 			$order = wc_get_order( $charge_array['order']['number'] );
--- a/includes/admin/class-wc-rest-payments-reader-controller.php
+++ b/includes/admin/class-wc-rest-payments-reader-controller.php
@@ -5,6 +5,7 @@
  * @package WooCommerce\Payments\Admin
  */
 
+use WCPay\Core\Server\Request\Get_Charge;
 use WCPay\Core\Server\Request\Get_Intent;
 use WCPay\Exceptions\API_Exception;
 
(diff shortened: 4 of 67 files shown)


includes/core/server/class-request.php
```diff
@@ -226,6 +228,26 @@ final public function send( $hook, ...$args ) {
...
+	 *
+	 * @throws Extend_Request_Exception
+	 * @throws Immutable_Parameter_Exception
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	final public function handle_request( $hook, ...$args ) {
+		try {
+			return $this->send( $hook, $args );
+		} catch ( API_Exception $e ) {
+			new WP_Error( $e->get_error_code(), $e->getMessage() );
```
RadoslavGeorgiev: `WP_Error`s are just objects, which are not thrown like exceptions, they need to be returned. And when returned, they still need to be handled. I don't understand how this method would help, let's see if the rest of the PR is going to clear that up.

includes/admin/class-wc-rest-payments-charges-controller.php
```diff
@@ -46,9 +46,8 @@ public function get_charge( $request ) {
 		$charge_id = $request->get_param( 'charge_id' );
 
 		try {
-			$wcpay_request = Get_Charge::create();
-			$wcpay_request->set_charge_id( $charge_id );
-			$charge = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
+			$wcpay_request = Get_Charge::create( $charge_id );
+			$charge        = $wcpay_request->send( 'wcpay_get_charge_request', $charge_id );
```
RadoslavGeorgiev: The charge ID is something that's already a parameter within the request, adding it to the filter here doesn't provide any benefit. Let's remove it.

includes/admin/class-wc-rest-payments-disputes-controller.php
```diff
@@ -85,13 +87,14 @@ public function register_routes() {
...
-		$filters   = $this->get_disputes_filters( $request );
+		$wcpay_request = List_Disputes::create();
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) ?? 'created' );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) ?? 'desc' );
+		$wcpay_request->set_filters( $this->get_disputes_filters( $request ) );
 
-		return $this->forward_request( 'list_disputes', [ $page, $page_size, $sort, $direction, $filters ] );
+		return $wcpay_request->handle_request( 'wcpay_list_disputes_request' );
```
RadoslavGeorgiev: Providing the `WP_Rest_Request $request` here could be really useful to this filter.

includes/admin/class-wc-rest-payments-reader-controller.php
```diff
@@ -279,9 +279,8 @@ public function generate_print_receipt( $request ) {
 
 			$charge         = $payment_intent->get_charge();
 			$charge_id      = $charge ? $charge->get_id() : null;
-			$charge_request = Get_Charge::create();
-			$charge_request->set_charge_id( $charge_id );
-			$charge_array = $charge_request->send( 'wcpay_get_charge_request', $charge_id );
+			$charge_request = Get_Charge::create( $charge_id );
+			$charge_array   = $charge_request->send( 'wcpay_get_charge_request', $charge_id );
```
RadoslavGeorgiev: Same point about the redundant `$charge_id` here.

includes/core/server/request/class-get-charge.php
```diff
@@ -18,28 +19,17 @@
...
-	/**
-	 * Charge id.
-	 *
-	 * @var string|null $intent_id
-	 */
-	private $charge_id = null;
 
 	/**
-	 * Set intent id.
+	 * Sets the intent ID, which will be used in the request URL.
```
RadoslavGeorgiev: `intent` needs to be replaced here? Same with the parameter description.

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,154 @@
...
+	private $uri = null;
+
+	/**
+	 * Returns the request's API.
+	 *
+	 * @return string
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function get_api(): string {
+		return $this->uri; // Will be changed on every request by overriding this method or use set_uri method.
```
RadoslavGeorgiev: I'm not sure I prefer the approach of a universal `Paginated` request here. Any reason against creating specific documents/deposits/authorizations classes? If everything they do is as simple as overwrite the `get_api()` method, it should be fairly easy to implement.

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Return exact same response from API server. In some cases, this is all we need.
+	 *
+	 * @param mixed $response Response from server.
+	 *
+	 * @return mixed|\WCPay\Core\Server\Response
+	 */
+	public function format_response( $response ) {
+		return $response;
```
RadoslavGeorgiev: Is there any reason to avoid the `Response` class here? I know it's this vs `to_array()`, and we'll be overwriting those responses with proper classes at some point in the future, so either option would be fine, but regressing like this in a sub-class feels weird.

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	/**
+	 * Set filters.
+	 *
+	 * @param array $filters Filterd to set.
+	 *
+	 * @return void
+	 */
+	public function set_filters( array $filters ) {
```
RadoslavGeorgiev, line 72: Why not make the setters explicit, and add setters for individual filters within request classes?

includes/admin/class-wc-rest-payments-authorizations-controller.php
```diff
@@ -58,11 +60,14 @@ public function register_routes() {
 	 * @param WP_REST_Request $request Full data about the request.
 	 */
 	public function get_authorizations( WP_REST_Request $request ) {
-		$page      = (int) $request->get_param( 'page' );
-		$page_size = (int) $request->get_param( 'pagesize' );
-		$sort      = $request->get_param( 'sort' );
-		$direction = $request->get_param( 'direction' );
-		return $this->forward_request( 'list_authorizations', [ $page, $page_size, $sort, $direction ] );
+		$wcpay_request = Paginated_Request::create();
```
RadoslavGeorgiev: Considering that `uri`, `page`, `page_size`, `sort_by`, and `sort_direction` are universal, do you think we could create a non-final `create_from_rest_request` method within the `Paginated` class, something like this:

```php
class Paginated {
	public static function from_rest_request( $request ) {
		$wcpay_request = static::create();
		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );	
		return $wcpay_request;
	}
}
```
This way sub-classes will be able to nicely use filters as well:

```php
class List_Disputes extends Paginated {
	public static function from_rest_request( $request ) {
		$filters = [
			'match'           => $request->get_param( 'match' ),
			'currency_is'     => $request->get_param( 'store_currency_is' ),
			'created_before'  => $request->get_param( 'date_before' ),
			'created_after'   => $request->get_param( 'date_after' ),
			'created_between' => $request->get_param( 'date_between' ),
			'search'          => $request->get_param( 'search' ),
			'status_is'       => $request->get_param( 'status_is' ),
			'status_is_not'   => $request->get_param( 'status_is_not' ),
		];

		$wcpay_request = parent::from_rest_request( $request );
		$wcpay_request->set_filters( $filters );
		return $wcpay_request;
	}
}
```

By doing so we will simultaneously add a good base for a bridge between REST and server requests.

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,80 @@
...
+ */
+class List_Transactions extends Paginated {
+
+	use Order_Info;
+
+	const DEFAULTS = [
+		'page'       => 0,
+		'pagesize'   => 25,
+		'sort'       => 'date',
+		'direction'  => 'desc',
```
RadoslavGeorgiev: Defaults are inherited from parents, at least `page` and `pagesize` do not to be re-defined here.

RadoslavGeorgiev (changes requested):

includes/core/server/class-request.php
zmaglica: The purpose of this code is to mimic the behavior of this function: https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-payments-rest-controller.php#L48

Instead of getting an array and parse exception for each request, this function will do that for us inside the request class and only in one place.

RadoslavGeorgiev: I should have revisited this comment :D

I see the point, let's rename it to something like `handle_rest_request` to make the purpose clear though, what do you think?

includes/core/server/request/class-get-charge.php
zmaglica: I will address this in intent branch.

includes/core/server/request/class-paginated.php
zmaglica: I wanted to avoid creating class, import it, create test for it and use it only in one place. It felt to me like we are creating a bunch of classes where only one line of code will be changed. But, I see your point here. Adding extra class is fine approach too, but considering that we have around 100+ API calls, we should be careful in creating a bunch of classes since it will be difficult to navigate through the code.

includes/core/server/class-request.php
zmaglica: Make sense. I will rename it to `handle_rest_request`

includes/core/server/request/class-paginated.php
zmaglica: I wanted to avoid calling ->to_array() for every class. I will refactor this and place it inside `handle_request` method

includes/core/server/request/class-paginated.php
RadoslavGeorgiev: > since it will be difficult to navigate through the code

IDEs for the win :)

If it's just the API method that's changing, we could have a single test for each of those sub-requests within a single test class, and just add one line there to prevent coverage decreases.

includes/core/server/request/class-paginated.php
```diff
@@ -38,7 +33,7 @@ class Paginated extends Request {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function get_api(): string {
-		return $this->uri; // Will be changed on every request by overriding this method or use set_uri method.
+		return ''; // Will be changed on every request that extends this class.
```
RadoslavGeorgiev: Does it make sense to keep this method? Leaving it abstract as the base request will enforce other paginated requests to implement it.

includes/core/server/request/class-list-transactions.php
zmaglica: This is fixed.

includes/admin/class-wc-rest-payments-authorizations-controller.php
zmaglica: All paginated request classes have this feature now.

includes/core/server/request/class-paginated.php
zmaglica: These filters have been added.

includes/core/server/request/class-paginated.php
zmaglica: Moved into `hanndle_request` function

includes/core/server/request/class-paginated.php
zmaglica: This is refactored and all classes or requests have their own class with dedicated URI and other things like filters and setters.

includes/admin/class-wc-rest-payments-reader-controller.php
zmaglica: Removed.

includes/admin/class-wc-rest-payments-disputes-controller.php
zmaglica: Added.

includes/admin/class-wc-rest-payments-charges-controller.php
zmaglica: Removed.

includes/core/server/class-request.php
zmaglica: The function name is renamed to `handle_rest_request`

includes/core/server/class-request.php
```diff
@@ -226,6 +229,33 @@ final public function send( $hook, ...$args ) {
...
+	 * @param mixed  ...$args      Other parameters for the hook.
+	 * @return mixed               Either the response array, or the correct object.
+	 *
+	 * @throws Extend_Request_Exception
+	 * @throws Immutable_Parameter_Exception
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	final public function handle_rest_request( $hook, ...$args ) {
+		try {
+			$data = $this->send( $hook, $args );
```
RadoslavGeorgiev: Don't we need to expand `$args` here as well?
```suggestion
			$data = $this->send( $hook, ...$args );
```

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @param WP_REST_Request $request Request object.
+	 *
+	 * @return Paginated
+	 */
+	public static function from_rest_request( $request ) {
+		$wcpay_request = static::create();
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
```
RadoslavGeorgiev: We need to check whether `page`, `pagesize`, `sort`, and `direction` are present within the REST request before setting them here. Removing one of them from the REST URL results in this right now:

```
Fatal error: Uncaught TypeError: Argument 1 passed to WCPay\Core\Server\Request\Paginated: :set_sort_by() must be of the type string,
null given, called in /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/request/class-paginated.php on line 57 and defined in /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/request/class-paginated.php on line 126
```

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,140 @@
...
+use WP_REST_Request;
+
+/**
+ * Request class for paginated requests.
+ */
+class Paginated extends Request {
+
+
+	const DEFAULTS = [
+		'page'      => 0,
```
RadoslavGeorgiev, line 20: Shouldn't this default to 1, in case the parameter is omitted?

includes/admin/class-wc-rest-payments-transactions-controller.php
```diff
@@ -76,13 +78,16 @@ public function register_routes() {
...
-		$filters    = $this->get_transactions_filters( $request );
-		return $this->forward_request( 'list_transactions', [ $page, $page_size, $sort, $direction, $filters, $deposit_id ] );
+
+		$wcpay_request = List_Transactions::from_rest_request( $request );
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
+		$wcpay_request->set_deposit_id( $request->get_param( 'deposit_id' ) );
+		$wcpay_request->set_filters( $this->get_transactions_filters( $request ) );
```
RadoslavGeorgiev: All of this is handled within the `List_Transactions` request already, let's remove it from this file.

includes/core/server/class-request.php
RadoslavGeorgiev: Alright. In any case, `WP_Error` is an object, which is not getting thrown in this case, and needs to be returned in order for the REST controller to handle it.

includes/admin/class-wc-rest-payments-transactions-controller.php
```diff
@@ -76,13 +78,16 @@ public function register_routes() {
...
-		$deposit_id = $request->get_param( 'deposit_id' );
-		$filters    = $this->get_transactions_filters( $request );
-		return $this->forward_request( 'list_transactions', [ $page, $page_size, $sort, $direction, $filters, $deposit_id ] );
+
+		$wcpay_request = List_Transactions::from_rest_request( $request );
+		$wcpay_request->set_page( (int) $request->get_param( 'page' ) );
+		$wcpay_request->set_page_size( (int) $request->get_param( 'pagesize' ) );
+		$wcpay_request->set_sort_by( $request->get_param( 'sort' ) );
+		$wcpay_request->set_sort_direction( $request->get_param( 'direction' ) );
+		$wcpay_request->set_deposit_id( $request->get_param( 'deposit_id' ) );
```
RadoslavGeorgiev: The deposit ID is not a part of `List_Transactions::from_rest_request` though, let's move it there before deleting the other lines here.

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+use WP_REST_Request;
+
+/**
+ * Request class for listing transactions.
+ */
+class List_Transactions extends Paginated {
+
+	use Order_Info;
+
+	const DEFAULTS = [
```
RadoslavGeorgiev: There are many cases where this should be `DEFAULT_PARAMS`, but is just `DEFAULTS` instead. Let's pick one, and stick with it. I'd prefer `DEFAULT_PARAMS` as it's already this way within the base requests PR, but I don't have a strong opinion against `DEFAULTS`.

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Used to provide custom check for setting filter keys. Some requests might have their own logic of which filters keys are mutable.
+	 *
+	 * @param string $key Key to check.
+	 *
+	 * @return bool
+	 */
+	public function is_filter_key_mutable( string $key ) {
+		// In most cases, the default parameters are keys that cannot be changed via filters.
+		return ! array_key_exists( $key, self::DEFAULTS );
```
RadoslavGeorgiev: Is there a strong reason why defaults are something that should be immutable here? If so, I'd just declare the same keys as in the `DEFAULT_PARAMS` array info `IMMUTABLE_PARAMS`, instead of introducing a new method.

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+		'sort'       => 'date',
+		'direction'  => 'desc',
+		'deposit_id' => null,
```
RadoslavGeorgiev: By introducing an empty deposit ID here, we're sending a blank parameter to the server with every request. If it's not required, let's remove the default value here.

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+
+	/**
+	 * Set date after.
+	 *
+	 * @param string $date_after Date after.
+	 *
+	 * @return void
+	 * @throws \WCPay\Core\Exceptions\Server\Request\Invalid_Request_Parameter_Exception
+	 */
+	public function set_date_after( string $date_after ) {
```
RadoslavGeorgiev: `set_date_after` and `set_date_between` appear a few times throughout the code. Could we maybe consolidate them into a trait?

image

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+	/**
+	 * Set date between.
+	 *
+	 * @param array $date_between Date between.
+	 *
+	 * @return void
+	 */
+	public function set_date_between( array $date_between ) {
+		// Add only if it's not empty.
+		if ( $date_between ) {
```
RadoslavGeorgiev: Since we're validating dates in the other methods, let's do it here as well. Considering the parameter is an array with numeric indexes (0 and 1), how about something like this?

```suggestion
		if ( is_array( $date_between ) && ! empty( $date_between ) ) {
			foreach ( $date_between as $date ) {
				$this->validate_date( $date );
			}
			$this->set_param( 'date_between', $date_between );
		}
```

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+
+	/**
+	 * Set search.
+	 *
+	 * @param string $search Search term.
+	 *
+	 * @return void
+	 */
+	public function set_search( string $search ) {
+		$this->set_param( 'search', $search );
```
RadoslavGeorgiev: This one is incorrect. If you take a look on the server, you'll see that `search` should be an array: https://github.com/Automattic/woocommerce-payments-server/blob/ce757b31f600ba0d2ee582978682856bf24031ce/server/wp-content/rest-api-plugins/endpoints/wcpay/class-transactions-controller.php#L88-L92

Also, there is a missing `WC_Payments_Utils::map_search_orders_to_charge_ids` call, which is otherwise present in the API client method: https://github.com/Automattic/woocommerce-payments/blob/7692800948ef80d3118c5593c32ee6776ea51022/includes/wc-payment-api/class-wc-payments-api-client.php#L666-L668

includes/core/server/request/class-list-transactions.php
```diff
@@ -0,0 +1,279 @@
...
+	 */
+	public static function from_rest_request( $request ) {
+		$wcpay_request       = parent::from_rest_request( $request );
+		$date_between_filter = $request->get_param( 'date_between' );
+		$user_timezone       = $request->get_param( 'user_timezone' );
+
+		if ( ! is_null( $date_between_filter ) ) {
+			$date_between_filter = array_map(
+				function ( $transaction_date ) use ( $user_timezone, $wcpay_request ) {
+					return $wcpay_request->format_transaction_date_with_timestamp( $transaction_date, $user_timezone );
```
RadoslavGeorgiev: Even if we set the return type of `parent::from_rest_request()` to be `static`, it's still being returned as `Paginated`, which does not have `format_transaction_date_with_timestamp`, causing linting errors.

What do you think about making the method static as well? It does not need any object-specific functionality.

RadoslavGeorgiev (changes requested):
I did a quick pass, focusing my testing on the list transactions endpoint. I'll review tests later, but so far it's looking pretty good :)

includes/core/server/request/class-paginated.php
zmaglica: Abstract function override is only allowed since PHP version 7.2. Currently, the minimum supported version of WooCommerce payment plugin is 7.0, so I didn't want to introduce a possible break change. We should bump minimum PHP supported version to 7.2 or leave it like this.

includes/core/server/request/class-paginated.php
zmaglica: If you take a look at `list_` endpoints in `WC_Payments_API_Client` class, you will find that the default value for page is zero. It felt wired to me too, but I kept it to zero because I didn't want to change previous behavior.

includes/core/server/request/class-paginated.php
RadoslavGeorgiev: Can't we just remove it from the `Paginated` class though? I don't think it needs to be overwritten.

includes/core/server/request/class-paginated.php
RadoslavGeorgiev: It happens to also work with 0, so let's leave it that way then :)

includes/core/server/request/class-paginated.php
zmaglica: image
PHPStorm cries about missing method in class, so I added it 😄 and I think that all child classes need to fill "body" of all abstract methods that are defined in parent abstract class (maybe I am wrong on this one, so I am not 100% certain)

RadoslavGeorgiev: The class was used directly, but now that it's not, we should just mark it as abstract as well. Then there should be no more complaints.

includes/core/server/request/class-list-transactions.php
zmaglica: A new trait has been added.

includes/core/server/request/class-list-transactions.php
zmaglica: Default Deposit have been removed.

includes/admin/class-wc-rest-payments-transactions-controller.php
zmaglica: Removed.

includes/core/server/request/class-paginated.php
zmaglica: This is fixed. Direction and sort is checked before calling set functions and page and page size have been cased to int with default values.

includes/core/server/class-request.php
zmaglica: Fixed.

includes/core/server/request/class-paginated.php
zmaglica: Good point. I really like approach. I make it abstract now.

includes/core/server/request/class-paginated.php
zmaglica: Make sense. I used `IMMUTABLE_PARAMS` constant to achieve the same results.

includes/core/server/request/class-list-transactions.php
zmaglica: This is refactored and moved into date params trait.

includes/core/server/request/class-list-transactions.php
zmaglica: Good point. Made this function static.

includes/core/server/request/class-list-transactions.php
zmaglica: This is fixed.

includes/core/server/request/class-paginated.php
```diff
@@ -0,0 +1,131 @@
...
+	 *
+	 * @param array $filters Filterd to set.
+	 *
+	 * @return void
+	 */
+	public function set_filters( array $filters ) {
+
+		foreach ( $filters as $key => $value ) {
+			// Nullable filters are not needed and mutable keys can only be changed with this method.
+			if ( null !== $value && ! in_array( $key, self::IMMUTABLE_PARAMS, true ) ) {
```
RadoslavGeorgiev: Thanks for switching from default parameters to an immutable declaration. With that in mind though, I don't see any reason to keep the `is_array( ... )` check here. If it's needed for whatever reason, I don't see overwriting paging parameters as mission-critical, so let consumers overwrite them :)

RadoslavGeorgiev (changes requested):
So close! One small nitpick, and I believe we'll be done with this one :)

RadoslavGeorgiev (commented):
All changes in the PR look good, let's make sure all tests are passing once updated with the contents of the `dev/create-and-setup-intent-class` :)

RadoslavGeorgiev (approved):

PR #5282, opened 2022-12-13
Integrate Stripe Link with Split-UPE/WooPay

### LINKED ISSUE #4720: Multiple UPE - Stripe Link compatibility

FangedParakeet:
Fixes #4720 

# Stripe Link's Awakening

This PR integrates Stripe Link with both the split-UPE implementation and WooPay. Stripe Link is an express checkout payment method that requires a UPE card payment element at checkout and responds to changes from the email input field. The split-UPE implementation does not use the UPE for the card element in order to maintain compatibility with WooPay and moreover WooPay also listens to changes in the email input field, which causes WooPay and Link's authentication modals to overlap each other disgracefully on the checkout.

To resolve these conflicts a separate UPE payment gateway is added to the checkout for Link and Link's automatic functionality is replaced by a payment request button, which when triggered interrupts activity from the WooPay checkout. The designs for this component, as well as WooPay's own checkout button have remained relatively fluid, so the design of this feature currently follows the design of existing payment request buttons. 

## A Stripe Link Refresher

To assist in understanding exactly what is going in this PR, here is a quick refresher on Stripe Link. Stripe Link is an express checkout payment method that allows a user to store their credit card information with Stripe, using a user's email address and phone number as authenticating keys. To register a new Stripe Link account, if an unrecognised email address is present in the email form field, a new checkbox to save one's payment information will appear underneath the payment element.

Stripe Link save details checkbox
_Link's special checkbox to save user details_

Upon returning to the checkout, entering the same registered email address will now open an OTP modal, which upon successful completion will authenticate the user and present a new modal displaying the user's previously saved addresses and payment methods, offering to autofill these values into the checkout. From this point onwards, the transaction completes like a regular UPE checkout.

Stripe Link OTP modal
_First OTP modal_

Stripe autofill modal
_Second autofill modal_

## Objectives

Ideally we would like to implement the following functionality.
- Stripe Link appears as an express checkout payment button.
- Stripe Link modal and WooPay modals do not appear at the same time.

## Obstacles

Stripe's Link modals are managed by a `linkAutofillModal` object ([reference here](https://stripe.com/docs/payments/link/autofill-modal#reference)), which provides us with a limited interface with the following functions.

- `launch({email: string}): void`
   - This is the equivalent of our `/user/exists` endpoint. However, instead of returning the result, if an email address is recognised Stripe will immediately open its OTP modal. This means we are never alerted when this request succeeds and have no way of knowing when or if the OTP modal will open after this request has been made. Additionally, if the OTP modal is closed by the user, it is not possible to restart the checkout flow--even with this function--and there is no way for us to know when this has occurred either. 
- `on("authenticated", (AuthenticationEvent) => void)`
   - This event will fire once the OTP has been entered correctly and the user has successfully authenticated.
- `on("autofill", (AutofillEvent) => void)`
   - Once a user has successfully authenticated and selected their desired shipping/billing address and payment method, this event will fire, containing the desired values for us to enter into our own checkout fields.
- `show(): void`
   - Once a user has successfully authenticated their account via OTP. A new modal will appear for a registered user to select their desired shipping/billing address and payment method. If _this_ modal is closed, it can be reopened using this `show()` function.
- `logout(): void`
   - This function can logout an authenticated user, exiting the Stripe Link checkout. (I don't currently see a use-case for this function.)

This creates the following challenges to our objectives stated above.

- We have no way of knowing whether a user's email address is a registered Stripe Link account, until after they have successfully validated via OTP.
- Once we "begin" the Stripe Link checkout, we have no way of controlling or restraining the OTP modal from appearing on the page. If we query for a user's registration and the user is found, the modal _will_ open on the page. 
- If we query Stripe Link for a user's registration and the OTP modal does not open, it could be because the query is still ongoing or it could be because the user has not been found and we have no way of distinguishing between either state. 

Ideally it would be nice if we could connect our WooPay and Link integrations by either checking both registration statuses and then performing some sort of action depending on the result. I.e. checking whether a user has a WooPay account, checking whether they have a Link account, and then presenting valid express checkout options depending on which accounts are available. This is not possible due to Link's above limitation, as we cannot know that a user has registered with Link without launching the modal and proceeding through the Link checkout process.

Alternatively, we could check whether a user is registered on one platform and then check the other platform afterwards. Unfortunately, this is also not possible within this current phase of the UPE, because WooPay's scripts and Stripe Link's scripts are sadly separated at checkout.

WooPay relies on the non-UPE `card` element, so its code lives over [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/index.js#L590) with the classic checkout script; Stripe Link relies on a UPE `payment` element--and must be initialised with the Stripe Elements object associated with this `payment` element--so these scripts must be placed [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe.js#L271) with the UPE checkout script. This means that we cannot trigger an action for one express checkout method from the other method. 

For example, [it was suggested that we could present an option to checkout with Link inside the WooPay modal](https://github.com/automattic/woocommerce-payments/issues/4720#issuecomment-1268747990). This is currently not feasible, since there is no trivial way for us to trigger an option to commence the Link checkout from the WooPay modal, since the scripts are enqueued separately on the page and are unable to interact with each other. 

(I say no _trivial_ way to do this, because I created a workaround by attaching a class to the email input to indicate that the Link checkout is ongoing and used this to block the WooPay modal and something similar could be done to communicate between the elements the other way around; but this would involve both scripts polling the page to check for these breadcrumbs and would have obvious negative effects on performance besides being very bodgy 😅 .)

**However, note that this approach _will_ be possible in the second phase of our split-UPE project, when WooPay can be triggered from within the UPE, as WooPay and Link will both be served from the UPE client script.**

## Outcomes

As a result of these conditions, this is the compromise that I have concluded upon in this PR. 

Stripe Link is present as a payment request button. Currently, it will not automatically listen to changes on the email input, as WooPay does. However, if a user selects to checkout with Stripe Link, we will begin sending the email input values to Link instead and regardless of the result (since this information is not available to us), the WooPay modal will be blocked by Link. In this scenario, selecting to checkout with Stripe Link means either the user has clicked the Stripe Link express checkout button or has selected the Link checkout gateway radio button.

So if a user has a single email address associated with both WooPay and Stripe Link, the following two paths are possible at checkout.

If the user arrives at the checkout and does not actively select Link, the WooPay modal will appear and Stripe will be none the wiser of the user's presence.

GIF of WooPay modal opening
_WooPay takes precedent when Link is not selected_

However, if the user selects Link while we are still querying WooPay, we will instead switch to the Link checkout, the WooPay modal will be blocked, and Link's modal will open instead.

GIF of Link modal opening
_Selecting Stripe Link blocks WooPay_

Note in the GIF above, that when the Link payment request button was clicked, in the bottom-right corner, the payment gateway also changed (with Link being selected off-screen). Also, note that in the previous scenario, if a user's email was registered to WooPay, but _not_ Link and a user clicked the Link express checkout button, then neither modal would open, since we have no way of knowing that a user's email is _not_ registered to Stripe Link. However, in this case, it's possible that a user could restart the WooPay checkout by clicking on the WooPay express checkout button after selecting Link and having nothing happen, though this action is not currently implemented on the WooPay button. 

CC @rogermattic since I have had to deviate delicately from your desired designs, so I would greatly appreciate your feedback on the destination I have arrived at. Hopefully, the GIFs and images in this lengthy description give you some idea of where that is. 

## Limitations of this PR

The Blocks checkout does not currently work with Stripe Link. This is currently very broken. Please do not try and go there or you will be very sad. This can be fixed in a separate issue, depending on the decisions made here.

I am also relatively certain that this will not work from the Pay for Order page either, though I haven't explicitly tested this yet.

Also, all the WCPay tests appear to be passing! 🎉 ...Despite me making wholesale changes to the Link checkout. 😞 This suggests to me that, if there is any test coverage for Stripe Link, these tests are quite sorely lacking. This will also have to be fixed in a separate issue, because I simply didn't get around to it here.

Finally, note one open bug that when both WooPay and Link are enabled as express checkout methods there are two `- OR -` separators visible above the checkout fields.

Duplicate separators
_Doppelganging separators_

This is an open issue beyond this PR that will occur if any two WCPay express checkout methods are enabled at checkout at the same time--regardless of whether one of them is Link or WooPay. This is because the express checkout methods are all [registered separately](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L369-L371) and each one adds a hook to place the separator on the page ([1](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-platform-checkout-button-handler.php#L85), [2](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L86)), so consequently the separator is duplicated when more than one express checkout button is available at checkout.

Ideally, the express checkout methods should all be registered together, so that we can group them together...but I tried to achieve this quickly myself and it caused much pain and broke many things, so I'll be leaving this for another day as well.

#### Testing instructions

**Sign-up flow**
1. Enable Stripe Link in your WCPay settings (you may need to request the capability as well) and set the currency at checkout to USD. 
2. Add item to cart, proceed to checkout.
3. Click the Link express checkout button. If the email input is empty, an error should appear prompting you to fill in this field. 
4. At this point, even with the field filled, nothing should happen, since you have not yet registered your email address yet. However, note that the selected payment gateway changes to Link. The "Save your info" checkbox should also be present underneath the payment information.
5. Fill in billing/shipping details and payment information. Select "Save your info" checkbox, enter phone number.
6. Complete checkout. Check order details and successful status. Check payment intent details in Stripe dashboard.
7. Add new item to cart. Proceed to checkout.
8. Enter previously registered email address. Stripe Link OTP modal should appear.

**Registration flow**
1. Enable Stripe Link in your WCPay settings (you may need to request the capability as well) and set the currency at checkout to USD. 
2. Add item to cart, proceed to checkout.
3. Click the Link express checkout button or Link payment gateway radio button. If the email input is empty, an error should appear prompting you to fill in this field. If email input is empty or set to an unregistered email address, "Save your info" checkbox should be present on Link payment gateway.
4. Enter Link registered email address into "Email address" field.
5. Link OTP modal should appear. Enter `000000` to successfully authenticate (note: Stripe will not actually send an SMS to your registered in test mode).
6. Link Autofill modal should now appear. Select existing payment method/details or add new details.
7. Close Autofill modal. You should be able to reopen the modal by clicking the Link express checkout button at the top of the page.
8. Close Autofill modal again. If you click another radio button, to change the selected payment gateway, upon returning to Link, the Autofill modal should reopen again with the correct information.
9. Select "Autofill checkout". Link should populate all relevant fields on checkout.
10. Complete checkout. Check order details and successful status. Check payment intent details in Stripe dashboard.

**WooPay flow**
1. Enable Stripe Link in your WCPay settings (you may need to request the capability as well) and set the currency at checkout to USD. 
2. Add item to cart, proceed to checkout.
3. Ensure Link is not the currently selected payment gateway. This time do not select Link express checkout button. 
4. Enter WooPay registered email address into "Email address" field.
5. WooPay spinner should display. WooPay modal should appear. (If you close the WooPay modal at this point, you should be able to summon the Link modal, by selecting the Link express checkout button at this stage).
6. Complete WooPay checkout. Check order details and successful status. Check payment intent details in Stripe dashboard.

- Confirm classic card checkout flow and/or UPE checkout flow for additional payment methods is unaffected. 
- Confirm Link is only present at checkout when currency is USD. 
- Try to break things. Hopefully, they don't break.

-------------------

**Post merge**


client/checkout/classic/upe.js
```diff
@@ -371,13 +392,18 @@ jQuery( function ( $ ) {
 		upeElement.mount( upeDOMElement );
 		unblockUI( $upeContainer );
 		upeElement.on( 'change', ( event ) => {
-			const selectedUPEPaymentType = event.value.type;
+			let selectedUPEPaymentType = event.value.type;
+			if ( 'card' === selectedUPEPaymentType ) {
```
FangedParakeet, line 398: Irritatingly, despite the payment element containing both `link` and `card` in its `payment_method_types` parameter, this result will always return as `card`. Go figure.

client/checkout/platform-checkout/email-input-iframe.js
```diff
@@ -254,6 +255,9 @@ export const handlePlatformCheckoutEmailInput = async (
 	iframeWrapper.addEventListener( 'click', closeIframe );
 
 	const openIframe = ( email ) => {
+		if ( isLinkCheckoutActive( platformCheckoutEmailInput ) ) {
```
FangedParakeet: This is my cheeky hack to interrupt the WooPay checkout modal, if Stripe Link has been selected. You can check the util function, but this is basically achieved by adding a new class to the email input and checking for its presence. 

FWIW the WooPay preloader still spins and this might be a little confusing, because ultimately (unless you change the payment gateway away from Link) nothing will happen. Let me know if you think this is bad!

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -657,10 +666,6 @@ public function process_redirect_payment( $order_id, $intent_id, $save_payment_m
 					'upe_payment_intent_error'
 				);
 			} else {
-				if ( $this->payment_method->get_id() !== $payment_method_type ) {
-					return;
-				}
-
```
FangedParakeet: So I think since we moved [the `maybe_process_upe_redirect` hook](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments-upe-checkout.php#L77) to the `WC_Payments_UPE_Checkout`, which gets [instantantiated once with its gateway set to an instance of the `card` gateway](https://github.com/Automattic/woocommerce-payments/blob/poc/upe-instances-multiplied/includes/class-wc-payments.php#L388), this line of code broke the checkout for every non-card payment method. 😅 

`$this->payment_method->get_id()` will always be `card`, which means this will exit for every non-card `$payment_method_type` payment method...which is every single UPE payment method (lol). This basically completes the transaction, but leaves the WC order status incomplete. I think this line was just defensive coding to begin with, so I think it can safely be removed, but let me know if you think I might be mistaken.

client/checkout/classic/style.scss
```diff
@@ -44,17 +44,32 @@
...
+	width: 100%;
+	background-color: #fff;
+	cursor: pointer;
+	margin: 0 auto;
+	border: 2px solid rgb( 247, 247, 247 );
+}
+#wcpay-stripe-link-button-wrapper button:active {
+	background-color: rgb( 247, 247, 247 );
+}
+
```
timur27, line 59: All these styles look great and seem to be in line with other styles we have, as per my personal experience. The only tiny thing that caught my eye was the button-click effect when the button is disabled. I would rather expect the button to be non-interactive when disabled to reduce confusion and avoid building up our users' hope that the button is functioning when in fact, it isn't.

<details>
  <summary>Here's the visual presentation of what I have in mind in order to be super clear</summary>


https://user-images.githubusercontent.com/22319444/208657908-b4caaa70-a80d-4437-a68d-9e84def66c4d.mov




</details>

If you agree that it will be more clear to disable the click effect, and if we can make such UX decisions, then the problem can be easily resolved by adding the following code: 

```css
#wcpay-stripe-link-button-wrapper button:disabled {
	background: none;
	outline: 0;
	box-shadow: none;
}
```

client/checkout/stripe-link/index.js
```diff
@@ -1,127 +1,298 @@
...
-			linkAutofill.launch( { email: event.target.value } );
-		} );
+	/**
+	 * Listener function for email input keyup event.
+	 *
+	 * @param {Object} event Keyup event.
+	 */
+	keyupHandler( event ) {
+		this.linkAutofill.launch( { email: event.target.value } );
+		this.disableRequestButton();
```
timur27: I would move these two lines into a separate function because: 
1. `linkAutofill.launch` is probably one of the most important functions in this class that deserves a separate method with a name that will underline that we are about to start the process instead of hiding it in keyup event handler. I'd say this will increase the readability of code for our successors. Someone might say it's nitpicking and not worth investing time in, and I understand this point of view. 😄
2. The same action is performed [here](https://github.com/Automattic/woocommerce-payments/pull/5282/files#diff-94d342a0f23678b924afd5eeed0d5b56f830d13c2a08173d018b9433906b6b31R283-R284) and it might be easy to miss synchronising two places if we will decide to remove `disableRequestButton()` from the process, for example. In the same way, someone might forget to add `disableRequestButton()` when implementing Link startup flow in a place other than email input.

Feel free to object if you feel differently! 🤗

client/checkout/stripe-link/index.js
```diff
@@ -1,127 +1,298 @@
...
+	enableRequestButton() {
+		jQuery( STRIPE_LINK_BUTTON_SELECTOR ).removeAttr( 'disabled' );
+	}
+
+	/**
+	 * Disable payment request button.
+	 */
+	disableRequestButton() {
+		jQuery( STRIPE_LINK_BUTTON_SELECTOR ).prop( 'disabled', true );
+	}
```
timur27: Nitpick: can we align these two statements and use 

```js
    jQuery( STRIPE_LINK_BUTTON_SELECTOR ).attr('disabled','disabled');
    jQuery( STRIPE_LINK_BUTTON_SELECTOR ).removeAttr( 'disabled' );
```
OR
```js
    $("input").prop('disabled', true);
    $("input").prop('disabled', false);
```

There are two ways to achieve this ([the second option seems to be more popular](https://stackoverflow.com/questions/13626517/how-to-remove-disabled-attribute-using-jquery)), depending on the jQuery version. Since you used both of them, I made the conclusion that both approaches work. So I thought maybe it could be a good idea to align them. But, maybe there's any particular reason why we use `removeAttr` since it removes the attribute completely instead of setting its value to `false`.

This is a really small thing and feel free to ignore it, if you feel there are more important things to address 👍

client/checkout/platform-checkout/email-input-iframe.js
timur27: The approach you applied is fine for me, as I personally don't know other ways to achieve this at the moment, and the fact that you figured this out and it works is more than enough for us! 🚀 

> the WooPay preloader still spins, and this might be a little confusing

Yep, I agree that this might be a little confusing. However, when I use UPE payment methods such as SEPA or iDEAL, after selecting them, the spinner for WooPay still loads as well. I'd say even if we decide to improve this, a separate issue would be the right choice here to improve everything at once.

includes/payment-methods/class-upe-payment-gateway.php
timur27: Yes, that's a good catch! Thanks for fixing it!

timur27 (commented):
Congrats on making it work despite of the limitations you faced, absolutely great job here! 🚀  I really enjoyed observing the Link payment method working. 

In terms of the code, there are a few small things I left in separate comments - looking forward to hearing from you on them 🙌🏼 

**Sign-up flow**
* Click the Link express checkout button. If the email input is empty, an error should appear prompting you to fill in this field. ✅ 
* Nothing should happen, since you have not yet registered your email address yet. However, note that the selected payment gateway changes to Link. The "Save your info" checkbox should also be present underneath the payment information. ✅ 
* Check order details and successful status. ✅ 
* Check payment intent details in Stripe dashboard ✅ 
* Enter previously registered email address. Stripe Link OTP modal should appear.
✅ 


**Registration flow**
* Click the Link express checkout button or Link payment gateway radio button. If the email input is empty, an error should appear prompting you to fill in this field. ✅ 
* If email input is empty or set to an unregistered email address, "Save your info" checkbox should be present on Link payment gateway. ✅ 
* Link OTP modal should appear ❓ - basically this works, if I enter my email to the email input field **after** the error prompting me to fill in this field was shown. However, if I open the shortcode checkout and fill in the empty email address field with an existing value, the Link OTP modal isn't showing up. I am pretty sure that this is the effect of all the limitations we faced and Link starts initialising only after we click the button or choose Link as a payment method, but wanted to double-check with you to sleep better 😸 
* Enter 000000 to successfully authenticate. Link Autofill modal should now appear. Select existing payment method/details or add new details ✅ 
* Close Autofill modal. You should be able to reopen the modal by clicking the Link express checkout button at the top of the page. ✅ 
* Close Autofill modal again. If you click another radio button, to change the selected payment gateway, upon returning to Link, the Autofill modal should reopen again with the correct information. ✅ 
* Select "Autofill checkout". Link should populate all relevant fields on checkout. ✅ 
* Complete checkout. Check order details and successful status. ✅ 
* Check payment intent details in Stripe dashboard. ✅ 

**WooPay flow** ✅ 
**My account -> Payment methods** Link card CRUD works as expected. ✅ 


-----
There are a few things which I noticed and which are worth discussing. I am looking forward to hearing from you, if you can reproduce them as well and if you feel it's something we need to address or not. 


1️⃣ **Previous successful Link checkout results in loading Link modal faster than WooPay modal for future checkouts**
This is actually the extension to the second compromise you explained in the PR description. 

I paid with Link once and the Order was successful. Now when I opened the shortcode checkout again, Link checkout loaded without any further actions from my side and did it faster than WooPay, thus blocking it. As I mentioned, this is something you already highlighted in your description, but I just wanted to let you know, that the fact of using the same payment method after successfully using them in the past implies, that we will face this race condition without even choosing Link from the list or clicking Link express checkout button. I am not sure how critical it can be for our users.

I guess there's not much we can do at the moment, but please let me know, if you will have any thoughts on this!

2️⃣ _Save your info for secure 1-click checkout with Link_ checkbox behavior
_Save your info for secure 1-click checkout with Link_ checkbox is not showing up when I open the checkout page with the registered email - this is ideally expected and works well.

After I change the email value to the one unknown for Link, the checkbox appears - again, as expected (by the way, I really like the styles here and the transformation effects). 

Now, when I want to get back to the registered email and update the email input field with it, the checkbox stays and does not disappear, which I was expecting it to do. 


https://user-images.githubusercontent.com/22319444/208696008-dc5f2092-e311-4668-889d-a4cd3682a70e.mov


Please let me know if you observe the same and if you think the point is valid and should be addressed. Looking at the number of limitations you successfully overcame, I am not sure if this isn’t another one that is hard to fix. Seems like this behavior could be handled by `linkAutofill` and is isolated.


3️⃣ **Link Autofill modal appears twice on the fresh setup**
After I clear the cache and start from scratch, I observe the autofill modal appearing twice in a row. After that, the issue did not occur - at least, I could not reproduce it after. Below is the screen recording, which is worth more than a million words.

https://user-images.githubusercontent.com/22319444/208715190-caee46c6-6e83-4811-b6ab-024e64ac334f.mov



❔ I have two questions basically to learn a bit from the current PR. 

1️⃣ Is _Save payment information to my account for future purchases_ checkbox expected per design here? It just was a bit unnatural for me to have both Link's saving checkbox and this one - but... this was only at first glance. Both checkboxes work as expected and _Save payment information to my account for future purchases_ successfully saves the card for the future use. 

image

2️⃣ Do you happen to know, why Link requires from us to set USD as the currency used - [the presentment currency in docs](https://stripe.com/docs/payments/link) is All currencies supported on cards. The background of this question is I wanted to play around with Link while having other split UPE payment methods provided in the state of the app. Not sure if that's possible - probably there are some obvious reasons which I'm simply unaware of - this is why this question appeared. 🙌🏼

------------------------------------

> The Blocks checkout does not currently work with Stripe Link. This is currently very broken. Please do not try and go there or you will be very sad. 

This is very sad to hear 😢 Please let me know, if I can help to create an issue, define its scope or anything else - I am happy to be a second head here.


> Also, all the WCPay tests appear to be passing! 🎉 ...Despite me making wholesale changes to the Link checkout. 😞 This suggests to me that, if there is any test coverage for Stripe Link, these tests are quite sorely lacking. This will also have to be fixed in a separate issue, because I simply didn't get around to it here.

Sounds like a good plan. I would also include some tests on the front-end side to ensure the behavior won't break in the future. 👍

client/checkout/classic/style.scss
FangedParakeet: Nice idea, looks much tidier! Implemented in 7b2dc8a4a29d1fa338a33f78007cfc908993bff2.

client/checkout/stripe-link/index.js
FangedParakeet: Makes sense to me, slick suggestion! Implemented in 311f061e7d2ca46e1700f06e6fb3cbc829024c10.

client/checkout/stripe-link/index.js
FangedParakeet: I promise you, I had tried using `.prop( 'disabled', false )` before, but for some reason it wouldn't enable the button. However, I've tried it again now and it is definitely working as it should, so I will assume that the prior incident was due to some other buffoonery on my part. Cheers for the very worthwhile nit-pick: fixed in c238221b3c94bd41f7d804cc9a6cc26fdbce877f.

FangedParakeet:
Thanks so much for the thorough review here, @timur27, and for countering the essay I left as a description with an essay of your own. 😆 Let me respond to your concerns each in turn.

> Link OTP modal should appear ❓ - basically this works, if I enter my email to the email input field after the error prompting me to fill in this field was shown. However, if I open the shortcode checkout and fill in the empty email address field with an existing value, the Link OTP modal isn't showing up. I am pretty sure that this is the effect of all the limitations we faced and Link starts initialising only after we click the button or choose Link as a payment method, but wanted to double-check with you to sleep better 😸

Due to the limitations I described that basically turn the Stripe Link startup flow into a runaway train once triggered, I had to make two compromises. Firstly, Stripe Link remains unconscious of the email field only until either the user clicks the express checkout button or the payment gateway radio button is selected. Consequently, as you've described, the Link OTP modal will not open until it is specifically summoned by the user.
 
This is essentially because the Stripe Link startup modal cannot be stopped once a request has been made to query for a Link user's account. If we were to do this automatically when the checkout loaded, we would not be able to stop the Link OTP modal from appearing, even if a user were to select the WooPay express checkout button instead. If an email was recognised as an account for both services, both modals would appear simultaneously. 

We can allow WooPay to check the email field for presence of an account automatically, because if a user instead selects Link, we _can_ still tell WooPay not to show the OTP modal, if accounts are found on both platforms...or at least I hacked it so that we can. 😅 

Let me know if that eases your slumber or if you still find yourself wide awake at night pondering whether we should handle this interaction any better.

FangedParakeet:
> 1️⃣ Previous successful Link checkout results in loading Link modal faster than WooPay modal for future checkouts
This is actually the extension to the second compromise you explained in the PR description.

Yep, this is indeed a compromise that I begrudgingly made. We will only query Link when Link is specifically selected: selection occurs if either the express checkout button is clicked or the Link payment gateway is selected. If you pay with Link once, when you return to a new checkout Link will automatically be selected on page load, because this was the most recently used payment gateway. 

This is WC Core functionality and it makes sense to me that if a user had checked out with Link immediately prior, it's likely that they would want to use Link the next time that they returned to said store's checkout as well. Even though Link is selected and is quicker than WooPay as well, theoretically a user who wanted to change their express checkout method could still click the WooPay button at this point, in order to checkout with another payment method. However, I don't think this functionality is yet supported from the WooPay button, but I'm pretty sure it will be shortly.

FangedParakeet:
> 2️⃣ Save your info for secure 1-click checkout with Link checkbox behavior
Save your info for secure 1-click checkout with Link checkbox is not showing up when I open the checkout page with the registered email - this is ideally expected and works well.

Sadly this is a limitation of Stripe Link that I was unable to find a workaround for. If a user ever closes the Link OTP modal, this essentially terminates any present or future ability to checkout with Link. Once the OTP modal has been closed, there are no functions available on the LinkAutofill API (e.g. `launch`, `show`, etc.) that will allow us to reopen the OTP modal. If this modal is closed, there is no way that I was able to show it again, should the end-user decide to change their mind. At least, if there is a way to reshow this modal, I was not able to find it.

Once the OTP modal has been successfully cleared, however, if the user decides to close the Autofill modal before prefilling the checkout fields, we can still reshow it with the `show` function in order to give the user another chance to use this functionality. 

Basically, once the user is logged in, we have more control over the modals, but if they decline the initial request to log in, there's no coming back. If you like, we can create a new issue for this scenario, because I admit that it is irritating; although, I am unsure and pessimistic that there currently exists a solution for this problem.

FangedParakeet:
> 3️⃣ Link Autofill modal appears twice on the fresh setup
After I clear the cache and start from scratch, I observe the autofill modal appearing twice in a row. After that, the issue did not occur - at least, I could not reproduce it after. Below is the screen recording, which is worth more than a million words.

I found a way to replicate this: behold the following.

1. On page load, ensure Link payment gateway is **not** selected. Firefox will remember the previously selected radio button on every page refresh (without needing to submit the form) or you can complete a transaction using CC to ensure this gateway is selected on the next checkout.
2. Enter email address as normal. 
3. Select Link Express Checkout button. This should change the payment gateway to Link and open the OTP modal.
4. Enter `000000` to pass OTP modal.
5. Autofill modal will now open. Select "Autofill checkout".
6. The checkout will flutter and for some reason the UPE's [ready event](https://github.com/Automattic/woocommerce-payments/blob/4720/stripe-link-integration/client/checkout/classic/upe.js#L260) will trigger. Autofill will reopen, since `stripeLinkButton.enable()` is fired again.

The `ready` event is required in the case wherein a user selects the Link Express checkout button, before the UPE has mounted completely. In this event, we want to wait for the UPE to mount and load before kicking off the rest of the Link login process. I don't know why this would fire twice, but clearly and annoyingly it is, so I've added a workaround fix for this in a465fdde23844bf17e7cd9957f816f87d4b37912. 

Let me know how that looks to you!

FangedParakeet:
> 1️⃣ Is Save payment information to my account for future purchases checkbox expected per design here? It just was a bit unnatural for me to have both Link's saving checkbox and this one - but... this was only at first glance. Both checkboxes work as expected and Save payment information to my account for future purchases successfully saves the card for the future use.

Nice catch! There is no need to include the save payment information checkbox on the Link gateway. Either the user will be a registered Link user, in which case their payment information will already be saved, or they will select Link's own checkbox in order to save their payment details with Link.

Removed the checkbox from the gateway in 184f840e20d153efca83a684aab77cdb5067fc31.

FangedParakeet:
> 2️⃣ Do you happen to know, why Link requires from us to set USD as the currency used

This has basically been set [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-link-payment-method.php#L29) in the code in times prior to this PR. However, FYI, while I believe Stripe is beginning to roll out Link to other markets, for now WCPay should only offer Link to US-based stores, which in turn must use USD with Link. So shall it be written, so shall it be done. 🤷

FangedParakeet:
> Please let me know, if I can help to create an issue, define its scope or anything else - I am happy to be a second head here.
...
> I would also include some tests on the front-end side to ensure the behavior won't break in the future. 👍

My plan is to create new issues for both of these tasks upon completion of this PR. Initially, I aspired to get everything done in one shot, but the size and scope of this PR would have bloated considerably and become even tougher for any mortal--even a superhero such as yourself--to sanely review. Moreover there are some user-flow and implementation design choices that I have made in this PR that I would rather seek confirmation approval for, before copying them over to the Blocks checkout. 

Let me know how that sounds to you!

FangedParakeet:
I think I responded to everything, @timur27, but let me know if I missed anything or if you have any more concerns. Cheers!

timur27:
@FangedParakeet Many thanks for all the quick updates you did - great job! I'll try to sum everything up in this message at once so it's easier for us to keep track of all the active threads.

* The Express Link button [is not clickable when disabled](https://github.com/Automattic/woocommerce-payments/pull/5282#discussion_r1053711326) ✅ 
* Autofill modal works well after [separating the code that launches it](https://github.com/Automattic/woocommerce-payments/pull/5282#discussion_r1053711765) into the function ✅ 
* Button enabling/disabling works well after [aligning jQuery functions](https://github.com/Automattic/woocommerce-payments/pull/5282#discussion_r1053714141) ✅ 
* [Autofill modal appearing twice fix](https://github.com/Automattic/woocommerce-payments/pull/5282#issuecomment-1360298427) works well. Quite tricky it is. 😄 ✅ 
* Save payment information checkbox [isn't rendered now after discovering it's redundant](https://github.com/Automattic/woocommerce-payments/pull/5282#issuecomment-1360390675). ✅ 

----------------------------------------

Thanks for all the explanations to my questions and argumentation for keeping some areas as they are because of the limitations - all is clear to me. ✅ 

----------------------------------------

One small thing to confirm, but in my opinion, it shouldn't prevent me from approving this PR, so I look forward to aligning on this, but I will approve in the meantime.

> Click the Link express checkout button or Link payment gateway radio button. If the email input is empty, an error should appear prompting you to fill in this field. If email input is empty or set to an unregistered email address, "Save your info" checkbox should be present on Link payment gateway.

to be more specific, this part
> If email input is empty OR set to an unregistered email address, "Save your info" checkbox should be present on the Link payment gateway.

I noticed that the "Save your info" checkbox is present only if the input is set to an unregistered address and not present when the email input is empty (as defined per testing instructions). I find it natural, but given the testing instructions, I'd like to double-check if there wasn't a typo in the description or if I misunderstood something. Thanks for clarifying! 🙌🏼

----------------------------------------

Things that possibly require creating follow-up issues - feel free to update the list in case I missed anything and let me know if I can help creating these: 
1. [The issue](https://github.com/Automattic/woocommerce-payments/pull/5282#:~:text=Ideally%2C%20the%20express%20checkout%20methods%20should%20all%20be%20registered%20together%2C%20so%20that%20we%20can%20group%20them%20together...but%20I%20tried%20to%20achieve%20this%20quickly%20myself%20and%20it%20caused%20much%20pain%20and%20broke%20many%20things%2C%20so%20I%27ll%20be%20leaving%20this%20for%20another%20day%20as%20well.) with duplicated separators.
2. Create/update tests for Link on the back-end and front-end.
3. Make the Link method work on the Blocks checkout page.

----------------------------------------

I am also leaving it up to you if you want to [align with Magda](https://github.com/Automattic/woocommerce-payments/pull/5282#:~:text=since%20I%20have%20had%20to%20deviate%20delicately%20from%20your%20desired%20designs%2C%20so%20I%20would%20greatly%20appreciate%20your%20feedback%20on%20the%20destination%20I%20have%20arrived%20at.%20Hopefully%2C%20the%20GIFs%20and%20images%20in%20this%20lengthy%20description%20give%20you%20some%20idea%20of%20where%20that%20is.) on designs before merging this PR or after.

----------------------------------------

It was great to review this PR! 🙌🏼

timur27 (approved):

FangedParakeet:
Thanks for another ravishing review, @timur27!

> I noticed that the "Save your info" checkbox is present only if the input is set to an unregistered address and not present when the email input is empty (as defined per testing instructions).

This is true! I added [this line](https://github.com/Automattic/woocommerce-payments/blob/4720/stripe-link-integration/client/checkout/stripe-link/index.js#L287) to check for the presence of an email address, since I thought it might be confusing for a guest user that clicks on the Stripe Link button that nothing appears to happen on the screen, unless they enter a valid email address.

However, this is a limitation that I have created and I believe we can send empty email addresses to the `launchAutofill` function, so if you think this error message is not essential, I can remove it.

> Things that possibly require creating follow-up issues - feel free to update the list in case I missed anything and let me know if I can help creating these:
> 1. The issue with duplicated separators.

Created #5339.

>2. Create/update tests for Link on the back-end and front-end.

Created #5338.

> 3. Make the Link method work on the Blocks checkout page.

Created #5340.

I'm not going to merge this just yet, since doing so will close #4720 and I've just reached out to Magda [here](https://github.com/Automattic/woocommerce-payments/issues/4720#issuecomment-1365940750) to hopefully confirm the approach here and I believe she's AFK until later this week. If all goes well, however, I'll look to merge and move on with our work here by the end of the week. 👍

timur27:
@FangedParakeet 
> if you think this error message is not essential, I can remove it

No, all is good for me, and no further changes are needed. I'm glad we're aligned now. Thanks 🙌🏼

FangedParakeet:
I've made several amends to this PR following the feedback from @stephendharmadi over [here](https://github.com/Automattic/woocommerce-payments/issues/4720#issuecomment-1376772755). When Link and WooPay are both enabled, both will automatically use the email input value to query for account registration. However, Link will patiently wait--by polling a status set on the input by WooPay--until the WooPay query is complete. The meat of this lies in the additions to the `launchAutofill` function over [here](https://github.com/Automattic/woocommerce-payments/blob/4720/stripe-link-integration/client/checkout/stripe-link/index.js#L53-L71), facilitated by some new utility functions added [here](https://github.com/Automattic/woocommerce-payments/blob/4720/stripe-link-integration/client/checkout/utils/link.js).

_If the user **is** a WooPay user, but **is not** a Link user_

WooPay will query for registration, return success, and the WooPay modal should open. Link will not be queried and this modal will not open.

_If the user **is not** a WooPay user, but **is** a Link user_

WooPay will query for registration, return failure, WooPay modal should not open. Link will now query for registration and Link's authentication modal should open. Spinner preloader should reappear when Link begins its query.

_If the user **is** a WooPay user, **and also** a Link user_

WooPay will query for registration, return success and the WooPay modal should open. Link will not yet be queried. However, if the user closes the WooPay modal, they can select the Link express checkout button and this will trigger a query to Link and open Link's authentication modal.

@timur27, please could you grace this PR with some more of your wisdom and let me know if it still is worthy of your blessings. If you could test these changes with both WooPay enabled and disabled, that would be perfect. Cheers!

timur27 (approved):
> If the user is a WooPay user, but is not a Link user

> WooPay will query for registration, return success, and the WooPay modal should open. Link will not be queried and this modal will not open.

Works as expected ✅ 

> If the user is not a WooPay user, but is a Link user

> WooPay will query for registration, return failure, WooPay modal should not open. Link will now query for registration and Link's authentication modal should open. Spinner preloader should reappear when Link begins its query.

Works as expected ✅ 

> If the user is a WooPay user, and also a Link user

> WooPay will query for registration, return success and the WooPay modal should open. Link will not yet be queried. However, if the user closes the WooPay modal, they can select the Link express checkout button and this will trigger a query to Link and open Link's authentication modal.

Works as expected ✅ 

The same way, Stripe Link behaves as expected when WooPay is disabled. Code looks good as well.

One small thing, which caught my eye is the case, when: 
1. Both WooPay and Stripe Link are enabled.
2. Email is registered on both platforms. 
3. WooPay OTP opens as expected and then let's assume is closed by a user.
4. Stripe Link Button is clicked by the user and Stripe OTP is displayed as expected. 
5. Now the user closes the OTP from Link which for us means that Link can't be used anymore (the technical limitation we discussed). 
6. The user refreshes their email input field which triggers the WooPay `/user/exists`
7. After WooPay OTP opens because of the email being present in the platform, Link button starts to be clickable again, despite of being disabled on step 5 and not functional anymore. 
8. The user can now click the Link button which triggers the spinner but eventually, nothing happens.

We know that nothing should happen in this case because of the technical limitations, but the question is - should we enable the Link button for the second time on the WooPay OTP modal opening if we [k]now that Link isn't functional anymore? I'm rather asking this question from the UX point of view, and this seems a very tiny thing for me, which doesn't block the PR as such. 

With that being said, I'm approving this PR once again. @FangedParakeet I really appreciate all your efforts with this PR, since it was not only technically challenging, but also required a lot of alignment on the UX, and looking at all the obstacles and limitations I understand that we aren't able to make this super seamless and ideal. It looks like you achieved the best shape in terms of UX you can. Thanks a lot! 🥇

FangedParakeet:
>We now that nothing should happen in this case because of the technical limitations but the question is - should we enable Link button for the second time on the WooPay OTP modal opening if we now, that Link isn't functional anymore?

I had considered this, but sadly I believe my hands are tied in this scenario. 

>should we enable Link button for the second time on the WooPay OTP modal opening if we [k]now, that Link isn't functional anymore?

While I have figured out a slightly bodgy (yet generally reliable) way to detect when Stripe's OTP modal is open, I don't have a very good way of learning that it has been closed by a user. The only way to learn this would be to poll the page repeatedly to check whether the OTP modal is present on the page to register that the OTP has decided to open: by continuing to poll the page for the modal's presence, we can infer that if the modal is no longer on the page and the user has not been authenticated, then the modal must have been closed. At this point, we could permanently disable the button. 

As far as I can tell, that's the only way to implement this change, though it might exhaust the page, since there will need to be an `interval` running constantly while Link is enabled. @timur27 let me know if you think this might be worth the effort or if you had a wiser way in mind to implement this feature!

Otherwise I am unable to distinguish between the scenario where the WooPay modal opens twice and, rather than being intentionally closed, the email address is simply not recognised by Link in the first instance, but registered on the second attempt.

timur27:
@FangedParakeet I don't have any solution from the top of my head to the concern we are discussing other than polling the page for the presence of the OTP modal window. And I fully agree with you that it's a suboptimal way.

FangedParakeet:
>And I fully agree with you that it's a suboptimal way.

OK, this wasn't quite the tacit acceptance of defeat that I was hoping you would give me here. 😂 I've reached out to design for final feedback [here](https://github.com/Automattic/woocommerce-payments/issues/4720#issuecomment-1385455888). In the meantime, since this idea is now playing on my mind as well, let me have another crack at permanently disabling the button when we estimate the modal being closed tomorrow morning and I'll see if the solution I described is workable or woeful. 🙏

timur27:
@FangedParakeet Please forgive me for not being clear here. In my last message, I wanted to emphasize that I agree with 

> As far as I can tell, that's the only way to implement this change, though it might exhaust the page, since there will need to be an interval running constantly while Link is enabled

and it's ok for me to move forward without any change if it is acceptable for the design team. As you now mentioned, since we have some time in between while waiting for the design feedback and we want to use it to improve the behavior, I'm more than happy and ready to re-review the PR quickly whenever/if needed!

FangedParakeet:
With much gratitude for your convincing coaxing, @timur27, I have added some code in 8fa07386dea58bb668ee43a8ce1e91be32a6de6d to watch out for the Link OTP modal and disable Link when we are confident that it has been closed by the user. I think you can constrain a re-review to specifically that commit, so let me know if that looks alright to you. Cheers!

timur27 (approved):
Works well! Once again, many thanks for the patience with all these small nitpicks. @FangedParakeet, I am happy to approve this one. 🚀

S-code652 (approved):

PR #5285, opened 2022-12-14
hide capture charge and cancel authorization on status complete/processing

### LINKED ISSUE #4789: [Authorizations] Updating order status to `Processing` before Capturing - Cannot refund order.
### Describe the bug

You can capture payment after setting the order status to processing. However:

* This does not add confirmation about the amount captured in order notes. 
* This also does not let you refund the order

### To Reproduce

1. Enable authorization
2. Create order in the frontend  (don’t capture the charge)
3. Change the order status to `Processing`
4. Capture charge now
5. Notice that order notes is not updated with confirmation about the capture:

    

6. Attempt to capture the order multiple times. No update on the order itself. 
7. On Stripe, you can see that attempt to capture failed since cannot be captured since it is already captured:

    
8. Access _Payments > Transactions_ to confirm that the order is indeed captured:
    
9. Notice that you cannot refund the order from the order details page even though it is successfully captured

    

### Actual behavior

* No confirmation about the captured amount
* Unable to refund the order

### Expected behavior

If the order status is updated to processing before capturing the charge, order notes should still confirm the captured amount if it was successfully captured, and it should let merchants refund the captured amount.

* OS: [e.g. iOS]: N/A
* Browser [e.g. chrome, safari]: N/A
* Version [e.g. 22]: N/A

### Additional context

Reported in 5575478-zen

dpaun1985:
Fixes #4789 

For a shop that has capture later option enabled, the merchant can change the order status from 'On hold' to Processing/ Complete. This could mean that the merchant already received the payment from another source.  If the order is already completed, it makes no sense to charge the client twice, and we need to hide the charge options from the order.

**Original issue:** merchant updated the order to a complete status, and after that tried to capture the charge. The capture was charged, but no logs were added to the order since it had a complete status. This led to a misleading of the merchant. 

Hide Capture charge and cancel authorization options for an order, when the merchant manually changes the status from 'On Hold' to 'Processing'/'Completed'. 

#### Testing instructions

* Go to `Woocommerce -> Settings -> Payments` and under 'Transactions' section enable `Issue an authorization on checkout, and capture later` for the shop
* Create an order 
* Change the order status from 'On Hold' to 'Processing' . Do NOT capture the charge
* The `Capture charge` and`Cancel authorization` should not be shown.  

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-4789-status-processing-before-capture
+++ b/changelog/fix-4789-status-processing-before-capture
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Hide capture charge and cancel authorization options on order when order is completed
--- a/changelog/fix-4789-status-processing-before-capture
+++ b/changelog/fix-4789-status-processing-before-capture
@@ -1,4 +1,4 @@
 Significance: minor
 Type: fix
 
-Hide capture charge and cancel authorization options on order when order is completed
+Hide "capture charge" and "cancel authorization" actions on order details page when order status is processing or completed.
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2101,7 +2101,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		if (
 			in_array(
 				$theorder->get_status(),
-				[ WC_Payments_Order_Service::STATUS_PROCESSING, WC_Payments_Order_Service::STATUS_COMPLETED ],
+				wc_get_is_paid_statuses(),
 				true
 			)
 		) {
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2098,13 +2098,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		}
 
 		// if order is already completed, we shouldn't capture the charge anymore.
-		if (
-			in_array(
-				$theorder->get_status(),
-				wc_get_is_paid_statuses(),
-				true
-			)
-		) {
+		if ( in_array( $theorder->get_status(), wc_get_is_paid_statuses(), true ) ) {
 			return $actions;
 		}
 
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -2097,6 +2097,17 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			return $actions;
 		}
 
+		// if order is already completed, we shouldn't capture the charge anymore.
+		if (
+			in_array(
+				$theorder->get_status(),
+				[ WC_Payments_Order_Service::STATUS_PROCESSING, WC_Payments_Order_Service::STATUS_COMPLETED ],
+				true
+			)
+		) {
+			return $actions;
+		}
+
 		$new_actions = [
 			'capture_charge'       => __( 'Capture charge', 'woocommerce-payments' ),
 			'cancel_authorization' => __( 'Cancel authorization', 'woocommerce-payments' ),
--- a/includes/class-wc-payments-order-service.php
+++ b/includes/class-wc-payments-order-service.php
@@ -17,12 +17,11 @@ class WC_Payments_Order_Service {
 	/**
 	 * Order status constants.
 	 */
-	const STATUS_CANCELLED  = 'cancelled';
-	const STATUS_COMPLETED  = 'completed';
-	const STATUS_FAILED     = 'failed';
-	const STATUS_ON_HOLD    = 'on-hold';
-	const STATUS_PENDING    = 'pending';
-	const STATUS_PROCESSING = 'processing';
+	const STATUS_CANCELLED = 'cancelled';
+	const STATUS_COMPLETED = 'completed';
+	const STATUS_FAILED    = 'failed';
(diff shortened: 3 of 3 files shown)


includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2097,6 +2097,17 @@ public function add_order_actions( $actions ) {
 			return $actions;
 		}
 
+		// if order is already completed, we shouldn't capture the charge anymore.
+		if (
+			in_array(
+				$theorder->get_status(),
+				[ WC_Payments_Order_Service::STATUS_PROCESSING, WC_Payments_Order_Service::STATUS_COMPLETED ],
```
kalessil: The hard coded list is not friendly to 3rd party extensions - the statuses can be overridden by `woocommerce_order_is_paid_statuses` hook. Consider using `wc_get_is_paid_statuses` function instead.

kalessil (changes requested):

includes/class-wc-payment-gateway-wcpay.php
dpaun1985: fixed in [8f955ce](https://github.com/Automattic/woocommerce-payments/pull/5285/commits/8f955cea51219f79877bffd9b553491bc7b2a940)

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2097,6 +2097,17 @@ public function add_order_actions( $actions ) {
 			return $actions;
 		}
 
+		// if order is already completed, we shouldn't capture the charge anymore.
```
kalessil, line 2100: Thanks for placing the comment!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2097,6 +2097,17 @@ public function add_order_actions( $actions ) {
...
+		if (
```
kalessil: Can we put `if ( ... ) {` on one line, please?

kalessil (approved):
LGTM! Note, I did code review only - no testing.

includes/class-wc-payments-order-service.php
```diff
@@ -17,11 +17,12 @@ class WC_Payments_Order_Service {
...
-	const STATUS_COMPLETED = 'completed';
-	const STATUS_FAILED    = 'failed';
-	const STATUS_ON_HOLD   = 'on-hold';
-	const STATUS_PENDING   = 'pending';
+	const STATUS_CANCELLED  = 'cancelled';
+	const STATUS_COMPLETED  = 'completed';
+	const STATUS_FAILED     = 'failed';
+	const STATUS_ON_HOLD    = 'on-hold';
+	const STATUS_PENDING    = 'pending';
+	const STATUS_PROCESSING = 'processing';
```
kalessil: Are those states correlating with WooCommerce core constants?

dpaun1985: the statuses relay to the order status, but since I won't use the constants anymore, I will undo the changes in this file.

includes/class-wc-payment-gateway-wcpay.php
dpaun1985: fixed in [222979d](https://github.com/Automattic/woocommerce-payments/pull/5285/commits/222979d301077252af3e6c02e12c16703d6851aa). thx

includes/class-wc-payments-order-service.php
kalessil: :+1:, mind creating a new low-priority issue (for a coll-down week, perhaps) to review and perhaps migrate the WooCommerce Payments codebase on WooCommerce core constants?

kalessil: Pre-condition for such migration is that constants from the core are including all of the values from this specific place.

dpaun1985: created an [issue](https://github.com/Automattic/woocommerce-payments/issues/5286) for this. Thank you!

kalessil: Thanks!

elazzabi (approved):
LGTM!

Question: This removes the button, but if I understand correctly, this can still happen if we send a direct API call. Should we block it at the API level too?

dpaun1985:
> Question: This removes the button, but if I understand correctly, this can still happen if we send a direct API call. Should we block it at the API level too?

The action is made via a POST action, not an API call. I expect that having nonce in the form, there will not be any security issues. 
However, if you consider, we can double-check also when the charge action is fired. What do you think @elazzabi  ?

changelog/fix-4789-status-processing-before-capture
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Hide capture charge and cancel authorization options on order when order is completed
```
anu-rock: Nitpick:

```suggestion
Hide "capture charge" and "cancel authorization" actions on order details page when order status is processing or completed.
```

anu-rock (commented):
The code changes LGTM and are aligned with the proposed solution of hiding authorization-related actions on the Order Details page.

My testing also confirms that it works.

* **On hold → Processing** - auth actions hidden ✅ 
* **Processing → On hold** - auth actions visible again ✅ 
* **On hold → Completed** - auth actions hidden ✅ 
* **Completed → On hold** - auth actions visible again ✅ 

However, even after marking an order as completed, it's still possible to capture it through its corresponding Payment Details page (Payments → Transactions).

@mgascam @kalessil - Should we deal with the above in a separate issue?

PR #5287, opened 2022-12-14
Update save my info section

### LINKED ISSUE #1505: Implement sorting mechanism for the checkout options list in settings
The items in the checkout options list can be sorted in the WooCommerce Payments settings screen. Users will be able to drag an element in the list and drop it in the desired position.

The checkout options will be displayed in this order on the checkout screen.

## Screen elements
- An icon at the beginning of each item representing the handle the user will press to drag the element.

## Screenshot
shot-2021-04-06-13-26-24

cesarcosta99:
Fixes 1505-gh-Automattic/woopay

Updates save my info section at checkout:

- Applies new design to both classic and Blocks checkout.
- Change its copies.
- Update tests to match new copies and the way interactions take place.

> **Note**
> 1. The copy next to the checkbox is still under discussion.
> 2. We don't know what's the link for the Learn more text yet.
>
> Check the PRD comments (link in the issue).

#### Testing instructions

Before running the tests, make sure you have Stripe events being delivered to your local WCPay server. That way the WooPay user can be correctly created.

**Run the following tests on both classic and Blocks checkouts.**

**Test the WooPay user is created**

1. Go to the store and add a product.
2. Go to checkout and fill in the form, type an email that does not belong to a WooPay user and a valid US phone.
3. The save my info section must be visible.
4. Check the UI matches the design. There will be little distinctions between classic and Blocks, but overall they should look the same.
5. Confirm the UI looks good in the mobile version.
6. Tick the checkbox and note the phone value has been copied over the phone inner input.
7. Place the order.
8. Check the WooPay user has been created.

**Test the save my info section is NOT displayed**

1. Go to the store and add a product.
2. Go to checkout and fill in the form, type an email that belongs to a WooPay user and a valid US phone.
3. The save my info section must not be visible.
4. Place the order.
5. Confirm the WooPay user was not created.

**Test the tooltip functionality**
1. Go to the store and add a product.
2. Go to checkout and fill in the form, type an email that does not belong to a WooPay user.
3. Go to the save my info section and pass the mouse over the `i` icon, next to the WooPay logo.
4. A tooltip must show up when you hover the mouse in the icon, and it should continue visible when you move the mouse to within the tooltip.

<table>
<tr>
<th>Classic checkout</th>
<th>Blocks checkout</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

-------------------

**Post merge**


client/components/platform-checkout/save-user/checkout-page-save-user.js
```diff
@@ -153,36 +167,107 @@ const CheckoutPageSaveUser = ( { isBlocksCheckout } ) => {
...
+					<Icon
+						icon={ info }
+						size={ isBlocksCheckout ? 24 : 36 }
+						className={ `info-icon ${
+							isInfoFlyoutVisible ? 'focused' : ''
+						}` }
+						onMouseOver={ setInfoFlyoutVisible }
+						onMouseOut={ setInfoFlyoutNotVisible }
+					/>
+					<div
```
cesarcosta99, line 230: There is the [flyout component](https://developer.wordpress.org/block-editor/reference-guides/components/flyout/) from the WP components, however the package version we use does not include it.

I thought about updating the version of the package, but I've seen some deprecated properties and didn't want to go through the risk of breaking things visually. It'd be hard to catch bugs if such thing happened.

client/components/platform-checkout/save-user/container.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+const Container = ( { children, isBlocksCheckout } ) => {
```
cesarcosta99, line 6: Mayisha shared her findings on 1447-gh-Automattic/woopay#issuecomment-1339026618 regarding the approach we should take when developing this.

I went a bit further and found out that it's not just the `FormStep` that's not exported, but we can't import anything from `@woocommerce/base-components/cart-checkout` as it would not be found.

>We can copy that existing component and make out own wrapper to render our remember me section as a stepped block but doing so would create UI mismatch if that component gets changed in any future version of the blocks plugin.

I went down on this approach for two reasons:

- We already rely on Blocks checkout classes for this feature [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/platform-checkout/index.js#L25-L32).
- The other possibility I see is integrating it via the Blocks plugin itself, like it's mentioned there. However, IMO, I don't see as a viable option for a small feature like this. This feature has changed a lot, and I don't think the required effort for implementing such approach worths this long shot.

client/settings/phone-input/index.tsx
```diff
@@ -96,6 +95,12 @@ const PhoneNumberInput = ( {
 		};
 	}, [ onValueChange, onValidationChange ] );
 
+	useEffect( () => {
+		if ( inputInstance && inputRef.current ) {
+			onValidationChange( inputInstance.isValidNumber() );
+		}
+	}, [ value, inputInstance, inputRef, onValidationChange ] );
+
```
cesarcosta99, line 105: Had to move this `onValidationChange` call to a new `useEffect` that included `value` in the dependency array because before this component was rendered after the checkbox was clicked – which means the validation was done at the mount.

Now that there is an animation, the component is always there and needs to rely on the `value` change to check the initial validation.

rafaelzaleski (commented):
When we copy over the number to the sign-up field, we need to either clean-up any non-numerical character or trigger an event in this field that will perform it's internal validation (`change` maybe?)

I tested with the phone `+1.646.532.9563` in the checkout field and the dots were copied over to the sign-up field. If a click in the sign-up field and remove one dot, the others are removed automatically. If I don't touch the sign-up field I'm able to submit the checkout with dots in the sign-up phone field.

rafaelzaleski (commented):
Overall it looks good and tests well. I left a comment about the validation when the phone number is copied from the WC checkout field.

I also think we should use relative units (percentages, rem, etc.) in CSS since the plugin is used with a huge variety of themes and it'd reduce the chances of it looking bad. What do you think?

cesarcosta99:
@rafaelzaleski thanks for reviewing. I've addressed your concerns in the last commits, could you please give it another review?

>I also think we should use relative units in CSS

I agree on this and changed it, however I left a few things out of this change:
- Borders, because they won't affect other elements and the general experience.
- The places where I've used the `$gap` variables, because ideally we should aim to keep a standard UI experience with default padding/margin etc. They're still in px, and I think we should change them as well, in the package, not exactly here.

rafaelzaleski (commented):
@cesarcosta99, is there a way to add the class `wc-block-components-text-input` to `PhoneNumberInputWrapper` if we are in the checkout block? 

I tested with another theme and the input didn't look good without this class.

Screenshot 2022-12-19 at 17 50 08

rafaelzaleski (approved):
Looks good and tests well! 🚢

PR #5288, opened 2022-12-14
Prevent processing the wrong kind of webhook for a gateway

### LINKED ISSUE #5266: Webhook processing should respect the test/live mode setting
### Describe the bug

First described in https://github.com/Automattic/woocommerce-payments-server/issues/2664, but recreated here because this is a client issue.

If a site is cloned using a method that results in a "duplicate" WCPay account, test payments on the cloned/staging site can mark **Failed** orders on the original/live site as **Paid/Processing**.

See this P2 post for more information about "duplicate" WCPay accounts: paJDYF-1Ss-p2

### To Reproduce

> **Note**
> I haven't attempted to reproduce this as yet.

1. Create a new WCPay account with a public-facing URL. Use WCPay Dev Tools to find the Blog ID.
2. Create a new order and mark it as **Failed**.
3. Clone this site to a staging site using the [All-in-One WP Migration plugin](https://wordpress.org/plugins/all-in-one-wp-migration/) (see paJDYF-1Ss-p2 for more info).
5. Ensure that the Blog ID is identical to the original site.
6. Ensure that WCPay test mode is enabled.
7. Pay for the Failed order. The order is no longer marked as **Failed** in the cloned/staging site.
8. ⚠️ The original site's Failed order is also no longer marked as **Failed**.

### Actual behavior

Orders on the original site can be modified when transactions occur on cloned staging sites in test mode.

### Expected behavior

Orders on the original site are not modified when transactions occur on cloned staging sites in test mode.

### Additional context

Related to https://github.com/Automattic/woocommerce-payments-server/issues/1596 .

#wcpay thread discussion: p1667267897841589-slack-CGGCLBN58
#woo-devs thread with details about the merchant's experience: p1667258886566859-slack-C7U3Y3VMY
ZD ticket: 5668326-zd-woothemes

jbordonado:
Fixes #5266 

- Before processing a webhook event, we verify that the event's mode (defined by the `livemode` flag in the event's payload) is identical to the gateway's mode. An exception is thrown if they don't match.

#### Testing instructions

As described in the issue:

* We need a live site
* Add a failed order to the live site
* Clone the live site to create a staging site
* Pay for the failed order on the staging site
* The order on the live site should still be marked as failed

The live and clone site should have WCPay dev installed so that we can redirect the API to the server (exposed through ngrok) using that branch

-------------------

**Post merge**


jbordonado:
Leaving here the zip file I built with these changes: 
[woocommerce-payments.zip](https://github.com/Automattic/woocommerce-payments/files/10245151/woocommerce-payments.zip)

deepakpathania:
Requested a review from Gamma as well since some folks there might have thoughts around the same.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -198,6 +200,30 @@ public function process( array $event_body ) {
...
+	 * @throws Invalid_Webhook_Data_Exception Event mode does not match the gateway mode.
+	 */
+	private function check_webhook_mode_mismatch( array $event_body ) {
+		$is_gateway_live_mode = ! $this->wcpay_gateway->is_in_test_mode();
+		$is_event_live_mode   = $this->read_webhook_property( $event_body, 'livemode' );
+
+		if ( $is_gateway_live_mode !== $is_event_live_mode ) {
+			$event_id = $this->read_webhook_property( $event_body, 'id' );
+
+			throw new Invalid_Webhook_Data_Exception(
```
RadoslavGeorgiev: I think it might be worth just logging the mismatch, and bailing without throwing an exception. Switching from one mode to another, and receiving a webhook for the old mode is not such an exceptional case, and we probably don't need to raise any alarms in that case.

RadoslavGeorgiev (approved):
I have a minor nitpick, but the PR looks great overall. Feel free to (not) implement this change, and :shipit: 

I cannot figure out any further ways to test the PR, but at the same time I don't think there's much that could go wrong with `is_in_test_mode()` and the `livemode` property. As long as this is being tested, and the test is covering for glaring holes in the implementation, we should be fine.

includes/class-wc-payments-webhook-processing-service.php
jbordonado: It makes sense.
I slightly changed the code to log an error and bail in case there's a mismatch if you'd like to review it again.
Reg. unit tests, I didn't find better ways to verify the change than checking that the process for an easy event (here a notification) is not made in case of a mismatch (test event + live gateway or the opposite). It was easier to check with exceptions, but that should be enough here

RadoslavGeorgiev (approved):
The changes look great! 🚢 🎊

includes/class-wc-payments-webhook-processing-service.php
RadoslavGeorgiev: Thank you for switching to logging! The test is (and was) sufficient IMO, marking this as resolved.

PR #5289, opened 2022-12-14
Express button product page

Mayisha:
This PR implements WooPay express checkout button on the Product page and it works the same as Google Pay/Apple Pay. On WooPay express button click it adds that product to the merchant's cart first and then initiates the platform checkout. Different quantities of the product, different variation and add-ons also works exactly same as other payment request buttons.

Some notable changes are-
- Added a hook `use-express-checkout-product-handler.js` which handles the product page related extra pre-processing. The logics and functions here are heavily borrowed from [payment-request file](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-request/index.js#L134). The only difference is we are not using jquery in the hook. Otherwise, the logics depend on the status of the same dom elements as the payment request button.
- had to make a change in the `checkout/utils/request.js` to properly translate a nested object to form data.

#### Testing instructions
- Install the latest [dev-tools](https://github.com/Automattic/woocommerce-payments-dev-tools). Go to the dev tools page and click the Enable the WooPay Express Checkout button checkbox and save the settings.
- Go to the WooCommerce Payments settings screen and click the Customize button of WooPay.
- Select `Product` page as the button location and save the settings.
- Go to a simple product page and ensure that you can checkout with WooPay express checkout button.
- Go to a variable product page, select different variations and ensure that you can checkout with WooPay express checkout button.
- Clicking the WooPay express checkout button adds that product to the merchant cart and WooPay checkout should have the same product in the cart (with the selected quantity)
- Install the `Product Add On` extension, add some add-ons to a product. Go to that product page, select some add-ons and click WooPay express checkout button. Ensure that the cart has this product with the add-ons.


client/checkout/platform-checkout/express-button/use-express-checkout-product-handler.js
```diff
@@ -0,0 +1,113 @@
+/**
```
Mayisha, line 1: The logic here is mostly copied from [payment-request functions](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-request/index.js#L134) but translated from jquery to React.

client/checkout/platform-checkout/express-button/use-express-checkout-product-handler.js
```diff
@@ -0,0 +1,113 @@
...
+		setIsAddToCartDisabled( getIsAddToCartDisabled() );
+
+		const onVariationChange = () =>
+			setIsAddToCartDisabled( getIsAddToCartDisabled() );
+
+		const variationList = document.querySelector( '.variations_form' );
+
+		if ( variationList ) {
+			variationList.addEventListener( 'change', onVariationChange );
+		}
```
Mayisha, line 94: In the [payment-request file](https://github.com/Automattic/woocommerce-payments/blob/develop/client/payment-request/index.js#L378), we are listening to the `woocommerce_variation_has_changed` event to identify any change in selected product variation.

However, this event is dispatched with the jquery `trigger` method which is not retrievable with native JS event listeners. So I have used the event listener on the variation list here.

Any thoughts on using jquery inside React hook/component here? I have tested locally that it works if I use jquery `on('woocommerce_variation_has_changed')` in this React hook.

client/checkout/utils/request.js
```diff
@@ -1,9 +1,21 @@
-export default async function request( url, args, options ) {
-	const data = new FormData();
-	for ( const key in args ) {
-		data.append( key, args[ key ] );
```
Mayisha, line 4: I had to update this function to consider the nested objects too when translating an object to formData. I discovered that this function was not considering the nested fields when [this](https://github.com/Automattic/woocommerce-payments/blob/061efdc2bc6cfdd7eff7542f045364631cbab725/client/checkout/platform-checkout/express-button/use-express-checkout-product-handler.js#L70) `paymentRequestAddToCart` function was failing to add the product to the cart because the [attribute](https://github.com/Automattic/woocommerce-payments/blob/061efdc2bc6cfdd7eff7542f045364631cbab725/client/checkout/platform-checkout/express-button/use-express-checkout-product-handler.js#L44) field was not a proper form field (it was coming as `[object object]`.

client/checkout/platform-checkout/express-button/use-express-checkout-product-handler.js
bborman22: I think what you have here works and I personally would want to avoid `jquery` in React if possible? I **think** this listener will pick up any call that triggers `woocommerce_variation_has_changed` based on this [block of code](woocommerce_variation_has_changed). My only concern would be around other plugins that might be using that custom event in different ways that we might have to listen in on, but couldn't find / think of any use cases? My preference would be to leave it as-is, but I'm totally open to listening to the custom event if there is more concern there.

bborman22 (commented):
Everything worked really well for me overall. I was able to run through the tests with simple and variable products and it all worked as expected. I left one comment on the custom jQuery event discussion and just had two other things to note:

> Platform checkout won't have the add ons as the support is not implemented yet

I was curious which product add ons plugin you used, I was assuming this one: https://woocommerce.com/products/product-add-ons/ I had seen some mention in the [extension compatibility project about Gravity Forms Product Add Ons](pdpOw2-1yu-p2#comment-1841), but wasn't sure if this is a plugin we should bring up in the project?

There is also an issue I encountered that was similar to something I ran in to before. If you test out these changes with the other Payment Request methods disabled, the feature won't work due to the following error: 

```
Uncaught TypeError: Cannot read properties of null (reading 'toString')
    at getPaymentRequestAjaxURL (utils.js?64e6:90:2)
    at WCPayAPI.paymentRequestAddToCart (index.js?a66a:620:48)
    at addToCart (use-express-checkout-product-handler.js?844e:67:14)
    at initPlatformCheckout (woopay-express-checkout-button.js?f5d0:60:4)
```

This error is essentially thrown as it tries to use `getPaymentRequestData( 'wc_ajax_url' )` in the `getPaymentRequestAjaxURL` function. I was wondering if we should be creating a parallel `api.wooPayExpressAddToCart` so that we don't have to depend on the `api.paymentRequestAddToCart` or if that would create too much repetition?  This has some similarities to what we were trying to solve with #5236. I'm not sure if there is anything to gain from looking at that solution?

I think we should come up with a solution to this scenario before merging, but was curious what you thought? Minus that one thing, this looks good to me!

Mayisha:
@bborman22 you are right about the product add-on. I am also checking with https://woocommerce.com/products/product-add-ons/ and this one is supposed to work fine. :ok_hand: 

I will fix the other issue :eyes:

Mayisha:
@bborman22 I have fixed the `Uncaught TypeError: Cannot read properties of null` issue that happens because of the unavailable `getPaymentRequestData( 'wc_ajax_url' )` when other payment request methods are disabled in https://github.com/Automattic/woocommerce-payments/pull/5289/commits/45ff4c7a5861123d2e12511e0da1c5e0ac4d6060. Let me know what you think about this solution. :thinking:

bborman22:
That approach definitely works for me. But in testing it I ran into another issue in that if the other payment request methods are disabled the `wc_ajax_wcpay_add_to_cart` hook will not be initialized and so the AJAX call to add to cart doesn't do anything. I did a quick test with it disabled and clicking the button didn't trigger anything, then I enabled the other payment requests and the button worked as expected. I'm not sure how far to go down in replicating the features from Payment Requests into Express Checkout, but since we already have that class set up with similarities I thought it might not be too much to add this and perhaps we could refactor at some point to create a shared "payment request" class? Let me know what you think though!

Mayisha:
I totally missed this, thanks @bborman22 for pointing that out. 

> since we already have that class set up with similarities I thought it might not be too much to add this and perhaps we could refactor at some point to create a shared "payment request" class?

I have added the required functions to the `WC_Payments_Payment_Request_Button_Handler` class in https://github.com/Automattic/woocommerce-payments/pull/5289/commits/5ceb87ce82b2e401d8b45264073edbd90dd0ce81. I have copied over the functions here. Refactoring at some point sounds like a good idea to me as now we have some common functions in both of the classes.

bborman22 (approved):
Awesome, it worked great for me now! LGTM! 🎉

PR #5296, opened 2022-12-15
Update UPE logic around the cards payment method and improve onboarding styling

### LINKED ISSUE #4114: Allow merchants to only enable credit card on UPE
### Describe the bug
If you try to enable UPE but don't want to enable a new payment method only credit card. You have to go back in the browser history or select a new payment method and then disable it later. This is not very intuitive and can make merchants end up with misconfigured sites.

### To Reproduce
- With Upe disabled
- Go to Payments > Settings
- Click on Enable in your store under "Enable the new WooCommerce Payments checkout experience"
- Try to save without selecting any other payment method

### Screenshots
Screen Shot 2022-04-08 at 5 41 31 PM

### Expected behavior
Merchants should be able to activate upe with credit card only

vladolaru:
Fixes #4114 

**Add active and locked card payments method when UPE is active + styling improvements**

The initial issue (#4114) surfaced the fact that a merchant could not complete the UPE onboarding _without_ selecting one of the specific UPE payment methods even though it was not a requirement since the cards payment method would suffice.

The Stripe UPE logic imposes that the cards payment method be mandatory, but our UI didn't reflect that reality. I decided to adjust _both_ the UPE onboarding wizard and the payments settings view to be consistent with this requirement. Previously, the cards payment checkbox list item was not present in the wizard, and the cards checkbox control in the settings view was not disabled when UPE was in use.

This PR fixes these logical and UX inconsistencies by **forcing the cards payment method checkbox to be checked and disabled** both in the wizard and the settings when UPE is active. To let the user know why this behavior is present, a `Required` label is added next to the cards payment method name, in both views.

At the same time, to help the user be better informed about requirements and possible next steps, **the wizard button label changes** from `Add payment methods` to `Continue` to capture better the multiple scenarios a user might face (either selecting UPE-specific payment methods or just sticking with card payments and test things). To ease the possible fear of users who are just testing, we **introduce a description under the payment methods selection box** with the mention: `You can always change or add more payment methods later.`

As [discussed](https://github.com/Automattic/woocommerce-payments/issues/4114#issuecomment-1344101862) with @rogermattic, all these changes bring _consistency and predictability_ to users who want to experiment with the UPE early access.

Throughout the initial exploration of the issue, several design inconsistencies surfaced. It seemed like a good moment to steer the current state of the UPE onboarding wizard closer to the original design intent. As a result, this PR introduces **styling changes around spacing, font weight and sizes, and alignments.** We are not at pixel-perfect fidelity (that was not my intention), but we are closer and can convey that increased level of quality that builds trust with our merchants.

The initial state of the UPE onboarding wizard
Markup on 2022-12-15 at 22:04:11

The updated state of the UPE onboarding wizard
Markup on 2022-12-15 at 22:01:39

The Payments Settings view after the changes
Markup 2022-12-15 at 21 57 21

#### Dev notes

In surfacing the cards payment method list item in the UPE onboarding wizard, I explored simply adding the `card` entry to the UPE payment methods list and letting the rest of the logic just display it: this _does_ make sense since the card payment method _is_ a valid (and required) UPE payment method, but it would have touched _a pretty large surface of the code;_ I understood that the UPE payment methods list is treated as _UPE-specific_ or _UPE-additional_ payment methods, so I left it as such.

In the end, I chose to _hard-code_ the cards payment method list item at the top since it is in line with the Stripe UPE requirements.

I had to introduce the behavior of `required` and/or `locked` payment methods. I chose to separate a payment method being required from it being locked (its input control disabled) to offer more future flexibility to our components in dealing with either combination. This behavioral addition is made at the general level, in our components.

Through minor refactoring of some variables, I aim to clarify that a payment method which is determined to be disabled (due to UPE-specific logic) doesn't necessarily refer to its input control (checkbox in our case) having the `disabled` HTML attribute.

In terms of CSS styling changes and additions, I aimed to keep my rules specific to avoid the risk of accidental overflows into other areas. There might be an argument to be made to make the styling more reusable, but I opted to be defensive and reduce the styling specificity when the need arises.

#### Testing instructions

1. On a dev site with WooCommerce Payments, checkout this PR's branch: `update/4114-upe-onboarding-with-just-credit-cards-method`
2. Don't enable additional UPE payment methods and finish the KYC flow to enable your account, or ensure that "Enable UPE checkout" is not checked in the **WCPay Dev Utils**.
3. Go to **Payments > Settings**.
4. The `Credit card / debit card` payment method should not be required, and you can activate or deactivate it at will.
5. Deactivate the `Credit card / debit card` payment method.
6. Click the button label "Enable in your store" to go to the UPE onboarding wizard.
7. In the wizard, the `Credit card / debit card` payment method should be checked, required, and without the possibility of unchecking it; at the same time, the "Continue" button should be active and allow you to finish your UPE onboarding without the need to select any UPE-specific payment methods.
8. Click the "Continue" button, then "Go to payments settings" button to return to the Settings view.
9. The `Credit card / debit card` payment method should be checked, required, and without the possibility of unchecking it.
10. Disable UPE from the three-dots right-side dropdown and the subsequent modal.
11. The `Credit card / debit card` payment method should remain active but no longer required or with the checkbox disabled.

Here is a quick rundown through the steps above:
Screen Capture on 2022-12-15 at 22-35-34

-------------------

**Post merge**


client/payment-methods/index.js
```diff
@@ -261,6 +261,9 @@ const PaymentMethods = () => {
 											getStatusAndRequirements( id )
 												.status
 									}
+									// The card payment method is required when UPE is active and it can't be removed.
```
ismaeldcom: I'm not sure why we were allowing `card` to be disabled, because it causes errors. But this change could cause some issues. Probably not a common case, but if for any chance a merchant as `card` disabled, after this change they will not be able to enable it anymore.

client/components/payment-methods-list/payment-method.js
```diff
@@ -32,14 +32,21 @@ const PaymentMethod = ( {
 	onUncheckClick,
 	className,
 	isAllowingManualCapture,
+	required,
+	locked,
```
ismaeldcom, line 36: I wonder if we really need those new props. We only have one case of a payment method being required and locked, that's `card`. As we are already doing something different for it here: https://github.com/Automattic/woocommerce-payments/blob/c4d3390e49359f00ce8875dbf78402500189a3e0/client/components/payment-methods-list/payment-method.js#L60

I think it would be simpler to just disable it, and add the required label when the payment method is `card`.

This is just my thought, I normally try to not overload a component with more props if those are not required.

client/components/payment-methods-list/payment-method.js
```diff
@@ -32,14 +32,21 @@ const PaymentMethod = ( {
...
 } ) => {
-	const disabled = upeCapabilityStatuses.INACTIVE === status;
+	const paymentMethodDisabled = upeCapabilityStatuses.INACTIVE === status;
```
ismaeldcom: **Nitpick**: I don't see the benefit of renaming this definition. I read `disabled` under `PaymentMethod` component as `payment method disabled`. Adding the same of the component to it seems a bit redundant. Anyway, it's just my opinion 🙂

client/additional-methods-setup/wizard/task-item.scss
```diff
@@ -114,4 +114,38 @@
...
+				margin-bottom: 0.5em;
+
+				ul {
+					.payment-method-checkbox {
+						margin-bottom: 1.5em;
+
+						&:last-child {
+							margin-bottom: 0.5em;
+						}
+					}
```
ismaeldcom: It's already defined there, so better to update it if needed instead of adding more definitions https://github.com/Automattic/woocommerce-payments/blob/ef11d7d9d587a3ba3de7e0cfbcd4455f4750f8fc/client/components/payment-methods-checkboxes/payment-method-checkbox.scss#L14-L17

client/components/payment-methods-checkboxes/payment-method-checkbox.scss
```diff
@@ -72,6 +78,14 @@
 		text-transform: initial;
 	}
 
+	&__required-label {
+		display: block;
+		font-size: 11px;
+		line-height: 1em;
+		color: $gray-700;
+		text-transform: initial;
```
ismaeldcom: As it's a child of the parent label `text-transform: initial` is already applied.

ismaeldcom (commented):
Thanks for taking care of it and going an extra mile with a better solution and style fixes. Overall, the solution looks good to me! I left some comments about specific parts of the code, and here I'll add more general comments to make it easier to read 🙂 

I'd like to mention here again, that we need to be careful about causing an issue with `card` disabled permanently without an option to enable it.

We should try to find a balance between pixel perfection and our existing style definitions, from WCPay, WC, and WP. Aiming for a style consistency among all the projects. I mention it, because I think most of the definitions on this block are unnecessary, and cause little deviations between our pages.

https://github.com/Automattic/woocommerce-payments/blob/c4d3390e49359f00ce8875dbf78402500189a3e0/client/additional-methods-setup/wizard/task-item.scss#L118-L150

Using the Figma file as reference some of those `px` sizes don't match. But most importantly, we should avoid using them, and use defined variables like `$gap`. Considering also if we really want to override the default component style to add a tiny pixel difference. I wrote before about it here: pcqRLn-mx-p2

I know it's tempting to fix some style issues on that page, because it as some major differences with the Figma design. But we will need another issue to properly tackle that. I'm not sure if you noticed it, but the wizard is not even responsive.

Also catches my attention how we are duplicating a lot of code between `PaymentMethodCheckbox` and `PaymentMethod` which share logic and style.

client/components/payment-methods-checkboxes/payment-method-checkbox.scss
vladolaru: Good catch.

client/additional-methods-setup/wizard/task-item.scss
vladolaru: I was reluctant to make changes higher up. Will do that.

client/components/payment-methods-list/payment-method.js
vladolaru: Nitpicks are good :) You are right; in the `PaymentMethod` component, it doesn't make any sense. I replaced it with the original `disabled`.

client/components/payment-methods-list/payment-method.js
vladolaru: Thanks for raising these concerns, Ismael. They are valid, and I agree we should try to keep our components' props on a short leash and prevent bloat.

I've thought more about it, and I think in our situation, the `required` and `locked` are valid since they allow us to keep external concerns out of the `PaymentMethod` component.

If we were to embed the logic and get rid of the props, we would have something like this `const required = 'card' === id && isUpeEnabled;`

This would lead us to couple the logic to the global UPE state and make our component less agnostic. I think your example is just at the border of acceptable since it only has stylistic ramifications while being required or locked (especially) has logical implications.

On the upside of adding the extra props, unit testing the behavior of each is much more clear-cut.

Please let me know your thoughts on this, Ismael. Much appreciated.

client/payment-methods/index.js
vladolaru: I agree that allowing the `card` payment method to be disabled is strange (we have a warning pop-up, though). But since I made sure that when UPE is active, the default state is with the `card` active, I don't see a situation where the merchant would get stuck.

When UPE is not active, the merchant can check/uncheck the `card` at will.
When UPE is active, the `card` will always be checked regardless of the previous value.

Am I missing something?

vladolaru:
Thank you for doing such a thorough review, Ismael! Much appreciated!

> I'd like to mention here again, that we need to be careful about causing an issue with card disabled permanently without an option to enable it.

I left a comment to explore this further.

> Using the Figma file as reference some of those px sizes don't match. But most importantly, we should avoid using them, and use defined variables like $gap. Considering also if we really want to override the default component style to add a tiny pixel difference. I wrote before about it here: pcqRLn-mx-p2

I actually don't like working with hard-coded values that are not tied to a design system to ensure consistency across views and time. My approach emerged as **I couldn't find a consistent use of variables and the like** (nor dev guidelines/comments in the codebase). So I told myself to polish this view without looking for pixel-perfect (I didn't try to follow the Figma designs to the T) but to get some good visual hierarchy as intended in the design.

So it was more about doing good for this particular situation (without breaking off from the rest of the UI) while we clean things up and establish better patterns and styling structure, as customers don't really care about our technical debt.

I made changes to make better use of the gap variables and reduce the more fine-grained styling changes.

> I know it's tempting to fix some style issues on that page, because it as some major differences with the Figma design. But we will need another issue to properly tackle that. I'm not sure if you noticed it, but the wizard is not even responsive.

Agree that we need to do further work. From what I see, this might be a pretty extensive issue. Maybe we could do a proper analysis of the current state of things, state our goals and intended outcomes, and split the work into logical issues. If we are not careful, this can quickly turn into a hefty rewrite of large parts of the WCPay client. I am keen to keep this discussion going, apart from this PR.

> Also catches my attention how we are duplicating a lot of code between PaymentMethodCheckbox and PaymentMethod which share logic and style.

I most definitely noticed this from the start :) But properly DRYing these components seemed to require a lot of work and discussions that would further delay the current issue. There seem to be many components in the client that are "in limbo" in terms of their role and relation to other components. I admit that I lack the history to know how things evolved and why they are as they are and where they should really be.
Please advise on how you see the next steps around this.

client/payment-methods/index.js
ismaeldcom: I've checked it again, it works as you described. It was probably an issue on my end while doing some changes to test different scenarios.

So everything good here 🙂

client/components/payment-methods-list/payment-method.js
ismaeldcom: I think `PaymentMethod` component is already highly tied to UPE, not reused anywhere, and contains too many props and external/internal logic for a presentational component. At this point adding more props will not do any harm 🙂 

My point here was about how easy it is with React (in this case) to overengineer components trying to abstract everything with the idea of reusing it later. But sometimes we don't need a reusable component, and it's easier to read and maintain if we design it accordingly.

Anyway, your solution looks good 👍

vladolaru: I agree about the overoptimization/over-abstraction tendency, and it is definitely something to look out for.

client/payment-methods/index.js
vladolaru: Great!

ismaeldcom (approved):
Thanks for the detailed replies and further work on the PR. It works as expected, and now the implementation looks even better and more aligned with the initial design.

We could still improve some parts, but that will be completely out of the scope 🙂 

Ready to merge! 🚀

PR #5301, opened 2022-12-16
Allow UPE methods to be saved during Blocks checkout

### LINKED ISSUE #5260: SEPA payment method not saved in Blocks checkout
### Describe the bug
In the split UPE project, SEPA payment methods are not saved from Blocks checkout. 

### To Reproduce

1. Buy a product using Blocks checkout and SEPA. Save SEPA during the process.
2. Open the Blocks checkout page again and check if you see the SEPA method saved previously. I expect it to appear, but it doesn’t. 🔴 
3. Open the Shortcode checkout page and check if you see the SEPA method saved previously. I expect it to appear, but it doesn’t. 🔴
4. Go to My account -> Payment methods and check if you see the SEPA method on the list. It’s not displayed there, and the list is empty. 🔴 
5. Add a new SEPA payment method in _My account -> Payment methods_ OR by saving the new SEPA method on the shortcode checkout page. After being added, it should appear on the list. Also, the previously added SEPA method starts to appear on the list after this action.

### Actual behavior
After the SEPA payment method is saved while checking out with Blocks, it does not appear in Blocks checkout, shortcode checkout, and _My account -> Payment methods_ page.

### Expected behavior
After the SEPA payment method is saved while checking out with Blocks, it should appear in Blocks checkout, shortcode checkout, and _My account -> Payment methods_ page.

### Additional context
[This](https://github.com/Automattic/woocommerce-payments/pull/5252#issuecomment-1342231008) comment is a good starting point, as it explains possible root causes and places to fix.

mdmoore:
Fixes #5260

This PR ensures that both shortcode and blocks checkouts are taken into account when determining whether the `save_payment_method` URL parameter is set. When saving payment methods from shortcode checkout, to check whether the save payment method checkbox is checked, we can use a variable name in a format similar to `wc-woocomerce_payments_sepa_debit-new-payment-method`, however in a blocks checkout, the gateway ID isn't present, leaving a variable in the format of `wc-sepa_debit-new-payment-method` in `$_POST` instead. 

#### Testing instructions

1. Remove all saved payment methods from your site.
2. From Blocks checkout, pay using SEPA and check the _Save payment information to my account for future purchases._ checkbox to save the payment method.
3. Visit **My Account > Payment Methods** and double check that the SEPA method that was just added appears on the page.
4. Repeat the above from shortcode checkout.


includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -445,7 +445,7 @@ public function process_payment( $order_id ) {
 		$payment_needed            = 0 < $converted_amount;
 		$selected_upe_payment_type = $this->stripe_id;
 		$payment_type              = $this->is_payment_recurring( $order_id ) ? Payment_Type::RECURRING() : Payment_Type::SINGLE();
-		$save_payment_method       = $payment_type->equals( Payment_Type::RECURRING() ) || ! empty( $_POST[ 'wc-' . self::GATEWAY_ID . '_' . $this->stripe_id . '-new-payment-method' ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$save_payment_method       = $payment_type->equals( Payment_Type::RECURRING() ) || ! empty( $_POST[ 'wc-' . self::GATEWAY_ID . '_' . $this->stripe_id . '-new-payment-method' ] ) || ! empty( $_POST[ 'wc-' . $this->stripe_id . '-new-payment-method' ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
FangedParakeet: This solution worked for me...but I believe this `POST` key is different in the blocks checkout because when we register the blocks payment method [the name parameter](https://github.com/Automattic/woocommerce-payments/blob/fix/5260-blocks-checkout-save-upe-methods/client/checkout/blocks/upe.js#L67) is set to the Stripe ID, rather than the gateway ID. You can see that previously we used to use the gateway ID--prefixed with `woocommerce_payments`--instead in the regular pre-partition UPE. 

I think if we used `upeMethods[ upeName ]` instead of just `upeName` for this name field, that would instead use the same gateway ID for both blocks and shortcode checkouts. This also would have the benefit of differentiating this gateway from other non-WCPay gateways that may also register `sepa_debit` as a name field, for whatever reason. 

Let me know your thoughts, @mdmoore--especially if this might have already been an idea that you contemplated and rejected. Cheers!

mdmoore: >I think if we used upeMethods[ upeName ] instead of just upeName for this name field, that would instead use the same gateway ID for both blocks and shortcode checkouts.

Thanks for examining this more closely even though the solution worked. This is definitely the way to go. I did indeed consider that we should be using the same gateway ID for both blocks and shortcode checkouts but didn't understand how blocks payment methods were getting their IDs. I now see that it's simply the `name` set in `registerPaymentMethod()`, which is fantastic! And, because @timur27 already took care of this in #5344, SEPA payment methods can now be saved when completing a blocks checkout on this PR.

**Side note:** After merging in the latest `poc/upe-instances-multiplied`, which included the gateway ID updates, and building the JS, the frontend was still using `sepa_debit` as the ID rather than `woocommerce_payments_sepa_debit`. It wasn't until I toyed around with the payment method name in `registerPaymentMethod()` that correct ID began showing.

mdmoore:
@timur27 Thanks for offering to review this one! With your changes in [#5344](https://github.com/Automattic/woocommerce-payments/pull/5344), the only thing I've done here is remove a check that's no longer valid. For context, it's the same one @FangedParakeet mentioned [here](https://github.com/Automattic/woocommerce-payments/pull/5282#discussion_r1051047165).

timur27 (approved):
Thanks for these changes @mdmoore, and sorry if I interrupted your work with my changes in the meantime - this was unintentional, as I worked on fixing the payment with an already saved SEPA. 😺 

I am now able to save SEPA payment method after you got rid of the if statement which blocked the token from being saved even with the `$save_payment_method` flag being `true`.

I tried to be a bit nitpicky here and look for the potential space where we could add a test that will help us not to break the logic in the future. But I found [this test](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/payment-methods/test-class-upe-payment-gateway.php#L1096-L1149), which tests everything that is needed. You probably will ask

> since it checks everything that's needed, why wasn't it failing when the token wasn't saved during the process?

Well, [this test uses payment method with _**card**_ id](https://github.com/Automattic/woocommerce-payments/blob/9e83a593fd1aa0d086ff24ce34ef6bc79c3bd1d1/tests/unit/payment-methods/test-class-upe-payment-gateway.php#L1077) (left side of the `if` you removed) and `$payment_method_type`(right side of the `if`) [is taken from the charge and also has a value equal to card](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/helpers/class-wc-helper-intention.php#L30). Effectively, the if statement that blocked us from saving on the localhost, was passing in the test suite. One might think that it's worth adding the identical test but with `SEPA` this time and asserting that the token is saved. But I wouldn't go this direction as we removed this logic, and such a test will look redundant.

To sum this up, after your changes and this quick research on the tests described above, I feel fully comfortable and safe with giving this PR the green light. 🚢  Thanks for working on this!

PR #5306, opened 2022-12-19
Let code freeze workflow pass botwoo token secret to release-pr workflow

### LINKED ISSUE #5305: Release code freeze workflow failure
### Describe the bug

"Create a PR to trunk" step fails with the following message: `gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable.`
It's not able to raise a PR using gh cli. 

Example failed run: https://github.com/Automattic/woocommerce-payments/actions/runs/3724726397

FYI, we recently switched from the GITHUB_TOKEN secret to the BOTWOO_TOKEN secret.

jbordonado:
Fixes #5305 

- Pass the BOTWOO_TOKEN secret from the code freeze workflow to the release-pr workflow

GitHub docs for reference:
- https://docs.github.com/en/actions/using-workflows/reusing-workflows#passing-secrets-to-nested-workflows
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callsecrets

#### Testing instructions

Hard to test the code freeze workflow without temporarily altering it, because the first job `check-code-freeze` will always output that it is not the time of freeze for now.

- I altered the workflow temporarily on a branch to always call the nested workflow with the release version 7.2.6
- It successfully created a PR with `botwoo` as the author; see the run: https://github.com/Automattic/woocommerce-payments/actions/runs/3729740012

I also tested the nested workflow again directly to ensure it still works:
- it created a PR for release 7.2.7 successfully; see the run: https://github.com/Automattic/woocommerce-payments/actions/runs/3729820808

-------------------

**Post merge**

### DIFF
--- a/.github/workflows/release-code-freeze.yml
+++ b/.github/workflows/release-code-freeze.yml
@@ -75,8 +75,7 @@ jobs:
     uses: ./.github/workflows/release-pr.yml
     with:
       releaseVersion: ${{ needs.check-code-freeze.outputs.nextReleaseVersion }}
-    secrets:
-      BOTWOO_TOKEN: ${{ secrets.BOTWOO_TOKEN }}
+    secrets: inherit
 
   slack-notification:
     name: "Send notification to Slack"
--- a/.github/workflows/release-code-freeze.yml
+++ b/.github/workflows/release-code-freeze.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       skipSlackPing:
-        description: "If true, the Slack ping will be skipped"
+        description: "Skip the Slack ping"
         type: boolean
         required: false
         default: false
@@ -75,6 +75,8 @@ jobs:
     uses: ./.github/workflows/release-pr.yml
     with:
       releaseVersion: ${{ needs.check-code-freeze.outputs.nextReleaseVersion }}
+    secrets:
+      BOTWOO_TOKEN: ${{ secrets.BOTWOO_TOKEN }}
 
   slack-notification:
     name: "Send notification to Slack"
--- a/changelog/fix-release-code-freeze-workflow-secret
+++ b/changelog/fix-release-code-freeze-workflow-secret
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+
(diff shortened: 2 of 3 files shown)


.github/workflows/release-code-freeze.yml
```diff
@@ -75,6 +75,8 @@ jobs:
     uses: ./.github/workflows/release-pr.yml
     with:
       releaseVersion: ${{ needs.check-code-freeze.outputs.nextReleaseVersion }}
+    secrets:
+      BOTWOO_TOKEN: ${{ secrets.BOTWOO_TOKEN }}
```
deepakpathania: Would we have any unintended side effects if we just `inherit` all secrets to the nested workflow? It'd be handy otherwise to avoid adding these individually for future scenarios we might miss.

changelog/fix-release-code-freeze-workflow-secret
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+
```
deepakpathania, line 5: Could we add something in the changelog for reference?

deepakpathania (commented):
Left a small clarifying comment.

.github/workflows/release-code-freeze.yml
jbordonado: I wouldn't expect side effects but I think the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) is always good to follow.
That said, I don't have a particularly strong opinion on this for this particular case

changelog/fix-release-code-freeze-workflow-secret
jbordonado: Ah. I meant to add a non-significant entry (so it doesn't show up in the actual changelog generated), but forgot to put any text. Fixed

.github/workflows/release-code-freeze.yml
deepakpathania: I agree in general, but in this case, it's easy to miss adding a new secret here, just like we missed, and then could be tricky to debug. 

So we should either add a comment here mentioning that one should add an entry here for it to be inherited in nested workflows, or `inherit` all to avoid the same. I'll let you decide which way to go. 😄

jbordonado: Made the nested workflow inherit all

Test run: https://github.com/Automattic/woocommerce-payments/actions/runs/3730484568

deepakpathania (approved):
Changes LGTM now. 🚢

PR #5309, opened 2022-12-19
Retry API requests on network failures.

### LINKED ISSUE #5189: Fix Duplicate Charges: Retry server(API) calls with idempotency headers
### Acceptance criteria

Add `try/catch` mechanisms to crucial server calls (ex. creating intentions), and use the same idempotency header for all of those calls within the same PHP request.

We should not try to use the same header within separate PHP requests (calls to the client), this is already explained in the results in this comment paJDYF-6wC-p2#comment-15189

### Testing instructions

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'
4. Observe expected behaviour

### Additional context

paJDYF-6wC-p2#comment-15189

dmvrtx:
Request is considered failed due to network reasons if response code is not set.

Fixes #5189 

API supports `Idempotency-Key` header with which POST request can be repeated and should not result in creation or repeated updated, but return last successful response instead.

API client already sets `Idempotency-Key` to a unique value for `POST` requests and for other requests it can be added with a `wcpay_api_request_headers` filter.

This PR creates a retry loop in API client which is broken on any of the following occurrences:

 - Response code is present.
 - Request doesn't have an `Idempotency-Key` header.
 - Reached number of possible retries, by default 3.
 - Request processing time is beyond the provided timeout.

#### Testing instructions

 * Covered by unit tests.

-------------------

**Post merge**

### DIFF
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -2116,14 +2116,6 @@ class WC_Payments_API_Client {
 				$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
 				$response_code = wp_remote_retrieve_response_code( $response );
 			} catch ( Connection_Exception $e ) {
-				/**
-				 * Error code 503 indicates there was no response, while 500
-				 * means there was a response and it was an error. We only want
-				 * to retry connection errors here.
-				 */
-				if ( 503 !== $e->get_http_code() ) {
-					throw $e;
-				}
 				$last_exception = $e;
 			}
 
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -31,6 +31,7 @@ class WC_Payments_API_Client {
 	const DELETE = 'DELETE';
 
 	const API_TIMEOUT_SECONDS = 70;
+	const API_RETRIES_LIMIT   = 3;
 
 	const ACCOUNTS_API                 = 'accounts';
 	const CAPABILITIES_API             = 'accounts/capabilities';
@@ -2088,20 +2089,34 @@ class WC_Payments_API_Client {
 			);
 		}
 
-		$response = $this->http_client->remote_request(
-			[
-				'url'             => $url,
-				'method'          => $method,
-				'headers'         => apply_filters( 'wcpay_api_request_headers', $headers ),
-				'timeout'         => self::API_TIMEOUT_SECONDS,
-				'connect_timeout' => self::API_TIMEOUT_SECONDS,
-			],
-			$body,
-			$is_site_specific,
-			$use_user_token
-		);
+		$headers        = apply_filters( 'wcpay_api_request_headers', $headers );
+		$stop_trying_at = time() + self::API_TIMEOUT_SECONDS;
+		$retries        = 0;
+		$retries_limit  = array_key_exists( 'Idempotency-Key', $headers ) ? self::API_RETRIES_LIMIT : 0;
 
-		$response = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+		while ( true ) {
+			$response = $this->http_client->remote_request(
+				[
+					'url'             => $url,
+					'method'          => $method,
+					'headers'         => $headers,
+					'timeout'         => self::API_TIMEOUT_SECONDS,
+					'connect_timeout' => self::API_TIMEOUT_SECONDS,
+				],
+				$body,
+				$is_site_specific,
+				$use_user_token
+			);
+
+			$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+			$response_code = wp_remote_retrieve_response_code( $response );
+
+			if ( $response_code || time() >= $stop_trying_at || $retries_limit === $retries ) {
+				break;
+			}
+
+			$retries++;
+		}
 
 		$this->check_response_for_errors( $response );
 
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -9,6 +9,7 @@ defined( 'ABSPATH' ) || exit;
 
 use WCPay\Exceptions\API_Exception;
 use WCPay\Exceptions\Amount_Too_Small_Exception;
+use WCPay\Exceptions\Connection_Exception;
 use WCPay\Fraud_Prevention\Fraud_Prevention_Service;
 use WCPay\Fraud_Prevention\Buyer_Fingerprinting_Service;
 use WCPay\Logger;
@@ -2095,23 +2096,41 @@ class WC_Payments_API_Client {
 		$retries_limit  = array_key_exists( 'Idempotency-Key', $headers ) ? self::API_RETRIES_LIMIT : 0;
 
 		while ( true ) {
-			$response = $this->http_client->remote_request(
-				[
-					'url'             => $url,
-					'method'          => $method,
-					'headers'         => $headers,
-					'timeout'         => self::API_TIMEOUT_SECONDS,
-					'connect_timeout' => self::API_TIMEOUT_SECONDS,
-				],
-				$body,
-				$is_site_specific,
-				$use_user_token
-			);
+			$response_code  = null;
+			$last_exception = null;
 
-			$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
-			$response_code = wp_remote_retrieve_response_code( $response );
+			try {
+				$response = $this->http_client->remote_request(
+					[
+						'url'             => $url,
+						'method'          => $method,
+						'headers'         => $headers,
+						'timeout'         => self::API_TIMEOUT_SECONDS,
+						'connect_timeout' => self::API_TIMEOUT_SECONDS,
+					],
+					$body,
+					$is_site_specific,
+					$use_user_token
+				);
+
+				$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+				$response_code = wp_remote_retrieve_response_code( $response );
+			} catch ( Connection_Exception $e ) {
+				/**
+				 * Error code 503 indicates there was no response, while 500
+				 * means there was a response and it was an error. We only want
+				 * to retry connection errors here.
+				 */
+				if ( 503 !== $e->get_http_code() ) {
+					throw $e;
+				}
+				$last_exception = $e;
+			}
 
 			if ( $response_code || time() >= $stop_trying_at || $retries_limit === $retries ) {
+				if ( null !== $last_exception ) {
+					throw $last_exception;
+				}
 				break;
 			}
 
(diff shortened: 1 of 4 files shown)


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2088,20 +2089,34 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
...
+				],
+				$body,
+				$is_site_specific,
+				$use_user_token
+			);
+
+			$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+			$response_code = wp_remote_retrieve_response_code( $response );
+
+			if ( $response_code || time() >= $stop_trying_at || $retries_limit === $retries ) {
```
kalessil, line 2123: This approach is not refactoring-friendly, better to move `time() >= $stop_trying_at || $retries_limit === $retries` into while condition (with necessary tweaks) - fewer risks of infinity loops after inaccurate changes of the body.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2088,20 +2089,34 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
...
-				'headers'         => apply_filters( 'wcpay_api_request_headers', $headers ),
-				'timeout'         => self::API_TIMEOUT_SECONDS,
-				'connect_timeout' => self::API_TIMEOUT_SECONDS,
-			],
-			$body,
-			$is_site_specific,
-			$use_user_token
-		);
+		$headers        = apply_filters( 'wcpay_api_request_headers', $headers );
+		$stop_trying_at = time() + self::API_TIMEOUT_SECONDS;
```
kalessil, line 2095: Are there any parts of the request pointing out its time?

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: It's covered by unit tests, so infinite loop shouldn't slip through. Splitting condition is making it worse in my opinion, as there will be two places at which loop can end. `while ( true )` is a good indicator that end condition is _inside the loop_.

kalessil: Fair enough =) Marking as resolved!

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: I'm not sure I understand the question.

kalessil: `time()` returns a timestamp, right?

With testing in mind (we got used to using filters for injecting time in tests), I wonder if we have to use `time()` call here or perhaps there is a timestamp already in the method scope?

dmvrtx: I haven't seen anything related in this method scope and on this stack level. There are timestamps down the stack. However, moving code there is not desirable, because it will be disconnected from the context of WCPay API client where idempotent requests make sense.

kalessil: Gotcha, marking as resolved.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2088,20 +2089,34 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
...
+					'timeout'         => self::API_TIMEOUT_SECONDS,
+					'connect_timeout' => self::API_TIMEOUT_SECONDS,
+				],
+				$body,
+				$is_site_specific,
+				$use_user_token
+			);
+
+			$response      = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+			$response_code = wp_remote_retrieve_response_code( $response );
```
htdat: It seems that relying on this is not accurate here. Looking futhrer downstream of `WC_Payments_Http::remote_request`, I notice that we throw `Connection_Exception` in  `WC_Payments_Http::make_request` for network issues. 

I think we should try/catch this exception here. Otherwise, the execution can not reach this code to allow retries. 

I tested this by: 

- Enable both client and server. 
- Client: Add products into cart, visit checkout page. 
- Server: Disable the server. 
- Client: Try to make a payment on the checkout page. 
- Client: check the WCPayment log file for today under `wp-content/uploads/wc-logs/` or ` docker/wordpress/wp-content/uploads/wc-logs/` for Docker. 
- There is only one log entry `ERROR HTTP_REQUEST_ERROR` - [code ref](https://github.com/Automattic/woocommerce-payments/blob/485926e28a9618470a9ab3f4ebc97f520c15be3b/includes/wc-payment-api/class-wc-payments-http.php#L111-L111)

htdat: Adding a try/catch seems to work well. With steps above, there are 4 log entries `ERROR HTTP_REQUEST_ERROR`. Here is my attempt https://gist.github.com/htdat/a50eb9310ae8cd8e5a01e9be27ba3367

dmvrtx: Thanks for checking this, @htdat! I'll add a try/catch loop and additional tests.

However, my concern is that `Connection_Exception` can be thrown in other cases too, where response code is present. It'd be prudent to not retry if response contains an error, but it adds complexity to checks.

What's worth trying is to modify `make_request` to return different HTTP code for empty response, e.g. 503. This will allow to distinguish errors in the retry loop.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2088,20 +2089,34 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
...
 
-		$response = apply_filters( 'wcpay_api_request_response', $response, $method, $url, $api );
+		while ( true ) {
+			$response = $this->http_client->remote_request(
+				[
+					'url'             => $url,
+					'method'          => $method,
+					'headers'         => $headers,
+					'timeout'         => self::API_TIMEOUT_SECONDS,
+					'connect_timeout' => self::API_TIMEOUT_SECONDS,
```
jbordonado: In relation to the server issue https://github.com/Automattic/woocommerce-payments-server/issues/2682, should we decrease the `connect_timeout` to 10s (the default value in WP) similarly?

dmvrtx: Same timeout value is applied to requests within admin interface, e.g. retrieving a list of transactions. Such requests can take longer time to be processed by server. However, we can apply different timeouts depending on which method is called in API client class, but I'd do it as a separate change.

jbordonado: Just noting that `connect_timeout` is about the connection phase to the server, not the time it takes for the server to process the request.

Quote from the [CURL_CONNECTTIMEOUT doc](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html):
> This only limits the connection phase, it has no impact once it has connected

According to Systems (details can be found in the linked issue), it should take a few milliseconds to connect if everything is healthy, so 70s seems excessive.

The regular `timeout` indeed makes more sense to be greater and it can be complicated to adjust it with relevant values depending on the API targeted, so I was not referring to that.

That can be a separate change, though, as you said 👍 you didn't alter this part

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: Addressed in fb33174b

htdat: > What's worth trying is to modify make_request to return different HTTP code for empty response, e.g. 503. This will allow to distinguish errors in the retry loop.

@dmvrtx - At first, I thought it's a good idea. However, when looking around the code you've made for `make_request` in https://github.com/Automattic/woocommerce-payments/commit/fb33174b81bcb8c3190f3b5c09d04e30bac811dc, it seems that `$exception_http_code` is always `503`. The reason is that if `is_array( $response ) === true`, the code will not execute the `if` block in this method, hence no exception will be thrown. 

I looked further to `Automattic\Jetpack\Connection\Client::remote_request` and `Automattic\Jetpack\Connection\Client::_wp_remote_request`, then finally `wp_remote_request`. If I understand correctly, it's really a network issue if these functions return a WP_Error response, i.e. a `Connection_Exception` throw in `make_request`. 

In other words, I think we can simplify the change by catching `Connection_Expection` only, which is enough to cover network issues. WDYT?

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: > This only limits the connection phase, it has no impact once it has connected

This makes sense, and I agree that it should be a separate issue/PR.

includes/wc-payment-api/class-wc-payments-api-client.php
dmvrtx: I agree, that the check I have for `$exception_http_code` looks fragile. However, the problem is that any filter can return `WP_Error` as well, even after request was successful, reducing the reliability of this check.

I really don't like the idea of sending sensitive payment data through the code stack with so many injection points allowing any plugin to read and even change information being sent. I agree to have a check on `is_wp_error()` alone, if we agree to implement a  separate `request` method for API calls which wouldn't allow for any interventions other than specified by the caller.

dmvrtx: Chanaged as @htdat proposed in 44fbe8cb

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
```diff
@@ -2146,6 +2147,154 @@ public function test_authorizations_summary_success() {
...
+	 * possible to retry.
+	 *
+	 * @throws Exception in case of the test failure.
+	 */
+	public function test_request_retries_post_on_network_failure_exception() {
+		$this->mock_http_client
+			->expects( $this->exactly( 4 ) )
+			->method( 'remote_request' )
+			->willThrowException(
+				new Connection_Exception( 'HTTP request failed', 'wcpay_http_request_failed', 503 )
```
htdat: Nitpick: I think we might want to use `500` for `http_code` to maintain the consistency.

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: I agree that at any point, other plugins/code can add a filter to return `WP_Error` and it can be annoying for a payment request. This is a nature of WordPress with many hooks like that. Someone can argue it's good for extensibility but it also creates a lot of headache if someone abuses it. 

Even we have a separate `request` method but still use `wp_remote_request`, which has filter `http_response` to modify any response, someone can still filter and change the `response`. That said, we might need to live with it at the moment and may move this discussion to a broader audience.

> https://github.com/Automattic/woocommerce-payments/commit/44fbe8cbf930b620195eff5663d15e3d986c9fed

LGTM! Though I have some minor comments regarding tests.

htdat (commented):
@dmvrtx - thanks for addressing the issue I raised. Code looks good now. Tests are great too. My only suggestion is to add another test case that one of the retries is successful. 

E.g. original request -> failed, first retry -> failed, second retry -> successful.

tests/unit/wc-payment-api/test-class-wc-payments-api-client.php
dmvrtx: Done in 78b35e56

dmvrtx:
> E.g. original request -> failed, first retry -> failed, second retry -> successful.

Done in 73fc1165

htdat (approved):
LGTM! Thank you for doing the awesome work here.

Issue #5261, opened 2022-12-08
The attempt to pay with a saved SEPA payment method in Blocks checkout fails


timur27:
### Describe the bug
In the split UPE project, the attempt to pay with a saved SEPA payment method in Blocks checkout fails. 

### To Reproduce
1. Save a new SEPA payment method in _My account -> Payment methods_
2. Add a product to the cart 
3. Open the Blocks checkout page and select the saved SEPA payment method 
4. Place an order
5. You will see `No payment method provided.` error notice on the top of the page.

The `/checkout` endpoint response is
```
{
  "code": "woocommerce_rest_checkout_missing_payment_method",
  "message": "No payment method provided.",
  "data": {
    "status": 400
  }
}
```

### Actual behavior
Users cannot pay with saved SEPA payment methods while using Blocks checkout.

### Expected behavior
Users should be able to pay with saved SEPA payment methods while using Blocks checkout.

### Additional context
This was encountered [in another PR](https://github.com/Automattic/woocommerce-payments/pull/5252#pullrequestreview-1207458645).

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5261?planning-poker) with Zenhub @mdmoore @timur27

PR #5344, opened 2022-12-28
Use the correct key for blocks payment method registration

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5261

There are two types of payment method ID in our system. I'll use SEPA for demonstration purposes. 
1. `sepa_debit`
2. `woocommerce_payments_sepa_debit` 

The first one is the Stripe payment method type ID. It should be used when e.g., we create a payment intent. The second one is the internal WCPay payment method ID.

The problem this PR fixes is that we used Stripe payment method type ID (`sepa_debit`) in the places where the WCPay payment method ID (`woocommerce_payments_sepa_debit`) was needed. Effectively, [the payment method was not found](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/assets/js/base/context/providers/cart-checkout/checkout-processor.ts#L116-L122) by the WooCommerce Blocks implementation and we were seeing the `No payment method provided ` error on the UI.

#### Testing instructions
1. Pay with a saved SEPA payment method in Blocks checkout and confirm there are no errors 
2. Pay with a new SEPA payment method in Blocks checkout and confirm there are no errors

-------------------

**Post merge**

### DIFF
--- a/client/checkout/blocks/upe-fields.js
+++ b/client/checkout/blocks/upe-fields.js
@@ -72,6 +72,7 @@ const WCPayUPEFields = ( {
 	},
 	emitResponse,
 	paymentMethodId,
+	upeMethods,
 	paymentIntentId,
 	paymentIntentSecret,
 	errorMessage,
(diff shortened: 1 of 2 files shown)


FangedParakeet (approved):
As promised, with these brand new edits, I found myself completely capable of checking out with both a saved SEPA payment method and a new SEPA payment method. I wasn't able to save a payment method while checking out, but it appears that will be handled in #5720.

Curiously enough, I actually suggested the approach in this PR over [here](https://github.com/Automattic/woocommerce-payments/pull/5301#discussion_r1052767796) in the respective PR to resolve #5720; however, I was unable to checkout and save a new SEPA payment method simultaneously, while using the amends within this PR. Just paging @mdmoore, since I wonder if these changes will affect his proposed resolution to the latter issue in #5301.

Anyways, for now, this appears to be working for me, so onwards we march! :shipit:

PR #5346, opened 2022-12-29
Check intent attached to order before processing payment

### LINKED ISSUE #5187: Fix Duplicate Charges: Store intent IDs whenever possible, check status
### Description

We need to associate the intent ID as early as possible to the order, meaning right before/after each `(create|create_and_confirm|update)_(setup_)intention` call.

### Acceptance criteria

1. Storing the intent early will help us to check the status of the intent _and_ the order in every call, related to payment processing, including:

*   `process_payment` of the main gateway
*   `create_payment_intent`, `process_payment`, and `process_redirect_payment` of the UPE gateway.

2. Within those calls, before creating a new intent, or performing any payment-related calls, we should:

- Check if the order is marked as paid -> redirect to the order received page.
- Check if the order has an intent ID -> retrieve the intent, and if successful, redirect to the order received page. If not, try to use the same intent for the rest of the process.

### Dev notes
A couple of things to keep in mind here:

*   Orders store `cart_hash`, meaning that we can compare the contents of the order with the current cart as a confirmation step. [See this example](https://href.li/?https://github.com/Automattic/woocommerce-payments/pull/5163/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R910)
*   While redirecting to the order received page, we need to add a (GET) parameter to the URL, and use it to show a message that we’ve used previously entered details (the customer has already committed to those details, so it should be fine). That’s because in the interim, they could have changed something in their address, their payment method, etc., but we are likely still using the old ones.

**UPE** **specifics**

Within this task, we should also make sure that the UPE gateway does not create a new intent when loading, using the one from the order instead. For this we need to do a few things:

1.  Check the order (if available) for an existing intent, and use it, instead of creating a new one. UPE has the necessary logic to update the intent within `process_payment`, so that’s fine.
2.  Consider removing calls to `remove_upe_payment_intent_from_session` in various places, especially [when handling failures](https://href.li/?https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L704). Check with Pulsar (@anu-rock): did you do this as a fail-safe, or were there particular issues you were facing?
3.  The WC session object might be unreliable, as my attempts proved it to be inconsistent. We should investigate when the session it stored (wherever it is stored), and when it is cleared.

### Additional context

paJDYF-6wC-p2#comment-15189

htdat:
Fixes #5187

- Check the status of attached intent during the payment processing.
- Mark the order successful if the intent is one of successful statuses, and redirect customers to the order-received of this order.

### Prep 

- Run `npm run watch` to build JS files.
- Feel free to test in multiple flows and different type cards: 
    - UPE and non-UPE flows.
    - classic and block checkout pages.
    - basic and 3DS cards. 
- Add this snippet to your site https://gist.github.com/htdat/6a52eec1be77b8442ac05401908a3627. You can add it with plugin `Code Snippets` or directly to the bottom of file `woocommerce-payments.php` in the root directory of this project.
- For basic (non-3DS) card in the non-UPE flow, we need to inject this diff to `woocommerce-payments.php` https://gist.github.com/htdat/09eef1ed0cbeec50a5354ca8da38d623 

### Start 

- Add some products, visit to the checkout page.
- Try to make an payment in the checkout page.
- See an error due to our snippets (trying to replicate potential issues that real customers can get).
- See this order (usually the newest order) in WP Admin > Order, confirm that it has not been paid.
- Try to make another payment again.
- Get redirected to `order-received` page of this order with this parameter `wcpay_previous_successful_intent=yes`, e.g.: http://localhost:8082/checkout/order-received/zyz/?key=wc_order_1b9oTbcXtWH0l&wcpay_previous_successful_intent=yes

Screenshot:
Markup on 2023-01-06 at 17:35:02

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : [link](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.3.0#check-intent-attached-to-order-before-processing-payment-pr-5346)


changelog/add-check-intent-attached-to-order-before-payment
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: update
+
+Check the status of attached intent during the payment processing.
```
RadoslavGeorgiev: This patch is pretty significant, I think making it a bit longer and non-technical could be benefitial.

How about something in the spirit of "Check the status of previously initiated payments and mark orders as processing instead of initiating a new payment".

htdat:
FYI. I am still looking into this [E2E test failure](https://github.com/Automattic/woocommerce-payments/actions/runs/3827260909/jobs/6511697984), which seems consistent: 

> FAIL tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-manage-payments.spec.js (125.827s)
>   Shopper > Subscriptions > Manage Payment Methods
>     ✓ should be able to purchase a subscription (13312ms)
>     ✓ should change a default payment method to a new one (8541ms)
>     ✕ should set a payment method to an already saved card (100007ms)

## Update

This is [debug.log](https://github.com/Automattic/woocommerce-payments/files/10338350/debug.log) from the E2E run above. 

> [03-Jan-2023 06:10:49 UTC] PHP Fatal error:  Uncaught Exception: Error: No route was found matching the URL and request method. in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php:797
> Stack trace:
> #0 /var/www/html/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-change-payment-gateway.php(334): WC_Payment_Gateway_WCPay->process_payment(93)

This error happens when a customer is trying to change a subscription (not order, its post type is `shop_subscription` in wp_posts) from the current payment method to another existing payment. 

With a subscription, we're saving a setup intent (starting with `seti_`) as order meta  `_intent_id`. Without checking this fact, we will send a GET request for a setup intention to the payment intention endpoint. 
https://github.com/Automattic/woocommerce-payments/blob/b5566a79245c48cb55324ec7f4279615f9ddd59e/includes/class-wc-payment-gateway-wcpay.php#L1951-L1951

I fixed it in this commit by checking intent ID prefix before processing further. https://github.com/Automattic/woocommerce-payments/pull/5346/commits/f3f7b5ef6f7655b82b980f8de70c5cd301258eeb

client/checkout/api/index.js
```diff
@@ -695,14 +695,19 @@ export default class WCPayAPI {
...
 	 * @return {boolean} Returns true if doing the redirection.
 	 */
-	handlePreviousOrderPaid( response ) {
+	handleDuplicatePayments( response ) {
 		let didRedirection = false;
-		if ( response.wcpay_upe_paid_for_previous_order && response.redirect ) {
+		const hasFlag =
+			response.wcpay_upe_paid_for_previous_order ||
+			response.wcpay_upe_previous_successful_intent;
+		if ( hasFlag && response.redirect ) {
```
RadoslavGeorgiev: I see how this works, and it makes sense, but I would maybe try making it a bit more readable:

```js
/**
 * Redirect to the order-received page for duplicate payments:
 *
 * @param {Object} response Response data to check if doing the redirect.
 * @return {boolean} Returns true if doing the redirection.
 */
handleDuplicatePayments( {
	wcpay_upe_paid_for_previous_order: previouslyPaid,
	wcpay_upe_previous_successful_intent: previousSuccessfulIntent,
	redirect,
} ) {
	if ( redirect ) {
		// Another order has the same cart content and was paid.
		if ( previouslyPaid ) {
			return ( window.location = redirect );
		}

		// Another intent has the equivalent successful status for the order.
		if ( previousSuccessfulIntent ) {
			return ( window.location = redirect );
		}
	}

	return false;
}
```

The change is mostly to bring comments closer to the code, I'm not insisting on it.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1919,6 +1929,51 @@ protected function get_account_branding_icon( $default_value = '' ): string {
...
+		}
+
+		// We only care about payment intent.
+		$is_payment_intent = 'pi_' === substr( $intent_id, 0, 3 );
+		if ( ! $is_payment_intent ) {
+			return;
+		}
+
+		$intent        = $this->payments_api_client->get_intent( $intent_id );
+		$intent_status = $intent->get_status();
```
RadoslavGeorgiev: An exception might occur during the `$api_client->get_intent()` call or if the response was erroneous, while trying to get the intent status. Right now if either happened, the whole payment process would fail, even though the exception would be caught, and the handling will be graceful.

Nevertheless, I'm curious if it makes sense to handle the exception here by just logging it, and returning `null`.

RadoslavGeorgiev (changes requested):
Code goods very well overall, a couple of nitpicks, but the approach seems bulletproof 🥳 

Just curious, are you planning to move the saving/storage of intents closer to API calls in a different PR?

changelog/add-check-intent-attached-to-order-before-payment
htdat: Makese sense, updated [a57a775](https://github.com/Automattic/woocommerce-payments/pull/5346/commits/a57a775cde59fec83353fff3b59703009f8222d7)

client/checkout/api/index.js
htdat: Great suggestion! Updated https://github.com/Automattic/woocommerce-payments/pull/5346/commits/290c0d2660e552291d61cc7a5963192fb0208baa

includes/class-wc-payment-gateway-wcpay.php
htdat: > Nevertheless, I'm curious if it makes sense to handle the exception here by just logging it, and returning null.

Agreed with this. Updated https://github.com/Automattic/woocommerce-payments/pull/5346/commits/cf3918ee6a76328db4f3322889c68357535c516f

htdat:
Thanks for the awesome feedback! 

@RadoslavGeorgiev - I've addressed your comments, and accepted all of them. This PR is ready for review again. 

Regarding: 

> Just curious, are you planning to move the saving/storage of intents closer to API calls in a different PR?

Good question but I would have mentioned earlier. In UPE gateway, `attach_intent_info_to_order` in [process_payment](https://github.com/Automattic/woocommerce-payments/blob/cf3918ee6a76328db4f3322889c68357535c516f/includes/payment-methods/class-upe-payment-gateway.php#L531) and [process_redirect_payment](https://github.com/Automattic/woocommerce-payments/blob/cf3918ee6a76328db4f3322889c68357535c516f/includes/payment-methods/class-upe-payment-gateway.php#L705) has already been close to the API calls 

For the main gateway, in this attach method is also already close to the API call in [update_order_status](https://github.com/Automattic/woocommerce-payments/blob/cf3918ee6a76328db4f3322889c68357535c516f/includes/class-wc-payment-gateway-wcpay.php#L2623-L2624) already. 

The only remaining attach method in [process_payment_for_order](https://github.com/Automattic/woocommerce-payments/blob/cf3918ee6a76328db4f3322889c68357535c516f/includes/class-wc-payment-gateway-wcpay.php#L1232) of the main gateway can be moved earlier, right after [create_and_confirm_intention](https://github.com/Automattic/woocommerce-payments/blob/cf3918ee6a76328db4f3322889c68357535c516f/includes/class-wc-payment-gateway-wcpay.php#L1065-L1090). However, reviewing all the complex code between the current position and the potential earlier position, I think it's acceptable to keep it in the current position. Otherwise, there are some pitfalls: 

- In many cases, we still need to call this attach method in the same position, then that's double database update calls.
- The code block between this method and two other methods `update_order_status_from_intent` and `attach_exchange_info_to_order` will be separated while they would be close to each other.

htdat:
> E2E Tests - Pull Request / WC - latest | blocks - shopper (pull_request) - [Details](https://github.com/Automattic/woocommerce-payments/actions/runs/3837591347/jobs/6533937785)

This E2E test failure is similar to a recent scheduled E2E run for the `develop` branch - [E2E Tests - All](https://github.com/Automattic/woocommerce-payments/actions/runs/3837809250).

htdat:
> E2E Tests - Pull Request / WC - latest | blocks - shopper (pull_request) - [Details](https://github.com/Automattic/woocommerce-payments/actions/runs/3837591347/jobs/6533937785)

This E2E test failure is similar to a recent scheduled E2E run for the `develop` branch - [E2E Tests - All](https://github.com/Automattic/woocommerce-payments/actions/runs/3837809250).

_Update_: Discussion regarding this: p1672840161807159-slack-C04D1QU76PK

RadoslavGeorgiev (approved):
The code looks and works great 🥳 🚢 

I've tested:
- UPE & non-UPE
- Blocks vs classic
- 3DS vs non-3DS. When testing with a 3DS card, I let it fail and then switched to 4242 to make sure the original payment method remains.

> In many cases, we still need to call this attach method in the same position, then that's double database update calls.

I think it's still worth simply storing the intent ID as early as possible. `$order->update_meta_data()` does not trigger a database update call by itself, that only happens when `$order->save()` or `$order->save_meta_data()` is called. Those would be called before marking the order as failed anyway, so adding a single update method call after the API call would make a lot of sense without any practical overhead. I'll leave that up to you though, the PR is approved anyway :)

PR #5354, opened 2023-01-02
Retry server redirection responses

### LINKED ISSUE #5115: Multiple UPE - Handle WCPay Server API Errors when fetching payment methods
### Describe the bug
An intermittent warning is occurring when fetching payment methods, such as on the **My Account > Payment Methods** or **Checkout** pages, or anywhere that saved payment methods might be displayed. When this happens, no payment methods are listed and the following is dumped on the page(along with a couple other warnings that cascade from this one):

`Warning: Illegal string offset 'data' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-customer-service.php on line 223`

This appears to be happening when the response coming back from the WCPay Server is returning an empty string rather than an array of payment methods. The request for payment methods originates [here in the token service](https://github.com/Automattic/woocommerce-payments/blob/338f2896bae6834c30bea614d7a5f2b21c7e2c4f/includes/class-wc-payments-token-service.php#L140) and ultimately fires a request to WCPay Server [here in the API client](https://github.com/Automattic/woocommerce-payments/blob/338f2896bae6834c30bea614d7a5f2b21c7e2c4f/includes/wc-payment-api/class-wc-payments-api-client.php#L1592).

### To Reproduce

It's a little difficult to reproduce since it happens so randomly. But, we've seen it often enough while adding, deleting, or listing payment methods from **My Account > Payment Methods**.

We should try to catch this error in the API client and retry the request when this occurs.

-----

Changes made should be merged into `poc/upe-instances-multiplied`, not develop.

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5115

#### What actually is the issue we try to address
WCPay client does not handle server redirection responses which are redirections because of the `doing_wp_cron` query parameter attached to the URL. There is a list of Cron Events in the WCPay Server (`WP Admin` -> `Tools` -> `Cron Events`) that are executed per the schedule on the local environment. 

After a particular Cron Event is run, the first request to the server always results in an `HTTP 302 Found` response with the empty body and the `Location` key-value pair provided in the header's array. Before this PR, our client code implementation assumed that the `body` part of the request is always non-empty and thus failed [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L223) while trying to get `data` value from the empty string.

<details>
  <summary>Please see what the error looks like from the UI perspective.</summary>

image

</details>

This PR: 
1. Improves the resilience of the WCPay client, which now retries the aforementioned `HTTP 302` server responses and thus fixes the error described in https://github.com/Automattic/woocommerce-payments/issues/5115.
2. Slightly modifies the implementation from https://github.com/Automattic/woocommerce-payments/pull/5309, which now allows the retry mechanism to be applied not only for requests with `Idempotency-Key` header but also for `GET`/`DELETE` requests, which [are idempotent by their nature](https://stripe.com/docs/api/idempotent_requests#:~:text=Sending%20idempotency%20keys%20in%20GET%20and%20DELETE%20requests%20has%20no%20effect%20and%20should%20be%20avoided%2C%20as%20these%20requests%20are%20idempotent%20by%20definition.).

#### Testing instructions
1. Checkout to the latest `develop` without the changes from this PR.
2. Comment out [these lines](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L215-L220) to avoid fetching payment methods from the db cache. With this prep step, we ensure that each request will eventually hit the server and thus allow us to reproduce the issue and test the fix properly.
3. On the merchant site, go to `My account` -> `Payment methods` and make sure you see the list of payment methods without any errors. 
4. Go to the server's admin panel (`http://localhost:8086/wp-admin/`) and then to `Tools` -> `Cron Events`.
5. Take any hook from the list and click `Edit`.
6. In the `Next Run` section, choose the time in the near future when you would like the event to be run. Let's say you set it to `13:15:00`.
8. Wait until `13:15:00` and refresh the `Payment methods` page after `13:15:00` as the first request on the merchant site. 
9. Confirm that you see the error `Warning: Illegal string offset 'data' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-customer-service.php on line 223`
10. Checkout to `fix/wcpay-server-api-redirects`, perform steps 2-8 once again on that branch and confirm that there's no error anymore.
-------------------

**Post merge**

### DIFF
--- a/includes/wc-payment-api/class-wc-payments-api-client.php
+++ b/includes/wc-payment-api/class-wc-payments-api-client.php
@@ -110,16 +110,25 @@ class WC_Payments_API_Client {
 	private $wcpay_db;
 
 	/**
+	 * Helper class to perform operations on a response.
+	 *
+	 * @var WC_Payments_Api_Utils
+	 */
+	private $api_utils;
+
+	/**
 	 * WC_Payments_API_Client constructor.
 	 *
-	 * @param string           $user_agent  - User agent string to report in requests.
-	 * @param WC_Payments_Http $http_client - Used to send HTTP requests.
-	 * @param WC_Payments_DB   $wcpay_db    - DB access wrapper.
+	 * @param string                $user_agent  - User agent string to report in requests.
+	 * @param WC_Payments_Http      $http_client - Used to send HTTP requests.
+	 * @param WC_Payments_DB        $wcpay_db    - DB access wrapper.
+	 * @param WC_Payments_Api_Utils $api_utils - Helper class to perform operations on a response.
 	 */
-	public function __construct( $user_agent, $http_client, $wcpay_db ) {
+	public function __construct( $user_agent, $http_client, $wcpay_db, $api_utils ) {
 		$this->user_agent  = $user_agent;
 		$this->http_client = $http_client;
 		$this->wcpay_db    = $wcpay_db;
+		$this->api_utils   = $api_utils;
 	}
 
 	/**
@@ -2094,7 +2103,6 @@ class WC_Payments_API_Client {
 		$headers        = apply_filters( 'wcpay_api_request_headers', $headers );
 		$stop_trying_at = time() + self::API_TIMEOUT_SECONDS;
 		$retries        = 0;
-		$retries_limit  = array_key_exists( 'Idempotency-Key', $headers ) ? self::API_RETRIES_LIMIT : 0;
 
 		while ( true ) {
 			$response_code  = null;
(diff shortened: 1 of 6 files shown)


includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2148,6 +2159,24 @@ protected function request( $params, $api, $method, $is_site_specific = true, $u
...
+	 * @return string - URL prepared for a retry.
+	 */
+	private function prepare_url_for_retry( $response, $url ) {
+		$response_code = wp_remote_retrieve_response_code( $response );
+
+		// If the redirect occurs because of the wp cron job, we need to retry the request with existing URL and avoid updating it for the sake of simplicity.
+		if ( 302 === $response_code && $this->api_utils->is_doing_wp_cron_query_parameter_present( $response ) ) {
+			return $url;
+		}
+		return $url;
```
htdat, line 2177: In case, we still follow this redirection to fix the issue: This function always returns `$url`. Why do we need it?

htdat (commented):
Have some thoughts here p1672911777864179-slack-CGGCLBN58

includes/wc-payment-api/class-wc-payments-api-client.php
timur27: Essentially, I created this method as a foundation for any possible future enhancements. If for some `302` redirects we'll see any need to actually use the new path coming from the `Location` header, we'll need to sanitize that new URL by removing Jetpack-related query parameters such as `token`, `signature`, or `nonce`. This method should be a good place to work with the `$url` then.

timur27:
Closing this PR in favor of the change in the local environment, as the problem is local-specific.

PR #5364, opened 2023-01-04
Disable email input WooPay when Express Checkout is enabled

Mayisha:
#### Description
The WooPay OTP popup is loaded when the email input has a valid WooPay email. With express checkout enabled, similar popup loads on the button click. When express checkout is enabled the email driven popup should not load. This PR adds a check for the express checkout feature flag to decide which popup should be loaded for initiating WooPay.

- Added a check for the express checkout feature flag. Loading the express checkout iframe if the feature flag is enabled, loading the email driven iframe when the feature flag is disabled.
- Created a `woopayCheckEmailInput` function to check for the existence of WooPay users based on the value of the email field. Logics are mainly extracted from the [email-input-iframe.js](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/platform-checkout/email-input-iframe.js). This is needed because based on the user's existence we need to render the `Save my information for a faster and secure checkout`. 

#### Testing instructions
**Express checkout disabled:** ( both shortcode and blocks checkout )
- Disable the express checkout feature flag from dev-tools.
- Go to the merchant checkout page and ensure that everything works as before. The WooPay OTP popup should load when an existing email is used in the email input. ( both shortcode and blocks checkout )
- When a new email is used in the email field `Save my information for a faster and secure checkout` checkbox should be present.

**Express checkout enabled:** ( both shortcode and blocks checkout )
- Enable the express checkout feature flag from dev-tools.
- Go to the merchant checkout page. Using an existing email on the email field should not trigger the WooPay popup.
-  Express checkout button should be present and clicking on that should show the WooPay popup.
- When a new email is used in the email field `Save my information for a faster and secure checkout` checkbox should be present.

### DIFF
--- a/changelog/fix--1513-single-woopay-popup
+++ b/changelog/fix--1513-single-woopay-popup
@@ -1,0 +1,3 @@
+Significance: patch
+Type: fix
+Comment: Added a check to load either the WooPay express checkout iframe or the email driven iframe. WooPay express checkout is behind a feature flag.
--- a/client/checkout/blocks/index.js
+++ b/client/checkout/blocks/index.js
@@ -22,6 +22,7 @@ import request from '../utils/request';
 import enqueueFraudScripts from 'fraud-scripts';
 import paymentRequestPaymentMethod from '../../payment-request/blocks';
 import { handlePlatformCheckoutEmailInput } from '../platform-checkout/email-input-iframe';
+import { woopayCheckEmailInput } from '../platform-checkout/woopay-check-email-input';
 import wooPayExpressCheckoutPaymentMethod from '../platform-checkout/express-button/woopay-express-checkout-payment-method';
 
 // Create an API object, which will be used throughout the checkout.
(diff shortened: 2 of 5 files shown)


brettshumaker (changes requested):
Nice work! This looks good to me and works how you describe it.

Looks like there's a separate issue with the functionality of the express button on the checkout pages - it's not picking up the email address from the input field so the iframe src URL contains `?email=undefined` and the user is always prompted to enter their email address. I think we can address _adding_ the email to the iframe URL in another issue but do you think we could handle _removing_ the email param in this PR?

Thanks!

Mayisha:
@brettshumaker I have added a check in https://github.com/Automattic/woocommerce-payments/pull/5364/commits/d515c72e148f1a5bb1a7cb986a062cdf88adcab9. I did not get rid of the `email` param completely as we are going to use it later.

brettshumaker (approved):
Looks good! Thanks for making that change.

PR #5373, opened 2023-01-06
Fix loading correct script dependencies even re-registering them in the same request

### LINKED ISSUE #5137: Conflict with Max Mega Menu
Max Mega Menu needs to load scripts from widgets.php onto the nav-menus.php page, this sometimes duplicates calls depending on how other plugins are coded. In this case, register_payment_scripts() is being called twice. Usually this is not a problem, but on line 431 of class-wc-payments-admin.php 'require_once' is being used. The first time, require_once correctly loads the dependencies array, but the second time, require_once returns null. This is causing the error.

Please can I suggest this change?

Line 431

from:

$script_asset      = file_exists( $script_asset_path ) ? require_once $script_asset_path : [ 'dependencies' => [] ];

to:

$script_asset      = file_exists( $script_asset_path ) ? include $script_asset_path : [ 'dependencies' => [] ];

### Describe the bug
When WooCommerce Payments and Max Mega Menu are installed together, there is a Warning/Error on the nav-menus.php page:

`An error of type E_ERROR was caused in line 558 of the file /home/ubuntu/www.trgfineart.com/wp-content/plugins/woocommerce-payments/includes/admin/class-wc-payments-admin.php. Error message: Uncaught TypeError: array_merge(): Argument #1 must be of type array, null given in /home/ubuntu/www.trgfineart.com/wp-content/plugins/woocommerce-payments/includes/admin/class-wc-payments-admin.php:558`
### To Reproduce

Install Max Mega Menu and WooCommerce Payments and navigate to Appearance > Menus.

### Actual behavior
Error is thrown

### Expected behavior
No error

htdat:
Fixes #5137

This issue does not happen in most cases as in a single request, we do not re-register our scripts and their dependencies.

However, with this specific plugin [Max Mega Menu](https://wordpress.org/plugins/megamenu/), they try to enqueue scripts again for `wp-admin/nav-menus.php` page to suit their specific needs: https://plugins.trac.wordpress.org/browser/megamenu/tags/3.0/megamenu.php#L169

That means, this line (and similar ones) will be called twice: 
https://github.com/Automattic/woocommerce-payments/blob/a79c5a70db6bc23c52e6e248d52f9481c220e457/includes/admin/class-wc-payments-admin.php#L433

And then `require_once` will return `true` rather than the file content itself.

- Simply change `require_once` to `require` can fix this issue. I think that's fine for us to reload in this case. 
- This approach is similar to WooCommerce Blocks plugin - [code ref](https://github.com/woocommerce/woocommerce-blocks/blob/d682079cda7d223a67af9d81e1cb72d8c99b1362/src/Assets/Api.php#L97-L99).

#### Testing instructions

* Enable debug.log in wp-config.php. https://wordpress.org/support/article/debugging-in-wordpress/
* Install https://wordpress.org/plugins/megamenu/ and activate it.
* Use the `develop` branch.
* Visit http://your-test.site/wp-admin/nav-menus.php
* Check wp-content/debug.log, get this kind of notices/errors. Note: depends on which PHP version you're using, it can be only a notice or a fatal error. 

> PHP Notice:  Trying to access array offset on value of type bool in /var/www/html/wp-content/plugins/woocommerce-payments/includes/admin/class-wc-payments-admin.php on line 437
* Check out this PR. 
* Re-visit http://your-test.site/wp-admin/nav-menus.php
* Confirm: no longer new error/notice in debug.log.
* Visit other WCPay pages under WP Admin > Payments, and the front-end checkout pages: confirm there is no relevant error/issue in debug.log or browser console.

-------------------

- [x] NO NEED. Covered with tests (or have a good reason not to test in description ☝️)
- [x] NO NEED. Tested on mobile (or does not apply)

**Post merge**

- [x] NO NEED. Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [x] NO NEED. Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [x] NO NEED. Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

### DIFF
--- a/changelog/fix-5137-confilict-max-mega-menu
+++ b/changelog/fix-5137-confilict-max-mega-menu
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Ensure script dependencies are loaded properly even when trying to register them again in the same request.
--- a/includes/admin/class-wc-payments-admin.php
+++ b/includes/admin/class-wc-payments-admin.php
@@ -430,7 +430,7 @@ class WC_Payments_Admin {
 	public function register_payments_scripts() {
 		$script_src_url    = plugins_url( 'dist/index.js', WCPAY_PLUGIN_FILE );
 		$script_asset_path = WCPAY_ABSPATH . 'dist/index.asset.php';
-		$script_asset      = file_exists( $script_asset_path ) ? require_once $script_asset_path : [ 'dependencies' => [] ];
+		$script_asset      = file_exists( $script_asset_path ) ? require $script_asset_path : [ 'dependencies' => [] ];
 		wp_register_script(
 			'WCPAY_DASH_APP',
 			$script_src_url,
@@ -539,7 +539,7 @@ class WC_Payments_Admin {
 
 		$tos_script_src_url    = plugins_url( 'dist/tos.js', WCPAY_PLUGIN_FILE );
 		$tos_script_asset_path = WCPAY_ABSPATH . 'dist/tos.asset.php';
-		$tos_script_asset      = file_exists( $tos_script_asset_path ) ? require_once $tos_script_asset_path : [ 'dependencies' => [] ];
+		$tos_script_asset      = file_exists( $tos_script_asset_path ) ? require $tos_script_asset_path : [ 'dependencies' => [] ];
 
 		wp_register_script(
 			'WCPAY_TOS',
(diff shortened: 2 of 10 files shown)


htdat:
FYI. The current E2E test failure is not related to this PR. More investigation here p1672932466830799-slack-C04D1QU76PK

jessy-p (approved):
LGTM! Tested, no longer getting error like `Trying to access array offset on value of type bool` in the logs on this branch.

PR #5378, opened 2023-01-06
Fix Tooltip rates display when there is a promo fee active

### LINKED ISSUE #5367: Tooltip not showing base and total rate when promo fees structure is applied
### Describe the bug

Whenever we have promo fee structure applied to an account, the credit card fees tooltip within `Payments > Settings` seems to show no fees against Base fee and Total per transaction fees.

### To Reproduce

1. Apply promo fees - eg `wcpay-promo-2022-us-incentive-20-off`
2. Browse to settings and mouse-over the fees related pill alongside `Credit card/ Debit card`

### Actual behavior
The tooltip shows no `Base Fee` and `Total per transaction` fees

### Screenshots
image

### Expected behavior
The tooltip should show the default `Base Fee` and resulting `Total per transaction` fees. ( details of promotional fee is shown in the main `Payments > Overview` page. 

Any

Not applicable. 

### Additional context
This bug was introduced as a side effect to a recent PR - #5299, where we made the tooltip show custom fees, when applicable.

nagpai:
Fixes #5367

The PR does the following changes: 

* It makes the tool tip show correct `Base` and `Total per transaction` fees within `Payments > Settings > Payment Methods` section.  There was a bug that made the tool tip have the fees missing whenever there was a promotional fee applied. 
* The PR also fixes the Pill component to show custom base fees, if applicable. Else it will show the rack rate base fee ( currently 2.9% + $0.30 

#### Assumption 

If there is a promo fee applied, the discount object would contain a `discount` property, and no `percentage_rate` nor `fixed_rate` .  Hence, the PR uses the absence of these as a way to identify a promo in action vs a custom fee. Kindly let me know if this is valid. Happy to rework, otherwise. Indicating [reference here](https://github.com/Automattic/woocommerce-payments/blob/67788e6cbf069e843d709e18e20d3d849eb8a7fd/client/utils/account-fees.tsx#L251)

#### Screenshots

<details><summary>

##### Before: 

</summary>
image

</details>
<details><summary>

##### After: 

</summary>

image

</details>

#### Testing instructions

* Checkout to this branch and apply a promo fee to the test site - say - `wcpay-promo-2022-us-incentive-20-off`
* Browse to `Payments > Settings > Payment Methods` section
* The Pill should read `From 2.9% + $0.30` ( rack rate base fees ) 
* Mouse over the pill, and you should see the same rate as Base fee, and correctly calculated `Total per transaction fee`
* As an added check, browse to `Payments > Overview` . The `Account Details` section should show a sub-section of `Active discounts`. This section would show the discounted rate ( with rack rates struck off ), and a progress bar indicating utilization of volume limit. 
* Change the fees to a custom fee and repeat the test steps above
* This time the pill, and the tooltip mouseover, should both show the custom base fee. ( check screenshot below ) 
image

-------------------

- [x] Tested on mobile - not applicable.

**Post merge**

### DIFF
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -317,6 +317,17 @@ export const formatMethodFeesDescription = (
 	} );
 };
 
+export const formatPillFeesDescription = (
+	accountFees: FeeStructure
+): string => {
+	const pillBaseFee = getTooltipAndPillBaseFee( accountFees );
+	return sprintf(
+		__( 'From %1$f%% + %2$s', 'woocommerce-payments' ),
+		formatFee( pillBaseFee.percentage_rate ),
+		formatCurrency( pillBaseFee.fixed_rate, pillBaseFee.currency )
+	);
+};
+
 export const getTransactionsPaymentMethodName = (
 	paymentMethod: PaymentMethod
 ): string => {
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -85,9 +85,7 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
-export const getTooltipAndPillBaseFee = (
-	accountFees: FeeStructure
-): BaseFee => {
+const getTooltipAndPillBaseFee = ( accountFees: FeeStructure ): BaseFee => {
 	const tooltipAndPillBaseFee = getCurrentBaseFee( accountFees );
 
 	if ( ! tooltipAndPillBaseFee.percentage_rate ) {
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -85,7 +85,9 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
-const getTooltipAndPillBaseFee = ( accountFees: FeeStructure ): BaseFee => {
+export const getTooltipAndPillBaseFee = (
+	accountFees: FeeStructure
+): BaseFee => {
 	const tooltipAndPillBaseFee = getCurrentBaseFee( accountFees );
 
 	if ( ! tooltipAndPillBaseFee.percentage_rate ) {
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -85,22 +85,36 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
+export const getTooltipAndPillBaseFee = (
+	accountFees: FeeStructure
+): BaseFee => {
+	const tooltipAndPillBaseFee = getCurrentBaseFee( accountFees );
+
+	if ( ! tooltipAndPillBaseFee.percentage_rate ) {
+		tooltipAndPillBaseFee.percentage_rate =
+			accountFees.base.percentage_rate;
+	}
+
+	if ( ! tooltipAndPillBaseFee.fixed_rate ) {
+		tooltipAndPillBaseFee.fixed_rate = accountFees.base.fixed_rate;
+	}
+
+	return tooltipAndPillBaseFee;
+};
+
 export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
-	const effectiveBaseFee = {
-		...accountFees.base,
-		...accountFees.discount[ 0 ],
-	};
+	const tooltipAndPillBaseFee = getTooltipAndPillBaseFee( accountFees );
 
 	const total = {
 		percentage_rate:
-			effectiveBaseFee.percentage_rate +
+			tooltipAndPillBaseFee.percentage_rate +
 			accountFees.additional.percentage_rate +
 			accountFees.fx.percentage_rate,
 		fixed_rate:
-			effectiveBaseFee.fixed_rate +
+			tooltipAndPillBaseFee.fixed_rate +
 			accountFees.additional.fixed_rate +
 			accountFees.fx.fixed_rate,
 		currency: accountFees.base.currency,
@@ -114,7 +128,7 @@ export const formatMethodFeesTooltip = (
 		<div className={ 'wcpay-fees-tooltip' }>
 			<div>
 				<div>Base fee</div>
-				<div>{ getFeeDescriptionString( effectiveBaseFee ) }</div>
+				<div>{ getFeeDescriptionString( tooltipAndPillBaseFee ) }</div>
 			</div>
 			{ hasFees( accountFees.additional ) ? (
 				<div>
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -106,11 +106,13 @@ export const formatMethodFeesTooltip = (
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
 	const currentBaseFee = getCurrentBaseFee( accountFees );
+	// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied. Use this to calculate the total fee too.
 	const currentFeeWithBaseFallBack = getCurrentFeeWithBaseFallback(
 		currentBaseFee,
 		accountFees.base
 	);
 
+	// Use the current fee with the base fallback, if applicable, to calculate the total fee. Eg. when we have a promotional discount fee applied.
 	const total = {
 		percentage_rate:
 			currentFeeWithBaseFallBack.percentage_rate +
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -95,7 +95,6 @@ export const formatMethodFeesTooltip = (
 		? currentBaseFee
 		: accountFees.base;
 
-	// Use the current fee with the base fallback, if applicable, to calculate the total fee. Eg. when we have a promotional discount fee applied.
 	const total = {
 		percentage_rate:
 			currentFeeWithBaseFallBack.percentage_rate +
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -89,15 +89,18 @@ export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
-	const currentBaseFee = getCurrentBaseFee( accountFees );
+	const effectiveBaseFee = {
+		...accountFees.base,
+		...accountFees.discount[ 0 ],
+	};
 
 	const total = {
 		percentage_rate:
-			currentBaseFee.percentage_rate +
+			effectiveBaseFee.percentage_rate +
 			accountFees.additional.percentage_rate +
 			accountFees.fx.percentage_rate,
 		fixed_rate:
-			currentBaseFee.fixed_rate +
+			effectiveBaseFee.fixed_rate +
 			accountFees.additional.fixed_rate +
 			accountFees.fx.fixed_rate,
 		currency: accountFees.base.currency,
@@ -111,7 +114,7 @@ export const formatMethodFeesTooltip = (
 		<div className={ 'wcpay-fees-tooltip' }>
 			<div>
 				<div>Base fee</div>
-				<div>{ getFeeDescriptionString( currentBaseFee ) }</div>
+				<div>{ getFeeDescriptionString( effectiveBaseFee ) }</div>
 			</div>
 			{ hasFees( accountFees.additional ) ? (
 				<div>
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -85,32 +85,15 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
-// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied.
-const getCurrentFeeWithBaseFallback = (
-	currentFee: BaseFee | DiscountFee,
-	baseFee: BaseFee
-) => {
-	if ( ! currentFee.percentage_rate ) {
-		currentFee.percentage_rate = baseFee.percentage_rate;
-	}
-
-	if ( ! currentFee.fixed_rate ) {
-		currentFee.fixed_rate = baseFee.fixed_rate;
-	}
-
-	return currentFee;
-};
-
 export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
 	const currentBaseFee = getCurrentBaseFee( accountFees );
 	// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied. Use this to calculate the total fee too.
-	const currentFeeWithBaseFallBack = getCurrentFeeWithBaseFallback(
-		currentBaseFee,
-		accountFees.base
-	);
+	const currentFeeWithBaseFallBack = currentBaseFee.percentage_rate
+		? currentBaseFee
+		: accountFees.base;
 
 	// Use the current fee with the base fallback, if applicable, to calculate the total fee. Eg. when we have a promotional discount fee applied.
 	const total = {
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -85,36 +85,39 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
-export const getTooltipAndPillBaseFee = (
-	accountFees: FeeStructure
-): BaseFee => {
-	const tooltipAndPillBaseFee = getCurrentBaseFee( accountFees );
-
-	if ( ! tooltipAndPillBaseFee.percentage_rate ) {
-		tooltipAndPillBaseFee.percentage_rate =
-			accountFees.base.percentage_rate;
+// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied.
+const getCurrentFeeWithBaseFallback = (
+	currentFee: BaseFee | DiscountFee,
+	baseFee: BaseFee
+) => {
+	if ( ! currentFee.percentage_rate ) {
+		currentFee.percentage_rate = baseFee.percentage_rate;
 	}
 
-	if ( ! tooltipAndPillBaseFee.fixed_rate ) {
-		tooltipAndPillBaseFee.fixed_rate = accountFees.base.fixed_rate;
+	if ( ! currentFee.fixed_rate ) {
+		currentFee.fixed_rate = baseFee.fixed_rate;
 	}
 
-	return tooltipAndPillBaseFee;
+	return currentFee;
 };
 
 export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
-	const tooltipAndPillBaseFee = getTooltipAndPillBaseFee( accountFees );
+	const currentBaseFee = getCurrentBaseFee( accountFees );
+	const currentFeeWithBaseFallBack = getCurrentFeeWithBaseFallback(
+		currentBaseFee,
+		accountFees.base
+	);
 
 	const total = {
 		percentage_rate:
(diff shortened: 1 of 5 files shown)


nagpai:
@kalessil - In the current state, the PR ends up fixing the issue with the tooltip. However, the pill indicating the base fee and fixed charge, ends up showing the discounted fees. 

Needed one small clarification regarding this. Let me know if this is correct: 

1. Whenever there is a custom fee - It should show the custom base fee and fixed fee 
2. Whenever there is a promo - it should show the rack rate. 

Should I create a new PR for fixing the pill, or add it here itself since it is a small one?

kalessil:
@nagpai, it should display:
- custom fee - if applicable
- or base fee

client/utils/account-fees.tsx
```diff
@@ -85,19 +85,36 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
+export const getTooltipAndPillBaseFee = (
```
mgascam: I think your analysis is accurate here: we need `percentage_rate` and `fixed_rate` and falling back to the base fee structure seems a good idea to me.

But I am not convinced about adding this new function. I found this part of the code a bit hard to follow, even before this PR, and adding this new function was not very helpful in understanding what is going on.

client/utils/account-fees.tsx
```diff
@@ -85,19 +85,36 @@ export const getCurrentBaseFee = (
...
+	}
+
+	return tooltipAndPillBaseFee;
+};
+
 export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
-	const currentBaseFee = getCurrentBaseFee( accountFees );
```
mgascam, line 92: I'd rather leave it as it was and not use the `getTooltipAndPillBaseFee`. Maybe it is just me, but makes things harder to understand :)

client/utils/account-fees.tsx
```diff
@@ -111,7 +128,7 @@ export const formatMethodFeesTooltip = (
 		<div className={ 'wcpay-fees-tooltip' }>
 			<div>
 				<div>Base fee</div>
-				<div>{ getFeeDescriptionString( currentBaseFee ) }</div>
+				<div>{ getFeeDescriptionString( tooltipAndPillBaseFee ) }</div>
```
mgascam: IMO this is the key to fixing this bug. I think you can agree with me that the problem we have is the generation of the fee description here, and also the 'Total per transaction' part. It shows blank because `getFeeDescriptionString` returns an empty string in the case where `percentage_rate` or `fixed_rate` is missing.

But I would go with a different approach. I'd like to introduce a new function. Just a wrapper around `getFeeDescriptionString`, something like:

```
// If the current base fee is missing a fixed or percentage rate, use the base fee's rate instead.
const getFeeDescriptionStringWithBaseFallback = (
	currentFee: BaseFee,
	baseFee: BaseFee
) => {
	if ( ! currentFee.percentage_rate ) {
		currentFee.percentage_rate = baseFee.percentage_rate;
	}

	if ( ! currentFee.fixed_rate ) {
		currentFee.fixed_rate = baseFee.fixed_rate;
	}

	return getFeeDescriptionString( currentFee );
};

```
So we would call this function in this line (and also to build the total):
```suggestion
				<div>{ getFeeDescriptionStringWithBaseFallback( currentBaseFee, accountFees.base ) }</div>
```
If you look carefully, this is just a different way of applying the fix you already have. IMO it is just shorter, and I also believe it helps to convey the intention and clarify things. 

What do you think?

client/utils/account-fees.tsx
```diff
@@ -302,6 +319,17 @@ export const formatMethodFeesDescription = (
 	} );
 };
 
+export const formatPillFeesDescription = (
```
mgascam: Following my previous comments, I think we could spare this function and stick to the original implementation.

client/utils/test/account-fees.tsx
```diff
@@ -265,4 +267,95 @@ describe( 'Account fees utility functions', () => {
 			);
 		} );
 	} );
+
+	describe( 'getTooltipAndPillBaseFee()', () => {
```
mgascam: Maybe we would not need this test suite?

mgascam (commented):
Thanks Nagesh, this is a tricky issue. I followed the testing instructions and the solution you propose seems to work.

But I have some concerns related to the complexity of the code. I'd go with a different approach 😄 . I hope I explained myself in the comments. Please have a look and let me know what you think.

kalessil (changes requested):
Works as was requested. I should admit having the base fee displayed on screenshot is rather confusing:
Screenshot from 2023-01-12 10-38-19

We should display there as it was before - with regards to discount. The tooltip behavior should stay as it is:
Screenshot from 2023-01-12 10-41-02

It'll be a mismatch, but that's a point where I can take the topic further to UI/UX folks.

client/utils/account-fees.tsx
nagpai: This is a beautiful and simplified alternative! I also like how it leaves the original function untouched, thereby avoiding the chance of any side-effects. Thanks so much, Miguel! I will rework this 👍

nagpai:
@kalessil 

>We should display there as it was before - with regards to discount. The tooltip behavior should stay as it is:

Thanks for your review. I have now made the Pill to show the discounted rate, while keeping the tool tip same. 

I hope this is fine now. 

image

client/utils/account-fees.tsx
nagpai: It looks like the Base Fee message part on the tooltip  requires the ready formatted string generated by `getFeeDescriptionString` function.  However for Total per transaction display, there is a `total` variable that requires the fee itself (with fall back if the percentage and fixed fee is absent).

Hence, I have created a function -[ getCurrentFeeWithBaseFallBack ](https://github.com/Automattic/woocommerce-payments/blob/ae30b742c5a0880919f363cfabc64dbd6b4e4c1c/client/utils/account-fees.tsx#L89)

The base fee section of the tooltip, uses the output of the above function, and gets a formatted string via existing [`getFeeDescriptionString`](https://github.com/Automattic/woocommerce-payments/blob/ae30b742c5a0880919f363cfabc64dbd6b4e4c1c/client/utils/account-fees.tsx#L57) function.

The total fee uses the currentFee computed with fallback, adds the other fees. This computed total fee is then displayed using the same `getFeeDescriptionString` function. 

If this is fine, will it make sense for me to write a test for `getCurrentFeeWithBaseFallback` ?  Kindly guide me. Thanks!

client/utils/account-fees.tsx
nagpai: This function is now redundant and has been deleted, since we are no longer wanting to show a separate rate on the Pill . We will show the same discounted rate.

client/utils/test/account-fees.tsx
nagpai: Removed this. I will write a new one for the new function if needed. Thanks

client/utils/account-fees.tsx
```diff
@@ -85,19 +85,39 @@ export const getCurrentBaseFee = (
...
+
+	return currentFee;
+};
+
 export const formatMethodFeesTooltip = (
 	accountFees: FeeStructure
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
 	const currentBaseFee = getCurrentBaseFee( accountFees );
+	const currentFeeWithBaseFallBack = getCurrentFeeWithBaseFallback(
```
mgascam: Thanks Nagesh!

Apologies for being so annoying, but I don't like very much the fact that we have `currentBaseFee` and `currentFeeWithBaseFallBack`. I still find it does not help with clarifying things. Maybe it is just a matter of adding more comments or simply personal preference. I'd like to hear other opinions.

In case it helps, I'd go with:

```diff
	index b2b9151a..72dcf3f7 100644
--- a/client/utils/account-fees.tsx
+++ b/client/utils/account-fees.tsx
@@ -106,18 +106,14 @@ export const formatMethodFeesTooltip = (
 ): JSX.Element => {
        if ( ! accountFees ) return <></>;
        const currentBaseFee = getCurrentBaseFee( accountFees );
-       const currentFeeWithBaseFallBack = getCurrentFeeWithBaseFallback(
-               currentBaseFee,
-               accountFees.base
-       );
 
        const total = {
                percentage_rate:
-                       currentFeeWithBaseFallBack.percentage_rate +
+                       currentBaseFee.percentage_rate +
                        accountFees.additional.percentage_rate +
                        accountFees.fx.percentage_rate,
                fixed_rate:
-                       currentFeeWithBaseFallBack.fixed_rate +
+                       currentBaseFee.fixed_rate +
                        accountFees.additional.fixed_rate +
                        accountFees.fx.fixed_rate,
                currency: accountFees.base.currency,
@@ -132,7 +128,12 @@ export const formatMethodFeesTooltip = (
                        <div>
                                <div>Base fee</div>
                                <div>
-                                       { getFeeDescriptionString( currentFeeWithBaseFallBack ) }
+                                       { getFeeDescriptionString(
+                                               getCurrentFeeWithBaseFallback(
+                                                       currentBaseFee,
+                                                       accountFees.base
+                                               )
+                                       ) }
                                </div>
                        </div>
                        { hasFees( accountFees.additional ) ? (
@@ -156,7 +157,9 @@ export const formatMethodFeesTooltip = (
                        <div>
                                <div>Total per transaction</div>
                                <div className={ 'wcpay-fees-tooltip__bold' }>
-                                       { getFeeDescriptionString( total ) }
+                                       { getFeeDescriptionString(
+                                               getCurrentFeeWithBaseFallback( total, accountFees.base )
+                                       ) }
                                </div>
                        </div>
                        { wcpaySettings &&

```

But, as I said, this may be just a personal preference as both approaches do the same.

mgascam (commented):
I am happy to approve the PR since it fixes the issue at hand. But at the same time not fully convinced if we can try to simplify things further. Please have a look at my latest comment and let me know what you think.

Thanks!

client/utils/account-fees.tsx
nagpai: >Apologies for being so annoying, but I don't like very much the fact that we have currentBaseFee and currentFeeWithBaseFallBack

No need to apologize, and this is not annoying at all. Infact your opinion and stress on making the code easily understandable makes a lot of sense. Let me go back to the drawing board and test out some ideas I see here.

nagpai: Things are slightly tricky here because of the `total` variable and how we need to add the fallback base rate, component-wise - for `percentage` and `fix rates`.  @mgascam , the last suggestion ends up showing the base fee as the total fee, without adding the additional and FX fees to it. 

image

This happens because the fallback rates, if applicable, need to be added separately for `percentage` and `fixed` rate, where we [first compute](https://github.com/Automattic/woocommerce-payments/blob/ae30b742c5a0880919f363cfabc64dbd6b4e4c1c/client/utils/account-fees.tsx#L109) `total`.  This adjusted rate needs to be then given as an input to the string generation function. That's why I opted to create a separate const earlier to store the current fee with base fall back ( if applicable ) . 

I have added comments in my last update here - [727d231](https://github.com/Automattic/woocommerce-payments/pull/5378/commits/727d2319a6df07fb47f77c2551b9022ad2064896). Let me know if this makes sense.

client/utils/test/account-fees.tsx
nagpai: Marking as resolved since I have removed this function for now

client/utils/account-fees.tsx
nagpai: Marking this as resolved for now, since the function `getTooltipAndPillBaseFee` has been removed. We are discussing separately about the new `getCurrentFeeWithBaseFallback` [function here](https://github.com/Automattic/woocommerce-payments/pull/5378#discussion_r1068383656) and [here](https://github.com/Automattic/woocommerce-payments/pull/5378#discussion_r1069570022) . Thanks

client/utils/account-fees.tsx
```diff
@@ -85,19 +85,41 @@ export const getCurrentBaseFee = (
 		: accountFees.base;
 };
 
+// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied.
+const getCurrentFeeWithBaseFallback = (
+	currentFee: BaseFee | DiscountFee,
+	baseFee: BaseFee
+) => {
+	if ( ! currentFee.percentage_rate ) {
```
kalessil: This logic is incorrect: it kind of blends fees for some cases into something weird.
We want to pick the base fee instead - from this angle we don't need the function at all.

client/utils/account-fees.tsx
kalessil: With reference to my other comment, IMO it'll be enough to do `currentFeeWithBaseFallBack = accountFees.base`.
And fixing the names is `currentBaseFee = accountFees.base`

kalessil (changes requested):

client/utils/account-fees.tsx
nagpai: Simplified and made amends here. I hope this is better. Request your review.

6db016a

client/utils/account-fees.tsx
nagpai: Thanks for the critical review and guidance. I have simplified this and made amends here 
6db016a

I hope this will work fine.

nagpai: @mgascam May I request your review too? Thanks!

client/utils/account-fees.tsx
```diff
@@ -90,14 +90,19 @@ export const formatMethodFeesTooltip = (
 ): JSX.Element => {
 	if ( ! accountFees ) return <></>;
 	const currentBaseFee = getCurrentBaseFee( accountFees );
+	// If the current fee doesn't have a fixed or percentage rate, use the base fee's rate. Eg. when there is a promotional discount fee applied. Use this to calculate the total fee too.
+	const currentFeeWithBaseFallBack = currentBaseFee.percentage_rate
```
mgascam, line 94: Do we need to check if `fixed_rate` exists in the `currentBaseFee` as well? Or is it enough to check the `percentage_rate`?

nagpai: The scenarios I thought of seeing the existing fees were 

1. Custom fees - here we have a `percentage_rate` (for sure) and `fixed_rate` (optionally) within the discount fee object ( that gets selected as the currentBaseFee)  - example - 2ecc4-pb/raw/

2. Promotional fees - the ones i have seen, currently, only have a discount percentage mentioned, without base percentage_rate, nor fixed fees. - example - 2ecc5-pb/raw/

I was unable to find a case where we do not have percentage fee and have fixed fee. Hence thought the presence or absence of percentage_rate  would be a good benchmark to decide whether to use the base fee or the discount fee object data. 

If I were to account for the fixed fee, I am wondering what would be the logic for situations where there is only a `fixed_rate`, or only `percentage_rate` mentioned in the discount fee object. Should we be falling back on Base fee rates only if both `fixed_rate` and `percentage_rate` are absent, or either.. ( sorry if I am unable to explain this in a concise way)

mgascam: Thanks for the detailed explanation @nagpai.

> I was unable to find a case where we do not have percentage fee and have fixed fee. Hence thought the presence or absence of percentage_rate would be a good benchmark to decide whether to use the base fee or the discount fee object data.

This makes sense to me as well. Marking conversation as resolved

client/utils/account-fees.tsx
nagpai: Marking this thread as resolved since the code has been changed.

nagpai:
Changing the status back to `author's reply` since the tests need a rework.

nagpai:
Tests have been reworked. Since these are among my first few tests written from scratch, I would appreciate a closer review of them 🙏 . Thanks!

client/utils/test/account-fees.tsx
```diff
@@ -265,4 +275,97 @@ describe( 'Account fees utility functions', () => {
...
+				currency: 'USD',
+			};
+
+			methodFees.fx = {
+				percentage_rate: 0.01,
+				fixed_rate: 0,
+				currency: 'USD',
+			};
+
+			const renderMethodFeesTooltip = () => {
```
mgascam: I'd suggest a different approach here, that can help in avoiding the repetition of this function. I'd say to pass the result of `render( formatMethodFeesTooltip( methodFees ) )` directly to the expect function below. Unless I am missing something.

I also suggest taking a snapshot of the HTML only. We are currently taking it from the result of calling the render function, which includes other stuff. I'd say that we should only care about the `container` property of the [rendered](https://testing-library.com/docs/react-testing-library/api#container-1) result.

client/utils/test/__snapshots__/account-fees.tsx.snap
```diff
@@ -0,0 +1,502 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Account fees utility functions formatMethodFeesTooltip() displays base fee, when only promo discount without percentage or fixed 1`] = `
+Object {
```
mgascam: I think is better to have snapshot of the rendered HTML, please look at my previous comment for more contet.

mgascam (changes requested):
I love that you added tests for the tooltip. I left a couple of suggestions about the snapshot generation.

client/utils/account-fees.tsx
```diff
@@ -90,14 +90,19 @@ export const formatMethodFeesTooltip = (
...
+		? currentBaseFee
+		: accountFees.base;
 
+	// Use the current fee with the base fallback, if applicable, to calculate the total fee. Eg. when we have a promotional discount fee applied.
```
kalessil: I believe we could omit this comment.

kalessil (approved):
Pre-approving. Leaving to Miguel guiding you on the testing approach.

client/utils/account-fees.tsx
nagpai: Sorry, I missed removing the redundant comment while removing older unwanted code. I have now removed it here 
b50fb9c

Thank you. (Marking as resolved)

client/utils/test/account-fees.tsx
nagpai: Thank you for guiding me through these great improvements! The redundant steps in my earlier code was a result of me checking an example code elsewhere. I will now remember to remove unwanted steps and simplifying the code for my future PRs, based on what you shared here.  I have made the changes here : ca527b5

client/utils/test/__snapshots__/account-fees.tsx.snap
nagpai: Thanks. Made the changes here ca527b5 

I can see the snapshot now showing only the HTML  ( `baseElement` property of the rendered object ) . I hope this is correct.

client/utils/test/account-fees.tsx
```diff
@@ -265,4 +275,91 @@ describe( 'Account fees utility functions', () => {
...
+				currency: 'USD',
+			};
+
+			methodFees.fx = {
+				percentage_rate: 0.01,
+				fixed_rate: 0,
+				currency: 'USD',
+			};
+
+			expect(
```
mgascam: Nice! But I'd use `container` instead of `baseElement`. Quoting the [docs](https://testing-library.com/docs/react-testing-library/api#baseelement-1):

> This is useful when the component you want to test renders something outside the container div, e.g. when you want to snapshot test your portal component which renders its HTML directly in the body.

I don't believe this is our case, so it seems [container](https://testing-library.com/docs/react-testing-library/api#container-1) is a better fit.

And what do you think about the following?:
```suggestion
			const { container } = render(
				formatMethodFeesTooltip( methodFees )
			);

			expect( container ).toMatchSnapshot();
```

mgascam (commented):
I added yet another nitpick comments 💅

client/utils/test/account-fees.tsx
nagpai: Notes to myself - Read the docs of the libraries too!  Thank you Miguel 🙌 

Fixed this here d687bae

mgascam (approved):
Looking good to me after latest changes!

client/utils/test/__snapshots__/account-fees.tsx.snap
nagpai: Thanks for all the time and details you gave me here! I learned a lot! 🙌

PR #5382, opened 2023-01-10
Fix ActionScheduler fatal error

reykjalin:

- Check if `WC()->session` is set before adding customer ID to the browser session when creating a customer object.

Some time between the release of WCPay 4.5.1 and 5.2.1 -- maybe because of some differences in WooCommerce versions? We're not sure why -- ActionScheduler jobs that created customers started resulting in fatal errors because the `WC()->session` object is not available in AS jobs.

Context in p1673303308974939-slack-CGGCLBN58?thread_ts=1673293381.897259&cid=CGGCLBN58

#### Testing instructions

> **Note**
> You can skip step (2) if your account on the test store already has a Stripe customer ID associated with it.

1. Make sure you're using the UPE checkout.
2. Add any amount of products to your cart and successfully complete a checkout of any amount.
3. Re-onboard WCPay to connect the store to a new WCPay account.
4. Add any amount of products to your cart and try to complete a checkout.
5. You should see a `No such customer` error:

image

6. Wait a couple seconds and then retry the payment.
7. It should now be successful.

On `develop` the payment retry will fail in step (6) and the AS job scheduled by the failed payment in step (4) will fail to run due to a fatal error that you can see in the logs.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-action-scheduler-fatal-error
+++ b/changelog/fix-action-scheduler-fatal-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent occasional fatal erros when creating customers via ActionScheduler jobs
--- a/changelog/fix-action-scheduler-fatal-error
+++ b/changelog/fix-action-scheduler-fatal-error
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Prevent occasional fatal erros when creating customers via ActionScheduler jobs
+Prevent occasional fatal errors when creating customers via ActionScheduler jobs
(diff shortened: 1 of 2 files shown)


tpaksu (approved):
Yes, I know this type of exceptions since #5228. Change looks good to me. 🚢 it.

changelog/fix-action-scheduler-fatal-error
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent occasional fatal erros when creating customers via ActionScheduler jobs
```
dechov: Errors* typo

reykjalin: Thanks for catching that! Fixed in 536c1fae 🙌

dechov (approved):

PR #5399, opened 2023-01-12
Single separator for express checkout methods

### LINKED ISSUE #4720: Multiple UPE - Stripe Link compatibility

### LINKED ISSUE #5339: Fix multiple separators when multiple express checkout methods appear simultaneously at checkout
### Describe the bug

Essentially when more than one express checkout method (e.g. Apple Pay, Google Pay, WooPay, etc.) is present at checkout, the `- OR -` separator, used to separate the express checkout buttons from the rest of the checkout or multiplied in proportion to the number of express checkout methods available. This is not a common issue, since most of these express checkout methods require rather unique environmental conditions in order to be present (e.g. Apple Pay only in Safari, Google Pay only in Chrome, etc.), so we don't find them together very commonly. However, with the introduction of WooPay and possibly Stripe Link (see #4720), this will now need to be addressed. 

This issue was first discovered in #5282, so please allow me to quote my own findings.

> Finally, note one open bug that when both WooPay and Link are enabled as express checkout methods there are two - OR - separators visible above the checkout fields.
> 
>This is an open issue beyond this PR that will occur if any two WCPay express checkout methods are enabled at checkout at the same time--regardless of whether one of them is Link or WooPay. This is because the express checkout methods are all [registered separately](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php?rgh-link-date=2022-12-13T15%3A49%3A43Z#L369-L371) and each one adds a hook to place the separator on the page ([1](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-platform-checkout-button-handler.php?rgh-link-date=2022-12-13T15%3A49%3A43Z#L85), [2](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php?rgh-link-date=2022-12-13T15%3A49%3A43Z#L86)), so consequently the separator is duplicated when more than one express checkout button is available at checkout.
>
>Ideally, the express checkout methods should all be registered together, so that we can group them together.

Multiple separators appear at checkout
_Superfluous separator snafu_

### To Reproduce

1. Enable more than one Express Checkout method at checkout. I would recommend the simplest pairing would be to enable WooPay and one of Apple Pay, Google Pay, or Link--if using the split-UPE branch (`poc/upe-instances-multiplied`) with changes from #5282 merged.
2. Add item to cart and proceed to shortcode checkout.
3. Witness the mayhem screenshotted above, as the `- OR -` separator is malevolently multiplied.

### Actual behavior

When multiple express checkout methods are registered to the shortcode checkout, the `- OR -` separator used to divide these checkout buttons from the regular checkout appears multiplied numerous times together beneath the express checkout buttons.

### Screenshots

Multiple separators appear at checkout
_Superfluous separator snafu_

### Expected behavior

Either the `- OR -` separator should appear in between each express checkout method or the express checkout methods should be grouped together in a single block--similar to the blocks checkout--with a single separator dividing them all from the rest of the checkout. I would prefer the latter solution, for the sake of consistency with blocks.

### Additional context

#5282

mdmoore:
Fixes #5339  

This branch is based off of `4720/stripe-link-intgration`. Since that hasn't yet been merged, it seems more feasible to merge it into that branch instead of `poc/upe-instances-multiplied`. 

This ensures that only a single **- OR -** separator is displayed after express checkout buttons rather than duplicated/stacked separators. The separator will appear on the product page, cart page, and checkout page.

image

#### Testing instructions

### DIFF
--- a/includes/class-wc-payments-platform-checkout-button-handler.php
+++ b/includes/class-wc-payments-platform-checkout-button-handler.php
@@ -78,13 +78,10 @@ class WC_Payments_Platform_Checkout_Button_Handler {
 		add_filter( 'wcpay_payment_fields_js_config', [ $this, 'add_platform_checkout_config' ] );
 
 		add_action( 'woocommerce_after_add_to_cart_quantity', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_after_add_to_cart_quantity', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
 
 		add_action( 'woocommerce_proceed_to_checkout', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_proceed_to_checkout', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
 
 		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
 	}
 
 	/**
--- a/includes/class-wc-payments-platform-checkout-button-handler.php
+++ b/includes/class-wc-payments-platform-checkout-button-handler.php
@@ -300,16 +300,4 @@ class WC_Payments_Platform_Checkout_Button_Handler {
 		<?php
 	}
 
-	/**
-	 * Display payment request button separator.
-	 */
-	public function display_platform_checkout_button_separator_html() {
-		if ( ! $this->should_show_platform_checkout_button() ) {
-			return;
-		}
-		?>
-		<p id="wcpay-payment-request-button-separator" style="margin-top:1.5em;text-align:center;">&mdash; <?php esc_html_e( 'OR', 'woocommerce-payments' ); ?> &mdash;</p>
-		<?php
-	}
-
 }
--- a/includes/class-wc-payments-stripe-link-button-handler.php
+++ b/includes/class-wc-payments-stripe-link-button-handler.php
@@ -55,7 +55,6 @@ class WC_Payments_Stripe_Link_Button_Handler {
 		}
 
 		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_button_html' ], -2 );
-		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_button_separator_html' ], -1 );
 
 		// Don't load for change payment method page.
 		if ( isset( $_GET['change_payment_method'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
--- a/includes/class-wc-payments-stripe-link-button-handler.php
+++ b/includes/class-wc-payments-stripe-link-button-handler.php
@@ -84,23 +84,6 @@ class WC_Payments_Stripe_Link_Button_Handler {
 	}
 
 	/**
-	 * Display payment request button separator.
-	 */
-	public function display_button_separator_html() {
-		if ( ! WC_Payments_Features::is_link_enabled() ) {
-			return;
-		}
-
-		if ( ! $this->is_checkout() ) {
-			return;
-		}
-		?>
-		<p id="wcpay-payment-request-button-separator" style="margin-top:1.5em;text-align:center;">&mdash; <?php esc_html_e( 'OR', 'woocommerce-payments' ); ?> &mdash;</p>
-		<?php
-	}
-
-
-	/**
 	 * Checks if this is the checkout page or content contains a cart block.
 	 *
 	 * @return boolean
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -470,6 +470,10 @@ class WC_Payments {
 			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
 			add_action( 'woocommerce_proceed_to_checkout', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
 			add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+
+			if ( WC_Payments_Features::is_payment_request_enabled() ) {
+				add_action( 'before_woocommerce_pay_form', [ __CLASS__, 'display_express_checkout_button_separator_html' ], 2 );
+			}
 		}
 
 		add_action( 'rest_api_init', [ __CLASS__, 'init_rest_api' ] );
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -466,6 +466,12 @@ class WC_Payments {
 			WC_Payments_Subscriptions::init( self::$api_client, self::$customer_service, self::get_gateway(), self::$account );
 		}
 
+		if ( WC_Payments_Features::is_express_checkout_enabled() ) {
+			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+			add_action( 'woocommerce_proceed_to_checkout', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+			add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+		}
+
 		add_action( 'rest_api_init', [ __CLASS__, 'init_rest_api' ] );
 		add_action( 'woocommerce_woocommerce_payments_updated', [ __CLASS__, 'set_plugin_activation_timestamp' ] );
 
@@ -1397,5 +1403,19 @@ class WC_Payments {
 		if ( ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) && file_exists( WCPAY_ABSPATH . 'dist/runtime.js' ) ) {
 			wp_enqueue_script( 'WCPAY_RUNTIME', plugins_url( 'dist/runtime.js', WCPAY_PLUGIN_FILE ), [], self::get_file_version( 'dist/runtime.js' ), true );
 		}
+	}
+
+	/**
+	 * Display payment request button separator.
+	 */
+	public static function display_express_checkout_button_separator_html() {
+		$should_hide = WC_Payments_Features::is_payment_request_enabled() && ! WC_Payments_Features::is_woopay_express_checkout_enabled() && ! WC_Payments_Features::is_link_enabled();
+
+		if ( WC_Payments_Features::is_payment_request_enabled() && ! self::$payment_request_button_handler->should_show_payment_request_button() ) {
+			return;
+		}
+		?>
+		<p id="wcpay-payment-request-button-separator" style="margin-top:1.5em;text-align:center;<?php echo $should_hide ? 'display:none;' : ''; ?>">&mdash; <?php esc_html_e( 'OR', 'woocommerce-payments' ); ?> &mdash;</p>
+		<?php
 	}
 }
--- a/includes/class-wc-payments-features.php
+++ b/includes/class-wc-payments-features.php
@@ -146,6 +146,24 @@ class WC_Payments_Features {
 	}
 
 	/**
+	 * Checks whether Auth & Capture (uncaptured transactions tab, capture from payment details page) is enabled.
+	 *
+	 * @return bool
+	 */
+	public static function is_auth_and_capture_enabled() {
+		return '1' === get_option( self::AUTH_AND_CAPTURE_FLAG_NAME, '1' );
+	}
+
+	/**
+	 * Checks whether Payment Request is enabled.
+	 *
+	 * @return bool
+	 */
+	public static function is_payment_request_enabled() {
+		return 'yes' === WC_Payments::get_gateway()->get_option( 'payment_request' );
+	}
+
+	/**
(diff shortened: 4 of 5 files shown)


includes/class-wc-payments.php
```diff
@@ -466,6 +466,12 @@ public static function init() {
 			WC_Payments_Subscriptions::init( self::$api_client, self::$customer_service, self::get_gateway(), self::$account );
 		}
 
+		if ( WC_Payments_Features::is_express_checkout_enabled() ) {
+			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+			add_action( 'woocommerce_proceed_to_checkout', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+			add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
```
timur27, line 472: We also have [this](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L86) type of action hook that renders the separator after the Payment Request button on the _Pay for order_ page. With this comment, I'd like to discuss if it's worth moving it here.

<p align="center">
image
</p>

_Pros:_
1. By moving the action callback hook to the `if` statement here, we'll render all the possible `--OR--` separators in one place, which was the ultimate idea of this `if` statement. 
2. Eventually, we can eliminate [this](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L796-L803) method and re-use `display_express_checkout_button_separator_html` to render.

_Cons:_
1. Since the separator on the _Pay for order_ page should be displayed only for the Payment Request button, we'll also need to add an additional `if (WC_Payments_Features::is_payment_request_enabled)` statement so that we don't render it when Apple Pay/Google Pay is disabled, but e.g. WooPay is enabled. 

After considering all the pros and cons, I personally would tend to move this here. At the same time, I'm happy to hear your opinion on this! 👍

timur27, line 472: After removing the additional separator coming from the Payment Request button on the Cart, Checkout, and Product Details pages, one might be surprised that the separator disappeared, but the button itself is still rendered _**after**_ the `--OR--` separator we have in place, and not before it. 

<p align="center">
image
</p>

This is because the separator rendering callbacks are hooked here with the higher priority `-1` than the Payment Request button (`1` prio [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L76), [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L79) and also [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L82)). 

We'd need to align the priority of the Payment Request button rendering with other buttons - I can see that both WooPay and Stripe Link have `-2` priority, so maybe this value should be our choice.

> 💡 Note: I think that the sequence of buttons rendered might be important from the business perspective. For instance, we might want to always render the WooPay Express Checkout button as the first one, then Stripe Link, and then the Payment Request button. Unfortunately, the sequence is unknown to me, but I think we can consult someone from the product or UX team to confirm this. That being said, if we set the callback hook priority for the Payment Request button to be `-2`, we'll get the following state, and I'm not sure if the `GPay -> WooPay -> Stripe Link` is the desired sequence. 

<p align="center">
image
</p>

includes/class-wc-payments-platform-checkout-button-handler.php
```diff
@@ -78,13 +78,10 @@ public function init() {
 		add_filter( 'wcpay_payment_fields_js_config', [ $this, 'add_platform_checkout_config' ] );
 
 		add_action( 'woocommerce_after_add_to_cart_quantity', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_after_add_to_cart_quantity', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
 
 		add_action( 'woocommerce_proceed_to_checkout', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_proceed_to_checkout', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
 
 		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_platform_checkout_button_html' ], -2 );
-		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_platform_checkout_button_separator_html' ], -1 );
```
timur27, line 87: Can we remove [this](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-platform-checkout-button-handler.php#L469-L479) method since it's not used anymore?

includes/class-wc-payments-stripe-link-button-handler.php
```diff
@@ -55,7 +55,6 @@ public function init() {
 		}
 
 		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_button_html' ], -2 );
-		add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_button_separator_html' ], -1 );
```
timur27, line 58: Similar to the previous comment - I think we don't need the `display_button_separator_html` method in this class anymore and can safely remove it. 

The links to the method I provide here are somehow broken, so to be explicit about the method I have in mind -`WC_Payments_Stripe_Link_Button_Handler::display_button_separator_html`

includes/class-wc-payments.php
```diff
@@ -1398,4 +1404,18 @@ public static function enqueue_dev_runtime_scripts() {
 			wp_enqueue_script( 'WCPAY_RUNTIME', plugins_url( 'dist/runtime.js', WCPAY_PLUGIN_FILE ), [], self::get_file_version( 'dist/runtime.js' ), true );
 		}
 	}
+
+	/**
+	 * Display payment request button separator.
+	 */
+	public static function display_express_checkout_button_separator_html() {
+		$should_hide = WC_Payments_Features::is_payment_request_enabled() && ! WC_Payments_Features::is_woopay_express_checkout_enabled() && ! WC_Payments_Features::is_link_enabled();
```
timur27, line 1416: I assume this flag is for us to avoid the situation when the separator is rendered both in this method and in `WC_Payments_Payment_Request_Button_Handler::display_payment_request_button_separator_html`. 

I think after we remove all the redundant callback hooks in `WC_Payments_Payment_Request_Button_Handler` I mentioned in another comment and [if we render the separator for the Payment Request button on the ](https://github.com/Automattic/woocommerce-payments/pull/5399#discussion_r1071005942)_Pay for order_ page, we'll eventually be able to remove this flag. Please let me know if I missed anything or if there is any other need to have this flag? Thanks!

timur27 (changes requested):
This one is a great step forward in our UPE project - thanks for your time working on this! 🚀 When I enable WooPay and Stripe Link payment methods, I get the expected result where `-- OR --` separator is rendered after the list of methods and not twice.

I noticed a few things when the Payment Request button was enabled, which I described in separate comments. These are all related to each other. Also, when the Payment Request button is enabled, I see it has its own `-- OR --` separator rendered again. Maybe this is an expected behavior, but I think it's worth confirming it with you.

<p align="center">
image</p>

At the same time, in Blocks, the Payment Request button is grouped together with the WooPay Express Checkout button.

<p align="center">
image
</p>


In case my assumptions about the behavior will be verified and confirmed - `class-wc-payments-payment-request-button-handler.php`
adds a button rendering callback to the `woocommerce_checkout_before_customer_details` action [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L83), which needs to be removed. 

We also render an additional -- OR -- for the Payment Request button [on the checkout cart](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L80) and [while being on the product details page](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L77).
<details>
  <summary>More details here</summary>


<table>
<tr>
<th>Checkout cart</th>
<th>Product details</th>
<tr>
<td>


image

</td>
<td>

image


</td>
</tr>
</table>


</details>

All the things from above can be resolved pretty quick, though. Please let me know what do you think regarding all my thoughts in the comments and I am happy to get back to this PR review, whenever needed!

includes/class-wc-payments.php
mdmoore: Nice catch! I agree with you that this `before_woocommerce_pay_form` hook should be moved as well.

mdmoore: Ah, I didn't notice this previously. It makes sense that there might be a preferred order for express checkout buttons. I'll apply the `-2` priority for now but will look into this and follow up with a new issue if needed.

includes/class-wc-payments-platform-checkout-button-handler.php
mdmoore: Yes, absolutely! Thanks for catching this .

includes/class-wc-payments-stripe-link-button-handler.php
mdmoore: Yes indeed!

includes/class-wc-payments.php
mdmoore: >I assume this flag is for us to avoid the situation when the separator is rendered both in this method and in WC_Payments_Payment_Request_Button_Handler::display_payment_request_button_separator_html

This `$should_hide` check is actually used to hide the separator when only payment request is enabled. The default for payment request buttons is to be hidden with `display: none;`. Then some frontend JS will check if Apple Pay or Google Pay is available in the browser and subsequently show the button and separator.

timur27: Many thanks for the explanation, Mike! 👍  This one is interesting. 

----------------------------------------

Is my understanding correct that the expected UI for the Payment Request button, being the only enabled Express Payment method available, is

<p align="center">
image
</p>

instead of
<p align="center">
image
</p>
?

----------------------------------------

Another thing is that I cannot see the difference between the two cases: `display: none` is provided to the element's styles or not. It seems to me that the browser (I'm using Chrome) removes `display: none`. In effect, regardless of whether the `$should_hide` flag is `true` or `false`, the HTML element is always rendered the same. Are you experiencing the same, and should we find a solution to this?

timur27 (commented):
Awesome, thank you for resolving all of the previous things so accurately!

The additional `--OR--` separator after the Payment Request button was successfully removed from the Checkout page, Checkout cart and Product details page. ✅ 

The separator rendering on the _Pay for order_ page under the Payment Request button was moved to the `class-wc-payments.php` class, as per our discussion. ✅ 

Hooks callbacks priorities were properly updated in order to display all possible buttons in a row and before the eventual `--OR--` separator. ✅ 

> Note to the above: as you proposed, we can follow-up and confirm on the buttons sequence while rendering, and I'm completely fine with such approach. 

Unused methods removed. ✅ 


I only left one question [here](https://github.com/Automattic/woocommerce-payments/pull/5399#discussion_r1073387443), and I would really appreciate it if you could help me clarify this.

includes/class-wc-payments.php
mdmoore: Thanks Timur! Regarding the expected UI when only Payment Request is enabled, we should expect to still the separator, such as in your second screenshot:

<p align="center">
image
</p>

This is what I'm experiencing now. Are you seeing no separator in this case?

>Another thing is that I cannot see the difference between the two cases: display: none is provided to the element's styles or not. It seems to me that the browser (I'm using Chrome) removes display: none. In effect, regardless of whether the $should_hide flag is true or false, the HTML element is always rendered the same. Are you experiencing the same, and should we find a solution to this?

It's a little confusing indeed. So whenever Payment Request is the only express payment method, `$should_hide` will be true and `display:none;` will be present in the HTML. This ensures that the Payment Request button and separator are hidden by default and displayed **only** when the browser has Apple Pay or Google Pay enabled. So, in Chrome, with Google Pay configured with a payment method, `display:none` is [removed by this .show() method](https://github.com/Automattic/woocommerce-payments/blob/f592aafc4bb27110c9718b01482146e2f9dbce4e/client/payment-request/index.js#L452) after the [HTML elements are collected above](https://github.com/Automattic/woocommerce-payments/blob/f592aafc4bb27110c9718b01482146e2f9dbce4e/client/payment-request/index.js#L443). 

To see this in action, enable only Payment Request express checkout methods and test using a different profile in Chrome that doesn't have a Google Pay card attached. You should see something like this: 

image

Let me know if anything above is confusing or if you're seeing something different on your end.

mdmoore:
@timur27 [Here's a link to my response](https://github.com/Automattic/woocommerce-payments/pull/5399#discussion_r1073633776) to your latest feedback since I'm not seeing that message here. Let me know if you have any questions!

includes/class-wc-payments.php
timur27: Wow, this was super tricky - I had no idea of the separate JS running in parallel, which actually adds the separator once again, and I wouldn't figure this out without you. 🥇 Thanks, I'm resolving the thread.

timur27:
Thanks, @mdmoore, for the collaboration here! Great job! Approving. 👍 🚢

timur27 (approved):

PR #5404, opened 2023-01-13
Fix currency change for manual subscription renewal.

### LINKED ISSUE #5356: Multi-currency incompatibility with Subscriptions renewal: Currency can be changed when cart contains renewal order
### Describe the bug
**Expected behavior**: Currency [can not be changed](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L149-L153) when renewing a subscription (example: a subscription of 85.79 USD should renew at 85.79 USD, changing currency in "Customer payment page" should not have any impact)

**Current behavior**: Currency can be changed when renewing a subscription and the renewal order can have a wrong total  (example: a subscription of 85.79 USD can renew at 85.79 EUR, changing currency in "Customer payment page" works and changes the currency but not the total)

### To Reproduce
- Set USD as default shop currency
- Enable multi currency, enable EUR and some other non-USD currencies
- Set customer currency to USD, purchase a subscription product
image
- Set the Payment Method to Manual Renewal through Admin UI of Subscription
image
- Renew the Subscription through Admin UI, this will create a renewal order
image
image
- Open the "Customer payment page" as the customer, change the currency to EUR or any other non-USD one
image
  - Expectation is that the currency of the renewal order can not be changed
  - Notice that the currency can be changed, the total shown in the non-USD currency is same as the USD total. Pay for the order, notice the renewal order is completed with the wrong total  
    image
    image

jessepearson:
Fixes #5356

This PR adds a fix to stop the ability to change the currency on the checkout page via the url if a subscription renewal is in the cart. There are other measures in place to hide the currency switcher, but the url appeared to still allow the currency to be changed.

* Add another check inside the `override_selected_currency` method within the `WooCommerceSubscriptions` compatibility class. It checks to see if there is  an `order_awaiting_payment` in the current session. If so, then it checks to see if there's a subscription renewal in that order.
* Add `order_contains_renewal` wrapper method to call Subscriptions `wcs_order_contains_renewal` function.
* Add Subscriptions helper functions and methods for testing.
* Add supporting tests.
* Remove `WCPay\MultiCurrency\Utils` being used in `WooCommerceSubscriptions` compatibility class since it wasn't actually used.

#### Testing instructions

* If you do not have a second currency added on your site, please add another, such as GBP, CAD, or EUR.
  * This is done under WooCommerce > Settings > Multi-Currency.
* Install WooCommerce Subscriptions on your test site.
  * This can be found in the `all-plugins` repo, ping me if you don't know what that is.
* Disable WCPay subscriptions in dev tools.
* Create a new simple subscription product.
* Note: For the next steps I used an incognito/private browser window to test, but it may work with your account on your site.
* Add the subscription product to your cart and complete the purchase.
* Back in the admin of your site go to WooCommerce > Subscriptions, then into your new subscription.
* At the top right under _Subscription actions_ choose _Create pending renewal order_ and click the _>_ or _Update_ button.
* Go back to the front end of your site into My Account > Orders  and click Pay for your new renewal order.
* In the address bar of your browser add `/?currency=EUR`, and then hit enter.
  * Note: EUR can be replaced by any other active currency on your site that is _not_ the default or order currency.
* The currency listed in the order items should not change. 

-------------------

**Post merge**


includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -142,7 +148,7 @@ public function maybe_disable_mixed_cart( $value ) {
 	 */
 	public function override_selected_currency( $return ) {
 		// If it's not false, return it.
-		if ( $return ) {
+		if ( $return || $this->running_filters ) {
```
tpaksu: Do you think it'll make sense to be more specific about the `running_filters` variable name here? Maybe it'll be confusing in the future, and someone tries to use it in another method in this file, which may confuse the script?

tpaksu (approved):
I've tested the code with having EUR as the initial currency, and USD as the override currency, the order page prices didn't change to USD, and the payment completed with EUR on a page having the currency override querystring parameter. Thanks for fixing this! I'm approving it with a nitpick, you are free to address it or not. LGTM!

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
jessepearson: I thought about that, as well, but I couldn't think of a better name. What would you suggest? Maybe `currently_running`?

tpaksu: No, I meant something like `$this->running_currency_override_filters` maybe. Because it'll only affect this method, and I guess it should be only affecting this method. Right?

jessepearson: Ah, okay, that makes sense.

PR #5407, opened 2023-01-15
Webhooks: Switch to CPT for storage, delay webhook processing.

### LINKED ISSUE #5188: Fix Duplicate Charges: Delay webhook processing, query scheduled before failure
### Description

We should delay webhook processing on the client.

### Acceptance criteria

1. Instead of processing the webhook event `payment_intent.succeeded` immediately, store it in a similar fashion to [the webhook reliability service](<https://github.com/Automattic/woocommerce-payments/blob/c6154cddf73f49e486ee9316da1075c9cf982418/includes/class-wc-payments-webhook-reliability-service.php#L20-L20>), and schedule it to be processed in 1 or 2 minutes.
2. Still, in combination with retries upon exceptions, we should be able to use the data from webhooks, in case they arrive in the meantime. This means that storage/scheduling should be done in a way that allows us to query scheduled events by either:
   * (1) order ID (without the necessity for a server call),
   * OR (2) from order meta (which requires a step to save the event ID to order meta before scheduling the event).
3. Within [the `catch` block of `WC_Payment_Gateway_WCPay::process_payment` ](<https://github.com/Automattic/woocommerce-payments/blob/c6154cddf73f49e486ee9316da1075c9cf982418/includes/class-wc-payment-gateway-wcpay.php#L676-L749>), if there is a stored intent (carried by the webhook), we’ll use it to proceed with the other post-payment steps in the gateway. Because of the retries we would have done before that, the webhook should be already accessible this way, given that webhooks are delivered normally, and without issues.

### Testing instructions

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'
4. Observe expected behaviour

### Dev notes

**Important consideration: If we can, we should do this based on merchant-specific flags. This way, if we notice any issues, we’ll be able to toggle this delay remotely.**

### Additional context

paJDYF-6wC-p2#comment-15189<br>p1669286462288109/1669219023.651449-slack-C04CH2FJ8Q0

### LINKED ISSUE #5279: Audit and optimize ActionScheduler usages
Linear issue: WOOPMNT-6366 (somehow this issue is not synced) 

See p7bje6-4BB-p2

Using ActionScheduler is not free. Spawning unnecessary actions can affect the site's performance.

Here's a list of improvements from a quick search for `schedule_job` across the code:
- We issue a bunch of actions after an order is processed. All of those could be rolled into one, which will take an order and perform the actions in bulk:
  - ~~`wcpay_update_customer_with_order_data`~~ Removed: see paJDYF-6T1-p2
  - `wcpay_update_saved_payment_method`
  - `wcpay_track_new_order` and `wcpay_track_update_order`
  - In addtion, the above 3 points could be moved to `WC_Payments_Order_Service` 
  - `wcpay_add_fee_breakdown_to_order_notes`
  - In the p2 post above there was also a suggestion of maintaining a list of orders needing attention and running a recurring action processing them in batches. 
  - There was also a suggestion to use the `shutdown` action hook (already used [here](https://github.com/Automattic/woocommerce-payments/blob/1ffd7b713db8588bbc59ad8f63dddf0abd65224d/includes/admin/tracks/tracks-loader.php#L44) and [here](https://github.com/Automattic/woocommerce-payments/blob/1ffd7b713db8588bbc59ad8f63dddf0abd65224d/includes/subscriptions/class-wc-payments-product-service.php#L84)) and avoid the ActionScheduler altogether, which would be preferable
- `wcpay_instant_deposit_reminder` seems to be an action used properly
- Failed webhook processing, we could try to process them all at once or batch them in bulks of, let's say, 5 (an arbitrary number):
  - `wcpay_webhook_process_event` shouldn't be issued per single webhook
- ~~Failed webhook fetching~~ (see [below](https://github.com/Automattic/woocommerce-payments/issues/5279#issuecomment-1357671038))
  - ~~`wcpay_webhook_fetch_events`~~
  - ~~This one's quite tricky as it hooks into the account update action, which it maybe shouldn't. In which case some backend work is also required, where we add a new endpoint.~~
  - ~~Seems to be run every time an account is refreshed, which heavily depends on the logic there, which is unrelated to webhook processing.~~
  - ~~Should be run using a scheduled recurring action instead (NOT a one-off action requeuing itself)~~
  - ~~Feel free to break this out into its own issue/task~~

The above list covers all the async actions from the time of writing, but there might be more in the future. In general:
- We shouldn't issue more than one job if the actions can be batched and processed inside one job
- We should use recurring actions more rather than inventing our own mechanisms using one-off actions

cc @Automattic/gamma

RadoslavGeorgiev:
Partially Fixes #5279. Fixes #5188.

(TBD better)

- [x] Add a custom post type for storing webhooks & related functionality (CRUD).
- [x] Store some (decided by `should_delay_event()`) webhooks in the post type upon their normal arrival.
- [x] Store all polled events in the post type.
- [x] Add webhook processing to the `catch` block of `process_payment` in order to check if there is a successful payment even though there was failure.
- [ ] Add a recurring action to process the events in the post type.

#### Testing instructions

*

-------------------

**Post merge**


includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -163,52 +158,183 @@ private function schedule_fetch_events() {
...
+			return;
 		}
 
-		return set_transient( $this->get_transient_name_for_event_id( $event_data['id'] ), $event_data, DAY_IN_SECONDS );
+		$post_arr = [
+			'post_title'   => $event['id'] . ' (' . $event['type'] . ')',
+			'post_name'    => $event['id'],
+			'post_type'    => self::POST_TYPE,
+			'post_content' => maybe_serialize( $event['data'] ),
+			'menu_order'   => $event['livemode'] ? 1 : 0,
```
RadoslavGeorgiev, line 220: Let's use a taxonomy instead.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -163,52 +158,183 @@ private function schedule_fetch_events() {
...
+
+		switch ( $event['type'] ) {
+			case 'payment_intent.succeeded':
+				$post_arr['post_parent'] = $event['data']['object']['metadata']['order_id'];
+				break;
+		}
+
+		$post_id = wp_insert_post( $post_arr );
+		if ( is_wp_error( $post_id ) || 0 <= $post_id ) {
+			Logger::error( 'Could not store event in CPT. Event ID: ' . $event['id'] );
```
RadoslavGeorgiev, line 234: Add the event data here to make debugging easier if necessary?

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -163,52 +158,183 @@ private function schedule_fetch_events() {
...
+	 * @return bool        Whether to delay the processing of the event.
 	 */
-	private function get_transient_name_for_event_id( string $event_id ): string {
-		// Use md5 to overcome the limit of transient name (172 characters) while Stripe event ID can be up to 255.
-		return 'wcpay_failed_event_' . md5( $event_id );
+	public function should_delay_event( array $event ) {
+		return in_array(
+			$event['type'],
+			[
+				'payment_intent.succeeded',
```
RadoslavGeorgiev, line 196: For SEPA payments, we should not delay the processing.

PR #5410, opened 2023-01-16
Load chunks properly with JS concatenation enabled

ismaeldcom:
With a recent incident report (paJDYF-6DK-p2) we noticed the need to fix chunks loading for sites with JS concatenation enabled. By default, chucks are loaded with a relative URL. JS concatenation will load them from a different path causing a 404 error. So we need to ensure they are loaded from the right path/URL.

- Override `__webpack_public_path__` using a dynamic value from `window.wcpayAssets.url` with the URL of WCPay dist path in the running site.
- Map every webpack entry to prepend the override, to avoid the need of manually importing `public-path.js` on every entry, or use a third party plugin.
- Inject `WCPAY_ASSETS` a new inline script in both WP admin and frontend, to make the plugin dist folder URL available on `window`.
- Use chunks hash as cache busting, keeping the convention of `?ver=[hash]` we already use to load JS scripts from PHP.

##### Prerequisites
- Install [Page Optimize](https://wordpress.org/plugins/page-optimize/) plugin.
- Go to **Settings → Performance** and enable `Concatenate scripts`.
- Edit `client/checkout/utils/fingerprint.js`, comment `FingerprintJS` import, and add the following code on `getFingerprint` first line:
```js
	const FingerprintJS = await import(
		/* webpackChunkName: "fp-agent" */
		'@fingerprintjs/fingerprintjs'
	);
```

##### Without this PR
- Run a webpack build in production mode `NODE_ENV=production node_modules/.bin/webpack build`.
- Add a product to the cart and go to checkout.
- You should see an error notice `Loading chunk N failed...`.

##### With this PR
- Run a webpack build in production mode `NODE_ENV=production node_modules/.bin/webpack build`.
- Add a product to the cart and go to checkout.
- Should work fine.

You can also check on browser inspector network tab, how enabling/disabling scripts concatenation makes `fp-agent` load from a different script.

-------------------

**Post merge**


daquinons (commented):
Thank you for this! :fire: The changes make sense and I've tested both scenarios 
Screenshot 2023-01-18 at 13 44 10

Screenshot 2023-01-18 at 13 46 00

I have a question about this part:
> You can also check on browser inspector network tab, how enabling/disabling scripts concatenation makes fp-agent load from a different script.

I tried to disable the script concatenation an still was loading from the same url 😕 Am I doing what is expected there? I feel like I'm missing something :)

ismaeldcom:
Thanks for the review 🙂 

>I tried to disable the script concatenation an still was loading from the same url 😕 Am I doing what is expected there? I feel like I'm missing something :)

The URL should be the same, because now it works as expected. The difference when you disable the script concatenation is on the network call initiator.

Without concatenation, the initiator is `upe_checkout.js`, one of our build entries:
Screenshot 2023-01-18 at 14 05 09

With concatenation, the initiator is `/_static/` for [Page Optimize](https://wordpress.org/plugins/page-optimize/), as it serves all concatenated scripts from a custom URL. In this case `upe_checkout.js` is included in the generated static file.
Screenshot 2023-01-18 at 14 06 01

daquinons:
> Thanks for the review 🙂
> 
> > I tried to disable the script concatenation an still was loading from the same url 😕 Am I doing what is expected there? I feel like I'm missing something :)
> 
> The URL should be the same, because now it works as expected. The difference when you disable the script concatenation is on the network call initiator.
> 
> Without concatenation, the initiator is `upe_checkout.js`, one of our build entries: Screenshot 2023-01-18 at 14 05 09
> 
> With concatenation, the initiator is `/_static/` for [Page Optimize](https://wordpress.org/plugins/page-optimize/), as it serves all concatenated scripts from a custom URL. In this case `upe_checkout.js` is included in the generated static file. Screenshot 2023-01-18 at 14 06 01

Aha!!! Now I saw it :D I wasn't looking at the initiator 💯

daquinons (approved):
After finding the last detail, I can approve it 🙌 Awesome @ismaeldcom

webpack/shared.js
```diff
@@ -1,39 +1,46 @@
...
+				'./client/multi-currency-analytics/index.js',
+			order: './client/order/index.js',
+			'subscriptions-empty-state':
+				'./client/subscriptions-empty-state/index.js',
+			'subscription-product-onboarding-modal':
+				'./client/subscription-product-onboarding/modal.js',
+			'subscription-product-onboarding-toast':
+				'./client/subscription-product-onboarding/toast.js',
+		},
+		// Override webpack public patch dynamically on every entry.
```
marcinbot: Typo: `patch` should be `path`

marcinbot (approved):
Looks good and the changes make sense. One small typo, but otherwise should be ready to :ship:

webpack/shared.js
ismaeldcom: Good catch, thanks! 🙂 _Addressed in ef06522_

PR #5415, opened 2023-01-17
Increase OTP popup height

Mayisha:
Increased the WooPay OTP popup's height by `50px` to make it visually better and accommodate the contents better in the popup viewport.


bborman22 (approved):
This looked good to me! I tested these out with the other WooPay design updates as well and all looked good.

LGTM!

PR #5416, opened 2023-01-17
Fix WooPay Express Checkout button error message display

brettshumaker:

If the WooPay service is unavailable, this PR adds notices to the appropriate `woocommerce-notices-wrapper` container instead of trying to place them under the email input field as it's not always available. For blocks cart/checkout pages, it adds them to the notices store and they are displayed immediately. For the shortcode cart/checkout and single product pages, it makes an ajax call to retrieve the HTML generated by `wc_add_notice()` and then adds that HTMl to the container. Let me know if there's a better way to achieve this for the non-blocks pages.

#### Testing instructions

* To test this, I was adding the following snippet to the top of `openIframe()` in `client/checkout/platform-checkout/express-button/express-button-iframe.js`. It causes the error notice to be added instead of needing to create a scenario where WooPay is unavailable.
```
showErrorMessage();
return;
```

* Check the single product page to ensure that the error notice is added in the correct location when the WooPay express checkout button is clicked.
* Check the shortcode cart page to ensure that the error notice is added in the correct location when the WooPay express checkout button is clicked.
* Check the shortcode checkout page to ensure that the error notice is added in the correct location when the WooPay express checkout button is clicked.
* Check the block cart page to ensure that the error notice is added in the correct location when the WooPay express checkout button is clicked.
* Check the block checkout page to ensure that the error notice is added in the correct location when the WooPay express checkout button is clicked.

It should look like this:
cart-shortcode checkout-shortcode cart-block checkout-block single-product

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-woopay-express-checkout-button-error-message-display
+++ b/changelog/fix-woopay-express-checkout-button-error-message-display
@@ -1,0 +1,5 @@
+Significance: patch
+Type: update
+Comment: Updates the way error notices are displayed for the WooPay express checkout button. This feature is behind a feature flag and is not yet public.
+
+
--- a/client/checkout/platform-checkout/express-button/express-checkout-iframe.js
+++ b/client/checkout/platform-checkout/express-button/express-checkout-iframe.js
@@ -5,7 +5,7 @@ import { __ } from '@wordpress/i18n';
 import { getConfig } from 'utils/checkout';
 import wcpayTracks from 'tracks';
 
-export const expressCheckoutIframe = async ( api ) => {
+export const expressCheckoutIframe = async ( api, context ) => {
 	let userEmail = '';
 
 	const parentDiv = document.body;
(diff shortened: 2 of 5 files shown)


bborman22:
Just making a note that we had talked briefly about this issue yesterday and discovered that the current test to determine if the page is using WC Blocks wasn't granular enough to catch different versions of the WC Blocks plugin. This was preventing the errors from displaying on the shortcode cart and checkout. The fix seemed to be to look for a `wcBlocks` entry in the `wcSettings` instead as that seemed to be consistent throughout the WC Blocks versions.

brettshumaker:
Noting that I've tested this with the latest WC (7.3) and the latest Blocks (9.4.0) plugins. Works fine except for the block cart page as there's a bug in the 9.4.0 Blocks plugin but it looks like that will be patched before that's put into core WC.

bborman22 (approved):
Works very well for me now! I tested with shortcode and blocks for all scenarios and worked as expected. I tested this out with WC 7.1.1 and WC Blocks 9.2 as well to give some different versions a test and worked as expected.

LGTM!

PR #5423, opened 2023-01-19
Ensure WCPay JS loaded on Checkout page for zero order totals

### LINKED ISSUE #4764: Payment form absent when item with price added to cart via AJAX on a free order
### Describe the bug
When an item with price 0 is added to the cart, it is expected that the payment form does not need to be displayed on the Checkout page. But, when an item with a price is added to the cart via AJAX on the checkout page, the payment form should appear, because it is no longer a 0 priced order.

### To Reproduce
1. Set up a product with a price of 0.
2. Set up two shipping types - Free Shipping and Flat rate at a non-zero cost
3. Add 0 priced item to cart and go to checkout page. Free shipping should be selected by default, so that order total is 0
4. Switch to Flat rate shipping on Checkout page so that order total is non-zero

### Actual behavior
The payment section loads, but there is no credit card form.

### Screenshots
image

### Expected behavior
When order total is 0 on Checkout page load, but is then updated to be non-0, the credit card form should appear and customer should be able to place the order.

### Additional context
p1662469528429629-slack-CGGCLBN58

### LINKED ISSUE #5277: Display CC fields when a 100% coupon code  that is added in cart is removed on checkout page
### Describe the bug

If a 100% coupon code that is originally added in the cart page is removed on the checkout page, no credit card fields are displayed unless you reload the page.

### To Reproduce

1. You should either try to checkout as a guest user, or try a user with no saved payment method on the store
2. Create a 100% coupon
3. Add a product to cart, and use the coupon
4. Proceed to checkout page and remove the coupon
5. Notice that the credit card fields are only displayed when you reload the page
6. Clicking on use a new payment method does not make a difference:

    

### Actual behavior

**When the coupon is originally added on the checkout page, everything works fine**

**When the coupon is first added on the cart page, the CC fields don't show up**

### Expected behavior

If there are no payment method saved, CC fields should always be displayed.

Any

Any

### Additional context

**Merchant reports**: 

5783438-zen

jessy-p:
When an item with price 0 is added to the cart, it is expected that the payment form does not need to be displayed on the Checkout page. But, when the cart is updated via AJAX on the checkout page to be non-zero, the payment form should appear, because it is no longer a 0 priced order.

Fixes #4764
Fixes #5277

- Even when cart is zero, make sure the JS is loaded.
- Do this only on checkout page, and not on blocks page because the Blocks Checkout already loads all the required JS. 

#### Testing instructions
1) Test with steps in #4764 
2) Test with steps in #5277 
3) Test with Checkout Blocks and UPE (all variations of checkout) and on other pages where the checkout form is used e.g. Pay for Order, My Accounts etc to ensure no regression 
4) Can test with any plugins that update order via AJAX on checkout page.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.6.0#ensure-wcpay-js-loaded-on-checkout-page-for-zero-order-totals-pr-5423
- [x] Not applicable - Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.


jessy-p:
Working only for non-UPE payment method, need to fix for UPE and test for block checkout and other checkout combinations.

jessy-p:
Fixed for UPE and blocks too. Ready for review @Automattic/gamma

fashuvo (commented):
Tested this locally. The fix works for Classic checkout and Blocks checkout, but it isn't working for UPE.

fashuvo (approved):
Changes LGTM. Works on UPE, Classic checkout, and Blocks checkout.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -629,6 +629,12 @@ public function register_scripts() {
 		);
 
 		wp_set_script_translations( 'WCPAY_CHECKOUT', 'woocommerce-payments' );
+
+		if ( ! WC()->cart->needs_payment() && is_checkout() && ! has_block( 'woocommerce/checkout' ) ) {
+			WC_Payments::get_gateway()->tokenization_script();
```
htdat, line 634: We may just directly `$this->tokenization_script()` as it's basically the same one but without calling an outside function to get another inside function.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -629,6 +629,12 @@ public function register_scripts() {
 		);
 
 		wp_set_script_translations( 'WCPAY_CHECKOUT', 'woocommerce-payments' );
+
+		if ( ! WC()->cart->needs_payment() && is_checkout() && ! has_block( 'woocommerce/checkout' ) ) {
```
htdat, line 633: Nitpick: moving `is_checkout()` as the first check as it can help terminate this code for most cases, which results in a little bit faster processing in general.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -135,6 +135,11 @@ public function register_scripts() {
 			WC_Payments::get_file_version( 'dist/upe_checkout.js' ),
 			true
 		);
+
+		if ( ! WC()->cart->needs_payment() && is_checkout() && ! has_block( 'woocommerce/checkout' ) ) {
+			WC_Payments::get_gateway()->tokenization_script();
+			WC_Payments::get_wc_payments_checkout()->enqueue_payment_scripts();
+		}
```
htdat, line 142: IMO, this whole block can be moved to a method in `WC_Payment_Gateway_WCPay` and then you can use that method for all of these places. Without that, we do not need to write the same code in three places. We can use this because of the inheritance:  

- UPE_Payment_Gateway extends WC_Payment_Gateway_WCPay
- UPE_Split_Payment_Gateway extends UPE_Payment_Gateway

Another possibility is to just hook this extracted method into the `wp_enqueue_scripts` filter. With that, you just need to write this code one time in the top-level class `WC_Payment_Gateway_WCPay`.

Something like this: https://gist.github.com/htdat/679c224b1c5a761be6082e44695b76ee

htdat (commented):
While reviewing this PR #5652, I did have a look at this issue too. All of these inline comments are more ideas, they're not meant to update/fix the code right away.

PR #5430, opened 2023-01-20
Add logging & order note when WCPay subscription is suspended

### LINKED ISSUE #5429: Add logging and/or order notes when WCPay Subscriptions are suspended (paused/on hold) to help diagnose unexpectedly paused subscriptions
### Description
Some WCPay merchants are reporting that WCPay Subscriptions (Stripe Billing, offsite renewal) are pausing soon after or during renewal process. This is rare, but serious, as the subscriptions are `Active` status but not paid.

We have discovered that affected subscriptions have likely been suspended, similar to merchant or subscriber putting subscription on hold (pausing). On hold is implemented by sending a request with `pause_collection: void` to WCPay Server => Stripe, and the affected subscriptions typically show this request.

https://github.com/Automattic/woocommerce-payments/blob/a79c5a70db6bc23c52e6e248d52f9481c220e457/includes/subscriptions/class-wc-payments-subscription-service.php#L454-L463

So the goal of this issue is to log all code paths that trigger `suspend_subscription()`, and/or add order notes, to help confirm if these suspensions are triggered by an error/interaction on merchant's sites.

### Testing instructions

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'
4. Observe expected behaviour

haszari:
Fixes #5429 

We've had reports that WCPay Subscriptions (Stripe Billing, offsite renewal) are pausing soon after or during renewal process. The pause is triggered by POSTing `pause_collection: void` to Stripe, which is usually sent when [a WCPay Subscription is suspended](https://github.com/Automattic/woocommerce-payments/blob/a79c5a70db6bc23c52e6e248d52f9481c220e457/includes/subscriptions/class-wc-payments-subscription-service.php#L462).

This PR adds logging and order notes to all code paths that trigger `suspend_subscription()`. These breadcrumbs will give us more clues to work out why subscriptions are getting unexpectedly suspended (in rare cases).

I've found 2 scenarios that trigger `suspend_subscription()`:

1. Admin user manually putting subscription on hold.
2. Handling of `invoice_upcoming` webhook. If the `next_payment` date is zero, subscription is suspended. 

Reviewers: 
- Let me know if there are other scenarios that we should add logging to.
- I'm not clear on when scenario 2 might occur, and if it's normal or unexpected.

##### Prerequisites
- Set up a site with WCPay onboarded. 
- Store location is USA so WCPay Subscriptions is available.
- Set up a subscription product.

##### Scenario 1 - admin user suspending subscriptions
- As a customer, purchase a subscription product.
  - You can skip this step if your store already has active subscriptions.
- As an admin, view the subscriptions list.
- Click `Suspend` link under an active subscription.
- View the subscription. 
- Confirm it has an order note indicating what triggered the suspend.

Screenshot 2023-01-23 at 1 49 49 PM

- Go to `WooCommerce > Status > Logs`. 
- View a recent WCPay log file, e.g. select `woocommerce-payments-[date]-[hash].log`, click `View`.
- Scroll or search for a log message about the suspended subscription, for example:

`2023-01-23T00:48:20+00:00 INFO Suspended WCPay Subscription because subscription status changed to on-hold. WC ID: 145 WCPay ID: sub_xxxxxxxxxxxx.`

##### Scenario 2 - subscription renewal with next payment date of `0`
TBC - I don't know how best to test this

-------------------

- [x] Covered with tests (or have a good reason not to test in description ☝️) _I don't think this is important enough functionality to enforce with automated tests, but I'm happy to add tests if reviewers think it's worthwhile_
- [x] Tested on mobile (or does not apply) _not applicable - uses standard existing UI (e.g. WC order notes)_

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.4.0#add-order-notes-and-logging-when-wcpay-subscriptions-are-suspended-5430
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable. _not applicable - not a critical flow_


includes/subscriptions/class-wc-payments-subscription-service.php
```diff
@@ -451,6 +451,25 @@ public function cancel_subscription( WC_Subscription $subscription ) {
...
 
+	/**
+	 * Suspends the WCPay subscription when a WC subscription is put on-hold.
+	 *
+	 * @param WC_Subscription $subscription The WC subscription that was suspended.
+	 *
+	 * @return void
+	 */
+	public function handle_subscription_status_on_hold( WC_Subscription $subscription ) {
+		$this->suspend_subscription( $subscription );
```
haszari, line 469: I've added logging that assumes the subscription is a WCPay Subscription. However, this might not be the case, e.g. a store that has WC Subscriptions active (currently, or in the past).

Reviewer question: should we check that the subscription is a WCPay Subscription before suspending, either here or in `suspend_subscription`?

haszari: Update: I added this as an early-exit precondition - 5b744756591df43eff5573e44604727163aabfde. When I was testing I `Suspend`ed a regular tokenised subscription, and the MIA WCPay ID broke the logging.

So that's a behaviour change, but maybe a good thing.

haszari:
Tagging @james-allan  for a specific review – you've up to speed on the unexpected subs pause issue. Keen for your thoughts on if these logs & order notes will help us hone in on the root cause of this rare problem.

james-allan:
> Handling of invoice_upcoming webhook. If the next_payment date is zero, subscription is suspended.
> I'm not clear on when scenario 2 [above] might occur, and if it's normal or unexpected.

This is a backstop of sorts. If Stripe notify us that they are about to charge an invoice but our subscription record doesn't have a next payment date, we tell Stripe to suspend the subscription to avoid them making that charge. 

This is not expected, but may occur under desynced situations.

includes/subscriptions/class-wc-payments-subscription-service.php
james-allan: The way this worked previously was the `suspend_subscription()` function would call `update_subscription()` and that function would check if the subscription had a wcpay subscription ID. 

https://github.com/Automattic/woocommerce-payments/blob/2c22e38c0997c716909d7a9a367ee1f3363ed488/includes/subscriptions/class-wc-payments-subscription-service.php#L834-L841

This would log entries into the WC Payments logs that could trip people up thinking there was an issue or error, when in fact the subscription just wasn't a WCPay Subscription and shouldn't have gotten to that point. 

So, exiting early when the subscription isn't a wcpay subscription is a good idea. 👍 

---- 

By introducing a new function that is hooked onto `woocommerce_subscription_status_on-hold`, we'll also need to update these instances where we unhook these callbacks where we don't intend to notify Stripe of a suspension. 

This is the only time we currently do this: 
https://github.com/Automattic/woocommerce-payments/blob/68856b78e0b41f4c4c4af66b0182f656d06dc138/includes/subscriptions/class-wc-payments-subscriptions-event-handler.php#L127-L129

james-allan: Rather than have to do all this updating though, we could update `suspend_subscription()` instead and just have it exit early and add the new order notes and logs.

james-allan (changes requested):
I left a comment in line about the `handle_subscription_status_on_hold()`. 

I also had a question though. What do you think if we added a stack trace to the log entry added when a subscription is put on-hold?

One of the issues with troubleshooting these issues is that we know Stripe was told to pause the subscription, we just don't know why.

These logs as they are now would confirm that this site did send the request and confirm the timestamp that it happened, however we wouldn't understand the flow that lead to it.

What do you think?

includes/subscriptions/class-wc-payments-subscription-service.php
haszari: Thanks – that makes sense. I didn't realise the implications of the new callback; the idea was that the logging happens outside `suspend_subscription`, nearer the "action" that triggered suspension, rather than generically inside `suspend`.

However, if we add a call stack, then the generic option would give us the same info (and more), and would be lower impact on the code base too.

haszari: Just realised we won't be able to add a specific order note for status change (on-hold => suspend). Maybe that's ok, I was hoping to add individual notes for the two scenarios.

haszari: Went with the decoupled status handler, so we get the actions logged and shown in notes: e5dce3f8a4c7ad8042052af2b8df38d69fb230d1

@james-allan  let me know what you think, happy to go with generic approach if that's better overall.

james-allan: This works for me.

includes/subscriptions/class-wc-payments-subscription-service.php
```diff
@@ -452,13 +452,59 @@ public function cancel_subscription( WC_Subscription $subscription ) {
...
+
+		// Add an order note as a visible record of suspend.
+		$subscription->add_order_note( __( 'Suspended WCPay Subscription because subscription status changed to on-hold.', 'woocommerce-payments' ) );
+
+		// Log that the subscription was suspended.
+		// Include a brief stack trace to help determine where status change originated.
+		// For example, admin user action, or a code interaction with customizations.
+		$e     = new Exception();
+		$trace = $e->getTraceAsString();
+		Logger::log(
```
haszari, line 479: Might be cool to add this as a feature of `Logger` class, so anyone logging can include/append a stack trace in a standard format.

james-allan: :nod:

james-allan (approved):
This looks good to me. 

I tested the following flows:

1. Putting subscriptions on-hold from admin row actions.
2. Putting the susbcription on-hold progamatically.
3. Suspending the subscription from the customer's my account page isn't allow on WCPay subscriptions. 

In both instances I confirmed the order note was added and looking in WooCommerce > Status > Logs I can see the entry with a stack trace. 

Screen Shot 2023-01-26 at 2 15 16 pm

Screen Shot 2023-01-26 at 2 15 48 pm

I also confirmed that if you call `WC_Payments_Subscriptions::get_event_handler()->handle_invoice_upcoming( $event_body );` with a mocked event body and mocked subscription without a next payment date, I confirmed the following note is added to the subscription: 

Screen Shot 2023-01-26 at 1 56 47 pm

haszari:
Thanks @james-allan !

Looking at the `invoice.upcoming` note, I realised it was missing key info – that this is happening in webhook handler. The `next_payment` date being 0 is only part of the story.

Screen Shot 2023-01-26 at 1 56 47 pm

Added more info in 77a99c70ff5deb9dbdb48a00a410d9086f9e8fbb 

> Suspended WCPay Subscription **in invoice.upcoming webhook handler** because subscription next_payment date is 0.

PR #5432, opened 2023-01-20
Use order currency API when overriding currency for a subscription switch to support HPOS

### LINKED ISSUE #5424: [HPOS] Replace call to `get_post_meta` in `MultiCurrency\Compatibility\WooCommerceSubscriptions` with HPOS-compatible order CRUD method.
### Describe the bug

Within the `WCPay\MultiCurrency\Compatibility\WooCommerceSubscriptions` class, the method [`override_selected_currency()`](https://github.com/Automattic/woocommerce-payments/blob/2c22e38c0997c716909d7a9a367ee1f3363ed488/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L143) calls the `get_post_meta` function that will not work correctly when a store has [HPOS](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#what-is-high-performance-order-storage-hpos) enabled.

This should be replaced with a [WC_Order CRUD method](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#what-is-high-performance-order-storage-hpos) to work with HPOS.

https://github.com/Automattic/woocommerce-payments/blob/2c22e38c0997c716909d7a9a367ee1f3363ed488/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L156-L158

### Additional details

> **Note**
> I've initially tried a quick replacement with `wc_get_order()`/`get_subscription()` here, but it results in an infinite loop.

This was identified while auditing WCPay subscription-related code for HPOS incompatible functions and hooks using the following regexp:

```
wpdb|get_post|get_post_field|get_post_status|get_post_type|get_post_type_object|get_posts|metadata_exists|get_post_meta|get_metadata|get_metadata_raw|get_metadata_default|get_metadata_by_mid|wp_insert_post|add_metadata|add_post_meta|wp_update_post|update_post_meta|update_metadata|update_metadata_by_mid|delete_metadata|delete_post_meta|delete_metadata_by_mid|delete_post_meta_by_key|wp_delete_post|wp_trash_post|wp_untrash_post|wp_transition_post_status|clean_post_cache|update_post_caches|update_postmeta_cache|post_exists|wp_count_post|save_post
```

See related PR #4228 for details on other subscription-related currency overrides within this function.

Jinksi:
Fixes #5424

This PR fixes overriding the currency when a customer begins a subscription switch and [HPOS](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#how-to-enable-hpos) is enabled.

It replaces a call to `get_post_meta()` (which is not HPOS compatible) with `WC_Order->get_currency()`.

See related PR #4228 for details on other subscription-related currency overrides within this function.

#### Testing instructions

1. [Set up multi currency.](https://woocommerce.com/document/payments/currencies/multi-currency-setup/) Needs to be activated in Woo settings and WCPay, and a currency switcher displayed for customer on store front end.
3. Set up 2 or more currencies (e.g. Euro and USD). Referred to below as **currency A** and **currency B**.
4. Enable [HPOS](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#how-to-enable-hpos) with syncing **disabled**.
5. Create a variable subscription product with 2 or more variations.
6. Purchase the variable subscription using **currency A**.
7. Revisit the shop page and switch to **currency B** with the currency selector.
8. Navigate to **My Account → Subscriptions** and select the purchased subscription.
9. Click the "Upgrade or Downgrade" button to initiate a subscription switch.
10. On `trunk`/`develop`, the subscription's currency will incorrectly show as the selected currency (**currency B**) rather than the currency the subscription was purchased with.
11. On this PR branch, the subscription's currency will correctly show as the currency it was originally purchased with: **currency A**.
12. Repeat from step 5 with HPOS disabled 

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : QA Testing N/A since this is specific to WooCommerce Subscriptions plugin feature.
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable. **No changes to critical flows**.
- [x] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable. Fix for **HPOS** stores only, no announcement required.


includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -152,9 +152,14 @@ public function override_selected_currency( $return ) {
 			return $order ? $order->get_currency() : $return;
 		}
 
-		$switch_id = $this->get_subscription_switch_id_from_superglobal();
-		if ( $switch_id ) {
-			return get_post_meta( $switch_id, '_order_currency', true );
+		// Ensure the subscription for a $switch_id is fetched only once per page load,
+		// avoiding an infinite loop that can occur on the product page when `get_subscription()` is used.
+		static $currency_fetched_for_switch_id = null;
```
Jinksi: Using a [`static` variable](https://www.php.net/manual/en/language.variables.scope.php#language.variables.scope.static) here to cache this value for subsequent calls to the same function in a single page load, preventing an infinite loop that occurs otherwise.

This allows subsequent calls to ignore the `if` statement below, skipping the call to `get_subscription` that causes the infinite loop.

tests/unit/multi-currency/compatibility/test-class-woocommerce-subscriptions.php
```diff
@@ -286,13 +286,20 @@ public function test_override_selected_currency_return_currency_code_for_switch_
...
-		$order = WC_Helper_Order::create_order();
-		$order->set_currency( 'JPY' );
-		$order->save();
+		// Note we're using a WC_Order as a stand-in for a true WC_Subscription.
+		$mock_subscription = WC_Helper_Order::create_order();
+		$mock_subscription->set_currency( 'JPY' );
+		$mock_subscription->save();
+
+		// Mock wcs_get_subscription to return our mock subscription.
+		WC_Subscriptions::set_wcs_get_subscription(
```
Jinksi, line 325: Using a mock `WC_Subscription` rather than `WC_Order`, as the `override_selected_currency` expects. This is consistent with neighbouring tests.

Jinksi:
e2e tests are failing with this seemingly-unrelated checkout error.

```
WooCommerce Blocks > Checkout failures › should throw an error that the card was declined due to invalid 3DS card

    Error: failed to find element matching selector "div.wc-block-components-notices > div > div.components-notice__content"
```

It looks like this is failing for [all PRs at the moment](https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-pull-request.yml).

shendy-a8c:
Is this PR ready for review @Jinksi ? Asking because even though you have requested for a review, there is no `needs review` label. It looks ready for review though. I might give it a review.

mattallan (approved):
Nice fix here @Jinksi! I like the approach and don't have any feedback based on the code changes. I have confirmed the issue on `trunk` with HPOs enabled, and confirmed it's fixed by this PR with both HPOS enabled and disabled.

Happy to approve this one.

I forgot to assign myself this issue in GH yesterday and I see @shendy-a8c has started looking into this one as well, so I'm happy for an extra set of eyes on these changes.

Jinksi:
> Is this PR ready for review @Jinksi ?

Thanks @shendy-a8c, I missed the `needs review` label (I put it on the issue), but I've added it now.

shendy-a8c:
Code diff looks good. However, even though the currency symbol is fixed, the amount is not. Do you see the same problem?

https://user-images.githubusercontent.com/73803630/214456269-ca5400b9-d2d8-46d9-995d-5fadcd95354c.mov

Jinksi:
Nice catch @shendy-a8c! I see the same issue with HPOS enabled and sync disabled.

I'll see if I can find the source.

Jinksi:
> Code diff looks good. However, even though the currency symbol is fixed, the amount is not. Do you see the same problem?

@shendy-a8c The issue was that the previous attempt from f202e311c58931ba146f2afe01fda435be270e41 returned the `$return` value on subsequent calls after the switch subscription currency was initially fetched and returned.

I've fixed this in bfc6be6, where the switch subscription currency is now cached and returned on subsequent calls. 

IMO it is a _bit_ clumsy. Keen to see if you have any different ideas to solve this one.

Jinksi:
Need to figure out why GH checks are showing 2 failing PHP unit tests, although these pass in my local env.

All of the failing checks are using WC beta, so may be unrelated to this PR.

```
1) WC_Payments_Order_Service_Test::test_mark_payment_capture_expired
Failed asserting that 'Order status changed from Pending payment to Canceled.' contains "Pending payment to Cancelled".

/home/runner/work/woocommerce-payments/woocommerce-payments/tests/unit/test-class-wc-payments-order-service.php:409

2) WC_Payments_Order_Service_Test::test_mark_payment_capture_cancelled
Failed asserting that 'Order status changed from Pending payment to Canceled.' contains "Pending payment to Cancelled".
```

shendy-a8c:
I can see that the problem is now resolved after your latest commits but yes I agree it's not too elegant solution.

How about if we remove the callback and re-add before and after we fetch the switch subscription?
I tried it [here](https://github.com/Automattic/woocommerce-payments/pull/5455) (based off `develop` but got a lot of failing tests) and [here](https://github.com/Automattic/woocommerce-payments/pull/5459) (based off your branch).
I also see this approach is pretty common in subscription-core, eg https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wc-product-subscription-variation.php#L164-L168.

About WC beta failing tests, I don't think it is because of your changes. I [made a small change](https://github.com/Automattic/woocommerce-payments/pull/5457) and the same tests fail.
It's now fixed by https://github.com/Automattic/woocommerce-payments/pull/5458.

shendy-a8c:
Also, I wonder why the same infinite recursive problem does not happen to the other `$this->get_subscription()` calls in that `override_selected_currency()` function. Should we also remove and re-add the hook callback before and after all `$this->get_subscription()` calls?

Jinksi:
It looks like the authors of #5404 also (just) ran into this issue, which was resolved by adding a new private class variable: `$running_override_selected_currency_filters`.

https://github.com/Automattic/woocommerce-payments/blob/27ff73a17f799f865526e6faad388155de0cfe6c/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L151-L153

https://github.com/Automattic/woocommerce-payments/blob/27ff73a17f799f865526e6faad388155de0cfe6c/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L161-L169

It is a similar approach to the `static` function variable.

@shendy-a8c I have used this approach in e7284fa, and it looks like it is working well.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -168,9 +168,14 @@ public function override_selected_currency( $return ) {
 			}
 		}
 
+		// The running_override_selected_currency_filters property is used to avoid an infinite loop
+		// that can occur on the product page when `get_subscription()` is used.
 		$switch_id = $this->get_subscription_switch_id_from_superglobal();
 		if ( $switch_id ) {
-			return get_post_meta( $switch_id, '_order_currency', true );
+			$this->running_override_selected_currency_filters = true;
```
Jinksi, line 175: Using new `$this->running_override_selected_currency_filters` introduced in #5404 to avoid infinite loops.

When `true`, the function will return early:

https://github.com/Automattic/woocommerce-payments/blob/e7284fafc9a3e41bfe340b0a4abb3c6727124c5d/includes/multi-currency/Compatibility/WooCommerceSubscriptions.php#L151-L153

haszari: I wonder if there's a clearer name for that class variable, to make it clear what it does - e.g. something like `inhibit_override_selected_currency` or `prevent_override_selected_currency_infinite_loop`.

Jinksi: The current naming makes sense to me, although I might suggest adding the `is_` prefix to indicate it is a bool ([along the lines of this article](https://michaelzanggl.com/articles/tips-on-naming-boolean-variables/)).

In my mind, the method's "running" state is what the variable holds, whereas the statement within the method handles the "inhibiting".

FWIW I didn't name this variable; it was introduced in #5404.

shendy-a8c (approved):
Looks good and tested well.

ps: when testing this PR, I notice [a bug](https://github.com/Automattic/woocommerce-payments/issues/5488) that isn't related to this PR.

https://user-images.githubusercontent.com/73803630/215522440-fa388e48-284e-4345-8a36-2625e102b06c.mov

PR #5438, opened 2023-01-20
Fix WooPay express checkout button rendering on cart shortcode

brettshumaker:
This PR fixes an issue with the WooPay express checkout button on the cart shortcode page. If the cart totals were updated, the button would not be re-rendered after the updated cart was shown. It renders the button during the `load` event on `window` and then uses the `updated_cart_totals` event to re-render.

#### Testing instructions

* Make sure the WooPay express button is enabled and is set to show on the cart page.
* Add a product to the cart and view cart.
* Update the item quantity or update the shipping method to make sure the button is re-rendered.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-woopay-express-checkout-button-on-shortcode-cart
+++ b/changelog/fix-woopay-express-checkout-button-on-shortcode-cart
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fix
+Comment: This fix is in relation to the WooPay express checkout button which is currently behind a feature flag and is not public.
+
+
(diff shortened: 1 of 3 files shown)


Mayisha (approved):
This is working well for me and the WooPay button is re-rendering now after cart updates. :tada: 

I am not a fan of mixing up JQuery with React but unfortunately, I can't think of any other solutions here. I tried to solve a similar case before when I found that the events dispatched with JQuery `trigger` is not caught by the regular React/JS event listeners.

:shipit: as things are working well.

PR #5446, opened 2023-01-24
Skip running blocks E2E tests

deepakpathania:
Fixes #5447 

#### Changes proposed in this Pull Request

Currently, the block E2E tests are failing because of a mismatch in error messages expected in tests and the ones coming from Stripe. There was [an issue](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/1633) created for the same in the past but it was closed without any progress and it has resurfaced. 
This is a stop gap solution to skip the blocks tests while we investigate the issue.

#### Testing instructions

* Ensure there are no blocks tests in the PR checks that run for this PR.
* Ensure there are no blocks checks in E2E - all pipeline.
    - Go to Actions -> E2E - all -> Run workflow -> Select this branch and notice the generated matrix output.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #5448, opened 2023-01-24
Add hard-coded card errors

RadoslavGeorgiev:

Add hard-coded card errors through a new `Card_Error` exception. This way messages will be static, and E2E tests will not fail.

Since the first commit of the PR fixed E2E tests, this PR also reverts https://github.com/Automattic/woocommerce-payments/pull/5446.

#### Testing instructions

1. Make sure all tests are passing.
2. Use cards, which will be declined from https://stripe.com/docs/testing#declined-payments to make sure you're seeing the right error message.
    - While doing this, I'd recommend adding a breakpoint on [this line](https://github.com/Automattic/woocommerce-payments/blob/fix/inconsistent-errors/includes/exceptions/class-card-error.php#L65) and comparing the API error message to the one, which is being returned.

-------------------

### DIFF
--- a/.github/workflows/e2e-pull-request.yml
+++ b/.github/workflows/e2e-pull-request.yml
@@ -30,8 +30,11 @@ jobs:
     strategy:
       fail-fast:     false
       matrix:
-        test_groups:   [ 'wcpay', 'subscriptions' ] # [TODO] Unskip blocks tests after investigating constant failures.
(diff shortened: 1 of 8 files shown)


RadoslavGeorgiev:
This PR was meant as a fix of an issue, which has since been fixed by other means. I'm closing it.

PR #5449, opened 2023-01-24
Skip email step when email field has a value.

Mayisha:
Checking the email input field on the merchant checkout page and sending that with the iframe src if the field has a value. This skips the email step in WooPay express checkout flow.

#### Testing instructions
- Enable WooPay express checkout on the merchant site.
- Go to the merchant checkout page after adding some products to the cart.
- Keep the email field on the merchant checkout page empty and click on the express checkout button.
- This should take you to the email input step of WooPay.
- Add a valid email on the email field on the merchant checkout page empty and click on the express checkout button.
- This should skip the email input step of WooPay and take you to the next page.
- Check the express checkout button on the cart and individual product pages. The express checkout button should always open the email input step of WooPay.
- Now disable WooPay express checkout on the merchant site.
- Ensure that the email triggered popup works as before and functionally nothing has changed there.

### DIFF
--- a/changelog/task-1406-skip-email-step
+++ b/changelog/task-1406-skip-email-step
@@ -1,0 +1,4 @@
+Significance: minor
+Type: add
+
+Skipping the email input step for WooPay express checkout flow when the email input field has already a value. The WooPay express checkout feature is behind a feature flag currently.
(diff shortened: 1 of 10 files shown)


brettshumaker (approved):
Great job. This all works as expected and I couldn't find a way to make it fail on shortcode or blocks checkout. 👍

PR #5451, opened 2023-01-24
Express Checkout separator - honor express checkout button settings

mdmoore:

This is a bit of an addendum to #5399. Shortly after merging #5399, @timur27 and I noticed that the button location/page settings weren't being honored. In some scenarios, the `- or -` separator was displayed even when no button was displayed. This PR corrects that by utilizing the same `should_show_*` checks used by the buttons. 

#### Testing instructions

5. Using the settings found for each express checkout button under its **Customize** option, disable the button on some pages and check that settings are being honored.

### DIFF
--- a/includes/class-wc-payments-stripe-link-button-handler.php
+++ b/includes/class-wc-payments-stripe-link-button-handler.php
@@ -63,14 +63,19 @@ class WC_Payments_Stripe_Link_Button_Handler {
 	}
 
 	/**
+	 * Checks if this is the Pay for Order page.
+	 *
+	 * @return boolean
+	 */
+	public function is_pay_for_order_page() {
+		return is_checkout() && isset( $_GET['pay_for_order'] ); // phpcs:ignore WordPress.Security.NonceVerification
+	}
+
+	/**
 	 * Display the payment request button.
 	 */
 	public function display_button_html() {
-		if ( ! WC_Payments_Features::is_link_enabled() ) {
-			return;
-		}
-
-		if ( ! $this->is_checkout() ) {
+		if ( ! $this->should_show_link_button() ) {
 			return;
 		}
 
@@ -81,6 +86,23 @@ class WC_Payments_Stripe_Link_Button_Handler {
 			</div>
 		</div>
 		<?php
+	}
+
+	/**
+	 * Indicates whether the Link express checkout button should be displayed.
+	 *
+	 * @return bool
+	 */
+	public function should_show_link_button() {
+		if ( ! WC_Payments_Features::is_link_enabled() ) {
+			return false;
+		}
+
+		if ( ! $this->is_checkout() || $this->is_pay_for_order_page() ) {
+			return false;
+		}
+
+		return true;
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


includes/class-wc-payments-stripe-link-button-handler.php
```diff
@@ -83,6 +88,23 @@ public function display_button_html() {
...
+	 * Indicates whether the Link express checkout button should be displayed.
+	 *
+	 * @return bool
+	 */
+	public function should_show_link_button() {
+		if ( ! WC_Payments_Features::is_link_enabled() ) {
+			return false;
+		}
+
+		if ( ! $this->is_checkout() || $this->is_pay_for_order_page() ) {
```
timur27, line 101: Thanks for a great job while adding this additional statement, as it absolutely makes sense since we didn't implement a Link on the Pay for Order page. 

Question for myself to extend my knowledge: do you happen to know if we plan to support Link on the Pay for an Order page in the future? I guess the `$this->is_pay_for_order_page()` part comes from some acceptance criteria, but I'm unsure where to find these.

includes/class-wc-payments.php
```diff
@@ -467,12 +467,13 @@ public static function init() {
 		}
 
 		if ( WC_Payments_Features::is_express_checkout_enabled() ) {
-			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
-			add_action( 'woocommerce_proceed_to_checkout', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
-			add_action( 'woocommerce_checkout_before_customer_details', [ __CLASS__, 'display_express_checkout_button_separator_html' ], -1 );
+			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'should_express_checkout_separator_be_displayed' ], -1 );
```
timur27: Would it make sense if we change the method name to 

```suggestion
			add_action( 'woocommerce_after_add_to_cart_quantity', [ __CLASS__, 'display_express_checkout_separator_if_necessary' ], -1 );
```

or similar? Currently, the method name suggests that the return value is of type `boolean`, whereas it's not true.

timur27 (changes requested):
Awesome, huge thanks for improving this area, Mike! From the functional point of view, I tested as many combinations as possible, and your fix works as expected! 🚀

I've left one question and one cosmetic suggestion, which should be easily resolvable, and I'm happy to approve right after! Thanks! 

Another thing, as we discussed a few days ago - it's worth refactoring this part of the app to encapsulate both button and separator rendering in one place. I'll make sure that we have an issue for this created.

timur27:
The follow-up issue to refactor the solution has been created https://github.com/Automattic/woocommerce-payments/issues/5472. @mdmoore Feel free to add anything to the description/labels if you find it necessary.

includes/class-wc-payments.php
mdmoore: Yes, this method name is much better. Thanks for the suggestion!

includes/class-wc-payments-stripe-link-button-handler.php
mdmoore: I'm not aware of any plans to add support for Link or WooPay express buttons on the Pay for Order page. My assumption is that support just hasn't been added yet since these buttons are new additions.

timur27 (approved):
Works as expected. Thanks, Mike! 🚢

PR #5454, opened 2023-01-25
Disable WooPay for unavailable countries

alefesouza:

This PR allows WooPay to be enabled via server side based on the user country.

#### Testing instructions

* Follow 1598-gh-Automattic/woopay testing instructions.

-------------------

**Post merge**


leonardola (approved):
This side of the changes looks good

leonardola:
On a side note. Will we block WooPay for merchants in other countries too? If so then we'll need to check if the merchant site address is in the US. Though I think this check already happens before the merchant can even activate WooPay. Worth checking that out.

includes/class-wc-payments-platform-checkout-button-handler.php
```diff
@@ -457,6 +458,12 @@ public function should_show_platform_checkout_button() {
 			return false;
 		}
 
+		// Check if WooPay is available in the user country.
+		$platform_checkout_utilities = new Platform_Checkout_Utilities();
+		if ( ! $platform_checkout_utilities->is_country_available( $this->gateway ) ) {
```
leonardola, line 463: If the merchant removes this check the option to create the WooPay account will show up. Additional checks must happen on the WooPay side when creating the new account. Also it's now possible to create and account directly in WooPay by accessing `/account` with a non existing account. So we should check it there too.

leonardola (approved):
Code looks and works good. But we should add other checks in the WooPay side.

leonardola (approved):
Looks good now. Thanks for the fixes

PR #5466, opened 2023-01-26
Re-enable email flow with express button

Mayisha:
- Reverted the changes made in https://github.com/Automattic/woocommerce-payments/pull/5364. In that PR we disabled the email triggered iframe if express checkout is enabled. After the revert, we will now always have email triggered iframe and have the express button if features flag is enabled.
- Added a check to see if there is already an OTP iframe open on the screen to make sure either the express triggered or the email triggered iframe is visible on the screen, not multiple at a time.

#### Testing instructions
- Disable WooPay express checkout. Go to the checkout page and make sure that using an existing WooPay email triggers the OTP iframe. (both on shortcode and blocks checkout)
- Enable WooPay express checkout. Go to the checkout page and make sure that using an existing WooPay email triggers the OTP iframe. (both on shortcode and blocks checkout)
- Click the express checkout button on the checkout page and make sure it opens the OTP iframe. (both on shortcode and blocks checkout)
- Use an existing WooPay email in the email field. While the email check is going on making the loader visible, quickly click the WooPay express checkout button. Wait for a while and make sure that there is only one OTP modal on the screen.
- Make sure that the express checkout button works on the shortcode cart, blocks cart and individual product page.


bborman22 (approved):
These changes look good to me and tested well! Also nice to get rid of the somewhat duplicated code for handling the email input. 

LGTM!

PR #5471, opened 2023-01-27
Add support phone and email in the admin settings page

### LINKED ISSUE #3981: Add Feature for Merchants To Update Their Support Phone Number
### Description

Stripe provides a call to update `business_profile.support_phone` in their [docs](https://stripe.com/docs/api/accounts/update#update_account-business_profile-support_phone). WooCommerce Payments doesn't include a field to change this, so merchants have no way to update this number. We update `business_profile.url` with the Customer bank statement field, and a field should be added here so merchants can update their support phone also. 

Merchants cannot change this by managing their accounts in Stripe. Changing the “Management and Ownership” phone and the main “Stripe Express account” phone does not update the Support Details phone number. 

### Acceptance criteria

* The field should be with Customer bank statement, as the two work together 
* It should update the connected account's support details, to provide a way for merchants to change what appears to their customers. 

### Designs

Should be added under the existing `Customer bank statement` field. 
Markup 2022-03-17 at 11 28 35

### Testing instructions

1. Go to Payments > Settings
2. Scroll down to Transactions and deposits
3. Enter the new phone number for the store's support 
4. Click Manage in Stripe, then Account
5. Click the store's settings under Professional Details 
6. Check that the phone has changed in Support Details.

### Dev notes

Stripe API docs for updating connected accounts: https://stripe.com/docs/api/accounts/update

### Additional context

Reported in 4868290-zen 
Support address and email may be helpful to add also, it looks like these are intended to be printed on the In Person Payments Receipt Template.

htdat:
Fixes #3981 

- Extract two new components `SupportEmailInput` and `SupportPhoneInput` from `client/card-readers/settings/sections/contacts-details.js`. 
- With that, we can use them for both `Card Readers` setting page and WCPay general settings pages.
- Add tests for the new components.

| Before      | After |
| ----------- | ----------- |
| settings-before  | settings-after-with-arrow  |

#### Testing instructions

- Checkout this PR, and run `npm start`.
- Visit wp-admin > Payments > Settings.
- See two new input fields.
- Edit and update them.

If possible, try to test with `Card Readers` settings page:

- In your local server, return `true` for `has_card_readers_available` flag/method, like this 2ef2d-pb
- Visit wp-admin > Payments > `Card Readers` > Receipt Details.
- Edit and update the same fields before and after this PR.

#### Note 1 

GitHub action throws a SOS warning about increasing sizes of JS and CSS for the admin settings page. I think this is fine as it brings a nice UI and validation for the phone input. Under the hood, it uses the `PhoneNumberInput` component, which is [based on 'intl-tel-input' package](https://github.com/Automattic/woocommerce-payments/blob/ef6339ff5a79ad8e26bf6f9c4f2ab8445b913310/client/settings/phone-input/index.tsx#L6-L6).

#### Note 2

> Full Stripe accounts allow the phone to be toggled off, so it would be great if WooCommerce Payments can offer that also.
> - From this comment https://github.com/Automattic/woocommerce-payments/issues/3981#issuecomment-1072658960

With the connected account, there is no option for this. Stripe always requires this number and here is the most detail I can get when trying to update this number via `curl`, Stripe Dashboard, and via API call through local server: 

Markup on 2023-01-31 at 16:00:16

-------------------

**Post merge**


htdat:
Ready to review! 

@elazzabi - I've added you as an optional reviewer as you worked on a relevant PR #4355

elazzabi (approved):
The code looks good. Thank you for working on this!

kalessil:
Out of curiosity, is there an opportunity to use TypeScript for new files?

Context: paJDYF-6MV-p2

htdat:
> Out of curiosity, is there an opportunity to use TypeScript for new files?

@kalessil - I tried to use TypeScript but the new files use some non-TypeScript functions such as `useAccountBusinessSupportEmail`, [useGetSavingError](https://github.com/Automattic/woocommerce-payments/blob/0c2ba6f94f391c0d20c53281386c97bd4760603a/client/data/settings/hooks.js#L479-L479) so that it requires more effort in other files to have the new files fully TypeScript-supported.

kalessil:
> @kalessil - I tried to use TypeScript but the new files use some non-TypeScript functions such as useAccountBusinessSupportEmail, [useGetSavingError](https://github.com/Automattic/woocommerce-payments/blob/0c2ba6f94f391c0d20c53281386c97bd4760603a/client/data/settings/hooks.js#L479-L479) so that it requires more effort in other files to have the new files fully TypeScript-supported.

Gotcha, thanks for sharing the details!

client/settings/support-phone-input/index.js
```diff
@@ -0,0 +1,73 @@
...
+		supportPhoneError = __(
+			'Support phone number cannot be empty, please specify.',
+			'woocommerce-payments'
+		);
+	}
+	const [ isPhoneValid, setPhoneValidity ] = useState( true );
+
+	if ( ! isPhoneValid ) {
+		supportPhoneError = __(
+			'Please enter a valid phone number.',
```
RadoslavGeorgiev: Not sure to which line to add this message, but as soon as I opened the WCPay settings page, I saw errors for those fields:

image

Does this mean that every WCPay merchant who updates the plugin will suddenly see errors like this? It did not prevent me from saving the page, and I'm not sure whether that's better because I might have wanted to make a quick change, or it's worse because they're required?

changelog/add-3981-support-phone-number
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+New support phone and email in the admin settings page.
```
RadoslavGeorgiev: Maybe change this to `New support phone and email settings`?

RadoslavGeorgiev (commented):
Great job @htdat , code looks clean and excellent, and also tests well! 👏 

Not marking this as a request for changes, as I'm not certain they're needed, but I have a couple things to discuss.

One weird issue I encountered was not being able to save settings at all initially. After digging a bit, it turned out that Stripe rejects my valid phone number as invalid. Afterwards the server return an exception, however the exception/error is not bubbled up from `WC_REST_Payments_Settings_Controller::update_account`, and I see a success pill, stating settings were saved correctly. This is an issue, but not one for this PR.

client/settings/support-phone-input/index.js
htdat: This is a very valid point. Thanks for pointing it out.

> Does this mean that every WCPay merchant who updates the plugin will suddenly see errors like this? 

You're right. We should avoid this. The logic around here is a bit tricky as Stripe does not require these fields when [creating](https://stripe.com/docs/api/accounts/create#create_account-business_profile-support_phone) or [updating](https://stripe.com/docs/api/accounts/update#update_account-business_profile-support_phone) an account, but also does not allow to unset them when they have already had some values (see my `Note 2` in the PR description).

I've added more logics to satisfy that fact from Stripe: https://github.com/Automattic/woocommerce-payments/pull/5471/commits/de42aafea658c03edb2b37dd37c6bee623ec80b9

- If email or support field has been empty, there will be no red notice, and merchants can submit these fields with empty values normally. 
- In case, these fields have some values, when merchants try to make them empty, red notice will display. 

It looks like the following: 

Markup 2023-02-02 at 18 36 36


> It did not prevent me from saving the page... 

I've also added more logic to disable the `Save Settings` button if these two fields are invalid. https://github.com/Automattic/woocommerce-payments/pull/5471/commits/4ff09b54af1a2d4e4a60a65356959d3694f33fae

changelog/add-3981-support-phone-number
htdat: I slightly update it https://github.com/Automattic/woocommerce-payments/pull/5471/commits/b3adffb30463c52d069e2adb3508feb19e83298a 

Actually, these fields have been already in the `Card Readers` page so I want to clarify them in the admin settings page.

htdat:
> Afterwards the server return an exception, however the exception/error is not bubbled up from WC_REST_Payments_Settings_Controller::update_account, and I see a success pill, stating settings were saved correctly. This is an issue, but not one for this PR.

I've debugged a bit and it seems that `WC_Payment_Gateway_WCPay::update_account` use `add_error` to display admin error notices in the old way rather than expose the error message from Stripe API to the rest request. 

I will debug a bit more and add a new issue for this.

--- 

@RadoslavGeorgiev - I've addressed other concerns in your inline comments too. Ready to review again.

client/settings/transactions/index.js
```diff
@@ -80,6 +84,9 @@ const Transactions = () => {
 					<span className="input-help-text" aria-hidden="true">
 						{ `${ accountStatementDescriptor.length } / ${ ACCOUNT_STATEMENT_MAX_LENGTH }` }
 					</span>
+
+					<SupportEmailInput onErrorMessage={ setSaveDisabled } />
+					<SupportPhoneInput onErrorMessage={ setSaveDisabled } />
```
RadoslavGeorgiev: The state of both inputs needs to be stored separately in order to determine whether saving is disabled or not. Right now:

1. I opened the page with valid values in both fields.
2. Deleted the email ➡️  The save button got disabled.
3. Deleted the phone number ➡️ The save button remain disabled.
4. Started typing an email (which is not validated) ➡️ The save button immediately got enabled, even though the phone number remains empty.

client/settings/support-phone-input/index.js
RadoslavGeorgiev: Nice! Thank you for addressing both concerns at once! :)

changelog/add-3981-support-phone-number
RadoslavGeorgiev: I understand, the message looks great now! 👌

RadoslavGeorgiev (changes requested):
> I will debug a bit more and add a new issue for this.

Thanks for digging into this! I'm not sure if it's applicable in contexts outside of the new fields, but in any case it would be nice to have another issue to keep track of it! 🤝 

---

The rest of the PR looks excellent, there's just one detail that remains to be fixed :)

client/settings/transactions/index.js
htdat: Another great catch! Fixed in this commit https://github.com/Automattic/woocommerce-payments/pull/5471/commits/a6a3b894ac75bdb5c9a76b5c36fcd97862d27c52

I've also updated Card Reader settings page too as it has the same issue. https://github.com/Automattic/woocommerce-payments/pull/5471/commits/404fde21f38f129b3db6226a442168809e38020a

htdat:
> The rest of the PR looks excellent, there's just one detail that remains to be fixed :)

@RadoslavGeorgiev - ready for review again.

htdat:
> I will debug a bit more and add a new issue for this.

Added this issue https://github.com/Automattic/woocommerce-payments/issues/5516
From my observation, it seems that Stripe only allows phone numbers respective to the country of the store - https://github.com/Automattic/woocommerce-payments/issues/5516#issuecomment-1415987684

RadoslavGeorgiev (approved):
Looks amazing! Thank you for the polished PR, :shipit:!

naman03malhotra (commented):
@htdat is is possible to load the `<PhoneNumberInput` component in a lazy fashion so that it doesn't effect the load times of the page?

htdat:
@naman03malhotra - as discussed with you via DM. I am summarizing the discussion here. 

I don’t see much slower in the page load time. In fact, the new package adds up the size of `dist/settings.js` but it is mostly about that size only, i.e. merchants' browsers need to fetch more ~66KB.  It does not use any external fetch/request, therefore, the rendering time will not be affected much. With your question about the lazy fashion, I am not sure how to achieve (or if there is any significant benefit with that laziness) because there is no async function insides this component.

Since it was validated to use in two other places (below), I think it’s generally fine. What I am doing here is to just abstract the existing component so that it can be reused.

- WooPay checkout page (for shoppers) - [code ref](https://github.com/Automattic/woocommerce-payments/blob/a9d2b7d162fd5f948eba7709d63ff12df8d27a8e/client/components/platform-checkout/save-user/checkout-page-save-user.js#L292-L292).
- wp-admin > Payments > Card Readers > Receipt Details (for merchants) - [code ref](https://github.com/Automattic/woocommerce-payments/blob/6f4e961d07ebe3cf5c64c9d16798af91dc005f37/client/card-readers/settings/sections/contacts-details.js#L99).

PR #5483, opened 2023-01-27
Pass blocks data to woopay

### LINKED ISSUE #5482: Pass blocks data fields to WooPay
### Description

To make the plugins that use the Newsletter block (Automatewoo, Mailpoet and Mailchimp) compatible with WooPay, we need pass in the relevant settings fields from the merchant store to WooPay. This can be done at the woopay checkout initiation by including a new field [here](https://github.com/Automattic/woocommerce-payments/blob/483a1da05d09f35a44c07d2a8f1839872f8c3fe2/includes/class-wc-payments.php#L1112).  

### Acceptance criteria

* Pass settings from Mailchimp, MailPoet and Automatewoo to WooPay
* Passed settings fields should be same as the ones used in the merchant store
* Settings should only be passed when the relevant plugin is active

malithsen:
Fixes #5482 

Changes introduced here extract Blocks settings data fields from Automatewoo, Mailchimp and Mailpoet and then pass them to WooPay when initiating the checkout. For Automatewoo and Mailchmp, the data fields are extracted by registering the blocks on a standalone [Integration Registry](https://github.com/woocommerce/woocommerce-blocks/blob/dea2050a69d0faaa45cf113ec40ad0b39c95613f/docs/third-party-developers/extensibility/checkout-block/integration-interface.md#L22). I didn't find a good way to follow the same approach for Mailpoet, so its data are extracted directly from its Settings instance without registering the plugin.

This data extraction mechanism is a stopgap until we have a more permanent API based approach to fetch these data for WooPay.

#### Testing instructions

Regression tests
* Make sure Automatewoo, Mailchimp and Mailpoet are *not* installed in the merchant.
* Add some products to the cart and go through WooPay checkout.
* Make sure checkout is functional and no errors show up in WC logs

Test passed data fields
* Install WooCommerce Blocks
* Install and activate Automatewoo
* Go to blocks checkout page
* Open devtools and note the fields in `wcSettings.automatewoo_data`
* Add a breakpoint [here](https://github.com/Automattic/woocommerce-payments/blob/483a1da05d09f35a44c07d2a8f1839872f8c3fe2/includes/class-wc-payments.php#L1140)
* Add a product to the cart and go to the checkout page.
* Use an already WooPay registered email to initiate the WooPay OTP prompt
* Go through OTP prompt, and note the value of [blocks_data](https://github.com/Automattic/woocommerce-payments/blob/483a1da05d09f35a44c07d2a8f1839872f8c3fe2/includes/class-wc-payments.php#L1138) at the breakpoint
* Make sure its values are same as the ones noted in `wcSettings.automatewoo_data`
* Follow the above steps again for Mailchimp and Mailpoet and match them against the values in `wcSettings.mailchimp-newsletter_data` and `wcSettings.mailpoet_data`

-------------------

**Post merge**


includes/compat/blocks/class-blocks-data-extractor.php
```diff
@@ -0,0 +1,114 @@
...
+	 * Get a list of available Blocks.
+	 */
+	private function get_available_blocks() {
+		$blocks = [];
+		if ( class_exists( '\AutomateWoo\Blocks\Marketing_Optin_Block' ) ) {
+			// phpcs:ignore
+			/**
+			 * @psalm-suppress UndefinedClass
+			 */
+			array_push( $blocks, new \Automatewoo\Blocks\Marketing_Optin_Block() );
```
a-danae: Should we replace `array_push` with `$blocks[] = ...`? It looks like it'd work the same for this case, and the PHP docs seem to [recommend](https://www.php.net/manual/en/function.array-push.php#:~:text=each%20passed%20value.-,Note,-%3A%20If%20you) it when that's the case.

Same thing about line 61 below, where we also use `array_push`.

includes/class-wc-payments.php
```diff
@@ -1082,6 +1083,8 @@ function ( $container ) {
 	 */
 	public static function ajax_init_platform_checkout() {
 		$is_nonce_valid = check_ajax_referer( 'wcpay_init_platform_checkout_nonce', false, false );
+		include_once WCPAY_ABSPATH . 'includes/compat/blocks/class-blocks-data-extractor.php';
+		$blocks_data_extractor = new Blocks_Data_Extractor();
```
a-danae: Would it be better to instantiate Blocks_Data_Extractor when we're sure we'll use it? We're registering the blocks in the class constructor, so there's some overhead there, and there's a nonce validation below that could bail out before using `$blocks_data_extractor`.

includes/compat/blocks/class-blocks-data-extractor.php
```diff
@@ -0,0 +1,114 @@
...
+	public function __construct() {
+		$this->integration_registry = new IntegrationRegistry();
+		$this->blocks               = $this->get_available_blocks();
+
+		$this->register_blocks();
+	}
+
+	/**
+	 * Get a list of available Blocks.
+	 */
```
a-danae, line 38: For all the docblocks of the methods that return something in this class, it'd probably be good to add the `@return` tag.

includes/compat/blocks/class-blocks-data-extractor.php
```diff
@@ -0,0 +1,114 @@
...
+	private $blocks = [];
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->integration_registry = new IntegrationRegistry();
+		$this->blocks               = $this->get_available_blocks();
+
+		$this->register_blocks();
```
a-danae: Two questions here:

1. Is there any advantage in running this code in the constructor, and then retrieving the data with `get_data` at some point later on, compared to doing it all together in the same call? It's all working well this way. I just think it makes it harder to follow how things play together.
2.  Should we unregister blocks after we get their data? I'm not sure if there are side effects of not doing it, maybe we could check with Nadir. In this context, I get this is only called in an ajax call and I think it's unlikely there are any.

a-danae (commented):
Thanks for your work here @malithsen ! Things work as described and the code makes sense to me. I left a couple of comments, probably nits.

> I didn't find a good way to follow the same approach for Mailpoet

I played a bit with this but couldn't get much either. I'm not a fan of this approach because any change in the settings Mailpoet uses would require an update in WCPay, but I don't have a different alternative to propose. 

@senadir, maybe you see a way to pass Mailpoet data without retrieving each setting directly, as we're doing now.

includes/compat/blocks/class-blocks-data-extractor.php
malithsen: Because we are using a separate `integration_registryy` I didn't see a reason to unregister. I was also not sure about the added overhead of doing everything in `get_data`. But having looked at it again following your comment, the methods we are calling in `integration_registry` seems to be pretty light, so moving that logic to `get_data` makes sense just to be cautious. I've done this in https://github.com/Automattic/woocommerce-payments/pull/5483/commits/714c44c7c0f0fc1baf64c088a527279929cb10ff. Thanks!

Still instantiating `integration_registry` under the constructor to avoid it being passed back and forth.

includes/compat/blocks/class-blocks-data-extractor.php
malithsen: Added in https://github.com/Automattic/woocommerce-payments/pull/5483/commits/714c44c7c0f0fc1baf64c088a527279929cb10ff

includes/class-wc-payments.php
malithsen: Makes sense. Changed in https://github.com/Automattic/woocommerce-payments/pull/5483/commits/714c44c7c0f0fc1baf64c088a527279929cb10ff

includes/compat/blocks/class-blocks-data-extractor.php
malithsen: Thanks! Changed in https://github.com/Automattic/woocommerce-payments/pull/5483/commits/714c44c7c0f0fc1baf64c088a527279929cb10ff

malithsen:
>  I'm not a fan of this approach because any change in the settings Mailpoet uses would require an update in WCPay, but I don't have a different alternative to propose.

@a-danae I totally agree, I am not a fan of the tight coupling with plugin code. The only reason for opening this PR in the current status is that we are looking for a short-term solution. The other approach I looked into was getting data directly from Block's `integration_registry`, but I didn't make much progress there. Slack thread: p1674696949200809-slack-C02TS23QJ1X

includes/compat/blocks/class-blocks-data-extractor.php
a-danae: Makes sense! Thanks Malith!

includes/class-wc-payments.php
```diff
@@ -1134,6 +1138,7 @@ public static function ajax_init_platform_checkout() {
 				'return_url'                     => wc_get_cart_url(),
 			],
 			'user_session'         => isset( $_REQUEST['user_session'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['user_session'] ) ) : null,
+			'blocks_data'          => $blocks_data_extractor->get_data(),
```
a-danae: Would this fit better into `store_data`? Sorry I didn't ask in the previous review. Things work fine like this. Wondering because we include plugin information from the merchant store there, like the subscriptions one. But again, all good.

a-danae (approved):
LGTM 👍 thanks for the explanations and for the updates!

Regarding tight coupling. Yup, I think we both agree. It'd be great if we could get all the extension data like we're doing for Mailchimp and AutomateWoo, but I understand they're implemented differently, and that it's not possible. And the current approach does the work.

Big thanks for looking for a way around it so we can still support these extensions 🙌

includes/class-wc-payments.php
malithsen: > Would this fit better into store_data?

I agree, thanks for pointing it out. Changed it in https://github.com/Automattic/woocommerce-payments/pull/5483/commits/26c9a2b77b5be3fe634b8644b4ee684536abd95e

a-danae: Thanks!

alefesouza (approved):
LGTM

PR #5494, opened 2023-01-31
Improve the decision-making process to determine if the save option should be displayed for a UPE payment method.

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5495

This PR enhances the logic for displaying the "Save payment method" checkbox for each UPE payment method. Before this change, the `showSaveOption` flag was set to `true` for all UPE methods, even though only SEPA can be saved. The checkbox was previously visible but then quickly hidden.

This PR moves the logic for the "Save payment method" checkbox to the back-end. The flag `showSaveOption` is now set there, making future adjustments to this feature more convenient. 

Finally, this PR considers the type of UPE payment method, ensuring that the `showSaveOption` flag is set to `false` for all methods other than SEPA.

The updated UPE logic is:
* For SEPA: run checks (returns `true` if "save cards" is enabled and cart does not have a subscription product, `false` otherwise)
* For all other non-SEPA UPE methods: return `false`.

#### Testing instructions

* Clear the cache of your browser and ideally, run `npm run build` and `npm run start`.
* Turn on all the available UPE methods, login to the existing account on the merchant site. 
* Go to the Blocks checkout and confirm that there's no checkbox to save the payment method, which quickly disappears after mounting a component for everything but SEPA and CC.
* Go to the shortcode checkout and confirm that the save checkbox is available only for SEPA and CC as well. 
* Place two orders from shortcode and blocks by using SEPA and saving it. Confirm that the payment method was successfully saved. 
* Add a subscription product to the cart and confirm if SEPA on the blocks checkout doesn't have the save checkbox available
* Select different payment methods and ensure that the save checkbox logic is correct and remains the same
* Try to break things by doing any other actions.

-------------------

**Post merge**

### DIFF
--- a/client/checkout/blocks/upe.js
+++ b/client/checkout/blocks/upe.js
@@ -96,10 +96,7 @@ Object.entries( enabledPaymentMethodsConfig ).map( ( [ upeName, upeConfig ] ) =>
 		ariaLabel: __( 'WooCommerce Payments', 'woocommerce-payments' ),
 		supports: {
 			showSavedCards: getUPEConfig( 'isSavedCardsEnabled' ) ?? false,
-			showSaveOption:
-				( getUPEConfig( 'isSavedCardsEnabled' ) &&
-					! getUPEConfig( 'cartContainsSubscription' ) ) ??
-				false,
+			showSaveOption: getUPEConfig( 'showSaveOption' ) ?? false,
 			features: getUPEConfig( 'features' ),
 		},
 	} )
--- a/client/checkout/blocks/upe.js
+++ b/client/checkout/blocks/upe.js
@@ -96,7 +96,7 @@ Object.entries( enabledPaymentMethodsConfig ).map( ( [ upeName, upeConfig ] ) =>
 		ariaLabel: __( 'WooCommerce Payments', 'woocommerce-payments' ),
 		supports: {
 			showSavedCards: getUPEConfig( 'isSavedCardsEnabled' ) ?? false,
-			showSaveOption: getUPEConfig( 'showSaveOption' ) ?? false,
+			showSaveOption: upeConfig.showSaveOption ?? false,
 			features: getUPEConfig( 'features' ),
 		},
 	} )
--- a/includes/class-wc-payments-upe-checkout.php
+++ b/includes/class-wc-payments-upe-checkout.php
@@ -183,10 +183,24 @@ class WC_Payments_UPE_Checkout extends WC_Payments_Checkout {
 					]
 				),
 				'icon'                 => $payment_method->get_icon(),
+				'showSaveOption'       => $this->should_upe_payment_method_show_save_option( $payment_method_id ),
 			];
 		}
 
 		return $settings;
+	}
+
+	/**
+	 * Checks if the save option for a payment method should be displayed or not.
+	 *
+	 * @param string $payment_method_id Payment method ID, e.g. sepa_debit or bancontact.
+	 * @return bool - True if the payment method is SEPA and the saved cards feature is enabled for the gateway and there is no subscription item in the cart, false otherwise.
+	 */
+	private function should_upe_payment_method_show_save_option( $payment_method_id ) {
+		if ( Payment_Method::SEPA === $payment_method_id ) {
+			return $this->gateway->is_saved_cards_enabled() && ! $this->gateway->is_subscription_item_in_cart();
+		}
+		return false;
 	}
 
 	/**
--- a/includes/class-wc-payments-upe-checkout.php
+++ b/includes/class-wc-payments-upe-checkout.php
@@ -183,7 +183,7 @@ class WC_Payments_UPE_Checkout extends WC_Payments_Checkout {
 					]
 				),
 				'icon'                 => $payment_method->get_icon(),
-				'showSaveOption'       => $this->should_upe_payment_method_show_save_option( $payment_method_id ),
+				'showSaveOption'       => $this->should_upe_payment_method_show_save_option( $payment_method ),
 			];
 		}
 
@@ -193,11 +193,11 @@ class WC_Payments_UPE_Checkout extends WC_Payments_Checkout {
 	/**
 	 * Checks if the save option for a payment method should be displayed or not.
 	 *
-	 * @param string $payment_method_id Payment method ID, e.g. sepa_debit or bancontact.
-	 * @return bool - True if the payment method is SEPA and the saved cards feature is enabled for the gateway and there is no subscription item in the cart, false otherwise.
+	 * @param string $payment_method Payment method.
+	 * @return bool - True if the payment method is reusable and the saved cards feature is enabled for the gateway and there is no subscription item in the cart, false otherwise.
 	 */
-	private function should_upe_payment_method_show_save_option( $payment_method_id ) {
-		if ( Payment_Method::SEPA === $payment_method_id ) {
+	private function should_upe_payment_method_show_save_option( $payment_method ) {
+		if ( $payment_method->is_reusable() ) {
 			return $this->gateway->is_saved_cards_enabled() && ! $this->gateway->is_subscription_item_in_cart();
 		}
 		return false;
--- a/includes/class-wc-payments-upe-checkout.php
+++ b/includes/class-wc-payments-upe-checkout.php
@@ -16,6 +16,7 @@ use WCPay\Constants\Payment_Method;
 use WCPay\Fraud_Prevention\Fraud_Prevention_Service;
 use WCPay\Payment_Methods\UPE_Payment_Gateway;
 use WCPay\Platform_Checkout\Platform_Checkout_Utilities;
+use WCPay\Payment_Methods\UPE_Payment_Method;
 
 
 /**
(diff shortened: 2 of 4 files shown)


client/checkout/blocks/upe.js
```diff
@@ -96,10 +96,7 @@ Object.entries( enabledPaymentMethodsConfig ).map( ( [ upeName, upeConfig ] ) =>
 		ariaLabel: __( 'WooCommerce Payments', 'woocommerce-payments' ),
 		supports: {
 			showSavedCards: getUPEConfig( 'isSavedCardsEnabled' ) ?? false,
-			showSaveOption:
-				( getUPEConfig( 'isSavedCardsEnabled' ) &&
-					! getUPEConfig( 'cartContainsSubscription' ) ) ??
-				false,
+			showSaveOption: getUPEConfig( 'showSaveOption' ) ?? false,
```
mdmoore: I noticed some strange behavior with this `showSaveOption` flag that the Checkout block uses. It seems that as long as it's `false` or doesn't exist, the save payment method checkbox functions correctly for each payment method. You can test that by simply setting the front-end `showSaveOption` flag to `false` or removing it completely.

In the line here, `showSaveOption` will always return false because `getUPEConfig( 'showSaveOption' )` attempts to reach a top level property named `showSaveOption` on the `woocommerce_payments_data` object when that property actually exists a couple levels lower on each payment method within `paymentMethodsConfig`. 

<details>
    <summary>Visual</summary>

image

</details>

We already have access to `paymentMethodsConfig` via `enabledPaymentMethodsConfig`. So if we wanted to handle the logic on the backend, we could use: 

```suggestion
			showSaveOption: upeConfig.showSaveOption ?? false,
```

Though, I'm wondering if we should just continue handling this logic on the front-end since the data is already available. The `is_reusable` flag on each payment method is used to determine if the payment method can be tokenized. Currently, this is only true for SEPA on our side, but we'll enable saving other payment methods in the future. So we could use something like:

```suggestion
			showSaveOption: upeConfig.isReusable
				? getUPEConfig( 'isSavedCardsEnabled' ) &&
				  ! getUPEConfig( 'cartContainsSubscription' )
				: false,
```

mdmoore (changes requested):
Thanks for working on this Timur! I left a couple comments. I'm anxious to hear your thoughts. Let me know if you have any questions or if I can clarify anything.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -183,12 +183,26 @@ public function get_enabled_payment_method_config() {
...
+	 *
+	 * @param string $payment_method_id Payment method ID, e.g. sepa_debit or bancontact.
+	 * @return bool - True if the payment method is SEPA and the saved cards feature is enabled for the gateway and there is no subscription item in the cart, false otherwise.
+	 */
+	private function should_upe_payment_method_show_save_option( $payment_method_id ) {
+		if ( Payment_Method::SEPA === $payment_method_id ) {
+			return $this->gateway->is_saved_cards_enabled() && ! $this->gateway->is_subscription_item_in_cart();
+		}
+		return false;
+	}
```
mdmoore:
```suggestion
	 * @return bool - True if the payment method is reusable and the saved cards feature is enabled for the gateway and there is no subscription item in the cart, false otherwise.
	 */
	private function should_upe_payment_method_show_save_option( $payment_method_id ) {
		if ( $this->gateway->wc_payments_get_payment_method_by_id( $payment_method_id )->is_reusable() ) {
			return $this->gateway->is_saved_cards_enabled() && ! $this->gateway->is_subscription_item_in_cart();
		}
		return false;
	}
```

As I mentioned above, the `is_reusable` flag inside each payment method determines whether a payment method type can be tokenized. Right now, we only support saving SEPA payment methods, but in the future we'll add more. My understanding is that Stripe supports tokenizing several of these payment method types already. If we decide to move the logic to the back-end here, it would be good to future proof the function rather that specifying SEPA. There are probably several other areas of WCPay that would benefit from the same idea.

mdmoore (changes requested):

client/checkout/blocks/upe.js
timur27: > attempts to reach a top level property named showSaveOption on the woocommerce_payments_data object when that property actually exists a couple levels lower

Oh, I planned to get it from the particular payment method, but I used the wrong place by mistake. 🤕 My bad. Many thanks for noticing this! Fixed.

> I'm wondering if we should just continue handling this logic on the front-end

Regardless of whether we have everything available on the UI or not, I'd prefer to keep any domain logic within the domain layer on the back end. Gateway/checkout classes are the ones in charge of giving us an idea of whether they can be saved.  

Another advantage of keeping the logic on the back end is that all the scripts (e.g. shortcode-classic/shortcode-upe/blocks-classic/blocks-upe) can re-use this flag instead of copying the cascaded if statement from each other and duplicate the code, which is an ideal environment for potential bugs in the future due to 'I updated there but forgot to update here'.

Let me know how it sounds to you and if you agree or you'd like to discuss more. Looking forward to hearing from you! 

> The is_reusable flag

I'll refer to this in another comment.

includes/class-wc-payments-upe-checkout.php
timur27: Thank you for explaining the "business" background of the `is_reusable` flag - I didn't know it reflects the subset of payment methods, which can be saved. I checked how the method works and confirmed that it's the equivalent of the SEPA subset. It turned out that Link is also tokenized = reusable = can be saved, and I checked this on `develop` as well. 

This can (even should) replace `Payment_Method::SEPA === $payment_method_id` to avoid any changes in the future when the set of tokenized methods is updated. This is a super important note from you, and thanks for it! Fixed. 

> 💡 instead of `$this->gateway->wc_payments_get_payment_method_by_id( $payment_method_id )->is_reusable()` I'll provide `$payment_method` as it's already available.

client/checkout/blocks/upe.js
timur27: > Another advantage of keeping the logic on the back end is that all the scripts (e.g. shortcode-classic/shortcode-upe/blocks-classic/blocks-upe) can re-use this flag instead of copying the cascaded if statement from each other and duplicate the code, which is an ideal environment for potential bugs in the future due to 'I updated there but forgot to update here'.

Quoting myself here. I've just noticed that we already have a perfect place that has benefited from such an approach - 3a435eb9cb4641a408697a6cad06ecea390445c7.

timur27:
@mdmoore Many thanks for the review - it looks like I've addressed/answered everything. Apart from the testing in the description, I'd like to ask you to test [this new change](https://github.com/Automattic/woocommerce-payments/pull/5494#discussion_r1093113289) as well. 

* select different payment methods and ensure that the save checkbox logic is correct **and remains the same**. While changing methods, `upeOnChange` will fire and thus update the gateway's state.

Additionally, to test everything out more thoroughly, you might want to add a subscription product to the cart and confirm if SEPA on the blocks checkout doesn't have the save checkbox available.

I've updated the description with these two scenarios.

client/checkout/blocks/upe.js
mdmoore: >Regardless of whether we have everything available on the UI or not, I'd prefer to keep any domain logic within the domain layer on the back end. Gateway/checkout classes are the ones in charge of giving us an idea of whether they can be saved.

>Another advantage of keeping the logic on the back end is that all the scripts (e.g. shortcode-classic/shortcode-upe/blocks-classic/blocks-upe) can re-use this flag instead of copying the cascaded if statement from each other and duplicate the code, which is an ideal environment for potential bugs in the future due to 'I updated there but forgot to update here'.

Yes, this all makes perfect sense to me. I'm all for keeping the logic in the back-end for those reasons. Thanks for expanding on the reasons for doing so.

includes/class-wc-payments-upe-checkout.php
mdmoore: >This can (even should) replace Payment_Method::SEPA === $payment_method_id to avoid any changes in the future when the set of tokenized methods is updated. This is a super important note from you, and thanks for it! Fixed.

Perfect!  Thank you 🙌. 

>I'll provide $payment_method as it's already available.

Ah, nice catch! Looks great.

mdmoore (approved):
Thanks for the changes, Timur! Everything is testing well per the instructions.

- Go to the Blocks checkout and confirm that there's no checkbox to save the payment method, which quickly disappears after mounting a component for everything but SEPA and CC. ✅ 
- Go to the shortcode checkout and confirm that the save checkbox is available only for SEPA and CC as well. ✅ 
- Place two orders from shortcode and blocks by using SEPA and saving it. Confirm that the payment method was successfully saved. ✅ 
- Add a subscription product to the cart and confirm if SEPA on the blocks checkout doesn't have the save checkbox available ✅ 
- Select different payment methods and ensure that the save checkbox logic is correct and remains the same ✅

**Note:** When only a subscription item is in the cart, the save checkbox shows up for CC. However, I tested this on the `poc/upe-instances-multiplied` branch and find the same behavior. So this isn't something introduced by your PR, but should be taken care of in another issue.

All good to me! 🚢

Issue #5397, opened 2023-01-12
Add skeleton loading state for UPE payment methods in blocks


FangedParakeet:
### Description

This is mainly noticeable when the split-UPE feature is enabled. Please see pdpOw2-Pn-p2#comment-2162 and surrounding conversation for more context.

Currently, in the blocks checkout, UPE elements are only mounted when their parent gateway is selected at checkout. This means the initial payment intent is created and the resulting element mounted only after the gateway radio button is selected. While these processes are occurring, the gateway at checkout remains static, jolting into life only once the element is finally ready to mount. Please lower your gaze to the below animation to see what I'm talking about.

![Janky UPE blocks loading](https://cdn-std.droplr.net/files/acc_1104206/ktwVqJ)
_This is what I'm talking about_

This issue is to add a skeleton loading state to this section while the element is being prepared. On the classic elements, there is [a ready state](https://stripe.com/docs/js/element/events/on_ready); if this is available for the Stripe React component as well, this would be an appropriate event to use to hide the loading state and show the mounted payment element.

### Acceptance criteria

* Add skeleton loading state for UPE payment methods in blocks checkout.
* Loading state removed when payment element is ready and mounted.

zmaglica:
I noticed that this issue is in Heisenberg backlog. @frosso does the team have this issue under its radar and do you folks plan to work on this issue? The reason why I am asking is that I have to prioritize it, and it looks like to me that this issue falls under your scope, so you can maybe prioritize it on your end.

FangedParakeet:
Just FYI, this is definitely under our radar and I have sort of begun resolving it already in a separate issue, by implementing a `LoadableBlock` element around the blocks payment elements, over [here](https://github.com/Automattic/woocommerce-payments/pull/5496/commits/ab81e884992cb41f5cee2307bbe82e11e5e19b89).

I think it would be nice to get this into our upcoming release, but we have a few other required issues that are ahead in terms of priority and since this is a non-essential enhancement, I'm not eager to commit wholeheartedly to completing this task with complete certainty within this sprint, since we will have limited availability due to the meetup next week.

timur27:
This is the same question I’ve asked [here](https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/5502).


[...]

Issue #5271, opened 2022-12-10
Multiple UPE - Spike: Explore scope of creating new feature flag for Split-UPE


FangedParakeet:
### Description

Following #4671 it has been decided that the best plan to rollout the first phase of the split-UPE project would be to create a new feature flag to be used in parallel with the existing UPE implementation. Downgrading the UPE for merchants would remove the additional payment methods from merchant checkouts, which would be unacceptable since these payment methods receive considerable usage. Moreover, we would prefer to gradually phase out the existing UPE implementation, to ensure that the split-UPE does not create any major conflicts in production, so replacing this existing UPE code entirely--as has been the current design--would also not be agreeable.

Consequently, our confirmed rollout plan must initially support the non-UPE gateway, legacy UPE gateway, and split-UPE gateway simultaneously. Once the final phase of the split-UPE has been implemented and released without incident, supporting [Stripe's new deferred intent payment flow](https://stripe.com/docs/payments/defer-intent-creation) and supporting compatibility with WooPay, we can plot to remove the redundant gateways and unify the payment gateways into one. 

Since our current POC branch #3968 unfortunately overwrites the previous UPE implementation, we will now have to work backwards somewhat to ensure that we can reintegrate the previous implementation and separate it from the split-UPE with a new feature flag. This issue is to try to scope out what would be required to create this separation and how much effort it would cost our squad. 

### Additional context

#4671 
pc2DNy-2qG-p2#comment-3837

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5271?planning-poker) with Zenhub @mdmoore @timur27

FangedParakeet:
Scanning through #3968, I've basically grouped the additions to the UPE made by the split-UPE implementation into the following categories.

- Multiple payment gateways registered to plugin, instead of single.
- Multiple payment elements mounted/registered to the shortcode/blocks checkout, instead of single element.
- Multiple payment intent caches for each payment method type, instead of single payment intent cache.
- Multiple AJAX functions to handle intent actions (create setup intent, create payment intent, update intent) for each payment method registered as hooks, instead of single hook.
- Testing instructions for payment gateway is dependent on the payment method type, instead of always displaying card testing instructions.
[...]

PR #5496, opened 2023-01-31
Adds and implements new separate feature flag for Split UPE feature

### LINKED ISSUE #5353: Multiple UPE - Create and implement split-UPE feature flag
### Description

Please see #5271 for more context on the necessity of this change.

For as long as we have been working on the split-UPE branch, we have simply re-used the existing UPE flag. Sadly this makes it challenging to roll this feature out gradually and also prevents us from downgrading as a fallback to the original undivided UPE. For this reason, we have decided to create a new split-UPE flag that we can use in parallel alongside the current UPE flag (`_wcpay_feature_upe`). This issue is to create this flag and implement it once and for all, hopefully following the wisdom shared in #5271.

This is a belated retroactive refactor that means we will have to undo some of our current work, in order to return previous UPE functionality and maintain it alongside the split-UPE code. This will probably be quite unpleasant. I'm sorry. 

### Acceptance criteria

* A new feature flag will be created for the split-UPE branch.
* Existing UPE feature flag will remain (`_wcpay_feature_upe`).
* A user should be able use the non-UPE checkout, UPE checkout, **and** split-UPE checkoutby enabling the respective feature flag. The split-UPE flag will override the existing UPE flag, but if the split-UPE flag is disabled, the user should be able to use either the legacy checkout or the UPE checkout, depending on the status of the single-UPE flag (`_wcpay_feature_upe`).
* Tests should be duplicated and separated for the split-UPE feature (separate tests for single-UPE and split-UPE).

### Additional context

#5271 
pc2DNy-2qG-p2#comment-3837

FangedParakeet:
Closes #5353. 

# Split UPE: A New Frontier

This PR adds a new feature flag `_wcpay_feature_upe_split` that can be used to toggle the new Split UPE implementation. With no flags enabled, the legacy card element will appear at checkout; with **only** the `_wcpay_feature_upe` feature flag enabled, the legacy UPE, singular "Popular payment methods" gateway will appear at checkout; with the `_wcpay_feature_upe_split` feature flag enabled, the split UPE gateways will all appear at checkout. This third feature flag will override the original UPE flag when present. 

In order to accomplish this, I have extracted all of the Split UPE-specific implementation code into a new payment gateway class, `UPE_Split_Payment_Gateway`, which extends the original `UPE_Payment_Gateway` class. Consequently, depending on the feature flag enabled, the default payment gatewag representing the plugin will either be an instance of `WC_Payment_Gateway_WCPay`, `UPE_Payment_Gateway`, or `UPE_Split_Payment_Gateway`.

**WIP.**

# Pending requirements

- ~Currently only supports use with the classic checkout. Blocks checkout compatibility is a requirement.~ _Edit: Currently compatible with blocks checkout._
- ~Tests for the regular and split UPE gateways have not yet been separated. Tests for all compatible payment gateways are a requirement.~ _Edit: tests are green and fully operational.

#### Testing instructions

Feature flag can be enabled by using Automattic/woocommerce-payments-dev-tools#80 on your test site. All payment flows with all three gateways (card, UPE, split UPE) should be fully operational, so all should be tested. 

-------------------

**Post merge**


FangedParakeet:
I still need to resolve the tests to ensure that they test both UPE gateways appropriately, but in the meantime, this PR is fully operational and should be working in both the classic shortcode and blocks checkouts. I'd recommend using Automattic/woocommerce-payments-dev-tools#80 to switch between feature flags, but if we can start to get this approach reviewed, we can unblock other rollout/feature-flag related tasks that are dependent on the finalisation of this implementation.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -473,9 +468,9 @@ public function process_payment( $order_id ) {
 		$currency                  = $order->get_currency();
 		$converted_amount          = WC_Payments_Utils::prepare_amount( $amount, $currency );
 		$payment_needed            = 0 < $converted_amount;
-		$selected_upe_payment_type = $this->stripe_id;
+		$selected_upe_payment_type = ! empty( $_POST['wcpay_selected_upe_payment_type'] ) ? wc_clean( wp_unslash( $_POST['wcpay_selected_upe_payment_type'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
 		$payment_type              = $this->is_payment_recurring( $order_id ) ? Payment_Type::RECURRING() : Payment_Type::SINGLE();
-		$save_payment_method       = $payment_type->equals( Payment_Type::RECURRING() ) || ! empty( $_POST[ 'wc-' . self::GATEWAY_ID . '_' . $this->stripe_id . '-new-payment-method' ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$save_payment_method       = $payment_type->equals( Payment_Type::RECURRING() ) || ! empty( $_POST[ 'wc-' . $this->id . '-new-payment-method' ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
FangedParakeet, line 473: I believe this change may resolve a few open issues around saved payments/subscriptions, since we were not otherwise correctly recognising that the save payment information checkbox was checked.

tests/e2e/utils/flows.js
```diff
@@ -276,7 +276,44 @@ export const merchantWCP = {
...
+
+		const isSplitUPEEnabled = await page.$(
+			'#_wcpay_feature_upe_split:checked'
+		);
+
+		if ( isSplitUPEEnabled ) {
+			await expect( page ).toClick( 'label', {
+				text: 'Enable Split UPE checkout',
+			} );
+		}
```
FangedParakeet, line 298: This E2E flow may require Automattic/woocommerce-payments-dev-tools#80 to be merged first.

client/checkout/blocks/upe-split-fields.js
```diff
@@ -0,0 +1,520 @@
...
+					</div>
+				</div>
+			);
+		}
+
+		return null;
+	}
+
+	return (
+		<LoadableBlock
```
FangedParakeet, line 489: This may somewhat resolve #5397.

client/checkout/api/index.js
```diff
@@ -338,14 +338,14 @@ export default class WCPayAPI {
...
-		return this.request(
-			buildAjaxURL(
-				getConfig( 'wcAjaxUrl' ),
-				`init_setup_intent_${ paymentMethodType }`
-			),
-			{ _ajax_nonce: getConfig( 'createSetupIntentNonce' ) }
-		).then( ( response ) => {
+	initSetupIntent( paymentMethodType = '' ) {
+		let path = 'init_setup_intent';
+		if ( '' !== paymentMethodType ) {
```
timur27: This method and [the one responsible for creating payment intent](https://github.com/Automattic/woocommerce-payments/pull/5496/files#diff-bc05e644f9133795b10f0696ea2862c0abeecfeed6f3ff2eaf48a691c701e14bR407) are using different ways to check the `paymentMethodType` presence - can we align them?

`if ( '' !== paymentMethodType )` vs `if (paymentMethodType)`

Can we use one of these from above everywhere? Or maybe we can use `this.options.isUPESplitEnabled` to perform the check - the same way we do [here](https://github.com/Automattic/woocommerce-payments/pull/5496/files#diff-bc05e644f9133795b10f0696ea2862c0abeecfeed6f3ff2eaf48a691c701e14bR455)?

All three options are acceptable, I think - just a matter of alignment.

client/checkout/blocks/style.scss
```diff
@@ -51,3 +51,4 @@ button.wcpay-stripelink-modal-trigger:hover {
 }
 
 @import '../platform-checkout/style';
+@import '../../components/loadable/style';
```
timur27, line 54: What's the reason for adding this, if you could explain?

client/checkout/blocks/upe-split-fields.js
```diff
@@ -0,0 +1,520 @@
...
+	// Checks whether there are errors within a field, and saves them for later reporting.
+	const upeOnChange = ( event ) => {
+		// Update WC Blocks gateway config based on selected UPE payment method.
+		if (
+			getUPEConfig( 'isSavedCardsEnabled' ) &&
+			! getUPEConfig( 'cartContainsSubscription' )
+		) {
+			gatewayConfig.supports.showSaveOption =
+				paymentMethodsConfig[ event.value.type ].isReusable;
+		}
```
timur27: Can we align these lines with the simplified approach we applied [here](https://github.com/Automattic/woocommerce-payments/pull/5496/files?diff=unified&w=0#diff-48ac7ae39ae4100b724967bc870908d0feba515258a2be4d971bfe44f9889c68R318-R319) last week? In case the link doesn't work -- I'm referring to lines `318-319` in `client/checkout/blocks/upe-fields.js`.

includes/class-wc-payments-features.php
```diff
@@ -14,19 +14,36 @@
...
 class WC_Payments_Features {
 	const UPE_FLAG_NAME                     = '_wcpay_feature_upe';
+	const UPE_SPLIT_FLAG_NAME               = '_wcpay_feature_upe_split';
 	const WCPAY_SUBSCRIPTIONS_FLAG_NAME     = '_wcpay_feature_subscriptions';
 	const WOOPAY_EXPRESS_CHECKOUT_FLAG_NAME = '_wcpay_feature_woopay_express_checkout';
 	const AUTH_AND_CAPTURE_FLAG_NAME        = '_wcpay_feature_auth_and_capture';
 
 	/**
-	 * Checks whether the UPE gateway is enabled
+	 * Checks whether amy UPE gateway is enabled.
```
timur27: sorry for this nitpick 
```suggestion
	 * Checks whether any UPE gateway is enabled.
```

includes/class-wc-payments-upe-checkout.php
```diff
@@ -105,6 +111,12 @@ public function get_payment_fields_js_config() {
 		$payment_fields['wcBlocksUPEAppearance']    = get_transient( UPE_Payment_Gateway::WC_BLOCKS_UPE_APPEARANCE_TRANSIENT );
 		$payment_fields['cartContainsSubscription'] = $this->gateway->is_subscription_item_in_cart();
 
+		if ( WC_Payments_Features::is_upe_legacy_enabled() ) {
+			$payment_fields['checkoutTitle']        = $this->gateway->get_checkout_title();
+			$payment_fields['upePaymentIntentData'] = $this->gateway->get_payment_intent_data_from_session();
+			$payment_fields['upeSetupIntentData']   = $this->gateway->get_setup_intent_data_from_session();
+		}
+
```
timur27, line 119: This class now handles the legacy UPE implementation (AJAX hooks definitions in the constructor confirm this) - why do we use this part of the code within an `if` then? Is the split UPE possibly executing this code (if yes, why)?

Shouldn't we just run it without a check? Or maybe the hooks in the constructor are not reflecting the state of the things?

includes/class-wc-payments-upe-blocks-payment-method.php
```diff
@@ -15,17 +15,30 @@ class WC_Payments_UPE_Blocks_Payment_Method extends WC_Payments_Blocks_Payment_M
...
+			'stripe',
+			'https://js.stripe.com/v3/',
+			[],
+			'3.0',
+			true
+		);
 
 		wp_register_script(
-			'WCPAY_BLOCKS_UPE_CHECKOUT',
+			'WCPAY_BLOCKS_CHECKOUT',
```
timur27, line 34: Isn't the same constant being used for the non-UPE block checkout? If so - is such an approach of using the same constant twice anyhow impacting the functionality? What's your opinion?

includes/class-wc-payments.php
```diff
@@ -364,32 +366,40 @@ public static function init() {
...
+			Eps_Payment_Method::class,
+			Link_Payment_Method::class,
+		];
+		foreach ( $payment_method_classes as $payment_method_class ) {
+			$payment_method                               = new $payment_method_class( self::$token_service );
+			$payment_methods[ $payment_method->get_id() ] = $payment_method;
+		}
+		if ( WC_Payments_Features::is_upe_split_enabled() ) {
 			foreach ( $payment_method_classes as $class ) {
 				$payment_method = new $class( self::$token_service );
```
timur27: Aren't we making redundant walkthroughs by `$payment_method_classes` here? I guess we create 2x10 `UPE_Payment_Method` object instances instead of 10. I understand that for the split UPE, we need to feed the gateway/methods maps, but then we could probably use `$payment_methods`'s values in the second `foreach` loop (`388`) as the collection to iterate on and skip line `389`, then use an already created payment method (from `384`) in `390-391`?

Another thought is that we might want to put these two if statements (for split and legacy UPE accordingly) within the loop and, thus, eliminate the need to loop twice?

includes/class-wc-payments.php
```diff
@@ -364,32 +366,40 @@ public static function init() {
...
+			$card_payments_checkout     = new WC_Payments_Checkout( self::$legacy_card_gateway, self::$platform_checkout_util, self::$account, self::$customer_service );
+			self::$wc_payments_checkout = new WC_Payments_UPE_Checkout( self::get_gateway(), self::$platform_checkout_util, self::$account, self::$customer_service );
+		} elseif ( WC_Payments_Features::is_upe_legacy_enabled() ) {
+			self::$card_gateway         = new UPE_Payment_Gateway( self::$api_client, self::$account, self::$customer_service, self::$token_service, self::$action_scheduler_service, $payment_methods, self::$failed_transaction_rate_limiter, self::$order_service );
 			self::$wc_payments_checkout = new WC_Payments_UPE_Checkout( self::get_gateway(), self::$platform_checkout_util, self::$account, self::$customer_service );
 		} else {
 			self::$card_gateway         = self::$legacy_card_gateway;
-			self::$wc_payments_checkout = $payments_checkout;
+			self::$wc_payments_checkout = new WC_Payments_Checkout( self::$legacy_card_gateway, self::$platform_checkout_util, self::$account, self::$customer_service );
 		}
```
timur27: I think we discussed [the idea of introducing the factory object](https://github.com/Automattic/woocommerce-payments/issues/5271#issuecomment-1371164973) to manage this state machine. I assume that it caused some headaches, and you feel that this would prolong this PR which is vital for our rollout - if this is the case, then I'm ok to leave this as it is. In this case, I personally would vote for refactoring it anyway but in a separate issue. What do you think about all this?

includes/class-wc-payments.php
```diff
@@ -569,6 +578,10 @@ public static function register_gateway( $gateways ) {
 			}
 
 			return array_merge( $gateways, $all_upe_gateways );
+		} elseif ( WC_Payments_Features::is_upe_enabled() ) {
+			$gateways[] = self::$card_gateway;
```
timur27: Cosmetic: we can just `return [ self::$card_gateway ];` here and in the next `else` statement too, and remove line `587`. I'm leaving it up to you if you feel it's a thing worth spending a few minutes on.

.github/workflows/e2e-pull-request.yml
```diff
@@ -30,11 +30,13 @@ jobs:
...
       fail-fast:     false
       matrix:
-        test_groups:   [ 'wcpay', 'subscriptions', 'upe' ] # [TODO] Unskip blocks tests after investigating constant failures.
+        test_groups:   [ 'wcpay', 'subscriptions', 'upe', 'upeSplit' ] # [TODO] Unskip blocks tests after investigating constant failures.
         test_branches: [ 'merchant', 'shopper' ]
         exclude:
           - test_groups: 'upe'
-            test_branches: 'merchant' 
+            test_branches: 'merchant'
+          - test_ groups: 'upeSplit'
```
timur27: This typo will make the e2e test suite

```suggestion
          - test_groups: 'upeSplit'
```

.github/workflows/e2e-pull-request.yml
```diff
@@ -30,11 +30,13 @@ jobs:
...
+            test_brances: 'merchant'
```
timur27:
```suggestion
            test_branches: 'merchant'
```

timur27 (commented):
Thanks @FangedParakeet for working on this! I did the review and left a couple of comments. Below you will find my testing round results. There are some things which don't work, but part of them in my opinion is not a result of your PR. Nevertheless, we'd need to see if they are valid and prioritise.

Also, e2e tests on poc branch [are green.](https://github.com/Automattic/woocommerce-payments/actions/runs/4128558161) whereas tests on this PR [fail for some reason](https://github.com/Automattic/woocommerce-payments/actions/runs/4128558161). 


### _**Legacy UPE**_

_My account -> Payment methods_ page
* Add a new card ✅ 
* Delete a card ✅

_Shortcode checkout_
* Pay with a new card without saving it ✅ 
* Pay with a new card and save it, confirm that it has been saved ✅ 
* Pay with a saved card ✅ 
* Pay with UPE method (Sofort) ✅ 
* Pay with UPE method (SEPA) ✅ 
* Pay with saved SEPA ✅ 
* Pay and save SEPA ✅ 

_Pay for order_
* with a saved CC ✅ 
* with a new CC ✅ 
* with a saved SEPA ✅ 
* with a new SEPA ✅ 
* with check payments ✅ 

_Blocks checkout_
* Pay with a new CC ✅ 
* Pay with a new CC and save it 🟠   -- there is no checkbox to save the payment method. However, the same behavior occurs on `develop`. On shortcode, CC can be saved.
* Pay with a saved CC ✅ 
* Pay with a new SEPA ✅ 
* Pay with a new SEPA and save it 🟠 -- there is no checkbox to save the payment method. However, the same behavior occurs on `develop`. On shortcode, SEPA can be saved. 
* Pay with a saved SEPA ✅ 
* Pay with an other UPE method (Sofort) ✅ 

_Subscriptions in Blocks_
* Pay with SEPA ✅ 
* Pay with a saved SEPA ✅ 
* Pay with CC ✅ 
* Pay with a saved CC ✅ 

_Subscriptions in shortcode_
* Pay with SEPA ✅ 
* Pay with CC ✅ 

_Settings page_
* all the available payment methods are visible in the _All payment methods_ page in _Payments -> Settings_ ✅ 

### _**Split UPE**_
_My account -> Payment methods_ page
* Add a new card ✅ 
* Delete a card ✅
* Add a new SEPA ✅ 
* Delete SEPA ✅ 

_Shortcode checkout_
* Pay with a new card without saving it ✅ 
* Pay with a new card and save it to the platform ✅
* Pay with a new card and save it to the merchant site (WooPay disabled), confirm that it has been saved  ✅ 
* Pay with a saved card ✅ 
* Pay with UPE method (Sofort) ✅ 
* Pay with UPE method (SEPA) ✅ 
* Pay with saved SEPA ✅ 
* Pay and save SEPA 🔴 - After I save SEPA, it doesn't appear on the list of the saved payment methods during the consequent checkout.
* The recurring payment with SEPA 🟠 - `Error: The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.` is shown. This might not be the result of this PR, but I encountered it while testing.

_Pay for order_
* with a saved CC  - will be fixed by https://github.com/Automattic/woocommerce-payments/pull/5513
* with a new CC - will be fixed by https://github.com/Automattic/woocommerce-payments/pull/5513
* with a saved SEPA 🟠  - getting an error `Invalid payment method. Please input a new card number.`. @timur27 to check if https://github.com/Automattic/woocommerce-payments/pull/5513 fixes this.
* with a new SEPA 🟠  - getting an error `Invalid payment method. Please input a new card number.`. @timur27 to check if https://github.com/Automattic/woocommerce-payments/pull/5513 fixes this. 
* with check payments ✅ 

_Blocks checkout_
* Pay with a new CC ✅ 
* Pay with a new CC and save it ✅ 
* Pay with a saved CC ✅ 
* Pay with a new SEPA ✅ 
* Pay with a new SEPA and save it 🟠 - after saving, SEPA doesn't appear on the list during the next checkout
* Pay with an other UPE method (Sofort) ✅ 

_Subscriptions in Blocks_
* Pay with SEPA 🔴  should be fixed by https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/5517
* Pay with a saved SEPA 🔴  should be fixed by https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/5517
* Pay with CC ✅ 
* Pay with a saved CC ✅ 

_Subscriptions in shortcode_
* Pay with SEPA ✅ 
* Pay with CC ✅ 

_Settings page_
* all the available payment methods are visible in the _All payment methods_ page in _Payments -> Settings_ ✅ 

### UPE disabled 
Shortcode checkout with CC ✅ 
Blocks with CC ✅ 
all the available payment methods are visible in the _All payment methods_ page in _Payments -> Settings_ ✅

client/checkout/blocks/style.scss
FangedParakeet: This `LoadableBlock` element is imported over [here](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/client/checkout/blocks/upe-split-fields.js#L36); however, for some reason these styles were not loaded when I imported the script, so I had to import the stylesheet manually.

includes/class-wc-payments-upe-checkout.php
FangedParakeet: If the split UPE is enabled, then `$this->gateway` will be an instance of `UPE_Split_Payment_Gateway`, whereas if the split UPE is not enabled, it will be an instance of `UPE_Payment_Gateway` (see `class-wc-payments.php` [lines 394](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/includes/class-wc-payments.php#L394) & [398](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/includes/class-wc-payments.php#L398)). In both the legacy implementation and the split UPE implementation, the `WC_Payments_UPE_Checkout` class is initialised (see `class-wc-payments.php` lines [396](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/includes/class-wc-payments.php#L398) & [399](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/includes/class-wc-payments.php#L398)). This is done to hydrate the `wcpay_upe_config` or `wcpay_config` JS variable, which will be used by either the legacy or split UPE front-end script (see `class-wc-payments-upe-checkout.php` line [237](https://github.com/Automattic/woocommerce-payments/blob/5353/add-split-upe-feature-flag/includes/class-wc-payments.php#L398)). 

These variables are only needed when the legacy UPE gateway is present and they have pretty irrelevant values when the split UPE is enabled, so I decided not to include them in the config when the split UPE is enabled. 

Not sure if I completely understood your question, so let me know if I missed the crux of your suggestion here!

includes/class-wc-payments-upe-blocks-payment-method.php
FangedParakeet: Essentially I have returned the state of the legacy UPE blocks implementation to [its current state on `develop`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-blocks-payment-method.php). When the legacy UPE is enabled or when the UPE is disabled, blocks scripts will be registered to the `WCPAY_BLOCKS_CHECKOUT` constant. However, when the split UPE is enabled, we will need to register split UPE scripts to `WCPAY_BLOCKS_UPE_CHECKOUT` and non-UPE scripts to `WCPAY_BLOCKS_CHECKOUT`, which is how the current implementation stands.

includes/class-wc-payments.php
FangedParakeet: Mind that the `$gateways` parameter in the `woocommerce_payment_gateways` filter contains an array of payment gateways--potentially including other non-WCPay gateways--so we will have to either do `$gateways[] = self::$card_gateway; return $gateways;` or `return array_merge( $gateways, [ self::$card_gateway ] );` to ensure that we still return all the previous and current gateways.

I've implemented the latter in bcd7d9e04cbed8cd23844dc4e294c74df1956e7a, because it's a one-liner and more consistent.

includes/class-wc-payments.php
FangedParakeet: I think using a factory object here is a fantastic idea. I would, however, prefer to resolve this in a separate issue, as I think this task would have pretty significant ramifications for the `WC_Payments` class and should be handled comprehensively and separately, rather than creating more complifications for this PR. 🙏

includes/class-wc-payments.php
FangedParakeet: The `UPE_Payment_Gateway` requires an array of payment methods and consequently so does the `UPE_Split_Payment_Gateway`, since it triggers the parent class's constructor. I've made the looping more efficient at the cost of some unavoidable code duplication in 3c3d75d46cc77cbec083a743953ab813f10912b8. Let me know if this will suffice. 🙏

client/checkout/blocks/upe-split-fields.js
FangedParakeet: Fixed in 17e7c8ac1ed7e8061dd002eea6d43e89892164d6. 👍

client/checkout/api/index.js
FangedParakeet: Fixed in afe6b556e4369a823a85f769053cc7e455038188. 👍

webpack/shared.js
```diff
@@ -12,11 +12,14 @@ module.exports = {
 			settings: './client/settings/index.js',
 			'blocks-checkout': './client/checkout/blocks/index.js',
 			'upe-blocks-checkout': './client/checkout/blocks/upe.js',
+			'upe-split-blocks-checkout':
+				'./client/checkout/blocks/upe_split.js',
```
timur27:
```suggestion
				'./client/checkout/blocks/upe-split.js',
```

PR #5513, opened 2023-02-02
Fix credit card and UPE payments on the Pay for order page

### LINKED ISSUE #5512: Credit card order payment fails with enabled UPE
With the UPE enabled, when paying for an order with the CC payment method, users are not redirected to the order summary page, instead, the page is in a loading state with the error in the console: 

```
upe.js?c32f:458 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'upeElement')
    at checkUPEForm (upe.js?c32f:458:36)
    at handleUPEOrderPay (upe.js?c32f:505:32)
    at HTMLFormElement.eval (upe.js?c32f:824:4)
    at HTMLFormElement.dispatch (jquery.js?ver=3.6.1:5430:27)
    at elemData.handle (jquery.js?ver=3.6.1:5234:28)
```

The bug was encountered during the first round of split UPE testing a few days ago. More details here: paJDYF-6WA#comment-15916. 

Acceptance criteria: 
* Pay for page allows users to pay with a credit card without errors when UPE is enabled and disabled

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5512

The code that handles payment on the Pay for order page listens to an event in both UPE and classic implementation. When a credit card is used, UPE implementation results in unexpected errors as it doesn't support CC but still tries to do it. This PR fixes the issue by blocking the UPE part execution on the Pay for order page when the legacy credit card is being used.

Update: additionally, this PR fixes UPE checkout from Pay for order page.

#### Testing instructions

* Pay for order with CC and confirm that there are no errors
* Pay for order with any of the UPE elements and confirm that there are no errors
* Perform a shortcode checkout purchase with both CC and UPE and confirm that everything works as expected
* Add a CC payment method in My account page and confirm, that the method is successfully added
* Add CC payment method in My account -> Payment methods page and ensure that no errors appear

-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe.js
+++ b/client/checkout/classic/upe.js
@@ -139,11 +139,11 @@ jQuery( function ( $ ) {
 	};
 
 	/**
-	 * Finds selected payment gateway and returns matching Stripe payment method for gateway.
+	 * Finds selected UPE payment gateway and returns matching Stripe payment method for gateway.
 	 *
-	 * @return {string} Stripe payment method type
+	 * @return {string} Stripe payment method type in case of type UPE, null otherwise
 	 */
-	const getSelectedGatewayPaymentMethod = () => {
+	const getSelectedUPEGatewayPaymentMethod = () => {
 		const gatewayCardId = getUPEConfig( 'gatewayId' );
 		let selectedGatewayId = null;
 
@@ -453,7 +453,7 @@ jQuery( function ( $ ) {
 	 * @return {boolean} false if incomplete.
 	 */
 	const checkUPEForm = async ( $form, returnUrl = '#' ) => {
-		const paymentMethodType = getSelectedGatewayPaymentMethod();
+		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
 		const upeComponents = gatewayUPEComponents[ paymentMethodType ];
 		const upeElement = upeComponents.upeElement;
 		const elements = upeComponents.elements;
@@ -513,7 +513,7 @@ jQuery( function ( $ ) {
 
 		try {
 			// Update payment intent with level3 data, customer and maybe setup for future use.
-			const paymentMethodType = getSelectedGatewayPaymentMethod();
+			const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
 			const upeComponents = gatewayUPEComponents[ paymentMethodType ];
 			const updateResponse = await api.updateIntent(
 				upeComponents.paymentIntentId,
@@ -563,7 +563,7 @@ jQuery( function ( $ ) {
 		blockUI( $form );
 
 		try {
-			const paymentMethodType = getSelectedGatewayPaymentMethod();
+			const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
 			const upeComponents = gatewayUPEComponents[ paymentMethodType ];
 			const { error } = await api.getStripe().confirmSetup( {
 				elements: upeComponents.elements,
@@ -769,7 +769,7 @@ jQuery( function ( $ ) {
 		.map( ( method ) => `checkout_place_order_${ method }` )
 		.join( ' ' );
 	$( 'form.checkout' ).on( checkoutEvents, function () {
-		const paymentMethodType = getSelectedGatewayPaymentMethod();
+		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
 		if ( ! isUsingSavedPaymentMethod( paymentMethodType ) ) {
 			const paymentIntentId =
 				gatewayUPEComponents[ paymentMethodType ].paymentIntentId;
@@ -789,7 +789,7 @@ jQuery( function ( $ ) {
 			return;
 		}
 		if ( ! $( '#wcpay-setup-intent' ).val() ) {
-			const paymentMethodType = getSelectedGatewayPaymentMethod();
+			const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
 			const paymentIntentId =
 				gatewayUPEComponents[ paymentMethodType ].paymentIntentId;
 			if ( isUPEEnabled && paymentIntentId ) {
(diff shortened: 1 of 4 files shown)


client/checkout/classic/upe.js
```diff
@@ -806,6 +806,13 @@ jQuery( function ( $ ) {
 
 	// Handle the Pay for Order form if WooCommerce Payments is chosen.
 	$( '#order_review' ).on( 'submit', () => {
+		// Skip handling legacy cards as UPE payment methods.
+		if (
+			'woocommerce_payments' ===
+			$( "#order_review input:checked[name='payment_method']" ).val()
+		) {
+			return;
+		}
```
mdmoore: This whole handler for the Pay for Order page is still tailored to the old UPE<sup>TM</sup>. This initial check for CC is needed and does allow CCs to skip the UPE payment flow. We have this `isWCPayChosen()` that checks if `card` is selected, so maybe we could use that here: 

```suggestion
		if ( isWCPayChosen() ) {
			return;
		}
```

Additionally, the [`isWCPayChosen()` condition on line 819](https://github.com/Automattic/woocommerce-payments/blob/7587606cecf15107080a19f2707dcce955e66628/client/checkout/classic/upe.js#L819) should be removed or possibly replaced with a check that ensures a UPE payment method is chosen. In the old UPE, there was only a single payment gateway for both CC and UPE methods, so this check was used to prevent this code from running for non-WCPay payment methods. UPE payments are currently failing on the Pay for Order page and that `isWCPayChosen()` check is the first reason. It prevents UPE payments from ever reaching the `handleUPEOrderPay()` function.  The problems don't stop there, however. Once `handleUPEOrderPay()` is reached, we get an error when updating the payment intent due to the Stripe payment method being `undefined`. A request like this is made - `http://localhost:8082/?wc-ajax=wcpay_update_payment_intent_undefined` when it should have the payment method appended, like - `http://localhost:8082/?wc-ajax=wcpay_update_payment_intent_bancontact`. The payment method is `undefined` because we're trying to grab the value of a [dom element that no longer exists here](https://github.com/Automattic/woocommerce-payments/blob/7587606cecf15107080a19f2707dcce955e66628/client/checkout/classic/upe.js#L522). We can simply replace that with `paymentMethodType` and UPE payments will work again. We're also trying to pull in `country` from a bygone dom element which is leading to the country being `undefined` as well. I'm not sure what the implications of this are or how we can pull in country from the Pay for Order page.

mdmoore (changes requested):
Thanks for creating this issue and immediately tackling it! CC payments did begin working thanks to the bail out of the UPE payment flow. Nice work there 🎉! 

I found that UPE payments are failing from the Pay for Order page due to some old UPE remnants. I left some ideas on what I think is going on and how to solve it. Let me know if you have any questons.

client/checkout/classic/upe.js
timur27: Thanks! 

> should be removed or possibly replaced with a check that ensures a UPE payment method is chosen

I'd say just removed - no need to add a check right after we performed an opposite check a few lines above. Removed `isWCPayChosen()` from this if statement. 👍 

I've also updated `form#add_payment_method` handling for adding payment methods and am using `isWCPayChosen()` there as well. May I kindly ask you to test adding CC payment methods in My account page as well? (I've also updated the PR description).

> We're also trying to pull in country from a bygone dom element which is leading to the country being undefined as well. I'm not sure what the implications of this are or how we can pull in country from the Pay for Order page.

With the old UPE on `develop` we send empty countries for this field as well - only CC provides a value, so I guess there shouldn't be any implications of this. This area should be further investigated and taken care of separately - I'll note it down and make sure we iterate here after split UPE is merged.

timur27:
Thanks, @mdmoore! We definitely can fix UPE within this PR as well. I've addressed everything. Looking forward to hearing from you.

client/checkout/classic/upe.js
mdmoore: >I'd say just removed - no need to add a check right after we performed an opposite check a few lines above. Removed isWCPayChosen() from this if statement. 👍

It does seem counterintuitive at first glance. The `isWCPayChosen()` condition that was removed was old UPEs way of preventing that block from executing for any payment method what was not part of WooCommerce Payments(CC and all UPE methods). Without a defensive check here, ensuring that a UPE payment method is selected, any selected payment gateway(e.g., Stripe, Cash on Delivery, PayPal) will be fed into `handleUPEOrderPay()` and will error. So we need something like a `isUPEMethodChosen()` here. As a quick test, I enabled Cash on Delivery and attempted to pay for an order. The UI was blocked indefinitely and an error appears in the console: 

```
Uncaught (in promise) TypeError: upeComponents is undefined
    checkUPEForm upe.js:458
    handleUPEOrderPay upe.js:505
    <anonymous> upe.js:814
    jQuery 9
    init checkout.min.js:1
    <anonymous> checkout.min.js:1
    jQuery 13
upe.js:458:21
```

>I've also updated form#add_payment_method handling for adding payment methods and am using isWCPayChosen() there as well. May I kindly ask you to test adding CC payment methods in My account page as well? (I've also updated the PR description).

Thanks for swapping that out as well. Adding CC methods from the My Account page is still working well.

>With the old UPE on develop we send empty countries for this field as well - only CC provides a value, so I guess there shouldn't be any implications of this. This area should be further investigated and taken care of separately - I'll note it down and make sure we iterate here after split UPE is merged.

That's great to know. Thanks for checking on this!

mdmoore (changes requested):
>We definitely can fix UPE within this PR as well.

Thanks for handling that in the PR! I left one more comment about preventing other payment gateways from being put into the UPE payment flow. Everything else seems to be working well!

client/checkout/classic/upe.js
timur27: @mdmoore Huge thanks for this vital and valid feedback. I didn't consider groups other than UPE and CC, which we work with daily.  I didn't think about any other payment method types, and your comment is super important. Thanks for it. Now coming back to the code: 

Given that there is more than one group of payment methods that the UPE implementation shouldn't handle, I think it is better not to check
```js
if (isSelectedPaymentMethodOfCCType() || isSelectedPaymentMethodOfBankTransferOrCashDepositEtc())
```


but to 
```js
if(isSelectedPaymentMethodNotAUPE())
```

so I just used `getSelectedGatewayPaymentMethod()` function that handles only UPE and returns `null` in any other way. Thanks to this method, we can exclude all the other payment methods. The commit is there, may I ask you to test this out and let me know, what do you think? Thanks.

mdmoore (approved):
Thanks @timur27 for addressing those issues.  Everything is testing well now.

✅ Pay for order with CC and confirm that there are no errors
✅ Pay for order with any of the UPE elements and confirm that there are no errors
✅ Perform a shortcode checkout purchase with both CC and UPE and confirm that everything works as expected
✅ Add a CC payment method in My account page and confirm, that the method is successfully added

🚢!

PR #5515, opened 2023-02-02
fix: add Link and WooPay compatibility tooltip

frosso:

Adding a tooltip that is displayed whenever the merchant attempts to enable both WooPay and Link at the same time.

I didn't run the changelog command because it's not needed - this is getting merged into a larger feature branch

Fixes https://github.com/Automattic/woocommerce-payments/issues/4720

#### Testing instructions

I recommend reviewing these changes with the "Hide whitespace changes" setting enabled - it'll show you that it's much fewer changes: https://github.com/Automattic/woocommerce-payments/pull/5515/files?w=1

- Go to the settings page
- Try to enable Link and WooPay at the same time
- A tooltip should appear
- Enabling/disabling them individually should not display the notice

With this changes, it's still possible that a merchant has both Link and WooPay enabled at the same time. But the number of merchants that had WooPay enabled is so small, that it doesn't matter.

-------------------

**Post merge**

### DIFF
--- a/client/settings/express-checkout/index.js
+++ b/client/settings/express-checkout/index.js
@@ -4,9 +4,10 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Card, CheckboxControl } from '@wordpress/components';
+import { Card, CheckboxControl, VisuallyHidden } from '@wordpress/components';
 import interpolateComponents from 'interpolate-components';
 import { useContext } from '@wordpress/element';
+import { Icon, warning } from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -24,6 +25,7 @@ import WooIcon from '../../gateway-icons/woo';
 import './style.scss';
 import WCPaySettingsContext from '../wcpay-settings-context';
 import LinkIcon from '../../gateway-icons/link';
+import Tooltip from 'components/tooltip';
 
 const ExpressCheckout = () => {
 	const [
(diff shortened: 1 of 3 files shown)


client/settings/express-checkout/test/index.test.js
```diff
@@ -50,11 +50,11 @@ describe( 'ExpressCheckout', () => {
 		const updateIsPaymentRequestEnabledHandler = jest.fn();
 
 		useGetAvailablePaymentMethodIds.mockReturnValue( [ 'link', 'card' ] );
-		useEnabledPaymentMethodIds.mockReturnValue( [ [ 'card', 'link' ] ] );
+		useEnabledPaymentMethodIds.mockReturnValue( [ [ 'card' ] ] );
```
frosso, line 53: I had to tweak the tests a bit: before, both WooPay and Link were enabled.
With these changes, this test shows only one of them enabled at the same time.

client/settings/express-checkout/test/index.test.js
```diff
@@ -172,4 +163,35 @@ describe( 'ExpressCheckout', () => {
 		const linkCheckbox = container.getByLabelText( 'Link by Stripe' );
 		expect( linkCheckbox ).not.toBeChecked();
 	} );
+
+	it( 'should prevent enabling both Link and WooPay at the same time', async () => {
```
frosso, line 167: Added a test for the added behavior.

client/settings/express-checkout/index.js
```diff
@@ -142,9 +177,13 @@ const ExpressCheckout = () => {
 							</div>
 						</li>
 					) }
-					<li className="express-checkout has-icon-border">
```
frosso, line 145: The previous changes had a `has-icon-border` classname wrapper. But it is not needed on this component. The SVGs already have a border.

client/settings/express-checkout/index.js
```diff
@@ -142,9 +177,13 @@ const ExpressCheckout = () => {
...
+					<li className="express-checkout">
 						<div className="express-checkout__checkbox">
 							<CheckboxControl
+								label={ __(
```
frosso, line 183: Added a label here, just for consistency

client/settings/express-checkout/index.js
```diff
@@ -208,16 +247,43 @@ const ExpressCheckout = () => {
...
+											<div
+												className="loadable-checkbox__icon-warning"
+												data-testid="loadable-checkbox-icon-warning"
+											>
+												<VisuallyHidden>
+													{ __(
+														'Link by Stripe cannot be enabled at checkout. Click to expand.',
+														'woocommerce-payments'
+													) }
+												</VisuallyHidden>
```
timur27, line 273: What's the purpose of these visually hidden components for both WooPay and Stripe Link?

client/settings/express-checkout/index.js
```diff
@@ -208,16 +247,43 @@ const ExpressCheckout = () => {
...
+											</div>
+										</div>
+									</Tooltip>
+								) : (
+									<CheckboxControl
+										label={ __(
+											'Link by Stripe',
+											'woocommerce-payments'
+										) }
```
timur27, line 282: I guess the `label` prop has been added to be more 'correct' technically when using this component from Blocks? I can see that we add it and hide it with the `display: none`, so I'm curious what was the reasoning behind it.

timur27 (approved):
Thanks for working on this, Francesco! Tests good and looks great as well. Approving this PR - I've just left two questions to better understand the idea behind some changes, but they shouldn't block us.

client/settings/express-checkout/index.js
frosso: I believe it has been added so that it can be more easily targeted within the tests. I noticed that some tests used it, others didn't.

client/settings/express-checkout/index.js
frosso: I added it so that screen readers can identify what the purpose of the tooltip.
The tooltip renders a `button` element around the SVG icon. The SVG icon is completely removed from the screen reader's text. So, without the `VisuallyHidden` element, a screen reader user would only "see" a `button` element with no text.
Adding a `VisuallyHidden` element adds some text for SR users, so that they can better understand the purpose of the `button` element.
An alternative could also have been to add an `aria-label` attribute on the `button` element.

I noticed that this article gives a decent overview of the problem: https://www.sarasoueidan.com/blog/accessible-icon-buttons/

PR #5520, opened 2023-02-04
Single separator for express checkout methods

### LINKED ISSUE #5518: Express Checkout buttons separators
### Description

https://github.com/Automattic/woocommerce-payments/pull/5399 and https://github.com/Automattic/woocommerce-payments/pull/5451 have been merged into the [`4720/stripe-link-integration`](https://github.com/Automattic/woocommerce-payments/tree/4720/stripe-link-integration) branch.

We can keep them there, but since the `4720/stripe-link-integration` won't get merged yet, we need to ensure that the fix is eventually merged into `develop`.

### To Reproduce

1. Enable more than one Express Checkout method at checkout. I would recommend the simplest pairing would be to enable WooPay and one of Apple Pay, Google Pay, or Link--if using the split-UPE branch (`poc/upe-instances-multiplied`) with changes from #5282 merged.
2. Add item to cart and proceed to shortcode checkout.
3. Witness the mayhem screenshotted above, as the `- OR -` separator is malevolently multiplied.

### Actual behavior

When multiple express checkout methods are registered to the shortcode checkout, the `- OR -` separator used to divide these checkout buttons from the regular checkout appears multiplied numerous times together beneath the express checkout buttons.

### Screenshots

Multiple separators appear at checkout
_Superfluous separator snafu_

### Expected behavior

Either the `- OR -` separator should appear in between each express checkout method or the express checkout methods should be grouped together in a single block--similar to the blocks checkout--with a single separator dividing them all from the rest of the checkout. I would prefer the latter solution, for the sake of consistency with blocks.

### Additional context

#5282

mdmoore:
Fixes #5518

This ensures that only a single `-OR-` separator is used with express checkout buttons such as Google Pay, Apple Pay, and WooPay. 

#### Testing instructions

5. Using the settings found for each express checkout button under its **Customize** option, disable the button on some pages and check that settings are being honored.

**Note**: This work was originally completed in https://github.com/Automattic/woocommerce-payments/pull/5399 and https://github.com/Automattic/woocommerce-payments/pull/5451 and was merged into the [4720/stripe-link-integration](https://github.com/Automattic/woocommerce-payments/tree/4720/stripe-link-integration) branch. Since that branch won't be merged yet, this PR moves the fix into the split-UPE branch.

### DIFF
--- a/includes/class-wc-payments-platform-checkout-button-handler.php
+++ b/includes/class-wc-payments-platform-checkout-button-handler.php
@@ -75,10 +75,7 @@ class WC_Payments_Platform_Checkout_Button_Handler {
 	 * @return bool
 	 */
 	public function is_woopay_enabled() {
-		if ( ! ( $this->is_platform_checkout_eligible && $this->is_platform_checkout_enabled && $this->is_platform_checkout_express_button_enabled ) ) {
-			return false;
-		}
-		return true;
+		return $this->is_platform_checkout_eligible && $this->is_platform_checkout_enabled && $this->is_platform_checkout_express_button_enabled;
 	}
 
 	/**
--- a/includes/class-wc-payments-platform-checkout-button-handler.php
+++ b/includes/class-wc-payments-platform-checkout-button-handler.php
@@ -49,6 +49,39 @@ class WC_Payments_Platform_Checkout_Button_Handler {
 	}
 
 	/**
+	 * Indicates eligibility for WooPay via feature flag.
+	 *
+	 * @var bool
+	 */
+	private $is_platform_checkout_eligible;
+
+	/**
+	 * Indicates whether WooPay is enabled.
+	 *
+	 * @var bool
+	 */
+	private $is_platform_checkout_enabled;
+
+	/**
+	 * Indicates whether WooPay express checkout is enabled.
+	 *
+	 * @var bool
+	 */
+	private $is_platform_checkout_express_button_enabled;
+
+	/**
+	 * Indicates whether WooPay and WooPay express checkout are enabled.
+	 *
+	 * @return bool
+	 */
+	public function is_woopay_enabled() {
+		if ( ! ( $this->is_platform_checkout_eligible && $this->is_platform_checkout_enabled && $this->is_platform_checkout_express_button_enabled ) ) {
+			return false;
+		}
+		return true;
+	}
+
+	/**
 	 * Initialize hooks.
 	 *
 	 * @return  void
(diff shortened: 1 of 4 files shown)


includes/class-wc-payments-platform-checkout-button-handler.php
```diff
@@ -48,6 +48,39 @@ public function __construct( WC_Payments_Account $account, WC_Payment_Gateway_WC
...
+	/**
+	 * Indicates whether WooPay and WooPay express checkout are enabled.
+	 *
+	 * @return bool
+	 */
+	public function is_woopay_enabled() {
+		if ( ! ( $this->is_platform_checkout_eligible && $this->is_platform_checkout_enabled && $this->is_platform_checkout_express_button_enabled ) ) {
+			return false;
+		}
+		return true;
```
frosso: It looks like this could be simplified:
```suggestion
		return $this->is_platform_checkout_eligible && $this->is_platform_checkout_enabled && $this->is_platform_checkout_express_button_enabled;
```

includes/class-wc-payments.php
```diff
@@ -457,6 +457,17 @@ public static function init() {
 			WC_Payments_Subscriptions::init( self::$api_client, self::$customer_service, self::get_gateway(), self::$account );
 		}
 
+		if ( WC_Payments_Features::is_express_checkout_enabled() ) {
```
frosso: I see that the `WC_Payments_Features` class is so far used for feature flags (which are set as options). But this function sets up a precedent to use it as a "utility" class for settings that might be enabled/disabled.

Since there's only one usage of `WC_Payments_Features::is_express_checkout_enabled()`, what do you think about modifying this with inline checks?
i.e.: 
```suggestion
		$is_woopay_express_checkout_enabled = WC_Payments_Features::is_woopay_express_checkout_enabled();
		if ( $is_woopay_express_checkout_enabled || self::get_gateway()->get_option( 'payment_request' ) ) {
```

And below:

```
if ( $is_woopay_express_checkout_enabled ) {
```

includes/class-wc-payments.php
```diff
@@ -1389,4 +1400,20 @@ public static function enqueue_dev_runtime_scripts() {
...
+
+	/**
+	 * Display express checkout separator only when express buttons are displayed.
+	 *
+	 * @return void
+	 */
+	public static function display_express_checkout_separator_if_necessary() {
+		$woopay          = self::$platform_checkout_button_handler->should_show_platform_checkout_button() && self::$platform_checkout_button_handler->is_woopay_enabled();
+		$payment_request = self::$payment_request_button_handler->should_show_payment_request_button();
+		$should_hide     = $payment_request && ! $woopay;
```
frosso, line 1414: Out of curiosity - why add `display: none` instead of just short-circuiting and not outputting any HTML? (maybe a comment to explain this in the codebase would also be useful)

It seems that there could be a possibility of rendering multiple elements on the page with the same `id` of `wcpay-payment-request-button-separator`. Ideally, all elements that have an `id` on a single page should have a unique `id` attribute value.

frosso (approved):
Changes look to be in line with the other PRs. Just a few minor questions/suggestions, but I don't want to delay this too much further.

includes/class-wc-payments.php
frosso: Discussed. We're good. The frontend is targeting the element, so that's why it's `display: none`.

includes/class-wc-payments-platform-checkout-button-handler.php
mdmoore: Yes, that looks much better! https://github.com/Automattic/woocommerce-payments/pull/5520/commits/ba2e62020f30d1383f86f70a26f5d1a029b77368

includes/class-wc-payments.php
mdmoore: Thanks for pointing this out! I see what you mean and it does make more sense to handle these checks inline for now. Changed in https://github.com/Automattic/woocommerce-payments/pull/5520/commits/ba2e62020f30d1383f86f70a26f5d1a029b77368.

asumaran:
@mdmoore I've noticed the order of the express checkout buttons has changed. Notice that the WooPay express checkout button is first in `WCPay 5.4.0`, which is no longer the case in the latest test release. Is that change intentional? 

| WCPay 5.4.0  | WCPay 5.5.0-test-2 |
| Screenshot 2023-02-17 at 1 44 59 PM  | Screenshot 2023-02-17 at 1 54 21 PM  |
| Screenshot 2023-02-17 at 1 45 05 PM  | Screenshot 2023-02-17 at 1 54 08 PM  |
| Screenshot 2023-02-17 at 1 45 20 PM  | Screenshot 2023-02-17 at 1 54 38 PM  |

mdmoore:
Hey @asumaran - Thanks for raising this! Order of express payment method buttons wasn't considered here so this wasn't intentional. But I see this was ironed out over here - p1676660676497189-slack-C022WMN88KG

PR #5532, opened 2023-02-09
Re-apply update customer optimization

### LINKED ISSUE #5417: Re-apply customer update optimization
## Description

As part of paJDYF-6T1-p2 we rolled back (#5383) a performance optimization we released with v4.6.0 that we should re-apply as soon as we're able.

Re-applying this requires 2 changes:

1. Make the customer update request asynchronous; and
2. Add a **synchronous** recovery flow where a customer has the wrong customer ID assigned.

Before we rolled back the optimization these were part of the same asynchronous function call which caused issues in cases where the customer ID did not exist in the current merchant account (e.g. after an account migration on the store). So now we need to split this functionality into 2.

The synchronous recovery logic will likely have to be on the payment intent creation call.

### Acceptance criteria

* Customer information is update asynchronously after a payment attempt (ideally only successful ones, but that may be out of scope for just re-applying the optimization).
* In cases where the customer ID is incorrect the payment flow should automatically recover by creating a new customer object before immediately retrying the payment.

### Testing asynchronous customer update

1. Add a product to your cart, complete a payment.
2. (Now) Note that the customer update request is sent synchronously.
3. (After fixing) The customer update request should be sent asynchronously, likely after the payment is completed.

### Testing incorrect customer ID recovery

1. Add a product to your cart, complete a payment.
2. Fix the customer ID for the current account such that it doesn't exist on the WCPayServer for this WCPay account (there are several ways to achieve this, see Dev notes).
3. Add a product to your cart, attempt to complete another payment.
4. The payment should be successful.
    - If the recovery flow is asynchronous (like it was before #5383) you'll see a "No such customer: `cus_abc123`" error notice and the payment will fail. Retrying the payment shortly after should result in a successful payment.

## Dev notes

Using ActionScheduler will have to be thoroughly considered. Running the customer update on the `shutdown` hook may be a better option. See discussion in paJDYF-6T1-p2#comment-15680 for details.

### Make sure the customer ID does not exist in Stripe

There are 2 easy ways to do this:

1. Manually edit the customer ID meta in your test store's DB to be bogus, e.g. `cus_doesnotexist`.
2. Re-onboard the store such that it's using a different WCPay account.

## Additional context

paJDYF-6T1-p2

reykjalin:
Fixes #5417 

- Make the request to update customer info asynchronous.

## Testing instructions

1. Set a breakpoint [here](https://github.com/Automattic/woocommerce-payments/blob/f2b19a3b06612b7225c3e6ad460f27d8fe205696/includes/class-wc-payment-gateway-wcpay.php#L3318).
2. Add a product to your cart, complete a payment.
3. Make sure the customer is redirected to the **Order Received** page after a successful payment even if you're still on the breakpoint.

-------------------

**Post merge**


reykjalin:
This is currently a proof of concept. If we're happy with the general approach we'll probably want to create a new `Payment_Intent_Service` where we encapsulate the retries, something along the lines of

```php
class Payment_Intent_Service {

  public function __construct( API_Client $api_client, Customer_Service $customer_service ) {
    $this->payments_api_client = $api_client;
    $this->customer_service = $customer_service;
  }

  public function create_intent( /* params passed to API client */ ) {
    try {
      $this->payments_api_client->create_and_confirm_intention(
        $converted_amount,
        $currency,
        $payment_information->get_payment_method(),
        $customer_id,
        $payment_information->is_using_manual_capture(),
        $save_payment_method_to_store,
        $payment_information->should_save_payment_method_to_platform(),
        $metadata,
        $this->get_level3_data_from_order( $order ),
        $payment_information->is_merchant_initiated(),
        $additional_api_parameters,
        $payment_methods,
        $payment_information->get_cvc_confirmation(),
        $payment_information->get_fingerprint()
      );
    } catch ( API_Exception $e ) {
      if (
        $e->get_error_code() === 'resource_missing'
        && 0 === strpos( strtolower( $e->getMessage() ), 'no such customer' )
      ) {
        // Create new customer and retry.
        $customer_data = WC_Payments_Customer_Service::map_customer_data( $order, new WC_Customer( $user->ID ) );
        $customer_id   = $this->customer_service->recreate_customer( $user, $customer_data );

        $this->payments_api_client->create_and_confirm_intention(
          $converted_amount,
          $currency,
          $payment_information->get_payment_method(),
          $customer_id,
          $payment_information->is_using_manual_capture(),
          $save_payment_method_to_store,
          $payment_information->should_save_payment_method_to_platform(),
          $metadata,
          $this->get_level3_data_from_order( $order ),
          $payment_information->is_merchant_initiated(),
          $additional_api_parameters,
          $payment_methods,
          $payment_information->get_cvc_confirmation(),
          $payment_information->get_fingerprint()
        );
      }
    }
  }

  public function update_intent( /* params passed to API client */ ) {
    try {
      $updated_payment_intent = $this->payments_api_client->update_intention(
        $payment_intent_id,
        $converted_amount,
        strtolower( $currency ),
        $save_payment_method,
        $customer_id,
        $this->get_metadata_from_order( $order, $payment_type ),
        $this->get_level3_data_from_order( $order ),
        $selected_upe_payment_type,
        $payment_country,
        $additional_api_parameters
      );
    } catch ( API_Exception $e ) {
      if (
        $e->get_error_code() === 'resource_missing'
        && 0 === strpos( strtolower( $e->getMessage() ), 'no such customer' )
      ) {
        // Create new customer and retry.
        $customer_data = WC_Payments_Customer_Service::map_customer_data( $order, new WC_Customer( $user->ID ) );
        $customer_id   = $this->customer_service->recreate_customer( $user, $customer_data );

        $updated_payment_intent = $this->payments_api_client->update_intention(
          $payment_intent_id,
          $converted_amount,
          strtolower( $currency ),
          $save_payment_method,
          $customer_id,
          $this->get_metadata_from_order( $order, $payment_type ),
          $this->get_level3_data_from_order( $order ),
          $selected_upe_payment_type,
          $payment_country,
          $additional_api_parameters
        );
      }
    }
  }
}
```

reykjalin:
I wonder if the `manage_customer_details_for_order()` call -- the one that triggers the async customer update -- should be moved to the bottom of the payment processing call?

https://github.com/Automattic/woocommerce-payments/blob/029adcab54e1fdd41dfcbe8a556db09b344a99ed/includes/class-wc-payment-gateway-wcpay.php#L961

Or maybe we need to separate the async customer update into a different function call, i.e. 1 function to get the `$user` and `$customer_id` based on info on the store and a separate call later to update the customer details asynchronously 🤔 


***

There might even be an opportunity to make the customer creation async, although that would be more complicated since we'd have to take saving payment methods into account when going through that process.

https://github.com/Automattic/woocommerce-payments/blob/029adcab54e1fdd41dfcbe8a556db09b344a99ed/includes/class-wc-payment-gateway-wcpay.php#L890-L894

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -957,10 +935,19 @@ public function process_payment_for_order( $cart, $payment_information, $additio
...
+		$customer_id = $this->customer_service->get_customer_id_by_user_id( $user->ID );
+		if ( null === $customer_id ) {
+			$customer_id = $this->create_customer_using_order( $order, $user );
+		} else {
+			// No need to update the customer object if we just created it.
+			$customer_details_options = [
+				'is_woopay' => filter_var( $metadata['paid_on_woopay'] ?? false, FILTER_VALIDATE_BOOLEAN ),
+			];
+			$this->update_customer_with_order_data_on_shutdown( $customer_id, $order, $customer_details_options );
+		}
```
reykjalin: This specific if-statement might be better as a `create_or_update_customer` function?

reykjalin:
I've hit a snag with reworking this that makes this difficult to complete: when the store has WooPay enabled and the customer chooses to not use WooPay and instead pay on the store while the customer ID associated with them is wrong - e.g. `cus_doesnotexist` as is mentioned in the testing instructions - the immediate retry implemented in the PR doesn't work:

image

To fully understand why, I need to explain the flow of payments when WooPay is enabled:

1. Customer comes to the checkout page and enters payment details.
2. The payment method is saved on the WCPay Platform account, not the merchant account.
3. Payment intent is created following 1 of 2 paths:
    1. If the customer opts into WooPay we create a customer object in the platform account and save the payment method there.
    2. If the customer does not opt into WooPay the payment method is cloned to the merchant account and used from there.

The problem we're running into is a function of 2 things:

1. The payment method used on the merchant store to initiate the payment **is a platform-level payment method**; and
2. When that payment method is cloned to the merchant account it's consumed **and can't be used again**.

Because we don't get any information on the cloned payment method with the `no such customer` error that can happen when the customer ID is incorrect the retry functionality will try again using **the platform level payment method**. When that happens that specific payment method has been consumed in the cloning process and no longer exists. This causes the payment to fail.

So far I can think of a couple ways to address this, but I'm not sure any of them are really good:

1. Always attach **the platform level payment method** to a new customer object in the platform account making it so we can attempt to clone it multiple times; or
2. Handle the error on the server side as suggested by @jrodger in https://github.com/Automattic/woocommerce-payments/issues/5417#issuecomment-1424295302 where we have access to the cloned payment method.

(1) is has data storage implications. Going with that solution would be complicated by the fact that we'd need to add an async job to purge that data from the platform account after processing a payment. And since it has to be async we'd likely have to store these temporary objects in a DB somewhere, further complicating the flow.

(2) would mean we'd have to add functionality to create the customer object from the server using information from the payment intent request and then add some tweaks to update the customer on the merchant store with the result, adding complexity but it seems manageable.

Note that we must create platform level payment methods when WooPay is enabled. There is no way to switch to the merchant account at any point - e.g. when the customer toggles "save my information" - because the whole Stripe SDK object is initialized on the page load and that determines where the payment method is created. Reloading that would unmount and remount the payment element causing any information the customer may have entered there to vanish.

***

We could try a different approach; manually add the cloned payment method to the error response and use that if it's present for the retry on the merchant store. It'd mean minor changes on the server, but would add more branches across the server and client repos that we have to keep in sync increasing the complexity.

***

I think I like (2) the most and handling this on the server. It saves us a round trip on requests, has no effect on how we store and purge data, and 99% of the changes are on the server instead of spreading the changes across the server and client repos.

I'm very much open to feedback on this though. @ricardo may have some additional insights here since he worked on the payment flow when WooPay is enabled and may be able to see a solution I can't think of.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -17,15 +17,13 @@
...
 use WCPay\Logger;
-use WCPay\Payment_Information;
 use WCPay\Session_Rate_Limiter;
 use WC_Order;
 use WC_Payments;
 use WC_Payments_Account;
 use WC_Payments_Action_Scheduler_Service;
 use WC_Payments_API_Client;
 use WC_Payments_Customer_Service;
-use WC_Payments_Explicit_Price_Formatter;
```
reykjalin, line 28: These imports were unused so I decided to remove them

includes/class-wc-payment-gateway-wcpay.php
reykjalin: Created and used the helper function in d27692549b43ccb509771c22eda34bde1175655d, f2186e12e4201f46fcf165e43b2459c6c126b37a, and f2b19a3b06612b7225c3e6ad460f27d8fe205696.

reykjalin:
Hmmm @marcinbot the shutdown hook doesn't seem to run before the response is sent to the frontend? 🤔 

When I put a breakpoint in [the function run on the hook](https://github.com/Automattic/woocommerce-payments/blob/4b922e84b1d8b96947f1dff89d8faa58e7915062/includes/class-wc-payment-gateway-wcpay.php#L3318) the customer isn't redirected until I hit "continue" in the debugger. Can you check if that's just on my end or if it also happens for you? 🤔 

I thought I tested this a while ago and got it to work right but for whatever reason it's not working now, not sure why.

jrodger:
I've removed Marcin and myself as reviewers for now. Add us back when ready and we'll get pinged on Slack (or add us back if this is ready for another review already, sorry if so).

reykjalin:
That's perfectly fine @jrodger, thanks for doing that. I thought I'd get back to this sooner, but haven't had the time yet 😕

reykjalin:
Closing this due to how old it is. We should look into this with fresh eyes, I think.

PR #5537, opened 2023-02-11
Fix SEPA payment methods to work with subscription products

timur27:
Fixes #

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/includes/payment-methods/class-upe-split-payment-gateway.php
+++ b/includes/payment-methods/class-upe-split-payment-gateway.php
@@ -8,6 +8,7 @@
 namespace WCPay\Payment_Methods;
 
 use Exception;
+use Square\Models\Payment;
 use WC_Payments_Features;
 use WCPay\Constants\Order_Status;
 use WCPay\Constants\Payment_Method;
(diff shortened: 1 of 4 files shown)


includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -93,6 +94,20 @@ public function __construct(
 			$this->id           = self::GATEWAY_ID . '_' . $this->stripe_id;
 			$this->method_title = "WooCommerce Payments ($this->title)";
 		}
+
+		add_filter( 'woocommerce_available_payment_gateways', [ $this, 'add_sepa_gateway' ], 100, 1 );
```
timur27: `100` was just a random high number to make sure the callback is being executed as the last one - feel free to improve on this

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -93,6 +94,20 @@ public function __construct(
...
+	}
+
+	/**
+	 * Adds SEPA gateway
+	 *
+	 * @param array $gateways                  - Filtered gateways.
+	 */
+	public function add_sepa_gateway( $gateways ) {
+		if ( WC_Payments_Features::is_upe_split_enabled() && 'sepa_debit' === $this->stripe_id ) {
```
timur27: We probably need to extend this `if` so that it checks for the subsccription cart only, because otherwise the sepa gateway might be added twice (in case of the non subscription product)

PR #5556, opened 2023-02-15
Re-enable and fix UPE shopper E2E tests

mdmoore:
Fixes 1650-gh-automattic/woopay

These e2e tests were skipped [here](https://github.com/Automattic/woocommerce-payments/commit/3d0b8b0f150fa8daee7901d96057ea7093f98500). This ensures that zip code is provided when filling details on the Add Payment Method page.

### DIFF
--- a/tests/e2e/specs/upe-split/shopper/shopper-upe-enabled-all-flows.spec.js
+++ b/tests/e2e/specs/upe-split/shopper/shopper-upe-enabled-all-flows.spec.js
@@ -22,7 +22,7 @@ const MIN_WAIT_TIME_BETWEEN_PAYMENT_METHODS = 20000;
 const sepaPaymentMethod = '#inspector-checkbox-control-8';
 const card = config.get( 'cards.basic' );
 
-describe.skip( 'Enabled Split UPE', () => {
+describe( 'Enabled Split UPE', () => {
 	beforeAll( async () => {
 		await merchant.login();
 		await merchantWCP.activateSplitUpe();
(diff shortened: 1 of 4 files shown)


tests/e2e/specs/upe-split/shopper/shopper-upe-enabled-all-flows.spec.js
```diff
@@ -48,73 +36,6 @@ describe.skip( 'Enabled Split UPE', () => {
 		await merchant.logout();
 	} );
 
-	describe( 'Blocks checkout', () => {
```
frosso, line 51: What was the reason for removing the Blocks Checkout tests from the Split UPE and legacy UPE? Are these tests covered somewhere else?

frosso (approved):
Changes seem to be passing - just a question about the removal of the tests on the blocks checkout.

tests/e2e/specs/upe-split/shopper/shopper-upe-enabled-all-flows.spec.js
mdmoore: All Blocks Checkout tests are broken/failing on `develop` at the moment, so it seems that tests involving Blocks are removed or otherwise suppressed. This is mentioned in this thread p1676324326527639/1676321242.264789-slack-C048UEB37MZ. I think this is the reason for the `SKIP_WC_BLOCKS_TESTS` flag as well.

frosso: Thanks for clarifying! Could you please log an issue for [the post-rollout epic](https://github.com/Automattic/woopay/issues/1631) to ensure we come back to this and fix the tests on the blocks checkout?

PR #5575, opened 2023-02-17
Replace existing manual scripts asset loading with dependency extraction plugin

### LINKED ISSUE #5359: Generated script dependencies are not used for some of them
Even if we load most of our dependencies in every page, we can end with some hard to replicate issues like #5347. Due to script dependencies not being loaded before.

We already have a working solution for this. Which relies on `@woocommerce/dependency-extraction-webpack-plugin` that generates and `*.asset.php` file with the script dependencies from WP and WC. But looks like we're not loading it in some cases.

We should always do something like:
https://github.com/Automattic/woocommerce-payments/blob/ef11d7d9d587a3ba3de7e0cfbcd4455f4750f8fc/includes/class-wc-payments-platform-checkout-button-handler.php#L102-L104

Instead of adding them manually like:
https://github.com/Automattic/woocommerce-payments/blob/2335fc77db2a79b323a723a8bf2c6967bb7e010f/includes/payment-methods/class-upe-payment-gateway.php#L142-L151

Getting the dependencies from the generated asset file, will allow us to automatically have the dependencies updated, without adding them manually.

zmaglica:
Fixes #5359

To load script dependencies, we used `@woocommerce/dependency-extraction-webpack-plugin` that generates *.asset.php file with the script dependencies from WP and WC. In some parts of code, we are still using the "traditional" or "default" way of loading scripts. This PR address this issue by replacing these parts of code with assets loading provided by the NPM plugin mentioned above.

#### Testing instructions

Make sure that all scripts are loaded and make sure that E2E tests and JS tests are loading fine.

*

-------------------

**Post merge**


jessy-p:
Tested and working well

jessy-p (approved):

jessy-p (commented):
Code coverage keeps failing, it is passing on other branches, please check

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -619,11 +619,14 @@ public function register_scripts() {
 		if ( $this->supports( 'tokenization' ) ) {
 			$script_dependencies[] = 'woocommerce-tokenization-form';
 		}
-
+		$script_src_url               = plugins_url( 'dist/checkout.js', WCPAY_PLUGIN_FILE );
```
naman03malhotra: Looks to me like a repeated piece of code in all the files. Can this be converted to a function or class?

zmaglica: Pushed new code that replaced all these manual imports with function.

includes/class-wc-payments.php
```diff
@@ -959,6 +959,32 @@ public static function get_file_version( $file ): string {
...
+	 *
+	 * @param string $handler Script handler.
+	 * @param string $script Script name.
+	 * @param array  $dependencies Additional dependencies.
+	 * @param bool   $in_footer  Load script in footer.
+	 * @param string $version Load with specific version.
+	 *
+	 * @return void
+	 */
+	public static function load_script_with_dependencies( string $handler, string $script, array $dependencies = [], bool $in_footer = true, string $version = null ) {
```
RadoslavGeorgiev: Should we rename this to `register_script_with_dependencies()`? Not only would this correspond better with what the method does, but will also not mislead anyone into thinking that the script would actually get loaded.

RadoslavGeorgiev: Also, let's just get rid of the `in_footer` parameter here altogether. There isn't a single call to this method with `$in_footer = false`, and IMO there shouldn't be any further. We should be fine making it a constant true.

includes/admin/class-wc-payments-admin.php
```diff
@@ -444,6 +444,7 @@ public function add_payments_menu() {
 	 * Register the CSS and JS scripts
 	 */
 	public function register_payments_scripts() {
+		WC_Payments::load_script_with_dependencies( 'WCPAY_DASH_APP', 'dist/index' );
```
RadoslavGeorgiev: Nice addition here, but the left-overs are still there.

includes/class-wc-payments-upe-split-blocks-payment-method.php
```diff
@@ -17,13 +17,8 @@ class WC_Payments_UPE_Split_Blocks_Payment_Method extends WC_Payments_Blocks_Pay
...
 		$classic_blocks_scripts = parent::get_payment_method_script_handles();
 
-		wp_register_script(
-			'WCPAY_BLOCKS_UPE_SPLIT_CHECKOUT',
-			plugins_url( 'dist/upe-split-blocks-checkout.js', WCPAY_PLUGIN_FILE ),
-			[ 'stripe' ],
-			'1.0.1',
-			true
-		);
+		WC_Payments::load_script_with_dependencies( 'WCPAY_BLOCKS_UPE_SPLIT_CHECKOUT', 'dist/upe-split-blocks-checkout', [ 'stripe' ], true, '1.0.1' );
```
RadoslavGeorgiev: I believe that the `1.0.1` version here was only added, because it's a required parameter, and someone wanted to get to `in_footer = true`. As I mentioned in another comment, let's remove the footer parameter altogether, and then neither the version, nor the boolean would be needed here.

RadoslavGeorgiev (changes requested):
Great work overall, left a couple of small comments.

includes/admin/class-wc-payments-admin.php
zmaglica: Addressed in c56251cc14c44e5ca4dd8b249a562c53c322afe6

includes/class-wc-payments-upe-split-blocks-payment-method.php
zmaglica: Fixed in 8c9a994cce0c7154f43e9cd005e9bbba0f99f7dd

includes/class-wc-payments.php
zmaglica: Good call on function name. The function name has been updated. More info can be found in this commit: 09e0f1e0580129272a70cfda3277342b40a4f9c0

htdat (commented):
LGTM too though I've found out some other places would use the new util method `register_script_with_dependencies` by searching and reviewing `wp_register_script` calls under the `/includes` folder.

https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/class-wc-payments-blocks-payment-method.php#L70-L76

https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/class-wc-payments-upe-blocks-payment-method.php#L33-L39


For this one, I think `register_script_with_dependencies` will need to be updated to use the `version` info from `*.asset.php` files

https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/multi-currency/CurrencySwitcherBlock.php#L55-L69

zmaglica:
> LGTM too though I've found out some other places would use the new util method `register_script_with_dependencies` by searching and reviewing `wp_register_script` calls under the `/includes` folder.
> 
> https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/class-wc-payments-blocks-payment-method.php#L70-L76
> 
> https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/class-wc-payments-upe-blocks-payment-method.php#L33-L39
> 
> For this one, I think `register_script_with_dependencies` will need to be updated to use the `version` info from `*.asset.php` files
> 
> https://github.com/Automattic/woocommerce-payments/blob/062fab3273445753a700909c7a6413307e598155/includes/multi-currency/CurrencySwitcherBlock.php#L55-L69

Addressed in 83f2836651897e764fbc76c57d3b4e9a9073d55c

htdat (approved):
🚢 it (except that @RadoslavGeorgiev and @jessy-p have other opinions).

RadoslavGeorgiev (approved):
LGTM :shipit: 👏

PR #5576, opened 2023-02-17
Fix express checkout button order

brettshumaker:

#5520 introduced changes to the separator displayed between express checkout buttons and the CTA button (or checkout form depending on the page) but the order of the buttons changed. This PR fixes the order so that the WooPay button is listed first.

| Before | After |
|-----|-----|
| after | after |

#### Testing instructions

* Apply patch
* Enable WooPay as well as Apple Pay and Google Pay.
* Visit any page that shows the express payment buttons (single product, cart, checkout) and ensure that WooPay is listed first. 
-------------------

**Post merge**

### DIFF
--- a/client/checkout/blocks/index.js
+++ b/client/checkout/blocks/index.js
@@ -65,8 +65,6 @@ registerPaymentMethod( {
 	},
 } );
 
-registerExpressPaymentMethod( paymentRequestPaymentMethod( api ) );
-
 // Call handlePlatformCheckoutEmailInput if platform checkout is enabled and this is the checkout page.
 if ( getConfig( 'isPlatformCheckoutEnabled' ) ) {
 	if (
@@ -79,6 +77,8 @@ if ( getConfig( 'isPlatformCheckoutEnabled' ) ) {
 		registerExpressPaymentMethod( wooPayExpressCheckoutPaymentMethod() );
 	}
 }
+
+registerExpressPaymentMethod( paymentRequestPaymentMethod( api ) );
 
 window.addEventListener( 'load', () => {
 	enqueueFraudScripts( getConfig( 'fraudServices' ) );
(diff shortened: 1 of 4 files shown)


asumaran:
@brettshumaker, what about the buttons in the checkout block? It hasn't changed compared to the previous WCPay version, but I noticed the WooPay button appears after GPay.

Screenshot 2023-02-17 at 2 36 24 PM

asumaran (approved):
Looks good! Tests as expected. I left a comment that probably belongs to the WC Blocks extension.

brettshumaker:
> what about the buttons in the checkout block? It hasn't changed compared to the previous WCPay version, but I noticed the WooPay button appears after GPay.

@asumaran  Good catch! I looked into this a bit and it looks like if I change the order in which we register the express payment buttons, I can get WooPay to show up first. Essentially moving this [block of code](https://github.com/Automattic/woocommerce-payments/blob/e04ea0dd455d48123dd3de87d83d520b0deed6ee/client/checkout/blocks/index.js#L70-L81) above line 68 in that file. I'm going to see if there's a better way to do this, but that did work for me.

client/checkout/blocks/index.js
```diff
@@ -65,8 +65,6 @@ registerPaymentMethod( {
 	},
 } );
 
-registerExpressPaymentMethod( paymentRequestPaymentMethod( api ) );
-
 // Call handlePlatformCheckoutEmailInput if platform checkout is enabled and this is the checkout page.
 if ( getConfig( 'isPlatformCheckoutEnabled' ) ) {
```
brettshumaker, line 69: @asumaran I didn't see a way to explicitly set the ordering of these methods with `registerExpressPaymentMethod()` so I went ahead with changing the order in which we register them. LMK your thoughts on this approach.

asumaran: I see no harm on doing this.

asumaran (approved):
LGTM!

PR #5585, opened 2023-02-20
Removed PHP ENUM dependency due to conflicts with multiple PHP versions

### LINKED ISSUE #5533: Current enum library in use is 1.7.7, causing errors when PHP version is 8.1+
### Describe the bug

[The message/error can be seen here](https://gist.github.com/anastas10s-afk/a92eda6a03e0567c362b5899afc278e0), literally overtaking the whole screen.

### To Reproduce

Suffices to have WooCommerce & WooCommerce Payments active on a site, where the PHP version is 8.1+, for the errors to start appearing.

### Actual behavior

WooCommerce Payments apparently [returns an error, with PHP versions 8.1.14 & 8.2.1](https://wordpress.org/support/topic/status-on-updating-depreciated-php-call-please/#post-16389912), in my staging installation.
[Nothing I could find on GitHub](https://github.com/Automattic/woocommerce-payments/issues?q=is%3Aissue+in%3Atitle+php+is%3Aopen).

### Screenshots

https://snipboard.io/vjslMy.jpg

### Expected behavior

Have the enum library updated, please.
The enum library currently in use is 1.7.7: 
https://github.com/Automattic/woocommerce-payments/blob/develop/composer.lock

While its newer versions contain a fix: 
https://github.com/myclabs/php-enum/releases

### Additional context

- https://wordpress.org/support/topic/error-with-php-8-1/

zmaglica:
Fixes #5533 

In newer versions of PHP (8.1 and 8.2), the enum package we use could throw warnings and errors because the package uses some deprecated functions which cause some errors and warnings. To overcome this issue, the ENUM package has been removed and implemented a built-in solution that mimics the package functionalities.

#### Testing instructions

Make sure that all tests are passing.

*

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-5533-remove-enum-dependecy
+++ b/changelog/fix-5533-remove-enum-dependecy
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fix
 
-Removed PHP ENUM dependency due to conflicts with multiple PHP versions.
+Replace PHP dependency myclabs/php-enum with a built-in solution due to conflicts with multiple PHP versions.
--- a/changelog/fix-5533-remove-enum-dependecy
+++ b/changelog/fix-5533-remove-enum-dependecy
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Removed PHP ENUM dependency due to conflicts with multiple PHP versions.
--- a/includes/constants/class-base-constant.php
+++ b/includes/constants/class-base-constant.php
@@ -61,7 +61,7 @@ abstract class Base_Constant {
 	 * @return bool
 	 */
 	final public function equals( $variable = null ): bool {
-		return $this->get_value() === $variable->get_value() && static::class === \get_class( $variable );
+		return $variable instanceof Base_Constant && $this->get_value() === $variable->get_value() && static::class === \get_class( $variable );
 	}
 
 	/**
--- a/includes/constants/class-base-constant.php
+++ b/includes/constants/class-base-constant.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * Class Base_Constant
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Constants;
+
+use ReflectionClass;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+
+/**
+ * Base constant class to hold common logic for all constants.
+ */
+abstract class Base_Constant {
+
+	/**
+	 * Enum value
+	 *
+	 * @var mixed
+	 */
+	protected $value;
+
+	/**
+	 * Class constructor.
+	 *
+	 * @param mixed $value Constant from class.
+	 */
+	public function __construct( $value ) {
+		if ( $value instanceof static ) {
+			$value = $value->get_value();
+		}
+		$this->value = $value;
+	}
+
+	/**
+	 * Get enum class value.
+	 *
+	 * @return mixed
+	 */
+	public function get_value() {
+		return $this->value;
+	}
+
+	/**
+	 * Compare to enums.
+	 *
+	 * @param mixed $variable Constant object to compare.
+	 *
+	 * @return bool
+	 */
+	final public function equals( $variable = null ): bool {
+		return $variable instanceof self && $this->get_value() === $variable->get_value() && static::class === \get_class( $variable );
+	}
+
+	/**
+	 * Find constant in class by value.
+	 *
+	 * @param string $value Value to find.
+	 *
+	 * @return int|string
+	 * @throws \InvalidArgumentException
+	 */
+	public static function search( string $value ) {
+		$class = new ReflectionClass( static::class );
+		$key   = array_search( $value, $class->getConstants(), true );
+		if ( false === $key ) {
+			throw new \InvalidArgumentException( "Constant with value '$value' does not exist." );
+		}
+		return $key;
+	}
+
+	/**
+	 * Used to created enum from constant names like CLASS::ConstantName().
+	 *
+	 * @param string $name Name of property or function.
+	 * @param array  $arguments Arguments of static call.
+	 *
+	 * @return static
+	 * @throws \InvalidArgumentException
+	 */
+	public static function __callStatic( $name, $arguments ) {
+		$class         = static::class;
+		$constant_name = $class . '::' . $name;
+		if ( ! defined( $constant_name ) ) {
+			throw new \InvalidArgumentException( "Constant with name '$name' does not exist." );
+		}
+
+		return new static( $name );
+	}
+
+	/**
+	 * Get real enum value.
+	 *
+	 * @return mixed|string
+	 */
+	public function __toString() {
+		return constant( \get_class( $this ) . '::' . $this->value );
+	}
+}
--- a/includes/constants/class-base-constant.php
+++ b/includes/constants/class-base-constant.php
@@ -30,11 +30,17 @@ abstract class Base_Constant {
 	 * Class constructor.
 	 *
 	 * @param mixed $value Constant from class.
+	 * @throws \InvalidArgumentException
 	 */
 	public function __construct( $value ) {
 		if ( $value instanceof static ) {
 			$value = $value->get_value();
+		} else {
+			if ( ! defined( static::class . "::$value" ) ) {
+				throw new \InvalidArgumentException( "Constant with name '$value' does not exist." );
+			}
 		}
+
 		$this->value = $value;
 	}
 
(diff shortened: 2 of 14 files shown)


includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * Compare to enums.
+	 *
+	 * @param mixed $variable Constant object to compare.
+	 *
+	 * @return bool
+	 */
+	final public function equals( $variable = null ): bool {
+		return $variable instanceof self && $this->get_value() === $variable->get_value() && static::class === \get_class( $variable );
```
marcinbot: Shouldn't this be `$variable instanceof static` instead of `self`? And wouldn't that be the same as `static::class === \get_class( $variable )` for the purposes of this function?

includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,105 @@
...
+		return new static( $name );
+	}
+
+	/**
+	 * Get real enum value.
+	 *
+	 * @return mixed|string
+	 */
+	public function __toString() {
+		return constant( \get_class( $this ) . '::' . $this->value );
```
marcinbot: I think `$this->get_value()` would be better in case we ever want to serialize the constants and pass them to JS.

includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,105 @@
...
+	/**
+	 * Class constructor.
+	 *
+	 * @param mixed $value Constant from class.
+	 */
+	public function __construct( $value ) {
+		if ( $value instanceof static ) {
+			$value = $value->get_value();
+		}
+		$this->value = $value;
```
marcinbot, line 44: I think there should be some validation to check that the `$value` is a valid constant on the class. We could use the `search` function for that.

includes/constants/class-base-constant.php
zmaglica: Yes, you are correct. I will remove instance of check.

includes/constants/class-base-constant.php
zmaglica: I think that a better option would be to move the constant validation form _callStatic to the constructor, so we don't have to use the search function. The search function is used to get a constant name from a constant value, so in this case it will not be good option to use.

marcinbot (commented):
The code looks good, but haven't tested everything yet.

includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Compare to enums.
+	 *
+	 * @param mixed $variable Constant object to compare.
+	 *
+	 * @return bool
+	 */
+	final public function equals( $variable = null ): bool {
+		return $this->get_value() === $variable->get_value() && static::class === \get_class( $variable );
```
htdat:
```suggestion
		return is_object( $variable ) && static::class === \get_class( $variable ) && $this->get_value() === $variable->get_value();
```

Tight up the conditions here: move the check class name as the second condition and a check with `is_object`? Otherwise, comparing with a random value (not created by this Enum) can yell errors. Then add the relevant tests too. 

You can apply this diff for the test file https://gist.github.com/htdat/c874e552412fc2c8fd85a33b0c91295a and run PHP tests to see these errors ` npm run test:php -- --filter=Base_Constant_Test`.

includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,106 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+
+/**
+ * Base constant class to hold common logic for all constants.
+ */
+abstract class Base_Constant {
```
htdat, line 20: Nitpick: Since we're really writing a base class for enums, shall we call this `Enum` rather than `Base_Constant` like the previously imported package?

tests/unit/test-class-base-constant.php
```diff
@@ -0,0 +1,52 @@
...
+class Base_Constant_Test extends WCPAY_UnitTestCase {
+
+	public function test_base_constant_will_create_constant() {
+		$class = Payment_Method::BASC();
+		$this->assertInstanceOf( Payment_Method::class, $class );
+		$this->assertSame( $class->get_value(), 'BASC' );
+		$this->assertSame( (string) $class, 'bacs_debit' );
+	}
+
+	public function test_get_const_name_by_value() {
```
htdat: May remove this one as it's a duplication of `test_class_will_return_const_name_if_searched_by_correct_const_value`.

includes/constants/class-base-constant.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Class constructor.
+	 *
+	 * @param mixed $value Constant from class.
+	 * @throws \InvalidArgumentException
+	 */
+	public function __construct( $value ) {
+		if ( $value instanceof static ) {
+			$value = $value->get_value();
```
htdat, line 37: Nitpick: I am not sure if it's only me but I found out `$value` and `get_value` confusing. We're actually referring to constant names here, so I think they would be `$name` and `get_name`? 

For example, here `BASC` is the constant name while `bacs_debit` is the constant value. 

https://github.com/Automattic/woocommerce-payments/blob/ea1865e88eae4c152fe4125ff29b5bd6b89a38c2/includes/constants/class-payment-method.php#L22

htdat (changes requested):
Besides my inline comments, I manually tested some flows with xDebug enabled to ensure that the relevant methods from the base class are covered: 

- Check out with non-UPE flow for testing `__callStatic` ([code ref](https://github.com/Automattic/woocommerce-payments/blob/ea1865e88eae4c152fe4125ff29b5bd6b89a38c2/includes/class-wc-payment-gateway-wcpay.php#L822-L823)), and `equals` ([code ref](https://github.com/Automattic/woocommerce-payments/blob/ea1865e88eae4c152fe4125ff29b5bd6b89a38c2/includes/class-wc-payment-gateway-wcpay.php#L1079-L1079))
- Enable MultiCurrency with additional payment methods such as `giropay` for testing `search` ([code ref](https://github.com/Automattic/woocommerce-payments/blob/ea1865e88eae4c152fe4125ff29b5bd6b89a38c2/includes/multi-currency/PaymentMethodsCompatibility.php#L69-L69))

For manual testing with GlobalStep, I think asking them to check wp-content/debug.log and ensuring no relevant error are quite enough.

includes/constants/class-base-constant.php
zmaglica: `is_object` validation is not necessary because get_class will return the name of the class, even for primitive types like int and string. The only thing that we could check is that the `$variable` is an instance of a base constant, so we don't have an issue with get_value() not existing on the passed variable.

includes/constants/class-base-constant.php
zmaglica: The reason why I put it as a base const since we're only using this for constants, and if you look from a different perspective, the enums are constants with steroids, and I wasn't sure where to put enum class since it felt wired to put it inside const folder with "Enum" name.

tests/unit/test-class-base-constant.php
zmaglica: Noted. I will fix this.

includes/constants/class-base-constant.php
zmaglica: This approach is a fallback legacy to Enum package we used before, and I mimic the function calls to make it backward compatible.

includes/constants/class-base-constant.php
htdat: > because get_class will return the name of the class, even for primitive types like int and string

This seems not... true. https://3v4l.org/tFM3p

> The only thing that we could check is that the $variable is an instance of a base constant, so we don't have an issue with get_value() not existing on the passed variable.

I think this works too.

changelog/fix-5533-remove-enum-dependecy
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Removed PHP ENUM dependency due to conflicts with multiple PHP versions.
```
htdat: Maybe also update the PR title a bit to reflect that we are not just going to "remove" the package but also add a replacement.

```suggestion
Replace PHP dependency myclabs/php-enum with a built-in solution due to conflicts with multiple PHP versions.
```

includes/constants/class-base-constant.php
zmaglica: Fixed in 7bc2338108d0882d11be026767037bd102509de7

tests/unit/test-class-base-constant.php
zmaglica: Fixed in 3caf6def25a21932cfa2e70f087a52edb7d07671

htdat (approved):
I am pre-approving PR. 
But it's worth double-checking to see why E2E tests failed. 

From quick looking at a run failure https://github.com/Automattic/woocommerce-payments/actions/runs/4321710982, this seems irrelevant to this change.

PR #5586, opened 2023-02-20
Release automation - Run smoke tests when zip file is created

### LINKED ISSUE #5528: Release Automation - Run E2E tests after building zip file
### Description

At the moment, smoke tests are run manually against the zip file created as part of release automation process. To improve dev efficiency and reduce incidents, we can run E2E tests against the zip file created, which includes most of the tests ran as part of smoke tests.

### Acceptance criteria

E2E tests should run immediately after the zip file is created as part of release automation.

achyuthajoy:
Fixes #5528 

The idea behind this PR is to automatically run smoke tests when a release zip is built. Since we don't have dedicated smoke tests suite at the moment, we'll be running the E2E - Pull Request test suite against the release zip file until dedicated smoke tests are available.

This PR will:
1. Run smoke tests with [build zip file workflow](https://github.com/Automattic/woocommerce-payments/actions/workflows/build-zip-file.yml). Allows skipping smoke tests if required.
2. Builds the zip file & run smoke tests with [Release - Create PR to trunk](https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/release-pr.yml). Allows skipping smoke tests if required.

#### Testing Build zip file  workflow

* Trigger a new manual action for [Build zip file](https://github.com/Automattic/woocommerce-payments/actions/workflows/build-zip-file.yml).
* Confirm that a checkbox `Skip running smoke tests` is displayed when the branch is selected.
* Trigger a new test against this branch without checking `Skip running smoke tests` and confirm that E2E - Pull request tests are run after build zip job is complete.
* Trigger a new test against this branch with `Skip running smoke tests` checked and confirm that E2E - Pull request tests are skipped.

#### Testing Release - Create PR to trunk  workflow

* Trigger a new manual action for [Release - Create PR to trunk](https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/release-pr.yml).
* Confirm that a checkbox `Skip running smoke tests` is displayed when the branch is selected.
* Trigger a new test against this branch with a specific release version & without checking `Skip running smoke tests`. Confirm that E2E - Pull request tests are run after build zip job is complete.
* Trigger a new test against this branch with a specific release version & `Skip running smoke tests` checked. Confirm that E2E - Pull request tests are skipped.

-------------------

**Post merge**
- [ ] Update release train model to include the changes introduced on this PR - @achyuthajoy


.github/workflows/build-zip-file.yml
```diff
@@ -2,15 +2,40 @@ name: "Build zip file"
 
 # This action will run when it is triggered manually.
 on:
+  workflow_call:
+    inputs:
+      skip-build-zip:
```
jbordonado: In which use case would we want to skip building the zip file within the `build-zip-file` workflow?
It feels like we don't need such input

.github/workflows/build-zip-file.yml
```diff
@@ -2,15 +2,40 @@ name: "Build zip file"
...
+        type: boolean
+        required: false
+        default: false
+        description: "Skip running build zip"
+      skip-smoke-tests:
+        type: boolean
+        required: false
+        default: false
+        description: "Skip running smoke tests"
+      repo-branch:
```
jbordonado: We can just checkout the relevant branch in the caller workflows before calling this one (and remove the `ref` param l.37/38). If we can avoid piling up inputs (and having to propagate some to child workflows) , that would be best in my opinion

.github/workflows/build-zip-file.yml
```diff
@@ -19,16 +44,26 @@ jobs:
...
           name: "woocommerce-payments"
-          path: zipfile
-          retention-days: 7
+          path: release
+          retention-days: 14
+
+  smoke-tests:
+    name: "Smoke tests"
+    needs: build-zip
+    if: ${{ inputs.skip-build-zip != true && inputs.skip-smoke-tests != true }}
```
jbordonado: Is `if: ${{ inputs.skip-smoke-tests != true }}` not enough?

Since the job depends on `build-zip`, which is skipped when `inputs.skip-build-zip != true`

if job B depends on job A, I would expect that job B is skipped automatically if job A is also skipped. Is that not the case?

.github/workflows/build-zip-file.yml
```diff
@@ -19,16 +44,26 @@ jobs:
...
         uses: ./.github/actions/build
 
-      - name: "Unzip the file (prevents double zip issue)"
+      - name: "Add file size notice & delete build zip"
         run: |
-          echo ":information_source: Unzipping to a folder to prevent double-zip issue ([known limitation](https://github.com/actions/upload-artifact#zipped-artifact-downloads)). Increase in artifact size is expected since we're uploading the folder." >> $GITHUB_STEP_SUMMARY
-          unzip ${{ steps.build_plugin.outputs.release-filename }} -d zipfile
-      
+          echo ":information_source: Ignore the artifact size mentioned since GitHub calculates the size of the source folder instead of the zip file created." >> $GITHUB_STEP_SUMMARY
+          rm -rf release/${{ steps.build_plugin.outputs.release-filename }}
```
jbordonado: Is this really needed?

I think the `npm run build` command already does that, so I'm not sure:
https://github.com/Automattic/woocommerce-payments/blob/d276e3a4efed3d380377cee9d33356ed541b265e/package.json#L23

.github/workflows/create-pre-release.yml
```diff
@@ -67,9 +79,19 @@ jobs:
...
           gh release create $VERSION --notes "$RELEASE_NOTES" --title "$TAG_MESSAGE" --target "develop" "--prerelease" $FILENAME
-          
+
       - name: "Trigger translations update on GlotPress"
         uses: ./.github/actions/trigger-translations
         with:
           release-version: ${{ steps.create_tag.outputs.trimmed-version }}
           glotpress-url: ${{ secrets.GLOTPRESS_IMPORT_URL }}
+
+  build-zip-smoke-tests:
```
jbordonado: I thought we agreed that it was not necessary to run the smoke tests for a pre-release package, given that we're running all our E2E tests on `develop` branch anyway, and the only change we're doing on the fly before building the package is adjust the version number in `package.json`

.github/workflows/build-zip-file.yml
```diff
@@ -2,15 +2,40 @@ name: "Build zip file"
```
jbordonado: Can we rename the file to `build-zip-and-run-smoke-tests.yml`, for instance?

And adjust the name l.1 accordingly

.github/workflows/build-zip-file.yml
```diff
@@ -2,15 +2,40 @@ name: "Build zip file"
 
 # This action will run when it is triggered manually.
```
jbordonado:
```suggestion
# This action will run inside the package creation workflows or when it is triggered manually
```

.github/workflows/e2e-pull-request.yml
```diff
@@ -6,22 +6,37 @@ on:
       - develop
       - trunk
   workflow_dispatch:
+  workflow_call:
+    inputs:
+      wcpay-use-build-artifact:
+        type: boolean
+        required: false
+        default: false
+      wcpay-artifact-directory:
```
jbordonado: I think, for now, we can remove the input and hardcode the default value in the below ENV variable.
We might never need it in the future, I don't see the target directory being ever different anytime soon, so my opinion is to keep it simple for a start

.github/workflows/release-code-freeze.yml
```diff
@@ -127,3 +137,13 @@ jobs:
                     }
                 ]
             }
+
+  build-zip-smoke-tests:
```
jbordonado: We discussed a few weeks ago that it would make more sense to include this in the child workflow, i.e. `release-pr.yml`, since  we will catch not only regular releases but also hotfix releases (for which the release lead would manually trigger `release-pr` instead of `release-code-freeze`)

.github/workflows/e2e-pull-request.yml
```diff
@@ -6,22 +6,37 @@ on:
...
+  WC_SUBSCRIPTIONS_REPO:    ${{ secrets.WC_SUBSCRIPTIONS_REPO }}
+  E2E_BLOG_ID:              ${{ secrets.E2E_BLOG_ID }}
+  E2E_BLOG_TOKEN:           ${{ secrets.E2E_BLOG_TOKEN }}
+  E2E_USER_TOKEN:           ${{ secrets.E2E_USER_TOKEN }}
+  WC_E2E_SCREENSHOTS:       1
+  E2E_SLACK_CHANNEL:        ${{ secrets.E2E_SLACK_CHANNEL }}
+  E2E_SLACK_TOKEN:          ${{ secrets.E2E_SLACK_TOKEN }}
+  E2E_USE_LOCAL_SERVER:     false
+  E2E_RESULT_FILEPATH:      'tests/e2e/results.json'
+  WCPAY_USE_BUILD_ARTIFACT: ${{ inputs.wcpay-use-build-artifact }}
```
jbordonado, line 34: Do you need these env variables to access them within `setup.sh`? Are they not available from the inputs directly?

If not, can we then use the env variables everywhere to be consistent? (l.71 & l.75)

jbordonado (changes requested):
I left a first round of comments, but I didn't test anything yet.
Let me know what you think, I'll start testing things out after the first changes (if you agree with some of my feedback hopefully 😄)

.github/workflows/build-zip-file.yml
jbordonado: Can we use `! inputs.skip-smoke-tests` instead of `inputs.skip-smoke-tests != true`?
I remember I had some problems with booleans and if conditions back then, but I see you've been using both notations in the PR, so maybe it works nicely everywhere, and we can be consistent?

.github/workflows/build-zip-file.yml
achyuthajoy: @jbordonado I understand it's always best to keep inputs to minimum. However, in this case, the caller workflow & build zip are 2 different workflows / jobs and hence to checkout a specific branch, that needs to be passed as an input.

I'm not sure whether it's possible to checkout a specific branch on one job/workflow & use it on another one without passing it as an input. For instance, the release branch is created from `develop` by the [release - code freeze workflow](https://github.com/Automattic/woocommerce-payments/actions/workflows/release-code-freeze.yml) and then calls the build zip workflow. If we don't pass the branch name to build zip, it would run on `develop`. 

Let me know your thoughts.

.github/workflows/build-zip-file.yml
achyuthajoy: I added it as a use case if someone wishes to just create the release branch & PR, but not build a zip file, which is a rare case. Happy to remove if you think it's not required.

And yes, we want smoke tests to be run, so build zip is a per-requisite.

.github/workflows/build-zip-file.yml
achyuthajoy: Ah yes it's not needed. I added the deletion step because somehow a build zip ended up inside the release folder, when I tried the build zip. It's not happening right now.

I'll remove it.

.github/workflows/build-zip-file.yml
jbordonado: Yeah, I don't see when we wouldn't want a zip out of the release branch, to be honest. And if we do, we can just ignore the zip file run for this one occurrence. So I vote to remove this input

.github/workflows/build-zip-file.yml
jbordonado: You're definitely right. I had something else in mind, but it looks like it's not possible actually.

.github/workflows/build-zip-file.yml
jbordonado: Without forgetting to remove the "delete build zip" part of the step's name too

.github/workflows/build-zip-file.yml
achyuthajoy: Removed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/06a924e2509b932eb181f15041d910b8d43d9a81

.github/workflows/build-zip-file.yml
achyuthajoy: Removed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/06a924e2509b932eb181f15041d910b8d43d9a81

.github/workflows/build-zip-file.yml
achyuthajoy: Removed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/b270fa50fc9a47d4041741b46c9374592d5e985b

.github/workflows/create-pre-release.yml
achyuthajoy: Removed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/c1afe1a34626f1512c7d379928d473097e73f645

.github/workflows/build-zip-file.yml
achyuthajoy: Updated in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/032c1da45c2e3e06c7d25fc05839e8fbeae5ed71 & https://github.com/Automattic/woocommerce-payments/pull/5586/commits/3bfe465a93a1f781e45928145acb65c735742135

.github/workflows/e2e-pull-request.yml
achyuthajoy: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/4dd1cdc983cae2b2a612cc644c5e72494da34c37

.github/workflows/e2e-pull-request.yml
achyuthajoy: The setup script can't access the inputs directly hence we use env vars. I've updated the references

.github/workflows/release-code-freeze.yml
achyuthajoy: Moved build zip & smoke tests to `release-pr` workflow - https://github.com/Automattic/woocommerce-payments/pull/5586/commits/bdc6da56a0be26c268e42144e0229bebfef5360f

.github/workflows/build-zip-file.yml
achyuthajoy: Updated in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/d15f4b18c70daa7a3431cb520699709a4e44a40e

achyuthajoy:
@jbordonado I've address all the reviews. Good for another round of review.

.github/workflows/release-code-freeze.yml
```diff
@@ -80,7 +80,7 @@ jobs:
   slack-notification:
     name: "Send notification to Slack"
     needs: [check-code-freeze, create-release-pr]
-    if: ${{ inputs.skipSlackPing != true }}
+    if: ${{ ! ( inputs.skipSlackPing && needs.create-release-pr.outputs.release-pr-created ) }}
```
jbordonado: Is the goal to avoid skipping the Slack ping in case the `build-zip-and-run-smoke-tests` job added in `release-pr.yml` were to fail?

Can we look into the [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) job's prop for that? I assume we can achieve that goal by just specifying that property on `build-zip-and-run-smoke-tests` job, but it would need to be checked.

.github/workflows/release-code-freeze.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
         with:
           # For posting a rich message using Block Kit (https://api.slack.com/messaging/interactivity)
-          payload: |       
+          payload: |
```
jbordonado, line 97: Can we slightly update the text from the Slack ping? I believe we can make it more concise and clear, and we can also mention that we now already built the zip file + run the smoke tests.

Here is what I had in mind: [Slack Block Kit Builder preview](https://app.slack.com/block-kit-builder/T024FN1V2#%7B%22blocks%22:%5B%7B%22type%22:%22header%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Code%20freeze%20notification%22,%22emoji%22:true%7D%7D,%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22The%20automated%20workflow%20just%20did%20the%20following%20actions:%22%7D%7D,%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22%E2%80%A2%20Create%20a%20release%20branch%20%60release/x.y.z%60%20%5Cn%20%E2%80%A2%20Raise%20a%20Pull%20Request%20using%20%60trunk%60%20as%20the%20base%20%5Cn%20%E2%80%A2%20Build%20a%20zip%20file%20and%20run%20the%20E2E%20tests%20against%20it%22%7D%7D,%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22If%20you%20absolutely%20need%20to%20include%20a%20PR%20that%20was%20not%20merged%20in%20%60develop%60%20yet,%20please%20consult%20the%20release%20lead.%22%7D%7D,%7B%22type%22:%22section%22,%22fields%22:%5B%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Version:*%5Cn$%7B%7B%20env.RELEASE_VERSION%20%7D%7D%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Planned%20date:*%5Cn%3Chttps://wcpay.wordpress.com/dev-resources/wc-payments-release-calendar/%7C$%7B%7B%20env.RELEASE_DATE%20%7D%7D%3E%22%7D%5D%7D%5D%7D)

Note: replace `release/x.y.z` by the corresponding env variables in the code, and remove `GITHUB_TAG_URL` env from the job, since it will not used anymore

Basically, keep it short with list items to enumerate what the workflow has done, and remove the link behind the `version` (the GH release doesn't exist yet when the ping is sent, so the link returns a 404).

Nice to have: 

- add a new section with 2 fields `Release zip file` and `Smoke tests run`, below the `Version` and `Planned date` section

OR (if not possible; not clear if we can retrieve a working link for the artifact before the workflow finishes)

- add a PR comment to the `release/x.y.z>trunk` PR with a link to the zip file and a link to the smoke tests run

I think it's best if we ship what you already built, so do you mind creating a new issue for this "nice to have" improvement?

jbordonado (changes requested):
I've been able to test successfully:
- Release code freeze, with or without smoke tests (skip flag)
- Release PR, with or without smoke tests (skip flag)

They used the newly created release branch to build the zip file, as expected.
The E2E tests then used the uploaded artifact, as expected.

I could not test the `build-zip-file-and-run-smoke-tests` workflow via a manual trigger, since it doesn't exist yet, and I didn't want to bother with a hacky way (add `push` trigger first, and trigger through gh cli). But I don't see any issue in the code for the case where it is triggered manually (and where it should use the branch where we triggered it). That should work fine.

I just had one comment on the way to keep the Slack ping even though the E2E tests would fail. I believe we can use a built-in field to achieve this. I also suggested a few changes on the Slack ping content, which shouldn't require much work (since I shared the exact format through the Block Kit link)

.github/workflows/release-code-freeze.yml
achyuthajoy: @jbordonado `continue-on-error` would work in this case, but there is a problem with the approach.

* If the release PR is created and build zip or smoke tests failed, we can make use of `continue-on-error` to send out the slack notification.
* But if the release PR creation failed for any reason (e.gbranch already exists), using `continue-on-error` would still send out the slack notification even if the PR is not created. Hence I added an output to confirm that the PR was ready before sending the slack alert.

WDYT?

.github/workflows/release-code-freeze.yml
achyuthajoy: > add a new section with 2 fields Release zip file and Smoke tests run, below the Version and Planned date section

@jbordonado We can link to the GH action run, which would also help with downloading the artifact & see the status of smoke test. It won't be possible to get the artifact data until the workflow is finished. 

I tried fetching the artifact data within the workflow and I got `{"total_count":0,"artifacts":[]}`

But after the workflow is finished I got:

```
{
  "total_count": 1,
  "artifacts": [
    {
      "id": 592122679,
      "node_id": "MDg6QXJ0aWZhY3Q1OTIxMjI2Nzk=",
      "name": "woocommerce-payments",
      "size_in_bytes": 14341765,
      "url": "https://api.github.com/repos/Automattic/woocommerce-payments/actions/artifacts/592122679",
      "archive_download_url": "https://api.github.com/repos/Automattic/woocommerce-payments/actions/artifacts/592122679/zip",
      "expired": false,
      "created_at": "2023-03-10T04:26:27Z",
      "updated_at": "2023-03-10T04:26:28Z",
      "expires_at": "2023-03-24T04:25:39Z",
      "workflow_run": {
        "id": 4381298159,
        "repository_id": 173384525,
        "head_repository_id": 173384525,
        "head_branch": "devops/automate-smoke-tests",
        "head_sha": "bc3dd1955540d17d339d12d086146fa95194b42e"
      }
    }
  ]
}
```

Here is the GitHub documentation related to artifacts.
> During a workflow run, you can use the [download-artifact](https://github.com/actions/download-artifact) action to download artifacts that were previously uploaded in the same workflow run. After a workflow run has been completed, you can download or delete artifacts on GitHub or using the REST API.

achyuthajoy: So the options we have are:

1. Add link to the workflow run in the slack notification which can be used to download the artifact & check smoke test status.
2. Move slack notification to a new workflow which can be triggered when required and runs after the current workflow is finished. Then we can fetch the artifact data and send it with the slack ping. I don't wish to complicate things more, so this is my least preferred option atm.

achyuthajoy: I will create new issues to improve the slack notification. We can look at them later

.github/workflows/release-code-freeze.yml
jbordonado: > But if the release PR creation failed for any reason (e.gbranch already exists), using continue-on-error would still send out the slack notification even if the PR is not created.

I thought we would add `continue-on-error: true` only on the `build-zip-and-run-smoke-tests` job inside `release-pr.yml`, meaning:
- the Slack job would run even when the `build-zip-and-run-smoke-tests` job inside `release-pr.yml` fails
- the Slack job would not run when the `prepare-release` job inside `release-pr.yml` fails, which is fine because we don't want to say on Slack that the release branch and PR are ready when it failed

But I realize now that this implies that if we trigger `release-pr.yml` manually and the smoke tests job fails, then we would not see this reflected on the workflow's run status, which may not be what we want (I'm not sure if that's important)

I'll let you decide if we keep it that way given these details

.github/workflows/release-code-freeze.yml
jbordonado: Option 1 is fine. But as I said, all of these can happen later in a new set of issues, I just wanted to lay out my ideas when I started to review the PR.

If we can just adjust the wording like on the link I shared (and keep the PR comment, or new text sections for later), that would be already very good

.github/workflows/release-code-freeze.yml
```diff
@@ -77,53 +77,80 @@ jobs:
...
-                            {
-                                "type": "mrkdwn",
-                                "text": "*Planned date:*\n<https://wcpay.wordpress.com/dev-resources/wc-payments-release-calendar/|${{ env.RELEASE_DATE }}>"
-                            }
-                        ]
-                    }
-                ]
-            }
+          script: |
+            github.rest.issues.createComment({
```
jbordonado: Using [gh pr comment](https://cli.github.com/manual/gh_pr_comment) directly could be easier and avoid introducing a 3rd party action:

`gh pr comment ${{ needs.create-release-pr.outputs.release-pr-id }} --body "Hi from GitHub CLI"`

Not sure if you're experimenting on that PR or if you plan also to ship this, I just wanted to mention again that I didn't expect this in this PR, so it's fine if you decide to revert this part.

achyuthajoy: I was just experimenting stuff :) to see if it works.

.github/workflows/release-code-freeze.yml
achyuthajoy: Doesn't look good if we mark a failed job successful. So let's stick to my approach for now and we'll improve later

.github/workflows/release-code-freeze.yml
achyuthajoy: I've added the link to the workflow run on slack I've added the workflow run link to the slack notification.

.github/workflows/release-code-freeze.yml
achyuthajoy: `gh pr` command works fine. I've removed the action.

achyuthajoy:
@jbordonado Good for another review. I've added the following:

* Updated the slack notification content
* Included link to workflow run on slack notification
* Included link to Pull request on slack notification
* Added step to create comment on PR with the workflow link

.github/workflows/release-code-freeze.yml
```diff
@@ -77,15 +77,15 @@ jobs:
       releaseVersion: ${{ needs.check-code-freeze.outputs.nextReleaseVersion }}
     secrets: inherit
 
-  slack-notification:
-    name: "Send notification to Slack"
+  slack-notification-pr-comment:
```
jbordonado: We can revert to the original name `slack-notification`

.github/workflows/release-code-freeze.yml
```diff
@@ -77,15 +77,15 @@ jobs:
...
+    name: "Send notification to Slack & comment on PR"
```
jbordonado: Same, we can revert

.github/workflows/release-pr.yml
```diff
@@ -84,28 +100,52 @@ jobs:
...
           BRANCH_NAME: ${{ steps.create_branch.outputs.branch-name }}
           CHANGELOG: ${{ steps.generate_changelog.outputs.changelog }}
           RELEASE_VERSION: ${{ steps.create_branch.outputs.trimmed-version }}
         run: |
           PR_BODY=$(echo -e ":warning: Please do not merge the PR from the GitHub interface. :warning:\n\n Instead, you can use the following command:\n\`\`\`\n git checkout trunk && git pull \n git merge --no-ff $BRANCH_NAME -m 'Merge $BRANCH_NAME into trunk' \n git push origin trunk \n\`\`\` \n Changelog: \n\`\`\`\n${CHANGELOG}\n\`\`\`")
-          gh pr create --title "Release branch for $RELEASE_VERSION" --body="$PR_BODY" --base="trunk"
+          PR_URL=$(gh pr create --title "Release branch for $RELEASE_VERSION" --body="$PR_BODY" --base="trunk")
+          PR_ID=${PR_URL##*/}
+          if [[ $PR_ID =~ ^[0-9]+$ ]]; then
+            echo "Release PR ID: $PR_ID"
```
jbordonado: This `echo` can be removed? I assume it was there during the testing

jbordonado (approved):
A few things to clean up after you moved the PR comment back to `release-pr`, but pre-approving now not to waste any more time.

I retested the code freeze workflow only, and it seems to work well.
Here is the Slack notification I received, for reference:

Screenshot 2023-03-10 at 11 41 38

Great work, I think this is a great improvement! 🥳

.github/workflows/release-code-freeze.yml
achyuthajoy: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/62038c903bca224966cac8a4f64d2afa6f546a7c

.github/workflows/release-code-freeze.yml
achyuthajoy: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/62038c903bca224966cac8a4f64d2afa6f546a7c

.github/workflows/release-pr.yml
achyuthajoy: Fixed in https://github.com/Automattic/woocommerce-payments/pull/5586/commits/294f03bd554f2a0fdb88753af87032b0fb64cdc0

PR #5591, opened 2023-02-21
Re-activate e2e tests for the shortcode split UPE

timur27:
This PR removes the redundant setup logic in the shortcode split UPE test, which, in turn, fixes the test being flaky.

#### Testing instructions

* All E2E test groups should be green.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-shortcode-upe-split
+++ b/changelog/fix-shortcode-upe-split
@@ -1,4 +1,0 @@
-Significance: patch
-Type: fix
-
-Fixes the split UPE shortcode checkout
--- a/changelog/fix-shortcode-upe-split
+++ b/changelog/fix-shortcode-upe-split
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixes the split UPE shortcode checkout
--- a/changelog/reenable-split-upe-e2e-tests
+++ b/changelog/reenable-split-upe-e2e-tests
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Re-enable split UPE E2E tests and ensure they pass.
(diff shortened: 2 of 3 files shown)


bborman22 (approved):
I haven't had much luck getting these tests to run locally (not because of this PR, just with my local setup) so I can't confirm it locally, but I do see all tests passing in the GH Actions. I also was following the discussion around why taking this approach and the future plans to re-enable these tests. I'll approve based on that, but if someone else on the team wanted to take a look locally if there is any value in that as I didn't want to hold this PR up any longer.

timur27:
Thanks, @bborman22. I analyzed the tests once more and found out that the reason behind these tests being flaky is probably not SEPA. 

If we focus on the first three lines in the test
```js
await setupProductCheckout(
    config.get( 'addresses.customer.billing' )
);
await shopper.goToCheckout();
await uiUnblocked();
```

we'll find out, that `setupProductCheckout` has its own `await shopper.goToCheckout(); await uiUnblocked();` inside. This makes 
```js
await shopper.goToCheckout();
await uiUnblocked();
```
redundant, and I guess this is why the test behaves unstable occasionally. E2E tests run locally vs with Actions seem to have a bit different setup, which is why it was so hard to establish the real reason behind the problem at the first attempt.

Having all this said, I'll now remove 

```js
await shopper.goToCheckout();
await uiUnblocked();
```

entirely, and this PR shouldn't be related with SEPA. Please let me know, if all this makes sense or if there are any other questions -- I'll be happy to answer them.

bborman22 (approved):
I can see the redundancy you mention in those calls and could see how that would cause some flaky-ness. I had a lot more success with running my tests locally and was able to get the `upeSplit` test suite to pass with these changes now same as within GH Actions.

LGTM!

PR #5594, opened 2023-02-21
Fix isUsingSavedPaymentMethod logic on the checkout processing with the split UPE enabled

### LINKED ISSUE #5593: UPE payment gateway payment doesn't work with the split UPE enabled
When the split UPE is enabled on the merchant site, shoppers see the following error during the shortcode checkout: 

`Invalid payment method. Please input a new card number.`

#### Steps to reproduce
1. On the merchant site, enable the split UPE feature in WCPay Dev Tools
2. Add a product and proceed to the shortcode checkout
3. Try to pay with any of the available UPE payment gateways

#### Expected behavior
Users should be successfully redirected to the order summary page or to the third-party to proceed with the payment.

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5593

This PR improves the logic for identifying if a saved payment method is used during the split UPE shortcode checkout. Previously, only some of the payment gateway selectors were verified without checking which payment method has been selected, which led to unexpected behavior.

##### Split UPE
As an admin, enable the split UPE feature. Rollback [this](https://github.com/Automattic/woocommerce-payments/commit/ca9580b7fda77341bb9155270f7c86aeab0952bb) commit on your local environment to be able to test SEPA as well. Use the shortcode checkout. Login as the existing user.

0. Disable WooPay and Link
1. Pay with a new CC
2. Pay with a saved CC 
3. Pay with a new SEPA 
4. Pay with a saved SEPA
5. Pay with any other UPE payment gateway - they are all non-tokenized, which means they cannot be saved. This flow should work as expected as well. 
6. Enable WooPay and perform steps 1-5 again. 
7. There's no need to perform steps 1-5 with WooPay disabled and Link enabled, because Link enabled uses a split UPE gateway for card, which is equal to when WooPay & Link are disabled (step 0).

##### Legacy UPE
1. Although the changes affect only `upe-split.js` script, I'd like to ask you to perform a few critical flows to ensure, that everything works as expected. 

##### Logged out
Log out and: 
1. Pay with CC 
2. Pay with SEPA
3. Pay with one of the rest UPE gateways

-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe-split.js
+++ b/client/checkout/classic/upe-split.js
@@ -700,39 +700,6 @@ jQuery( function ( $ ) {
 	};
 
 	/**
-	 * Checks if the customer is using a saved payment method.
-	 *
-	 * @param {string} paymentMethodType Stripe payment method type ID.
-	 * @return {boolean} Boolean indicating whether or not a saved payment method is being used.
-	 */
-	function isUsingSavedPaymentMethod( paymentMethodType ) {
-		const ccSavedPaymentMethodSelector =
-			'#wc-woocommerce_payments-payment-token-new';
-		const sepaPaymentMethodSelector =
-			'#wc-woocommerce_payments_sepa_debit-payment-token-new';
-		const tokenizedPaymentMethods = [ 'sepa_debit', 'card' ];
-
-		// The payment method used cannot be saved, which means that we are not using saved tokens for sure.
-		if ( ! tokenizedPaymentMethods.includes( paymentMethodType ) ) {
-			return false;
-		}
-
-		if ( 'sepa_debit' === paymentMethodType ) {
-			return (
-				$( sepaPaymentMethodSelector ).length &&
-				! $( sepaPaymentMethodSelector ).is( ':checked' )
-			);
-		}
-
-		if ( 'card' === paymentMethodType ) {
-			return (
-				$( ccSavedPaymentMethodSelector ).length &&
-				! $( ccSavedPaymentMethodSelector ).is( ':checked' )
-			);
-		}
-	}
-
-	/**
 	 * Returns the cached setup intent.
 	 *
 	 * @param {string} paymentMethodType Stripe payment method type ID.
@@ -871,3 +838,36 @@ jQuery( function ( $ ) {
 		}
 	} );
 } );
+
+/**
+ * Checks if the customer is using a saved payment method.
+ *
+ * @param {string} paymentMethodType Stripe payment method type ID.
+ * @return {boolean} Boolean indicating whether or not a saved payment method is being used.
+ */
+export function isUsingSavedPaymentMethod( paymentMethodType ) {
+	const ccSavedPaymentMethodSelector =
+		'#wc-woocommerce_payments-payment-token-new';
+	const sepaPaymentMethodSelector =
+		'#wc-woocommerce_payments_sepa_debit-payment-token-new';
+	const tokenizedPaymentMethods = [ 'sepa_debit', 'card' ];
+
+	// The payment method used cannot be saved, which means that we are not using saved tokens for sure.
+	if ( ! tokenizedPaymentMethods.includes( paymentMethodType ) ) {
+		return false;
+	}
+
+	if ( 'sepa_debit' === paymentMethodType ) {
+		return (
+			null !== document.querySelector( sepaPaymentMethodSelector ) &&
+			! document.querySelector( sepaPaymentMethodSelector ).checked
+		);
+	}
+
+	if ( 'card' === paymentMethodType ) {
+		return (
+			null !== document.querySelector( ccSavedPaymentMethodSelector ) &&
+			! document.querySelector( ccSavedPaymentMethodSelector ).checked
+		);
+	}
+}
--- a/client/checkout/classic/upe-split.js
+++ b/client/checkout/classic/upe-split.js
@@ -702,15 +702,34 @@ jQuery( function ( $ ) {
 	/**
 	 * Checks if the customer is using a saved payment method.
 	 *
+	 * @param {string} paymentMethodType Stripe payment method type ID.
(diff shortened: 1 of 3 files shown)


client/checkout/classic/upe-split.js
```diff
@@ -702,11 +702,14 @@ jQuery( function ( $ ) {
 	/**
 	 * Checks if the customer is using a saved payment method.
 	 *
+	 * @param {string} paymentMethodType Stripe payment method type ID.
 	 * @return {boolean} Boolean indicating whether or not a saved payment method is being used.
 	 */
-	function isUsingSavedPaymentMethod() {
+	function isUsingSavedPaymentMethod( paymentMethodType ) {
 		const paymentMethodSelector =
```
FangedParakeet:
```suggestion
	function isUsingSavedPaymentMethod() {
	    const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
		const paymentMethodSelector =
```
I would probably move this inside the function, since this value isn't really dependent on something variable that can only be gathered from outside the function--but this isn't really that important.

FangedParakeet: Then again, having it in this current order does remove you having to call `getSelectedUPEGatewayPaymentMethod()` twice in the checkout event, so there's a benefit there. Ah, I'll leave it up to you, haha.

client/checkout/classic/upe-split.js
```diff
@@ -702,11 +702,14 @@ jQuery( function ( $ ) {
...
-			'#wc-woocommerce_payments-payment-token-new';
+			'#wc-woocommerce_payments_' +
+			paymentMethodType +
+			'-payment-token-new';
```
FangedParakeet: Just repeating a comment I made on Slack, for the purposes of persistence:

> Just FYI, when Link is enabled AFAIK, the card gateway will appear as a split UPE gateway as well, though its ID will be `woocommerce_payments`, as opposed to `woocommerce_payments_card`. Think you'll need to handle that case inside `isUsingSavedPaymentMethod()` as well.

FangedParakeet:
This isn't really relevant until SEPA is both enabled generally and for saved payments, but might we also have to have some similarly customised logic [here](https://github.com/Automattic/woocommerce-payments/blob/fix/shortcode-upe-split/client/checkout/classic/upe-split.js#L828-L832) (and probably a few other places) as well, when saving a new payment method?

client/checkout/classic/upe-split.js
timur27: Eventually I didn't move `const paymentMethodType = getSelectedUPEGatewayPaymentMethod();` to the method, because of 

> Then again, having it in this current order does remove you having to call getSelectedUPEGatewayPaymentMethod() twice in the checkout event

Thanks for all your thoughts here!

client/checkout/classic/upe-split.js
timur27: If I understand your suggestion correctly, the goal is to ensure that `upe-split.js` handles credit card because there are some cases when CC is used as a split UPE gateway. Enabled Link is one of them, but disabled Link & WooPay is another as well. If we are on the same page, this PR should cover it now. Thanks!

client/checkout/classic/upe-split.js
```diff
@@ -848,3 +838,36 @@ jQuery( function ( $ ) {
...
+			! document.querySelector( sepaPaymentMethodSelector ).checked
+		);
+	}
+
+	if ( 'card' === paymentMethodType ) {
+		return (
+			null !== document.querySelector( ccSavedPaymentMethodSelector ) &&
+			! document.querySelector( ccSavedPaymentMethodSelector ).checked
+		);
+	}
```
FangedParakeet:
```suggestion
	const prefix = '#wc-woocommerce_payments-payment';
	const suffix = '-token-new';
	const savedPaymentSelector = 'card' === paymentMethodType ? prefix + suffix : prefix + '_' + paymentMethodType + suffix;

	return (
		null !== document.querySelector( savedPaymentSelector ) &&
		! document.querySelector( savedPaymentSelector ).checked
	);
```
This appears to be working fine, but I'd suggest that it's not tremendously future-proofed. Right now only the SEPA and card payment methods support saved payment methods through WCPay. However, Bancontact, iDEAL, and Sofort all also support recurring payments through Stripe and it's hopefully only a matter of time until we switch them on.

In a glowing future where all payment gateways are combined into the split UPE gateway, we would have to add new blocks for each of these payment methods, when enabling saved payments for these new methods. This will probably get tiresome to maintain at some point. 

If it were up to me, I'd probably rewrite this in a more dynamic fashion, using this suggestion as an example. That being said, this is working effectively right now and your more specific code is probably more efficient than my suggestion right now as well. Let me know your thoughts: I'm eager to hear them!

FangedParakeet (commented):
I was unable to test the SEPA stuff, because when the SEPA capability is enabled, the plugin reverts to the legacy UPE by default and I couldn't figure out how to neutralise that. But everything else appears to be working as expected for me!

I've left one comment about my thoughts on how you've implemented the `isUsingSavedPaymentMethod()` function, but it's more of a question of taste rather than a correction, so just let me know your thoughts on it and then we can march on with merging in this PR. 👍

client/checkout/classic/upe-split.js
timur27: Pretty fair suggestion it is! I've updated the implementation 57f19b6ec20d21398b775790360a919a146ce0cc

timur27:
@FangedParakeet you can apply [this trick](https://github.com/Automattic/woocommerce-payments/pull/5594#:~:text=Rollback%20this%20commit%20on%20your%20local%20environment%20to%20be%20able%20to%20test%20SEPA%20as%20well.) to be able to test with SEPA.

FangedParakeet (approved):
LGTM. Cheers for responding to my small suggestion! Tests well and appears to be working as expected. 👍

PR #5600, opened 2023-02-22
Fixes jumping focus on split UPE blocks gateway select

### LINKED ISSUE #5557: Split UPE: blocks checkout focus shifts on gateway selection
### Describe the bug

When a new Split UPE gateway is selected in the blocks checkout, the page's focus is moved to the top of the checkout page, rather than within the gateway's inner content.
Please investigate whether the issue is within the checkout block or is due to the Split UPE.

Shortcode checkout is not affected by this bug.

### To Reproduce

- Create a page with the shortcode checkout
- Ensure you have multiple UPE gateways enabled for checkout
- Add some products to the cart
- Select one of the UPE gateways
- Focus is scrolling the page to the top

### Expected behavior

The page should not scroll to the top.

FangedParakeet:
Fixes #5557 

# Fixing the frustrating focus forever

## The problem

When the split UPE is enabled, for some reason whenever a new split UPE gateway is selected in the blocks checkout, the page springs upwards and the focus is shifted to some secret spot near the top of the checkout page. This is very irritating.

## _Why_ is this happening?

The UPE checkout scrapes computed styles from the checkout page in order to customise the appearance of the payment element in a way that is coherent with the rest of the checkout page. This is done by adding hidden elements to the DOM, then using `window.getComputedStyle` on these hidden elements to compute the style applied to these elements. 

`window.getComputedStyle` can only provide a snapshot of the current styles applied to an element, so element pseudo states (e.g. hover, **focus**, etc.) must be manually triggered in order to compute these styles. This means that in order to compute the styles of a hidden element in a pseudo state--for example, the focus state of an element--this pseudo state must _actually be triggered_ prior to reading the styling with `window.getComputedStyle`.

We can see this focus being triggered [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L245-L247) when input focus styles are computed [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/upe-styles/index.js#L322-L326). Unfortunately, this is unavoidable, as it is apparently not possible to compute pseudo state styles without actually triggering the pseudo state. 

But why does this happen so frequently _every single time_ a payment gateway is selected while using the split UPE? We should be [caching these styles](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe-split-fields.js#L400), so that they [only need to be computed once](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/blocks/upe-split-fields.js#L406-L408) instead of being computed numerous times for each separate payment element. However, if you test on `develop`, you will be bemused and likely not amused to discover that, for some reason, this caching does not appear to occur on the blocks checkout and `getUPEConfig( 'wcBlocksUPEAppearance' )` always returns `null`. 

Blistering barnacles! What the devil is this madness?

## The solution

So it turned out that this value for `$_POST[ 'is_blocks_checkout' ]` was always unset in the AJAX request to save the UPE appearance, because apparently [if we send the value as a boolean](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/api/index.js#L532), this is ignored by the request. If you hunt around to find this AJAX request in your network tab, you will find that the `is_blocks_checkout` parameter is always missing. 

Changing this parameter's value from a boolean value to a string, returns the parameter to the request and allows it to be parsed correctly by `rest_sanitize_boolean` on the admin AJAX endpoint. We could also send `0` or `1` as values, but I choose boolean strings just because (I'll willingly bend on this delicate declaration if implored!). 

Now the page will jump once (unavoidably and just as in the regular UPE checkout) and then never again! Huzzah!

#### Testing instructions

1. Enable split gateway UPE feature, remove SEPA capability if necessary, and add additional payment methods.
2. Proceed to blocks checkout. 
3. Toggle between various split UPE payment gateways.
4. Note that page will jump **at most** one time and more likely never.
5. Check that regular UPE blocks checkout and regular non-UPE checkout still cache the appearance as expected. 
6. Thank your lucky stars.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-5557-blocks-checkout-focus-changes-on-gateway-select
+++ b/changelog/fix-5557-blocks-checkout-focus-changes-on-gateway-select
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixes focus change on gateway select with split UPE enabled.
--- a/client/checkout/api/index.js
+++ b/client/checkout/api/index.js
@@ -523,7 +523,7 @@ export default class WCPayAPI {
 	 * Saves the calculated UPE appearance values in a transient.
 	 *
 	 * @param {Object} appearance The UPE appearance object with style values
-	 * @param {boolean} isBlocksCheckout True if save request is for Blocks Checkout. Default false.
+	 * @param {string} isBlocksCheckout 'true' if save request is for Blocks Checkout. Default 'false'.
 	 *
 	 * @return {Promise} The final promise for the request to the server.
 	 */
--- a/client/checkout/api/index.js
+++ b/client/checkout/api/index.js
@@ -527,7 +527,7 @@ export default class WCPayAPI {
 	 *
 	 * @return {Promise} The final promise for the request to the server.
 	 */
-	saveUPEAppearance( appearance, isBlocksCheckout = false ) {
+	saveUPEAppearance( appearance, isBlocksCheckout = 'false' ) {
 		return this.request( getConfig( 'ajaxUrl' ), {
 			is_blocks_checkout: isBlocksCheckout,
 			appearance: JSON.stringify( appearance ),
--- a/client/checkout/blocks/upe-fields.js
+++ b/client/checkout/blocks/upe-fields.js
@@ -385,7 +385,7 @@ const ConsumableWCPayFields = ( { api, ...props } ) => {
 		async function generateUPEAppearance() {
 			// Generate UPE input styles.
 			const upeAppearance = getAppearance( true );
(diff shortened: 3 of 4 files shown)


timur27 (approved):
This one is interesting! 😄 And it works for me. Thanks Samir. 🥇 

https://user-images.githubusercontent.com/22319444/220677294-bc869c30-26e1-4dc0-9c3b-753a59eeb37d.mov

PR #5569, opened 2023-02-16
Implement Things To Do for Progressive Onboarding (WCPay overview)

oaratovskyi:
Fixes #4748 

#### Description

This PR adds new logic of notices in Things To Do on the WCPay overview page for Progressive Onboarding.

#### Changes proposed in this Pull Request

- Added `getVerifyBankAccountTask()` function that returns progressive onboarding notice data, update relevant copies in strings.tsx
- Used the newly created function in `getTasks()` WCPay overview page
- Added `created` and `progressiveOnboarding` data to `wcpayAccount.accountStatus`

#### Testing instructions

**Pre-requisites**

- Checkout to `add/4748-server-changes-for-wcpay-overview-notices` on the server.
- Enable Progressive Onboading feature in WCPay Dev
- Change some server code (see instructions in the server PR)

**PO. No first transaction**

- Complete successful PO onboarding, (just payments, don't add a bank account).
- Go to the WCPay overview page. check that Payments are _enabled_, but Deposits are _disabled_.
- Observe there is no `Things To Do` block about proceeding further.

**PO. more than 14 days after account creation**

- Change the account creation date in the server, so it's more than 14 days.
- Refresh the cache on the client
- You should see this notice

image

**PO. more than 30 days after account creation**

- Change the account creation date in the server, so it's more than 30 days.
- Refresh the cache on the client
- You should see this notice

image

- Roll back your account creation date.
- Refresh the cache

**PO. First payment**

- Make a successful payment
- Refresh the client cache
- Go to the WCPay overview page
- Verify you see the notice:

image

**PO. more than 7 days from the first payment**

- Change the date of the first payment in the server, so it's more than 7 days.
- Refresh the client cache
- Go to the WCPay overview page
- Verify you see the notice:

image

**PO. more than 21 days from the first payment**

- Change the date of the first payment in the server, so it's more than 21 days.
- Refresh the client cache
- Go to the WCPay overview page
- Verify you see the notice:

image

**PO. more than 30 days from the first payment**

- Change the date of the first payment in the server, so it's more than 30 days.
- Refresh the client cache
- Go to the WCPay overview page
- Verify you see the notice:

image

- roll back the first transaction date
- Refresh the client cache

**PO. Balance rising:  $1000 TPV reached**

- Make more payments so TPV is more than $1000
- Refresh the client cache
- Verify you see the notice the same as in **PO. more than 7 days from the first payment**

**PO. Near thresholds: $3000 TPV reached**

- Make more payments so TPV is more than $3000
[...]

PR #5609, opened 2023-02-24
Cooldown – Optimize asset bundling and loading

ismaeldcom:
Right now we load most of the assets by creating a JS React file for SVG images or importing them from the `/assets` folder, which result in base64 data URI. The issue with this approach is that some images get bundled multiple times, because we have multiple entries configured.

In addition, we also load some of those images from PHP and CSS files, so we can't just generate the new files on `dist/` as some of them will be duplicated.

This PR includes the following changes:
- Update webpack config to inline `svg/png` files by default. And generate a URL to files in `assets/` folders, for those imported with the query `?assets`. Using the same pattern, we already use for chunks and load assets from PHP, `[file]?ver=[hash]` as cache busting
- Add a new alias for `assets` to easily import them.
- Replace every JS SVG component with SVG files.
- Adjust SCSS rules replacing `svg` selectors with `img`.
- Remove unused assets, duplicated ones, and those with a Gridicons version.
- Optimize every SVG and PNG files.
- Update jest config to handles assets imported with query, and update every affected snapshot.

#### Testing instructions
As we don't have visual regression tests, and this PR affect many different components, the best way to test it is to visually check that there is no difference between `develop` and this branch. Checking all the images from WCPay (Overview, Settings, Transactions, etc) load correctly, without any markup difference.

>**Note**: Some payment method icons will be slighly different, because I unified them to avoid having multiple versions of the same.

-------------------

**Post merge**


htdat:
👋 @ismaeldcom, this is your cooldown issue, or it's a part of your team Moltres's sprint? 
If none of them is correct, I think you may create an issue for this PR to avoid getting it lost (as it's a nice idea and observation). 

FYI. I am currently doing triage as the Platform Porter PcreKM-yM-p2. I want to know the info above to categorize the issue/PR properly.

ismaeldcom:
👋 Yes! This is a cooldown issue 🙂 Should we use a label or something to make the triage easier? Maybe just adding cooldown to the title 🤔

htdat:
> Should we use a label or something to make the triage easier?

I don't think we need it as we do not have many issues/PRs like this. More infrequently used labels just make everyone more confused :D 

> Maybe just adding cooldown to the title.

Yes, or writing it somewhere in the PR description would be enough.

_Update: I've had another look, it seems that this issue has been in the `Current Focus` pipeline of Moltres' ZenHub board. That's also enough for me. I would have looked at this before._

jbordonado:
How did you optimize the SVG files? Have you used something like https://github.com/svg/svgo?

Maybe we could have a pre-commit hook for this, to avoid any unoptimized assets

ismaeldcom:
>How did you optimize the SVG files? Have you used something like https://github.com/svg/svgo?

I've used [SVGOMG](https://jakearchibald.github.io/svgomg/) to manually optimize each file.

>Maybe we could have a pre-commit hook for this, to avoid any unoptimized assets

I thought about adding webpack ImageMinimizerWebpackPlugin, but for SVG files it's better to do it manually to adjust the precision. Although using it with a safe configuration could be a good option, at least better than no optimization at all 🙂

client/components/payment-delete-illustration/test/index.tsx
```diff
@@ -10,7 +10,7 @@ import { render } from '@testing-library/react';
  * Internal dependencies
  */
 import PaymentDeleteIllustration from '..';
-import WooCardIcon from '../../../gateway-icons/woo-card';
+import WooCardIcon from 'assets/images/payment-methods/woo.svg?asset';
```
oaratovskyi: I wonder if it should be `woo-card.svg`?

oaratovskyi (approved):
Thanks, @ismaeldcom for this huuuge amount of work. I see the difference in our build size and it's dramatic, great work 🔥 

The only thing that I am curious about is the next change when the logo became bigger - is there any reason behind it?
image

Anyway, I've tested the following pages and haven't noticed any differences:
- Connect account
- Overview
- Deposit list
- Deposit details
- Transaction list
- Transaction details
- Dispute list
- Dispute details
- Settings
- Multi-Currency

I've also tested regular checkout as a shopper. I saw that there are changes in WooPay, but I haven't tested it, not sure how.

Pre-approving 🟢 
Let me know if you want me to test more pages and looking forward to your answer on the bigger logo!

htdat:
👋 Hi all, this seems a low-risk PR but I see it's touching many files. I would suggest merging it after the next release of WCPay 5.6 (Mar 15) or after the cut-off commit for WCPay 5.6 freeze RC (Mar 12). Then it can go through all three testing cycles for WCPay 5.7.

client/components/payment-delete-illustration/test/index.tsx
ismaeldcom: Nice catch! Addressed in 114c334

ismaeldcom:
Thanks @oaratovskyi for reviewing this large PR! I think you've tested pretty much every page where images were used. If you're curious about, you can test WooPay by enabling platform checkout in **WCPay Dev**:

image

>The only thing that I am curious about is the next change when the logo became bigger - is there any reason behind it?

That's expected, as part of the optimization, I get rid of `width` and `height` definitions in SVG files, and kept only `viewbox`. That might make some images looks bigger or smaller. But in the end we're applying a fixed size later in JS/CSS.

>I would suggest merging it after the next release of WCPay 5.6 (Mar 15) or after the cut-off commit for WCPay 5.6 freeze RC (Mar 12). Then it can go through all three testing cycles for WCPay 5.7.

💯  agree @htdat! I also prefer to merge cooldown PRs at the beginning of a release to have plenty of testing cycles. Just in case 🙂

oaratovskyi:
> I also prefer to merge cooldown PRs at the beginning of a release to have plenty of testing cycles

100% agree

oaratovskyi:
@ismaeldcom there is also a conflict in `eps.svg`, I removed the comment in my [WCPay Things To Do PR](https://github.com/Automattic/woocommerce-payments/pull/5569/files), I shouldn't have done it :harold

ismaeldcom:
>there is also a conflict in eps.svg, I removed the comment in my [WCPay Things To Do PR](https://github.com/Automattic/woocommerce-payments/pull/5569/files)

Yes, I noticed it while reviewing your PR 🙂  No problem, I'll fix it before merging, in case more conflicts arise.

oaratovskyi:
Checked WooPay setting and checkout page, LGTM!

PR #5638, opened 2023-02-28
Advanced fraud tools section

tpaksu:
Fixes Automattic/woocommerce-payments-server#2797

This PR adds a new screen to WCPay, that holds the advanced fraud protection settings, designed and proposed in WIM5w2Y10SCNBFdMLSVftj-fi-1300%3A194524. Also contains the list item enabling the user to switch to the advanced protection level, and the button that sends the user to the advanced protection level configuration page on the payments settings page. 

### The advanced protection level selection list item:
image

### The page that opens after clicking the "Configure" button:
image

Also this PR contains a new "Amount input" component that's used in the "Purchase price threshold" rule, which allows the user enter only a float in it.

image

#### Testing instructions

* Checkout this branch. 
* (Optional) Nothing is changed on `package.json`, but to be sure, you can run `npm ci` to get the matching packages on the `package-lock.json` file.
* Rebuild the scripts with `npm run build:client`
* Enable the fraud protection feature by running `npm run wp option update wcpay_fraud_protection_settings_active 1`
* Go to WCPay Settings, and you should see the Fraud Protection section on the end of the form, before the Advanced link.
* Check if the section contains the "Advanced" checkbox, with the description matching the Figma design. 
* Check if the "Configure" button is only enabled when the advanced protection level is selected.
* Check if the "Configure" button when clicked sends you to the Advanced protection settings page.
* Check if the design of Advanced protection settings matches with the Figma design (Don't bother about the container width's and text breaking on different parts, because that's designed for different screen sizes). 
* Check if enabling/disabling some settings are saved when you click the "Save Changes" button, and not immediately saved by changing the values.
* Check the save isn't triggered when there are some validation errors on forms containing extra fields, such as "Order velocity", "Purchase price threshold" and "Order item count" rules. The warnings will display a yellow box, and the errors will display a red box as stated in the figma. 
* Try to break things. 

-------------------

**Post merge**


tpaksu:
Building #5685 on top of this PR, because it has the necessary UI to build the rules from.

dwainm:
- [x] Check if the section contains the "Advanced" checkbox, with the description matching the Figma design.
- [x] Check if the "Configure" button is only enabled when the advanced protection level is selected.
- [x] Check if the "Configure" button when clicked sends you to the Advanced protection settings page.
- [x] Check if the design of Advanced protection settings matches with the Figma design (Don't bother about the container width's and text breaking on different parts, because that's designed for different screen sizes).
- [x] Check if enabling/disabling some settings are saved when you click the "Save Changes" button, and not immediately saved by changing the values.
- [x] Check the save isn't triggered when there are some validation errors on forms containing extra fields, such as "Order velocity", "Purchase price threshold" and "Order item count" rules. The warnings will display a yellow box, and the errors will display a red box as stated in the figma.

dwainm (changes requested):
Thanks for the great testing steps @tpaksu !

I found one issue. Order items threshold, both fields: Non-numeric characters aren't causing a warning (it's gone when you refresh, but I think we should not allow user to save): 
Screenshot 2023-03-10 at 15 33 47

tpaksu:
Can it be saved @dwainm? If it’s the case, then it definitely needs a fix. Thanks for catching it!

dwainm:
@tpaksu , it can be saved, so not a significant issue. However, it's not consistent with the rest of the page. The other inputs all only allow numerical inputs.

tpaksu:
@dwainm you can test it again.

dwainm (approved):
Testing all the flows and can confirm this works.

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
```diff
@@ -0,0 +1,197 @@
...
+						value={ minItemsCount }
+						type="number"
+						onChange={ setMinItemsCount }
+						help={ __(
+							'Leave blank for no limit',
+							'woocommerce-payments'
+						) }
+						min={ 1 }
+						step={ 1 }
+						onBlur={ () => {
```
eduardoumpierre: What about creating a function to reuse this logic for both `minItemsCount` and `maxItemsCount`?

This is an alternative:
```js
const handleInputBlur = (itemsCount, setItemsCount) => () => {
  if ( ! ( '' + itemsCount ).match( /^\d+$/ ) ) {
    setItemsCount( '' );
  }
}
```

And then it can be used as:
```js
onBlur={handleInputBlur(minItemsCount, setMinItemsCount)}
onBlur={handleInputBlur(maxItemsCount, setMaxItemsCount)}
```

client/settings/fraud-protection/advanced-settings/cards/international-ip-address.js
```diff
@@ -0,0 +1,54 @@
...
+					'This filter screens for IP addresses outside of your ' +
+						'{{supportedCountriesLink}}supported countries{{/supportedCountriesLink}}.',
+					'woocommerce-payments'
+				),
+				components: {
+					supportedCountriesLink: <Link href="#" />,
+				},
+			} ) }
+		>
+			<div>
```
eduardoumpierre: Do we need this extra `<div>` for all rule cards?

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
```diff
@@ -0,0 +1,197 @@
...
+							if ( ! ( '' + maxItemsCount ).match( /^\d+$/ ) ) {
+								setMaxItemsCount( '' );
+							}
+						} }
+					/>
+				</div>
+			</div>
+			{ ! parseInt( minItemsCount, 10 ) &&
```
eduardoumpierre: This condition could be moved to a variable such as `isItemRangeSet` (just flipping the actual condition).

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
```diff
@@ -0,0 +1,197 @@
...
+						<br />
+						<FraudProtectionRuleCardNotice type={ 'warning' }>
+							{ __(
+								'An item range must be set for this filter to take effect.',
+								'woocommerce-payments'
+							) }
+						</FraudProtectionRuleCardNotice>
+					</div>
+				) }
+			{ parseInt( minItemsCount, 10 ) > parseInt( maxItemsCount, 10 ) ? (
```
eduardoumpierre: This condition could also be moved to a variable.

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
```diff
@@ -0,0 +1,197 @@
...
+			</FraudProtectionRuleDescription>
+		</div>
+	</FraudProtectionRuleCard>
+);
+
+export const OrderItemsThresholdValidation = (
+	settings,
+	setValidationError
+) => {
+	const key = 'order_items_threshold';
```
eduardoumpierre: Instead of storing just the `key`, we could use object destructuring to get the values from `settings[key]`, like:

```js
const key = 'order_items_threshold';
const { enabled, min_items: minItems, max_items: maxItems } = settings[ key ] || {};

if ( enabled ) {
  if ( ! parseInt( minItems, 10 ) && ! parseInt( maxItems, 10 ) ) {
  ...
```

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
```diff
@@ -0,0 +1,188 @@
...
+							'Minimum purchase price',
+							'woocommerce-payments'
+						) }
+					</label>
+					<AmountInput
+						id={ 'fraud-protection-purchase-price-minimum' }
+						prefix={ '$' }
+						placeholder={ '0.00' }
+						value={ minAmount }
+						onChange={ ( val ) => {
```
eduardoumpierre: This line could be replaced by:
```js
onChange={ setMinAmount }
```

As the parameter order is the same, we can just pass the function reference.

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
```diff
@@ -0,0 +1,188 @@
...
+		</div>
+	</FraudProtectionRuleCard>
+);
+
+export const PurchasePriceThresholdValidation = (
+	settings,
+	setValidationError
+) => {
+	const key = 'purchase_price_threshold';
+	if ( settings[ key ].enabled ) {
```
eduardoumpierre: We could also benefit from object destructuring or we can find a way to extract the duplicated logic to a different function.

client/settings/fraud-protection/advanced-settings/index.js
```diff
@@ -0,0 +1,199 @@
...
+			document
+				.querySelector( '.woocommerce-layout__header-heading' )
+				.after( saveButton );
+		}
+	} );
+
+	const validateSettings = ( fraudProtectionSettings ) => {
+		setValidationError( null );
+		let validationResult = true;
+		validationResult &&= OrderItemsThresholdValidation(
```
eduardoumpierre: What do you think about this approach?

```js
const isValid = [
  OrderItemsThresholdValidation,
  OrderVelocityValidation,
  PurchasePriceThresholdValidation,
]
  .map( ( validationFn ) =>
    validationFn( fraudProtectionSettings, setValidationError )
  )
  .every( Boolean );
```

client/settings/fraud-protection/advanced-settings/index.js
```diff
@@ -0,0 +1,199 @@
...
+			setIsSavingSettings( false );
+		} else {
+			window.scrollTo( {
+				top: 0,
+			} );
+		}
+	};
+
+	// Hack to make "WooCommerce > Settings" the active selected menu item.
+	const wcSettingsMenuItem = document.querySelector(
```
eduardoumpierre: Can this be moved to a `useEffect` hook?

client/settings/fraud-protection/advanced-settings/rule-toggle.js
```diff
@@ -0,0 +1,141 @@
...
+
+	// Set global object values from input changes.
+	useEffect( () => {
+		if (
+			advancedFraudProtectionSettings &&
+			advancedFraudProtectionSettings[ setting ]
+		) {
+			advancedFraudProtectionSettings[ setting ].enabled = toggleState;
+			advancedFraudProtectionSettings[ setting ].block = checkState;
+			setAdvancedFraudProtectionSettings(
```
eduardoumpierre, line 55: In order to prevent immutability issues, I'd suggest changing the update approach to something like this:

```js
setAdvancedFraudProtectionSettings( ( settings ) => ( {
  ...settings,
  [ setting ]: {
    ...settings[ setting ],
    block: checkState,
    enabled: toggleState,
  },
} ) );
```

eduardoumpierre (commented):
I still need to follow the testing instructions, but the code is well-tested and easy to follow.

I left some small comments and suggestions.

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
tpaksu: Addressed in [fb841a9](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/fb841a90d36c4cf19988faa34ae8c8bbe2bc2a09). But those changes will need a re-test :)

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
tpaksu: Addressed in [fb841a9](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/fb841a90d36c4cf19988faa34ae8c8bbe2bc2a09) alongside with the other variable assignment fixes.

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
tpaksu: Addressed in [fb841a9](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/fb841a90d36c4cf19988faa34ae8c8bbe2bc2a09).

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.js
tpaksu: Used the object destructuring directly in the parameter in [b80055d](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/b80055d7577fd928eb0e3e9258e3774cdc3614b8).

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
tpaksu: Addressed this too (don't know which one :P)

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
tpaksu: Removed all the key usage and added object destructuring in [b80055d](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/b80055d7577fd928eb0e3e9258e3774cdc3614b8).

client/settings/fraud-protection/advanced-settings/index.js
tpaksu: Needed to tweak a bit to include keys for the object destructuring in the validation method parameters, thanks for suggesting this! Addressed in [b80055d](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/b80055d7577fd928eb0e3e9258e3774cdc3614b8).

client/settings/fraud-protection/advanced-settings/index.js
tpaksu: Thanks for suggesting this, didn't know that needed to be done in an `useEffect` wrapper.

client/settings/fraud-protection/advanced-settings/rule-toggle.js
tpaksu: I need to understand this a bit better to fix it, we own the global object and modifying only it, this may need to be updated in each card toggle, right? Because some cards have extra settings, and I may need to think more about it.

client/settings/fraud-protection/advanced-settings/cards/international-ip-address.js
tpaksu: Good catch! They're redundant. Removed in [3389a35](https://github.com/Automattic/woocommerce-payments/pull/5638/commits/3389a355839c4a87b06a18e802fbb7e8250ddee1).

tpaksu:
Thank you @eduardoumpierre for reading the code and finding those issues!

client/settings/fraud-protection/advanced-settings/index.js
eduardoumpierre: It's just to prevent running it multiple times during the component lifespan. As the component contains a state, it will most likely re-render multiple times, and having this logic in the component body will make it run multiple times as well.

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
```diff
@@ -0,0 +1,178 @@
...
+			<strong>Limits</strong>
+			<div className="fraud-protection-rule-toggle-children-horizontal-form">
+				<div className="fraud-protection-rule-toggle-children-vertical-form">
+					<label htmlFor="fraud-protection-purchase-price-minimum">
+						{ __(
+							'Minimum purchase price',
+							'woocommerce-payments'
+						) }
+					</label>
+					<AmountInput
```
eduardoumpierre, line 67: It seems the default value for the `AmountInput` is `0` rather than empty, which causes an error message to show whenever we fill in the `min_amount` first.

image

Is this the expected behavior or should the initial value be an empty string?

eduardoumpierre: I'm also able to save the settings even with the error message. The `max_amount` field returns as an empty string after refreshing the page.

eduardoumpierre: The input allows me to add multiple `-` which invalidates the input value when submitted. (It shows an empty string after saving and refreshing the page)

<table>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
<tr>
<td>
Before submitting
</td>
<td>
After submitting
</td>
</tr>
</table>

eduardoumpierre (commented):
I followed the testing instructions and it worked as expected. The code looks good and is well-tested.

I'm not approving yet because I want to confirm a behavior from a comment that I've left.

Thanks for your great work on this issue so far!

client/settings/fraud-protection/advanced-settings/cards/purchase-price-threshold.js
tpaksu: > Is this the expected behavior or should the initial value be an empty string?

I believe it's the expected behavior, because the help text says "Leave blank for no limit", and it's not blank. You can also save 0 - 0 as the min max amounts, because it'll pass the checks. Treating 0 as "blank" doesn't seem right to me. But being able to save it with an error message shown is not right.

eduardoumpierre (approved):
I followed the testing instructions and everything worked as expected. The code is easy to follow and well-tested.

Thanks for your work on this issue. LGTM! :shipit:

client/settings/fraud-protection/advanced-settings/rule-toggle.js
tpaksu: It broke everything after I switched to that approach, so I've reverted it back. It'll also be safe to use as is.

PR #5650, opened 2023-02-28
Automatically disable Link when WooPay is also enabled

### LINKED ISSUE #5648: Can't disable WooPay or Link after upgrading to 5.5
### Describe the bug
Merchants that had Link active and migrated to WC Payments 5.5 and got WooPay auto enabled can't use any of those methods because one needs the other to be disabled but as both are enabled it's not possible do disable any of them.

slack link: p1677604444196419-slack-C022WMN88KG

### To Reproduce
- Enable UPE and Split UPE via Dev Tools
- Enable WooPay and Link at the same time. To allow this, remove [these lines](https://github.com/Automattic/woocommerce-payments/blob/trunk/client/settings/express-checkout/index.js#L82-L107) and [these lines](https://github.com/Automattic/woocommerce-payments/blob/trunk/client/settings/express-checkout/index.js#L325-L350)
- Now revert the changes in the code
- Reload the settings page and you should see the following

https://user-images.githubusercontent.com/6342964/221988548-2e7aca8a-f957-4962-90b7-71388cf38b50.mov

### Expected behavior
When both WooPay and Link are enabled at the same time we should disable Link and keep WooPay enabled. Later the merchant can choose between WooPay or Link.

leonardola:
Fixes #5648

This PR disables Link when WooPay is active. To allow merchants to later choose between WooPay or Link.

#### Testing instructions
- While on `trunk` branch
- Enable UPE and Split UPE via Dev Tools
- Enable WooPay and Link at the same time. To allow this, remove [these lines](https://github.com/Automattic/woocommerce-payments/blob/trunk/client/settings/express-checkout/index.js#L82-L107) and [these lines](https://github.com/Automattic/woocommerce-payments/blob/trunk/client/settings/express-checkout/index.js#L325-L350)
- Now revert the changes in the code
- Reload the settings page and you should see the following

https://user-images.githubusercontent.com/6342964/221988548-2e7aca8a-f957-4962-90b7-71388cf38b50.mov

- Checkout to this PR branch and [re]load any page
- Now Link should be inactive and it is now possible do disable WooPay and then enable Link if necessary.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -643,8 +643,9 @@ class WC_Payments {
 
 			$payment_methods = self::$card_gateway->get_payment_method_ids_enabled_at_checkout();
 
-			if ( in_array( 'link', $payment_methods, true ) && self::$platform_checkout_button_handler->is_woopay_enabled() ) {
-				$key = array_search( 'link', $payment_methods, true );
+			$key = array_search( 'link', $payment_methods, true );
+
+			if ( $key !== false && self::$platform_checkout_button_handler->is_woopay_enabled() ) {
 				unset( $payment_methods[ $key ] );
 
 				self::get_gateway()->update_option( 'upe_enabled_payment_method_ids', $payment_methods );
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -640,6 +640,16 @@ class WC_Payments {
 	 */
 	public static function register_gateway( $gateways ) {
 		if ( WC_Payments_Features::is_upe_split_enabled() ) {
+
+			$payment_methods = self::$card_gateway->get_payment_method_ids_enabled_at_checkout();
+
+			if ( in_array( 'link', $payment_methods, true ) && self::$platform_checkout_button_handler->is_woopay_enabled() ) {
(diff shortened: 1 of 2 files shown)


includes/class-wc-payments.php
```diff
@@ -560,14 +560,24 @@ public static function get_plugin_headers() {
 	 */
 	public static function register_gateway( $gateways ) {
 		if ( WC_Payments_Features::is_upe_split_enabled() ) {
+
+			$payment_methods = self::$card_gateway->get_payment_method_ids_enabled_at_checkout();
+
+			if ( in_array( 'link', $payment_methods, true ) && self::$platform_checkout_button_handler->is_woopay_enabled() ) {
+				$key = array_search( 'link', $payment_methods, true );
```
alefesouza:
```suggestion
			$key = array_search( 'link', $payment_methods, true );

			if ( $key !== false && self::$platform_checkout_button_handler->is_woopay_enabled() ) {
```

As `array_search` returns false if a value doesn't exists in the array, wouldn't be better to use it to check if the key exists?

alefesouza (approved):
It works as expected and code looks good to me, just a simple suggestion that does not block anything.

PR #5653, opened 2023-03-01
UPE: intent caching tweaks

### LINKED ISSUE #5310: Payment initially succeeds in Stripe, then fails with "The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made."
**Describe the bug**

Payment initially succeeds in Stripe, then fails with "The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made." 

* Seemingly at random, the `.succeeded` webhooks appear not to be arriving.
* The last response received by the site is the payment intent confirmed
* The `payment_intent.succeeded` is not logged on the site
WC Payments say the` payment_intent.succeeded` webhook succeeds (200 response)
Resending it fails without explanation
* `wcpay_track_new_order` appears to have completed without error but there is no corresponding entry in the payments log
* In the site transaction details, no order ID is shown but the payment is shown as authorised: 
* The transactions list shows this in multiple cases: 
* We believe the re-used `pi_ `is something of a red herring, created because the previous payment by the same customer failed to complete correctly. 

* Note: Would it be possible to focus on why the payment is not completed and the transaction is not retaining the order ID?

**To Reproduce**

We don't have the steps to reproduce at this time. Please see p1670273606982649-slack-C02C00HCJ8J

**Expected behavior**

Orders should be shown as pending once the payment has cleared, not cancelled. 

**Screenshots**

In the site transaction details, no order ID is shown but the payment is shown as authorized: 

The transactions list shows this in multiple cases: 

**Environment (please complete the following information):**

 - WP Version: 6.1.1
 - WC Version: 7.1.0
 - Stripe Plugin Version: 7.0.1
 - Browser [e.g. chrome, safari] and Version
 - Any other plugins installed: 

**Additional context**
5764782-zd-woothemes

Other potential instances: 
5778834-zd-woothemes
5226400-zd-woothemes
5634064-zd-woothemes
5745918-zd-woothemes

kalessil:
Fixes #5310 

- Re-works UPE intent cache cleanup (hooks onto order state change + 2 places where order status change might not happen)
- Implements error logging as order note when payment fails at intent update phase

#### Testing instructions

* TBD

#### To be done

- Impact analysis and testing scenarios for:
    - [ ] UPE checkout
    - [ ] Split UPE checkout
    - [ ] Blocks checkout
    - [ ] Subscriptions

-------------------

**Post merge**


includes/class-wc-payments-upe-checkout.php
```diff
@@ -161,6 +163,31 @@ public function get_payment_fields_js_config() {
...
+	/**
+	 * This method extracts payment intent data and assumes failed state of operating flow, when the intent already
+	 * bound to an order or intent is processed but order state is not updated yet. Hence, we'll do some checks to
+	 * ensure buyers session stores a usable intent.
+	 *
+	 * Note: I know we should not enter a failed state, but real-live environments behaviour is unpredictable.
+	 *
+	 * @return string|null
+	 */
+	private function get_payment_intent_data_from_session() {
```
kalessil, line 175: This part of PR clearly states we are still determining where precisely the root cause.

Could changes in hooks we are bound to ('woocommerce_order_status_changed' now) solve the missing intent cleanup? Possibly, but we need steps to reproduce the original issue to confirm that.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -554,6 +570,11 @@ public function process_payment( $order_id ) {
 				} catch ( Amount_Too_Small_Exception $e ) {
 					// This code would only be reached if the cache has already expired.
 					throw new Exception( WC_Payments_Utils::get_filtered_error_message( $e ) );
+				} catch ( Exception $e ) {
+					// If something goes on with API, forget the cached intent as we have no clues what exactly is wrong.
+					// It might be connectivity issues, intent state not allowing the update or post update period.
+					self::remove_upe_payment_intent_from_session();
```
kalessil: Keeping this for now. If the order changes status when an exception happens here, we can drop this statement.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -196,7 +196,23 @@ public function update_payment_intent_ajax() {
...
+					$order_id,
+					$save_payment_method,
+					$selected_upe_payment_type,
+					$payment_country,
+					$fingerprint
+				);
+			} catch ( Exception $e ) {
+				// If something goes on with API, forget the cached intent as we have no clues what exactly is wrong.
+				// It might be connectivity issues, intent state not allowing the update or post update period.
+				self::remove_upe_payment_intent_from_session();
```
kalessil, line 208: Keeping this for now. If the order changes status when an exception happens here, we can drop this statement.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -161,6 +163,31 @@ public function get_payment_fields_js_config() {
...
+			list( $cart_hash, $intent_id, $client_secret ) = explode( '-', $intent_data, 3 );
+			// TODO: should we inject this via constructor?
+			$order = ( new WC_Payments_DB() )->order_from_intent_id( $intent_id );
+			if ( ! $order && ! $this->gateway->is_intent_being_processed( $intent_id ) ) {
+				return $intent_data;
+			}
+			// Clean the cache, to recover it from failed state of our flows.
+			$this->gateway->remove_upe_setup_intent_from_session();
+		}
+		return null;
```
dmvrtx:
```suggestion
			if ( $order || $this->gateway->is_intent_being_processed( $intent_id ) ) {
				// Clean the cache, to recover it from failed state of our flows.
				$this->gateway->remove_upe_setup_intent_from_session();
				$intent_data = null;
			}
		}
		return $intent_data;
```

Less complexity in the `if` statement and just one exit point. What do you think?

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1174,25 +1187,31 @@ public function get_setup_intent_data_from_session( $payment_method = false ) {
...
+	 *
+	 * @return void
 	 */
 	public static function remove_upe_payment_intent_from_session() {
-		if ( isset( WC()->session ) ) {
-			WC()->session->__unset( self::KEY_UPE_PAYMENT_INTENT );
+		$woocommerce = WC();
+		if ( isset( $woocommerce->session ) && $woocommerce->session->__isset( self::KEY_UPE_PAYMENT_INTENT ) ) {
+			$woocommerce->session->__unset( self::KEY_UPE_PAYMENT_INTENT );
 		}
```
dmvrtx:
```suggestion
		if ( isset( $woocommerce->session ) && isset( $woocommerce->session->self::KEY_UPE_PAYMENT_INTENT ) ) {
			unset( $woocommerce->session->self::KEY_UPE_PAYMENT_INTENT );
		}
```

I'm not sure about the actual semantics with using a `const` here, but the idea of magic methods is that we don't call them directly but allow PHP magic to do the trick 😉

dmvrtx (commented):
Thank you for looking into making this part more reliable. Looking at the changes I start to think that there is a gap in WooCommerce API for payment plugins and it might require us looking upstream to properly solve such issues.

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -378,26 +378,32 @@ public function get_selected_stripe_payment_type_id() {
...
+	 * @return void
 	 */
 	private function add_upe_payment_intent_to_session( string $intent_id = '', string $client_secret = '' ) {
-		$cart_hash = 'undefined';
+		$woocommerce = WC();
+		if ( isset( $woocommerce->session ) ) {
+			$cart_hash = 'undefined';
+			if ( isset( $_COOKIE['woocommerce_cart_hash'] ) ) {
+				$cart_hash = sanitize_text_field( wp_unslash( $_COOKIE['woocommerce_cart_hash'] ) );
+			}
```
htdat: Not directly related to this PR, but I think it would be better to use the value from `WC()->cart`: 


```suggestion
			$cart_hash = isset( $woocommerce->cart ) ? $woocommerce->cart->get_cart_hash() : 'undefined';

```

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1133,12 +1150,8 @@ public function log_payment_error_ajax() {
 
 			$this->order_service->mark_payment_failed( $order, $intent_id, $intent_status, $charge_id, $error_message );
 
-			self::remove_upe_payment_intent_from_session();
-
 			wp_send_json_success();
 		} catch ( Exception $e ) {
-			self::remove_upe_payment_intent_from_session();
-
```
htdat, line 1137: I don't think we should delete `remove_upe_payment_intent_from_session` in these places, as `log_payment_error_ajax` handles errors happening during the confirmation stage of shoppers/buyers in the checkout page. 

IMO, as we will surely run this method in both `try` and `catch`, we should run it early, before try-catch.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -774,10 +795,8 @@ public function process_redirect_payment( $order_id, $intent_id, $save_payment_m
 
 				$this->order_service->attach_intent_info_to_order( $order, $intent_id, $status, $payment_method_id, $customer_id, $charge_id, $currency );
 				$this->attach_exchange_info_to_order( $order, $charge_id );
-				$this->update_order_status_from_intent( $order, $intent_id, $status, $charge_id );
 				$this->set_payment_method_title_for_order( $order, $payment_method_type, $payment_method_details );
-
-				self::remove_upe_payment_intent_from_session();
```
htdat, line 776: I also don't think we should delete `remove_upe_payment_intent_from_session` in this method `process_redirect_payment`, as this method is triggered when shoppers/buyers are redirected to the thank-you page. 

At this stage, we should clear up the payment intent session. Otherwise, it will be used for the shopper's next order.

includes/payment-methods/class-upe-payment-gateway.php
htdat: Better to use `set` with the `null` value like [the core](https://github.com/woocommerce/woocommerce/blob/733bcc88cf5e9186b1ae61eb89e30d4eeac74040/plugins/woocommerce/includes/class-wc-cart-session.php#L231-L239). I also think that makes sense as the default value when [trying to get a not-existing session key](https://github.com/woocommerce/woocommerce/blob/733bcc88cf5e9186b1ae61eb89e30d4eeac74040/plugins/woocommerce/includes/abstracts/abstract-wc-session.php#L101-L102) is `null`.

includes/payment-methods/class-upe-payment-gateway.php
htdat: +1 for this handling.

includes/class-wc-payments-upe-checkout.php
htdat: +1 for this improvement.

htdat (commented):
Thanks for working on this. I am leaving some early comments.

includes/class-wc-payments-upe-checkout.php
kalessil: Like the proposal, addressed in f05961ddc8beffc1ead89ebd05864d981c659b1d.

includes/payment-methods/class-upe-payment-gateway.php
kalessil: Sounds good, addressed in 12aad6142388dd0946526e0ef29c4b7d2599c80a.

includes/payment-methods/class-upe-split-payment-gateway.php
kalessil: Nice! Addressed in 0b19357da700f17bbad9bb204cb5cf5a65d992f2.

includes/payment-methods/class-upe-payment-gateway.php
kalessil: Note `$this->update_order_status_from_intent( $order, $intent_id, $status, $charge_id );` call be fore the removed one.
It will trigger an order status update, and as we listen to the hook for order status change, cached intent will be cleared.

includes/payment-methods/class-upe-payment-gateway.php
kalessil: Inside try body: similar to https://github.com/Automattic/woocommerce-payments/pull/5653/files#r1122708333 - `$this->order_service->mark_payment_failed( $order, $intent_id, $intent_status, $charge_id, $error_message );` will trigger order status change hook and we'll perform intent cache reset.

The one in the catch statement makes no sense - look carefully into exceptions potentially happening in try-statement. Those are validations of posted data or errors getting intents/charges. The latest are read operations, so I don't see cases in there which do require the cache reset.

But I'm open to reviving this if there is a case in try we have a case requiring the cache cleaning.

includes/payment-methods/class-upe-payment-gateway.php
kalessil: This is a must-handling. Reiterated with 82dfe6b68c304fc92f3e755bb67cd64cf07f0d86

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -551,9 +569,16 @@ public function process_payment( $order_id ) {
...
 					}
 					$updated_payment_intent = $request->send( 'wcpay_update_intention_request', $order, $payment_intent_id );
-				} catch ( Amount_Too_Small_Exception $e ) {
-					// This code would only be reached if the cache has already expired.
-					throw new Exception( WC_Payments_Utils::get_filtered_error_message( $e ) );
+				} catch ( Exception $e ) {
+					try {
+						$intent = Get_Intention::create( $payment_intent_id )->send( 'wcpay_get_intention_request' );
+						$charge = $intent->get_charge();
+						$this->order_service->mark_payment_failed( $order, $payment_intent_id, $intent->get_status(), $charge ? $charge->get_id() : '', $e->getMessage() );
```
kalessil, line 574: Note: that adds extra order notes, exposing the original issue and cases when APIs are down.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -803,8 +801,6 @@ public function process_payment( $order_id ) {
 				$order->add_order_note( $note );
 			}
 
-			UPE_Payment_Gateway::remove_upe_payment_intent_from_session();
```
kalessil, line 829: Relocating this into UPE-gateway - intent caching is UPE specific and should not be added here in the first place.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -720,8 +720,6 @@ public function process_payment( $order_id ) {
 				];
 			}
 
-			UPE_Payment_Gateway::remove_upe_payment_intent_from_session();
```
kalessil, line 746: Relocating this into UPE-gateway - intent caching is UPE specific and should not be added here in the first place.

includes/payment-methods/class-upe-payment-gateway.php
kalessil: From support PoV I believe it's a good thing to have, but I would love to hear more thoughts on this change.

kalessil:
A new batch of changes, and it makes sense to describe the current version.

Three major decisions:
- rely on order status change to clear cached intent ID
- improve error handling in UPE gateway to mark order failed and create a note for failing intent updates
- remove most of the clear cached intent calls and keep them only in the place where order status change can not be guaranteed

Tradeoffs:
- we have to keep the intent verification (vs order and intent status), as the generated checkout pages are not refreshing the intent id. 

This means it should be possible to open two pages and complete the order on one page; voila, the second open one uses a processed intent. Changes preventing duplicate orders/charges will likely cover this, but I suggest keeping the introduced verifications for now - I don't have better ideas on how to sync cached intent ID.

kalessil:
I need inputs from more folks:
- @FangedParakeet: are there changes from the PR which need to be adopted from `UPE_Payment_Gateway` to `UPE_Split_Payment_Gateway`?
- @Automattic/helix: could this PR affect subscriptions? What test scenarios would you recommend performing?
- @htdat: based on the duplicate charges prevention project, which test scenarios for checkout blocks would you recommend performing?

includes/class-wc-payments-upe-checkout.php
```diff
@@ -161,6 +163,30 @@ public function get_payment_fields_js_config() {
...
+	 * Note: I know we should not enter a failed state, but real-live environments behaviour is unpredictable.
+	 *
+	 * @return string|null
+	 */
+	private function get_payment_intent_data_from_session() {
+		$intent_data = $this->gateway->get_payment_intent_data_from_session();
+		if ( $intent_data ) {
+			// Extract the intent and ensure it's not bound to an order yet nor being processed.
+			list( $cart_hash, $intent_id, $client_secret ) = explode( '-', $intent_data, 3 );
+			$order = ( new WC_Payments_DB() )->order_from_intent_id( $intent_id );
```
anu-rock, line 180: Can we instead get the instance of `WC_Payments_DB` from DI, like in webhook service?

https://github.com/Automattic/woocommerce-payments/blob/d276e3a4efed3d380377cee9d33356ed541b265e/includes/class-wc-payments-webhook-processing-service.php#L112

anu-rock (commented):
The code changes look good based on my first-pass review. Keeping the caching logic exclusively in UPE gateway is a good idea. And thanks for various reliability improvements. I'll come back after some functional testing.

includes/class-wc-payments-upe-checkout.php
KarlisJ: +1

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -577,7 +604,14 @@ public function process_payment( $order_id ) {
...
 			}
 		} else {
-			return $this->parent_process_payment( $order_id );
+			try {
+				return $this->parent_process_payment( $order_id );
+			} catch ( Exception $e ) {
+				// The invoked method already handles order status changes on success and partially on failure.
+				// But, not all exceptions are triggering order status changes, so we have to clear intent cache.
+				// The intent caching is UPE-specific, so we have to keep it in the UPE-gateways only.
+				self::remove_upe_payment_intent_from_session();
```
FangedParakeet, line 613:
```suggestion
				self::remove_upe_payment_intent_from_session();
				throw $e;
```
Do we not need to rethrow the exception here as well? Otherwise if the parent gateway class throws an exception, for whatever reason, won't we just catch it and then continue on to return sucess instead of failure?

brucealdridge:
I have removed Helix from the review-queue due to Subscriptions being passed off to Quark. I have asked them to review.

PR #5654, opened 2023-03-01
Fixes the save cards checkbox appearance for WCPay gateway

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5655

This PR adds the check that prevents displaying the save checkbox if a user is not logged in. The logic is now the same as in [the UPE checkout](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L307-L309).

<table>
<tr>
<th>Before the fix (logged out)</th>
<th>After the fix (logged out)</th>
<tr>
<td>

image

</td>
<td>

image

</td>
</table>

#### Testing instructions

1. As an admin, Disable UPE
2. Ensure you are logged out
3. Proceed to the checkout page with a simple product and confirm that there is no checkbox under the CC gateway fields
4. Log in as an existing user, proceed to the checkout page and confirm that the checkbox now appears

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-save-card-checkbox-appearance
+++ b/changelog/fix-save-card-checkbox-appearance
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Avoid rendering save cards checkbox for logged out users
--- a/includes/class-wc-payments-checkout.php
+++ b/includes/class-wc-payments-checkout.php
@@ -202,7 +202,9 @@ class WC_Payments_Checkout {
 				<?php
 				if ( $this->gateway->is_saved_cards_enabled() ) {
 					$force_save_payment = ( $display_tokenization && ! apply_filters( 'wc_payments_display_save_payment_method_checkbox', $display_tokenization ) ) || is_add_payment_method_page();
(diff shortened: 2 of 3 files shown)


RadoslavGeorgiev (approved):
Thank you! Changes look good, and work as advertised. :shipit: 

FWIW it would be nice to toggle the visibility of the checkbox when customers decide to register on checkout, because then you could get from no registration to registration + stored payment method in one step. Right now (iregardless of this PR) you'd need to do it in two steps.

shendy-a8c:
I took a look at this PR a bit and wondered why it worked in 5.4 without this change.
So far, I find that the checkbox is there but is hidden by WC's code [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/legacy/js/frontend/tokenization-form.js#L95).
With 5.5 (`trunk`), when I put a breakpoint at that WC code, it's still executing but the checkbox is not hidden. I still can't figure out why and what change in 5.5 that is not in 5.4 that causes that unexpected behavior.

Anyway, this PR fixes the problem, so looks good to me.

Issue #4853, opened 2022-10-06
Progressive Onboarding: Collect the MCC and pre-fill it in Stripe's KYC


dmallory42:
### Description

After [this issue](https://github.com/Automattic/woocommerce-payments/issues/4744) that tackles the needed UI components and coupled with [this WCPay server issue](https://github.com/Automattic/woocommerce-payments-server/issues/3193), we need to introduce an MCC select control in our onboarding wizard to allow merchants to select the appropriate MCC for their business. 

After that, we can pass along this MCC and have it _prefilled_ when redirecting the merchant to Stripe KYC. [According to Stripe](https://stripe.com/docs/connect/express-accounts#create-account), any prefilled information won't be asked from the merchant in Stripe's onboarding flow. Nice!

For the PO MVP, we decided to only present in the MCC control the sub-list of MCCs that Stripe uses for its "Industry" control. For fetching the MCCs list to use in the PO wizard, we will use the WCPay API endpoint introduced in [this PR](https://github.com/Automattic/woocommerce-payments-server/pull/3180).

#### Notes

Brandon raised some thoughts on this on Slack (p1665001733545039/1664352339.162549-slack-C03KTTK2YMA) and recommended that we treat the WCPay experience as agnostic as possible.

### Acceptance criteria

* MCC is selectable on the onboarding form
* The selected MCC is passed to [Stripe's Account API](https://stripe.com/docs/api/accounts/create) and is prefilled in Stripe's KYC flow.

darcie:
We had a request from a merchant to update MCC in 5744324-zen, adding to support the needs of this feature.

dmallory42:
Hi @darcie, it looks like that merchant's need is slightly different - this issue would be for collecting the MCC on initial account onboarding, it looks like what they would have needed is a way to update the MCC *after* the account had been created.

cc'ing @rogermattic - it may be worth considering as part of the PO work (or in future) that it could be good to allow merchants to update details about their Stripe account in the WCPay admin section? It looks like it would be technically possible to update via API.

rogermattic:
Thanks for pinging me @dmallory42 

> it may be worth considering as part of the PO work (or in future) that it could be good to allow merchants to update details about their Stripe account in the WCPay admin section? It looks like it would be technically possible to update via API.

Great idea, I'll definitely like to have it in Settings as one of the details that the merchant can edit! I'll add it to my wireframes.

vladolaru:
[...]

Issue #4744, opened 2022-09-12
Create or extend components needed for Progressive Onboarding


dmallory42:
### Description

This issue will cover either creating or identifying and extending any components which will be needed for Progressive Onboarding. 

Example/psuedocode of the type of components we will need (these could be extensions of existing components or brand new components):

`<OnboardingForm>`
`<OnboardingFormStep>`
`<SocialSecurityInput>`
`<DateOfBirthInput>`

#### Required components so far:
- Searchable custom select control with categories for MCC.
- International phone number field.
- Loading bar.
- Update CustomSelectControl to match other form inputs style.
- Notice component – Use the new BannerNotice component added recently.

### Acceptance criteria

* Either create or identify and then extend any components related to Progressive Onboarding, as defined in the designs paJDYF-5IU-p2.

ismaeldcom:
Until we have the final design we cannot know for sure the required components. But for now, I see we will need a `TextControl` (input) with a formatting feature.

For example, to format date and SSN inputs value:

<img width="250" alt="Screenshot 2022-09-28 at 11 59 43" src="https://user-images.githubusercontent.com/7670276/192750631-505077c0-cdf0-49a3-86e4-2bf60bfdccb6.png">

Issue #5476, opened 2023-01-27
Submit and Redirect to Stripe


dmallory42:
### Description

We will need to adjust our existing logic, currently we pass a progressive flag from the client, but this will need to change since the intelligent router will decide whether to pass the user to PO or not. This issue handles any changes required to the Stripe redirect performed by the onboarding form.

### Acceptance Criteria

* We appropriately redirect users to either the normal Stripe KYC, or the PO KYC, with values prefilled.

oaratovskyi:
Reasonable to take after https://github.com/Automattic/woocommerce-payments/issues/5475

ismaeldcom:
Will be probably handled within #5779

oaratovskyi:
@ismaeldcom I also thought so, but it does not work when `prefill` is passed. We'll need to tweak plugin PHP side or maybe even server. I'll do that in #5779 + it will be easier to review and test one by one

ismaeldcom:
🤔 It should work, at least it was working before. If the prefilled data is invalid (according to Stripe) it will create an empty account, that can be the case.

ismaeldcom:
Oh! I found it while reviewing your PR, we will need to update [the current prefill](https://github.com/Automattic/woocommerce-payments/blob/28e46d6ff48a2135b97eb10d68b6052c48b33305/includes/class-wc-payments-account.php#L1008-L1015) to make it work like `progressive`. Right now we're only reading country and business/company type from the URL. So yes, it will be handled in a different PR 🙁

PR #5670, opened 2023-03-06
Initial progressive onboarding form

### LINKED ISSUE #5473: Minimal Implementation of Progressive Onboarding Form
### Description

This issue will cover the creation of the form for Progressive Onboarding. This will be a minimal form (not concerned with the design, just to add the fields and ensure they are getting passed to the server). It may depend on the components developed in #4744.

Validation of the form will be covered by #5474.

### Acceptance criteria

* The form contains all of the expected fields and is passed to the server.

ismaeldcom:
Fixes #5473 

- Implement `<Stepper/>`, a reusable component to show one step at a time, exposing `nextStep` method and `onComplete` event. It checks for step prop `name` and falls back to `index`, to have a string reference to the step, and use it later for tracking purposes (for example).
- Create steps `Personal`, `Business`, and `Store` details based on the design, with the expected fields and values.
- On Stepper complete event, use `fromDotNotation` on the onboarding data to convert dot notation keys to nested objects, removing undefined/null values. And append it as query to `connectUrl`.
- Use a simple way to render fields which can be extended/refactored to implement validation and tracks, without overengineering it at first.

The following TODO items will be covered as part of other issues:
- #4853 – Populate MCC options and add MCC field test. After we have codes defined on the server.
- #4744 – MCC Select control must provide search functionality and create a notice component.

#### Testing instructions
- Put a breakpoint on the server to debug `$account_data` withing `Onboarding_Controller`, and run the debugger.
- Go to **WCPay Dev Utils** and enable `Force onboarding`, `Enable Progressive Onboarding`, and `Force the plugin to act as disconnected...` if you already have an account connected.
- Navigate to `/wp-admin/admin.php?page=wc-admin&path=/payments/onboarding-prototype`.
- The form should have 3 steps with the fields included in the design.
- Complete every step in the form, checking that business type and company structure and rendered based on the selected country.
- After clicking on `Continue` on the third step, the server debugger should trigger and `$account_data` should contain the values from the filled form fields, similar to:

```
email: "john@doe.com"
business_name: "John Doe LLC"
url: "https://johndoe.com"
country: "US",
company:
    structure: "single_member_llc"
individual:
    first_name: "John"
    last_name: "Doe"
phone: "000000000"
business_type: "company"
annual_revenue: "less_than_250k"
go_live_timeframe: "within_1month"
```

-------------------

**Post merge**


client/onboarding-prototype/types.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Internal dependencies
+ */
+
+export type OnboardingFields = {
+	email?: string;
+	'individual.first_name'?: string;
```
dmallory42, line 7: I'm guessing that it wouldn't be possible to have `individual` as a key representing an object containing `first_name` and `last_name` here? I am assuming it's like this because of the dot notation to nested object logic, but just checking/confirming!

dmallory42 (approved):
This looks really good, thanks for working on it @ismaeldcom! I can see already that this gives us a great framework to add design to the form. 

I left one minor comment but it's more of a question - the other thing I noticed (not sure if this is known) is that the location dropdown for `Where is your business legally registered?` is empty for me, with no options to select?

Pre-approving because I think this more than achieves the goal of creating the initial form which can be built on to add the design and complex UI elements. 🙂

client/onboarding-prototype/types.ts
ismaeldcom: I wrote it flattened to avoid having issues with React not detecting nested state changes, and make it easier to enforce types on functions like `getFieldProps`. I was tempted to write it nested and just flatten with a function to use it in the state, but then we will need to deal with different types for the same structure based on where we use it.

Anyway, we can do it, if you think it would be easier to read 🙂

ismaeldcom:
>I noticed (not sure if this is known) is that the location dropdown for Where is your business legally registered? is empty for me, with no options to select?

It should load the content from the server, it works for me even without cache 🤔 Could you check if it's throwing an error on your side, or what's the response from the server? You can check `get_cached_business_types` for that.

client/onboarding-prototype/types.ts
dmallory42: That makes sense, thanks for explaining! 

I don't think it's too hard to read currently, and it sounds like there's a good reason to keep it as it is to avoid issues with nested state changes and enforcing types, so I'd vote for keeping it as it is. 🙂

dmallory42:
> Could you check if it's throwing an error on your side, or what's the response from the server? You can check get_cached_business_types for that.

It looks like it is just some server weirdness - I'm getting some unusual behaviour on certain requests - possibly related to some work I was doing related to cron jobs in another PR, but it's unrelated to the PR. 

I can see that the logic is making the correct call to the server and that the data would get passed back to the client properly, so I don't think it's a blocker/issue. 🙂

client/onboarding-prototype/steps/business-details.tsx
```diff
@@ -0,0 +1,105 @@
...
+				selectedBusinessType.structures.length > 0 && (
+					<CustomSelectControl
+						{ ...getSelectFieldProps( 'company.structure' ) }
+						value={ selectedCompanyStructure }
+						options={ selectedBusinessType.structures }
+					/>
+				) }
+			<CustomSelectControl
+				{ ...getSelectFieldProps( 'mcc' ) }
+				// @ismaeldcom TODO: The select control must provide search functionality.
```
oaratovskyi: I wonder about these todos - should we link GH issues to those?

client/onboarding-prototype/steps/test/business-details.tsx
```diff
@@ -0,0 +1,131 @@
...
+				<BusinessDetails />
+			</OnboardingContextProvider>
+		);
+		const businessNameField = screen.getByLabelText(
+			strings.fields.business_name
+		);
+		const urlField = screen.getByLabelText( strings.fields.url );
+		const countryField = screen.getByText( strings.placeholders.country );
+
+		const mccField = screen.getByText( strings.placeholders.mcc );
```
oaratovskyi: GH actions say 'mccField' is assigned a value but never used, it probably because of the todo below?
> // @ismaeldcom TODO: Add mcc field test

oaratovskyi:
> It should load the content from the server, it works for me even without cache 🤔 Could you check if it's throwing an error on your side, or what's the response from the server? You can check get_cached_business_types for that.

@ismaeldcom works for me, I see the list of countries!

oaratovskyi:
@ismaeldcom I see that it is not described in the testing instructions, but I am just curious about next: what issue should implement in sending those props to Stripe form? Currently Stripe KYC is not pre-filled as it was in prev implementation

client/onboarding-prototype/steps/business-details.tsx
ismaeldcom: Sounds like a good idea, do we have a syntax for that?

client/onboarding-prototype/steps/test/business-details.tsx
ismaeldcom: Yep, we can remove that line, but I left it in purpose to make it clear that we need to use it.

ismaeldcom:
>what issue should implement in sending those props to Stripe form? Currently Stripe KYC is not pre-filled as it was in prev implementation

I think it will fit #5476 well. What do you think?

oaratovskyi:
> I think it will fit https://github.com/Automattic/woocommerce-payments/issues/5476 well. What do you think?

Perfect, thanks! Didn't notice this one

client/onboarding-prototype/steps/test/business-details.tsx
oaratovskyi: Gotcha! Let it be here, we'll address it in #4853

client/onboarding-prototype/steps/business-details.tsx
oaratovskyi: We don't. But I do like this:

```
// TODO GH-4853 <comment>
```

I realize that is clear to me, but not for everyone, so I don't insist on doing that 😅

oaratovskyi (approved):

client/onboarding-prototype/steps/business-details.tsx
ismaeldcom: Looks good to me! _Addressed in 6c12726_

PR #5679, opened 2023-03-07
Added Cancel Confirmation Modal

### LINKED ISSUE #5603: Create warning modal for order status change to Cancelled
Similar to #2814, we should implement a warning modal when merchants attempt to change the order status on a successful WC Pay order to `Cancelled`. (And perhaps other statuses?)

Like with the Refunded status change, a lack of a warning modal here can sometimes lead to painful and expensive consequences for the merchant and/or us. P2 thread pbUcTB-2e0-p2

jessy-p:
Fixes #5603 

- Added a confirmation modal when the user chooses to change the status of Order to Cancelled.

#### Testing instructions
1. Place an order.
2. On the admin dashboard 'Edit Order' screen, select "Cancelled" from the Order Status dropdown
3. A confirmation modal should be displayed with the option to "Do Nothing" or "Cancel Order", which warns the merchant that Cancel order will not refund the customer.
4. On click of Do Nothing, the order status dropdown is reset back to its previous status, on click of Cancel Order, it is retained as Cancelled.

Additional Testing Instructions: 
- If the order is not successful, the confirmation modal should not be displayed. e.g. an unpaid order is allowed to be cancelled without showing any confirmation modal.
- Test that there are no regressions in the Refund Confirmation Modal introduced via [PR#4987](https://github.com/Automattic/woocommerce-payments/pull/4987) as that code is also refactored.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.7.0#test-cancel-confirmation-modal
- [x] Not applicable, added E2E tests - Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.


client/order/cancel-confirm-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+		isCancelConfirmationModalOpen,
+		setIsCancelConfirmationModalOpen,
+	] = useState( true );
+
+	const closeModal = () => {
+		setIsCancelConfirmationModalOpen( false );
+	};
+
+	const doNotCancel = () => {
+		jQuery( '#order_status' ).val( originalOrderStatus ).change();
```
naman03malhotra: Is it possible to use native JS function here instead of jQuery? 
It possibly can be:
```js
document.getElementById('order_status').value = originalOrderStatus;
```
Though it would be good idea to double check if we are using jQuery for legacy reasons.
cc @ismaeldcom for input

Note for Frontend Guild: Can we start the discussion to get rid of jQuery and promote using native JS, jQuery is a bloated library. Possibly, we can reduce bundle size by getting rid of that.

client/order/cancel-confirm-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+					className="cancel-confirmation-modal"
+					actions={
+						<>
+							<Button isSecondary onClick={ doNotCancel }>
+								{ __( 'Do Nothing', 'woocommerce-payments' ) }
+							</Button>
+							<Button isPrimary onClick={ cancelOrder }>
+								{ __( 'Cancel order', 'woocommerce-payments' ) }
+							</Button>
+						</>
```
naman03malhotra: Maybe a good idea to store this fragment in a variable for better code readability.

client/order/cancel-confirm-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+					}
+				>
+					<p>
+						{ interpolateComponents( {
+							mixedString: __(
+								'Are you trying to issue a refund for this order? If so, please click ' +
+									'{{doNothingBold/}} and see our documentation on {{howtoIssueRefunds/}}. If you want ' +
+									'to mark this order as Cancelled without issuing a refund, click {{cancelOrderBold/}}.',
+								'woocommerce-payments'
```
naman03malhotra: Also, this one in a separate variable for better code readability.

client/order/index.js
```diff
@@ -42,30 +44,53 @@ jQuery( function ( $ ) {
...
+			<RefundConfirmationModal
+				orderStatus={ originalStatus }
+				refundAmount={ refundAmount }
+				formattedRefundAmount={ getConfig( 'formattedRefundAmount' ) }
+				refundedAmount={ getConfig( 'refundedAmount' ) }
+			/>
+		);
+	}
+
+	function renderModal( modalToRender ) {
```
naman03malhotra, line 94: Nice, I like how you separated it into a different function :)

client/order/index.js
```diff
@@ -42,30 +44,53 @@ jQuery( function ( $ ) {
...
-					formattedRefundAmount={ getConfig(
-						'formattedRefundAmount'
-					) }
-					refundedAmount={ getConfig( 'refundedAmount' ) }
-				/>,
-				container
+			renderModal(
+				<CancelConfirmationModal
+					originalOrderStatus={ originalStatus }
+				/>
```
naman03malhotra, line 62: Since this flow has been refactored, adding tests for the cases covered here would be nice :) This would increase the code coverage and confidence for future refactors.

client/order/index.js
```diff
@@ -42,30 +44,53 @@ jQuery( function ( $ ) {
...
 		}
 	} );
+
+	function renderRefundConfirmationModal(
+		originalStatus,
+		canRefund,
+		refundAmount
+	) {
+		if ( ! canRefund ) {
+			alert( __( 'Order cannot be refunded', 'woocommerce-payments' ) );
```
naman03malhotra: Just a suggestion: is it possible to use an internal library dialog UI component instead of an alert? 
IMO, alerts are harsh and outdated.

naman03malhotra (commented):
Overall looks great, I left some suggestions on JS and react-specific things nothing major. I haven't tested it yet, I'll have another round of review later.

client/order/cancel-confirm-modal/index.js
jessy-p: `jQuery( '#order_status' ).val( originalOrderStatus ).change();`

can only be replaced by the following lines, let me know what you think

```
const orderStatusElement = document.getElementById( 'order_status' );
orderStatusElement.value = originalOrderStatus;
orderStatusElement.dispatchEvent( new Event( 'change' ) );
```

client/order/cancel-confirm-modal/index.js
jessy-p: Done

client/order/cancel-confirm-modal/index.js
jessy-p: Done

client/order/index.js
jessy-p: Will check, was going by the fact that existing refund errors use an alert here on order edit
https://github.com/Automattic/woocommerce/blob/trunk/plugins/woocommerce/client/legacy/js/admin/meta-boxes-order.js#L998

client/order/cancel-confirm-modal/index.js
ismaeldcom: >Though it would be good idea to double check if we are using jQuery for legacy reasons.

Not needed anymore, we dropped IE support ~6 months ago, so we can embrace vanilla now 🙂 

>Note for Frontend Guild: Can we start the discussion to get rid of jQuery and promote using native JS, jQuery is a bloated library. Possibly, we can reduce bundle size by getting rid of that.

I'd love to 🤔 Although I'm not sure if rewriting all the working code, just to get rid of jQuery is something we should do soon. But start writing new code in vanilla sounds like a good option.

And no issue with the bundle, we're using the global jQuery included in WP.

>can only be replaced by the following lines, let me know what you think

I'd replace `getElementById ` by `querySelector` it's faster, more consistent, and allow using any selector. But it's just a nitpick 🙂

botwoo:

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5603-cancel-confirmation-modal) :rocket:

- Latest commit: 5f47ba3a5671ab5ccd8a32d835d2b80675e58bc4
- Build time: 2023-03-21 12:09:13 UTC

client/order/index.js
jessy-p: Added this style of error notices. Hope it is noticeable enough, it is used in other WCPay screens on errors.
image

client/order/cancel-confirm-modal/index.js
jessy-p: Done

client/order/index.js
jessy-p: Done

client/order/cancel-confirm-modal/index.js
```diff
@@ -0,0 +1,83 @@
...
+	const confirmationMessage = interpolateComponents( {
+		mixedString: __(
+			'Are you trying to issue a refund for this order? If so, please click ' +
+				'{{doNothingBold/}} and see our documentation on {{howtoIssueRefunds/}}. If you want ' +
+				'to mark this order as Cancelled without issuing a refund, click {{cancelOrderBold/}}.',
+			'woocommerce-payments'
+		),
+		components: {
+			doNothingBold: <b>Do Nothing</b>,
+			cancelOrderBold: <b>Cancel Order</b>,
```
htdat: Use i18n functions for these two strings?

Also, since these strings are used in two places: the message itself and the button labels? May declare `const` for them and use the new const(s) in two places.

client/order/refund-confirm-modal/index.js
```diff
@@ -71,7 +75,9 @@ const RefundConfirmationModal = ( {
 					window.location.reload();
 				} else {
 					resetOrderStatus();
-					window.alert( response.data.error );
+					dispatch( 'core/notices' ).createErrorNotice(
```
htdat, line 78: 👍 This is a nice improvement!

client/order/index.js
```diff
@@ -42,30 +44,57 @@ jQuery( function ( $ ) {
...
+			renderRefundConfirmationModal(
+				originalStatus,
+				canRefund,
+				refundAmount
+			);
+		} else if (
+			'wc-cancelled' === this.value &&
+			'wc-cancelled' !== originalStatus
+		) {
+			if ( ! canRefund || 0 >= refundAmount ) {
```
htdat, line 56: WDYT if we simplify the condition of showing the cancel modal a bit? Basically, move it into a separate `if` condition, and combine all conditions into that `if`. 

```suggestion
		} 
		
		if (
			'wc-cancelled' === this.value &&
			'wc-cancelled' !== originalStatus && 
			canRefund &&
			refundAmount > 0
```

tests/e2e/specs/wcpay/merchant/merchant-orders-status-change.spec.js
```diff
@@ -0,0 +1,235 @@
...
+const refundCancelSelector =
+	'.refund-confirmation-modal .wcpay-confirmation-modal__footer .is-secondary';
+const refundConfirmSelector =
+	'.refund-confirmation-modal .wcpay-confirmation-modal__footer .is-primary';
+const selectedOrderStatusSelector = '.wc-order-status > span';
+
+let orderId;
+
+describe( 'Order > Status Change', () => {
+	describe( 'Change Status of order to Refunded', () => {
```
htdat: Nitpick: 
```suggestion
	describe( 'Change Status of order to Cancelled', () => {
```

tests/e2e/specs/wcpay/merchant/merchant-orders-status-change.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+			// Click on Refund order.
+			await expect( page ).toClick( refundConfirmSelector );
+
+			// Wait for refund to be processed
+			await uiUnblocked();
+			await page.waitForNavigation( { waitUntil: 'networkidle0' } );
+
+			// Get the order price
+			const priceElement = await page.$( '.woocommerce-Price-amount' );
```
htdat: Running `document.querySelectorAll('.woocommerce-Price-amount')` on any wp-admin order page, it will return multiple NodeList/Element(s) so I think we might choose a more specific selector, prefer a unique ID or a combination with it rather than a class name.

```suggestion
			const priceElement = await page.$( '#woocommerce-order-items .total .woocommerce-Price-amount' );
```

htdat (commented):
I manually tested and the PR works well. There are some inline comments. After fixing them, I think it's ready to merge (except that if other people have different opinions).

client/order/index.js
jessy-p: First set of condition = Did user click on cancelled
Second set of  condition = Is this order refundable.
Also, I feel like it reads easier when scanning the file due to similarity with the previous refund modal code, and I believe there are more modals to come.

tests/e2e/specs/wcpay/merchant/merchant-orders-status-change.spec.js
jessy-p: Fixed, copy paste error!

tests/e2e/specs/wcpay/merchant/merchant-orders-status-change.spec.js
jessy-p: Changed

client/order/cancel-confirm-modal/index.js
jessy-p: Done.

client/order/index.js
htdat: Thanks for providing more details. Not a blocker here for me to approve this PR.

htdat (approved):
LGTM! 

✅ Manually re-tested both Cancellation and Refund Modals.

PR #5685, opened 2023-03-07
Add JSON rule builder for the rule engine

tpaksu:
Fixes Automattic/woocommerce-payments-server#3043

This adds the communication layer between the fraud protection settings and the server by creating a ruleset on the UI itself and sending it to the server as is, and also handles the storage of the local copy of it. The local copy will last 1 day, and will updated when the cache expires (uses WP transients). 

Also, this PR contains the part which saves the fraud protection level, and modifies the ruleset on the server according to the setting selected. 

This PR also includes: 
- the inclusion of the Basic protection level as the default, 
- the indication of AVS and CVC checks according to Stripe account settings on the advanced protection settings cards,
- the escalation of protection level to advanced, when the settings are saved from the advanced protection settings page.
- the removal of order velocity rule, and the fixes for the fraud protection levels help modals according to p1679314937335329-slack-C01B8KNUYSW.

#### Testing instructions

**Preparation**
* Checkout this branch
* Run `npm run build:client` to update the client scripts
* Open your client DB and check if there are 
    * `_transient_wcpay_fraud_protection_settings` 
    * `_transient_timeout_wcpay_fraud_protection_settings`
    *  `current_protection_level`
options in `wp_options` table. If they exists, delete them. Also check if you have the feature flag (`wcpay_fraud_protection_settings_active`) enabled (set as `1`) to see the UI on the same table. 

    You can use `SELECT * FROM wp_options where option_name in ('current_protection_level', '_transient_wcpay_fraud_protection_settings', '_transient_timeout_wcpay_fraud_protection_settings');`
to query the necessary things to delete.
* Open your server DB and empty the `wcpay_fraud_rulesets` table.

**Testing**

*Initializing the ruleset config on the server*

* When you open a page that requests the settings, 
    * the protection level option should be set as **basic**, 
    * the transient will contain an empty array, 
    * and the same ruleset (empty array) will be stored on the server for the account. 
    * And if you refresh that page, no duplicate records should be created, the system will use the transient value.

*Updating the settings with presets*

* When you change the fraud protection level on the _Payments > Settings_ page, 
    * the **current_protection_level** should change accordingly.
    * the transient will contain the relevant settings, matching with the settings shown in the relevant modal.
    * the server ruleset table will contain a new row, having a matching ruleset with the transient value.
    * When you select Advanced and click Save Settings, the last ruleset definition should stay intact, and only the advanced level will be set on the related option.
    * When you click Configure, you should see the last selected preset's values on the form.

*Updating the settings with advanced definitions*
* While you have a preset level (something different than the Advanced level) saved as your protection level, click the **Advanced** radio button and click **Configure**. Modify your settings a bit, and click **Save changes**.
* You should see a **Current protection level is set to "advanced".** message just before the **Settings Saved** message.
    * the protection level option should be set as **advanced**, 
    * the transient will contain the matching ruleset, 
    * and the same ruleset will be stored on the server for the account. 
    * And if you click "Save changes" again on that page without changing anything, no duplicate records should be created on the server.

_Testing synchronization errors_
1. The client loses the configuration
* Delete the options and transients from the client DB. 
* Refresh the advanced settings page, and see if the settings are fetched back from the server.
2. The server loses the configuration
* Delete all the rulesets for the current account on the DB
* Set the `_transient_timeout_wcpay_fraud_protection_settings` to `0`, to simulate transient expiration
* Refresh the advanced settings page
* Check the server DB to see if the client ruleset settings are saved back to the server ruleset table.
3. The client and server has different settings
* Select and save some presets from the Payments > Settings page, then select Advanced, save, and go to the advanced settings page. Modify some settings on the form and save it.
* Delete the last saved ruleset for the current account on the DB.
* Set the `_transient_timeout_wcpay_fraud_protection_settings` to `0`, to simulate transient expiration
* Refresh the advanced settings page
* Check if the advanced settings page now contains the last ruleset record on the server DB fraud rulesets table. (The source of truth needs to be the server on this case)

-------------------

**Post merge**


tpaksu:
@marcinbot hey 👋! I'm between two minds here, asking again for your help, would you prefer a switch case statement for the rule JSON builder like now, or would you suggest writing a factory type approach, where each rule has it's own class, and the related part of the JSON is built within them?

includes/fraud-prevention/class-fraud-risk-tools.php
```diff
@@ -88,44 +114,266 @@ public function init_advanced_settings_page() {
...
+			}
+		);
+
+		$rule_configuration = [];
+		foreach ( $enabled_settings as $key => $enabled_setting ) {
+			switch ( $key ) {
+				case self::RULE_AVS_MISMATCH:
+					$rule_configuration[] = [
+						'key'     => $key,
+						'outcome' => $enabled_setting['block'] ? self::FRAUD_OUTCOME_BLOCK : self::FRAUD_OUTCOME_REVIEW,
```
marcinbot: The `$enabled_setting['block'] ? self::FRAUD_OUTCOME_BLOCK : self::FRAUD_OUTCOME_REVIEW,` bit gets repeated across all the rules and could be extracted to a variable before the `switch` statement.

tpaksu:
@marcinbot just realized that we don't serialize those rule configurations to JSON when sending, they are given to the Request body as an array directly, because it'll get double encoded when sent, and the server can't decode it. Instead of `JsonSerializable` I'm putting a `to_array` method to each.

client/settings/fraud-protection/advanced-settings/constants.js
```diff
@@ -0,0 +1,339 @@
...
+							key: Checks.CHECK_ITEM_COUNT,
+							operator: CheckOperators.OPERATOR_GT,
+							value:
+								parseInt( ruleConfiguration.max_items, 10 ) ??
+								null,
+						},
+					],
+				};
+			} else {
+				ruleBase.check = parseInt(
```
eduardoumpierre: Is this `parseInt` duplicated?

client/settings/fraud-protection/advanced-settings/index.js
```diff
@@ -101,8 +107,20 @@ const FraudProtectionAdvancedSettingsPage = () => {
 	};
 
 	const handleSaveSettings = async () => {
-		if ( validateSettings( settings.advanced_fraud_protection_settings ) ) {
+		if ( validateSettings( advancedFraudProtectionSettings ) ) {
 			setIsSavingSettings( true );
+			if ( 'advanced' !== currentProtectionLevel ) {
```
eduardoumpierre: What do you think about moving the `advanced` to a constant?

client/data/settings/selectors.js
```diff
@@ -201,7 +201,6 @@ export const getPlatformCheckoutLocations = ( state ) => {
 	);
 };
 
-// TODO: Finish implementation once server API endpoints are deployed.
 export const getCurrentProtectionLevel = ( state ) => {
 	return getSettings( state ).current_protection_level || 'standard';
```
eduardoumpierre: What do you think about moving the `standard` to a constant?

tpaksu: Moved in [773170d](https://github.com/Automattic/woocommerce-payments/pull/5685/commits/773170da25e7fd0251de70ebc538c9f68c15d8bf).

client/settings/fraud-protection/advanced-settings/index.js
tpaksu: Moved in [773170d](https://github.com/Automattic/woocommerce-payments/pull/5685/commits/773170da25e7fd0251de70ebc538c9f68c15d8bf).

client/settings/fraud-protection/advanced-settings/constants.js
tpaksu: Thanks for catching this, yes it's indeed duplicated. Fixed in [773170d](https://github.com/Automattic/woocommerce-payments/pull/5685/commits/773170da25e7fd0251de70ebc538c9f68c15d8bf).

botwoo:

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/3043-json-generator-for-rule-engine) :rocket:

- Latest commit: a7d567f4120f6098a9c69c238a6b74725f08b39b
- Build time: 2023-03-22 11:59:42 UTC

client/settings/fraud-protection/advanced-settings/cards/address-mismatch.js
```diff
@@ -40,20 +37,6 @@ const AddressMismatchRuleCard = () => (
 				'woocommerce-payments'
 			) }
 		</FraudProtectionRuleDescription>
-		<FraudProtectionRuleCardNotice type="info">
```
tpaksu, line 43: Removing rule notice after a decision from [this discussion](https://www.figma.com/file/WIM5w2Y10SCNBFdMLSVftj?node-id=2523:263284#388311039)

client/settings/fraud-protection/advanced-settings/cards/avs-mismatch.js
```diff
@@ -13,38 +13,48 @@ import FraudProtectionRuleCard from '../rule-card';
...
-				'file with the card issuer.',
-			'woocommerce-payments'
-		) }
-	>
-		<FraudProtectionRuleDescription>
-			{ __(
-				'Buyers who can provide the street number and post code on file with the issuing bank ' +
-					'are more likely to be the actual account holder. AVS matches, however, are not a guarantee.',
+const AVSMismatchRuleCard = () => {
+	const declineOnAVSFailure =
```
tpaksu, line 17: Showing different messages according to the platform setting for AVS checks related to the design on [Figma](https://www.figma.com/file/WIM5w2Y10SCNBFdMLSVftj/%5BWCPay%5D-Fraud-%26-Risk-Tools?node-id=2930-263372&t=V91lulvTyXqJtCTK-0)

client/settings/fraud-protection/advanced-settings/cards/cvc-verification.js
```diff
@@ -13,37 +13,47 @@ import FraudProtectionRuleCard from '../rule-card';
...
-			'woocommerce-payments'
-		) }
-	>
-		<FraudProtectionRuleDescription>
-			{ __(
-				'Because the card security code appears only on the card and not on receipts or statements, the card security code ' +
-					'provides some assurance that the physical card is in the possession of the buyer.',
+const CVCVerificationRuleCard = () => {
+	const declineOnCVCFailure =
+		window.wcpaySettings?.accountStatus?.fraudProtection
```
tpaksu, line 18: Showing different messages according to the CVC check settings on the platform, its design is [here](https://www.figma.com/file/WIM5w2Y10SCNBFdMLSVftj/%5BWCPay%5D-Fraud-%26-Risk-Tools?node-id=2930-263372&t=V91lulvTyXqJtCTK-0).

client/settings/fraud-protection/advanced-settings/cards/order-velocity.js
tpaksu, line 1: Removed order velocity related code according to [this](https://a8c.slack.com/archives/C01B8KNUYSW/p1679314937335329) discussion.

client/settings/fraud-protection/advanced-settings/index.js
```diff
@@ -170,9 +186,6 @@ const FraudProtectionAdvancedSettingsPage = () => {
 						<LoadableBlock isLoading={ isLoading } numLines={ 20 }>
 							<AddressMismatchRuleCard />
 						</LoadableBlock>
-						<LoadableBlock isLoading={ isLoading } numLines={ 20 }>
```
tpaksu, line 173: Removed order velocity related code according to [this](https://a8c.slack.com/archives/C01B8KNUYSW/p1679314937335329) discussion.

client/settings/fraud-protection/components/fp-help-text/index.js
```diff
@@ -3,24 +3,32 @@
...
 				'Allows you to fine-tune the level of filtering according to your business needs.',
 				'woocommerce-payments'
 			);
-		default:
+		case ProtectionLevel.STANDARD:
 			return __(
 				"Provides a standard level of filtering that's suitable for most businesses.",
 				'woocommerce-payments'
 			);
+		case ProtectionLevel.BASIC:
```
tpaksu, line 25: Adding "Basic" protection level with any filters included (CVC and AVS are all included in the platform level). The text it shows as the help text needs a discussion and redesign.

client/settings/fraud-protection/components/fp-modal/index.js
```diff
@@ -18,6 +18,12 @@ export const HighFraudProtectionModal = ( {
 	setHighModalOpen,
 	storeCurrency,
 } ) => {
+	const { declineOnAVSFailure, declineOnCVCFailure } = window.wcpaySettings
```
tpaksu, line 21: Added dynamics to AVS and CVC check texts, because we need to show/hide them according to the Stripe account settings for the merchant. If AVS checks are disabled, we need to hide them from both lists, for example.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2256,11 +2272,30 @@ protected function get_advanced_fraud_protection_settings() {
 			return [];
 		}
 
-		$settings = get_option( 'advanced_fraud_protection_settings', false );
-		if ( ! $settings ) {
-			return Fraud_Risk_Tools::get_default_protection_settings();
+		// Check if we have local cache available before pulling it from the server.
+		$cached_server_settings = get_transient( 'wcpay_fraud_protection_settings' );
```
tpaksu, line 2299: This is the place where all magic happens. If there's no ruleset on the server saved for  the merchant, we save the "Basic" protection level ruleset as the first one automatically. This part needs a deep investigation.

allie500:
Thanks for working on this Taha! I've spent most of my day testing this in detail and I've found some issues in my testing. Please see below. 

_Initializing the ruleset config on the server_

- When you open a page that requests the settings,
  - the protection level option should be set as basic,
  - the transient will contain an empty array,
  - and the same ruleset (empty array) will be stored on the server for the account.
  - And if you refresh that page, no duplicate records should be created, the system will use the transient value.

Everything worked as described for the _Initializing the ruleset config on the server_ tests.

---

_Updating the settings with presets_

- When you change the fraud protection level on the Payments > Settings page,
  - the current_protection_level should change accordingly.
  - the transient will contain the relevant settings, matching with the settings shown in the relevant modal.
  - the server ruleset table will contain a new row, having a matching ruleset with the transient value.
  - When you select Advanced and click Save Settings, the last ruleset definition should stay intact, and only the advanced level will be set on the related option.
  - When you click Configure, you should see the last selected preset's values on the form.

**Tested with Basic preset:**
Everything works except for the last two tests. In my test the current protection level changed to advanced as expected but the ruleset definition did not stay intact. It reverted to the standard rule set. When I clicked configure I saw the standard ruleset values on the form.


**Tested with Standard preset:**
Everything worked as described for the _Updating the settings with presets_ tests. 


**Tested with High preset:**
Everything works except for the last two tests. In my test the current protection level changed to advanced as expected but the ruleset definition did not stay intact. It reverted to the standard rule set. When I clicked configure I saw the standard ruleset values on the form.

---

_Updating the settings with advanced definitions_

- While you have a preset level (something different than the Advanced level) saved as your protection level, click the Advanced radio button and click Configure. Modify your settings a bit, and click Save changes.
- You should see a Current protection level is set to "advanced". message just before the Settings Saved message.
  - the protection level option should be set as advanced,
  - the transient will contain the matching ruleset,
  - and the same ruleset will be stored on the server for the account.
  - And if you click "Save changes" again on that page without changing anything, no duplicate records should be created on the server.

**Testing this starting from the Basic preset:**

Before testing this, I reset my level to Basic and confirmed that the protection level and settings transient were basic, they were. I selected Advanced, clicked configure, modified my settings a bit, and then clicked save settings. I got the expected "Current protection level is set to "advanced"". message just before the Settings Saved message.

The protection level option was correctly set to 'advanced' in the client DB. 

The transient did not contain the matching ruleset. It still had the empty array from the basic preset.

The same ruleset was not stored on the server. It was also still just the empty array from basic preset.

All the changes on the advanced settings configure form screen were reverted to the basic preset.

Further clicks of the save settings button did not save my custom configuration. No new records were created in the server DB table.


**Testing this starting from the Standard preset:**

Set level to standard, saved settings, confirmed settings in client and server DBs prior to re-running the test steps.

Pretty much the same result as with Basic. The settings don't get saved to either the transient option or the Server DB table. 

The IP address, Billing address, and Address mismatch rules in the advanced config form all reverted back to the Standard preset. This is the same behavior as Basic, but in this test I also modified the two threshold rules since they were set with standard. I simply set the max price from $1k to $5k and the items threshold from 10 to 25. Those changes persisted in the form but weren't saved on the backend of the client and not on the server either.


**Testing this starting from the High preset:**

Same behavior as Standard preset test run. Rules reverted to the high preset rules in the IP address, Billing address, and Address mismatch rules on the form, the client DB and server DB both still had the high preset. The two threshold form fields still had my custom changes but those changes weren't saved to the DBs.

---

_Testing synchronization errors_

1. The client loses the configuration

- Delete the options and transients from the client DB.
- Refresh the advanced settings page, and see if the settings are fetched back from the server.

I tested this a couple of different times with different preset levels.

First I set the protection level to standard and confirmed settings in both the client and server DB. I then confirmed that the advanced settings page was correctly populated with the Standard preset rules. 

After deleting the three options from the client DB, I refreshed the advanced settings page. The advanced setting page values were correct. Next I checked the client DB and the transient timeout option was set. The transient settings option existed with the correct data. The `current_protection_level` option did not exist.

Testing with High preset.
Confirmed High preset settings were right in both DBs and on the advanced settings page. Same behavior as standard. 

With this scenario it looks like the only thing missing is correctly setting the `current_protection_level` in the client DB. That option was not set for either of my tests. As a result, when you check the fraud protection level on the WC Pay settings page the Basic radio button is selected, even though the actual rules are set to something else. 

---

2. The server loses the configuration

- Delete all the rulesets for the current account on the DB
- Set the _transient_timeout_wcpay_fraud_protection_settings to 0, to simulate transient expiration
- Refresh the advanced settings page
- Check the server DB to see if the client ruleset settings are saved back to the server ruleset table.

Before testing this, I reset the client to the standard preset and confirmed all settings were correct in the client and server DBs. Next I clicked advanced radio and then the configure button. I confirmed that the form was populated with the standard preset rules.

After deleting the rulesets from the DB and setting the timeout transient to 0, I refreshed the page. The page loaded with the Basic preset rules (nothing turned on). The client DB had an updated transient timeout. The ruleset was an empty array and the protection setting was basic. The server picked up the basic ruleset config: `[]`.

When I navigated back to the WCPay Settings page to set up for test 3 below, I got an illegal string offset warning in the client `debug.log` file. This only happened once though. I couldn't get the warning when I re-tested this flow.

<details>
<summary><strong>Illegal string offset warning log</strong></summary>

```
[21-Mar-2023 16:40:20 UTC] PHP Warning:  Illegal string offset 'ruleset_config' in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php on line 2283
[21-Mar-2023 16:40:20 UTC] PHP Stack trace:
[21-Mar-2023 16:40:20 UTC] PHP   1. {main}() /var/www/html/index.php:0
[21-Mar-2023 16:40:20 UTC] PHP   2. require() /var/www/html/index.php:17
[21-Mar-2023 16:40:20 UTC] PHP   3. wp() /var/www/html/wp-blog-header.php:16
[21-Mar-2023 16:40:20 UTC] PHP   4. WP->main() /var/www/html/wp-includes/functions.php:1332
[21-Mar-2023 16:40:20 UTC] PHP   5. WP->parse_request() /var/www/html/wp-includes/class-wp.php:780
[21-Mar-2023 16:40:20 UTC] PHP   6. do_action_ref_array() /var/www/html/wp-includes/class-wp.php:399
[21-Mar-2023 16:40:20 UTC] PHP   7. WP_Hook->do_action() /var/www/html/wp-includes/plugin.php:565
[21-Mar-2023 16:40:20 UTC] PHP   8. WP_Hook->apply_filters() /var/www/html/wp-includes/class-wp-hook.php:332
[21-Mar-2023 16:40:20 UTC] PHP   9. rest_api_loaded() /var/www/html/wp-includes/class-wp-hook.php:308
[21-Mar-2023 16:40:20 UTC] PHP  10. WP_REST_Server->serve_request() /var/www/html/wp-includes/rest-api.php:410
[21-Mar-2023 16:40:20 UTC] PHP  11. WP_REST_Server->dispatch() /var/www/html/wp-includes/rest-api/class-wp-rest-server.php:442
[21-Mar-2023 16:40:20 UTC] PHP  12. WP_REST_Server->respond_to_request() /var/www/html/wp-includes/rest-api/class-wp-rest-server.php:1018
[21-Mar-2023 16:40:20 UTC] PHP  13. WC_REST_Payments_Settings_Controller->get_settings() /var/www/html/wp-includes/rest-api/class-wp-rest-server.php:1171
[21-Mar-2023 16:40:20 UTC] PHP  14. WCPay\Payment_Methods\UPE_Split_Payment_Gateway->get_option() /var/www/html/wp-content/plugins/woocommerce-payments/includes/admin/class-wc-rest-payments-settings-controller.php:415
[21-Mar-2023 16:40:20 UTC] PHP  15. WCPay\Payment_Methods\UPE_Split_Payment_Gateway->get_advanced_fraud_protection_settings() /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php:1701
```

</details>

---

3. The client and server has different settings

- Select and save some presets from the Payments > Settings page, then select Advanced, save, and go to the advanced settings page. Modify some settings on the form and save it.
- Delete the last saved ruleset for the current account on the DB.
- Set the _transient_timeout_wcpay_fraud_protection_settings to 0, to simulate transient expiration
- Refresh the advanced settings page
- Check if the advanced settings page now contains the last ruleset record on the server DB fraud rulesets table. (The source of truth needs to be the server on this case)

I modified this test flow a bit as previous testing failed for me when trying to save custom advanced rulesets. So instead, I just saved the Standard preset as the current protection level. Verified that everything was correctly saved on both the client and server DBs. Next, I clicked advanced radio button, then the configure button. I verified that the standard preset rules were populated in the form. 

After confirming the setup, I deleted the last saved ruleset from the server DB, and set the transient timeout to 0 on the client DB. After refreshing the page, I found that the client had reverted to the server's stored ruleset. In this case, Basic. But the current protection level option in the client DB was still set to standard.

In some of the previous tests, I found that the test would fail and the settings would revert to Basic as well. So to double test this, I set the current protection level to high and confirmed that this setting was correct on the client and server DBs. 

Then I set the current protection level to standard and confirmed that this setting was correct on the client and server DBs. At this point the current ruleset on the server is standard and the previous ruleset is high. 

I clicked the Advanced radio button and then the configure button. I confirmed that the form was filled with the Standard preset. 

Next I deleted the the last saved ruleset on the server DB (standard ruleset) and set the timeout transient to 0. I refreshed the advanced settings page and it reloaded with the high ruleset. However, the `current_protection_level` option in the client DB was still set to `standard`. 

So in this scenario we also need to make sure that the `current_protection_level` is correctly set.

includes/admin/class-wc-rest-payments-settings-controller.php
```diff
@@ -763,21 +768,58 @@ private function update_platform_checkout_enabled_locations( WP_REST_Request $re
...
+			case 'standard':
+				$ruleset_config = Fraud_Risk_Tools::get_standard_protection_settings();
+				break;
+			case 'high':
+				$ruleset_config = Fraud_Risk_Tools::get_high_protection_settings();
+				break;
+			case 'advanced':
+				$referer                   = $request->get_header( 'referer' );
+				$is_advanced_settings_page = 0 < strpos( $referer, 'fraud-protection' );
+				if ( $is_advanced_settings_page ) {
```
eduardoumpierre: This if statement seems to be inverted according to the comments.

When we change the protection levels to **Basic -> Standard -> Advanced** without refreshing the page, it changes successfully to **Standard**. After saving the **Advanced**, it is reset to **Basic**.

It works as expected If we change it to `if ( ! $is_advanced_settings_page )`.

eduardoumpierre (commented):
I've followed the testing instructions and faced the same behaviors that Allie described. I was able to identify one issue in the code that will hopefully fix some scenarios.

Great work on this issue so far, Taha!

includes/admin/class-wc-rest-payments-settings-controller.php
tpaksu: Thank you for catching this and letting me free from some time of digging deep! Yes, that is the culprit that causes it!

tpaksu: Fixed in [9b2dd0a](https://github.com/Automattic/woocommerce-payments/pull/5685/commits/9b2dd0a842f7f829e973ab3b6d264edd14f1eb91).

tpaksu:
Thanks @allie500 and @eduardoumpierre for your reviews! 

I think the only problem left here is losing the fraud protection level when the client loses the `current_protection_level` option. That's because we don't save the level to the server, but I can try and match the retrieved ruleset with the presets we have, and if it matches with a preset defined, I can set the protection level accordingly. That seems to be the only way I can do that. What do you think?

allie500:
> Thanks @allie500 and @eduardoumpierre for your reviews!
> 
> I think the only problem left here is losing the fraud protection level when the client loses the `current_protection_level` option. That's because we don't save the level to the server, but I can try and match the retrieved ruleset with the presets we have, and if it matches with a preset defined, I can set the protection level accordingly. That seems to be the only way I can do that. What do you think?

That's the only way I can think to do it as well. Also, if the retrieved ruleset from the server doesn't match the `basic`, `standard`, or `high` rulesets you could then just set the `current_protection_level` option to `advanced`.

tpaksu:
Added the notice we [discussed](p1679431659816549/1679395176.042179-slack-C01B8KNUYSW) on Slack in [bba4906](https://github.com/Automattic/woocommerce-payments/pull/5685/commits/bba4906750260ff9ab530782b2d035732a9dbe51). To test this, change `fraud_ruleset` to `fraud_rulesets` in 

https://github.com/Automattic/woocommerce-payments/blob/bba4906750260ff9ab530782b2d035732a9dbe51/includes/wc-payment-api/class-wc-payments-api-client.php#L74 

and clear the client options and transients. You should see a fetch error on both screens, preventing users from editing the configurations. The notices should look like: 

On http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Ffraud-protection
image

On http://localhost:8082/wp-admin/admin.php?page=wc-settings&tab=checkout&section=woocommerce_payments
image

eduardoumpierre (approved):
I've followed the testing instructions and everything worked as expected. I also tested the error notice and it showed correctly. There are some complex parts, but the code most of the code is easy to follow. It also has excellent test coverage.

Thanks for working on this issue. LGTM! 🚀

PR #5687, opened 2023-03-08
Fix fatal error for undefined function

Mayisha:
Added a check to avoid the fatal error caused by the function `woocommerce_store_api_register_update_callback` when it does not exist (undefined).

Reported in p1678186111824589-slack-C022WMN88KG  

#### Testing instructions
- Create checkout page with the checkout block.
- Ensure WooPay is enabled in you WCPay settings.
- Go to the shop, add products to the cart and go to the block checkout page.
- Fill in the details, select the checkbox `Save my information for a faster and secure checkout` and provide a phone number.

- Place order.
- Ensure that your WooPay account has been created.


jbordonado (approved):
Thanks for fixing this, looks good on my end code-wise, I'll let others have a look too

brettshumaker (approved):

PR #5692, opened 2023-03-09
Update Subscriptions Core to 5.5.0

mattallan:
This PR updates our subscriptions-core library version from 5.3.1 to 5.5.0.

There [weren't many changes](https://github.com/Automattic/woocommerce-subscriptions-core/blob/af1d20e53b5c6d791c60944514fa67d0b505f657/changelog.txt#L3-L8) between these versions that related to WC Payments however, I've copied over the relevant changelog entries.

---

Note that the following changelog is not included in WC Payments changelogs.:

```
When HPOS is enabled, changing your address while paying for a renewal order will update the address on the subscription.
```

This fix isn't relevant to WC Pay Subscriptions because they don't go through the regular "pay for failed renewal order" flow, instead, they go through the change payment method flow.


brucealdridge (approved):
Reviewed the changes and compared changelog against subs-core 5.3.1

PR #5722, opened 2023-03-10
Update Subscriptions Core to latest tagged 5.5.0 version

mattallan:
See previous version bump #5692 

Subscriptions Core 5.5.0 was re-tagged to include a last-minute change before the version was shipped in WC Subscriptions and WC Payments.

This small PR updates the `composer.lock` to fetch the latest version

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1004,12 +1004,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/Automattic/woocommerce-subscriptions-core.git",
-                "reference": "af1d20e53b5c6d791c60944514fa67d0b505f657"
+                "reference": "5f114e1fad79196a24f189236c9e7fb04482b8a2"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/Automattic/woocommerce-subscriptions-core/zipball/af1d20e53b5c6d791c60944514fa67d0b505f657",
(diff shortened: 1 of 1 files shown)


Jinksi (approved):
Looks good @mattallan 🙌 

* On `develop`, after removing the `vendor` dir, `composer install` results in the previous subs-core 5.5.0 library version in `vendor/woocommerce/subscriptions-core` with only 2 changelog entries.
* On this branch, after removing the `vendor` dir, `composer install` results in the current subs-core 5.5.0 library version in `vendor/woocommerce/subscriptions-core` with all 3 changelog entries.

```
= 5.5.0 - 2023-03-10 =
* Fix - When HPOS is enabled, changing your address while paying for a renewal order will update the address on the subscription. #413
* Fix - Prevent admin error notices being shown for the "subscription trial end" event that was caused by no callbacks being attached to this scheduled action. #414
* Fix - Prevent fatal error when copying the `_shipping_address` meta data where the value is not an array. #417
```

shendy-a8c:
Changelog check fails. I think you will still need to run npm run changelog and it's possible that it won't add anything to the `changelog.txt` by choosing 'patch'.

mattallan:
I can merge without needing to add a changelog since it was already added as part of this PR: https://github.com/Automattic/woocommerce-payments/pull/5692

Thanks @Jinksi and @shendy-a8c

PR #5739, opened 2023-03-10
Added e2e tests for Free Cart & Cart modified on Checkout page

### LINKED ISSUE #5738: E2E test for Free Cart & Cart modified on Checkout page
### Description
Added E2E test for #5277.

### Acceptance criteria
- E2E tests to test that when a free cart is updated via AJAX on the checkout page to be non-zero, the shopper is able to checkout using WCPay.

### Testing instructions
- E2E tests pass
- If #5423 is fully/partially regressed, tests fail

jessy-p:
Fixes #5738
Tests for #5423 

E2E tests to test that when a free cart is updated via AJAX on the checkout page to be non-zero, the shopper is able to checkout using WCPay.

#### Testing instructions
- E2E tests pass
- If #5423 is fully/partially regressed, tests fail

-------------------

**Post merge**


tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+	afterAll( async () => {
+		// Clear the cart at the end so it's ready for another test
+		await shopperWCP.emptyCart();
+	} );
+
+	it( 'checkout a free cart', async () => {
+		await shopper.goToCheckout();
+		await shopper.fillBillingDetails( billingDetails );
+		await page.waitFor( 1000 );
```
deepakpathania: Would waiting for a selector be better here? This could lead to flakiness.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( productName );
+		await shopper.goToCart();
+		await clearAndFillInput( '#coupon_code', 'free' );
+		await expect( page ).toClick( 'button[name="apply_coupon"]' );
+	} );
+
+	afterAll( async () => {
+		// Clear the cart at the end so it's ready for another test
+		await shopperWCP.emptyCart();
```
deepakpathania: We're adding to card before each test and emptying only after all are done - is that intended?

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+		await shopper.fillBillingDetails( billingDetails );
+		await page.waitFor( 1000 );
+		await uiUnblocked();
+		await shopper.placeOrder();
+		await expect( page ).toMatch( 'Order received' );
+	} );
+
+	it( 'update a free cart via AJAX and checkout with WCPay', async () => {
+		await shopper.goToCheckout();
+		await page.click( '.woocommerce-remove-coupon' );
```
deepakpathania: It is usually recommended to store selectors as constants so that it is obvious what they represent.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+		await shopper.placeOrder();
+		await expect( page ).toMatch( 'Order received' );
+	} );
+
+	it( 'update a free cart via AJAX and checkout with WCPay Blocks', async () => {
+		await shopperWCP.openCheckoutWCB();
+		await shopperWCP.fillBillingDetailsWCB( billingDetails );
+		await fillCardDetailsWCB( page, card );
+		await page.waitForSelector(
+			'.wc-block-components-main button:not(:disabled)'
```
deepakpathania: Same as above.

deepakpathania (changes requested):
Left some comments.

> E2E tests to test that when a free cart is updated via AJAX on the checkout page to be non-zero, the shopper is able to checkout using WCPay.

It isn't obvious to me from the tests that we're checking this behavior btw. 🤔

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+	it( 'checkout a free cart', async () => {
+		await shopper.goToCheckout();
+		await shopper.fillBillingDetails( billingDetails );
+		await page.waitFor( 1000 );
+		await uiUnblocked();
+		await shopper.placeOrder();
+		await expect( page ).toMatch( 'Order received' );
+	} );
+
+	it( 'update a free cart via AJAX and checkout with WCPay', async () => {
```
htdat: I think we may want to clarify the action? Just saying `update a free cart via AJAX` does not transfer the meaning of the action, IMO.

```suggestion
	it( 'remove the free coupon, can still checkout with classic checkout', async () => {
```

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,83 @@
...
+		await shopper.goToCheckout();
+		await page.click( '.woocommerce-remove-coupon' );
+		await uiUnblocked();
+		await setupCheckout( billingDetails );
+		await fillCardDetails( page, card );
+		await shopper.placeOrder();
+		await expect( page ).toMatch( 'Order received' );
+	} );
+
+	it( 'update a free cart via AJAX and checkout with WCPay Blocks', async () => {
```
htdat: "WCPay Blocks" does not sound right, may use the "Blocks checkout"? 

```suggestion
	it( 'remove the free coupon, can still checkout with Blocks checkout', async () => {
```



Also, I think for testing blocks checkout, we should move tests to this folder `tests/e2e/specs/blocks/shopper`? @deepakpathania may consult on this.

deepakpathania: Yeah, makes sense. We could then move the `goToCheckout` step to `beforeEach` as well for the rest of the tests.

jessy-p: That would mix up this particular scenario (Checkout of free cart and removing a free coupon and checkout) with the general flow tests. Do you think that might be confusing?

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Each test ends in Place Order, so it is not causing any issue if the tests are successful. Similar to the existing `tests/e2e/specs/wcpay/shopper/shopper-checkout-purchase.spec.js`. Since we are only checking "Order Received", having multiple items in the cart in case one of the test fails does not affect the next tests in both cases.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Not sure, followed the same as the existing utility function `setupCheckout` in `utils/payments`. I cannot reuse the utility function as such, as I don't want the last line of the utility function.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Copied from tests in `specs/blocks/shopper/`. I guess it makes sure that the button is enabled. Changed here

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Done

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Improved all test descriptions.

jessy-p:
> It isn't obvious to me from the tests that we're checking this behaviour btw.

We are adding tests for the changes of #5423  and hopefully preventing recurrence of issues like [Payment form absent when item with price added to cart via AJAX on a free order](https://github.com/Automattic/woocommerce-payments/issues/4764) and [Display CC fields when a 100% coupon code that is added in cart is removed on checkout page](https://github.com/Automattic/woocommerce-payments/issues/5277)
To keep the tests simple, I have used the latter. I think the descriptions earlier reflected the general purpose nature of the change and was confusing, I have updated it to be simple and hopefully, they now better describe the tests.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Could we create a seperate issue to fix this for both [`setupCheckout`](https://github.com/Automattic/woocommerce-payments/blob/5e61ad2629836ecd49583a2668862f6e2e6b9967/tests/e2e/utils/payments.js#L243) and this code?

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Done

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
jessy-p: Improved all test descriptions.

htdat (commented):
This is what I am seeing, it seems not well-organized. 
Markup on 2023-03-15 at 14:37:27

I think totally, we should have 8 tests that are generated from this matrix: 

- Checkout page types: Classic vs Blocks.
- Flows: Non-UPE vs UPE. 
- Tests: Place an order with the free coupon, vs. Place an order with the free coupon then remove it.

Eventually, it would be something like the following. And with that, I think it still makes sense to separate Block vs Classic checkouts as mentioned here https://github.com/Automattic/woocommerce-payments/pull/5739#discussion_r1135520985:

- Use a free coupon in Classic Checkout - non-UPE flow
    - Place order normally. 
    - Remove free coupon, place order normally.
- Use a free coupon in Blocks Checkout - non-UPE flow
    - Place order normally. 
    - Remove free coupon, place order normally.
- Use a free coupon in Classic Checkout - UPE flow
    - Place order normally. 
    - Remove free coupon, place order normally.
- Use a free coupon in Blocks Checkout - UPE flow
    - Place order normally. 
    - Remove free coupon, place order normally.

jessy-p:
Use a free coupon in Classic Checkout - non-UPE flow
- Place order normally. - This is the first test L56
- Remove free coupon, place order normally. - This is the second test L65

Use a free coupon in Blocks Checkout - non-UPE flow
- Place order normally. - This is not tested
- Remove free coupon, place order normally. - This is L75

Use a free coupon in Classic Checkout - UPE flow
- Place order normally. - Would be same as L56, since the payment form is the one affected by UPE/non-UPE
- Remove free coupon, place order normally. - this is L109

Use a free coupon in Blocks Checkout - UPE flow - not tested
- Place order normally.
- Remove free coupon, place order normally.

@htdat 
I don't know if we should try to be too extensive in E2E testing, rather we should try to have fewer representative tests.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
deepakpathania: I noticed you're already using it below so something similar could be done here.
```
await page.waitForSelector( blocksPageEnabledSelector );
```
But not a blocker, so I'll let you take a call.

tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js
```diff
@@ -0,0 +1,118 @@
...
+		await page.waitForSelector( blocksPageEnabledSelector );
+		await expect( page ).toClick( 'button', { text: 'Place Order' } );
+		await page.waitForSelector( 'div.woocommerce-order' );
+		await expect( page ).toMatch( 'p', {
+			text: 'Thank you. Your order has been received.',
+		} );
+	} );
+} );
+
+describe( 'Checkout with free coupon & after updating cart on Checkout page: UPE', () => {
```
deepakpathania: We could potentially have an outer `describe` block with this description with nested `UPE` and `classic` describe blocks. Not a blocker though.

jessy-p: Done

deepakpathania: We could just have the inner describe blocks as `classic` and `UPE` I guess instead of repeating since the overall test description would concatenate anyways.

jessy-p: Done

deepakpathania (approved):
Pre-approving since my open comments are non-blockers. ✅

jessy-p:
@htdat  @deepakpathania  Have you tried running these tests locally?  I am consistently unable to run it locally since yesterday (Tried multiple setups). Before that, while developing, they were running fine on local.

deepakpathania:
@jessy-p Have you tried running without the `-dev` flag in the e2e test command? Or running a test suite in isolation?

deepakpathania:
I tried running in isolation locally and am seeing the following error: 

```
Error: Cannot accept dialog which is already handled!
```
with the following message
```
  message: 'Caught error after test environment was torn down\n' +
    '\n' +
    'Protocol error (Runtime.callFunctionOn): Target closed.'
```
    
When running in dev mode, I can see that the created `checkout-wcb` page isn't available for some reason. 

Screenshot 2023-03-16 at 12 23 45 PM

htdat:
> Have you tried running these tests locally? I am consistently unable to run it locally since yesterday (Tried multiple setups). Before that, while developing, they were running fine on local.

I could finally run this command `npm run test:e2e-dev tests/e2e/specs/wcpay/shopper/shopper-checkout-free-cart.spec.js` and got this result: 

Markup on 2023-03-16 at 14:46:58

But before that, I needed to fix several issues on my end: 

- Activate `woo-gutenberg-products-block` plugin.
- Check E2E server, and resolve the issue missing `WOOPAY_URL` constant.

htdat (approved):
My last review https://github.com/Automattic/woocommerce-payments/pull/5739#pullrequestreview-1340835281 was explained by Jessy that we do not need to add all E2E tests.

jessy-p:
After installing and activating WooCommerce Blocks plugin, it is working on local, thanks @htdat 
This is caused by SKIP_WC_BLOCKS_TESTS=1 as mentioned in readme.
I'm removing the blocks test for now, as it looks like block tests are not meant to be part of the shopper folder.

botwoo:

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5738-E2E-test-free-cart-js) :rocket:

- Latest commit: b5ec377453a6f9282b8c349ea4ff1fc5624b8d65
- Build time: 2023-03-20 06:51:19 UTC

PR #5740, opened 2023-03-10
Use concurrency to stop obsolete workflows

### LINKED ISSUE #5361: Stop obsolete workflows from running
### Description
When a workflow is triggered, if the previous run was not done yet, it will continue running. 
We can use [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) to stop wasting resources and time in such cases.

This is especially true for E2E tests. When a PR is opened, if you push several commits within a few minutes, you'll trigger each time a new run and those take ~10 minutes to complete.
Also, if too many runs are happening at the same time, there is a lack of resources and the next runs are queued and will need to wait for other runs to complete before they can be executed.

### Acceptance criteria

* Where deemed appropriate (we may want to keep old workflow runs running depending on the type of action it is doing, to be defined), old workflow runs are canceled and only one run happens at a time

fashuvo:
Fixes #5361

In our current setup, triggering a new workflow while a previous one is also running, will result in both workflows running at the same time. If multiple commits are pushed within a few minutes, this may require the PR author to wait for all the tests to finish.

This PR resolves the issue by adding [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) support to our repo. We chose to cancel the previous workflow run when a new commit is pushed to the repo. As a result, the workflow will always run for the latest commit only.

#### Testing instructions

* Once merged, this PR should run the workflows for the latest commits only.

-------------------

**Post merge**


.github/workflows/bundle-size.yml
```diff
@@ -4,6 +4,10 @@ on:
     pull_request:
     workflow_dispatch:
```
jbordonado, line 5: Out of the scope of this PR but this trigger doesn't seem supported by `preactjs/compressed-size-action`. Should be removed

jbordonado (approved):

.github/workflows/bundle-size.yml
fashuvo: Noted. Will try to address this separately.

deepakpathania: @fashuvo Let's try to create follow-up issues in order to not forget about these in the future.

PR #5745, opened 2023-03-10
Add tests for UPE scripts

### LINKED ISSUE #5674: Improve unit test coverage for UPE scripts
### Description

Currently, unit test coverage is lacking for UPE checkout scripts in [client/checkout/blocks](https://github.com/Automattic/woocommerce-payments/tree/develop/client/checkout/blocks) and [client/checkout/classic](https://github.com/Automattic/woocommerce-payments/tree/develop/client/checkout/classic). 

### Acceptance criteria
Several new tests have been added, covering the logic and methods in the UPE checkout scripts.

mdmoore:
Fixes #5674 

This adds unit tests for a few of the functions within the UPE classic checkout. The UPE scripts are wrapped in a `
$( document ).ready()` block with functions defined within, sharing global variables throughout the block. Without a larger refactor, testing much of the script(s) isn't possible. However, a few functions could be pulled out of the block and exported, making testing more feasible.

#### Testing instructions

1. Checkout `upe/increase-unit-test-coverage` and run the following tests
    - `npm run test:js -- client/checkout/classic/test/upe.test.js`
    - `npm run test:js -- client/checkout/classic/test/upe-split.test.js`

-------------------

**Post merge**


client/checkout/classic/test/upe-split.test.js
```diff
@@ -72,3 +77,101 @@ describe( 'isUsingSavedPaymentMethod', () => {
 		expect( isUsingSavedPaymentMethod( paymentMethodType ) ).toBe( false );
 	} );
 } );
+
+describe( 'getSelectedUPEGatewayPaymentMethod', () => {
```
FangedParakeet: Correct me if I'm mistaken, but I think it's convention to have one single `describe` block per test file...However, you can nest `describe` blocks inside other `describe` blocks, so I think all you'd need to do to conform would be to wrap all of the blocks in one patriarchal wrapper block. 

I'd probably recommend something like the below.

```js
describe( 'UPE split checkout', () => {
	let container;
	
	beforeEach( () => {
		// container setup...
	}

	afterEach( () => {
		// container tear-down...
	}
	
	describe( 'isUsingSavedPaymentMethod', () => {
		// isUsingSavedPaymentMethod tests...
	}
	
	describe( 'getSelectedUPEGatewayPaymentMethod', () => {
		beforeEach( () => {
			// ...
		}

		afterEach( () => {
			// ...
		}

		// getSelectedUPEGatewayPaymentMethod tests...
	}
	
	// other describe blocks...
}
```

This way you can keep the same custom setup and tear-down functions for each function you're testing. Let me know if that sounds reasonable or like a terrible idea!

client/checkout/classic/test/upe.test.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Internal dependencies
+ */
+//import * as CheckoutUtils from 'utils/checkout';
```
FangedParakeet:
```suggestion
```
Think you forgot to delete this comment. 🔍

FangedParakeet (changes requested):
Tests looking and working good for me! Left a pair of picky suggestions essentially about aesthetics, but I think your setup and test logic doesn't need much adaptation. Nice one.

client/checkout/classic/test/upe-split.test.js
mdmoore: Ah, thanks for mentioning this. I was actually looking for a convention for `describe` blocks when working on these but found WCPay uses both [nested](https://github.com/Automattic/woocommerce-payments/blob/develop/client/disputes/test/index.tsx) and [multiple top level](https://github.com/Automattic/woocommerce-payments/blob/develop/client/card-readers/settings/sections/test/index.test.js) patterns and [sometimes both](https://github.com/Automattic/woocommerce-payments/blob/develop/client/data/disputes/test/resolvers.js). I'm all for making things more descriptive though and wrapping the blocks seems like a good way to do so. [ec059f2](https://github.com/Automattic/woocommerce-payments/pull/5745/commits/ec059f28ca706fb5b723187202321fa3c8f670e3) takes care of it!

client/checkout/classic/test/upe.test.js
mdmoore: Yup. Thanks for catching it!

client/checkout/classic/test/upe-split.test.js
FangedParakeet: >I was actually looking for a convention for describe blocks when working on these but found WCPay uses both [nested](https://github.com/Automattic/woocommerce-payments/blob/develop/client/disputes/test/index.tsx?rgh-link-date=2023-03-13T17%3A49%3A03Z) and [multiple top level](https://github.com/Automattic/woocommerce-payments/blob/develop/client/card-readers/settings/sections/test/index.test.js?rgh-link-date=2023-03-13T17%3A49%3A03Z) patterns and [sometimes both](https://github.com/Automattic/woocommerce-payments/blob/develop/client/data/disputes/test/resolvers.js?rgh-link-date=2023-03-13T17%3A49%3A03Z).

Whoops, my bad! I think in the PHP unit tests there is always one parent group of tests per file, so I incorrectly assumed it was the same here. Thanks for doing it as such anyways!

client/checkout/classic/test/upe-split.test.js
```diff
@@ -1,74 +1,191 @@
...
+				<li class="wc_payment_method payment_method_woocommerce_payments">
+					<input id="payment_method_woocommerce_payments" type="radio" class="input-radio">
+				</li>
+				<li class="wc_payment_method payment_method_woocommerce_payments_bancontact">
+					<input id="payment_method_woocommerce_payments_bancontact" type="radio" class="input-radio">
+				</li>
+			</ul>
+			`;
+			document.body.appendChild( container );
+		} );
```
FangedParakeet, line 109: It's a small suggestion, but couldn't you move the `beforeAll` and `afterAll` from both child `describe` blocks (`isUsingSavedPaymentMethod` and `getSelectedUPEGatewayPaymentMethod`) and change them to `beforeEach` and `afterEach` respectively? That way, we would create one container for each `describe` test block, remove it upon completion of those particular tests, and then recreate it for the next block.

```js
describe( 'UPE split checkout', () => {
	let container;
	
	beforeEach( () => {
		container = document.createElement( 'div' );
		container.innerHTML = `
			<ul class="wc_payment_methods payment_methods methods">
				<li class="wc_payment_method payment_method_woocommerce_payments">
					<input id="payment_method_woocommerce_payments" type="radio" class="input-radio">
				</li>
				<li class="wc_payment_method payment_method_woocommerce_payments_bancontact">
					<input id="payment_method_woocommerce_payments_bancontact" type="radio" class="input-radio">
				</li>
			</ul>
		`;
		document.body.appendChild( container );
	} );

	afterEach( () => {
		document.body.removeChild( container );
		container = null;
	} );
	
	describe( 'isUsingSavedPaymentMethod', () => {
		// isUsingSavedPaymentMethod tests...
	}
	
	describe( 'getSelectedUPEGatewayPaymentMethod', () => {
		// getSelectedUPEGatewayPaymentMethod tests...
	}
	
	describe( 'getSetupIntentFromSession', () => {
		// getSelectedUPEGatewayPaymentMethod tests...
	}
}
```

You don't need the container for the final group of tests (`getSetupIntentFromSession`), but I suppose it doesn't hurt. Otherwise if you want to avoid creating this container for the third block of tests, I would extract the `beforeAll` and `afterAll` functionality into a separate function, so that you could reuse it in the first two blocks of tests.

```js
let container;
const setUpContainer = () => {
	container = document.createElement( 'div' );
	container.innerHTML = `
		<ul class="wc_payment_methods payment_methods methods">
			<li class="wc_payment_method payment_method_woocommerce_payments">
				<input id="payment_method_woocommerce_payments" type="radio" class="input-radio">
			</li>
			<li class="wc_payment_method payment_method_woocommerce_payments_bancontact">
				<input id="payment_method_woocommerce_payments_bancontact" type="radio" class="input-radio">
			</li>
		</ul>
	`;
	document.body.appendChild( container );
}
const tearDownContainer = () => {
	document.body.removeChild( container );
	container = null;
}
describe( 'UPE split checkout', () => {

	describe( 'isUsingSavedPaymentMethod', () => {
		beforeAll( setUpContainer );
		afterAll( tearDownContainer );
		// isUsingSavedPaymentMethod tests...
	}
	
	describe( 'getSelectedUPEGatewayPaymentMethod', () => {
		beforeAll( setUpContainer );
		afterAll( tearDownContainer );
		// getSelectedUPEGatewayPaymentMethod tests...
	}
	
	describe( 'getSetupIntentFromSession', () => {
		// getSelectedUPEGatewayPaymentMethod tests...
	}
}
```

This basically allows us to avoid having copied code in the same file. Let me know your thoughts and if either of these approaches takes your fancy. Cheers!

mdmoore: @FangedParakeet I think this would make sense if the both blocks of mocked HTML were the same. One block is mocking the new payment method radio button with the "Use a new payment method" label,  while the other block is mocking the radio button for individual payment method selection. Is the suggestion here to combine the 2 HTML blocks? If so, that could work.

FangedParakeet: Ah, I see. This is another classic case of me not properly reading something right in front of me. I thought those HTML container elements were basically the same, but in fact they're drastically different. In that case, continue on as normal, nothing to see here. My bad! 😅

FangedParakeet (approved):
LGTM. Think this is a great start to grow some test coverage for our UPE client-side scripts. Nice one! 🚢

PR #5761, opened 2023-03-13
Add new workflow to run E2E tests on a test Atomic site

jbordonado:
Fixes https://github.com/Automattic/woocommerce-payments-meta/issues/77

- Add a new GitHub workflow to run E2E tests against an Atomic site
- Introduce a way to distinguish between different sets of JSON data being used by the tests setup (based on the `NODE_ENV` parameter)
- Update E2E documentation accordingly

As per p9o2xV-2S7-p2, the strategy will be to create proper Atomic sites we would like to test against, as opposed to using ephemeral sites or dev sites.

I decided to create one, to begin with: https://wcpaytestecomm.wpcomstaging.com/. This is a WPCOM eCommerce site running on PHP7.4 with WordPress latest and WooCommerce latest.

#### Limitations

- When an Atomic site is first created, there are things to update/change manually. The issue to perform this in an automated way: https://github.com/Automattic/woocommerce-payments-meta/issues/82
- Similarly, should we need an "env setup" step before running them (apart from the global settings mentioned above), it would also be manual. An example would be that before running E2E tests, we should always upload the zip file built from the branch itself or taken from a build artifact (code freeze workflow). Issue: https://github.com/Automattic/woocommerce-payments-meta/issues/83.
- Related to the point above, we'd need to upload the new WCPay version after each release, until it is implemented.

#### Testing instructions

- GitHub CLI: `gh workflow run e2e-tests-atomic.yml --ref=test/add-atomic-e2e-tests-workflow` to run the new workflow manually. Latest run: https://github.com/Automattic/woocommerce-payments/actions/runs/4467275891
- Locally: `E2E_GROUP=wcpay E2E_BRANCH=shopper npm run test:e2e-dev -- --NODE_ENV=atomic` to run all E2E tests from the `wcpay/shopper` folder on the Atomic site

Note: the rest of the commands for running locally the tests on the Docker-based environment don't change (you may or may not specify --NODE_ENV=test, which is the default anyway)

-------------------

**Post merge**


tests/e2e/README.md
```diff
@@ -173,14 +173,22 @@ Handy utility scripts for managing environment:
 
 <br>
 
+For running E2E tests on an Atomic site:
+
+`E2E_GROUP=atomic npm run test:e2e-dev -- --NODE_ENV=atomic`
```
deepakpathania: Do we need both a separate group and env for this to be distinguished?

tests/e2e/config/env.setup.js
```diff
@@ -0,0 +1,8 @@
+const config = require( 'config' );
+
+global.process.env = {
+	...global.process.env,
+	// Remove the trailing slash from jest sequencer WORDPRESS_URL.
+	WP_BASE_URL: config.get( 'url' ),
+	PUPPETEER_SLOWMO: true,
```
deepakpathania: I assume this was added for debugging and wouldn't be needed for final result?

tests/e2e/config/jest.config.js
```diff
@@ -45,15 +44,17 @@ if ( process.env.E2E_GROUP ) {
 		allowedPaths.push( e2ePaths[ process.env.E2E_GROUP ] );
 	}
 } else {
+	// The 'atomic' folder is a temporary measure until we are able to run all E2E tests for Atomic sites.
```
deepakpathania: Did we remove the folder mentioned here in subsequent commits?

tests/e2e/config/jest.config.js
```diff
@@ -45,15 +44,17 @@ if ( process.env.E2E_GROUP ) {
...
+	// The only way to run E2E tests locally for the Atomic site is to use E2E_GROUP=atomic, to avoid duplicate tests in other cases.
```
deepakpathania: Does this mean the default `npm run test:e2e` run certain tests twice?

deepakpathania (commented):
Left some comments, will continue with functionality testing.

deepakpathania:
Tested locally and the test suite passes in the non `dev` environment, which is what the GH actions use. Having some issues in the `dev` mode with the Chromium window but that might be a local issue.

Screenshot 2023-03-16 at 12 09 22 PM

tests/e2e/README.md
jbordonado: - `E2E_GROUP` is used to figure out which top folder to use within `tests/e2e` to fetch the .spec files we need to run
- `NODE_ENV` is here to target the right JSON file to use for the test data

The `atomic` folder was deleted in the last commit 4cdafe7272b76956965e6f12dff161f3abafa978 because I figured the remaining failing tests from `wcpay | shopper` were failing because UPE was enabled by default on the Atomic site, so I'll update that comment.

tests/e2e/config/jest.config.js
jbordonado: Yep, will remove this comment

tests/e2e/config/jest.config.js
jbordonado: It used to ignore the `atomic` paths from the `allowedPaths` for the regular tests.
Not relevant anymore as the `atomic` folder is gone, the comment will be removed

jbordonado:
I made a bunch of updates, here is the latest run: https://github.com/Automattic/woocommerce-payments/actions/runs/4439226720

We're now running wcpay `shopper` and `merchant` tests, they are passing locally, so I expect that they should also pass on the CI.
I also added `upe` and `upeSplit` tests , although I'm not sure how that's supposed to work since both test suites are executed in parallel I believe. These features are exclusive of each other (activating UPE split means you can't test UPE regular). But it passes on [E2E Tests - Pull Request](https://github.com/Automattic/woocommerce-payments/actions/runs/4439183465) workflow, so maybe I'm understanding it wrong.

Making `subscriptions` tests work requires more time as I was not able to import the sample XML data.
Also, I cannot use the import CLI command on the Atomic site (`wp import` is restricted for Atomic sites), so I'm not sure there'll be a way to avoid doing it manually through the Products screen for future sites..

jbordonado:
Now that I thought about it with a clear head, it makes sense.

The regular E2E tests are running in an isolated Docker container for each combination of group/branch, while for Atomic tests, we are running the tests on the same live site, which makes it impossible to run all the group/branch combinations in parallel.

I've updated the workflow to run each combination sequentially, and they all passed: https://github.com/Automattic/woocommerce-payments/actions/runs/4445925765/jobs/7806631092. The whole run took ~21min (- subscription ones)

I'm not sure there is a way around this because we can't realistically set up as many identical Atomic sites as we have group/branch combinations (4 at the moment).

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5761` or branch name `test/add-atomic-e2e-tests-workflow` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=test/add-atomic-e2e-tests-workflow) :rocket:

---

Build info: 

- Latest commit: dd7d5b4a0773e6c2c8f8f42c97390088468f0a76
- Build time: 2023-03-23 09:19:34 UTC

jbordonado:
I looked at the `subscriptions` tests, and there is an issue for most tests due to this line: 
`await withRestApi.deleteCustomerByEmail( customerBilling.email );` 

This API from [@woocommerce/e2e-utils](https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/e2e-utils#rest-api-withrestapi) attempts to reassign (parameter `reassign` in the query) all posts to an user with ID=1 before deleting the other targeted user. See below:

https://github.com/woocommerce/woocommerce/blob/c3bec0c827ca4ba4cabfc76879ce7d019a8a842b/packages/js/e2e-utils/src/flows/with-rest-api.js#L374-L394

The problem is that we won't have any user with ID=1 on an Atomic site.

Quote from [a post](https://atomicp2.wordpress.com/2019/07/19/broken-user-tokens-being-created-with-incorrect-external_user_id/#comment-1130) on Atomic P2:
> When the Atomic site is created, all users are given a user ID that matches their WP.com user ID

In my case, I have a user with several digits, so any new user added will be just that digit + 1, so it's never starting from ID = 1.

Given that deleting the subscription user before most of these individual tests is mandatory (otherwise, an error is thrown at checkout), it prevents us from running them consistently on the CI for now.

Just adding a piece of code like this to the API make it work:
```
const allUsers = await client.get(userEndpoint);
const reassignUserId = allUsers.data?.find( user => user.name === 'admin')?.id || 1;
```

In this case, I assume by default we would want to reassign any post from a deleted user to the default admin of the site (username=admin).
So a code that would take the ID from the user with username="admin" by default, or `1` if there's no such user (today's behavior) would solve that problem I'm experiencing

tests/e2e/specs/wcpay/merchant/merchant-admin-analytics.spec.js
```diff
@@ -36,10 +36,9 @@ describe( 'Admin Order Analytics', () => {
 		// Go to Action Scheduler
 		await merchantWCP.openActionScheduler( 'pending', orderId );
 		const importOrderRun = await page.$x(
-			'//code[contains(text(), "0 => ' +
-				orderId +
-				'")]/ancestor::tr//span[contains(@class, "run")]/a'
+			`//tr[contains(., "0 => ${ orderId }") and contains(., "${ actionSchedulerHook }")]//span[contains(@class, "run")]/a`
```
jbordonado, line 39: In the case of an Atomic site (having different plugins enabled by default), it is necessary to be more specific in the selector with the action's name, because there usually is another action at the top of the list (coming from AutomateWoo for instance).

tests/e2e/specs/wcpay/shopper/shopper-pay-for-order.spec.js
```diff
@@ -14,7 +10,6 @@ const { shopperWCP } = require( '../../../utils' );
 describe( 'Shopper > Pay for Order', () => {
 	beforeAll( async () => {
 		await shopper.login();
-		await createSimpleProduct();
```
jbordonado, line 17: This would pollute the long-lived Atomic site with countless "Beanie" products, whereas there already exists one by default (since the initial setup phase requires importing some products through a CSV file).
So I didn't see a point in keeping this, we'll just target the existing "Beanie" product.

.github/workflows/e2e-tests-atomic.yml
```diff
@@ -0,0 +1,70 @@
...
+  
+  # Run UPE tests against WP Nightly & WC latest.
+  wp-nightly-tests-upe-batches:
+    name: "WP - nightly | WC - latest | ${{ matrix.test_groups }} - ${{ matrix.test_branches }}"
+    runs-on: ubuntu-22.04
+    needs: wp-nightly-tests-first-batch
+    if: success() || failure()
+    strategy:
+      fail-fast: false
+      max-parallel: 1 # Only one UPE group at a time to avoid conflicts between them
```
jbordonado: The `upe` and `upeSplit` can't be executed at the same time (conflicts on the setting itself), but to keep the file concise, I grouped them in the same matrix using a `max-parallel` parameter of 1

jbordonado:
To summarise:
- `subscriptions` tests cannot be executed reliably right now, due to an issue with the `deleteCustomerByEmail` API in the context of an Atomic site (wrong ID used in the query to reassign deleted user's posts)
- `upe` and `upeSplit` must run one after the other, only after the first batch is done, to avoid settings conflicts between group/branch combinations
- we'll need to implement a proper `setup` step (see https://github.com/Automattic/woocommerce-payments-meta/issues/82 and https://github.com/Automattic/woocommerce-payments-meta/issues/83) to keep things easy in the future, when we'll want a different Atomic site

Latest run for reference: https://github.com/Automattic/woocommerce-payments/actions/runs/4467275891

.github/workflows/e2e-tests-atomic.yml
```diff
@@ -0,0 +1,52 @@
...
+  cancel-in-progress: true
+
+jobs:
+  # Run tests against WP Nightly & WC latest. Only one combination at a time.
+  wp-nightly-tests:
+    name: "WP - nightly | WC - latest | ${{ matrix.test_groups }} - ${{ matrix.test_branches }}"
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
jbordonado, line 30: Since we use the same live site for each combination, we cannot run them in parallel due to conflicts on settings etc...

tests/e2e/utils/payments.js
```diff
@@ -32,16 +32,16 @@ export async function fillCardDetails( page, card ) {
...
 		} );
+		await page.waitFor( 1000 );
 
 		const cardCvcInput = await stripeFrame.waitForSelector(
 			'[name="cvc"]'
 		);
 		await cardCvcInput.type( card.cvc, { delay: 20 } );
+		await page.waitFor( 1000 );
 
-		if ( null !== ( await page.$( '#add_payment_method' ) ) ) {
```
jbordonado, line 41: This is not a good check because this just checks whether the form is displayed.
The displayed zip code field depends on the country selected (US), and we are currently not changing the country field in our tests. In my case, "France" was selected by default and the zip field was not present, hence the test failed.

tests/e2e/utils/payments.js
```diff
@@ -56,19 +56,24 @@ export async function fillCardDetails( page, card ) {
 			{ timeout: 30000 }
 		);
 		await cardNumberInput.type( card.number, { delay: 20 } );
+		await page.waitFor( 1000 );
```
jbordonado, line 59: This test would sometimes fail (especially in headless mode, rarely in interactive mode) due to an issue with the input value typed in the field. No idea what exactly is the chain of events resulting in this (I'm assuming something related to Puppeteer typing too fast somehow) I find that adding these pauses helped with the stability of this test overall...

jbordonado:
Latest run: https://github.com/Automattic/woocommerce-payments/actions/runs/4489423989

Only `wcpay | shopper`, `upe` and `upeSplit` tests

.github/workflows/e2e-tests-atomic.yml
```diff
@@ -0,0 +1,52 @@
...
+      matrix:
+        test_groups: [ 'wcpay', 'subscriptions', 'upe', 'upeSplit' ]
```
deepakpathania, line 32: Should we keep this matrix updated based on the ones we're planning to begin with instead of the `exclude` groups? 🤔 We can then add and update here once we enable more, not sure which approach seems better so starting this conversation.

tests/e2e/utils/payments.js
```diff
@@ -56,19 +56,24 @@ export async function fillCardDetails( page, card ) {
...
 		const cardDateInput = await stripeFrame.waitForSelector(
-			'[name="exp-date"]'
+			'[name="exp-date"]',
+			{ timeout: 30000 }
 		);
 
 		await cardDateInput.type( card.expires.month + card.expires.year, {
 			delay: 20,
 		} );
+		await page.waitFor( 1000 );
```
deepakpathania, line 69: Not a huge fan of timeouts like these in tests tbh since we never really know whether they're useful, I have seen waiting for certain selectors or for the page to finish with network calls is better in terms of maintainability. WDYT?

.github/workflows/e2e-tests-atomic.yml
deepakpathania: This could be an issue considering the linear nature of this, the tests would take upwards of an hour if we enable them all and run them linearly so I guess we'll always have to run a subset of tests on atomic anyways. Unless we write specific tests only for atomic.

deepakpathania (commented):
Left some comments.

.github/workflows/e2e-tests-atomic.yml
jbordonado: I don't have a preference.

I initially chose this approach to explicitly see what is currently missing compared to the regular E2E tests, but that's a detail, let me know if you prefer the other way around

tests/e2e/utils/payments.js
jbordonado: I agree 100%, but in this case, this is not about network calls, just about typing a value in a field. I tried multiple unsuccessful solutions before using this, which seems to help.

I ran a lot of times in a case where `tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js` would have 1 or 2 tests fail because somehow the card number input didn't reflect exactly what we tried to fill in.
I'm not sure exactly what the problem is, something related to Puppeteer typing very fast (too fast?) and not waiting correctly for the next step...?

Take this run for instance: https://github.com/Automattic/woocommerce-payments/actions/runs/4487977959

The test `should throw an error that the card expiration date is in the past` should type the card number `4242424242424242`. Still, you can see in the artifact's screenshots that the card number's last 4 digits are `0004`, and so instead of seeing the expected error, we see the "card number is invalid" error because obviously this card number is wrong. I'm not even sure how we end up with this number.

There's an issue on the Puppeteer repo (https://github.com/puppeteer/puppeteer/issues/1648) that is closed, but people are still adding comments to complain about this problem, even on recent versions.

I'm open to suggestions if you have any though

.github/workflows/e2e-tests-atomic.yml
jbordonado: Yeah, that's something we'll need to think about.

Is it an issue if we just need to run them once a day (when most people are not working timezone-wise) and for the release (upon code freeze, which is on a Sunday, so it should not block anything)? I don't think we should worry about the time

We can also focus on a set of specific smoke tests for Atomic (which should take no more than 10 or 15 mins) for faster feedback loop on some changes.

deepakpathania (approved):
Tests are passing locally for me in `dev` mode and in CI. Headless mode locally seems to get stuck but that could be a local set up issue.

Screenshot 2023-03-23 at 3 01 14 PM


Things to note:
- single hosted site so multiple runs in parallel can lead to issues.
- should be used as an additional area of coverage instead of a required check for now since tests cannot be parallelized.
- limited tests being run on Atomic due to the nature of tests being too brittle and confined to non atomic environment

.github/workflows/e2e-tests-atomic.yml
deepakpathania: Ok with this for now.

tests/e2e/utils/payments.js
deepakpathania: Fair, let's revisit in the future in case we're able to debug the flaky nature better.

jbordonado:
Additional things to note:
- The site could be stuck in an improper state for the next runs if a test is not properly written (teardown) or fails to cleanup once it is done 
- We need to see how much setup we can automate when the site is first created (https://github.com/Automattic/woocommerce-payments-meta/issues/82)
- We need to see how much setup/cleanup we can automate when we're about to run the tests on the site (https://github.com/Automattic/woocommerce-payments-meta/issues/83), so that the site is close to the initial setup when it was first created.

Issue #4748, opened 2022-09-12
Implement Things To Do for Progressive Onboarding (WCPay overview)


dmallory42:
### Description

We need to implement Things To Do to prompt a Progressive Onboarding user to complete onboarding in the WCPay overview. The urgency should depend on how close they are to the time/volume thresholds.

Rules: 
- After payments are enabled: no notification
- After the first sale is complete: default info notification
- Balance is rising or 7 days from the first sale (1000 USD now in design):: the orange/warning color notif and slightly different message
- Nearing thresholds - 21 days after the first sale or reaching 3000 USD (as it's now in design): The most urgent notification with the red color and error icon
- After the limit is hit: Not sure how to handle this yet, but I think we should also have a message for when the user's payments have been disabled because they hit a limit (either 30 days or $5000 USD)
- Successful onboarding: no notification

### Relevant Links

Designs for PO can be found here: s12xuy2TEGi3F91coVmxpF-fi-15788%3A257304

Unsure whether the exact wording of inbox notes has been defined, we can use this issue or the Slack channel to communicate with the Design team about this prior to working on this issue.

PR #5766, opened 2023-03-14
Fix bug with showing WC tasks when there is no WCPay account

oaratovskyi:
Fixes bug related to merge of #4748 

#### Description
GlobalStep found an error p1678777814525669-slack-C011ENB20Q1. When there is no WCPay account onboarded (right after install of WCPay), the task list in WC Home is not shown.

- Change the deconstruction of the params so the code will execute in any case. For example, Reconnect WPCOM task will be available even if there is no WCPay account (nothing new, was working like that before).
- Add test to cover this case

#### Testing instructions

**Develop**
* Remove the account from local
* Clear the cache
* Open WC Home and check there are no WC tasks at all

**The fix**
* Check out the branch `fix/4748-check-if-account-status-is-not-error`
* Build the changes locally `npm run watch`
* Open WC Home and check there are WC Tasks but without WCPay tasks
* Onboard the account
* Check that task list is still visible

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-4748-check-if-account-status-is-not-error
+++ b/changelog/fix-4748-check-if-account-status-is-not-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix bug with showing WC tasks when there is no WCPay account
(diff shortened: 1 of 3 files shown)


ismaeldcom (approved):
Thanks for taking care of the fix so quickly! The solution works as expected, and missing test will ensure we don't miss it again 👍

PR #5768, opened 2023-03-14
Add workflow "Build live branch for PR"

### LINKED ISSUE #5624: Build up WCPay live branches: add a new workflow
### Description

We'd like to have an easy way for testing any WCPay branch/PR. 

### Acceptance criteria

* Build up a new GitHub workflow, or integrate one of current GitHub workflows to send built files info from GitHub Actions to betadownload.jetpack.me ([see the same for Woo core](https://github.com/woocommerce/woocommerce/blob/a45f9c4f5544b7a2b4d116baac6800ee0317758a/.github/workflows/pr-build-live-branch.yml#L86)). This is an effort to build zip file: https://github.com/Automattic/woocommerce-payments/pull/5523

Moving two following tasks regarding integrating Jurassic Ninja to a separate meta repo issue: https://github.com/Automattic/woocommerce-payments-meta/issues/92

### Additional context

- pc2DNy-2Dm-p2#comment-4249
- pcShBQ-BZ-p2
- paJDYF-3gj-p2
- pdqwkG-eT-p2
- https://github.com/Automattic/jurassic.ninja/pull/274

htdat:
Fixes #5624

Add a new GitHub workflow, besides building the file, and notifying the Jetpack Beta builder (betadownload.jetpack.me), it would do something more for that purpose: 

- Play around the version to have the dev version, and modify the built files.
- Tweak the file folder of artifacts in accordance with Jetpack Beta Builder. See p1678811397206069/1678259021.463009-slack-CBG1CP4EN
- Add and update comment to indicate where to download the zip file. See the `botwoo` comment below. This would be useful to download the build file too. I think it might be helpful for people outside A8C.

A note on this comment: 

> Building a zip file already exists: https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/build-zip-and-run-smoke-tests.yml
> 
> Source: https://github.com/Automattic/woocommerce-payments/issues/5624#issuecomment-1464028911

I am not using this workflow because:

1. the artifact retention day for live branch needs to be only 1 day.
2. Jetpack Beta Builder, i.e. betadownload.jetpack.me, has a specific requirement for the artifact. 

#### Testing instructions

- Check the run for the new action in this PR.
- Check the comment from `botwoo`.
- Confirm: https://betadownload.jetpack.me/woocommerce-payments-branches.json showing the new file for this PR. (Usually, it's almost instant, but sometimes I guess you might wait a bit). 

It would be similar to other files such as https://betadownload.jetpack.me/woocommerce-branches.json, which is eventually used to populate the options for WCPay in Jurassic Ninja.

*

-------------------

- [x] NO NEED Covered with tests (or have a good reason not to test in description ☝️)

**Post merge**

- [x] NO NEED Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [x] NO NEED Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [x] NO NEED Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.


botwoo:

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5624-live-branch-support) :rocket:

- Latest commit: 9274327c05e63f2f4b2bb00a32d11ed96b2b67e9
- Build time: 2023-03-17 08:45:57 UTC

htdat:
Ready to review.

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,123 @@
+name: Build live branch for PR
+
+on:
+  pull_request:
```
htdat, line 4: May change this (equal to manual triggers) and we will slowly roll out this workflow to all PRs later? 

```suggestion
  workflow_dispatch:
```

naman03malhotra: I agree, manual should be fine to begin with and later we can discuss about creating it for every PR

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,125 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
deepakpathania, line 8: This would cancel the workflow if a subsequent commit is pushed and the previous one isn't complete. Do you want that behavior for this workflow or would having one for each commit be better?

deepakpathania (commented):
I might be missing something but is this more about getting a zip for a branch while publishing that zip to a directory? With the PR title, I expected that we would get a URL (maybe JN or similar) with the plugin zip installed from the branch but that doesn't seem to be the case. 🤔

.github/workflows/pr-build-live-branch.yml
htdat: Yes, I'd like that behavior. Because eventually betadownload.jetpack.me will only save the last build.

htdat:
> I might be missing something but is this more about getting a zip for a branch while publishing that zip to a directory? 

Yes, it builds the zip file, and then sends the meta information to betadownload.jetpack.me.
And then betadownload.jetpack.me will process this information including: 

- Grab the artifact file, and save in its server. E.g. for this PR it's this link https://betadownload.jetpack.me/data/woocommerce-payments/add_5624-live-branch-support/woocommerce-payments-dev.zip
- Generate the info in https://betadownload.jetpack.me/woocommerce-payments-branches.json

> With the PR title, I expected that we would get a URL (maybe JN or similar) with the plugin zip installed from the branch but that doesn't seem to be the case. 🤔

Yeah, the `live branch` concept is from Jetpack I guess so I agree that it's a bit confusing.

htdat:
@deepakpathania - Anyway, this is the first step. Eventually, I will need to create TamperMonkey script like this one https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-beta-tester/userscripts/wc-live-branches.user.js, which be handled in another issue https://github.com/Automattic/woocommerce-payments-meta/issues/92

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: php .github/workflows/scripts/get-next-version.php # outputs.NEXT_RELEASE_VERSION
+
+      - name: "Bump dev version"
+        id: bump-dev-version
+        env: 
+          NEXT_VERSION: ${{ steps.next-version.outputs.NEXT_RELEASE_VERSION }}
+          PR_NUMBER: ${{ github.event.number }}
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          CURRENT_VERSION=$(jq '.version' package.json -r)
```
jbordonado, line 43: It's already available in `steps.current-version.outputs.VERSION`

jbordonado: It's already available in `steps.current-version.outputs.VERSION`

deepakpathania:
Thanks for the clarification, @htdat. It makes a lot more sense to me now.

I believe we could potentially leverage the existing `Build zip` workflow and add the required metadata and call to the registry for this purpose. Or maybe split that workflow into [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to reuse some of that functionality. With an additional workflow like this, we have some repetitive code spread across workflows and any future changes to them would need to be duplicated.

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,125 @@
...
+
+      - name: "Build the plugin"
+        id: build_plugin
+        uses: ./.github/actions/build
+
+      - name: "Update the structure of artifacts for Jetpack Beta Builder" 
+        run: |
+          cd release
+          mv woocommerce-payments woocommerce-payments-dev
+          zip -q -r "woocommerce-payments-dev.zip" "woocommerce-payments-dev/"
```
jbordonado, line 68: By doing this, when you click to download the artifact on the GitHub UI, you will get a double zip (a zipped folder containing the right zip file), and we would not be able to use that directly on a WP site to upload the plugin manually.
We'd need to unzip the downloaded content, and then use the zip inside, which is cumbersome.

This is what the file size notice below is about in the step after this.
In [the other workflow](https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/build-zip-and-run-smoke-tests.yml), we are simply uploading the release folder, to avoid this double-zip issue.

You may just:
- go into `release` folder
- rename the folder `woocommerce-payments` to `woocommerce-payments-dev`

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,125 @@
...
+          
+          :x: This build file is NOT ready for production. It's available for testing only.
+
+          - Latest commit: ${PR_HEAD_SHA}
+          - Build file in the 'Artifacts' section: https://github.com/Automattic/woocommerce-payments/actions/runs/${GITHUB_RUN_ID}
+          - Expired at: ${EXPIRY_TIME}
+
+          _Note: this comment is updated when a new commit is pushed to this PR._"
+
+          PR_URL="https://github.com/Automattic/woocommerce-payments/pull/${PR_NUMBER}"
```
jbordonado: We can use the GitHub Context here:
- `github.server_url` for `https://github.com` 
- `github.repository` for `Automattic/woocommerce-payments`

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,125 @@
+name: Build live branch for PR
```
jbordonado: "Prepare branch for Jetpack Beta Builder" would better describe what the workflow does?

Similarly for the workflow's filename

We can also add a general comment above to point out what it ends up doing exactly (make the artifact built available in `woocommerce-payments-branches.json` for X purpose etc...)

.github/workflows/pr-build-live-branch.yml
jbordonado: In case we want to switch to a manual trigger, I guess the code will need to be adjusted, as it tries to grab values only available in a `pull_request` context, and also it tries to add a PR comment.

jbordonado (commented):
This will be a great addition and facilitate the testing of features, thanks for working on this 👍 

I left some comments.

I also agree with Deepak regarding the code duplication and the potential for reusability.
We can keep this new functionality separate from the other `build-zip-file` workflow. However, we can still promote reusability by introducing composite actions with smart inputs, etc...
For instance, I like that you added a PR comment with helpful info regarding the zip being built, which we should also be doing in the original workflow. And the original workflow already fixed the double zip issue, which you reintroduced here (the same mistake as when we first built the original workflow). We will improve maintainability if we take the time.

That being said, maybe you'd like to go ahead with this, and leave a refactor for the next steps. I don't think it's a big problem to merge this after taking care of the feedback., we're not talking about a super hard workflow with a lot of steps here

.github/workflows/pr-build-live-branch.yml
htdat: I see the point of the double-zip, however, it has been done with the purpose to making the artifact compatible with the betadownload.jetpack.me requirements. You can see more in my PR description. That's also the reason I did not use https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/build-zip-and-run-smoke-tests.yml

jbordonado: Okay, I see. So it would be nice to add a note about this next to the artifact link on the PR comment (the problem is not obvious because unzipping the downloaded content on MacOS directly unzips the zip inside, too, so you never actually see that there is a double zip thing unless you inspect the zip with `zipinfo` command for instance. It can waste quite some time to understand the problem if you're unfamiliar with it).

.github/workflows/pr-build-live-branch.yml
htdat: I can change this but I just want it consistent with the same concept made by Jetpack and Woo core. So that it can be easier to recognize across A8C: 

- Jetpack Live Branch PCYsg-goI-p2
- WooCommerce Live Branch pcShBQ-BZ-p2

.github/workflows/pr-build-live-branch.yml
htdat: Anyway, the main purpose of this artifact is to feed into betadownload.jetpack.me.

The comment is just something nicer. I've just learned that by feeding this info into betadownload.jetpack.me, we do not need to do anything else, just use a link like this: 

- https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments={{github.head_ref}}

For example, with this PR, this link would work for any A12s (that has the access to have more options in JN) https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5624-live-branch-support

I will update this comment to expose this info rather than show the GitHub action run link.

jbordonado: Very nice.

So we wouldn't need the Tampermonkey script?

htdat: > So we wouldn't need the Tampermonkey script?

That's right. We do not need it for a quick test. But we can add more options and it brings a better experience with the Tampermonkey script, such as the following one from Woo core. But again, this is for the other task :D 

Markup on 2023-03-16 at 17:52:58

.github/workflows/pr-build-live-branch.yml
htdat: This is not actually used so I just simply removed it.

.github/workflows/pr-build-live-branch.yml
htdat: I am going to simply use `gh pr comment {PR_NUMBER}` - commit 4cc91b57177476138d276d7d0dfbc2806c297f59

.github/workflows/pr-build-live-branch.yml
htdat: Updated the comment in this comment 4cc91b57177476138d276d7d0dfbc2806c297f59
You can also the updated comment https://github.com/Automattic/woocommerce-payments/pull/5768#issuecomment-1469972956

.github/workflows/pr-build-live-branch.yml
htdat: Yeah, it turns out that I still need the PR number to inform Jetpack Beta Builder so I think it would be better to use `pull_request` context. 

But more important as discussed in https://github.com/Automattic/woocommerce-payments/pull/5768#discussion_r1138451184, we can bring the benefit right away with the new comment including the link to spin up a new JN site.

.github/workflows/pr-build-live-branch.yml
htdat: I chose a balanced approach here: update the GitHub workflow name 3214566f40c904aa013f0358a9a7bcd876359c72, but keep the filename to have its consistency with [Woo core](https://github.com/woocommerce/woocommerce/blob/31ec8d8352a5a2b904eb299a32997ebfa961236f/.github/workflows/pr-build-live-branch.yml).

htdat:
@deepakpathania and @jbordonado  - ready to review again! 

Two notable changes: 

- 7b099f7 - Extract get-dev-version and get-plugin-data to better organize the code (and may help for potential refactor). 
- 4cc91b5 - Update the comment.

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,132 @@
...
+          GITHUB_TOKEN: ${{ secrets.BOTWOO_TOKEN }}
+          PR_NUMBER: ${{ github.event.number }}
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          PR_HEAD_REF: ${{ github.head_ref }}
+        run: |
+          BODY="### Test the build on Jurassic Ninja 
+          
+          This feature is only available for A12s. Please log in into https://jurassic.ninja/specialops/ before clicking the link below:
+          
+          https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=${PR_HEAD_REF}
```
deepakpathania: Possible to make this a `href` with let's say:
```
[Launch JN with branch](link)
```

deepakpathania (commented):
Left one cosmetic comment. Would also encourage to open a follow up issue for refactoring using composite actions.

jbordonado (approved):
Pre-approving, changes look good.

I think we're good to go after taking care of Deepak's latest comment (improve the link with text + follow-up issue)

.github/workflows/pr-build-live-branch.yml
htdat: Addressed 593486aad251d44f87f4245d11693960396e1bf8

deepakpathania (approved):
LGTM now ✅

htdat:
> Would also encourage to open a follow up issue for refactoring using composite actions. 

cc @jbordonado @deepakpathania - Opening this issue https://github.com/Automattic/woocommerce-payments/issues/5803

I will try to work on this in the next cooldown. But otherwise, Harmony may help move the issue forward. Thanks for helpful and constructive reviews here.

PR #5791, opened 2023-03-16
Add progressive onboarding form validation

### LINKED ISSUE #5474: Progressive Onboarding Form: Implement Validation
### Description

Add any validation logic needed on the form. This should match Stripe’s validation, to prevent us passing values that Stripe will not accept.

This will be a follow up to #5473.

### Acceptance criteria

* The form fields are validated in a way which matches both our server's, and Stripe's specifications of expected values.

ismaeldcom:
Fixes #5474

- Install `@testing-library/react-hooks` to make React hooks testing easier.
- Implement a reusable `Field` component, to wrap `TextControl` and `CustomSelectControl`, and provide error state support.
- Create `OnboardingTextField` and `OnboardingSelectField` to use `Field` while connecting them to `OnboardingContext`. Setting up label, value, change, and validation by `name`. Allowing to render fields by just choosing the onboarding field name, like: `<OnboardingTextField name="email" />`.
- Create `OnboardingForm` component to provide submit on `enter`, validation before submit, and `nextStep()` call.
- Add `errors` and `touched` states to `OnboardingContext` to be able to track modified fields and their errors. Allowing us to show errors only after a field has been edited or the form submitted.
- Create `useValidation` hook to handle field touched and errors on change, and return `validate()` and `error()` for field components. The error message comes from `strings` and there is a generic one in case there is no matching error.
- Basic validation logic, considering our current fields I think we don't need any third party library for validation. Making all fields required and checking email seems enough.
- Update every step to use the new simplified implementation.

#### Question
Should we use a custom phone field? We already use it for WooPay. In the design it looks like a regular text field. But I wonder if we need to get it in a specific format, like separated country code. As a reference, looks like Stripe Accounts API accepts a string with no additional validation.

```js
return alert( JSON.stringify( fromDotNotation( data ), null, 2 ) );
```

- Go to `/wp-admin/admin.php?page=wc-admin&path=/payments/onboarding-prototype`.
- Fill the first step fields using an invalid email (email without @).
- The email field should be in error state – red border and error message below.
- Fix the error, and click on **Continue**.
- Click on **Continue** in the second step.
- Errors should be shown for every field.
- Continue until the full flow is done.
- The alert should contain the provided data – If you opted to use the mentioned snippet. Otherwise, you can check it by using breakpoints or looking at prefilled details on Stripe.

-------------------

**Post merge**


client/onboarding-prototype/strings.tsx
```diff
@@ -73,6 +73,12 @@ export default {
 			'woocommerce-payments'
 		),
 	},
+	errors: {
+		generic: __( 'Please provide a response', 'woocommerce-payments' ),
+		email: __( 'Please provide a valid email', 'woocommerce-payments' ),
+		phone: __( 'Please provide a valid phone', 'woocommerce-payments' ),
```
dmallory42: Nitpick: I think `Please provide a valid phone number` reads better here. What do you think?

dmallory42 (approved):
Looks great and works well, thanks for working on this! I only had some minor comments so I'm pre-approving this!

I really like the approach taken here, everything is very well organised and makes it easy to add more validation rules in future!

> Should we use a custom phone field? We already use it for WooPay. In the design it looks like a regular text field. But I wonder if we need to get it in a specific format, like separated country code. As a reference, looks like Stripe Accounts API accepts a string with no additional validation.

It looks like Stripe [expect](https://stackoverflow.com/questions/57797200/correct-phone-number-format-for-stripe-connect-create-customer) the value here to be in international phone number format - I found a [helpful resource](https://support.twilio.com/hc/en-us/articles/223183008-Formatting-International-Phone-Numbers) about the format. I think it would be good in the future to have something like [this](https://uxplanet.org/phone-number-field-design-best-practices-23957cbd86d5) with validation enforcing that format, but I think that feels like a follow-on task rather than part of the initial issue. What do you think?

I was thinking something like this:
Screenshot 2023-03-21 at 11 10 56

I had one more small nitpick 🙂 with the email field, as soon as you start typing it will highlight as red with the `Please provide a valid email` message. Would it be possible to either enforce a bit of a delay, or wait until the user stops focusing on the field before trying to validate it?

client/onboarding-prototype/strings.tsx
ismaeldcom: Definitely! I was in automatic mode adding the name of the field to the error 😄  Addressed in 0fcefdb

botwoo:

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5474-po-form-validation) :rocket:

- Latest commit: a5c1371ae9d456069dd0966d47266da2a2550ac8
- Build time: 2023-03-21 12:06:20 UTC

ismaeldcom:
Thanks!

>but I think that feels like a follow-on task rather than part of the initial issue. What do you think?
>I was thinking something like this:

I agree, we can tackle it in #4744 or create a new issue. We already have an external lib `intl-tel-input` used in settings and WooPay, although it doesn't like well integrated, and it's a heavy dependency. We can probably look a bit for an alternative. It looks like this:
Screenshot 2023-03-21 at 08 40 48

>Would it be possible to either enforce a bit of a delay, or wait until the user stops focusing on the field before trying to validate it?

I thought about it too, and ended leaving it as it is. Another option could be to validate only on submit, in this way there are no errors while you're filling the form, which will be the case most of the time. What do you think?

dmallory42:
> I agree, we can tackle it in https://github.com/Automattic/woocommerce-payments/issues/4744 or create a new issue. We already have an external lib intl-tel-input used in settings and WooPay, although it doesn't like well integrated, and it's a heavy dependency. We can probably look a bit for an alternative. It looks like this:

Agreed, it looks like a good candidate to tackle in #4744 IMO! 

> I thought about it too, and ended leaving it as it is. Another option could be to validate only on submit, in this way there are no errors while you're filling the form, which will be the case most of the time. What do you think?

It's tough because it can save time to see the errors before submitting, but maybe on submit is a bit friendlier for users, and most of the time there would be no errors.

I don't think there's a "wrong" answer really - I wonder though, what does the main Woo Onboarding use? It could be good to take the same approach for a consistent feel.

ismaeldcom:
I checked both WC and Stripe flows, and chosen to go with the Stripe one. Woo Onboarding only have one field with validation, country/region, because the rest of them are optional. And they choose to disable the continue button.

Maybe we should switch to the WC way, it's just that I don't like disabled buttons, they look odd and have low accessibility due to the poor contrast 🙂

dmallory42:
> I checked both WC and Stripe flows, and chosen to go with the Stripe one. Woo Onboarding only have one field with validation, country/region, because the rest of them are optional. And they choose to disable the continue button.
> 
> Maybe we should switch to the WC way, it's just that I don't like disabled buttons, they look odd and have low accessibility due to the poor contrast 🙂

That makes sense, thanks for checking it! I think this is good to go since we'll tackle the phone input in another issue. Thanks for working on this!

Issue #5770, opened 2023-03-14
New UPE - Initial minimum viable product


bborman22:
### Description
Please refer to pc2DNy-2e2-p2 for more details.

> [Stripe has informed us that they have] added two important payment flow features with our desires in mind: the payment intent is no longer required prior to creating the payment method and the payment element can be used singularly to create the payment method, allowing us to create and confirm the payment intent ourselves on the Payments Server. This feature would solve our security concerns over exposed payment intent information on the client and would allow the UPE to be compatible with WooPay, since we can use our legacy payment intent confirmation flow with the UPE.
>
> Unfortunately, this feature is not yet ready. Stripe plans to open up a private beta for this feature by the end of October and plans to fully release the feature by the end of Q1 next year.

Please refer to [this file](https://wootransactp2.files.wordpress.com/2022/10/new-integration-path-for-payment-element.pdf.zip) for the current description of this new payment flow. 

If you scroll down to the section titled "Server-side confirmation", you will see that this almost exactly matches our current payment flow for our legacy card element, where the payment intent is created and confirmed simultaneously and subsequent to creating the payment method. This will allow us to resolve almost all of our current security and integration concerns with our existing UPE implementation. 

This issue exists to create the initial PoC implementation of this new payment flow, once the tool has been released to beta and we are given access to it by Stripe. This issue is simply to implement the basic shortcode checkout payment flow and there will be other issues created to support other checkout and payment flows. It's also assumed that this tool will be in a fluid state while in limited beta release, so this PoC will not need to be considered in a final state once completed. 

### Acceptance criteria

* Implements Enhanced UPE shortcode checkout flow.

### Additional context

[Stripe UPE – Options to solve WooPay compatibility and WC Pay loss of control over the payments UI/UX](pc2DNy-2e2-p2)
[New integration path for Payment Element](https://wootransactp2.files.wordpress.com/2022/10/new-integration-path-for-payment-element.pdf.zip)

FangedParakeet:
Just a heads up, but I've renamed this issue from a POC to an MVP, because the scope of our efforts is much greater than a puny prototype and we are consequently expending as much energy to possible to create an end-product that may be limited in its initial ability, but that is still robust enough to be released and reused.

timur27:
@FangedParakeet There's one thing regarding the 3DS authentication for credit cards we'd need to find the best approach to handle. 

For the legacy card (non-UPE), the flow looks the following: 

[...]

Issue #5028, opened 2022-11-01
Streamline UPE/non-UPE code branching and eliminate UPE feature flag checks


FangedParakeet:
### Description
This issue is to be tackled post `6.7.0` release, once all the merchants are migrated to the split UPE with deferred intent creation. The further description is based on this assumption. 

We have four different gateway approaches in the plugin: 
* legacy card
* legacy UPE
* split UPE
* deferred intent creation UPE

The first three are not used anymore. At the same time, there are a lot of places in the code where we check which flag out of these above is enabled and execute the flow according to it. This means that we need to remove all the UPE-related code branching other than deferred UPE.

Acceptance criteria
* eliminate all branching where any of the following is a part: `is_upe_deferred_intent_enabled()`, `is_upe_split_enabled()`, `is_upe_legacy_enabled()` by keeping only the code related to the deferred intent creation UPE
* remove `is_upe_deferred_intent_enabled()`, `is_upe_split_enabled()` and `is_upe_legacy_enabled()` located in `class-wc-payments-features.php`, as well as clean up all the areas that depend on those checks
* improve test coverage wherever possible

frosso:
Discussed internally - We're going to clean this up gradually in multiple PRs.

So we'll likely clean up each version of the UPE, one by one.
- Legacy card
- Split UPE
- Legacy UPE (likely last - it has a larger impact)
- Logic related to the Deferred Intent UPE feature flag (Deferred Intent UPE is the winner)

PR #5811, opened 2023-03-17
Adds new UPE feature flag

### LINKED ISSUE #5682: New UPE: Create new feature flag for upcoming implementation
### Description

Here are the [existing UPE feature flags](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-features.php#L16-L17). We will want to develop our new UPE implementation in #5019 without having any impact on this existing implementation. Consequently we will cower behind the cover of a feature flag until we are ready to face the light of day.

This issue is to add this new feature flag that will be used in the development of #5019. 

### Acceptance criteria

- Unique new feature flag must exist.
- Flag must have a cool name (I think New UPE appears to be the current convention, but always open to suggestions).
- New feature flag will override other UPE flags when enabled.
- New feature flag must be included in [existing UPE logic](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-features.php#L37-L62).
- (For the foreseeable future, [feature flag eligibility](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-features.php#L74) will still be denied when SEPA enabled.)

### Additional context

#5019

FangedParakeet:
Fixes #5682 

Adds a new feature flag for the new UPE/deferred intent feature to hide behind. 

Please see, review, and approve Automattic/woocommerce-payments-dev-tools#86 to assist with toggling this feature flag. This flag doesn't do anything yet, but will be used in #5770. Decisions on how enablement and disablement of this feature flag can be left for other issues.

#### Testing instructions

This change does not yet effect any changes, so you will have to check your `wp_options` table to ensure the flag has been added and enabled on your site.

1. Pull the changes in this PR and Automattic/woocommerce-payments-dev-tools#86 and ensure both are in use on your site.
2. Go to the Dev Tools, ensure the checkbox titled "Enable Split UPE checkout with deferred intent creation" is checked, then submit the page.
3. On refresh, this checkbox should remain checked.
4. In your site's `wp_options` search for the entry with `option_name` of ~`woocommerce_woocommerce_payments_settings`. Within this PHP serialised array, you should find a key with value of `_wcpay_feature_upe_deferred_intent` now~ `_wcpay_feature_upe_deferred_intent` with a value of `1`.
5. Return to dev tools, uncheck the same checkbox, and submit the page.
6. On refresh, the checkbox should remain unchecked. In your site's options table, ~in `woocommerce_woocommerce_payments_settings` object~, `_wcpay_feature_upe_deferred_intent` should now ~have a value of `0`~ be removed from the table.
7. Lucky number seven.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5811` or branch name `add/5682-add-upe-deferred-intent-feature-flag` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5682-add-upe-deferred-intent-feature-flag) :rocket:

---

Build info: 

- Latest commit: 02229b8571f24dd2e87657cdb5e45050b3f33695
- Build time: 2023-03-30 22:33:14 UTC

includes/class-wc-payments-features.php
```diff
@@ -69,6 +71,13 @@ public static function is_upe_split_enabled() {
 		return '1' === get_option( self::UPE_SPLIT_FLAG_NAME, '0' ) && self::is_upe_split_eligible();
 	}
 
+	/**
+	 * Checks whether the Split UPE with deferred intent is enabled
+	 */
+	public static function is_upe_deferred_intent_enabled() {
+		return '1' === get_option( self::UPE_DEFERRED_INTENT_FLAG_NAME, '0' ) && self::is_upe_split_eligible();
+	}
```
timur27, line 82: Could we cover this method with tests? 

Something like 
```php
	public function test_is_upe_deferred_intent_enabled_returns_true() {
		add_filter(
			'pre_option__wcpay_feature_upe_deferred_intent',
			function ( $pre_option, $option, $default ) {
				return '1';
			},
			10,
			3
		);
		$this->mock_cache->method( 'get' )->willReturn( [
			'capabilities' => [
				'sepa_debit_payments' => 'inactive'
			]
		]
		);
		$this->assertTrue( WC_Payments_Features::is_upe_deferred_intent_enabled() );
	}
```

to test the method returning `true` would be sufficient. And two more for the `false` scenario. Thank you!

timur27 (commented):
Nice one! All the changes make sense for me. The only thing are tests, and I left a separate comment for them.

In terms of manual tests - I've tried to verify the flag value in the database. For me, it's always a separate row in the db table for this flag and not the part of `woocommerce_woocommerce_payments_settings` option. The row appears in the table when the checkbox is checked, and disappears for the unchecked checkbox.

image

And I can't see `_wcpay_feature_upe_deferred_intent` key in the `woocommerce_woocommerce_payments_settings`'s value. I'm pretty sure everything is working fine, but I'd like to confirm on the testing steps to ensure that I'm doing everything as expected.

includes/class-wc-payments-features.php
FangedParakeet: I liked this example so much that I fixed the existing UPE feature flag tests to use this more atomic structure and added a few new assertions for the new feature flag as well--see 005e17986bf1814682fa5ad3b035f3192835da2f if you don't believe me. Nice one!

tests/unit/test-class-wc-payments-features.php
```diff
@@ -160,7 +159,7 @@ function ( $pre_option, $option, $default ) {
 
 	public function test_is_woopay_express_checkout_enabled_returns_false_when_platform_checkout_eligible_is_false() {
 		add_filter(
-			'pre_option__' . WC_Payments_Features::PROGRESSIVE_ONBOARDING_FLAG_NAME,
+			'pre_option_' . WC_Payments_Features::PROGRESSIVE_ONBOARDING_FLAG_NAME,
```
FangedParakeet, line 162: Oops, fixed this teeny typo I found dormant in the tests here.

FangedParakeet:
> For me, it's always a separate row in the db table for this flag and not the part of woocommerce_woocommerce_payments_settings option.

I don't know why I got the feature flag options mixed up with the payments settings, which also uses the `wp_options` as I described. 🤦  You're totally correct: `_wcpay_feature_upe_deferred_intent` should be out on its own. 

By the way, I used the key name `_wcpay_feature_upe_deferred_intent` instead of something like `_wcpay_feature_new_upe`, because I felt that despite New UPE being the currently accepted and more widely known moniker, "new" is a term that quickly grows old. 😁 However, you are the skipper of the rest of this implementation, so if you feel a name change here might be more consistent, please let me know and I'll snap to attention and follow your lead!

timur27:
Just another minor thing. 

> Decisions on how enablement and disablement of this feature flag can be left for other issues.

Should we create one now to have this in the backlog? I haven't found an issue that would cover the enablement logic, but let me know if there is some. 👍

timur27:
> By the way, I used the key name _wcpay_feature_upe_deferred_intent instead of something like _wcpay_feature_new_upe, because I felt that despite New UPE being the currently accepted and more widely known moniker, "new" is a term that quickly grows old.

This makes sense to me, and I found `_wcpay_feature_upe_deferred_intent` to be more acceptable than all the other currently known options, which in turn would end up with `_wcpay_feature_new_upe` or `wcpay_feature_enhanced_upe`.

I went through [the Stripe documentation](https://stripe.com/docs/payments/finalize-payments-on-the-server) and attempted to find out any naming convention that would distinguish the create-payment-intent-later approach from all the others but haven't found anything like that. 

Having all this said, `_wcpay_feature_upe_deferred_intent` is fine for me. If the upe with deferred intent will eventually be the only UPE available for customers, then we could update/unify the naming to be `_wcpay_feature_upe`.

timur27:
> 6. On refresh, the checkbox should remain unchecked. In your site's options table, in woocommerce_woocommerce_payments_settings object, _wcpay_feature_upe_deferred_intent should now have a value of 0.

I think when the checkbox is unchecked, there's no row with a `0` value in the db table at all, since what [this method](https://github.com/Automattic/woocommerce-payments-dev-tools/blob/700b195cad544bdbf52f5139dd427aada05a9bdd/woocommerce-payments-dev-tools.php#L469-L477) essentially does [it runs the `delete_option` for the unchecked option](https://github.com/Automattic/woocommerce-payments-dev-tools/blob/700b195cad544bdbf52f5139dd427aada05a9bdd/woocommerce-payments-dev-tools.php#L476) - the function that removes an option from a table. 

I've checked the already existing `_wcpay_feature_upe_split` feature flag against this theory and can confirm that it behaves the same (the row is removed with the unchecked option). Nothing needs to be changed, I think only amending the PR description would be fine, in case you confirm my essay here. Thanks!

includes/class-wc-payments-features.php
```diff
@@ -58,8 +59,12 @@ public static function is_upe_legacy_enabled() {
 			return true;
 		}
 
+		$upe_split_flag_value    = '1' === get_option( self::UPE_SPLIT_FLAG_NAME, '0' );
+		$upe_deferred_flag_value = '1' === get_option( self::UPE_DEFERRED_INTENT_FLAG_NAME, '0' );
+
 		// if the merchant is not eligible for the Split UPE, but they have the flag enabled, fallback to the "legacy" UPE (for now).
-		return '1' === get_option( self::UPE_SPLIT_FLAG_NAME, '0' ) && ! self::is_upe_split_eligible();
+		return ( $upe_split_flag_value || $upe_deferred_flag_value )
+			&& ! self::is_upe_split_eligible();
```
timur27, line 67: I thought the legacy UPE fallback with enabled SEPA was worth re-testing due to these changes. I've tested the fallback flow for both split UPE and deferred UPE and can confirm that it: 

1. Falls back successfully to the legacy UPE with SEPA enabled and split UPE enabled ✅
2. Falls back successfully to the legacy UPE with SEPA enabled and deferred UPE enabled ✅
3. Doesn't fall back to the legacy UPE with SEPA disabled and split/deferred UPE enabled ✅

includes/class-wc-payments-features.php
```diff
@@ -263,6 +275,8 @@ public static function to_array() {
 		return array_filter(
 			[
 				'upe'                     => self::is_upe_enabled(),
+				'upeSplit'                => self::is_upe_split_enabled(),
+				'upeDeferred'             => self::is_upe_deferred_intent_enabled(),
```
timur27, line 279: I suppose the intention here is to supply flag values to the front-end, since [the values are added to the JS script here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-payments-admin.php#L539-L543)? If yes, shouldn't we add these two values to [the schema](https://github.com/Automattic/woocommerce-payments/blob/develop/client/globals.d.ts#L10-L16)?

includes/class-wc-payments-features.php
timur27: Awesome, I noticed you covered the deferred flag within the test_split_upe_enabled_with_eligible_merchant, which is perfectly fine for me. 💯

includes/class-wc-payments-features.php
FangedParakeet: My intention here was _not_ actually to supply this to the front-end, but rather to make this available for the unit tests. I wanted to be able to include the UPE flags in [this array](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/tests/unit/test-class-wc-payments-features.php#L22-L24), so that I could use the `pre_option_` hook to set these options, as you suggested, and also use [this `tear_down()` function to remove the filters](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/tests/unit/test-class-wc-payments-features.php#L42-L45) after use.

However, merely adding the UPE flags to the `FLAG_OPTION_NAME_TO_FRONTEND_KEY_MAPPING` array constant would cause [this test](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/tests/unit/test-class-wc-payments-features.php#L64) to fail. We could add them to the front-end schema, though I am a little loathe to do so, since that doesn't seem completely required and many of the other settings returned in this array are not necessarily included in the schema either.

Although, if you did plan on using any of these constants on the front-end, this would be a worthy reason for their inclusion here--so let me know if that might be the case!

FangedParakeet:
> Should we create one now to have this in the backlog? I haven't found an issue that would cover the enablement logic, but let me know if there is some. 👍

When I referred to the enablement logic, I was more referring to how you intended on using functions like [this one](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/includes/class-wc-payments-features.php#L39) (e.g. considering new UPE as split or not), where `is_upe_deferred_intent_enabled()` would actually be used, or whether we need any interaction with Tracks or other feature flags (possibly requiring new functions like [this one](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/includes/class-wc-payments-features.php#L101)) when the feature flag is enabled/disabled.

I'm leaving this all in your hands, in the main integration issue, so hopefully that's what you had intended as well! We do not have an issue to plot a rollout plan for this feature, which I believe is closer to what you were asking about here. This would probably include the aforementioned Tracks integration. I'll create an issue for that this morning, as you've presumably suggested. 👍 But correct me, if I've misunderstood what you were asking at all!

FangedParakeet:
> If the upe with deferred intent will eventually be the only UPE available for customers, then we could update/unify the naming to be _wcpay_feature_upe.

I think this will eventually be the plan--likely handled in #5028.

> Nothing needs to be changed, I think only amending the PR description would be fine, in case you confirm my essay here. 

Essay confirmed! 😂 And I believe I already corrected my disinformation in the PR testing instructions. 👍 

Let me know if we're all clear to continue here, @timur27, or if there's anything else I can clean up or failed to address!

includes/class-wc-payments-features.php
timur27: Oh, I see now. 

> Although, if you did plan on using any of these constants on the front-end

No, I've had no such plans.

> We could add them to the front-end schema, though I am a little loathe to do so, since that doesn't seem completely required and many of the other settings returned in this array are not necessarily included in the schema either.

Makes sense to me, I would feel the same. Let's not add them to the schema then.

timur27:
> We do not have an issue to plot a rollout plan for this feature, which I believe is closer to what you were asking about here.

Yes, this is what I meant! Thanks for creating an issue - looking forward to estimating there. 

> When I referred to the enablement logic, I was more referring to how you intended on using functions like [this one](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/includes/class-wc-payments-features.php#L39) (e.g. considering new UPE as split or not), where is_upe_deferred_intent_enabled() would actually be used, or whether we need any interaction with Tracks or other feature flags (possibly requiring new functions like [this one](https://github.com/Automattic/woocommerce-payments/blob/add/5682-add-upe-deferred-intent-feature-flag/includes/class-wc-payments-features.php#L101)) when the feature flag is enabled/disabled.
I'm leaving this all in your hands, in the main integration issue, so hopefully that's what you had intended as well!

I haven't covered this in the POC, because I wasn't aware it was in the scope. We can continue the discussion [here](https://github.com/Automattic/woocommerce-payments/pull/5869) or [here](https://github.com/Automattic/woocommerce-payments/issues/5770), whether I need to cover something else on top of the existing implementation. Meanwhile, I think there's nothing else that blocks this PR from being merged, so I'm approving it!


PS I've updated the description for the 6th step to be more in line with reality. 👍

timur27 (approved):

PR #5852, opened 2023-03-24
Storing transaction fee in order metadata

### LINKED ISSUE #3257: Storing transaction fee in WooCommerce order post meta
### Description
Storing additional transaction data (specifically the transaction fee) in the post meta of the WooCommerce order, so it's accessible to 3rd party WooCommerce integrations.

### Designs
For example, major payment gateways like Stripe and PayPal commonly store this data in the post meta, and (optionally) display it in the WooCommerce order UI as well (less important to display in the UI than to actually have available in the post meta)

Screen Shot 2021-10-29 at 09 13 59

### Dev notes

See https://github.com/Automattic/woocommerce-payments/issues/3257#issuecomment-1313715165 for implementation details.

3rd party integrations - like our integration that connects WooCommerce to other platforms, can use this data when retrieving an order from WooCommerce, which can be very helpful for reporting and accounting purposes. 

### Additional context
Related to ticket number 4412706-zen referenced in https://github.com/Automattic/woocommerce-payments/issues/1484

jessy-p:
Fixes #3257

- The `application_fee_amount` is stored in order metadata as `_wcpay_transaction_fee`
- Introduced a function `attach_transaction_fee_to_order` in `WC_Payments_Order_Service`
- Saves at different points to cover all checkout flows
            -   `process_payment_for_order` (Regular Order)
            -  `process_redirect_payment` ( UPE)
            -  `process_webhook_payment_intent_succeeded` (Sometimes the redirect processing is not successful and charge data is found missing on order)
-  Displays the Transaction Fee on Order Screen

##### Screenshot
image

#### Testing instructions
- Checkout using different flows - Classic checkout, Block checkout, UPE, non-UPE, other Payment Methods, using saved card
- Verify that the fee charged by Stripe is same as the value saved into order metadata with key `_wcpay_transaction_fee`
- Verify that the same value is displayed on the Order Edit screen.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-5.7.0#test-order-transaction-fee-display


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5852` or branch name `add/3257-store-transaction-fee-meta` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/3257-store-transaction-fee-meta) :rocket:

---

Build info: 

- Latest commit: 3ccf491441c38a3e9d523d73fd571019541ef38b
- Build time: 2023-03-31 11:05:23 UTC

jessy-p:
@souravdebnath1986 Is the Admin screen modification OK (screenshot above)?
Also, is it required? https://github.com/Automattic/woocommerce-payments/pull/5852#discussion_r1150158717

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1270,6 +1270,8 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 			$payment_method_details = $charge ? $charge->get_payment_method_details() : [];
 			$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
 
+			$this->order_service->attach_transaction_fee_to_order( $order, $charge );
```
htdat: May move up this together with others similar actions such as `attach_intent_info_to_order` and `attach_exchange_info_to_order` to update metadata so that the same actions would come together? That would be similar to what is being done in UPE gateway class. 

:x: Also, 3DS cards through non-UPE flow will not work properly with the current position of this action:

- [This code blocks](https://github.com/Automattic/woocommerce-payments/blob/ffdbd96644b23d9a5e1ba2b424c27849e333a358/includes/class-wc-payment-gateway-wcpay.php#L1228-L1251) handles 3DS cards, i.e. `requires_action` intent status, and have something for `$return`. 
- Since `$return` is no longer null, the whole `process_payment_for_order` will be terminated earlier - [code ref](https://github.com/Automattic/woocommerce-payments/blob/ffdbd96644b23d9a5e1ba2b424c27849e333a358/includes/class-wc-payment-gateway-wcpay.php#L1259-L1261).

includes/class-wc-payments-order-service.php
```diff
@@ -643,6 +650,29 @@ public function attach_intent_info_to_order( $order, $intent_id, $intent_status,
...
+	/**
+	 * Given the charge, adds the application_fee_amount from the charge to the given order as metadata.
+	 *
+	 * @param WC_Order               $order The order to update.
+	 * @param WC_Payments_API_Charge $charge The charge to get the application_fee_amount from.
+	 */
+	public function attach_transaction_fee_to_order( $order, $charge ) {
+		try {
+			if ( $charge && $charge->get_application_fee_amount() ) {
+				$order->add_meta_data(
```
htdat: I think we can use update_meta_data, which can ensure that this meta key is unique.

_Edit:_ I see the same pattern of using update_meta_data for other meta keys. 

```suggestion
				$order->update_meta_data(
```

includes/class-wc-payments-order-service.php
```diff
@@ -643,6 +650,29 @@ public function attach_intent_info_to_order( $order, $intent_id, $intent_status,
...
+					self::WCPAY_TRANSACTION_FEE_META_KEY,
+					WC_Payments_Utils::interpret_stripe_amount( $charge->get_application_fee_amount(), $charge->get_currency() ),
+					true
+				);
+				$order->save_meta_data();
+			}
+		} catch ( Exception $e ) {
+			// Log the error and don't block checkout.
+			Logger::log( 'Error when saving application fee for the order ' . $order->get_id() . ': ' . $e->getMessage() );
```
htdat: Nitpick: using `transaction fee` as this is how we call it while Stripe uses another term `application fee`.

includes/class-wc-payments-webhook-processing-service.php
```diff
@@ -466,6 +467,11 @@ private function process_webhook_payment_intent_succeeded( $event_body ) {
 			$meta_data_to_update['_stripe_mandate_id'] = $mandate_id;
 		}
 
+		$application_fee_amount = $charges_data[0]['application_fee_amount'] ?? null;
+		if ( $application_fee_amount ) {
+			$meta_data_to_update['_wcpay_transaction_fee'] = WC_Payments_Utils::interpret_stripe_amount( $application_fee_amount, $currency );
+		}
+
```
htdat, line 475: Ideally, this would use the new order service method `attach_transaction_fee_to_order` but it seems it's not a trivial effort to build up `WC_Payments_API_Charge`. 

Just a flag here as a potential improvement to take advantage of Order_Service.

includes/class-wc-payments-order-service.php
```diff
@@ -66,13 +66,20 @@ class WC_Payments_Order_Service {
...
 	 * @const string
 	 */
 	const WCPAY_REFUND_STATUS_META_KEY = '_wcpay_refund_status';
 
+	/**
+	 * Meta key used to store WCPay transaction fee.
+	 *
+	 * @const string
+	 */
+	const WCPAY_TRANSACTION_FEE_META_KEY = '_wcpay_transaction_fee';
```
htdat, line 98: Can we use this constant for every place using this meta key?

includes/admin/class-wc-payments-admin.php
```diff
@@ -918,6 +919,30 @@ public function show_woopay_payment_method_name_admin( $order_id ) {
 		<?php
 	}
 
+	/**
+	 * Display the _wcpay_transaction_fee from order metadata to the Order Edit screen on admin.
+	 *
+	 * @param int $order_id order_id.
+	 */
+	public function display_wcpay_transaction_fee( $order_id ) {
```
htdat, line 927: I don’t think we need to expose the transaction fee like that. We have already added the fee breakdown note, which includes this info.

htdat (changes requested):
Thanks for working on this @jessy-p. I've left some initial feedback. 

Also, I just did a few manual testings. I will run more manual tests in the next review round.

Edit: some relevant PHPUnit tests failed. It's worth looking at them too.

includes/admin/class-wc-payments-admin.php
jessy-p: It is mentioned as a requirement #3257. I've checked with Product/Design.

includes/class-wc-payments-order-service.php
jessy-p: Done

includes/class-wc-payments-order-service.php
jessy-p: Changed.

includes/class-wc-payments-webhook-processing-service.php
jessy-p: Yes, building `WC_Payments_API_Charge` will be an overhead and including `wcpay_transaction_fee` into the already existing update and save is the more optimal way, so we are not using `attach_transaction_fee_to_order` over here.

includes/class-wc-payments-order-service.php
jessy-p: I don't think we need to introduce`WC_Payments_Order_Service` in the Admin dashboard display logic, just to use the constant. 

When we introduce a central constants file for all metadata keys we may move this, along with other metadata names used in the `WC_Payments_Webhook_Processing_Service`

jessy-p: I don't think we need to introduce`WC_Payments_Order_Service` in the Admin dashboard display logic, just to use the constant. 

When we introduce a central constants file for all metadata keys we may move this, along with other metadata names used in the `WC_Payments_Webhook_Processing_Service`

htdat: > I don't think we need to introduce`WC_Payments_Order_Service` in the Admin dashboard display logic, 

I did not mean this. 

> just to use the constant. 

And yes with this. 

> When we introduce a central constants file for all metadata keys we may move this, along with other metadata names used in the `WC_Payments_Webhook_Processing_Service` 

IMO, it's just easier to start using it now rather than waiting until a refactor. But anyway, it's not a blocker for this PR.

includes/admin/class-wc-payments-admin.php
htdat: If we have the confirmation from the Product/Design, let's do it. I just feel it redundant with what we have already had.

includes/class-wc-payment-gateway-wcpay.php
jessy-p: I guess it was placed there  because existing `$charge` object was being retrieved there. I moved it now, and all scenarios seem to work.
The new position does not help with 3DS though. 3DS with UPE works, possibly because of the code in the redirect methods. Non-UPE checkout + 3DS does not work either ways. Will look for another solution for that.

jessy-p: I guess it was placed there  because existing `$charge` object was being retrieved there. I moved it now, and all scenarios seem to work.
The new position does not help with 3DS though. 3DS with UPE works, possibly because of the code in the redirect methods. Non-UPE checkout + 3DS does not work either ways. Will look for another solution for that.

jessy-p: The 3DS + non-UPE orders gets updated with `_charge_id` only during webhook processing, so I think we should be OK with the similar solution for `_wcpay_transaction_fee` as well (it is working in that way now)

includes/admin/class-wc-payments-admin.php
jessy-p: @souravdebnath1986  Please let us know Previous question: https://github.com/Automattic/woocommerce-payments/pull/5852#issuecomment-1482657910

naman03malhotra: @elizaan36 can you please comment from the design's point of view? 
For context: please see the description of the PR and the screenshot.

elizaan36: I was going to mention the same thing as @htdat, that the transaction fee is broken down in the order notes and this is redundant. 

However, it sounds like more of an issue with extensibility that reporting and accounting integrations will find this new placement useful. If it helps with fees reporting overall then I'd say it's a good addition.

And it's exactly aligned with the Stripe placement and style so the design itself looks good to me.

marcinbot: One thing to consider is that there could be a lot of order notes (added by any other extensions), and the notes can also be deleted. Whereas this will always be present on the order page.

jessy-p: Based on the comments, I'm going to leave it in.

includes/admin/class-wc-payments-admin.php
```diff
@@ -918,6 +919,30 @@ public function show_woopay_payment_method_name_admin( $order_id ) {
...
+		}
+		?>
+		<tr>
+			<td class="label wcpay-transaction-fee">
+				<?php echo wc_help_tip( __( 'This represents the fee WooCommerce Payments collects for the transaction.', 'woocommerce-payments' ) ); // phpcs:ignore WordPress.Security.EscapeOutput ?>
+				<?php esc_html_e( 'Transaction Fee:', 'woocommerce-payments' ); ?>
+			</td>
+			<td width="1%"></td>
+			<td class="total">
+				-<<?php echo wp_kses( wc_price( $order->get_meta( '_wcpay_transaction_fee' ), [ 'currency' => $order->get_currency() ] ), 'post' ); ?>
```
htdat: Redundant `<` - [screenshot](https://user-images.githubusercontent.com/10045087/228738334-31bcb4d9-e80d-4776-abf6-c0f0e159170a.png)

includes/class-wc-payment-gateway-wcpay.php
htdat: Ah, I see now. For 3DS + non-UPE flow, before we re-direct to the order-received page, we send a AJAX request and it's handled in `WC_Payment_Gateway_WCPay::update_order_status`, so I think putting it inside this handling would be enough: 

https://github.com/Automattic/woocommerce-payments/blob/50ab8e33720c764c2e9a75b094083ae581b1770a/includes/class-wc-payment-gateway-wcpay.php#L2745-L2746

includes/class-wc-payments-order-service.php
```diff
@@ -875,6 +882,28 @@ private function mark_order_held_for_review_for_fraud( $order, $intent_data ) {
...
 
+	/**
+	 * Given the charge, adds the application_fee_amount from the charge to the given order as metadata.
+	 *
+	 * @param WC_Order               $order The order to update.
+	 * @param WC_Payments_API_Charge $charge The charge to get the application_fee_amount from.
+	 */
+	public function attach_transaction_fee_to_order( $order, $charge ) {
+		try {
+			if ( $charge && null !== $charge->get_application_fee_amount() ) {
```
htdat, line 901: IMO, we should declare param types here, and then force consumers to check these param types before passing them to use this method. 

With this, we do not need to check `$charge` inside this method (currently, consumers can actually pass `null`) and give the right expectation for consumers. 

Not a blocker but it would reduce the cognitive load in using this method in the future. 

```suggestion
	public function attach_transaction_fee_to_order( WC_Order $order, WC_Payments_API_Charge $charge ) {
		try {
			if ( null !== $charge->get_application_fee_amount() ) {
```

htdat (commented):
✅ I manually tested with UPE, non-UPE, block and non-block checkouts, and some subscription products. 

❓ They all work pretty well, except for the use case of non-UPE + 3DS cards. For that case, I think adding `attach_transaction_fee_to_order` in the `update_order_status` AJAX would work.
https://github.com/Automattic/woocommerce-payments/blob/50ab8e33720c764c2e9a75b094083ae581b1770a/includes/class-wc-payment-gateway-wcpay.php#L2745-L2746

--- 

With so many flows, I think it's worth checking sites tested by GlobalStep and double check to ensure all orders have this `_wcpay_transaction_fee` meta key.

includes/admin/class-wc-payments-admin.php
jessy-p: Sorry, typo! Fixed.

includes/class-wc-payments-order-service.php
jessy-p: > With this, we do not need to check `$charge` inside this method (currently, consumers can actually pass `null`) 
> Not a blocker but it would reduce the cognitive load in using this method in the future.

There are few flows which have (validly) `$charge` as null at some point, I would rather have the cognitive load within the function than its consumers (multiple points of the payment processing functions), since the code there is already quite nested and difficult to read.

jessy-p:
>adding attach_transaction_fee_to_order in the update_order_status AJAX would work.

since `$charge_id` is being added in webhook, I thought the `application_fee_amount` which is also charge related, makes sense to be updated there, and it also works for other flows that do not work otherwise. (Split UPE + non-card Payment Methods)

includes/class-wc-payments-order-service.php
```diff
@@ -883,6 +890,28 @@ private function mark_order_held_for_review_for_fraud( $order, $intent_data ) {
 		$this->set_intention_status_for_order( $order, $intent_data['intent_status'] );
 	}
 
+	/**
+	 * Given the charge, adds the application_fee_amount from the charge to the given order as metadata.
+	 *
+	 * @param WC_Order               $order The order to update.
+	 * @param WC_Payments_API_Charge $charge The charge to get the application_fee_amount from.
```
htdat: Nitpick: Moving from https://github.com/Automattic/woocommerce-payments/pull/5852/files#r1154097958, at least, we might update this doc block. 

```suggestion
	 * @param WC_Order                    $order The order to update.
	 * @param null|WC_Payments_API_Charge $charge The charge to get the application_fee_amount from.
```

htdat (approved):
Thanks for addressing my review. 🚢 Pre-approve the PR.

> since `$charge_id` is being added in webhook, I thought the `application_fee_amount` which is also charge related, makes sense to be updated there...

I think we should treat webhook as fallback only, and not totally rely on it. Ideally, we would have a post-payment process to handle all of these actions at once rather than scatter in many flows like this. However, we're not being at that state. 

> and it also works for other flows that do not work otherwise. (Split UPE + non-card Payment Methods).

Acknowledge that there is a bug here and that this process prevents reaching to your `attach_transaction_fee_to_order` method in `UPE_Payment_Gateway::process_redirect_payment`. https://github.com/Automattic/woocommerce-payments/issues/5854 - p1679565189430509-slack-CU6SYV31A

PR #5853, opened 2023-03-24
[POC] Use Store API Token for requests to the Store API

asumaran:
Fixes #

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -11,6 +11,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 use Automattic\WooCommerce\StoreApi\StoreApi;
 use Automattic\WooCommerce\StoreApi\RoutesController;
+use Automattic\WooCommerce\StoreApi\Routes\V1\AbstractCartRoute;
+use Automattic\WooCommerce\StoreApi\SchemaController;
+use Automattic\WooCommerce\StoreApi\Schemas\ExtendSchema;
+use Automattic\WooCommerce\StoreApi\Formatters;
+use Automattic\WooCommerce\StoreApi\Formatters\MoneyFormatter;
+use Automattic\WooCommerce\StoreApi\Formatters\HtmlFormatter;
+use Automattic\WooCommerce\StoreApi\Formatters\CurrencyFormatter;
 use WCPay\Core\Mode;
 use WCPay\Core\Server\Request;
 use WCPay\Logger;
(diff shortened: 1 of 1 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5853` or branch name `use-store-api-token` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=use-store-api-token) :rocket:

---

Build info: 

- Latest commit: 7a4ec080017920089f293b110f31c8ff624007b2
- Build time: 2023-03-24 17:07:23 UTC

includes/class-wc-payments.php
```diff
@@ -40,6 +47,18 @@
 use WCPay\Core\WC_Payments_Customer_Service_API;
 use WCPay\Blocks_Data_Extractor;
 
+class WooPay_Store_Api_Token extends AbstractCartRoute {
```
c-shultz, line 50: Presumably, this class is present here because this is a PoC, so it will need to be moved out?

c-shultz: When it does get moved out, please ensure that it's only loaded when absolutely necessary (when store is WooPay eligible and opted-in)

c-shultz (changes requested):

includes/class-wc-payments.php
asumaran: > Presumably, this class is present here because this is a PoC, so it will need to be moved out?

Yes, that's correct. It was placed here just to quickly build the POC.

c-shultz:
I understand from @senadir that stores will need WooCommerce 7.1+ to support this approach, so two things:
- Make sure we are testing against older versions of WooCommerce and handling errors correctly
- Let's just disable WooPay if a merchant doesn't have a minimum version of WooCommerce installed

@pierorocca any thoughts on how we would want to gracefully handle an incompatibility with an older version of WooCommerce? For context, this is one of several approaches to fix WooPay, so we shouldn't over-invest here until we decide on the final approach.


1. My preferred approach would would be to just simply hide WooPay settings altogether for lower versions of WooCommerce. We do have WooCommerce 7.1+ listed as the minimum version for WooPay compatibility already.

1. Alternatively, we could uncheck/disable the opt-in checkbox for WooPay and display a simple note that WooCommerce 7.1+ being required for WooPay. Given that this is targeted as patch fix.

Given that this is targeting a patch fix, I might suggest we do number 1 now, and consider if number 2 is needed in a future release.

c-shultz:
Speaking of patch fixes, @asumaran, this should be based off of `trunk` instead of `develop` so it's ready for a patch.

includes/class-wc-payments.php
rafaelzaleski: One suggestion I have is to consider moving the platform checkout code out of `class-wc-payments.php` (we have many functions for platform checkout functionality inside this class). This could help to reduce the file size, make it easier to maintain, and improve code readability. 
Overall, I think this change would help to organize our codebase better and make it more manageable in the long term.

pierorocca:
> My preferred approach would would be to just simply hide WooPay settings altogether for lower versions of WooCommerce. We do have WooCommerce 7.1+ listed as the minimum version for WooPay compatibility already

Apologies as I'm not seeing my GH notifications. @c-shultz option 1 makes it clean while option 2 creates a visual incentive for a merchant to upgrade. Over time however, I think the need for Option 2 is low to nil so it would be a short lived benefit.

PR #5864, opened 2023-03-27
Fix overview screen type error when the account has no deposit scheduled

### LINKED ISSUE #5863: Simplify deposits UI – type error when the account has no deposit scheduled
For new WCPay accounts with no scheduled deposit, the following error will appear when viewing the [`DepositsOverview` card](https://github.com/Automattic/woocommerce-payments/blob/8d62900a1137f065a562a38f54c72d361c1b7f3a/client/components/deposits-overview/index.tsx#L15) on the Payments → Overview screen.

This is only visible if the `_wcpay_feature_simplify_deposits_ui` feature flag is enabled.

```
Uncaught TypeError: can't access property "date", nextScheduled is undefined
```

This error is being thrown because the `overview` object has no `nextScheduled` property:

https://github.com/Automattic/woocommerce-payments/blob/8d62900a1137f065a562a38f54c72d361c1b7f3a/client/components/deposits-overview/utils.ts#L21-L30

### Steps to reproduce

1. Onboard to WCPay with a new account.
2. Enable the `_wcpay_feature_simplify_deposits_ui` feature flag: Run `update_option( '_wcpay_feature_simplify_deposits_ui', '1' );` in [WP Console](https://wordpress.org/plugins/wp-console/).
3. View the Payments → Overview screen. See the error.

### Notes
The `Overview` type should declare `nextScheduled` as an optional property.

https://github.com/Automattic/woocommerce-payments/blob/8d62900a1137f065a562a38f54c72d361c1b7f3a/client/types/account-overview.d.ts#L43-L50

Jinksi:
Fixes #5863

This PR fixes the error `TypeError` that occurs on the Payments → Overview screen when an account has no scheduled upcoming deposit.

The `nextScheduled`, `lastPaid` and `instant` properties of the type `AccountOverview.Overview`  have been changed to optional, since they can be returned as `undefined` in some cases.

https://github.com/Automattic/woocommerce-payments/blob/279046a35872335d77bea203c9b5b3085aa198a5/client/types/account-overview.d.ts#L43-L50

I've also made the `overview` prop of the `getNextDeposit()` function optional, which can be `undefined` if `useAllDepositsOverviews` has not resolved the data yet. This was throwing another error once the error above was fixed.

https://github.com/Automattic/woocommerce-payments/blob/279046a35872335d77bea203c9b5b3085aa198a5/client/components/deposits-overview/utils.ts#L21-L23

#### Testing instructions

1. Onboard to WCPay with a new account.
2. Enable the `_wcpay_feature_simplify_deposits_ui` feature flag: Run `update_option( '_wcpay_feature_simplify_deposits_ui', '1' );` in [WP Console](https://wordpress.org/plugins/wp-console/).
3. View the Payments → Overview screen.
4. Ensure no error is thrown in the console.
5. Ensure that the currency shown in the estimated deposits table matches the selected/default currency.

image

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5864` or branch name `fix/5863-account-overview-nextScheduled-typeerror` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/5863-account-overview-nextScheduled-typeerror) :rocket:

---

Build info: 

- Latest commit: aaa7a2b1716e4d94f0847b5a1b988c29ffa4e43b
- Build time: 2023-03-30 20:19:56 UTC

shendy-a8c (commented):
Tested with `_wcpay_feature_simplify_deposits_ui` on and off. I do see the TypeError problem on base branch and I don't see it in this PR branch, so definitely this PR fixes the issue.

I have some questions though.

client/types/account-overview.d.ts
```diff
@@ -42,11 +42,11 @@ declare namespace AccountOverview {
 
 	interface Overview {
 		currency: string;
-		lastPaid: Deposit;
-		nextScheduled: Deposit;
+		lastPaid?: Deposit;
```
shendy-a8c: Just curious, what's the difference between adding the `?` to make that property optional and adding ` | undefined`?

```suggestion
		lastPaid: Deposit | undefined;
```

Jinksi: `lastPaid: Deposit | undefined;` would mean the object needs the `lastPaid` property to be present, but it can have an `undefined` value.

If we're confident that the `AccountOverview.Overview` response will _always_ have a `lastPaid` property in the response it would be more accurate to use `lastPaid: Deposit | undefined;`.

client/types/account-overview.d.ts
```diff
@@ -42,11 +42,11 @@ declare namespace AccountOverview {
...
+		nextScheduled?: Deposit;
 		pending: Balance;
```
shendy-a8c: I still don't quite understand fully why you make lastPaid, nextScheduled, and instant optional and why only those 3? Why didn't you make `pending` and `available` also optional?

Looking at https://github.com/Automattic/woocommerce-payments/blob/fix/5863-account-overview-nextScheduled-typeerror/client/data/deposits/selectors.js#L86-L93
potentially all of them can be optional.

However, the next line https://github.com/Automattic/woocommerce-payments/blob/fix/5863-account-overview-nextScheduled-typeerror/client/data/deposits/selectors.js#L97
is setting each key: `lastPaid`, `nextScheduled`, `pending`, `available`, and `instant`.

Also, looking at server code, the value of `deposit.next_scheduled` will always be defined. At most, it'll be empty array, so I don't understand why we have to handle when it's undefined.

shendy-a8c: I am just rubber ducking. Now I can see why. It's because when it's an empty array, it won't this [forEach](https://github.com/Automattic/woocommerce-payments/blob/fix/5863-account-overview-nextScheduled-typeerror/client/data/deposits/selectors.js#L82) won't even run even once, right?

client/types/account-overview.d.ts
shendy-a8c: Thank you for the explanation.

client/types/account-overview.d.ts
shendy-a8c: I also can now see why `pending` and `available` are always defined.

shendy-a8c (approved):
Looks good!

client/types/account-overview.d.ts
Jinksi: As you mentioned, this is where the `undefined` value originates from: `client/data/deposits/selectors.js` `getAllDepositsOverviews()`:

https://github.com/Automattic/woocommerce-payments/blob/b644b98c5c3a83c0969749762d64f7d25c18bda3/client/data/deposits/selectors.js#L86-L93

So I think you're correct that we can safely use `lastPaid: Deposit | undefined;`.

We can also update `pending`, `available` and `instant`.

> Also, looking at server code, the value of deposit.next_scheduled will always be defined.

This `AccountOverview.Overview` type is being used by `AccountOverview.OverviewsResponse.overview.currencies`:

https://github.com/Automattic/woocommerce-payments/blob/b644b98c5c3a83c0969749762d64f7d25c18bda3/client/types/account-overview.d.ts#L52-L58

So the response from the server is being restructured in the `getAllDepositsOverviews()` code above, and the `undefined` value for these properties is being introduced there.

Logging the value of `currencies` from [this line](https://github.com/Automattic/woocommerce-payments/blob/b644b98c5c3a83c0969749762d64f7d25c18bda3/client/data/deposits/selectors.js#L101) confirms that `nextScheduled` can be `undefined` but is present on the currency object:

image

I'll update the `AccountOverview.Overview` type to ensure all properties can be `undefined`. ~Since this relates only to the `currencies` array, perhaps we could extend `AccountOverview.Overview` instead?~ This type is only used by `AccountOverview.OverviewsResponse.overviews.currencies` AFAICT.

client/types/account-overview.d.ts
Jinksi: Replied in https://github.com/Automattic/woocommerce-payments/pull/5864#discussion_r1152627256

client/types/account-overview.d.ts
Jinksi: I've updated this in cebf521.

@shendy what do you think of this solution?

shendy-a8c: Looks good @Jinksi.

Issue #5808, opened 2023-03-17
Users encountering an error while attempting to pay for order using UPE method


timur27:
### Describe the bug
When trying to Pay for order with any of the available UPE methods, user sees an error _**Invalid payment method. Please input a new card number.**_

### To Reproduce

Enable split UPE and:
1. Buy a product using the shortcode checkout and UPE method (e.g. Sofort) 
2. On the third-party test payment page, choose _**Fail test payment**_
3. After redirected back to the checkout page, go to _**My account -> Orders**_, find the failed order and click _**Pay**_
4. On the _**Pay for order**_ page, choose any of the available UPE methods and click _**Pay for order**_

### Actual behavior
Users are seeing _**Invalid payment method. Please input a new card number.**_ error notice. 

### Screenshots

![image.png](https://images.zenhubusercontent.com/59de7a6cb0222d5de47b4727/62610a52-2727-4392-a270-c89f379009c7)

### Expected behavior

Users shouldn be able to pay for order successfully, using any of the available payment method on the list.

### Desktop (please complete the following information):

* OS: MacOS
* Browser: Chrome

timur27:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5808?planning-poker) with Zenhub @FangedParakeet @mdmoore

FangedParakeet:
I now have a PR ready that will (eventually) resolve this issue, but I am going to mark this issue temporarily blocked for now, since I have also learned that this functionality is also blocked by #5854 and we will not be able to return this functionality, until a hero appears to save us from despair by resolving that particular issue.

> I have identified three issues that are causing these frustrating failures.
>
>1. If order currency and customer selected currency do not match, if the customer selected currency is not supported by available payment method, payment intent will produce an error and payment elements will fail to mount.
>2. Pay for order form incorrectly detects when WCPay payment method is selected at checkout.
>3. Payment intent confirms successfully, but process_redirect_payment function exits early, order is not updated, and instead marked as a failure.

Please see #6177 for more specific details on these maladies and their various cures.

Issue #5025, opened 2022-10-31
Enhanced UPE - WCPay Compatibility: 3DS Authentication


FangedParakeet:
### Description
Currently, 3DS authentication works for happy path flows only after https://github.com/Automattic/woocommerce-payments/pull/5869. This issue's ultimate goal is to ensure that all the edge cases are also covered and fully support the 3DS authentication. 

### Acceptance criteria
* 3DS cards saving for the physical products (by clicking the save checkbox) works as expected
* 3DS cards saving for the subscriptions (without clicking the save checkbox) works as expected
* Subscription renewal using WooCommerce Subscriptions plugin for the subscriptions bought using a 3DS card works as expected

The first two ACs above should be met on the shortcode checkout, blocks checkout, _Pay for order_ page, _Change payment for subscriptions_ page and _Add payment method_ page.

### Additional context
[Stripe 3DS test cards](https://stripe.com/docs/testing#regulatory-cards)

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5025?planning-poker) with Zenhub @mdmoore @timur27

Issue #5023, opened 2022-10-31
Enhance My Account section to render and fully function payment gateways on subpages


FangedParakeet:
### Description
This issue ensures that payment gateways are being rendered and fully functional in different smaller pages within the _My account_ on the merchant site. Pages to cover: 

1. _Orders -> Pay_ (for order)
2. _Subscriptions_ -> _Change payment_
3. _Payment methods_ -> _Add payment method_

For the first two pages, the ultimate goal of this issue is to ensure that users can both add a new payment method or choose from the saved payment methods list. For the third page, the goal is to ensure that users can add new payment methods.

### Acceptance criteria
* Should be able to successfully complete order and transaction from Pay for Order page by using the saved payment method or a new one.
* Should be able to successfully add a new payment method 
* Should be able to successfully change the payment method for the existing subscription by choosing the saved payment method or adding a new one

### Engineering notes
[The MVC of the deferred intent UPE](https://github.com/Automattic/woocommerce-payments/pull/5869) was implemented for the shortcode in `stripe-checkout.js`. When working on this issue, creating separate scripts per subpage is highly recommended to avoid one script being responsible for multiple things and thus increase code readability. If separate scripts like `pay-for-order-handler.js`/`add-payment-method.js` or similar will be created, the `stripe-checkout.js` should be better renamed to `shortcode-checkout.js`. 

Another hint would be to look for possible common areas between `stripe-checkout.js` and newly created scripts for _My account_ subpages. If common areas will be found - try to merge them into one place.

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5023?planning-poker) with Zenhub @mdmoore @timur27

timur27:
I gave it a `3` because of the amount of screens that need to be addressed within this issue. Also, compared to [this issue](https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/6227) which is estimated for 2SP, _**My account**_ section related issue seems to be more complex. 


@FangedParakeet could you share your opinion on why you gave 2SP so that we could align and maybe you can convince us. Thank you!

FangedParakeet:
>1. Orders -> Pay (for order)

This will be completed in #5808, so can be ignored here.

 >2. Subscriptions -> Change payment
[...]

Issue #5020, opened 2022-10-31
Enhanced UPE - Blocks integration


FangedParakeet:
### Description

Stripe will soon release a new refactor of their UPE offering new payment integration flows. The most crucial addition is that they will now offer server side payment intent confirmation for the UPE, meaning that the payment intent is no longer required prior to presenting the UI element and the intent can be created and confirmed after collecting the payment information into a payment method.

This necessitates a refactor for both shortcode and blocks checkouts, where both create a payment intent as the first order of business, before loading any UI elements. This new refactored blocks implementation may now have significant resemblance to the previous legacy card element implementation. This issue is to complete this blocks integration. 

This issue is currently blocked until Stripe accepts us into their beta release program for their new enhanced UPE tool. This is also likely blocked by #5019. Although it would not be impossible for these two pieces to be completed in parallel, there will likely be significant overlap on the back-end portions of both implementations, so it would be preferable to complete the PoC prior to this effort. 

### Acceptance criteria

* Able to successfully complete checkout with new payment method via Blocks checkout.

### Additional context

[Stripe UPE – Options to solve WooPay compatibility and WC Pay loss of control over the payments UI/UX](pc2DNy-2e2-p2)
#4603

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5020?planning-poker) with Zenhub @mdmoore @timur27

timur27:
I think we need to update the description and the ACs.

From a technical standpoint, the success of this issue would be the working Blocks checkout page. However, we need to consider which payment methods should be working after this issue is completed. In addition to card and non-card gateways, there is also WooPay and Stripe Link. Should we include them both in the scope? They also need to be implemented for the shortcode, so a separate issue to cover Stripe Link & WooPay for both shortcode and Blocks seems to be a good option to consider as well.

Furthermore, any reusable components from issue #5869 should be re-used in this issue, and we can emphasize this in the ACs as well.

PR #5869, opened 2023-03-27
Deferred intent creation for split UPE gateways

### LINKED ISSUE #5770: New UPE - Initial minimum viable product

timur27:
Closes https://github.com/Automattic/woocommerce-payments/issues/5770

#### What's the purpose of this PR
This PR is the first step in the UPE project for deferred payment intent creation. Previously, creating a payment intent was required for the Stripe payment UI element, resulting in excess intents and an increased risk of carding attacks. This PR implements the new approach by creating and confirming payment intents on the server, improving conversion metrics, security and, eventually, allowing to support WooPay and Stripe Link simultaneously. [The Stripe documentation](https://stripe.com/docs/payments/finalize-payments-on-the-server) for the deferred intent creation is a good place to understand the approach.

Assuming that the deferred intent UPE feature is enabled via [the feature flag](https://github.com/Automattic/woocommerce-payments/pull/5811) in WCPay Dev Settings, the following outcomes will be observed **on the shortcode checkout page**:

* Stripe Payment Elements for particular payment methods appear on the page successfully, while there is no intent created and the _Payments_ section in the Stripe Dashboard doesn't have any new entries
* A payment intent is created on the server after users click _Place order_ button
* The intent is created only for the payment method that is used while placing an order
##### Front end
`upe-deferred-intent-creation/event-handlers.js` script is the entry point of the implementation, as it's rolled out by the `class-upe-split-payment-gateway.php`. As the name suggests, it handles the main events that eventually lead to rendering elements or handling the checkout. It follows the way events are handled for the split UPE in `classic/upe-split.js`. 

`upe-deferred-intent-creation/stripe-checkout.js` handles the Stripe Payment Element creation and mounting, as well as checkout handling. This script also provides helper functions to achieve the ultimate goal of rendering and using the Stripe payment elements. While reviewing, it's worth starting with the `mountStripePaymentElement` and `checkout` functions. A lot of implementation here comes from `classic/upe-split.js` and `classic/index.js` scripts.

`upe-deferred-intent-creation/3ds-flow-handling.js` handles the authentication for 3DS cards and hopefully can be re-used while working on other screens - _Blocks checkout_ and _My account_ pages. The implementation is mostly taken from the `maybeShowAuthenticationModal()` function which is introduced in every UPE strategy.

Compared to the previous implementations, the client code was split between multiple scripts, which should help while working further on the project. It also allows decoupling and removing jQuery wherever possible, making the testing a painless process with the current WCPay environment configuration. Different features are possible placed in different scripts, which increases code readability, is easier to maintain and extend.

##### Back end
The legacy `class-wc-payment-gateway-wcpay.php` is re-used heavily, and the only thing that was added there was the additional `return_url` parameter for non-card elements so that it can handle the third-party authentication flow. 

1. Update the WCPay Dev Tools plugin in your local environment to be based on [this PR](https://github.com/Automattic/woocommerce-payments-dev-tools/pull/86) or newer. 
2. In WCPay Dev Tools, uncheck the **_Enable UPE checkout (legacy) and Enable Split UPE checkout_** feature flags. At the same time, check the _Enable Split UPE checkout with deferred intent creation_ flag.
3. Set the currency to EUR on the merchant store
4. Add a simple product to the cart, checkout using the shortcode and card, and confirm that only one payment intent is created and processed successfully in the Stripe Dashboard. The intent should be created _**after**_ the _Place order_ button is clicked, opposite to the split UPE implementation where the intent is created before Stripe Payment Element rendering.
5. Repeat the above step. Instead of the card, pay with any other payment gateway (e.g., `giropay`, `eps` etc.) and confirm a single intent approach works as expected when a payment is accepted on the test third-party authentication page.
6. Repeat the above step but reject the payment on the test third-party authentication page and confirm that you are redirected to the checkout page and that the payment intent is not confirmed on the Stripe side. The order status should be `Failed`.
7. Checkout with 3DS (`4000002760003184`) and 3DSv2 (`4000000000003220`) card. Confirm that the bank authentication pop-up modal appears on the screen.
8. Test the transaction failure for any of the above 3DS cards - confirm that the error notice is successfully shown to the user on the checkout page after the transaction failure, and confirm that the Order status is `Failed`. 
9. Repeat the attempt but authorize the transaction this time. Confirm that the order status has changed from `Failed` to `Processing` in the `WooCommerce -> Orders`. Confirm that the order status/history are in line with the payment status/history in Stripe Dashboard.
10. Save the default card and confirm that it's saved successfully. After clicking the save checkbox, confirm that the Terms are appearing within the Payment Element.
11. Purchase a subscription and pay with a default card. After purchase, ensure the subscription was purchased successfully, and the card is saved.
12. Using [the WooCommerce Subscriptions plugin](https://github.com/woocommerce/woocommerce-subscriptions/releases), renew the previously bought subscription. Confirm that the status of the subscription is active and the Stripe dashboard has the payment entry for the renewal, which is not in status `Succeeded`

#### Testing the existing UPE implementations
For the legacy UPE and the split UPE, a few things in the code have been modified to allow the deferred payment intent creation to re-use some duplicated code. This is why I'd like to ask you to test the critical flows. Good examples are: 
* Paying with a card when non-UPE is enabled
* Card payment and non-card payment with the legacy UPE enabled
* Card payment and non-card payment with the split UPE enabled

[Results of the first test round](https://docs.google.com/spreadsheets/d/1DIxtcrEiknLHGmn5HiYGONNo_bwhy1LAxmhuWGdMD6k/edit#gid=0)
[Follow-up issues as part of the project](https://github.com/Automattic/woocommerce-payments/pull/5869#issuecomment-1521522039)

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5869` or branch name `poc/deferred-upe-intent-creation` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=poc/deferred-upe-intent-creation) :rocket:

---

Build info: 

- Latest commit: 0ab4dbcb53ef63d93ff383c71305021713769bca
- Build time: 2023-05-05 16:24:34 UTC

timur27:
@FangedParakeet @mdmoore Given it's a POC that doesn't have everything covered and probably isn't perfect in terms of code structure, you might find something you wish could be improved, both technically and functionally. I'd appreciate it if you could leave all your findings as part of the review. I will gather all of them and ensure we have an issue in our backlog that brings shortcode checkout with deferred intent creation approach to the perfect state.

changelog/add-5682-add-upe-deferred-intent-feature-flag
FangedParakeet, line 1: [Don't think you need this one anymore](https://github.com/Automattic/woocommerce-payments/blob/develop/changelog/add-5682-add-upe-deferred-intent-feature-flag).

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+	 *
+	 * @param {string} paymentMethodType The type of payment method, e.g. 'card', 'giropay', etc.
+	 * @param {string} domElement The selector of the DOM element of particular payment method to mount the UPE element to.
+	 */
+	async function createAndMountPaymentElement(
+		paymentMethodType,
+		domElement
+	) {
+		const options = {
+			mode: 'payment',
```
FangedParakeet: Just as a caveat, I'm not sure if we use Stripe subscriptions, but if the `amount` value is 0, I think we would want to use [`setup` as mode value](https://stripe.com/docs/js/elements_object/create_without_intent#stripe_elements_no_intent-options-mode) here.

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+				),
+			},
+		};
+
+		if ( getUPEConfig( 'cartContainsSubscription' ) ) {
+			upeSettings.terms = getTerms( paymentMethods, 'always' );
+		}
+
+		return upeSettings;
+	}
```
FangedParakeet: Is this basically borrowed from [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L318-L326)? I concede that things like adding payment methods or the order pay page are outside the scope of this PR, but at some point it seems like it would probably make sense to move this to a shared utility function.

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+
+	const handleCheckout = async ( $form, paymentMethodType ) => {
+		blockUI( $form );
+		const elements = gatewayUPEComponents[ paymentMethodType ].elements;
+		elements.submit().then( ( result ) => {
+			if ( result.error ) {
+				showErrorCheckout( result.error.message );
+			}
+		} );
+		const pm = await api.getStripe().createPaymentMethod( {
```
FangedParakeet:
```suggestion
		const paymentMethod = await api.getStripe().createPaymentMethod( {
```
This is pretty pedantic and isn't really something that I'm particularly bothered about, but I think [using full-names for variables is more accepted form](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/#naming-conventions).

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+			const fields = {
+				...$form.serializeArray().reduce( ( obj, field ) => {
+					obj[ field.name ] = field.value;
+					return obj;
+				}, {} ),
+				wc_payment_method: pm.paymentMethod.id,
+				'wcpay-fingerprint': fingerprint ? fingerprint : '',
+			};
+
+			const response = await api.processCheckout( fields );
```
FangedParakeet: The `processCheckout` function was added explicitly for the UPE, because in the previous UPE checkout, since the intent could only be confirmed on the client-side and we had to override WC from redirecting the user on form submit, to allow Stripe to do so instead, we had to manually complete the WC checkout via AJAX to create the order, before confirming the intent and allowing Stripe to perform the redirect.

In this flow, we no longer need to confirm the intent through the payment element and Stripe no longer performs the URL redirect (AFAIA--not sure if this presumption only applies to the card payment method). Have you considered simply submitting the form and allowing the regular WC checkout flow to proceed. I think if you just did `$form.submit()` at this point, it would basically achieve the same end-result as you've managed here without needing to serialize and restructure the form fields, as you're doing above, or manually redirecting the user using `window.location.href`.

In the legacy card, we create the payment method on the client, as you've done here, [include a hidden field on the page](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-checkout.php#L200) to store the payment method ID, and then simply [populate this field and submit the checkout form regularly](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/index.js#L311-L314).

I'm honestly not sure which path is more performant or preferable, but am just curious if you were already aware of this and had made a conscious decision to complete the checkout in this way. Let me know!

client/checkout/classic/upe-split.js
```diff
@@ -321,7 +288,9 @@ jQuery( function ( $ ) {
 		}
 		if ( isCheckout && ! ( isOrderPay || isChangingPayment ) ) {
 			upeSettings.fields = {
-				billingDetails: hiddenBillingFields,
+				billingDetails: getHiddenBillingFields(
+					getUPEConfig( 'enabledBillingFields' )
+				),
```
FangedParakeet: Nice, liking this simplification. 👍

includes/class-payment-information.php
```diff
@@ -249,7 +249,7 @@ public static function from_payment_request(
 	 * @return string
 	 */
 	public static function get_payment_method_from_request( array $request ): string {
-		foreach ( [ 'wcpay-payment-method', 'wcpay-payment-method-sepa' ] as $key ) {
+		foreach ( [ 'wcpay-payment-method', 'wcpay-payment-method-sepa', 'wc_payment_method' ] as $key ) {
```
FangedParakeet: FYI, this `wcpay-payment-method` key is hyphenated instead of being separated by underscores like other PHP variables, because it's manually inserted into the DOM over [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-checkout.php#L200) by the `WC_Payments_Checkout` class. I left another comment about this, but was just wondering if you were of this. If we did the same thing, presumably we would not need to add another uncannily similar key over here.

includes/class-wc-payments-features.php
FangedParakeet, line 1: I _think_ resolving merge conflicts will probably remove this from the diff.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -115,6 +115,10 @@ public function get_payment_fields_js_config() {
 			$payment_fields['checkoutTitle']        = $this->gateway->get_checkout_title();
 			$payment_fields['upePaymentIntentData'] = $this->gateway->get_payment_intent_data_from_session();
 			$payment_fields['upeSetupIntentData']   = $this->gateway->get_setup_intent_data_from_session();
+		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+			$payment_fields['checkoutTitle'] = $this->gateway->get_checkout_title();
+			$payment_fields['currency']      = get_woocommerce_currency();
+			$payment_fields['cartTotal']     = number_format( WC()->cart->total, 2, '', '' );
```
```suggestion
			$payment_fields['cartTotal']     = WC_Payments_Utils::prepare_amount( WC()->cart->get_total( '' ), get_woocommerce_currency() );
```
Any reason you elected not to use this utility function, like we do over [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L330), for example?

includes/class-wc-payments.php
```diff
@@ -461,6 +463,20 @@ public static function init() {
...
+			$payment_methods = [];
+			foreach ( $payment_method_classes as $payment_method_class ) {
+				$payment_method                               = new $payment_method_class( self::$token_service );
+				$payment_methods[ $payment_method->get_id() ] = $payment_method;
+			}
+			foreach ( $payment_methods as $payment_method ) {
+				self::$upe_payment_method_map[ $payment_method->get_id() ]  = $payment_method;
+				self::$upe_payment_gateway_map[ $payment_method->get_id() ] = new New_UPE_Payment_Gateway( self::$api_client, self::$account, self::$customer_service, self::$token_service, self::$action_scheduler_service, $payment_method, $payment_methods, self::$failed_transaction_rate_limiter, self::$order_service );
+			}
+
```
FangedParakeet: There is so much repeated code here and I have to really squint to differentiate what's going on here now. 😂 This isn't really your fault at all and I totally understand what you've had to do here, but I really think we need to refactor the logic here--perhaps in a separate issue--so that this becomes a bit more maintainable. Lord knows I've contributed to the mess here, so my hands are bloodied too, but I don't think we can leave this like this forever lol. Let me know if you happen to feel the same way too!

includes/payment-methods/class-new-upe-payment-gateway.php
```diff
@@ -0,0 +1,80 @@
...
+			[],
+			'3.0',
+			true
+		);
+
+		$script_dependencies = [ 'stripe', 'wc-checkout', 'wp-i18n' ];
+		if ( $this->supports( 'tokenization' ) ) {
+			$script_dependencies[] = 'woocommerce-tokenization-form';
+		}
+		WC_Payments::register_script_with_dependencies( 'wcpay-upe-checkout', 'dist/new_upe_checkout', $script_dependencies );
```
FangedParakeet: Is this line the only difference between this class and the `UPE_Split_Payment_Gateway` class it extends? Could this just an `if` branch inside the `UPE_Split_Payment_Gateway::register_scripts` function? If so, might that make the logic in `WC_Payments` a little simpler? 😁

FangedParakeet (commented):
I need to test this a little more--having some frustrating issues with my local environment, which are confusing my outcomes--but...this implementation looks pretty good to me! Nice one. It's very reassuring and relieving that we're able to use so much of the legacy payments code: finally the UPE has come full circle.

I've got a handful of comments here that are mainly either about style or pedantic quibbles, so hopefully there's enough here for you to mull over, while I test this more comprehensively. I'm eager to learn your inspirations for a few of your integration choices, so I'll await your responses with bated anticipation in the meantime.

client/checkout/classic/new-upe.js
timur27: Nice catch - the checkout page failed to load for trial subscriptions since creating elements instance for zero-amount in `payment` mode was declined by Stripe. https://github.com/Automattic/woocommerce-payments/pull/5869/commits/e58a0bbc6a83bd886be40a6a26e829cb0729affe fixes this.

client/checkout/classic/new-upe.js
timur27: I merged the duplicated logic in https://github.com/Automattic/woocommerce-payments/pull/5869/commits/749fb72d83db5040c4c6ed2cc14fc8ef628b0aa8. 

Although it's a good step right now, I thought that we need to keep in mind, that at some point, `upe-split.js`/`upe.js` will be removed and this logic will probably be moved back to the only one UPE script. This means we might prefer to avoid the roundtrip (move the function to the utility script and then back to the upe script) at some point to save time. But again, this particular case is absolutely reasonable.

client/checkout/classic/new-upe.js
timur27: Pedantic but crucial - thanks for it! As part of https://github.com/Automattic/woocommerce-payments/pull/5869/commits/1d22a4c02b72815410512861eb7db60f7eda279e, we now use the fluent interface and the name has disappeared.

client/checkout/classic/new-upe.js
timur27: I knew about form submission in the non-UPE implementation. Still, I went with the AJAX call first because I was eager to build the POC quicker since applying the AJAX-call-approach was just a matter of copying and applying it to the new implementation. 

Your arguments to try the form submission instead are absolutely justified - we could achieve less complexity and replace the unnecessary AJAX call with the clearer form submission. 

I've compared the AJAX call vs. form submission regarding the execution pathways, and they perform the same steps. With form submission, [we are redirected to the `checkout` thanks to the `action` parameter here](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/templates/checkout/form-checkout.php#L32) which [eventually leads us to the checkout event execution](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-ajax.php#L479-L483) and finally, [the payment processing](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-checkout.php#L1050). The same path is observed via the AJAX call approach, obviously, since it's getting handled by [this implementation](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-ajax.php#L479-L483) as well.

I've tested the form submission approach and eventually applied it https://github.com/Automattic/woocommerce-payments/pull/5869/commits/1d22a4c02b72815410512861eb7db60f7eda279e. Due to this change, [I've also rolled back the `processCheckout` API definition](https://github.com/Automattic/woocommerce-payments/pull/5869/commits/48832a45e6f0b8049f23a2439fd72637f2477568) to reduce the probability of breaking the existing implementation without reason. For me, it's now looking way more elegant. I'm looking forward to hearing your opinion on this. We can definitely re-iterate here, add something else, or even roll back to the previous approach easily.

includes/class-payment-information.php
timur27: I changed the approach in https://github.com/Automattic/woocommerce-payments/pull/5869/commits/85cace1efdad218d989f8583edd2ab00c1ca9107, and we now re-use the `wcpay-payment-method` because as you mentioned, there is no reason to create an additional key here.

> 💡 Note: As you probably noticed in the above-linked commit, I attach the element on the front-end, instead of [in these areas](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L304). The reason for that is, that the `payment_fields` method on the back-end, the same way as `display_gateway_html()` method is being run multiple times, depending on the amount of the enabled UPE methods. I want to avoid attaching multiple hidden fields to insert the payment method id.

includes/class-wc-payments-upe-checkout.php
timur27: The `prepare_amount` utility function in `WC_Payments_Utils` was overlooked. Changes applied https://github.com/Automattic/woocommerce-payments/pull/5869/commits/9b34bd2a2f9f52b07c1d039240859a9f8623bc48. Thanks!

includes/class-wc-payments.php
timur27: I can't agree more! I remember us chatting about the UPE factory. Because of [this comment](https://github.com/Automattic/woocommerce-payments/pull/5869#discussion_r1155030491), all the changes in `class-wc-payments.php` were removed for now. Still, we need to think about the refactor here, but also keep in mind that our efforts should eventually pay off, given that there will be only one UPE implementation in the end. I'll make sure to come back to this topic with the whole team in a few days and for now, we can resolve this thread, if that sounds reasonable for you.

includes/payment-methods/class-new-upe-payment-gateway.php
timur27: It makes sense! Please see https://github.com/Automattic/woocommerce-payments/pull/5869/commits/85f568d14b044669e47ba39fa5c047eccb403ca9.

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+	 */
+	async function createAndMountPaymentElement(
+		paymentMethodType,
+		domElement
+	) {
+		const amount = Number( getUPEConfig( 'cartTotal' ) );
+		const options = {
+			mode: 1 > amount ? 'setup' : 'payment',
+			currency: getUPEConfig( 'currency' ).toLowerCase(),
+			amount: Number( getUPEConfig( 'cartTotal' ) ),
```
FangedParakeet:
```suggestion
			amount: amount,
```
Could just reuse the var you defined above, no?

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+	 * No Stripe PaymentIntent is created at this point.
+	 *
+	 * @param {string} paymentMethodType The type of payment method, e.g. 'card', 'giropay', etc.
+	 * @param {string} domElement The selector of the DOM element of particular payment method to mount the UPE element to.
+	 */
+	async function createAndMountPaymentElement(
+		paymentMethodType,
+		domElement
+	) {
+		const amount = Number( getUPEConfig( 'cartTotal' ) );
```
FangedParakeet: I believe there are third-party plugins that can add/remove items from the cart at checkout. This could create a scenario where the amount changes at checkout somehow and we need to switch from a payment intent to a setup intent on remounting the element upon updated checkout. I'm pretty sure the legacy UPE is able to handle this somehow. I'm not expecting all these edge-cases to be supported here, but just leaving a note that we may want to double-check this at some later time--if you haven't tried it out already!

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+
+		upeElement.mount( domElement );
+		gatewayUPEComponents[ paymentMethodType ].upeElement = upeElement;
+	}
+
+	let paymentMethodGenerated;
+
+	const handleCheckout = ( $form, paymentMethodType ) => {
+		if ( paymentMethodGenerated ) {
+			paymentMethodGenerated = null;
```
FangedParakeet:
```suggestion
			paymentMethodGenerated = false;
```
(I know--this is borrowed from the legacy card implementation, but...) AFAICT, the only values possible for `paymentMethodGenerated` are `true` or `null`. Why not `true` or `false` instead?

client/checkout/classic/new-upe.js
```diff
@@ -0,0 +1,242 @@
...
+		.map( ( method ) => `checkout_place_order_${ method }` )
+		.join( ' ' );
+	$( 'form.checkout' ).on( checkoutEvents, function () {
+		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
+		return handleCheckout( $( this ), paymentMethodType );
+	} );
+
+	function appendPaymentMethodIdToForm( $form, paymentMethodId ) {
+		// Remove any existing payment method inputs.
+		$form.find( 'input[name="wcpay-payment-method"]' ).remove();
```
FangedParakeet: Can there be multiple of these inputs on the page? Shouldn't this only get added once for the card payment method? Card payment method should be ever present, I believe, so I think we can rely on this input being present at least once.

FangedParakeet (commented):
Got a few more teeny comments and clarifications for you to munch on with your cornflakes, but thanks for responding to my first fistful of feedback--this is looking really nice, clean, and simple now! 🙌

client/checkout/classic/new-upe.js
timur27: Sure, fixed in 40031cf9a4c6c420589b640bde7eba2745070546

client/checkout/classic/new-upe.js
timur27: This is a very good observation. There are multiple layers of this. 

#### The amount changes, but the intent type remains to be of type `payment`.
> TL;DR this is covered everywhere.

In this case, legacy UPE/split UPE are handling it but on the back-end, meaning that [the payment intent gets updated](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L239-L256) with the new amount. The Stripe Card Element handles this as well in the right way because it fetches the amount only after the checkout button is clicked, and the amount is proper at that stage. Our new implementation with deferred intent creation handles it because the amount is updated after clicking the _Place order_ button right before the intent is created and confirmed. 

#### The amount changes, implying the intent type change `setup` <-> `payment`.
This is an interesting case, and I think you are mentioning exactly this. I found out that the intent type is not changed in any of the implementations. For the legacy card, things run smoothly since [we don't need to provide the intent type when rendering the UI element](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/index.js#L54-L57). For the legacy UPE and split UPE, the expected behavior is that the available payment methods are updated. If there are eight different gateways for a physical product and it gets removed in the meantime + a subscription with a free trial is added to the cart, and all the gateways except the CC should disappear, which doesn't occur. Eventually, users will see

image
for the split UPE and for the current deferred intent flow implementation

image
for the legacy UPE.


The legacy UPE + split UPE should be fixed within a separate issue. I'll try to make things work for the deferred intent flow. Per my research, there is a `updated_cart_totals` event that is triggered, and we can then [update the element dynamically](https://stripe.com/docs/payments/finalize-payments-on-the-server?platform=web&type=payment#dynamic-updates) which hopefully allows us to successfully change the flow. I've tested it quickly, and it seems that the `updated_cart_totals` isn't fired for the async cart update, but I'll check this more deeply.

client/checkout/classic/new-upe.js
timur27: Agree with you, updated in 48609a2b00e9a82a175018f2b16c0be667aa7a28. Thanks!

client/checkout/classic/new-upe.js
timur27: With the current implementation, this input is added only once for the payment gateway used for the purchase without the risk of being added more than once due to the point of time it gets added (after checkout processing starts).

This removal is rather a countermeasure from the weird cases when the input is already on the page. However, I can't think of any real use case when this could happen. I've removed this line in 02f4657e549bd178ffeb882a7a6ba09b3303a0c4. Please let me know if I somehow misunderstood your question or the purpose of this thread as such.

client/checkout/classic/new-upe.js
timur27: @FangedParakeet There are a few things to consider, and I'd like to hear your thoughts on what and how we would like to prioritize here.

#### No third-party plugins
Without third-party plugins, this can break as described above for any payment gateways (legacy UPE/split UPE etc.). And to make this happen, one needs to update the cart in a separate browser tab and then place an order on the initial checkout page in another tab. Given the nature of PHP and how it's rolled out, there is no easy way to fix this, and I don't think we should search for ways to update the rolled-out scripts in a separate tab. As you rightly pointed out yesterday, one good option here would be to polish the error message and make it clearer for users. As discussed yesterday, I will create a separate issue soon.

#### Third-party plugins
Regarding third-party plugins that allow modifying the total on the checkout page. It looks like`updated_cart_totals` DOM event should be triggered after such changes, which makes it possible for us to handle the event and [dynamically update the Stripe Elements](https://stripe.com/docs/payments/finalize-payments-on-the-server?platform=web&type=payment#dynamic-updates) whenever the change occurs. However, there are two concerns I have in mind: 

1. I'm not sure we want to add the support logic for the third-party plugins in the main gateway script that handles payment processing for WooCommerce Payments, and even if we want, I think this could be done as part of the issue I mentioned above for all the other UPE types since the problem isn't local. 
2. Those gateways that can't be used for subscription purchasing will disappear after the Stripe Element for them is updated. Then, we will have an empty `div` (see below) and need to find a way to remove this without refreshing the page.
image

3. Another thing is that third-party plugins might allow changing the amount by updating the item count in the cart, which is not a problem for us since they won't be able to replace a physical product with a subscription. But plugins that allow changing the products from the checkout page (is it even possible/reasonable?) could be a problem.

While not impossible, all this feels like a new feature for the existing payments, making me feel we might want to brainstorm ideas and address this defect across all the UPEs separately. Please let me know what you think about all this - I'm happy to provide more information if needed.

timur27: One more update: I've tried to apply a coupon which is one of the possible options to verify some of our assumptions quickly. After applying a coupon, the `updated_checkout` event is fired, which can then be handled w/o any problem. But, after I applied a 100% discount the payment gateways were automatically updated and disappeared (which is good since no payment is needed). 


https://user-images.githubusercontent.com/22319444/230667403-f5cf292b-8f65-4a5d-a666-39465639c123.mov

Given this, I'm curious if we need to address this problem at all? 
* When the total drops to zero due to coupons - gateways disappear. 
* In order to change the product list completely and move from a physical product to a subscription with a free trial, users still need to go back to the cart page.

FangedParakeet: >While not impossible, all this feels like a new feature for the existing payments, making me feel we might want to brainstorm ideas and address this defect across all the UPEs separately.

For now, I'd rather focus on achieving parity with the legacy UPE rather than comprehensively perfecting our implementation. So in any scenario, I think it's fine for us to return an error, if the existing UPE implementation also returns an error; but if the legacy implementation has some mystical manner of handling such edge cases, we will need to devise one ourselves as well. 

> In order to change the product list completely and move from a physical product to a subscription with a free trial, users still need to go back to the cart page.

The only situation I can think would be if a user had two items in their cart: a regular simple product and a free-trial subscription product. At this point, presumably we'd complete the checkout using a payment intent; however, if a coupon were added to remove the simple product, reducing the checkout total to zero, we would still need to collect payment for the subscription product, but we would now need a setup intent and would not charge this payment method at this time. I will need to test to see how the legacy implementation handles this case, but--if it is even a scenario we will need to entertain--it is definitely an edge case that can be resolved in a separate issue.

timur27: > if a user had two items in their cart: a regular simple product and a free-trial subscription product. At this point, presumably we'd complete the checkout using a payment intent; however, if a coupon were added to remove the simple product, reducing the checkout total to zero, we would still need to collect payment for the subscription product, but we would now need a setup intent and would not charge this payment method at this time. 

Legacy UPE and split UPE are failing here. Legacy card and deferred intent creation flow handle this scenario. The fact that the intent is created after order placement seems to play a key role, and the Stripe UI Element's `mode` is not relevant in this use-case.

UPDATE: The issue created: https://github.com/Automattic/woocommerce-payments/issues/6030

timur27:
@FangedParakeet I attempted to improve the readability and quality of code as part of 1006c7bde2bde5b708d42db0416133e4b66d0585. I did this because after not reading the code for a few days and getting back to it, I noticed that some things were unclear to me immediately. The code was hard to read for someone who didn't work on the functionality before, tests were difficult to implement because of being limited by jQuery, etc. 

I decoupled jQuery from all the other logic - it's now placed in `event-handlers.js` script. Unfortunately, we cannot eliminate it completely at this stage because events like `updated_checkout` are triggered by jQuery and cannot be observed by vanilla JS.

I moved all the non-jQuery logic to the `stripe-checkout.js` script. The implementation still handles jQuery form in some places because some existing utility functions are still designed for it. But we are one step further here to fully remove jQuery, I believe. 

Script functions now have better readability, they also try to be responsible for only one task. 

I started to add tests into `stripe-checkout.test.js` but I'll aim to add more within the next 24 hours. In the meantime, I think there will be no other big shifts in the implementation, so I decided to push this for your review now. Thank you for all your time and efforts while reviewing this!

includes/class-wc-payments-upe-checkout.php
```diff
@@ -115,6 +115,10 @@ public function get_payment_fields_js_config() {
 			$payment_fields['checkoutTitle']        = $this->gateway->get_checkout_title();
 			$payment_fields['upePaymentIntentData'] = $this->gateway->get_payment_intent_data_from_session();
 			$payment_fields['upeSetupIntentData']   = $this->gateway->get_setup_intent_data_from_session();
+		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+			$payment_fields['checkoutTitle'] = $this->gateway->get_checkout_title();
+			$payment_fields['currency']      = get_woocommerce_currency();
+			$payment_fields['cartTotal']     = WC_Payments_Utils::prepare_amount( WC()->cart->get_total( '' ), get_woocommerce_currency() );
```
FangedParakeet: I tested this PR on a fresh WP site and when I was adding a currency changer widget, I had to comment out this line, otherwise I was met by the following unfriendly and fatal PHP error.

```
Fatal error: Uncaught Error: Call to a member function get_total() on null in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-upe-checkout.php on line 121 Error: Call to a member function get_total() on null in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-upe-checkout.php on line 121
```

I didn't actually realise that our code could be called on these pages, but I would guess that these scripts are just arbitrarily included along with other footer elements on this widget preview page. This error prevented anything from loading on the page at all: not including our scripts on this page would probably a more ideal solution, but in any case, this is a fatal error we will obviously want to prevent. Simply checking for the presence of a cart object would probably do the trick, in which case maybe there are more config fields we may want to remove as well.

FangedParakeet (commented):
I've just finishing putting this PR through its paces by plying it against our [WCPay critical test flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows) on a fresh (Docker) local WP site. Generally, the main shortcode checkout functions worked as expected. 🎉 But there were a few rather important things that were broken along the way.

Please check out the complete list of my tests over [here](https://docs.google.com/spreadsheets/d/1DIxtcrEiknLHGmn5HiYGONNo_bwhy1LAxmhuWGdMD6k/edit#gid=0). I skipped anything to do with the Blocks checkout understandably, as well as some more complicated onboarding flows and any transaction/dispute flows, which were unfortunately not possible to complete. 

Anyways, enough suspense--here are the things that did not work as expected for me: please let me know if you're able to replicate any of these events or whether I need to provide more details for you!

### Checkout with non-card UPE methods

I noticed that for payment methods with delayed confirmation, normally when the order is received the order status will change to pending payment and a note will be added indicating that the payment has started.

A payment was started using WooCommerce Payments
_Initial order note_

This status will be updated later once the payment is authorised and successfully confirmed; I think this is usually completed via a webhook from Stripe (itself via the WCPay Server). While testing this PR, I discovered that this order status never updated for me and remained frozen at "Pending payment". 

When selecting the payment intent ID from the order details page, I could see that despite the status remaining constant, the intent had been confirmed successfully and Stripe's dashboard would also attest this. However, we failed to update the order status despite this great success. 

I also noticed that on the payment intent details page, the order number field was left blank.

 Order number missing
 _Order number missing_
 
This suggests that some connecting metadata might be misplaced whilst processing the checkout: either the order number is not being stored on the intent in WCPay (the order details are present on the intent in Stripe's dashboard) or the intent ID is not being added correctly as metadata of the order. If this is indeed true, I think (and optimistically hope) that this will also be the cause of most of the other mysterious mishaps that I encountered.

### Checkout after failing an order

If I failed a transaction, for example by selecting to fail the transaction on Stripe's confirmation redirection page, upon returning to checkout if I tried to complete a valid transaction during this same checkout, I found myself unable to continue. 

The initial failed transaction returned an error as expected, but upon retrying to complete the checkout--this time with valid information--I was redirected to the Order Received page, but shown an error message stating that, "Unfortunately your order cannot be processed as the originating bank/merchant has declined your transaction. Please attempt your purchase again." The order status remained `failed`, however, in Stripe's dashboard I could see that the payment intent had been confirmed successfully and the customer presumably charged. 

### Pay for order

This one broke pretty quickly. The payment element did not even load: instead the standard error message was presented stating, "We're not able to process this payment. Please refresh the page and try again." In the JS console, there was an error relating that `encryptedValue is undefined`. 

AFAIA the pay for order page is not entirely functional with the split-UPE as it currently stands, so I'm unsure if this error is related to #5808.

### Subscriptions

Impressively almost all of these flows proceeded happily as intended! However, I was unable to change the payment method or add a new payment method to an existing subscription. I believe this relies on the same checkout page that is used for the Pay for Order flow and the same error as above presented itself in this scenario as well.

### Refunds

When a refund was successful, I was happy to see that status change notes were added to the order; however, when [the refund failed](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows#refund-failure), these notes were unusually not added to the order. This could be related to the missing metadata issue mentioned above.

### Transactions

From individual orders I could see valid transaction details, by clicking the payment intent ID link on the order itself. However, if I visited the dashboard at **Payments > Transactions** the table was completely empty, stating, "No data to display." The data is clearly there, but not being presented in this table. I suspect that some metadata link is once more absent, so hopefully this can be a quick fix.

That's basically it! I hope that many of these issues might have the same root cause, but I didn't have time to investigate too deeply. I didn't really find any obvious vulnerabilities along the way, but there are some breaking effects that extend beyond the additional UPE payment methods. 

I'd be happy creating new issues to investigate and resolve these issues--unless you are either unable to replicate any or are able to fix them yourself with ease within this PR. We can later decide whether we merge resolutions into this PR branch or merge the lot straight into `develop`. I'll save that decision for a squad meeting discussion.

Anyways, fantastic work--the large majority of critical test flows were supported by this PR, so that's a pretty prominent success. Nice one!

includes/class-wc-payments-upe-checkout.php
timur27: That's a good catch. I was able to reproduce it by simply opening the Widgets page. Fixed by 1996305c34df5510438c015fc2d6632f0d1bb1f5.

timur27:
Thanks for the thorough testing and the report, Samir :rocket:
#### Checkout with non-card UPE methods
The behavior with the order status not changing you described is not a bug, in my opinion. This behavior is specific to the Sofort payment method. Sofort is a payment method with delayed notification (while other non-card gateways like giropay, EPS, iDEAL, P24, Bancontact are immediate notification type of payment methods). In the real-life scenario in non-testing mode, a payment typically takes 2 to 14 business days [[source](https://stripe.com/docs/payments/sofort/accept-a-payment)], whereas in the test mode all the purchases made with Sofort remain pending for 3 minutes [[source](https://stripe.com/docs/billing/subscriptions/sofort#testing)].
> :bulb: From [this documentation](https://stripe.com/docs/payments/payment-methods#payment-notification) I conclude that ACH debits also have a delayed notification approach.

I’ve tested it once more on my local, using Sofort. After the order is placed _and_ the payment is authorized via the third-party authorization page, the Order status is indeed _Pending payment_. The pi status on the Stripe side is also _Pending_, so there’s no discrepancy between the Stripe dashboard and WC Orders page at this stage.



image

After 3 minutes, the status of the payment intent on the Stripe side changes from _Pending_ to _Succeeded_ and, assuming our local WCPay Server environment listens to the webhooks, the Order status in WooCommerce changes to _Processing_ as well.
image

And the statuses history updates as well


with the order id present in the transaction details as well


Stripe payment status is _Succeeded_ at this point as well, in line with the WCPay.
image

image

The same behavior is observed in the split-UPE implementation.
That said, please let me know if this clarifies our concerns or if something still needs an alignment. :+1:

#### Checkout after failing an order
There are two things to consider here. First is the Order status update, and the second is the _Unfortunately your order cannot be processed as the originating bank/merchant has declined your transaction. Please attempt your purchase again._ message on the _Order received_ page when you checkout successfully after a failure.
The Order status on the WooCommerce side updates for me successfully and is in line with the Stripe dashboard. The status changes immediately if you use a payment method with an immediate notification system. (e.g. giropay). Have you used Sofort for this test and if yes, is the status `Failed` for you after 3 minutes as well?

| giropay | Sofort |
|------------------------|------------------------|
| Image 1 Description | Image 2 Description |




The second thing is the text about the originating bank/merchant declining the transaction. I can reproduce it, and I can also reproduce it in the split UPE. The root cause of this issue is that the order is still in the status `failed` when performing a successful checkout (webhook from Stripe doesn't arrive yet), and [this WC Core implementation](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/templates/checkout/thankyou.php#L29-L31) is executed. We need to cover this fix in a separate issue, I think. Please let me know if you agree or not.



#### Pay for order
The _Page for order_  page was considered by me at the same level of user value as Blocks checkout, shortcode checkout, and My account page. It wasn’t in the scope, and thus there was no implementation, which is why it expectedly failed. I believe this page should be implemented in a separate script as a separate functionality, possibly re-using some pieces implemented by `stripe-checkout.js`, and events to start the _Pay for order_ processing should be observed by the `event-handlers.js` script. I think it's a good candidate to be covered by a separate issue. Please let me know if it would align with this MVP expectation to have the _Pay for order_ page not implemented yet, and I'm happy to create an issue for this.

#### Subscriptions
The same situation here as with the _Pay for order_ page - adding/changing payment methods wasn’t considered part of the scope. Given that the script handling this use-case is the same as the script for handling the _Pay for order_ page, I think handling them together within a separate issue would be a good option. I'm looking forward to hearing your opinion on this.


#### Payment methods
This is something I came to when reviewing your tests. I think this page needs to be considered as not implemented, the same way the above _Pay for order_ and _Subscriptions_ are, and possibly covered by a separate issue. Currently, there's no possibility of adding a new payment method when the deferred intent creation UPE is enabled. Please let me know if you agree or not.

#### Refunds
I’ve followed [the steps described](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows#refund-failure), and I was able to test the feature successfully. There is step nr 7, which is tricky because it can easily be missed. It says one should refresh the page to see the unsuccessful notes. Could you please let me know if you can still reproduce this issue or if it’s not the issue anymore?

<p align="center">
  
</p>


#### Transactions
I can successfully list all the transactions on the _Payments -> Transactions_ page, and I'm not sure why we have this difference. Please let me know if there are any more steps I can perform to break the page.

<p align="center">
  
</p>


#### The summary of the failed use cases and what's left to be done with them.

Please let me know if you think it's better to address any of the below points differently. The _To-Do_ column created by me has only the proposals in it, so I'm happy to discuss if any changes are needed.

| Use-case                                               | To-Do
|--------------------------------------------------------|--------------------------------------------------------
| Checkout failures (with various cards)                 | Re-test, since the initial test was performed with Sofort payment which has a 3-minute notification delay |
| Refund failure                                         | Re-test, since it works for @timur27 |
| View transactions                                      | Re-test, since it works for @timur27 |
| Filter transactions                                    | Re-test, since it works for @timur27 |
| Search transactions                                    | Re-test, since it works for @timur27 |
| Download transactions as CSV file                      | Re-test, since it works for @timur27 |
| View transaction details                               | Re-test, since it works for @timur27 |
| Dispute created notifications | Re-test after the above rows are passing |
| Manual capture from Uncaptured tab in Transactions page| Re-test after the above rows are passing |
| View Uncaptured transactions in Transactions page      | Re-test after the above rows are passing |
| 3DS authentication for cards | Needs a re-test after the latest changes |

timur27:
The proposed list of use cases to be covered by follow-up issues: 

| Use-case                                               | To-Do
|--------------------------------------------------------|--------------------------------------------------------
| Add a new payment method to an existing subscription                 | Address within the issue X (to be created) |
| Change payment method of the existing Subscription to new card | Address within the issue X (to be created). |
| Change payment method of the existing Subscription to saved card | Address within the issue X together with the previous row|
| Pay for order flow                 | Address within the issue X (to be created). Decide if the above rows should be addressed together with the Pay for order flow and create one issue for them all, or create two separate issues |
| My account -> Payment methods flow | Address within the issue Y (to be created) | 
| 3DS cards saving for the physical products | To be covered in a separate issue Z (to be created) |
| 3DS cards saving for the subscription products | To be covered in a separate issue (probably together with the previous row ) | 
| Subscription renewal using WooCommerce Subscriptions plugin for the subscriptions bought using a 3DS card | To be covered in a separate issue (probably together with the previous row) |

FangedParakeet:
Just finished retesting this PR on a JN this time, so edging slightly closer to a production environment. Focussing on my failures from the former effort, behold my findings this time.

### Checkout with non-card UPE methods (status not updating)

**Unable to replicate!** 🎉  

However, one thing worth noting, but while an order is in a pending status, we don't show any order details on the payment intent details page.

Sofort PI with missing order details
_Sofort pending transaction with missing order details_

The order has been created and these details have been attached to the intent itself, on Stripe's side. However, presumably because [we exit from processing the redirect payment if the status is pending](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L716-L724), we don't connect the transaction to the present order in WCPay. 

This isn't an enormous tragedy, since these details do successfully update, when the payment succeeds and the webhook completes the order. 

Sofort PI has order details when paid
_Sofort transaction has order details after payment is made_

This is also not a bug exclusive to this implementation and likely affects all versions of the UPE, so that lowers its priority somewhat, but I think we should still create a new issue to resolve this issue.


### Checkout after failing an order

**Still able to replicate.** 😞 

> The root cause of this issue is that the order is still in the status failed when performing a successful checkout (webhook from Stripe doesn't arrive yet), and [this WC Core implementation](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/templates/checkout/thankyou.php?rgh-link-date=2023-04-17T12%3A39%3A03Z#L29-L31) is executed. We need to cover this fix in a separate issue, I think.

Agree, absolutely. Just to reiterate, I expect almost all of the issues noted in this round of testing to be resolved in new and separate issues. 


### Pay for order

**Still broke.** 😞 


### Refund failure (notes missing)

**Unable to replicate!** 🎉 


### Transactions not displaying

**Unable to replicate!** 🎉 


### 3DS Authentication

**Works...?** 🎉 

This was working fine for me and I think we've discussed that any perceived issues are not worthy of worry, so I think this is fine...?


### My account -> Payment methods flow

**Works...?** 🎉 

Was there any issue with this page? This functionality appeared to be working for me. 


## New issues

So here at last is my final list of issues that I believe will remain pending once this PR is merged--*to be solved in separate issues*.

- Order status missing when payment is pending.
- Pay for order page 
- Add/change payment method for subscription (possibly solved by above issue)

...And I think, excepting anything else already pending for the existing incarnation of the split-UPE gateway, I believe that's it, which is honestly a fantastically short list of pending items. Nice one, you smashed it!

Let me know if anything is amiss in my testing again or if there are any suspicious scenarios that I neglected to test, but otherwise I think this puts us right on track with the remainder of this implementation.

timur27:
Thanks for the another testing round, @FangedParakeet! 

> 3DS Authentication
Works...? 🎉

3DS authentication works for the happy path scenarios when users only pay with it and authenticate/fail the transaction. If they want to save the 3DS payment method or renew the subscription bought with 3DS method, they won't be able to achieve this yet. This is why, I summarized these non-working use-cases in the last issue below. 

>My account -> Payment methods flow
Works...? 🎉

_Add payment method_ doesn't work, gateways aren't loaded. The same is with the gateways on the Pay for order and Change payment method for subscription pages, so I gathered them together and mentioned in the issue below. 

| Use-case                                               | Issue | Backlog | Status |
|--------------------------------------------------------|--------------------------------------------------------|--------------------------------------------------------|--------------------------------------------------------
| Checkout after failing an order | https://github.com/Automattic/woocommerce-payments/issues/6141 | Split UPE post-rollout | [Needs estimation](https://github.com/Automattic/woocommerce-payments/issues/6141#issuecomment-1521416269) |
| Checkout with non-card payment method with delayed notification (Sofort) | https://github.com/Automattic/woocommerce-payments/issues/6142| Split UPE post-rollout | [Needs estimation](https://github.com/Automattic/woocommerce-payments/issues/6142#issuecomment-1521454478) | 
| _Pay for order_ , _Change payment for Subscriptions_ and _Add payment method_ |https://github.com/Automattic/woocommerce-payments/issues/5023| New UPE - Deferred payment intent integration | Description was updated and needs to be reviewed. After confirming the acceptance criteria with @FangedParakeet  - start the estimation process | 
| 3DS cards saving for the physical products, 3DS cards saving for the subscription products, Subscription renewal using WooCommerce Subscriptions plugin for the subscriptions bought using a 3DS card  | https://github.com/Automattic/woocommerce-payments/issues/5025 | New UPE - Deferred payment intent integration | Description was updated and needs to be reviewed. After confirming the acceptance criteria with @FangedParakeet  - start the estimation process. Also, two blocking issues were linked to it - make sure everyone agrees with it |

bborman22:
It looks like this PR is in a good spot now with a really thorough review. And it seems we have a good grasp on the current limitations and a plan to address them. I'm dropping in the recent PR review questions here to confirm the need for some additional reviews.

**Does this PR potentially add or impact the functionality/accessibility of external APIs/interfaces?**
From what I can see here, there isn't any impact to external APIs particularly in regards to a security risk. The only things that caught my eye was the new client code, but my understanding is that is heavily based on existing client code? @timur27 would you be able to confirm that my understanding is correct here or point out areas where there might be some security concerns based on APIs or interfaces that may be more accessible now or introduced new parameters?

**Does this PR potentially impact the flow of payments?**
This PR will directly impact the Split UPE flow of payments though it is behind a feature flag. Due to the large nature of the PR I was hoping we could get a review from @dechov (or someone from the architecture team) to make sure this approach makes sense in the larger plugin structure. I was also hoping @RadoslavGeorgiev might have a chance to review this given the other efforts/considerations to refactor some of the existing gateways? Let me know if there would be any concerns with getting reviews for this PR though and we can address them!

**Does the PR description effectively document the changes that are being made so reviewers have good context?**
The PR links out to the existing issue to cover the why we are making the change, but perhaps it could be improved a bit with a short summary of why we're making the change in the PR. I think it could also be helpful to give a short summary about the implementation, particularly around what we might be re-using or creating new, and why we might be creating all new client code, just to preemptively answer some questions that might come up from reviews.

timur27:
@bborman22 Answering your questions:
#### Inputs
With the `event-handlers.js` being rolled out by the back-end gateway and being the entry point of all the new client code, we have two new input fields that are coming from the back-end to the client code - these are _[the currency and the cart total amount parameters](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/includes/class-wc-payments-upe-checkout.php#L120-L122)_. These fields are used to provide [three new parameters](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js#L135-L137) to [the new Stripe API method](https://stripe.com/docs/js/elements_object/create_without_intent). I don't see any problem with getting the values and exposing it to the client, because even if updated by a malicious actor for any reason, they won't be able to introduce any risks, since they are used as part of [the UI Payment Element mounting only](https://stripe.com/docs/payments/finalize-payments-on-the-server?platform=web&type=payment#add-the-payment-element-to-your-checkout-page), and the final values are still checked and updated on the back-end after users place an order.

#### Outputs
As mentioned earlier, we now use [the API method](https://stripe.com/docs/js/elements_object/create_without_intent) to create elements that do not require us to provide the client secret, as opposed to [the method used in the split UPE and the legacy UPE](https://stripe.com/docs/js/elements_object/create). 

Another new API call is occurring [here](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js#L69), but it's used only to [validate the form fields,](https://stripe.com/docs/js/elements/submit) and actually, everything is encapsulated on the Stripe side at this point, and we don't provide any data at all, not to mention any sensitive data.

Lastly, [we provide a new parameter](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/includes/class-wc-payment-gateway-wcpay.php#L1122) (`return_url`) in the legacy gateway implementation for non-card payment methods, which adds the support of the payment methods that require bank authentication. This field is passed as a request parameter to the WCPay Server and consequently to Stripe as it's [one of its accepted parameters](https://stripe.com/docs/payments/finalize-payments-on-the-server?platform=web&type=payment#submit-payment).

#### Code re-use and other security aspects
We
* don't manage/handle/expose/pass any data related to the payment intent or client secret, compared to the legacy UPE and the split UPE
* took the existing implementation as the baseline for script functions like `event-handlers.js`/`stripe-checkout.js`/`3ds-flow-handling.js`, by simplifying it in some places and decoupling the big functions into the smaller ones so that every function is responsible only for one thing which is way easier to test and maintain
* rollout the fraud scripts as [the first action after the page is loaded](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js#L23), the same way we do in [the previous implementation](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-split.js#L45)
* [try to acquire the fingerprint and stop elements mounting if no fingerprint is present](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js#L167-L175), and we are doing it at the same point in time as in [the previous implementation](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-split.js#L159-L169)
* [create the Stripe payment method object at the same time in the 
process and mount it to the hidden HTML input element](https://github.com/Automattic/woocommerce-payments/blob/8c40fac726b0730bdf88f26dc085d9b3925d3b87/client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js#L208-L211), similarly to the previous implementations. After trying to challenge this approach in p1680782020798119-slack-CGGCLBN58, the conclusion was that it could remain as it is and there are no risks with such an approach

timur27:
> Does the PR description effectively document the changes that are being made so reviewers have good context?
The PR links out to the existing issue to cover the why we are making the change, but perhaps it could be improved a bit with a short summary of why we're making the change in the PR. I think it could also be helpful to give a short summary about the implementation, particularly around what we might be re-using or creating new, and why we might be creating all new client code, just to preemptively answer some questions that might come up from reviews.


@bborman22 The PR description has been updated. Thanks!

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1117,7 +1117,25 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 				$request->set_off_session( $payment_information->is_merchant_initiated() );
 				$request->set_payment_methods( $payment_methods );
 				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
-				// Make sure that setting fingerprint is performed after setting metadata becaouse metadata will override any values you set before for metadata param.
+
+				if ( 1 === count( $payment_methods ) && 'card' !== $payment_methods[0] ) {
+					$request->set_return_url(
```
RadoslavGeorgiev, line 1123: Two questions here:

1. (Maybe I'm unprepared for this one) How does SEPA work here? I assume that we get IBAN fields, but there might still be a redirect?
2. This is very UPE-specific, but there is no indication here. I understand the plan is to eventually clean up and combine everything, but could you mention it in a comment here in the interim? Looking at the code by itself, it is not at all obvious why, how, and when this return URL would be used.

client/checkout/utils/upe.js
```diff
@@ -64,3 +70,140 @@ export const getPaymentIntentFromSession = (
...
+			.getAttribute( 'id' );
+	} else if (
+		null !== document.querySelector( 'li.woocommerce-PaymentMethod' )
+	) {
+		selectedGatewayId = document
+			.querySelector(
+				'li.woocommerce-PaymentMethod input.input-radio:checked'
+			)
+			.getAttribute( 'id' );
+	}
```
RadoslavGeorgiev: This could be significantly simplified:
```suggestion
	const radio = document.querySelector( 'li.wc_payment_method input.input-radio:checked, li.woocommerce-PaymentMethod input.input-radio:checked' );
	if ( null !== radio ) {
		selectedGatewayId = radio.id;
	}
```

Also, is there any reason to use `getAttribute( 'id' )` instead of just `id`? It's a native HTML prop.

RadoslavGeorgiev (commented):
Code looks good, great job! I only left a couple of small questions. Just curious, what are your plans regarding blocks support?

includes/class-wc-payment-gateway-wcpay.php
timur27: 1. This is a good callout! SEPA support was not considered part of this PR, which was basically because it's not fully compatible with the split UPE and is currently disabled. Attaching the `return_url` to this payment method doesn't make sense, but currently, SEPA won't work even without attaching the `return_url`. I'll consult the creation of an issue to implement SEPA for the deferred intent creation with the team and will link it here. UPDATE: we'll cover SEPA in https://github.com/Automattic/woocommerce-payments/issues/6205.
2. Fair suggestion - thank you! I've updated the code in 42ac641c74405b95b19fca442e56813e30d7bb46

client/checkout/utils/upe.js
timur27: Excellent suggestion, and there's no reason to use `getAttribute('id')` instead of `id`. I've mostly copied this code from another place, but I fully agree that we better improve it as well. 749d7d9229b6076e3ce8b4bff8716a1da14b3270

timur27:
@RadoslavGeorgiev, thanks for your review. I addressed two threads. 

> what are your plans regarding blocks support?

Blocks support will be introduced in https://github.com/Automattic/woocommerce-payments/issues/5020.

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -119,8 +119,8 @@ public function register_scripts() {
 		if ( $this->supports( 'tokenization' ) ) {
 			$script_dependencies[] = 'woocommerce-tokenization-form';
 		}
-		WC_Payments::register_script_with_dependencies( 'wcpay-upe-checkout', 'dist/upe_split_checkout', $script_dependencies );
-
+		$script = ( WC_Payments_Features::is_upe_deferred_intent_enabled() && ! WC_Payments_Features::is_upe_split_enabled() ) ? 'dist/upe_with_deferred_intent_creation_checkout' : 'dist/upe_split_checkout';
```
FangedParakeet:
```suggestion
		$script = ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) ? 'dist/upe_with_deferred_intent_creation_checkout' : 'dist/upe_split_checkout';
```
Presumably if we reach this point, then at least one of `is_upe_deferred_intent_enabled()` and `is_upe_split_enabled()` is `true`. The logic here ensures that if the latter is true--and the original split-UPE flag is enabled--the new deferred intent flag would be overridden and the original split-UPE would be returned instead. Is this intended/expected?

As with the split-UPE release--where when the split-UPE is eligible and enabled, all other UPE flags are ignored--it makes sense to me to have the deferred intent flag similarly override all of its ancestors as well. That way we can simply enable the feature by enabling the single flag, without having to worry about disabling the others. 

Let me know your thoughts or if I'm missing anything here!

includes/class-wc-payments-upe-checkout.php
```diff
@@ -115,6 +115,11 @@ public function get_payment_fields_js_config() {
 			$payment_fields['checkoutTitle']        = $this->gateway->get_checkout_title();
 			$payment_fields['upePaymentIntentData'] = $this->gateway->get_payment_intent_data_from_session();
 			$payment_fields['upeSetupIntentData']   = $this->gateway->get_setup_intent_data_from_session();
+		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+			$payment_fields['checkoutTitle'] = $this->gateway->get_checkout_title();
```
```suggestion
```
I think there's old, revived logic that is better off remaining at rest at play here. This parameter is used to rename the CC gateway to "Popular payment methods" when it contains more than one payment method: this will never be the case with deferred intent split-UPE. The CC gateway should *always* just be labelled "Credit card / debit card" now; I can see that now the CC gateway is instead labelled "Popular payment methods" at checkout.

It's a small cosmetic issue, so happy to resolve this separately, if needs be.

client/checkout/classic/upe-deferred-intent-creation/3ds-flow-handling.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+import { isWCPayChosen } from 'wcpay/checkout/utils/upe';
+import { getConfig } from 'wcpay/utils/checkout';
+import showErrorCheckout from 'wcpay/checkout/utils/show-error-checkout';
+
+const getPaymentMethodId = () => {
+	return isWCPayChosen()
+		? document.querySelector( '#wcpay-payment-method-sepa' )?.value ?? ''
+		: document.querySelector( '#wcpay-payment-method' )?.value ?? '';
+};
```
FangedParakeet, line 12: Can you enlighten me on what this function is doing? My sleepy noggin is a little bit lost.

If WCPay is chosen--and the WCPay gateway (`#payment_method_woocommerce_payments`) is checked--we return the SEPA payment method? Since this is in use to handle 3DS cards that require authentication, why should we ever return a SEPA payment method?

In `showAuthenticationModalIfRequired()` below, this returns an empty string for me, which makes sense given the logic of the function, but doesn't seem like it's intended. This might explain why saving 3DS cards is apparently non-functional, because when `shouldSavePaymentPaymentMethod()` is `true`, it looks like we will just send the empty string as the `payment_method_id` parameter.

Let me know if I'm misunderstanding anything here! I know we have another issue to resolve this, so I'm not arguing that this is resolved here immediately, but just wondering if there's anything soaring gently over my head. Cheers!

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
FangedParakeet, line 1: Just wanna say, I'm a big, big fan of the major refactor you've performed here: code looks squeaky clean and I'm overjoyed to not have to scroll frantically through thousands of lines of code in a single file to find the one piece I'm looking for. The hero we didn't deserve, but the one we needed. 🙌

client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js
```diff
@@ -0,0 +1,242 @@
...
+export const checkout = ( api, jQueryForm, paymentMethodType ) => {
+	if ( hasCheckoutCompleted ) {
+		hasCheckoutCompleted = false;
+		return;
+	}
+
+	blockUI( jQueryForm );
+
+	const elements = gatewayUPEComponents[ paymentMethodType ].elements;
+	validateElements( elements, jQueryForm );
```
FangedParakeet, line 204: If `elements.submit()` returns an error, we should exit out of this function and not continue with calling `createStripePaymentMethod` below. If I leave one of the credit card fields empty and hit "Place order" at checkout, `validateElements()` displays an error underneath the empty field, but I also get an error at the top of the checkout stating that "paymentMethodObject.paymentMethod is undefined", which I presume is caused by `createStripePaymentMethod`, which is summoned when not yet ready.

client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js
```diff
@@ -0,0 +1,242 @@
...
+ * form UI to prevent duplicate submission and validates the Stripe elements. It then creates a Stripe payment method
+ * object and appends the necessary data to the form for checkout completion. Finally, it submits the form and prevents
+ * the default form submission from WC Core.
+ *
+ * @param {Object} api The API object used to create the Stripe payment method.
+ * @param {Object} jQueryForm The jQuery object for the form being submitted.
+ * @param {string} paymentMethodType The type of Stripe payment method being used.
+ * @return {boolean} return false to prevent the default form submission from WC Core.
+ */
+let hasCheckoutCompleted;
```
FangedParakeet, line 194:
```suggestion
let hasCheckoutCompleted = false;
```
Not a huge deal, but if we don't need to check for this variable's `undefined` state, why not be more explicit and just declare this as `false` from the beginning?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1076,7 +1076,7 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 
 			if ( ! empty( $upe_payment_method ) && 'woocommerce_payments' !== $upe_payment_method ) {
 				$payment_methods = [ str_replace( 'woocommerce_payments_', '', $upe_payment_method ) ];
-			} elseif ( WC_Payments_Features::is_upe_split_enabled() ) {
+			} elseif ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 				$payment_methods = [ 'card' ];
```
FangedParakeet, line 1080: I think the only way to get here is via a checkout with a saved payment method. Right now the only possible payment method for this flow is cards, but when SEPA is added, we may need to revisit this. Disregard for now, but just commenting more as a reminder for myself. 😅

FangedParakeet (commented):
Just have a few last little clarifications here, but this is working pretty well for me so far--beyond the other issues that have already been identified and captured in separate issues! 

Also, P.S. I'm seeing Stripe's new skeleton loading state on the payment element for the first time--I quite like it actually and think it fits with our checkout and loading states (particularly in blocks) pretty well!

includes/payment-methods/class-upe-split-payment-gateway.php
timur27: The current implementation is something I introduced consciously, and I think I brought it one day to our catch-up. I wanted to bring an extra layer of protection for the deferred UPE and ignore it when it's enabled with the split UPE. I thought we'd review the enablement strategy as part of our current dedicated issue in our backlog. That's the background.

I have no objections and can definitely make the deferred intent UPE a priority across all the UPEs. I updated the code in 4de930d1f392ffaa0b1df32201ae7a482c26ba5a. Please not, that I also changed the order between the legacy UPE and the deferred intent UPE in one place, in order to make deferred UPE work as expected, when both flags are enabled. 

Having that said, I would vote for testing out all the 7 possible combinations of feature flags that are currently possible, then noting down those that for some reason should be amended and consider addressing them as part of the enablement strategy issue or any other issue on an earlier stage. Here are the possible combinations: 

Example: 
| Legacy UPE | Split UPE | Deferred UPE | The rolled out UPE     |
|------------|-----------|--------------|-------------|
| Enabled    | Enabled   | Enabled      | deferred |
| Enabled    | Enabled   | Disabled     | split |
| Enabled    | Disabled  | Enabled      | deferred |
| Enabled    | Disabled  | Disabled     | legacy |
| Disabled   | Enabled   | Enabled      | deferred |
| Disabled   | Enabled   | Disabled     | split |
| Disabled   | Disabled  | Enabled      | deferred |
| Disabled   | Disabled  | Disabled     | legacy card |

includes/class-wc-payments-upe-checkout.php
timur27: Sure, I removed the parameter in 0ab4dbcb53ef63d93ff383c71305021713769bca and deferred intent works without any disruptions afterward.

client/checkout/classic/upe-deferred-intent-creation/3ds-flow-handling.js
timur27: You're right; this function working wrong for the 3DS payments is probably why saving 3DS cards doesn't work. I'd vote for addressing this in https://github.com/Automattic/woocommerce-payments/issues/5025.

client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js
timur27: As discussed f2f, we'll leave it as it is due to some side-effects this small change brings.

client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js
timur27: As discussed f2f, we'll cover this in another issue due to the side-effects the change currently introduces.

includes/class-wc-payments-upe-checkout.php
FangedParakeet: I can still see the gateway titled "Popular payment methods," but no matter: created #6227. 👍

client/checkout/classic/upe-deferred-intent-creation/stripe-checkout.js
FangedParakeet: Created #6230.

FangedParakeet (approved):
LGTM! Thanks for making those last minute changes. 

Here are a few pending issues that we'll try to solve in our next sprint:
- #5020 
- #5023 
- #5025 
- #6141 
- #6142 
- #6227 
- #6230 

None of these are vulnerabilities and all are outside the scope of this initial MVP. As always, fantastic work, @timur27; thanks for your input @RadoslavGeorgiev & @dechov! I think we're ready to move this into `develop` and invite GlobalTest to do their worst with it. 

Let's go! 🚢

PR #5879, opened 2023-03-28
Migrate get_account_data API to the core Request class

htdat:
Fixes 
A part of https://github.com/Automattic/woocommerce-payments-meta/issues/87

This PR is going to migrate the first simple API first, then I will migrate subsequent endpoints in other PRs based on this one.

#### Testing instructions

* Use WCPay Dev tool, and vist the setting page. 
* Click `clear` for `Account cache content`.
* See no error in the PHP error log.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5879` or branch name `update/migrate-account-apis-to-request-class` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/migrate-account-apis-to-request-class) :rocket:

---

Build info: 

- Latest commit: 9d13dcf979483c0c5ae6fe304b5c54974479e174
- Build time: 2023-04-14 04:27:06 UTC

includes/core/server/request/class-get-account.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace WCPay\Core\Server\Request;
+
+use WCPay\Core\Server\Request;
+use WC_Payments_API_Client;
+
+/**
+ * Request class for getting intents.
+ */
+class Get_Account extends Request {
```
htdat, line 16: @Automattic/gamma I'd like to have early feedback regarding to this trait `Use_Test_Mode_Only_When_Dev_Mode`. We're having 4 requests like this under the current class `WC_Payments_API_Client`.

includes/core/server/request/class-get-account.php
```diff
@@ -0,0 +1,44 @@
...
+
+	/**
+	 * Formats the response from the server.
+	 *
+	 * @param  mixed $response The response from `WC_Payments_API_Client::request`.
+	 * @return mixed           The same response from REST API.
+	 */
+	public function format_response( $response ) {
+		return $response;
+	}
```
htdat: @Automattic/gamma - I'd like to have feedback on this too. Does this sound right? 

Or should I do not extend this method, and just use something like this  in consumers?

```php
$response = Get_Account::create()->send('hook_name');

$response->to_array(); 
```

htdat: FYI. I am removing the custom `format_response` method so that it can get the advantage of using the Response class.

includes/core/server/request/class-get-account.php
```diff
@@ -0,0 +1,34 @@
...
+namespace WCPay\Core\Server\Request;
+
+use WCPay\Core\Server\Request;
+use WC_Payments_API_Client;
+
+/**
+ * Request class for getting intents.
+ */
+class Get_Account extends Request {
+	use Use_Test_Mode_Only_When_Dev_Mode;
```
htdat, line 17: Given that, we're having 4 requests like this under the current class `WC_Payments_API_Client`, I am approaching with this `trait`.

includes/core/server/request/class-get-account.php
htdat: Marking as resolved, and moving this convo to https://github.com/Automattic/woocommerce-payments/pull/5879#discussion_r1155454194

includes/core/server/request/trait-use-test-mode-only-when-dev-mode.php
```diff
@@ -0,0 +1,31 @@
...
+	 * Class constructor.
+	 *
+	 * @param  WC_Payments_API_Client     $api_client  Api client.
+	 * @param  WC_Payments_Http_Interface $http_interface  Http interface.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function __construct( WC_Payments_API_Client $api_client, WC_Payments_Http_Interface $http_interface ) {
+		parent::__construct( $api_client, $http_interface );
+		$this->set_param( 'test_mode', WC_Payments::mode()->is_dev() );
```
zmaglica: I would avoid this approach. There is `DEFAULT_PARAMS` inside the abstract request class that could be used for these cases. Overriding the constructor makes things difficult because you will have to maintain the constructor param injection in multiple places.

htdat: The problem is that the dev mode value can be retrieved via an expression only, which does not support for const like `DEFAULT_PARAMS`.

Markup on 2023-04-03 at 17:34:24

> Overriding the constructor makes things difficult because you will have to maintain the constructor param injection in multiple places.

Not really, as in our usage, we use `create()`, which hides the constructor initialization. I myself don't like it just because it hides the constructor in a trait, which makes things hard to read. However, by introducing a trait, we can easily signify which classes have the same attribute.

zmaglica: Rather than creating a new trait that overrides the constructor, a better approach could be to add a specific function to the existing trait you added that handles the dev mode requirement in intent requests. This way, we can build upon our previous work and improve the trait's functionality without creating unnecessary complexity in the code.

For example, you can put that `test_mode` is required and inside trait add something like this:

```
public function add_test_mode_paramameter() {
      $this->set_param( 'test_mode', WC_Payments::mode()->is_dev() );
}
```

and call this function when you call other's setters. Or you could create `set_test_mode` function and fill it manually.

htdat: @zmaglica - awesome feedback. I've addressed it here 6604de1e3f6aa6e6388811fc652bab5d4469094f 

I am marking this as resolved now. In case, you have any feedback regarding the new approach, feel free to leave inline comments.

htdat:
Addressed the last feedback https://github.com/Automattic/woocommerce-payments/pull/5879#discussion_r1161511019

Ready for another review round.

includes/core/server/request/trait-use-test-mode-only-when-dev-mode.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Trait for requests that we only send a test mode request if in dev mode.
+ */
+trait Use_Test_Mode_Only_When_Dev_Mode {
+	/**
+	 * Set the test_mode parameter to true only if in dev mode.
+	 *
+	 * @return void
+	 */
+	public function set_test_mode_only_when_dev_mode() {
```
zmaglica: I would rename this function to simply `set_test_mode` or `add_test_mode_parameter` to avoid confusion, since the function name says something like that test mode will be set only when dev mode is enabled.

includes/core/server/request/class-get-account.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace WCPay\Core\Server\Request;
+
+use WCPay\Core\Server\Request;
+use WC_Payments_API_Client;
+
+/**
+ * Request class for getting intents.
+ */
```
RadoslavGeorgiev: Please update this class description :)

includes/core/server/request/trait-use-test-mode-only-when-dev-mode.php
RadoslavGeorgiev: I'm wondering whether this should be approached through an external method at all. The way the request worked within the API client did not allow the mode to be set externally, and maybe we should keep it similar. How about extending the `get_default_params` method within `Get_Account`, and including this straight in defaults?

htdat: @RadoslavGeorgiev - yeah, I would agree the point of avoiding an external method at all, and proposed extending the constructor to set_param `test_mode` right there https://github.com/Automattic/woocommerce-payments/pull/5879#discussion_r1155710619, which would be even more simple and straightforward than extend `get_default_params`. 

However, its readability is reduced and the idea of overriding the constructor in a trait is not really nice. Therefore, I went with the current approach, though it requires an external method.

htdat: > I would rename this function to simply set_test_mode or add_test_mode_parameter to avoid confusion, since the function name says something like that test mode will be set only when dev mode is enabled.

@zmaglica - I tried to balance this approach :D Your proposed names do not explicitly the relevance with `dev_mode` so I am using this name `set_test_mode_according_to_dev_mode` - commit c6e5724a3e58ba8009c5921426f90332eac2bed2

includes/core/server/request/class-get-account.php
htdat: 😬 - good point. Addressed 4c3cb3d834ea0db76bf3c676ba9a15d36ed7a74d

htdat:
Ready for another review :D

includes/core/server/request/trait-use-test-mode-only-when-dev-mode.php
RadoslavGeorgiev: > However, its readability is reduced and the idea of overriding the constructor in a trait is not really nice. Therefore, I went with the current approach, though it requires an external method.

I do agree that `get_default_params` should be obfuscated within a trait, however I'm not sure it's enough code to warrant a trait, especially if it could be automatically done within the request:

```php
public static function get_default_params() {
	$params              = parent::get_default_params();
	$params['test_mode'] = WC_Payments::mode()->is_dev(); // Explain why.
	return $params;
}
```

Using `get_default_params` to append a dynamic default parameter seems like very straight-forward and logical behavior anyway.

htdat: > however I'm not sure it's enough code to warrant a trait, especially if it could be automatically done within the request. 

The code is pretty simple, but I'd prefer to use trait as currently, there are 4 methods using this only dev-mode pattern in class `WC_Payments_API_Client` - [1](https://github.com/Automattic/woocommerce-payments/blob/4c2f691c6e0fda63292fa20385a99350da4216f7/includes/wc-payment-api/class-wc-payments-api-client.php#L861-L861), [2](https://github.com/Automattic/woocommerce-payments/blob/4c2f691c6e0fda63292fa20385a99350da4216f7/includes/wc-payment-api/class-wc-payments-api-client.php#L878-L878), [3](https://github.com/Automattic/woocommerce-payments/blob/4c2f691c6e0fda63292fa20385a99350da4216f7/includes/wc-payment-api/class-wc-payments-api-client.php#L897-L897), [4](https://github.com/Automattic/woocommerce-payments/blob/4c2f691c6e0fda63292fa20385a99350da4216f7/includes/wc-payment-api/class-wc-payments-api-client.php#L1032-L1032). IMO, using trait here will increase readability and reusability

> Using `get_default_params` to append a dynamic default parameter seems like very straight-forward and logical behavior anyway.

After looking at this again, I agree this point and better than overriding the parent class `Request`'s constructor. WDYT about this approach @zmaglica?

zmaglica: I like Rado's approach too, but since we have multiple request classes that use the same parameter, maybe we can use the best of both worlds. Inside Trait, add code that Rado shared and use that trait inside these endpoints to load dev mode param.

htdat: Great that we have the alignment here. Addressed 9d13dcf979483c0c5ae6fe304b5c54974479e174

htdat:
Ready to review again after addressing here https://github.com/Automattic/woocommerce-payments/pull/5879#discussion_r1166271547!

zmaglica (approved):

PR #5897, opened 2023-03-29
Improve authorization capture notification message in payment details page

### LINKED ISSUE #5192: Improve capture notification section in payment details page
### Description

Improve the capture notification section that appears in the payment details section when it is pending to be captured.

Screenshot 2022-11-25 at 09 17 00

### Acceptance criteria

* Include a link to the capture section in public documentation.
* Replace ‘You need to capture before DATE’ with something like ‘You have 2 days left’ as the deadline to capture the payment gets closer.

### Designs

TBD

### Additional context

P2
paJDYF-4N1-p2#comment-14945

nagpai:
Fixes #5192 

Improve the capture notification section that appears in the payment details section when it is pending to be captured.
* Include a link to the capture section in public documentation.
* Improve the deadline message to indicate a relative date / time , to help the merchant understand the deadline/ urgency better . eg - `You need to capture this charge in 6 days`

#### Screenshots 

Message with the tooltip:

Screenshot 2023-03-30 at 5 16 39 PM

#### Screenshots with different duration

**Hours:** 
image

**Minutes:**
image

#### Testing instructions

* Within `Payments > Settings > Transactions` make sure to enable `Issue an authorization on checkout, and capture later` , and click the blue `Save Changes` button
* From the front end, create at least three to four orders 
* Browse to `Payments > Transactions > Uncaptured tab`
* Click on any uncaptured transactions to open its `Payment Details` page
* The yellow box with the notice should read in a format similar to the screenshot shown above
* If you mouseover for a few seconds, the tooltip should show the actual deadline date, as shown in the screenshot. 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5897` or branch name `update/5192-improve-capture-notification-details` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/5192-improve-capture-notification-details) :rocket:

---

Build info: 

- Latest commit: b154e9d01b470867a3f8562bd89ed60de58b24bf
- Build time: 2023-04-20 03:53:03 UTC

kalessil:
@nagpai, how will the message look like for:
- 2 hours
- 40 minutes
- 15 seconds
?

nagpai:
@kalessil . I have updated the additional screenshots in the main comment above. I got these by changing the `authorization.created` param to a specific date within [here](https://github.com/Automattic/woocommerce-payments/blob/d2d487f9270bfb2ac3f44bc2328d1a9f0877420c/client/payment-details/summary/index.tsx#L373-L387)
  
  eg:
  
```
<b title={ dateI18n(
	'M j, Y / g:iA',
	moment
		.utc( [2023, 2, 24, 4, 18, 0 ]  )
		.add( 7, 'days' )
) }
>
{ sprintf(
	__( ' %s', 'woocommerce-payments' ),
	moment
		.utc( [2023, 2, 24, 4, 18, 0 ] )
		.add( 7, 'days' )
		.fromNow( true )
) }
</b>
```

I hope this is fine.

I will need to write a test for this, since the [existing test](https://github.com/Automattic/woocommerce-payments/blob/d2d487f9270bfb2ac3f44bc2328d1a9f0877420c/client/payment-details/summary/test/index.tsx#L210) will not work due to changed structure of the message.

kalessil:
Thanks, @nagpai ! No need - the new cases are rather for me to ensure see live how it'll look like for other dates. We can turn this draft into a PR now 🎉

nagpai:
@kalessil Thank you! I had to remove [this failing test](https://github.com/Automattic/woocommerce-payments/pull/5897/commits/2e22a2f8f228b449e5aacaab992d1ac7829d10c3#diff-c9d5dc585da0daf11bbc87caf598730591d551ba1371612bb80c0a64dac8dc79L210) - `renders capture section correctly`

kalessil:
@nagpai : thanks! As a note, we don't drop the tests until there are functional changes, so please recover the removed one and ensure it gets green again.

client/payment-details/summary/index.tsx
```diff
@@ -353,17 +354,37 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+									<b
+										title={ dateI18n(
 											'M j, Y / g:iA',
 											moment
 												.utc( authorization.created )
 												.add( 7, 'days' )
 										) }
+									>
+										{ sprintf(
+											__( ' %s', 'woocommerce-payments' ),
```
kalessil: LOL, @nagpai  is learning how to glue components together.

Do we need `__` here?
If yes, then a comment explaining that it ensures the proper translation locale is picked will be needed.

kalessil (approved):
LGTM! Left a minor comment though =)

client/payment-details/summary/index.tsx
nagpai: >LOL, @nagpai is learning how to glue components together. 

Thanks, I hope they remain stuck  Haha.. 

>Do we need __ here?
If yes, then a comment explaining that it ensures the proper translation locale is picked will be needed.

I have added a comment here f885ef6 .

nagpai: Sorry, removed some unwanted spaces in the comment. -4983028
Thanks for the review and approval! 🙌

anu-rock:
A general comment purely based on the screenshots above: it isn't immediately clear that hovering over the relative time (eg. 6 days) will reveal more details via a tooltip. We can perhaps use the [`<abbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) tag or a dotted border via CSS to make it more intuitive.

nagpai:
@anu-rock  Thanks for the good improvement callout there 🙏  . I have added it here d94e327

image

kalessil (approved):
LGTM!

client/payment-details/summary/index.tsx
```diff
@@ -367,18 +368,47 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
-									</b>
+									>
+										<b>
+											{ sprintf(
+												// eslint-disable-next-line @wordpress/i18n-no-placeholders-only
+												__(
+													'%s',
+													'woocommerce-payments'
+												),
+												// Wrapped in __() to ensure translation of relative time strings inserted by moment.fromNow() (e.g. "days", "hours", "minutes", etc.)
```
mgascam: I'm afraid that this will not work because `moment.fromNow()` is dynamic, so translators will only see that we want to translate '%s', but not the actual time left. A tip in case you did not know, If you run `npm run build:client` you can have a look at the .pot file that translators will use. I believe this is what it shows for this PR:

```
#: dist/index.js:1
#: client/payment-details/summary/index.tsx:192
msgid "%s"
msgstr ""
```

Instead, I can think of a couple of options:
1. We would have more than one sentence for translators, depending on how close we are to the deadline. If it is days something like ` __( 'You need to capture this charge in %s days' ), $timeToCapture )` and if it is hours ` __( 'You need to capture this charge in %s hours' ), $timeToCapture )`. I believe you get the point.
2. Maybe moment.js is capable of localizing that string on its own. I believe we are able to know the locale of the user. If something like this is possible, it will be less verbose than the other option.

Or maybe there's another, more clever way of doing this. I'll also share this [i18n Best Practices](https://fieldguide.automattic.com/i18n/developers/best-practices/) link in case it is useful.

mgascam (changes requested):
The code is working fine, but I believe there is an issue with the translation of the time left for capturing the charge.

client/payment-details/summary/index.tsx
nagpai: @mgascam . Thanks for the awesome pointers and help! 

I have managed to make changes that use Moment's i18N feature.  Two questions / concerns I wanted to check with you and others. 

1.  The [total package size](https://github.com/Automattic/woocommerce-payments/pull/5897#issuecomment-1488653592) seems to increase by 7% ( 73KB ) . Would that be fine or unacceptable? If not, I am happy to work via alternative method of using individual strings ( although verbose ) 

2. I am unable to find a solution to the typescript error I get for including the `moment-with-locales` [component here ](https://github.com/Automattic/woocommerce-payments/pull/5897/files#diff-a41f2e2fca1befa933534f9c72b4d1552d19af998a9dc2252f592d31e10b7a59R10) . I had to add an `ts error expected` exclusion note for now. If I could get some pointers or help. 

Thanks so much

mgascam: > I have managed to make changes that use Moment's i18N feature. Two questions / concerns I wanted to check with you and others.
> 
>     1. The [total package size](https://github.com/Automattic/woocommerce-payments/pull/5897#issuecomment-1488653592) seems to increase by 7% ( 73KB ) . Would that be fine or unacceptable? If not, I am happy to work via alternative method of using individual strings ( although verbose )

I'm glad that you worked your way through it, congrats! But If you look at the size of the `index.js`, it grows  +72.6 kB (+32%). I'd say that's a bit too much, specially if we are not making a heavy use of moment's locales in other places.  I'd rather go the direction of having individual strings in this case.

>     2. I am unable to find a solution to the typescript error I get for including the `moment-with-locales` [component here ](https://github.com/Automattic/woocommerce-payments/pull/5897/files#diff-a41f2e2fca1befa933534f9c72b4d1552d19af998a9dc2252f592d31e10b7a59R10) . I had to add an `ts error expected` exclusion note for now. If I could get some pointers or help.

~~Perhaps not relevant after my previous comment, but maybe you are getting an error because we are missing the type definitions of the moment library. As you may know, this type definitions can be often found on [github](https://github.com/aquavit/DefinitelyTyped/blob/master/moment/moment.d.ts) and added to your project (and those are not added to the bundle, in case you wondered)~~ Well, I just checked and moment includes its own `moment.d.ts` type definitions so please ignore @nagpai

mgascam (changes requested):
Looking at the size increase, I think we should discard using `moment` locales.

client/payment-details/summary/index.tsx
nagpai: Hi @mgascam . When you get time can you please review my latest changes? 

1. I have removed MomentJS and its locale, and tried implementing using plain vanilla. It seems to work fine in manual tests. 

2. On the test snapshots however something strange is happening, where you may kindly guide me. The current date and time is mocked at `Jan 1, 2023, 01:00 AM` .. The snapshots however seem to render `Jan 7, 2023, 3:00PM` , instead of what I would expect `Jan 8, 2023, 01:00AM`.  I may be doing some silly mistake there. For now to make the tests pass, I have matched the expected string to what the snapshot ends up throwing. Would appreciate your review and guidance 🙌

client/payment-details/summary/index.tsx
```diff
@@ -367,18 +413,76 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+														authorization.created
+													).getHours(),
+													new Date(
+														authorization.created
+													).getMinutes(),
+													new Date(
+														authorization.created
+													).getSeconds()
+												)
+											)
```
mgascam: We might be missing the `timezone` argument in the call to the dateI18n function. 
```suggestion
											),
											'UTC'
```
Can you give it a try and let me know if that improves things?

client/payment-details/summary/index.tsx
```diff
@@ -367,18 +413,76 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+													Date.UTC(
+														new Date(
+															authorization.created
+														).getFullYear(),
+														new Date(
+															authorization.created
+														).getMonth(),
+														new Date(
+															authorization.created
+														).getDate() + 7,
```
mgascam: This does not seem to be correct. If I understand correctly, we are creating a new Date object, getting the date and then adding `7.` But if the date is 31 then we would end passing 38 to the date argument of Date.UTC, which I think is incorrect (it should be an integer between 1 and 31). I'd go with a different approach: first parse the `authorization.created` to have an integer timestamp, so you could add the seven days and calculate the time remaining. I am thinking that moment could be handy in this case 🥲 

Also, I think we should try to avoid the repetition of Date here. Maybe this can be handled inside the `relativeTimeRemaining` function?

client/payment-details/summary/test/__snapshots__/index.tsx.snap
```diff
@@ -24,9 +24,9 @@ exports[`PaymentDetailsSummary correctly renders a charge 1`] = `
               usd
             </span>
             <span
-              class="chip chip-light"
+              class="chip chip-primary"
             >
-              Paid
+              Payment authorized
```
mgascam, line 29: I was not expecting to see this change of status in the payment TBH. I guess it comes from changes in `develop` branch 🤔

client/payment-details/summary/test/__snapshots__/index.tsx.snap
```diff
@@ -272,6 +310,22 @@ exports[`PaymentDetailsSummary renders a charge with subscriptions 1`] = `
         <div
           class="payment-details-summary__section"
         >
+          <div
```
mgascam, line 313: Maybe this is why we see changes in the status of the payment?

client/payment-details/summary/test/__snapshots__/index.tsx.snap
```diff
@@ -698,37 +754,11 @@ exports[`PaymentDetailsSummary renders capture section correctly 1`] = `
         </li>
       </ul>
     </div>
-    <div
```
mgascam, line 701: Is the removal of the capture section expected? 🤔

client/payment-details/summary/test/index.tsx
```diff
@@ -207,92 +212,108 @@ describe( 'PaymentDetailsSummary', () => {
...
-			paymentIntent: paymentIntentMock,
-			fraudOutcome: latestFraudOutcomeMock,
+		test( 'renders capture section correctly', () => {
+			mockUseAuthorization.mockReturnValueOnce( {
+				authorization: {
+					captured: false,
+					charge_id: 'ch_mock',
+					amount: 1000,
+					currency: 'usd',
+					created: new Date( Date.now() ).toISOString(),
```
mgascam: I think I understand your intent here, but please note that this date format may be different from the one that the server is returning. IIRC the server has` 2023-04-12 14:24:12` but in this case we have `2023-04-12T14:24:12.752Z`. Note the missing timezone information.

mgascam (changes requested):
I am not convinced that we are going in the right direction with latest changes. Maybe it was a bad idea to remove the moment library, sorry about that.

I left some comments with hints that I hope are helpful.

client/payment-details/summary/test/index.tsx
nagpai: Thanks for this callout! I have used moment in the fixed commit. It seems to work fine now.

client/payment-details/summary/test/__snapshots__/index.tsx.snap
nagpai: I think the snapshot is changing since we have replaced the static date showing in the capture notice, with a relative time. I hope my understanding is correct here?

client/payment-details/summary/index.tsx
nagpai: Thanks for your guidance, @mgascam ! It now seems to be working fine!

client/payment-details/summary/index.tsx
nagpai: I have added it just now 👍 . Thanks

client/payment-details/summary/index.tsx
nagpai: Ah.. I missed this simple detail ( about 7 being added to the date without checking the possibility of largest date of the month getting exceeded ). I have implemented the whole thing with Moment, that made it simpler too. I would have loved to do this with vanilla JS, but looks like moment results in crisp and readable code.

client/payment-details/summary/test/__snapshots__/index.tsx.snap
nagpai: Maybe due to the snapshot changes, as explained in the earlier note. ( static date changed to relative time )

client/payment-details/summary/test/__snapshots__/index.tsx.snap
mgascam: Hmm, I'd say that the date is not relevant for the status of the payment, but would need to dig into that to be sure. Unfortunately, I don't have the capacity at the moment.

client/payment-details/summary/index.tsx
```diff
@@ -175,6 +176,51 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 		paymentIntent
 	);
 
+	const relativeTimeRemaining = ( deadlineDate: Moment ): string => {
```
mgascam: This is a nitpick comment. I like to have verbs in the name of the functions. I'd suggest `calculateRelativeTimeRemaining` but maybe it's just me. Feel free to ignore it.

mgascam (commented):
It's great to see that the remaining time calculation seems to be fixed. 
Before approval, I'd like to understand better the changes in the snapshot. I was not expecting to see some of them, that seem unrelated to the purpose of this PR.

client/payment-details/summary/test/index.tsx
```diff
@@ -207,92 +212,108 @@ describe( 'PaymentDetailsSummary', () => {
...
+			const charge = getBaseCharge();
+			charge.captured = false;
+
+			const container = renderCharge( charge );
+
+			expect(
+				screen.getByRole( 'button', { name: /Capture/i } )
+			).toBeInTheDocument();
+
+			expect(
```
mgascam: I'd say that this assertion seems to be duplicated with the` toMatchSnapshot()` below.

nagpai: It was indeed redundant! I have retained the snapshot assertion and removed the exact element compare.

client/payment-details/summary/index.tsx
nagpai: I have renamed this :)

client/payment-details/summary/test/__snapshots__/index.tsx.snap
nagpai: The status here seems to be correct as expected - `Payment authorized`. Not sure why it was showing `Paid` in the status earlier though. 🤔

client/payment-details/summary/test/__snapshots__/index.tsx.snap
nagpai: During the time of this PR there was indeed another one added related to Fraud review. The snapshot related to that seems to have gotten inserted. I hope I am reading this correctly?

kalessil (changes requested):
Judging by the custom formatter re-invention, it looks like we are following the wrong direction when it comes to translation.

Please note [this](https://github.com/moment/moment/blob/e96809208c9d1b1bbe22d605e76985770024de42/src/lib/locale/base-config.js#L29) part of moment js config, which injects the [patterns](https://github.com/moment/moment/blob/e96809208c9d1b1bbe22d605e76985770024de42/src/lib/locale/relative.js#L1) used for formatting.

With my limited experience at the moment js, I still believe we can inject into the config already translated patterns. That will keep original approach simplicity and allows to skip extra library loading.

client/payment-details/summary/index.tsx
nagpai: Marking this as resolved

nagpai:
HI @kalessil  - Thanks for the valuable hint. I have made use of moment.updateLocale() , and WP i18n function this time for translation. - 93450af

Thanks to guidance from @mgascam  earlier, I ran a build and was able to find the relative time strings available for translation in the `/languages/woocommerce-payments.pot`. I hope that is the correct place to check?

client/payment-details/summary/index.tsx
```diff
@@ -171,6 +172,19 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 
 	const isFraudOutcomeReview = isOnHoldByFraudTools( charge, paymentIntent );
 
+	moment.updateLocale( 'en', {
```
kalessil, line 176: That looks much simpler now. Can we add a comment explaining that we inject the translated messages to align on used locales and not expand the bundle size by moment translations library?

client/payment-details/summary/index.tsx
```diff
@@ -3,11 +3,12 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { sprintf, __ } from '@wordpress/i18n';
```
kalessil: There are unused imported symbols in this file, can we clean them up please?

nagpai: Sorry about missing this. I have cleaned it up here d023fe1

client/payment-details/summary/index.tsx
nagpai: Thanks. I have added it here 
d023fe1

kalessil (approved):
I am approving the current approach. Leaving the final approval to @mgascam.

aheckler:
Looking at the screenshots shared above, I think we should make some slight changes to the language.

Screenshot taken on 2023-04-19 at 18 52 19 UTC@2x

"You need to capture this charge in 6 days" makes it sound like I must come back to my dashboard in 6 days and capture the charge then. Of course, that's not true. I can capture the charge at any time before it expires.

An improved version might be: "You must capture this charge within the next 6 days." This makes it clear that I don't have to capture at any _specific_ time, only within a certain time window.

nagpai:
@aheckler Thanks for the feedback. I have changed the text here : b154e9d

mgascam (approved):
LGTM!

PR #1134, opened 2020-11-30
Migrate from Travis to GitHub actions

kalessil:
#### Changes proposed in this Pull Request

* Migrates CI to GitHubActions
* It'll be a follow-up PR for E2E (runs, but all tests are failing, needs investigation) and PHP8 tests (MySQL connectivity is broken, needs investigation).

#### Testing instructions

* The PR should have 3 checks from GitHub actions and passing.

-------------------

- [x] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in the description ☝️)
- [x] Tested on mobile (or does not apply)

PR #5924, opened 2023-03-30
Bump WP and PHP versions and unpin composer version in GH workflows

### LINKED ISSUE #5333: Update PHP compatibility workflow
### Describe the bug
The workflow started to fail once composer released version 2.5.1: https://github.com/Automattic/woocommerce-payments/actions/runs/3764384197

image

We temporarily fixed it by pinning the version to 2.5.0 in #5332.

As mentioned in p1671788515519959/1671784565.257749-slack-C04D1QU76PK, what we really need to do is:
- bump min WP version to 5.9 in `phpcs-compat.xml.dist` file
- bump PHP version to 7.2 in `phpcs-compat.xml.dist` file
- Check if we need to update other places too?

Note: bumping min WP version in that file should be part of the [wcpay:bump-versions](https://github.com/woocommerce/woorelease-extensions/blob/trunk/class-woocommerce-payments-bump-versions.php) command, since we're already doing that for `phpcs.xml.dist` file. Ideally, we could also pass a PHP version as a parameter to bump the PHP version in every place we need to.

jbordonado:
Fixes #5333 

- Bump WP 'tested up to' version to `6.2` and minimum version to `6.0`
- Bump PHP minimum version to `7.2` (WC bumped their PHP min requirement to 7.2 during their [6.5 release](https://github.com/woocommerce/woocommerce/commit/819494f2ed3b7dabb95f58db3af35dc3b6ef4731))
- Unpin composer version (2.5.0) in the `PHP compatibility` workflow and `setup-repo` composite action
- Require PHP `>=7.2` for our project ([any versions of 7.2, ..., 8.2](https://semver.madewithlove.com/?package=php&constraint=%3E%3D7.2&stability=stable))

Bonus changes:
- Clean up the readme.txt and workflows a bit
- Use this opportunity to rename the L-2 variables for WC and WP to something more generic `XX_MIN_SUPPORTED_VERSION` (we should soon follow a  L-1 policy, like WC plans to do for their WP requirements)
- Note that in parallel I reworked the `wcpay:bump-versions` command in https://github.com/woocommerce/woorelease-extensions/pull/17, which takes into account all these changes and has been tested on this branch already.

#### Testing instructions

- All the PR checks should pass, including the PHP compatibility workflow
- WP minimum version is bumped everywhere
- PHP minimum version is bumped everywhere

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5924` or branch name `update/php-compatibility-workflow-and-php-version` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/php-compatibility-workflow-and-php-version) :rocket:

---

Build info: 

- Latest commit: f293c1744dc66f68243660b22d45fb7fc8de8bf5
- Build time: 2023-04-04 09:18:59 UTC

.github/workflows/compatibility.yml
```diff
@@ -4,15 +4,16 @@ on:
   pull_request
 
 env:
-  WC_L2_VERSION: '7.3.0'
-  WP_L2_VERSION: '5.9'
+  WC_MIN_VERSION: '7.3.0'
```
deepakpathania: Should we leave a comment here explaining that this is `L-2` for now? The previous variable names made it obvious.

.github/workflows/compatibility.yml
```diff
@@ -4,15 +4,16 @@ on:
...
+  WP_MIN_VERSION: '6.0'
+  PHP_MIN_VERSION: '7.2'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
-  generate-matrix:
+  generate-first-matrix:
```
deepakpathania: `first` and `second` in this context are not that clear, could we rename them to indicate the purpose instead?

.github/workflows/compatibility.yml
```diff
@@ -36,40 +37,47 @@ jobs:
...
+  generate-second-matrix:
+    name: "Generate the matrix for compatibility-woocommerce-beta dynamically"
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.generate_matrix.outputs.matrix }}
+    steps:
+      - name: "Generate matrix"
+        id: generate_matrix
+        run: |
+          PHP_VERSIONS=$( echo "[\"$PHP_MIN_VERSION\", \"8.0\", \"8.1\"]" )
```
deepakpathania: Will this require a manual change once we bump this further? Should we create the subsequent versions also dynamically or is it intended to be bumped like this to avoid breaking changes?

.github/workflows/php-lint-test.yml
```diff
@@ -6,8 +6,9 @@ on:
 
 env:
   WP_VERSION:        latest
-  WC_L2_VERSION: '7.3.0'  # the min supported version as per L-2 policy
+  WC_MIN_VERSION: '7.3.0'
```
deepakpathania: The comment should stay IMO for clarity.

deepakpathania (commented):
Left some comments.

.github/workflows/compatibility.yml
jbordonado: I don't know.

We are describing our version policy here: https://github.com/Automattic/woocommerce-payments#version-support-policy
And we are not supposed to update these variables manually, as we have a dedicated `wcpay:bump-versions` command for this.

I can add a comment linking to the version policy, but that seems redundant if I do that in every concerned file.

WDYT?

.github/workflows/compatibility.yml
jbordonado: I considered that, since each matrix job is immediately followed by the job that needs it, and the two jobs (`woocommerce-compatibility` and  `compatibility-woocommerce-beta`) are doing the same thing, we could not bother finding fancy names for the matrix generation. Anyway, I will rename them to include the job it will be used in, it can only improve things.

On that workflow in particular, I understood from the comments that there are 2 different jobs because we want to allow a part of the testing matrix to fail. But I fail to see the goal here. If the second job fails, the whole workflow will be marked as failed anyway. Do you have more history on this decision?
It seems like we could just have these two jobs grouped into one, and adjust the matrix a little to run the same combinations as before... Not really important but just curious on this

.github/workflows/compatibility.yml
jbordonado: I don't have enough history here, I'm not sure how we previously bumped the "max versions" for PHP.
I guess there would be concerns about breaking changes and compatibility for sure, so it doesn't seem wise to bump it at the same time as we bump the min version.

We can plan another task to allow bumping the `max` and `max - 1` PHP versions based on a new parameter `php_max_version`, but doesn't seem that urgent

.github/workflows/php-lint-test.yml
jbordonado: Similar thoughts as https://github.com/Automattic/woocommerce-payments/pull/5924#discussion_r1155636653

Will update based on your reply

.github/workflows/compatibility.yml
deepakpathania: Um, I believe we could just rename it to something like `WC_MIN_SUPPORTED_VERSION` and that could be a little more explicit?

.github/workflows/compatibility.yml
deepakpathania: Fair.

.github/workflows/php-lint-test.yml
deepakpathania: Replied on the other issue.

deepakpathania (commented):
Done with another round of review.

.github/workflows/compatibility.yml
deepakpathania: I don't have the relevant context either but I just noticed that in the original PR, we had `continue-on-error` set as `true` for the second matrix which was then maybe removed later, [ref](https://github.com/Automattic/woocommerce-payments/pull/1134/files#diff-0ff6bcc45c7c09bfcb5419c83559f190add597ec3b82ad238ddb1d6e43e9517eR82). 

So maybe that was the original thought to allow failures but wasn't considered in some follow up.

.github/workflows/compatibility.yml
jbordonado: Yeah that works better, done!

Also updated accordingly the parallel PR in the extensions repo

.github/workflows/compatibility.yml
jbordonado: Yeah, it was done temporarily to avoid failing due to PHP 8 build failures.
I think we're safe to regroup the jobs now since we've been able to run CI jobs with PHP 8 for a long time. The next time we work on that file, it'd be good to keep it in mind

deepakpathania (approved):
Changes LGTM now. ✅

htdat (approved):
🚢 Awesome work, thanks for working on this issue @jbordonado. 

Anyway, I notice a pattern of using `composer self-update 2.0.6` in both bash and GitHub yml files such as the following (even in the `check-changelog.yml` made by me). I think it's because we used to be stuck with PHP 7.0. It would be nice if this can be addressed in the future too. 

https://github.com/Automattic/woocommerce-payments/blob/ae116e1808b36910a2bb58bea847ad148da4bc18/.github/workflows/php-lint-test.yml#L35

https://github.com/Automattic/woocommerce-payments/blob/ae116e1808b36910a2bb58bea847ad148da4bc18/bin/run-ci-tests.bash#L11

jbordonado:
> 🚢 Awesome work, thanks for working on this issue @jbordonado.
> 
> Anyway, I notice a pattern of using `composer self-update 2.0.6` in both bash and GitHub yml files such as the following (even in the `check-changelog.yml` made by me). I think it's because we used to be stuck with PHP 7.0. It would be nice if this can be addressed in the future too.
> 
> https://github.com/Automattic/woocommerce-payments/blob/ae116e1808b36910a2bb58bea847ad148da4bc18/.github/workflows/php-lint-test.yml#L35
> 
> https://github.com/Automattic/woocommerce-payments/blob/ae116e1808b36910a2bb58bea847ad148da4bc18/bin/run-ci-tests.bash#L11

Thanks for double-checking things, much appreciated 🙏 

Yep, I also noticed this, and wondered what is the reason.
I'll make a note of it, good feedback!

PR #5927, opened 2023-03-31
Update mock_wcpay_request to properly return the mock response

htdat:
Fixes #

I noticed this issue when trying to switch the approach for my own question here https://github.com/Automattic/woocommerce-payments/pull/5879/files#r1153037010

The current `mock_wcpay_request` method is just mocking the response for `WC_Payments_API_Client::send_request`. 

To use the response from this request, I still need to mock `format_response` even my request concrete class does not extend `format_response`. 

With this change, I just use: 

```php
$mock_response = [
   // something here.
]; 
$this->wcpay_mock_request( Concrete_Request::class, 1, null, $mock_response);
```

Otherwise, I need to do something like, which is both verbose and not what I am expecting from reading [the doc](https://github.com/Automattic/woocommerce-payments/blob/547a0d2bbbfc3f505077cf051c8c669b71eb34a9/includes/core/server/CONTRIBUTING.md#testing) as `$mock_response` does not give me the expected result when calling `mock_wcpay_request`:

```php
$mock_response = [
	// something here.
];
$request = $this->wcpay_mock_request( Concrete_Request::class, 1);
$request->expect( 'format_response' )
	->willReturn ( new Response( $mock_response ) ); 
``` 

Also the doc mention that I just need to mock `format_response` in case I have custom `format_response` method.

#### Testing instructions

* No specific instruction for this, ideally no current tests fails. 
* Focus on my reasoning and see if it makes sense to change. 
* It's OK to discard this PR if we have a better way to get the same thing. Note that:  I am lacking of the full knowledge from the phase 1 of WCPay as Platform.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5927` or branch name `update/mock_wcpay_request-add-response` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/mock_wcpay_request-add-response) :rocket:

---

Build info: 

- Latest commit: 68665e54260c61bfb6652bb692163669db606aaf
- Build time: 2023-06-19 09:48:34 UTC

htdat:
FYI. I've just updated the testing instructions per @jessy-p's suggestion.

tests/WCPAY_UnitTestCase.php
```diff
@@ -64,6 +65,13 @@ function( $argument ) use ( $request_class, $request ) {
 			)
 			->willReturn( $response );
```
zmaglica, line 66: I like the idea here, and we can keep it to make it more resilient to changes, but is there a reason why Request Class won't extend parent abstract request class?

RadoslavGeorgiev (changes requested):
@htdat could you please provide a couple of examples featuring this change?

The base request class [has an implementation of `format_response`](https://github.com/Automattic/woocommerce-payments/blob/update/mock_wcpay_request-add-response/includes/core/server/class-request.php#L271-L273), so AFAICT the `if` will never get entered.

Some context: In the initial batch of requests, we simulated `format_response` because  intent-related requests would try to deserialize the intent object. The assumption was that if a `$response` was provided to `mock_wcpay_request`, we would return it from the `API_Client`'s `send_request` method, and it would be returned as usual. Is this not the case?

htdat:
@RadoslavGeorgiev and @zmaglica - Thanks for your reviews and feedback. They're very helpful for me! 

> The base request class [has an implementation of `format_response`](https://github.com/Automattic/woocommerce-payments/blob/update/mock_wcpay_request-add-response/includes/core/server/class-request.php#L271-L273), so AFAICT the `if` will never get entered.

🤦 This is a valid point! I missed it somehow.

---- 

> The assumption was that if a `$response` was provided to `mock_wcpay_request`, we would return it from the `API_Client`'s `send_request` method, and it would be returned as usual. Is this not the case?

Actually... it's not the case, and I am trying to fix it here. 
With `Get_Account` request class introduced in https://github.com/Automattic/woocommerce-payments/pull/5879, I tried to follow up [Standard mock](https://github.com/Automattic/woocommerce-payments/blob/14b49378b81972dc813777b02c905595d1a136ba/includes/core/server/CONTRIBUTING.md#testing) like the following:

```php
		$this->mock_wcpay_request( Get_Account::class, 1, null,
			[
				'account_id'               => 'acc_test',
				'live_publishable_key'     => 'pk_test_',
				'test_publishable_key'     => 'pk_live_',
				'has_pending_requirements' => true,
				'current_deadline'         => 12345,
				'is_live'                  => true,
			]
		);
```

Here is [the diff](https://gist.github.com/htdat/f143ea225be411e4208f7f4d328397b2#file-test-pr-5927-diff) to apply against the `develop` branch, run `npm run test:php -- --filter WC_Payments_Account_Test`, and get this error:


> WC_Payments_Account_Test::test_maybe_redirect_to_onboarding_account_connected_with_mock_wcpay_request_modification
> Error: Call to a member function to_array() on null
> 
> --- Complete result: https://gist.github.com/htdat/f143ea225be411e4208f7f4d328397b2#file-php-test-result-txt

That's because `format_response` of the core Request class is NOT mocked, or told to use the implementation in the real class. Therefore, when [`send`](https://github.com/Automattic/woocommerce-payments/blob/14b49378b81972dc813777b02c905595d1a136ba/includes/core/server/class-request.php#L232-L236) is called, PHPUnit is just trying to mock it and return `null`. That causes the error above. 

A _simple fix_ would use `setMethodsExcept( ['format_response'] )` to tell PHPUnit using the original implementation of `format_response` when building [this `$request` mock](https://github.com/Automattic/woocommerce-payments/blob/14b49378b81972dc813777b02c905595d1a136ba/tests/WCPAY_UnitTestCase.php#L51-L53). With that, it works well for the new test case in my diff... However, all of the current tests with mock `format_response` will cause errors because they are not expected to mock `format_response`. 

So... with the fact that, none of current tests are using three latest parameters `$response, $api_client_mock, $http_mock`, I think we can apply the fix in my current PR, i.e. just mocking the expected response when `$response` is provided - commit db820b2c22c9725a3d1911ef1cbfc3becc7473e3

--- 

> could you please provide a couple of examples featuring this change?

Besides correcting the expectation mentioned in the doc, it would reduce the verbosity of writing tests a bit. For example, with the current limitation (without this PR), I need to write the same mock above like this: 
https://github.com/Automattic/woocommerce-payments/blob/14b49378b81972dc813777b02c905595d1a136ba/tests/unit/test-class-wc-payments-account.php#L173-L187

tests/WCPAY_UnitTestCase.php
htdat: In many cases, especially for `GET` requests, the consumers just expect an array rather than an object like intentions. Some examples are in my recent PRs: 
- Get_Account https://github.com/Automattic/woocommerce-payments/pull/5879,
- Get_Account_Login_Data and Get_Account_Capital_Link https://github.com/Automattic/woocommerce-payments/pull/5948.

htdat:
Anyway, ready to review again.

naman03malhotra:
@htdat I can see this has been added to the sprint automatically. Is there an issue to track for this? Are you planning to continue working on this during this sprint, maybe during the cooldown?

naman03malhotra:
@jessy-p can you please take a look when you get some time?

jessy-p:
Able to replicate the issue with the provided diff and the fix on this branch does fix the issue. 👍

htdat:
Thanks a lot @jessy-p! 

@RadoslavGeorgiev - could you have a look at this issue when you have time too? As I mentioned above, it's OK to discard my proposal here. If so, we might change [the current MD doc](https://github.com/Automattic/woocommerce-payments/blob/547a0d2bbbfc3f505077cf051c8c669b71eb34a9/includes/core/server/CONTRIBUTING.md#testing) a bit to reflect the correctness, i.e. in the `Standard mock` section, `$mock_response` would be a Response object rather than a simple array.

tests/WCPAY_UnitTestCase.php
```diff
@@ -61,8 +62,14 @@ function( $argument ) use ( $request_class, $request ) {
...
 				)
-			)
-			->willReturn( $response );
+			);
+
+		if ( ! is_null( $response ) ) {
+			$request
+				->expects( $this->exactly( $total_api_calls ) )
+				->method( 'format_response' )
+				->willReturn( new Response( $response ) );
```
RadoslavGeorgiev:
```suggestion
				->willReturn( $response instanceof Response ? $response : new Response( $response ) );
```

How about checking whether the object is already a response? This way we'd allow mocks of requests, which have custom responses.

RadoslavGeorgiev (approved):
Once again, thanks for working on this Dat! 🙇 

Now I understand the idea. Last time I was a bit misguided because of the check whether `format_response()` exists. As far as I'm concerned, this PR works and can be shipped.

I have one suggestion, but you are welcome to skip it.

tests/WCPAY_UnitTestCase.php
htdat: Nice addition, I've committed your suggestion!

PR #5937, opened 2023-03-31
Express checkout button display class

### LINKED ISSUE #5472: Refactor payment method buttons and separators rendering
#### Why is the refactoring needed?
There was a fix implemented [here](https://github.com/Automattic/woocommerce-payments/pull/5399) and [here](https://github.com/Automattic/woocommerce-payments/pull/5451) that resolves the issue of displaying multiple separators on the checkout, cart, product details pages. 

After the fix, the button and separator displaying implementations are separated from each other. With that, a few drawbacks are currently visible, and these should be resolved once the refactoring is completed: 
1. In a nutshell, the logic deciding whether we should render the button and the separator is duplicated (see `WC_Payments::display_express_checkout_separator_if_necessary` for the separator and `WC_Payments_*_Button_Handler::init` for `Platform_Checkout`, `Payment_Request` and `Stripe_Link` accordingly. This will lead to situations when e.g. the WooPay button logic is changed in the future, we might easily forget to update the logic in the separator rendering class, which will immediately result in displaying the separator alone when the button is disabled.
2. Separator rendering logic is located in `class-wc-payments.php`, which doesn't seem to be its direct responsibility.
3. There's limited test coverage due to static methods in the `class-wc-payments.php` class.

Ideally, we could create a new class that will merge and orchestrate all the logic from both button and separator rendering in one place and test it out properly.

#### Acceptance criteria
1. Button and separator displaying logic is merged into one place.
2. [The bug with the multiple separators](https://github.com/Automattic/woocommerce-payments/issues/5339) doesn't appear (no regression after the changes within this issue).
3. All the logic is covered with tests.

mdmoore:

The goal of this PR is to combine the logic of the express checkout button display and the separator that follows the button. The logic is currently duplicated across multiple files and is prone to errors that could lead to undesired display issues. This centralizes the logic into a new class, `WC_Payments_Express_Checkout_Button_Display_Handler`. This class will handle initialization and logic for any express payment method buttons. 

This class doesn't add a new class to help prevent redundancy when adding new express payment methods, as mentioned in 1545-gh-Automattic/woopay. 

Closes #5472 

#### Testing instructions

The goal of testing should be to ensure that WooPay buttons and Payment Request Buttons all continue to function as expected and that the `-or-` separator is never duplicated or displaying when it shouldn't be.

1. Enable WooPay buttons and Payment Requests
2. Check that buttons are appearing in the appropriate locations(Product page, Cart page, Checkout page).
3. In the WCPay settings, toggle on/off button display on different pages and ensure those settings are respected.
4. Test that payments via each button are working as expected. This guide works well for testing Apple Pay: paJDYF-1Ys-p2 . For WooPay, a successful redirect to WooPay is enough.

-------------------

**Post merge**

- [] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5937` or branch name `add/5472-express-checkout-button-class` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5472-express-checkout-button-class) :rocket:

---

Build info: 

- Latest commit: 96d9d85bcb72a977185844280c16b42d6d9fce14
- Build time: 2023-04-28 18:08:17 UTC

includes/class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -0,0 +1,102 @@
...
+		$this->gateway                          = $gateway;
+		$this->payment_request_button_handler   = $payment_request_button_handler;
+		$this->platform_checkout_button_handler = $platform_checkout_button_handler;
+
+		$this->platform_checkout_button_handler->init();
+		$this->payment_request_button_handler->init();
+
+		$is_woopay_express_checkout_enabled = WC_Payments_Features::is_woopay_express_checkout_enabled();
+
+		if ( $is_woopay_express_checkout_enabled || $this->gateway->get_option( 'payment_request' ) ) {
```
FangedParakeet:
```suggestion
		if ( $is_woopay_express_checkout_enabled || 'yes' === $this->gateway->get_option( 'payment_request' ) ) {
```
AFAIA, the value of `$this->gateway->get_option( 'payment_request' )` can be either `yes` or `no`. Presumably, we are checking for more than the presence of this setting, but rather that it has also been enabled as well. The same goes for the identical check below on line 60, but correct me, if I'm misunderstanding something here!

includes/class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	/**
+	 * Display express checkout separator only when express buttons are displayed.
+	 *
+	 * @return void
+	 */
+	public function display_express_checkout_separator_if_necessary() {
+		$woopay          = $this->platform_checkout_button_handler->is_woopay_enabled() && $this->platform_checkout_button_handler->should_show_platform_checkout_button();
```
FangedParakeet:
```suggestion
		$woopay          = $this->platform_checkout_button_handler->should_show_platform_checkout_button();
```
Looks like [`should_show_platform_checkout_button()` will be `false` when `is_woopay_enabled()` is `false`](should_show_platform_checkout_button), so this first check appears redundant here.

includes/class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -0,0 +1,102 @@
...
+		$payment_request = $this->payment_request_button_handler->should_show_payment_request_button();
+		$should_hide     = $payment_request && ! $woopay;
+		if ( $woopay || $payment_request ) {
```
FangedParakeet:
```suggestion
		if ( $should_show_woopay || $should_show_payment_request ) {
```
This is a really minor quibble and one you're welcome to ignore, but I'd probably use more descriptive variable names here. 🤷

FangedParakeet:
```suggestion
		if ( ! $should_hide ) {
```
Actually, shouldn't we just check for this instead? Is there any reason to add this HTML to the DOM, if we are going to be hiding it?

includes/class-wc-payments.php
```diff
@@ -487,10 +495,13 @@ public static function init() {
 
 		self::maybe_register_platform_checkout_hooks();
 
-		// Payment Request and Apple Pay.
-		self::$payment_request_button_handler   = new WC_Payments_Payment_Request_Button_Handler( self::$account, self::get_gateway() );
+		self::$apple_pay_registration = new WC_Payments_Apple_Pay_Registration( self::$api_client, self::$account, self::get_gateway() );
+
+		self::$payment_request_button_handler = new WC_Payments_Payment_Request_Button_Handler( self::$account, self::get_gateway() );
```
```suggestion
		$payment_request_button_handler = new WC_Payments_Payment_Request_Button_Handler( self::$account, self::get_gateway() );
```
Do `$payment_request_button_handler` and `$platform_checkout_button_handler` still need to be static variables? I can't see them used anywhere else in this class and it looks like the `WC_Payments_Express_Checkout_Button_Display_Handler` now handles functionality once encapsulated by these other two classes.

tests/unit/test-class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -0,0 +1,286 @@
...
+			->method( 'should_show_platform_checkout_button' )
+			->willReturn( true );
+
+		$this->mock_payment_request_button_handler
+			->method( 'should_show_payment_request_button' )
+			->willReturn( true );
+
+		$this->express_checkout_button_display_handler->display_express_checkout_separator_if_necessary();
+
+		$this->assertStringContainsString( 'wcpay-payment-request-button-separator', ob_get_contents() );
```
FangedParakeet, line 166:
```suggestion
		ob_start();
		$this->assertStringContainsString( 'wcpay-payment-request-button-separator', ob_get_contents() );
		ob_end_clean();
```
Just a heads up, I think you need to manually open, close, and clean this buffer to ensure other tests are not affected. This applies for everywhere else where you've used `ob_get_contents()` as well. Let me know if I'm misunderstanding anything though!

FangedParakeet (commented):
Implementation here looks really neat and tidy--and I don't really have any issues with the design. Just left a few comments on minor details: please let me know your thoughts and then I'll give this another run through. Cheers! 🙏

includes/class-wc-payments-express-checkout-button-display-handler.php
mdmoore: >the value of $this->gateway->get_option( 'payment_request' ) can be either yes or no

That's correct! Nice catch.

includes/class-wc-payments-express-checkout-button-display-handler.php
mdmoore: >$should_show_woopay || $should_show_payment_request

I'm fine with these name changes. Thanks for the suggestion!

>Actually, shouldn't we just check for this instead? Is there any reason to add this HTML to the DOM, if we are going to be hiding it?

The `$should_hide` flag is only true when Payment Request is enabled and WooPay express checkout is not. We do indeed have to add this HTML in this case, but we hide it by default and show it only when the browser has a Google Pay or Apple Pay payment method stored. There's more on this [here](https://github.com/Automattic/woocommerce-payments/pull/5399#discussion_r1073633776).

includes/class-wc-payments.php
mdmoore: Good point! There's no reason for `$payment_request_button_handler` and `$platform_checkout_button_handler` to remain static variables since they're just being passed into the constructor of `WC_Payments_Express_Checkout_Button_Display_Handler` and not requiring any direct access via the class.

tests/unit/test-class-wc-payments-express-checkout-button-display-handler.php
mdmoore: I do believe you're correct. Thanks for pointing it out!

mdmoore:
Thanks @FangedParakeet for the review. I've addressed everything now if you'd like to give this another look at your convenience! 

The [Environment - WC beta (beta, latest, latest, 7.2) ](https://github.com/Automattic/woocommerce-payments/actions/runs/4749834067/jobs/8437461777?pr=5937) job is failing for all PRs right now due to WC 7.7 requiring PHP 7.3 while PHP 7.2 is being installed.

FangedParakeet (approved):
LGTM! Nice one for answering all my feedback and finding such an elegant solution to this rather technical issue that has kept our brows thoroughly ruffled for quite a long time. Very happy to have such a neat and maintainable solution to this problematic issue at last.

:shipit:

PR #5942, opened 2023-04-01
Add checks to prevent interacting with multi-currency if it is disabled

eduardoumpierre:
Fixes Automattic/woocommerce-payments-server#3209

#### Testing instructions

* Don't checkout to this branch yet
* Run `npm run start`
* Go to **WooCommerce > Settings** and change the store currency for any other than USD, and save changes
* Go to **Payments > Settings**
* Click on "Advanced settings", uncheck "Enable Multi-Currency", click on "Save changes"
* Refresh the page
* You should see a notice "Error retrieving currencies."
* Open the standard modal
* It should contain "An order exceeds the equivalent of $1,000.00 USD in undefined."
* **Checkout to this branch**
* Refresh the page
* There should be no notice
* Open the same modal – it should contain "An order exceeds the equivalent of $1,000.00 USD"
* The unit tests should pass

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5942` or branch name `fix/3209-multi-currency` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/3209-multi-currency) :rocket:

---

Build info: 

- Latest commit: 212774b7af21b909b24d6de4402b066240792cbe
- Build time: 2023-04-03 19:16:47 UTC

client/data/multi-currency/selectors.js
```diff
@@ -17,7 +17,7 @@ const getMultiCurrencyState = ( state ) => {
 };
 
 export const getCurrencies = ( state ) => {
-	return getMultiCurrencyState( state ).currencies || [];
+	return getMultiCurrencyState( state ).currencies || {};
```
tpaksu, line 20: Did you double check that this won't break the MC settings UI?

eduardoumpierre: It shouldn't as the actual data structure is an object, but I'll check that!

eduardoumpierre: It works because both will result in `undefined` when trying to access a key that doesn't exist.

```js
const a = [];
const b = {};

a.available // undefined
b.available // undefined
```

I'll keep the change because it makes more sense to be an object than an array. Is that ok for you?

tpaksu (approved):
Code is clean and understandable. The extra tests you've written shouldn't be needed for these changes, but it's weird that sometimes the coverage need shows as 60%, as it usually was 92%. The automated tests pass, the manual testing went well too, I've also checked the data passed to the server shows 1000$ when the presets are saved, and they were saved correctly. Thanks for working on this, LGTM.

PR #5959, opened 2023-04-04
Migrate update_account API to the core Request class

htdat:
Fixes #
A part of https://github.com/Automattic/woocommerce-payments-meta/issues/87

**ℹ️ NOTICE** 

- This PR is based on https://github.com/Automattic/woocommerce-payments/pull/5879's PR branch. Hence, 5879 should be merged first though this PR can be reviewed at any time.

- Migrate update_account API to the core class. 
- Add docs and tests.

Not directly relate to this PR:
 
- c93f8b70c7eccf4d599581451470f2523410b83a - Remove const ACCOUNT_FIELDS_TO_UPDATE, and add comment for ACCOUNT_SETTINGS_MAPPING to reduce the duplication.

#### Testing instructions

* Go to WP Admin > Payments > Settings, and update some settings such as email address, phone, statement descriptor, Payout/Deposit options.

-------------------

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-payments-settings-controller.php
+++ b/includes/admin/class-wc-rest-payments-settings-controller.php
@@ -610,7 +610,7 @@ class WC_REST_Payments_Settings_Controller extends WC_Payments_REST_Controller {
 	 */
 	private function update_account( WP_REST_Request $request ) {
 		$updated_fields_callback = function ( $value, string $key ) {
-			return in_array( $key, array_keys( WC_Payment_Gateway_WCPay::ACCOUNT_SETTINGS_MAPPING ), true ) &&
+			return array_key_exists( $key, WC_Payment_Gateway_WCPay::ACCOUNT_SETTINGS_MAPPING ) &&
 				$this->wcpay_gateway->get_option( $key ) !== $value;
 		};
 		// Filter out fields that are unchanged or not in the list of fields to update.
--- a/includes/admin/class-wc-rest-payments-settings-controller.php
+++ b/includes/admin/class-wc-rest-payments-settings-controller.php
@@ -13,23 +13,6 @@ defined( 'ABSPATH' ) || exit;
  * REST controller for settings.
  */
 class WC_REST_Payments_Settings_Controller extends WC_Payments_REST_Controller {
-
-	const ACCOUNT_FIELDS_TO_UPDATE = [
-		'account_statement_descriptor',
-		'account_business_name',
-		'account_business_url',
-		'account_business_support_address',
-		'account_business_support_email',
-		'account_business_support_phone',
-		'account_branding_logo',
-		'account_branding_icon',
-		'account_branding_primary_color',
-		'account_branding_secondary_color',
-		'deposit_schedule_interval',
-		'deposit_schedule_monthly_anchor',
-		'deposit_schedule_weekly_anchor',
-	];
-
 	/**
 	 * Endpoint path.
 	 *
@@ -627,7 +610,7 @@ class WC_REST_Payments_Settings_Controller extends WC_Payments_REST_Controller {
 	 */
 	private function update_account( WP_REST_Request $request ) {
 		$updated_fields_callback = function ( $value, string $key ) {
-			return in_array( $key, static::ACCOUNT_FIELDS_TO_UPDATE, true ) &&
+			return in_array( $key, array_keys( WC_Payment_Gateway_WCPay::ACCOUNT_SETTINGS_MAPPING ), true ) &&
 				$this->wcpay_gateway->get_option( $key ) !== $value;
 		};
 		// Filter out fields that are unchanged or not in the list of fields to update.
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -1280,12 +1280,7 @@ class WC_Payments_Account {
 				return;
 			}
 
-			$request = Update_Account::create();
-			// Dynamically set the request parameters.
-			foreach ( $stripe_account_settings as $param => $value ) {
-				call_user_func( [ $request, "set_{$param}" ], $value );
-			}
-
+			$request         = Update_Account::from_account_settings( $stripe_account_settings );
 			$response        = $request->send( 'wcpay_update_account_settings' );
 			$updated_account = $response->to_array();
 
@@ -1325,12 +1320,14 @@ class WC_Payments_Account {
 	public function possibly_update_wcpay_account_locale( $option_name, $old_value, $new_value ) {
 		if ( 'WPLANG' === $option_name && $this->is_stripe_connected() ) {
 			try {
-				$locale  = $new_value ?? 'en_US';
-				$request = Update_Account::create();
-				$request->set_locale( $locale );
+				$account_settings = [
+					'locale' => $new_value ? $new_value : 'en_US',
+				];
 
+				$request         = Update_Account::from_account_settings( $account_settings );
 				$response        = $request->send( 'wcpay_update_account_settings' );
 				$updated_account = $response->to_array();
+
 				$this->database_cache->add( Database_Cache::ACCOUNT_KEY, $updated_account );
 			} catch ( Exception $e ) {
 				Logger::error( __( 'Failed to update Account locale. ', 'woocommerce-payments' ) . $e );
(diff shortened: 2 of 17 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `5959` or branch name `dev/migrate-update-account-data` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=dev/migrate-update-account-data) :rocket:

---

Build info: 

- Latest commit: 2d495342fb66797f17e118affd53e37b207f94eb
- Build time: 2023-04-14 09:03:06 UTC

includes/class-wc-payments-account.php
```diff
@@ -1230,7 +1231,16 @@ public function update_stripe_account( $stripe_account_settings ) {
 				Logger::info( 'Skip updating account settings. Nothing is changed.' );
 				return;
 			}
-			$updated_account = $this->payments_api_client->update_account( $stripe_account_settings );
+
+			$request = Update_Account::create();
+			// Dynamically set the request parameters.
+			foreach ( $stripe_account_settings as $param => $value ) {
```
zmaglica: Instead of looping, you can create a static function and fill the request from that function. In this example, the code can trigger an exception if one of the param doesn't have a setter function. You can take a look at the Paginated request class and see how it is done there.

includes/class-wc-payments-account.php
```diff
@@ -1267,10 +1277,12 @@ private function settings_changed( $changes = [] ) {
 	public function possibly_update_wcpay_account_locale( $option_name, $old_value, $new_value ) {
 		if ( 'WPLANG' === $option_name && $this->is_stripe_connected() ) {
 			try {
-				$account_settings = [
-					'locale' => $new_value ? $new_value : 'en_US',
-				];
-				$updated_account  = $this->payments_api_client->update_account( $account_settings );
+				$locale  = $new_value ?? 'en_US';
+				$request = Update_Account::create();
+				$request->set_locale( $locale );
```
zmaglica: Nitpick. You don't have to define $locale variable. You can move definition as a function call argument.

includes/class-wc-payments-account.php
htdat: Another great feedback with the idea of creating a static function! Addressed at 9be5c979d8f967cc7c0308637035cda37d1531d3

includes/class-wc-payments-account.php
htdat: No longer relevant. Closing this.

htdat:
Addressed feedback https://github.com/Automattic/woocommerce-payments/pull/5959#discussion_r1161571544 - ready for another review

includes/core/server/request/class-update-account.php
```diff
@@ -0,0 +1,234 @@
...
+				'wcpay_core_invalid_request_parameter_account_settings_empty'
+			);
+		}
+
+		$wcpay_request = static::create();
+
+		foreach ( $account_settings as $param_name => $value ) {
+			$param_setter = 'set_' . $param_name;
+
+			if ( ! method_exists( static::class, $param_setter ) ) {
```
zmaglica: I wouldn't throw an exception here. I would simply skip it and continue with the loop. In case someone adds something to the account settings, this API endpoint would crash.

htdat: Make sense, addressed bcf2bc6d20bbfc277df710755a7dd8193084d162

htdat:
Ready to review again.

tests/unit/core/server/request/test-class-core-update-account-request.php
```diff
@@ -0,0 +1,79 @@
...
+	public function test_from_account_settings_with_empty_input_will_throw_exception() {
+		$this->expectException( Invalid_Request_Parameter_Exception::class );
+		$this->expectExceptionMessage( 'No account settings provided' );
+
+		Update_Account::from_account_settings( [] );
+	}
+
+	public function test_from_account_settings_with_non_existing_param_continue_to_run() {
+		$none_existing_param = 'none_existing_param';
+		$this->expectException( Invalid_Request_Parameter_Exception::class );
```
zmaglica: Nitpick: Move this assert exceptions after calling `Update_Account::from_account_settings` to be sure that exception will be thrown when `$request->get_param( $none_existing_param )` is called

includes/admin/class-wc-rest-payments-settings-controller.php
```diff
@@ -627,7 +610,7 @@ private function update_is_wcpay_subscriptions_enabled( WP_REST_Request $request
 	 */
 	private function update_account( WP_REST_Request $request ) {
 		$updated_fields_callback = function ( $value, string $key ) {
-			return in_array( $key, static::ACCOUNT_FIELDS_TO_UPDATE, true ) &&
+			return in_array( $key, array_keys( WC_Payment_Gateway_WCPay::ACCOUNT_SETTINGS_MAPPING ), true ) &&
```
zmaglica: Can you maybe use [array_key_exists](https://www.php.net/manual/en/function.array-key-exists.php) here and avoid calling in_array and array_keys?
The code would look something like this 

`return array_key_exists( $key, WC_Payment_Gateway_WCPay::ACCOUNT_SETTINGS_MAPPING ) &&`

includes/core/server/request/class-update-account.php
```diff
@@ -0,0 +1,227 @@
...
+	public function get_method(): string {
+		return 'POST';
+	}
+
+	/**
+	 * If true, the request will be signed with the user token rather than blog token.
+	 *
+	 * @return bool
+	 */
+	public function should_use_user_token(): bool {
```
zmaglica, line 39: Perhaps this function name could be a bit misleading. Either pass an argument and make it default to true or name the function to something like 'use_user_token'

includes/core/server/request/class-update-account.php
```diff
@@ -0,0 +1,227 @@
...
+				'wcpay_core_invalid_request_parameter_account_settings_empty'
+			);
+		}
+
+		$wcpay_request = static::create();
+
+		foreach ( $account_settings as $param_name => $value ) {
+			$param_setter = 'set_' . $param_name;
+
+			if ( ! method_exists( static::class, $param_setter ) ) {
```
zmaglica: Really small nitpick. You don't have to fix it, but to save a couple of lines of code, you could do something like this:

```
if ( method_exists( static::class, $param_setter ) ) {
       $wcpay_request->{$param_setter}( $value );
}
```

You are already checking if the method exists, and you don't have to continue the loop, since it will continue by the default. You can change logic and call the setter function if the method exists, and the loop will continue as it should.

includes/core/server/request/class-update-account.php
htdat: This is to override the parent `Request::should_use_user_token` method. If needed, we can open a new PR to fix it for the Request class. But anyway, it should not be addressed in this PR.

includes/core/server/request/class-update-account.php
htdat: Good point, addressed 5ae1bf0fcd01d01430b569e663675a7e717ed9f7

includes/admin/class-wc-rest-payments-settings-controller.php
htdat: Really good point! Addressed 9c2ecbe91e721026d9e28aace54c23b598473214

tests/unit/core/server/request/test-class-core-update-account-request.php
htdat: Good point. Addressed 03ee2e8542347e6dcb3f4c60e19e3d05a68d6541

htdat:
@zmaglica - thanks for a lot of great feedback. I've addressed them and make this PR ready to review again.

includes/core/server/request/class-update-account.php
zmaglica: Yes, you are right. Feel free to ignore my comment.

zmaglica (approved):

PR #5804, opened 2023-03-17
Expand Stripe Link test coverage

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5338

#### Changes proposed in this Pull Request
This PR expands the E2E coverage for Stripe Link-related functionality. The flow now runs the shortcode checkout with UPE/split UPE, and the blocks checkout tests Link flow with UPE enabled. 

One more unit test has been added to cover the implementation in `class-wc-payments.php`.

#### Testing instructions
1. Run e2e tests locally and ensure they are green.
2. Re-run the GH Actions flow for this PR and ensure the tests are green. 

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #6008, opened 2023-04-11
Revert "Expand Stripe Link test coverage"

timur27:
Reverts Automattic/woocommerce-payments#5804

When run concurrently by multiple jobs, Stripe Link tests may fail because of the Rate Limiting. We are getting `429`, and re-tries don't help, as it seems that the job that fails is blocked for a longer period. There was an attempt to randomly delay each test suite (https://github.com/Automattic/woocommerce-payments/pull/6006) so that we could reduce the probability of a failure in the most efficient way, but after running the suite in 3 concurrent jobs, there still were some failures. 

[Example of failure.](https://github.com/Automattic/woocommerce-payments/actions/runs/4670590467/jobs/8272086277)

The possibility of concurrent test execution for Stripe Link in the test mode (without any mocks etc.) should be reviewed. In the meantime, to avoid `develop` from failing because of rate limiting, this PR could be reverted.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6008` or branch name `revert-5804-improvement/test-coverage-for-stripe-link` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=revert-5804-improvement/test-coverage-for-stripe-link) :rocket:

---

Build info: 

- Latest commit: 379def01a23c79c3d662403ac5aa1d75b63cb3ad
- Build time: 2023-04-12 07:32:17 UTC

bborman22 (approved):
The changes look good in the sense that it is a direct revert of the previous PR. This would get us back to the initial state successfully and I can see that all the GH Actions have passed including E2E tests now again.

My one question would be if it's necessary to revert all the changes or if it would be possible to just exclude the Stripe Link tests until we can get them passing? I saw you mentioned that there might not be a good way to E2E test Stripe Link long term due to this concurrent tests and  rate limiting, so maybe a revert is best (it's also the cleanest to get our pipeline green), but wanted to ask the question in case there is a way we could save most of this work without interfering with the pipeline for others and then makes any attempt at a fix a little bit easier in the future?

Otherwise, this revert PR looks good to me!

timur27:
@bborman22 I'm currently discussing the purpose of E2E tests for Stripe Link with a broader audience, and I think they may eventually be removed. It will be also much faster and cleaner (as you mentioned) to revert the changes fully, with immediate effect. Having this in mind, I would like to merge this PR, reopen https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/5338 and review the approach, find other ways to cover this functionality or worst case bring back the E2E tests with some slight modifications.

PR #6032, opened 2023-04-13
Convert the fraud protection settings files to TypeScript

### LINKED ISSUE #6031: Convert the fraud protection settings files to TypeScript
### Description

The [fraud protection settings](https://github.com/Automattic/woocommerce-payments/tree/develop/client/settings/fraud-protection) files are currently written in JavaScript. This code should be converted to TypeScript. Context here: paJDYF-6MV-p2

### Acceptance Criteria 

- JavaScript is rewritten to TypeScript.
- All tests pass.

allie500:
Fixes #6031 

This pull request changes all the JavaScript files to TypeScript files in the `client/settings/fraud-protection` directory and its subdirectories.

#### Testing instructions

- Checkout PR branch.
- Build the client: `npm run build:client`
- All tests should pass.
- Make sure the feature flag option is set to `1`: `npm run wp update option wcpay_fraud_protection_settings_active 1`
- Visit the WCPay settings page and scroll to the bottom. The fraud protection section should look like the following screenshot:

pr6032_wcpay-settings-fraud-protection

- Click the question mark icon next to Basic. A modal should appear and look like the following screenshot:

pr6032_basic-protection-modal png

- Perform each of the following actions on the modal to confirm that it closes with each action:
  - Click the 'Got it' button.
  - Click the 'X' button.
  - Click anywhere outside of the modal.

- Click the advanced radio button, then the configure button.
- Check that each of the following advanced rule cards are rendered on the page:
  - AVS Mismatch
  - CVC Verification
  - International IP Address
  - IP Address Mismatch
  - Address Mismatch
  - Purchase Price Threshold
  - Order Items Threshold

- Make changes to each rule ensuring that the toggle works for each rule. Set each rule to only hold for review. For purchase price threshold rule set minimum to 10 and max to 100. For order items threshold rule set minimum to 2, max to 20. Save changes.
- Inspect the saved ruleset in your local client and local server DBs to ensure that the rules were saved correctly.

Client DB:
 
```sql
SELECT * FROM `wp_options` WHERE `option_name` = '_transient_wcpay_fraud_protection_settings';
```

Server DB:

```sql
SELECT * FROM `wcpay_fraud_rulesets` ORDER BY `wcpay_fraud_rulesets`.`wcpay_fraud_ruleset_id` DESC;
```

- Now set each of the rules to block. Save changes.
- Inspect the saved ruleset in your local client and local server DBs to ensure that the rules were saved correctly.
- Now remove the purchase price threshold and order items threshold minimums. Save settings.
- Inspect the saved ruleset to ensure that the purchase price threshold and order items threshold rules were saved correctly.
- Now re-add the purchase price threshold and order items threshold minimums. Remove the purchase price threshold and order items threshold maximums. 
  - You should receive the following error notice on the order items threshold when there is a minimum value but no max value: _Maximum item count must be greater than the minimum item count._ Delete the minumum value, you should get the following notice: _An item range must be set for this filter to take effect._ 
  - Toggle the order items threshold filter off. 
  - Save settings.
- Inspect the saved ruleset to ensure that the purchase price threshold and order items threshold rules were saved correctly.
- Test the same notices on the purchase price threshold rule:
  - Delete both the min and max values, you should get the following notice: _A price range must be set for this filter to take effect._
  - Put a value in the minimum field, then put a smaller value in the max field. You should get the following error notice: _Maximum purchase price must be greater than the minimum purchase price._
- Now set the following advanced config:
  - Disable filtering for: International IP Address, IP Address Mismatch, and Address Mismatch.
  - Purchase price threshold: leave min field empty, set max field to 50, uncheck the block payment box.
  - Order items threshold: leave min field empty, set max field to 10, check the block payment box.
  - Save settings.
- Check that a new `wcpay_fraud_rule_id` was created for this new saved ruleset on the server DB.
- Now click save settings again. 
- Check that a new `wcpay_fraud_rule_id` was ***NOT*** created for this new saved ruleset on the server DB.
- Now, go back to the WCPay settings page, set your fraud protection level to basic and save settings. Check that the fraud ruleset was correctly saved on the client DB and server DB.
- Lastly, click save settings again and check that a new `wcpay_fraud_rule_id` was ***NOT*** created for this new saved ruleset on the server DB.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6032` or branch name `update/6031-convert-fraud-protection-settings-to-typescript` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/6031-convert-fraud-protection-settings-to-typescript) :rocket:

---

Build info: 

- Latest commit: 9e1caf5b32cf66fe6fff05731c93fd0b700a29c0
- Build time: 2023-05-11 18:51:42 UTC

client/settings/fraud-protection/advanced-settings/allow-countries-notice.tsx
```diff
@@ -12,7 +12,13 @@ import FraudPreventionSettingsContext from './context';
 import FraudProtectionRuleCardNotice from './rule-card-notice';
 import { getSettingCountries, getSupportedCountriesType } from './utils';
 
-const getNoticeText = ( filterType, blocking ) => {
+declare global {
+	interface Window {
+		wcSettings: Record< string, any >;
+	}
+}
```
ismaeldcom: We should move this global declaration to our `globals.d.ts`, similar to how we declare `wcpaySettings` or `wcTracks` there.

client/components/amount-input/index.tsx
```diff
@@ -4,30 +4,40 @@
...
 } ) => {
-	const validateInput = useCallback( ( subject ) => {
-		// Only allow decimals, a single dot, and more decimals (or an empty value).
-		return /^(\d+\.?\d*)?$/m.test( subject );
-	}, [] );
+	// Only allow decimals, a single dot, and more decimals (or an empty value).
+	const validateInput = useCallback(
+		( subject ) => /^(\d+\.?\d*)?$/m.test( subject ),
+		[]
+	);
```
ismaeldcom, line 28: If we use `useCallback` just to get a memoized function that will never change, it's better to define it outside the component, especially for pure functions.

React memoization also has a performance hit, so it should be used only when it could increase performance.

client/data/settings/selectors.js
```diff
@@ -213,5 +213,5 @@ export const getCurrentProtectionLevel = ( state ) => {
 };
 
 export const getAdvancedFraudProtectionSettings = ( state ) => {
-	return getSettings( state ).advanced_fraud_protection_settings || EMPTY_OBJ;
+	return getSettings( state ).advanced_fraud_protection_settings || EMPTY_ARR;
```
tpaksu, line 214: Good catch! We started to deliver the settings as a rule array, so it should be initialized as an array.

client/components/amount-input/index.tsx
```diff
@@ -4,30 +4,40 @@
...
 
 	useEffect( () => {
 		if ( ! validateInput( value ) ) {
 			onChange( '' );
 		}
 	}, [ validateInput, value, onChange ] );
 
-	if ( isNaN( value ) || null === value ) value = '';
+	if ( isNaN( Number( value ) ) || null === value ) value = '';
```
ismaeldcom: With the new declared type (`value: string`) it shouldn't be `null`.

I'm curious about why are you assigning an empty string to `value` when it is not a number and also calling `onChange('')` with `useEffect` 🤔

If you want to force the use of numbers, you can declare value like that and cast it later on the `<input/>`. And for the change event, it will be easier to just add an `else` on `handleChange` to return an empty string. In this way you don't need to keep casting `value` and event value every time you use `<AmountInput/>`.

client/settings/fraud-protection/advanced-settings/allow-countries-notice.tsx
```diff
@@ -66,12 +78,11 @@ const AllowedCountriesNotice = ( { setting } ) => {
...
-					.map( ( countryCode ) => {
-						return (
+					.map(
+						( countryCode: string ) =>
 							window.wcSettings.countries[ countryCode ] ?? false
-						);
-					} )
-					.filter( ( element ) => element )
+					)
+					.filter( ( element: string ) => element )
```
ismaeldcom: Nothing wrong here, just wondering, did you get any error here without declaring `map` and `filter` param types? It should properly infer `string` type from `settingCountries: string[]`. And I've seen it also on other files.

client/settings/fraud-protection/advanced-settings/context.ts
ismaeldcom, line 1: If you'd like to know more about some Context patterns, I've use one of my favorites in [the Stepper component](https://github.com/Automattic/woocommerce-payments/blob/584305c7c2f397a1f2faaf55b779e9db38bf53ac/client/components/stepper/index.tsx). In the end it's the same, just a nice wrapper over it to make type declaration, initialization, and use of the context a bit more straightforward.

client/settings/fraud-protection/advanced-settings/index.tsx
```diff
@@ -37,8 +37,14 @@ import './../style.scss';
...
 import wcpayTracks from 'tracks';
+import {
+	UseCurrentProtectionLevel,
+	UseAdvancedFraudPreventionSettings,
+	ProtectionSettingsUI,
+	UseSettings,
+} from '../interfaces';
 
-const observerEventMapping = {
+const observerEventMapping: { [ key: string ]: string } = {
```
ismaeldcom: I think you've already used it in this PR, but as a reminder, `{ [ key: string ]: string }` can be written  like `Record< string, string>` too.

client/settings/fraud-protection/advanced-settings/index.tsx
```diff
@@ -72,28 +78,32 @@ const Breadcrumb = () => (
...
 	return headerElement && ReactDOM.createPortal( children, headerElement );
 };
 
-const FraudProtectionAdvancedSettingsPage = () => {
-	const { saveSettings, isLoading, isSaving } = useSettings();
+const FraudProtectionAdvancedSettingsPage: React.FC = () => {
+	const { saveSettings, isLoading, isSaving } = useSettings() as UseSettings;
 
-	const cardObserver = useRef( null );
+	const cardObserver = useRef< IntersectionObserver | null >( null );
```
ismaeldcom: No need to add `| null` with `useRef`, as it already expects its generic type to be nullable.

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
```diff
@@ -28,7 +35,7 @@ const FraudProtectionRuleCardNotice = ( { type, children } ) => {
 
 	return (
 		<BannerNotice
-			status={ type }
+			status={ type } // Type 'string' is not assignable to type 'Status | undefined'.
```
tpaksu: Is this a todo comment?

client/settings/fraud-protection/advanced-settings/utils.ts
```diff
@@ -0,0 +1,353 @@
...
+					};
+					break;
+				case Rules.RULE_IP_ADDRESS_MISMATCH:
+					parsedUIConfig[ rule.key ] = {
+						enabled: true,
+						block: rule.outcome === Outcomes.BLOCK,
+					};
+					break;
+				case Rules.RULE_ORDER_ITEMS_THRESHOLD:
+					const minItems = findCheck(
```
tpaksu: Maybe we should also rename these variables with a `Check` suffix, like `minItemsCheck`, just to increase clarity.

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
```diff
@@ -11,7 +11,14 @@ import './../style.scss';
 import BannerNotice from 'wcpay/components/banner-notice';
 import { TipIcon } from 'wcpay/icons';
 
-const FraudProtectionRuleCardNotice = ( { type, children } ) => {
+interface FraudProtectionRuleCardNoticeProps {
+	type: any;
+}
```
ismaeldcom: If you're checking supported types later, why not enforcing it here with template literal types? Like `'error' | 'warning' | 'info'`

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.tsx
```diff
@@ -13,18 +19,25 @@ import FraudProtectionRuleToggle from '../rule-toggle';
...
+const OrderItemsThresholdCustomForm: React.FC< OrderItemsThresholdCustomFormProps > = ( {
+	setting,
+} ) => {
 	const { protectionSettingsUI, setProtectionSettingsUI } = useContext(
 		FraudPreventionSettingsContext
 	);
 
 	const minItemsTemp = parseInt(
-		protectionSettingsUI[ setting ].min_items,
+		protectionSettingsUI[ setting ].min_items?.toString() || '',
```
ismaeldcom: A quick shorthand for `?.toString() || ''` is `+ ''`, it will cast any type to string. One of those JS strange things...

client/settings/fraud-protection/index.tsx
```diff
@@ -13,20 +13,22 @@ import { ProtectionLevels } from './components';
...
-					<h4>
-						<span id="fraud-protection-card-title">
-							{ __(
-								'Set your payment risk level',
-								'woocommerce-payments'
-							) }
-						</span>
-					</h4>
-					<ProtectionLevels />
+					<>
```
tpaksu: Why did we need this fragment wrapper here?

client/settings/fraud-protection/advanced-settings/rule-card.tsx
```diff
@@ -10,7 +10,18 @@ import { Card } from '@wordpress/components';
 import './../style.scss';
 import CardBody from '../../card-body';
 
-const FraudProtectionRuleCard = ( { title, description, children, id } ) => {
+interface FraudProtectionRuleCardProps {
+	title: string;
+	description: string | React.ReactNode;
```
ismaeldcom: `ReactNode` already includes `string` type.

client/settings/fraud-protection/interfaces.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * External dependencies
+ */
+import { Dispatch, SetStateAction } from 'react';
+
+export interface FraudPreventionSetting {
+	block: boolean;
+	enabled: boolean;
+	min_items?: string | number | null;
```
tpaksu: Will it be a good idea to split this into multiple interfaces? I think, for example, the purchase price threshold can initialize without `min_amount` in the provided setting object this way.

client/settings/fraud-protection/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+	operator: string;
+	checks: FraudProtectionSettingsSingleCheck[];
+}
+
+export type FraudProtectionSettingsCheck =
+	| FraudProtectionSettingsSingleCheck
+	| FraudProtectionSettingsMultipleChecks
+	| null;
+
+export interface AdvancedFraudProtectionSettings {
```
tpaksu: This should be a definition for a `FraudProtectionRule` object. `AdvancedFraudProtectionSettings` is an array of `FraudProtectionRule`.

client/settings/fraud-protection/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+	key: string;
+	outcome: string;
+	check: FraudProtectionSettingsCheck;
+}
+
+export type UseCurrentProtectionLevel = [ string, ( level: string ) => void ];
```
tpaksu: Is this naming standard in our code base?

client/settings/fraud-protection/advanced-settings/test/index.test.tsx
```diff
@@ -35,13 +35,63 @@ jest.mock( '@wordpress/data', () => ( {
...
+			}
+		>;
+		isMultiCurrencyEnabled: string;
+	};
+};
+
+const mockUseCurrentProtectionLevel = useCurrentProtectionLevel as jest.MockedFunction<
+	() => [ string, ( level: string ) => void ]
+>;
+
```
ismaeldcom: We have installed `ts-jest/utils` which exports `mocked` a handy method to avoid those mock declarations. We just need to wrap any TS method within `mocked()` and it will allow us to mock while keeping the type declaration.

tpaksu (changes requested):
Thanks for working on this! The code looks good, but before starting testing it, I have found some minor issues in the code, mostly about naming conventions. Once you address those, I'll be happy to start manual testing.

client/settings/fraud-protection/index.tsx
ismaeldcom: Update: Taha beat me here! 🙂

No need for a React fragment within `<CardBody>`

client/settings/fraud-protection/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+import { Dispatch, SetStateAction } from 'react';
+
+export interface FraudPreventionSetting {
+	block: boolean;
+	enabled: boolean;
+	min_items?: string | number | null;
+	max_items?: string | number | null;
+	min_amount?: string | number | null;
+	max_amount?: string | number | null;
```
ismaeldcom: Following the `<AmountInput/>` comment, it would better to declare those as optional numbers, and avoid `string` and `null` if we only want numbers there. Although that could mean a bit more refactor.

I like to think of a TS migration more like – make the code type oriented, rather than make TSlint happy 😄

ismaeldcom (commented):
Thanks for the effort to contribute to the TS migration! 😄

I've finished the code review, everything looks good! I just left some nitpicks and slightly enhancements here and there. Let me know if you need more details on any of them 🙂

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
allie500: We ended up going with 'any', b/c if we use `'error' | 'warning' | 'info'` we end up with an error on this test that we had to modify to accpt an 'invalid' type. Before it was testing if no type was passed and TS didn't like that. 
https://github.com/Automattic/woocommerce-payments/blob/ed593379de71cdb8e7c3c76645681830b64d4839/client/settings/fraud-protection/advanced-settings/test/rule-card-notice.test.tsx#L38

I'll leave it as 'any' for now but if you have any other suggestions, I'm all ears.

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
allie500: No I think it was just a comment we put there as we were dealing with some TS linting issues with it and a related test at the time. I guess we just forgot to remove it. Thanks for flagging. Removed in [3fba24d](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/3fba24df9ea4c5e43daf0ee2ffa7695457f13f15).

client/settings/fraud-protection/advanced-settings/utils.ts
allie500: That's a good idea! I've made that change in [484a4ec](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/484a4ec68b48aeb62e34fa4806572dd0983c5cb8).

client/settings/fraud-protection/index.tsx
allie500: ~~We need a fragment wrapper because `<Card />` and `<FraudProtectionTour />` are on the same level.~~

I saw that you were referring to the fragment inside `<CardBody />` not the outer fragment. We had added it before because the `<h4>` and the `<ProtectionLevels />` were at the same level and TS was complaining about it. I've removed it in [82ce8c2](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/82ce8c264520073658396aed9a8ddcd4edf394ff).

tpaksu: Where is `<FraudProtectionTour />`?

allie500: We had added it before because the `<h4>` and the `<ProtectionLevels />` were at the same level and TS was complaining about it. I've removed it in [82ce8c2](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/82ce8c264520073658396aed9a8ddcd4edf394ff). TS wasn't complaining about it any more. 🤷

allie500: I thought you were referring to the outer fragment at first.

ismaeldcom: JSX expressions must have a single parent, but usually components' children have `ReactNode[]` type, so they don't need a Fragment.

client/settings/fraud-protection/interfaces.ts
allie500: I'm not sure I understand. Could you expand on this? Since it is a type, I thought that it would need to be in `UpperCamelCase`.

client/settings/fraud-protection/interfaces.ts
allie500: Thanks for flagging this and explaining. I've updated it in [991151c](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/991151c184e833b5807093b895f81973a8525927).

client/settings/fraud-protection/interfaces.ts
allie500: I'm not sure I understand. Could you expand on this? Since it is a type, I thought that it would need to be in `UpperCamelCase`.

client/settings/fraud-protection/interfaces.ts
allie500: I think it could be a good idea because it could make it easier to to understand and maintain the code. That said, we do have some pending changes to make to the advanced settings based on feedback from the demo. So I don't know if we should do it now or as part of those pending changes. What do you think @eduardoumpierre?

If we move forward with it now, I can split it out into the following five interfaces:
- International IP Address
- IP Address Mismatch
- Address Mismatch
- Purchase Price Threshold
- Order Items Threshold

client/settings/fraud-protection/interfaces.ts
tpaksu: `useCurrentProtectionLevel` is a redux hook, and is described with a verb (`useXXX`), and capitalizing it will still keep it as a verb, whereas the type should be a noun IMO. Maybe using `UseCurrentProtectionHookType` or something would fit better? Wdyt?

tpaksu: For example, this page (https://react-redux.js.org/using-react-redux/usage-with-typescript) shows the naming for `useAppDispatch` as `AppDispatch`, so we can tweak the naming with a suffix like `CurrentProtectionLevelHook` maybe.

allie500: Okay, I understand what you mean now! I like `UseCurrentProtectionLevelHook`, that way it maintains the full name of the hook but includes that suffix saying what it is (a Redux hook) and makes the type a noun. 

WDYT? Thumbs up for `UseCurrentProtectionLevelHook`?

client/settings/fraud-protection/advanced-settings/allow-countries-notice.tsx
allie500: Thanks for the advice. Fixed in [f318424](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/f318424198410c9919f9df3a5b0832caf2d2ca85).

client/components/amount-input/index.tsx
allie500: I think I took the correct approach based on this feedback in [09b7a98](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/09b7a9845f62e3fec04e675545a99c5a550f675e). Could you have a look and let me know if the changes are what you were looking for?

client/components/amount-input/index.tsx
allie500: I think I took the correct approach based on this feedback in [09b7a98](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/09b7a9845f62e3fec04e675545a99c5a550f675e). Could you have a look and let me know if the changes are what you were looking for?

client/settings/fraud-protection/interfaces.ts
tpaksu: I think if we get rid of the `Use` prefixes, and append `Hook`, they may look better.

client/settings/fraud-protection/interfaces.ts
tpaksu: No need for five interfaces, just a standard one, and two additional ones for the rules that require extra parameters (purchase price and order items), so we can use a combined type, like described here: https://ultimatecourses.com/blog/use-intersection-types-to-combine-types-in-typescript

client/settings/fraud-protection/advanced-settings/context.ts
tpaksu: I think this comment is for me as I'm the author of this ignorantly written context 😬 I like patterns, but this code only needed a global variable holder.

client/settings/fraud-protection/interfaces.ts
allie500: Got it. Fixed in [1b56724](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/1b56724105f1379f1c84461174a4ec0a8ae72a96).

client/settings/fraud-protection/advanced-settings/allow-countries-notice.tsx
allie500: I'm not sure why we did that. I removed them and TS didn't complain. Addressed in [8e18987](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/8e18987eaabbb3579936208afaf829aef485599b).

client/settings/fraud-protection/advanced-settings/cards/order-items-threshold.tsx
allie500: Thanks for this tip! I've updated these instances in this file in [8e18987](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/8e18987eaabbb3579936208afaf829aef485599b).

client/settings/fraud-protection/advanced-settings/rule-card.tsx
allie500: Thanks for flagging this one. I've fixed it in [d99ee3e](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/d99ee3e5914d25ac196ea89aed8e7f3fb00bdc02).

client/settings/fraud-protection/advanced-settings/test/index.test.tsx
allie500: Thanks for sharing this info. I was able to use it on `mockUseCurrentProtectionLevel` and `mockUseAdvancedFraudProtectionSettings` in [d99ee3e](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/d99ee3e5914d25ac196ea89aed8e7f3fb00bdc02).

client/settings/fraud-protection/advanced-settings/context.ts
ismaeldcom: Nothing wrong with it! 😄 I just like to share some Context with TS patterns, because they tend to be quite messy with bigger one. This case is a quite simple one.

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
ismaeldcom: There are different ways to do it, an example can be:

```tsx
// rule-card-notice.tsx
const supportedTypes = [ 'error', 'warning', 'info' ] as const; // `const` will make its type literal array.

interface FraudProtectionRuleCardNoticeProps {
	type: typeof supportedTypes[ number ]; // With `typeof` and `[number]` we get the template literal.
}

// rule-card-notice.test.tsx
<FraudProtectionRuleCardNotice type={ 'invalid' as any
/* You can cast to any to avoid any TS error and test it,
emulating the behavior of using the component from a JS file. */  }>
```

client/settings/fraud-protection/interfaces.ts
allie500: Okay we've addressed this in [e771f99](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/e771f99dd0ff3557e0eb947de2def477714faed9). I think you should now be good to re-review and test. 

Let us know if you have any further feedback.

client/settings/fraud-protection/advanced-settings/index.tsx
allie500: Thanks for catching this one. Fixed in [e771f99](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/e771f99dd0ff3557e0eb947de2def477714faed9).

client/settings/fraud-protection/advanced-settings/index.tsx
allie500: I tried removing ` | null` but when I did, I got the following error on line 216:

> Cannot assign to 'current' because it is a read-only property

```TypeScript
cardObserver.current = new IntersectionObserver( observerCallback );
```

So I left it as is.

client/settings/fraud-protection/advanced-settings/rule-card-notice.tsx
allie500: @eduardoumpierre helped me with this a few days ago and he came up with the same solution. Great minds think alike. 😁

client/settings/fraud-protection/interfaces.ts
allie500: Thanks for the suggestion here. We talked about it a bit during our last pairing session and we think this refactor would be better tackled as a cooldown week activity so that we can get this issue finished up so we can move on to some of the other advanced fraud protection rule changes we need to make. Those changes are waiting for this TS conversion to be merged.

client/settings/fraud-protection/advanced-settings/index.tsx
ismaeldcom: That's right, the way to do it is to use `undefined` instead of `null`. Defining it like `useRef< IntersectionObserver >();`

client/settings/fraud-protection/advanced-settings/utils.ts
```diff
@@ -96,117 +98,83 @@ const buildRuleset = (
...
-							key: Checks.CHECK_ITEM_COUNT,
-							operator: CheckOperators.OPERATOR_GT,
-							value:
-								parseInt(
-									ruleConfiguration?.max_items?.toString() ||
-										'',
-									10
-								) ?? null,
-					  };
+			if ( isOrderItemsThresholdSetting( ruleConfiguration ) ) {
```
tpaksu, line 101: Why didn't we directly use 

```ts
if ( ruleConfiguration instanceof FraudPreventionOrderItemsThresholdSetting ) {
```

or (if that only works on class instances)

```ts
if ( typeof ruleConfiguration === 'FraudPreventionOrderItemsThresholdSetting' ) {
```

here? I think a single check like this will be enough than creating separate functions to check the type.

client/settings/fraud-protection/interfaces.ts
tpaksu: I was talking about *all* variables that have the `UseXXX` pattern, you only changed one of them :)

allie500: 🤦 Thanks for pointing that out! I've addressed it in [394a1a3](https://github.com/Automattic/woocommerce-payments/pull/6032/commits/394a1a3cfa4bdb873a5e8ed28309256e64053ca4).

client/settings/fraud-protection/advanced-settings/utils.ts
allie500: Thanks for the suggestion here. I gave it a try but both approaches result in errors.

```typescript
if ( ruleConfiguration instanceof FraudPreventionOrderItemsThresholdSetting ) {
```

> 'FraudPreventionOrderItemsThresholdSetting' only refers to a type, but is being used as a value here.

```typescript
if ( typeof ruleConfiguration === 'FraudPreventionOrderItemsThresholdSetting' ) {
```
> This comparison appears to be unintentional because the types '"string" | "number" | "bigint" | "boolean" | "symbol" | "undefined" | "object" | "function"' and '"FraudPreventionOrderItemsThresholdSetting"' have no overlap.

tpaksu: Gotcha. It's also written in https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates that it can't be done.

tpaksu (changes requested):
Tested all and they work, except the amount input doesn't clear and doesn't apply the decimal point anymore. Once you fix that, I'll be happy to approve. Thanks for working on this!

tpaksu (approved):
Ok I've tested the amount input and it seems to work as expected, sending the correct values to the server. We can ship it. Thanks for taking this on and nailing it! LGTM!

PR #6061, opened 2023-04-18
Update naming convention from "platform checkout" to "woopay"

### LINKED ISSUE #6070: Migrate platform checkout naming convention to WooPay naming convention on Client
@zmaglica commented on [Thu Mar 16 2023](https://github.com/Automattic/woocommerce-payments-meta/issues/96)

### Description
Currently, we have mixed naming conventions between WooPay and Platform checkout in our codebase. WooPay and Platform checkout are used interchangeably to refer to the same thing, which can be confusing for new developers who are not aware of this naming convention.

This issue has become more prominent as WooPay has grown in popularity, and it's important that we standardize our naming conventions to avoid confusion and maintain code quality.

### Impact
The mixed naming conventions can impact our development process in a number of ways. For example, it can make it more difficult to understand code, which can slow down development and increase the likelihood of errors. Additionally, inconsistent naming conventions can make it harder to maintain the codebase and can lead to longer-term issues down the road.

### Solution
To address this issue, we need to standardize on a single naming convention. Specifically, we should use "WooPay" consistently throughout the codebase.

To make this change, we will need to migrate all comments, file names, constants, namespaces, and other instances of "Platform checkout" to "WooPay". To do this, you can use your favorite IDE and search for string occurrences with "Platform checkout" keyword. Be sure to take extra care to ensure that all instances are updated, and to test thoroughly to avoid introducing new issues.

### Here are the specific steps you can follow to update the codebase:

1. Search for all instances of "Platform checkout" in the codebase using your IDE.
2. Replace each instance with "WooPay".
3. Double-check that all instances have been updated, and that no issues have been introduced.
4. Run any necessary tests to ensure that the changes have not caused any regressions.
5. By standardizing on a single naming convention, we can avoid confusion and maintain code quality.

htdat:
Fixes https://github.com/Automattic/woocommerce-payments/issues/6070

- Change all "platform checkout" texts to "woopay" according to their forms.
- Only change variables, parameters (used internally in merchant/client sites, e.g. the front-end code), const names. 
- DO NOT change option name keys (saved in the wp_options), params and endpoints to interact with the server.

#### Testing instructions

- All PHP and JS tests are passed. 
- I would suggest looking at my inline comments and verify they're all reasonable.
- Go through all changes and spot out non-reasonable changes.
- (Maybe) Run this through weekly WooPay tests by GlobalStep too.

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6061` or branch name `dev/migrate-to-woopay-naming` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=dev/migrate-to-woopay-naming) :rocket:

---

Build info: 

- Latest commit: ae9f278bb658f546aafdc68d6ddf0e5cf448b358
- Build time: 2023-05-11 11:04:15 UTC

client/components/woopay/save-user/checkout-page-save-user.js
```diff
@@ -78,17 +78,17 @@ const CheckoutPageSaveUser = ( { isBlocksCheckout } ) => {
...
-						save_user_in_platform_checkout: isSaveDetailsChecked,
-						platform_checkout_source_url: window.location.href,
-						platform_checkout_is_blocks: true,
-						platform_checkout_viewport: `${ viewportWidth }x${ viewportHeight }`,
-						platform_checkout_user_phone_field: {
+						save_user_in_woopay: isSaveDetailsChecked,
+						woopay_source_url: window.location.href,
+						woopay_is_blocks: true,
+						woopay_viewport: `${ viewportWidth }x${ viewportHeight }`,
+						woopay_user_phone_field: {
```
htdat, line 85: I think these changes should be fine as they're all handled within the merchant site. We should preserve similar metedata keys in `WooPay_Save_User::maybe_add_userdata_to_metadata` as they interact with WPCOM.

client/tracks/index.js
```diff
@@ -76,16 +76,14 @@ const events = {
...
-	PLATFORM_CHECKOUT_EXPRESS_BUTTON_CLICKED:
-		'platform_checkout_express_button_clicked',
+	WOOPAY_OFFERED: 'woopay_offered',
+	WOOPAY_OTP_START: 'woopay_otp_prompt_start',
+	WOOPAY_OTP_COMPLETE: 'woopay_otp_prompt_complete',
+	WOOPAY_OTP_FAILED: 'woopay_otp_prompt_failed',
+	WOOPAY_AUTO_REDIRECT: 'woopay_auto_redirect',
+	WOOPAY_SKIPPED: 'woopay_skipped',
+	WOOPAY_EXPRESS_BUTTON_OFFERED: 'woopay_express_button_offered',
+	WOOPAY_EXPRESS_BUTTON_CLICKED: 'woopay_express_button_clicked',
```
htdat, line 92: These track event name changes would be OK based on this convo
p1679328700028019/1679325891.095749-slack-C022WMN88KG

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -873,8 +873,8 @@ protected function prepare_payment_information( $order ) {
 			$payment_information->must_save_payment_method_to_store();
 		}
 
-		if ( $this->platform_checkout_util->should_save_platform_customer() ) {
-			do_action( 'woocommerce_payments_save_user_in_platform_checkout' );
+		if ( $this->woopay_util->should_save_platform_customer() ) {
+			do_action( 'woocommerce_payments_save_user_in_woopay' );
```
htdat, line 877: Since this is only for our internal usage, so I think it's fine to change this action name.

includes/woopay-user/class-woopay-extension.php
```diff
@@ -1,18 +1,18 @@
...
 
 /**
- * Class that registers a blocks extension callback for platform checkout new user.
+ * Class that registers a blocks extension callback for woopay new user.
  */
-class Platform_Checkout_Extension {
-	const EXTENSION_NAMESPACE           = 'platform-checkout';
-	const PLATFORM_CHECKOUT_SESSION_KEY = 'platform-checkout-user-data';
+class WooPay_Extension {
+	const EXTENSION_NAMESPACE = 'woopay';
```
htdat, line 14: This change would be OK as the front-end value is updated too.

includes/woopay-user/class-woopay-extension.php
```diff
@@ -1,18 +1,18 @@
...
+	const WOOPAY_SESSION_KEY  = 'woopay-user-data';
```
htdat, line 15: This would be ok as this session will be destroyed when payment is complete. See `WooPay_Save_User::maybe_clear_session_key`

includes/admin/class-wc-rest-payments-settings-controller.php
```diff
@@ -384,10 +384,10 @@ public function get_settings(): WP_REST_Response {
...
 				'is_saved_cards_enabled'              => $this->wcpay_gateway->is_saved_cards_enabled(),
 				'is_card_present_eligible'            => $this->wcpay_gateway->is_card_present_eligible(),
-				'is_platform_checkout_enabled'        => 'yes' === $this->wcpay_gateway->get_option( 'platform_checkout' ),
-				'platform_checkout_custom_message'    => $this->wcpay_gateway->get_option( 'platform_checkout_custom_message' ),
-				'platform_checkout_store_logo'        => $this->wcpay_gateway->get_option( 'platform_checkout_store_logo' ),
-				'platform_checkout_enabled_locations' => $this->wcpay_gateway->get_option( 'platform_checkout_button_locations', [] ),
+				'is_woopay_enabled'                   => 'yes' === $this->wcpay_gateway->get_option( 'platform_checkout' ),
+				'woopay_custom_message'               => $this->wcpay_gateway->get_option( 'platform_checkout_custom_message' ),
+				'woopay_store_logo'                   => $this->wcpay_gateway->get_option( 'platform_checkout_store_logo' ),
+				'woopay_enabled_locations'            => $this->wcpay_gateway->get_option( 'platform_checkout_button_locations', [] ),
```
htdat, line 390: IMO, these should be OK as they preserve option names while the setting keys are updated here as well as the front-end (React) code.

includes/woopay/class-woopay-utilities.php
```diff
@@ -95,8 +95,8 @@ public function get_woopay_available_countries() {
 			return $this->get_persisted_available_countries();
 		}
 
-		$platform_checkout_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'https://pay.woo.com';
-		$url                    = $platform_checkout_host . '/wp-json/platform-checkout/v1/user/available-countries';
+		$woopay_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'https://pay.woo.com';
+		$url         = $woopay_host . '/wp-json/platform-checkout/v1/user/available-countries';
```
htdat, line 132: - Keep `PLATFORM_CHECKOUT_HOST` as it's being used in the woopay plugin. If we can change in the woopay plugin, this const name would change `WOOPAY_HOST` too.
- The endpoint `/wp-json/platform-checkout` is preserved.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -38,7 +38,7 @@ class WC_Payments_API_Client {
 
 	const ACCOUNTS_API                 = 'accounts';
 	const CAPABILITIES_API             = 'accounts/capabilities';
-	const PLATFORM_CHECKOUT_API        = 'accounts/platform_checkout';
+	const WOOPAY_API                   = 'accounts/platform_checkout';
```
htdat, line 41: Only change the const name, but preserve its value.

includes/class-wc-payments-features.php
```diff
@@ -181,11 +181,11 @@ public static function is_wcpay_subscriptions_eligible() {
...
-	 * Checks whether platform checkout is enabled.
+	 * Checks whether woopay is enabled.
 	 *
 	 * @return bool
 	 */
-	public static function is_platform_checkout_eligible() {
+	public static function is_woopay_eligible() {
 		// read directly from cache, ignore cache expiration check.
 		$account = WC_Payments::get_database_cache()->get( WCPay\Database_Cache::ACCOUNT_KEY, true );
 		return is_array( $account ) && ( $account['platform_checkout_eligible'] ?? false );
```
htdat: Only change the function name while preserve the account setting key `platform_checkout_eligible` as its value is retrieved from WPCOM.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1085,14 +1085,14 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 			// The sanitize_user call here is deliberate: it seems the most appropriate sanitization function
 			// for a string that will only contain latin alphanumeric characters and underscores.
 			// phpcs:ignore WordPress.Security.NonceVerification.Missing
-			$platform_checkout_intent_id = sanitize_user( wp_unslash( $_POST['platform-checkout-intent'] ?? '' ), true );
+			$woopay_intent_id = sanitize_user( wp_unslash( $_POST['platform-checkout-intent'] ?? '' ), true );
```
htdat, line 1088: Only change the var name and keep POST param `platform-checkout-intent` as it's being used in the woopay plugin.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -305,32 +298,29 @@ export const handlePlatformCheckoutEmailInput = async (
...
-		const PlatformCheckoutUserCheckEvent = new CustomEvent(
-			'PlatformCheckoutUserCheck',
-			{
-				detail: {
-					isRegisteredUser: userExist,
-				},
-			}
-		);
-		window.dispatchEvent( PlatformCheckoutUserCheckEvent );
+		const WooPayUserCheckEvent = new CustomEvent( 'WooPayUserCheck', {
```
timur27: I know this approach has been taken from the previous implementation, but I don't see a reason for the event name to start with the capital letter. 
```suggestion
		const woopayUserCheckEvent = new CustomEvent( 'woopayUserCheck', {
```

timur27: As we touched this function, could we align the event name with the function name by renaming the event to be `woopayUserExistEvent` or other way around, renaming the function to be called `dispatchUserCheckEvent`? I personally prefer the former, but am fine with any of the options.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -305,32 +298,29 @@ export const handlePlatformCheckoutEmailInput = async (
...
 			document
-				.querySelector( '#platform-checkout-subscriptions-login-error' )
+				.querySelector( '#woopay-subscriptions-login-error' )
 				.remove();
-			hasPlatformCheckoutSubscriptionLoginError = false;
+			hasWooPaySubscriptionLoginError = false;
 		}
 
-		if ( getConfig( 'platformCheckoutNeedLogin' ) ) {
+		if ( getConfig( 'woopayNeedLogin' ) ) {
```
timur27:
```suggestion
		if ( getConfig( 'woopayNeedsLogin' ) ) {
```

client/checkout/woopay/email-input-iframe.js
```diff
@@ -277,9 +270,9 @@ export const handlePlatformCheckoutEmailInput = async (
 
 	// Store if the subscription login error is being shown
 	// to remove it when change the e-mail address.
-	let hasPlatformCheckoutSubscriptionLoginError = false;
+	let hasWooPaySubscriptionLoginError = false;
```
timur27: I've noticed that we use `wooPay` and `woopay` interchangeably now. Could we align these two and have the unified approach? After going through lines `216`, `249` and other similar, I would expect this variable to be called 

```suggestion
	let hasWoopaySubscriptionLoginError = false;
```

client/checkout/woopay/express-button/index.js
```diff
@@ -12,7 +12,7 @@ import { WoopayExpressCheckoutButton } from './woopay-express-checkout-button';
 import WCPayAPI from '../../api';
 import request from '../../utils/request';
 
-const renderPlatformCheckoutExpressButton = () => {
+const renderWooPayExpressButton = () => {
```
timur27: In this particular case, it seems to me that the `checkout` word is a crucial part of the naming. The button is called WooPay Express Checkout Button in WCPay Dev Settings and in [the feature description](https://github.com/Automattic/woocommerce-payments/pull/5164). Having that said, I'd re-name it to:
```suggestion
const renderWooPayExpressCheckoutButton = () => {
```

client/checkout/woopay/express-button/test/index.test.js
```diff
@@ -20,13 +20,10 @@ jest.mock( '../woopay-express-checkout-button', () => ( {
 	},
 } ) );
 
-describe( 'renderPlatformCheckoutExpressButton', () => {
+describe( 'renderWooPayExpressButton', () => {
```
timur27, line 23: The same as in the previous comment: 
```suggestion
describe( 'renderWooPayExpressCheckoutButton', () => {
```

client/checkout/woopay/express-button/woopay-express-checkout-button.js
```diff
@@ -39,23 +39,23 @@ export const WoopayExpressCheckoutButton = ( {
 	useEffect( () => {
 		if ( ! isPreview ) {
 			wcpayTracks.recordUserEvent(
-				wcpayTracks.events.PLATFORM_CHECKOUT_EXPRESS_BUTTON_OFFERED,
+				wcpayTracks.events.WOOPAY_EXPRESS_BUTTON_OFFERED,
```
timur27, line 42: I'd leave `checkout` word in this constant:
```suggestion
				wcpayTracks.events.WOOPAY_EXPRESS_CHECKOUT_BUTTON_OFFERED,
```

client/checkout/woopay/express-button/woopay-express-checkout-button.js
```diff
@@ -39,23 +39,23 @@ export const WoopayExpressCheckoutButton = ( {
...
+	const initWooPay = ( e ) => {
 		e.preventDefault();
 
 		if ( isPreview ) {
 			return; // eslint-disable-line no-useless-return
 		}
 
 		wcpayTracks.recordUserEvent(
-			wcpayTracks.events.PLATFORM_CHECKOUT_EXPRESS_BUTTON_CLICKED,
+			wcpayTracks.events.WOOPAY_EXPRESS_BUTTON_CLICKED,
```
timur27, line 58: Same as above: 

```suggestion
			wcpayTracks.events.WOOPAY_EXPRESS_CHECKOUT_BUTTON_CLICKED,
```

client/components/woopay/hooks/use-woopay-user.js
```diff
@@ -3,31 +3,31 @@
...
+			'WooPayUserCheck',
+			handleWooPayUserCheck
 		);
 
 		return () => {
 			currentWindowRef.removeEventListener(
-				'PlatformCheckoutUserCheck',
-				handlePlatformCheckoutUserCheck
+				'WooPayUserCheck',
+				handleWooPayUserCheck
```
timur27: Is there a reason we start the event name with a capital letter? Why don't we define it `wooPayUserCheck` instead of `WooPayUserCheck`?

> Note: I know this was the case for the previous implementation as well, so I'm rather thinking now if we can improve here additionally, since we are already working with this codebase.

client/components/woopay/save-user/checkout-page-save-user.js
timur27: I agree with you. After reading [the docs](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/docs/third-party-developers/extensibility/rest-api/extend-rest-api-update-cart.md#the-solution), it looks like as long as the `EXTENSION_NAMESPACE` in `WooPay_Extension` is equal to `woopay`, the updated keys will be handled in `WooPay_Save_User::maybe_add_userdata_to_metadata`.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -76,7 +74,7 @@ export const handlePlatformCheckoutEmailInput = async (
 		( 'undefined' !== typeof performance &&
 			'back_forward' ===
 				performance.getEntriesByType( 'navigation' )[ 0 ].type ) ||
-		'true' === searchParams.get( 'skip_platform_checkout' );
+		'true' === searchParams.get( 'skip_woopay' );
 
 	// Track the current state of the header. This default
 	// value should match the default state on the platform.
```
timur27, line 79: Shouldn't we change this as well?
```suggestion
	// value should match the default state on WooPay.
```

client/tracks/index.js
timur27, line 1: Is there a reason why we didn't update the _platform_-related naming in `recordUserEvent` function?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -677,7 +677,7 @@ public function save_payment_method_checkbox( $force_checked = false ) {
 	 */
 	public function should_use_stripe_platform_on_checkout_page() {
```
timur27, line 678: How about changing this function's name to 
```suggestion
	public function should_use_woopay_on_checkout_page() {
```
and eventually the comment describing this method?

I'm not sure where `stripe` comes from here in the name.

includes/class-wc-payment-gateway-wcpay.php
timur27: [Confirmed](https://github.com/search?q=org%3AAutomattic+woocommerce_payments_save_user_in_platform_checkout&type=code) that the hook with this name is defined and used only within WCPay.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -873,8 +873,8 @@ protected function prepare_payment_information( $order ) {
 			$payment_information->must_save_payment_method_to_store();
 		}
 
-		if ( $this->platform_checkout_util->should_save_platform_customer() ) {
-			do_action( 'woocommerce_payments_save_user_in_platform_checkout' );
+		if ( $this->woopay_util->should_save_platform_customer() ) {
+			do_action( 'woocommerce_payments_save_user_in_woopay' );
 			$payment_information->must_save_payment_method_to_platform();
```
timur27, line 878: I'm curious why we didn't update this method definition to be
```suggestion
			$payment_information->must_save_payment_method_to_woopay();
```

includes/class-wc-payment-gateway-wcpay.php
timur27: In other words, it's possible to update the request parameter name, but it should be done simultaneously with the WooPay changes, or right after - is this a correct understanding? I'm curious if we have plans to still update this place and the ones similar to it or is there any other hard limitation which forces us to skip the naming change as part of the _platform to woopay naming_ migration?

includes/class-wc-payments-checkout.php
```diff
@@ -117,17 +117,17 @@ public function get_payment_fields_js_config() {
...
 			'isUPESplitEnabled'              => WC_Payments_Features::is_upe_split_enabled(),
 			'isSavedCardsEnabled'            => $this->gateway->is_saved_cards_enabled(),
-			'isPlatformCheckoutEnabled'      => $this->platform_checkout_util->should_enable_platform_checkout( $this->gateway ),
-			'isWoopayExpressCheckoutEnabled' => $this->platform_checkout_util->is_woopay_express_checkout_enabled(),
+			'isWooPayEnabled'                => $this->woopay_util->should_enable_woopay( $this->gateway ),
+			'isWoopayExpressCheckoutEnabled' => $this->woopay_util->is_woopay_express_checkout_enabled(),
 			'isClientEncryptionEnabled'      => WC_Payments_Features::is_client_secret_encryption_enabled(),
-			'platformCheckoutHost'           => defined( 'PLATFORM_CHECKOUT_FRONTEND_HOST' ) ? PLATFORM_CHECKOUT_FRONTEND_HOST : 'https://pay.woo.com',
+			'woopayHost'                     => defined( 'PLATFORM_CHECKOUT_FRONTEND_HOST' ) ? PLATFORM_CHECKOUT_FRONTEND_HOST : 'https://pay.woo.com',
 			'platformTrackerNonce'           => wp_create_nonce( 'platform_tracks_nonce' ),
```
timur27, line 124: I've already asked about this in some other comment - essentially I don't know why we didn't change this nonce name - could you please let me know if there are any reasons?

includes/class-wc-payments-checkout.php
timur27, line 1: In line `201`, we didn't update `wcpay-is-platform-payment-method` - is the reason the same as [described here](https://github.com/Automattic/woocommerce-payments/pull/6061/files#r1172521655)?

includes/class-wc-payments-features.php
timur27: What needs to be done in order to be able to update the key? Obviously, not within this issue, but overall I'm curious what we need to do to be able to retrieve `woopay_checkout_eligible` from WPCOM?

includes/class-wc-payments-woopay-button-handler.php
timur27, line 1: We haven't updated the `wcpay_platform_checkout_button_is_product_supported` ([seems to be used internally only](https://github.com/search?q=org%3AAutomattic+wcpay_platform_checkout_button_is_product_supported&type=code)) and `wcpay_platform_checkout_button_are_cart_items_supported` (as well, [used internally only](https://github.com/search?q=org%3AAutomattic+wcpay_platform_checkout_button_are_cart_items_supported&type=code)). Is there any plan to tackle these within another issue?

includes/class-wc-payments.php
```diff
@@ -1397,8 +1397,8 @@ public static function ajax_init_platform_checkout() {
 
 		$account_id = self::get_account_service()->get_stripe_account_id();
 
-		$platform_checkout_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'https://pay.woo.com';
-		$url                    = $platform_checkout_host . '/wp-json/platform-checkout/v1/init';
+		$woopay_host = defined( 'PLATFORM_CHECKOUT_HOST' ) ? PLATFORM_CHECKOUT_HOST : 'https://pay.woo.com';
```
timur27, line 1384: `PLATFORM_CHECKOUT_HOST` seems to be defined only in WCPay, but might have dependencies on WooPay - do we plan to update this somewhere else?


UPDATE: I now see your comment [here](https://github.com/Automattic/woocommerce-payments/pull/6061/files#r1172509038). I'm assuming that we won't wait until the value will be changed on the WooPay side and update the constant value afterward as part of this PR, but instead accomplish this in another issue?

timur27 (commented):
Thanks @htdat! I've left a couple of comments, but overall the PR looks good and I'm looking forward to going through it once more after the threads are resolved. I'll also test it then.

There are a couple of things that were described in the comments, but more high-level, they are about: 
* `woopay`/`wooPay`/`WooPay` seem to be used interchangeably across the code and I'm curious if we can align these.
* WooPay Express **_Checkout_** Button - I think the `checkout` part is crucial here as I've only heard people using this nomenclature when talking about this feature.
* I'm totally fine this PR won't update the options or constants which depend on the WooPay/other systems, but I'm still curious if we have this effort planned within any other issue.
* There are a couple of methods which remain being defined with the use of `platform` and I'm looking forward to hearing if it's something we just need to cover within other issue or not. 

Thanks a lot for your time answering/resolving the threads I've started!

htdat:
@timur27 To provide more context, in this PR, I used [`git grep`](https://gist.github.com/htdat/16038402d4318f854b845988b1f3f76e) to find out the forms of `platform checkout` and update them in three first commits:

- WOOPAY replaces `PLATFORM_CHECKOUT` - commit https://github.com/Automattic/woocommerce-payments/pull/6061/commits/7d1f04339b8193a784699306d10d98d0aca24705
- WooPay replaces these texts (commit https://github.com/Automattic/woocommerce-payments/pull/6061/commits/425136e7825fd171821b95cc2e55444506bb2930):
```
Platform Checkout
Platform checkout
Platform_Checkout
```

- `woopay` replaces these texts (commit https://github.com/Automattic/woocommerce-payments/pull/6061/commits/07a123b0db7766413faab1abc8a655ef91b588d8):
```
platform checkout
platform-checkout
platformCheckout
platform_checkout
``` 

then manually review them to ensure that they're not option keys, REST API params and endpoints interacting with the server. 

> `woopay`/`wooPay`/`WooPay` seem to be used interchangeably across the code and I'm curious if we can align these.

That's a good question! Eventually, I think we still only need two forms: `woopay` (lower-cases) and `Woopay` (sentence case). That said, I think we might narrow this effort to current changes only. Extending this change to existing texts seems excessive for this PR.

includes/class-wc-payments-features.php
zmaglica: There is an MC page where you can enable or disable this key. Moreover, there is a also a functionality to manually set any key or config, and you can manually update this key.

includes/class-wc-payment-gateway-wcpay.php
zmaglica: The idea is not to touch request parameters, route names, config values and responses from requests and functions to make it backward compatible. The main reason for that is that we were not entirely sure what could go wrong if we decided to change something in between and at that time, WooPay was disabled, so to be safe we decided to update the codebase only where we updated function names, files, namespaces, comments, variables and other values that could be easily updated by keeping the key items same.

includes/class-wc-payments-checkout.php
zmaglica: I am not entirely sure @timur27. @htdat do you maybe know the reason?

includes/class-wc-payments-checkout.php
zmaglica: I think it could be that reason. @htdat could you please confirm it, so we are 100% sure?

includes/class-wc-payments.php
zmaglica: @timur27 I think this another issue would be a better option to accomplish this. I think I mentioned this in previous comments, but we are trying to limit the scope of migration to file names, variables, functions, namespaces and other things that will be backward compatibly without changing config values, responses, return of functions and endpoints.

client/checkout/woopay/email-input-iframe.js
zmaglica: Make sense. I will address this in another commit where I will fix conflicts too.

includes/class-wc-payments-woopay-button-handler.php
zmaglica: I will address this in upcoming commits. Thank you for pointing it out.

includes/class-wc-payments-checkout.php
htdat: It's simply that I could not search it per my methodology https://github.com/Automattic/woocommerce-payments/pull/6061#issuecomment-1521293291 

Therefore, yes, we can search and change it. Just double check and change it in the front-end code too.

includes/class-wc-payments-checkout.php
htdat: It's simply that I could not search it per my methodology https://github.com/Automattic/woocommerce-payments/pull/6061#issuecomment-1521293291 

However, for this specific one, I am not sure how it's being used during the payment process. We might double check it and change all places accordingly.

client/checkout/woopay/email-input-iframe.js
zmaglica: I left the first option since it is more self describing and explanatory.

includes/class-wc-payments-woopay-button-handler.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/fd3f74b017b872a7c3aa3d1e751219a5024af783

client/components/woopay/hooks/use-woopay-user.js
bborman22: Just wanted to second this, this needs to match the case in [this line](https://github.com/Automattic/woocommerce-payments/pull/6061/files#diff-3ba3b857d1b29585cd53445c992d8b20ae6b27c92774ec5d2e7010b1d23a079aR301). This seems to be the only `CustomEvent` we're emitting, but to match the common event names elsewhere, lets go with `woopayUserCheck` if that works?

client/checkout/woopay/express-button/express-checkout-iframe.js
```diff
@@ -209,33 +207,32 @@ export const expressCheckoutIframe = async ( api, context, emailSelector ) => {
...
+		if ( ! getConfig( 'woopayHost' ).startsWith( e.origin ) ) {
 			return;
 		}
 
 		switch ( e.data.action ) {
 			case 'otp_email_submitted':
 				userEmail = e.data.userEmail;
 				break;
-			case 'redirect_to_platform_checkout':
+			case 'redirect_to_woopay':
```
bborman22, line 219: These action names are coming from WooPay server so would require coordination there to work. For now what about keeping both cases for now? I'm unsure how we'll ultimately handle the backwards compatibility here though longer term, so open to other suggestions.

client/checkout/woopay/express-button/express-checkout-iframe.js
```diff
@@ -209,33 +207,32 @@ export const expressCheckoutIframe = async ( api, context, emailSelector ) => {
...
-				api.initPlatformCheckout(
-					userEmail,
-					e.data.platformCheckoutUserSession
-				).then( ( response ) => {
-					if ( 'success' === response.result ) {
-						window.location = response.url;
-					} else {
-						showErrorMessage();
-						closeIframe( false );
+				api.initWooPay( userEmail, e.data.woopayUserSession ).then(
```
bborman22: `e.data.woopayUserSession` needs to remain `e.data.platformCheckoutUserSession` since this is coming for WooPay server as well and would require coordination as well.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -497,32 +484,28 @@ export const handlePlatformCheckoutEmailInput = async (
...
 
 	window.addEventListener( 'message', ( e ) => {
-		if ( ! getConfig( 'platformCheckoutHost' ).startsWith( e.origin ) ) {
+		if ( ! getConfig( 'woopayHost' ).startsWith( e.origin ) ) {
 			return;
 		}
 
 		switch ( e.data.action ) {
-			case 'auto_redirect_to_platform_checkout':
+			case 'auto_redirect_to_woopay':
```
bborman22, line 458: This should have the same issue as the other action names coming from WooPay, but I actually didn't encounter any problems. Let's keep both cases for now though just to be sure.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -497,32 +484,28 @@ export const handlePlatformCheckoutEmailInput = async (
...
 				hasCheckedLoginSession = true;
-				api.initPlatformCheckout(
-					'',
-					e.data.platformCheckoutUserSession
-				)
+				api.initWooPay( '', e.data.woopayUserSession )
```
bborman22: Same issue as other areas, needs to remain `platformCheckoutUserSession`.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -546,13 +529,13 @@ export const handlePlatformCheckoutEmailInput = async (
 				hasCheckedLoginSession = true;
 				closeLoginSessionIframe();
 				break;
-			case 'redirect_to_platform_checkout':
+			case 'redirect_to_woopay':
```
bborman22, line 493: Same issue as above for action names and multiple cases.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -546,13 +529,13 @@ export const handlePlatformCheckoutEmailInput = async (
...
 				wcpayTracks.recordUserEvent(
-					wcpayTracks.events.PLATFORM_CHECKOUT_OTP_COMPLETE
+					wcpayTracks.events.WOOPAY_OTP_COMPLETE
 				);
-				api.initPlatformCheckout(
-					platformCheckoutEmailInput.value,
-					e.data.platformCheckoutUserSession
+				api.initWooPay(
+					woopayEmailInput.value,
+					e.data.woopayUserSession
```
bborman22: Same issue with keeping `e.data.platformCheckoutUserSession`.

bborman22 (changes requested):
I was able to take a look at the PR and give it a full testing with the critical flows for WooPay. I only found 2 issues that just require a couple updates that are pretty simple. The big one is the iframe `postMessage` action names to be kept in sync with what WooPay is sending. The other one is just to make sure that the `isRegisteredUser` value is updated correctly by matching event case names.

I know this PR can be particularly tricky with the conflicts, so let me know if it would be helpful to schedule a time for a review to try and get it pushed through in a tighter window to avoid too much conflict resolutions?

client/checkout/woopay/email-input-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

client/checkout/woopay/email-input-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

client/checkout/woopay/email-input-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

client/checkout/woopay/email-input-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

client/checkout/woopay/express-button/express-checkout-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

client/checkout/woopay/express-button/express-checkout-iframe.js
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6061/commits/5dade92dafb2f7f088cff7f5d3408af76e7b40f6

bborman22 (approved):
Thank you for addressing those issues! All the code looks good and I've been able to test every WooPay critical flow and the base payment processing outside of WooPay with no issues.

I think this PR is good to get merged now!

PR #5895, opened 2023-03-29
Add/Update required components for progressive onboarding

ismaeldcom:
Fixes #4744

#### Changes proposed in this Pull Request
Implement the following components following PO design.
- **LoadBar** – An infinite load bar.
- **PhoneNumberControl** – Implemented similar to Stripe KYC one, in absence of our own design. It could be replaced later by paJDYF-85M-p2.
  - Country codes coming from `intl-tel-input`.
  - Use native `select` and `input` elements.
  - Selecting a country or clicking on the country code switch focus to the input.
  - Defaults to `wcpaySetting.connect.country` or `US`
  - Returns a phone number in E164 format.
- **GroupedSelectControl**
  - Group provided options by `option.group` which should match `group.key`.
  - Allow to toggle groups visibility, showing the first one by default.
  - Optionally searchable, filtering by option name, and also by the optional property `context`. Restoring visibility state after clearing search.
  - Allow selection by both mouse and keyboard.

Update `CustomSelectControl` to match final design and implement `useUniqueId` hook using `lodash` under the hood.

#### Testing instructions

They should match our design and behave according to our needs.

>**Note** Those are just the components' implementation without using them directly. To check their functionality we need to add somewhere a code like the example below.

<details>
<summary>Example code</summary>

```javascript
import LoadBar from 'components/load-bar';
import PhoneNumberControl from 'components/phone-number-control';
import GroupedSelectControl from 'components/grouped-select-control';
import CustomSelectControl from 'components/custom-select-control';

...

const [ option, setOption ] = React.useState< any >();
const [ phone, setPhone ] = React.useState( '' );

...

<LoadBar />
<PhoneNumberControl
	value={ phone }
	onChange={ setPhone }
	label="Phone Number Control"
/>
<GroupedSelectControl
	label="Grouped Select Control"
	options={ options }
	groups={ groups }
	value={ option }
	onChange={ setOption }
	searchable
	placeholder="Select an option"
/>

<CustomSelectControl
	placeholder="Select an option"
	label="Custom Select Control"
	options={ options }
/>
```
</details>

#### Screenshot and videos

<table><tr>
<td>custom-select-control</td><td>

https://user-images.githubusercontent.com/7670276/230898888-82866f5a-5ee2-4d5f-87a9-23ca96a7b56c.mov
</td><td>

https://user-images.githubusercontent.com/7670276/230898884-d620a982-5dd0-4d7b-b48d-6147815eaccb.mov
</td></tr><tr><td colspan=3>

[...]

PR #6167, opened 2023-04-27
Fetch, cache, and localize for JS onboarding wizard fields data from PHP, collect and prefill MCC

### LINKED ISSUE #4853: Progressive Onboarding: Collect the MCC and pre-fill it in Stripe's KYC

### LINKED ISSUE #6104: Preload progressive onboarding server data
### Description
For the new onboarding flow we do two requests to the server, for business type and MCC data, and use them as options on multiple form fields.

Right now, while the data is requested the component just doesn't work, because there are no options. We need to preload that data ASAP, to reduce the possibility of showing a loading indicator later.

But also handle the loading state, to show a loading indicator, on those components, if for any reason the data is still loading.

If the data could not be loaded, the form cannot be completed, so we should handle the error. Here I see two options:
- Show a message to inform the merchant, recommending trying again later or contact support.
- Fail silently and redirect to full onboarding as usual.

### Acceptance criteria
- Both business type and MCC data are loaded as soon as possible.
- Affected components show a loading state if the data is still not available.
- Handle loading data error.

vladolaru:
Fixes #6104
Fixes #4853 

This PR ended up handling two issues in the interest of speed of development and reduced back-and-forth.

First, we use the REST endpoint introduced by [this paired server PR](https://github.com/Automattic/woocommerce-payments-server/pull/3310) to fetch all the fields data in one request (business types and industry/MCC fields, right now). This data is cached in the DB for a week. 

Furthermore, the data is provided directly from PHP for JS consumption rather than being fetched async through a client REST endpoint. By doing it this way, we reduce the client logic complexity (no need for loading states for controls) and increase the client-side UI speed since the data is available immediately. We do end up with a higher HTML payload, but when I considered the current state of the markup (huge, mainly due to the WP media dependency), I decided that the downside in performance is marginal.

Second, we implement the industry/MCC onboarding wizard control with some adjustments to the `GroupedSelectField` component (introduced in [this PR](https://github.com/Automattic/woocommerce-payments/pull/5895)). The adjustments are needed so we are able to use bespoke select options that are custom-ordered (groups and items in a group). The MCC field has these requirements to allow for a focus on the merchant experience rather than programmatic generation.

Here's what the new onboarding prototype MCC control looks like:
Screenshot 2023-05-02 at 12 01 17
Screenshot 2023-05-02 at 12 01 56

The collected MCC is passed along with the rest of the onboarding data to the server for processing, storage, and prefilling in the Stripe account (implemented in [the paired PR](https://github.com/Automattic/woocommerce-payments-server/pull/3310)).

#### Testing instructions

* On a dev site with WooCommerce Payments, checkout this PR's branch: `add/6104-preload-and-cache-onboarding-wizard-fields-data`
* On the WCpay server used by this client, checkout [the paired PR's](https://github.com/Automattic/woocommerce-payments-server/pull/3310) branch: `add/c6104-single-global-route-for-onboarding-wizard-fields-data`
* Make sure you force the WCPay client to act as disconnected from the server via the [WCPay Dev Tools admin page](http://localhost:8082/wp-admin/admin.php?page=wcpaydev)
* Go to [the onboarding prototype page](http://localhost:8082/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fonboarding-prototype)
* On the third step, you should see the industry control with all the grouped options (labeled "What type of goods or services does your business sell?")
* Select, search, clear the search, close and reopen, select again; basically, smoke test the control
* At the same time, smoke test the business type and structure control (labeled "Where is your business legally registered?")
* Go through the entire onboarding, and you should end up with a Stripe account that has the MCC prefilled (check the Stripe account's dashboard for this); for example, if you chose the "Consulting" option, the Stripe account should have "Professional services (other)" as the industry (see screenshot below)
Screenshot 2023-05-02 at 12 10 18

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6167` or branch name `add/6104-preload-and-cache-onboarding-wizard-fields-data` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/6104-preload-and-cache-onboarding-wizard-fields-data) :rocket:

---

Build info: 

- Latest commit: 8b3fad8c7f013d42b197f7f26372c42d374fd391
- Build time: 2023-05-04 07:11:06 UTC

client/globals.d.ts
ismaeldcom, line 1: We use `globals.d.ts` to define global properties like `window`, a good example is `wcpaySettings`. But local types should be defined within their directory, in this case it could be `client/onboarding-prototype/`, or in `client/types` if it can be used in multiple places.

client/components/grouped-select-control/index.tsx
```diff
@@ -5,81 +5,62 @@ import React, { useRef, useState } from 'react';
...
 
 /**
  * Internal Dependencies
  */
 import './style.scss';
+import { OnboardingFields } from 'wcpay/onboarding-prototype/types';
 
-export interface Item {
+export interface ListItem {
+	type: string;
```
ismaeldcom: `type` is not used anywhere, as `group` and `items` is used to differentiate them.

client/components/grouped-select-control/index.tsx
```diff
@@ -5,81 +5,62 @@ import React, { useRef, useState } from 'react';
...
 import classNames from 'classnames';
 import { __ } from '@wordpress/i18n';
-import { useSelect } from 'downshift';
+import { useSelect, UseSelectState } from 'downshift';
 
 /**
  * Internal Dependencies
  */
 import './style.scss';
+import { OnboardingFields } from 'wcpay/onboarding-prototype/types';
```
ismaeldcom: We can get rid of this unused export.

ismaeldcom (commented):
Thanks for going the extra mile by updating and unifying business structures with the MCC and simplifying the implementation by adding the onboarding data to `wcpaySettings`.

>We do end up with a higher HTML payload, but when I considered the current state of the markup (huge, mainly due to the WP media dependency), I decided that the downside in performance is marginal.

We can squeeze it a bit by removing unneeded properties from `mccs_display_tree` like: `type` and item `id`, although that will be fine-tuning it too much 😄 Why are we using the string based ID instead of `mcc` code? 🤔 I understood we will be using MCC code on every place (client/server).

>The adjustments are needed so we are able to use bespoke select options that are custom-ordered (groups and items in a group).

The adjustments look good, but weren't needed to have ordered groups/options, that was already supported with the previous implementation.

I left a few small comments and there are also a couple of issues to solve. Everything else looks good so far.
- Four JS tests are failing.
- Some items on `itemsToRender.map` are rendered with the same `key`, something we should avoid.
- Toggling groups doesn't seem to work properly after the adjustments, if you collapse `Most popularar` and expend `Retail`, an additional `Clothing and accesories` option gets added. Probably related with the previous point.

client/components/grouped-select-control/index.tsx
vladolaru: Yes. I have removed it.

vladolaru:
Thank you @ismaeldcom for reviewing this PR. Much appreciated. Now, let's get going :D 

> Four JS tests are failing.

Fix those (after some headaches to understand what is going on) in https://github.com/Automattic/woocommerce-payments/pull/6167/commits/7b357b33acc083a25894bc34ee23a21501409975

> * Some items on itemsToRender.map are rendered with the same key, something we should avoid.
> * Toggling groups doesn't seem to work properly after the adjustments, if you collapse Most popularar and expend Retail, an additional Clothing and accesories option gets added. Probably related with the previous point.

Yes, these are related, but I already fixed them on server to have unique IDs. I don't see this behavior anymore. Could you double check that you don't have a old cached version of the fields data from the server (probably best to delete the `wcpay_onboarding_fields_data__en_US` DB option on the client)

> Why are we using the string based ID instead of mcc code? 🤔 I understood we will be using MCC code on every place (client/server).

Yeah. That was the original plan from a while back but the reality has changed. I've opted for Stripe-like string IDs so we can capture (and store) the merchant intent in relation to his or hers business MCC. This is needed because in the options we have many entries that translate to the same MCC (number). By using unique string IDs for these entries we know what option they chose. For example, we can differentiate between someone who sells "Books" vs "Music or other media" (these have the same 5815 MCC). Things gain more weight when dealing with "Regulated and age-restricted products". Here "Pharmacies or pharmaceuticals", "Supplements or nutraceuticals", "Marijuana dispensaries", and "Marijuana-related products" all share the same 5912 MCC 😄 
I think this signal is an important signal for us internally (analytics, risk, etc), and I believe this is the reason Stripe also opted to do it like that in their KYC.

> The adjustments look good, but weren't needed to have ordered groups/options, that was already supported with the previous implementation.

Hmm. You are right. I don't know why it stuck into my mind that there was a sort when constructing the `mergedList` in the previous implementation state.

I will tackle the rest of your comments also.

client/components/grouped-select-control/index.tsx
vladolaru: Yes. I was aiming to be more declarative about the intent, but I refactored things in https://github.com/Automattic/woocommerce-payments/pull/6167/commits/204fa68c457dc0fcaabdb733439b20747fef706c

client/globals.d.ts
vladolaru: I agree, but how could I manage the current case where `wcpaySettings` has `onboardingFieldsData`? If I don't hoist the `Country` and `MccsDisplayTreeItem` types up to the global level, I can't use them to type things in `wcpaySettings`.

How would you approach this?

ismaeldcom: As those types are specific for `wcpaySettings` you can declare them [inside of it](https://github.com/Automattic/woocommerce-payments/blob/e564f363ca9b4b6f1e0c6f8c0d3a3493418fed9a/client/globals.d.ts#L7). And if you need to reference its type later, just use `typeof wcpaySettings...`

ismaeldcom (approved):
Everything works as expected, so I'm preapproving it now! You can merge after solving the `globals.d.ts` comment.

>Yes, these are related, but I already fixed them on server to have unique IDs. I don't see this behavior anymore. Could you double check that you don't have a old cached version of the fields data from the server

That was it! 🙂 

>I think this signal is an important signal for us internally (analytics, risk, etc), and I believe this is the reason Stripe also opted to do it like that in their KYC.

Totally agree with you, thanks for the explanation! I wasn't aware of the later decision.

vladolaru:
Excellent!

> Totally agree with you, thanks for the explanation! I wasn't aware of the later decision.

It was easy to slip by as I just took it and implemented it in [this PR](https://github.com/Automattic/woocommerce-payments-server/pull/3270) 😀 🐐

PR #6170, opened 2023-04-27
Supply correct payment method instance to process_redirect_payment

### LINKED ISSUE #5781: UPE: Purchasing same item twice in a row gives intent error
### Describe the bug
If you purchase an item then add the same item to your cart again and try to purchase it again, you get this error:

> Error: The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.

### To Reproduce

1. Make sure UPE is enabled in your store.
2. Add a simple product to your cart and complete checkout.
3. Add the item to your cart again and try to complete checkout.
4. Receive error.

### Additional context
May be fixed by #5653

### LINKED ISSUE #5854: The wrong gateway prevents process_redirect_payment method from being executed fully
### Describe the bug
There's a class `WC_Payments_UPE_Checkout` in the codebase, which depends on the `UPE_Payment_Gateway` through the `$gateway` variable. 

The `$gateway` variable is initialised with the card gateway object for the split UPE [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L467). Then, a group of hooks is defined in [the constructor of the `WC_Payments_UPE_Checkout`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L77-L89) that relies on the `$gateway` during their execution. 

[There's also the hook to change the `$gateway` value to differ from the `card`
](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L75), and this was expected to be executed before any of the hooks described previously. But it turned out that the gateway is set [only when rendering the payment fields](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-split-payment-gateway.php#L316).

Effectively, the code for [some part of these hooks](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L77-L89) is executed with a wrong gateway (card), instead of a proper one. [`maybe_process_upe_redirect`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L80) is one of them and was the starting point of this issue.

### To Reproduce
Turn on the debug mode on your local for the merchant site and set the debug to listen on [this line](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L768).

1. Use the shortcode checkout and one of the non-card elements that require third-party transaction confirmation to purchase a product (e.g. `giropay`). 
2. After confirming the transaction on t third-party site, you will be redirected to the _**Order received**_ page. At this point, the debugger set previously should stop the code execution. 
3. Go inside the `get_selected_payment_method` and confirm that `$payment_method_type` equals to `giropay`, whereas `$this->stripe_id` equals to `card`. 
4. Confirm that because of the above, the execution flow is escaped at [this stage](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L770), while it should be processed further.

mdmoore:
Fixes #5854, Fixes #5781 

The crux of the the problem here is an [early return in `process_redirect_payment`](https://github.com/Automattic/woocommerce-payments/blob/9fa0bbbe007c1072776f13b401aeb530a4403f37/includes/payment-methods/class-upe-payment-gateway.php#L768). It's this early return that prevents the UPE payment intent from being removed from the session, causing the issue described in #5781. `UPE_Split_Payment_Gateway` extends `UPE_Payment_Gateway` and overrides `get_selected_payment_method` to redefine the logic with a simple check against `$this->stripe_id`. The problem is that `$this->stripe_id` is always `card` due to hooks in the parent class resetting the stripe id for it's own uses. So redirect payment methods such as `giropay` are always compared to `card` and we return `false`. That's fine as that's what `UPE_Payment_Gateway` needs and `UPE_Split_Payment_Gateway` doesn't need the `stripe_id` at this point in the execution. We really only need an instance of the payment method which we can get from the `upe_payment_method_map` via `WC_Payments::get_payment_method_by_id()`. 

### Testing

The goal is to test that `process_redirect_payment` fully completes. This would mean that [`remove_upe_intent_from_session()`](https://github.com/Automattic/woocommerce-payments/blob/9fa0bbbe007c1072776f13b401aeb530a4403f37/includes/payment-methods/class-upe-payment-gateway.php#L792) runs as well, which covers testing for both #5854 and #5781. Testing with Xdebug breakpoints has been problematic because the hooks that initiate this process fire multiple times, making it confusing as to whats actually occurring. I've found it best to just use `Logger::log()` to ensure the process completes.

1. Add a log statement immediately after [this line](https://github.com/Automattic/woocommerce-payments/blob/9fa0bbbe007c1072776f13b401aeb530a4403f37/includes/payment-methods/class-upe-payment-gateway.php#L829) indicating that the process has completed, e.g. `Logger::log( 'process_redirect_payment has completed' )`
2. Enable Split UPE Checkout via WCPay Dev Tools.
3. Go to the client DB and delete all rows in the `wp_woocommerce_sessions table`. This will ensure no `upe_payment_intent` is present yet.
4. Complete a checkout using a redirect payment method such as Giropay, making sure to **Authorize Test Payment** at Stripe.
5. Repeat step 4 at least 2 more times, using the same product and payment method. Reports have indicated that the payment intent error in #5781 requires 3 purchases.
6. Check logs for the message used in step 1. 
7. Ensure that the following error doesn't occur on the checkout page: 

>Error: The parameter application_fee_amount cannot be updated on a PaymentIntent after a capture has already been made.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6170` or branch name `fix/5854-use-correct-payment-method-id` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/5854-use-correct-payment-method-id) :rocket:

---

Build info: 

- Latest commit: f8340d4b1b4af7641741d75d708560a93cd723ac
- Build time: 2023-05-10 17:59:22 UTC

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -160,13 +160,31 @@ public function should_use_stripe_platform_on_checkout_page() {
...
+		$payment_method = $this->get_selected_payment_method( $payment_method_type );
+		if ( ! $payment_method ) {
+			return;
 		}
 
-		return $this->payment_method;
-	}
+		$payment_method_title = $payment_method->get_title( $payment_method_details );
 
+		$payment_gateway = 'card' === $payment_method->get_id() ? self::GATEWAY_ID : self::GATEWAY_ID . '_' . $payment_method_title;
```
FangedParakeet:
```suggestion
		$payment_gateway = $this->id;
```
~I think you can safely trust the `id` parameter at this stage of the execution. However, when I thought a little more about this, since this is the only difference between this function and the function it overrides in `UPE_Payment_Gateway`, why don't you just change this line over [here](https://github.com/Automattic/woocommerce-payments/blob/fix/5854-use-correct-payment-method-id/includes/payment-methods/class-upe-payment-gateway.php#L895) to use `$this->id` instead of `self::GATEWAY_ID`? Then we wouldn't need to override this function at all.~

Actually, disregard this entirely. This is always being called from the card gateway instance inside `WC_Payments_UPE_Checkout`: `stripe_id` will always be `card` and `id` will always be `woocommerce_payments`, so this will not work. 🤦

FangedParakeet (approved):
Tested with a few different LPMs and all the metadata appears to be lining up now: transaction details look correct, with the correct order details and payment method details, and I was unable to produce the well-known and dreaded "application_fee_amount" error message. 

Changes LGTM. It's worth noting that for now, the parent function `UPE_Payment_Gateway::set_payment_method_title_for_order` is still required for some payment flows with the split-UPE, but when the deferred intent UPE reaches fully operational parity, we will need to remove it. 

But that'll be a story for another day. For today, let's :shipit:.

isaacshed:
Unfortunately your order cannot be processed as the originating bank/merchant has declined your transaction. Please attempt your purchase again.


Quickteller Business Payment Gateway

PR #6175, opened 2023-04-28
Update Platform Checkout Tracking conditions

malithsen:

With the recent changes in the usage tracking( p2: p2y3YZ-6L8-p2) , we can now collect usage data specific to WooPay regardless of the usage tracking opt-in status. This PR does the following so that we can receive `platform_checkout_enabled` and `platform_checkout_disabled` events regardless of usage tracking opt-in status.

- A minor refactor to change where additional event properties are being recorded.
- Introduce `maybe_record_admin_event` to record events emitted from the admin area
- Use `maybe_record_admin_event` to capture `platform_checkout_enabled` and `platform_checkout_disabled` events.

#### Testing instructions

Regression Test
- Make sure WooPay is enabled.
- Add a product to the cart and go to the checkout page.
- In ~5 minutes, check the Tracks Live View for the `woocommerceanalytics_order_checkout_start` event.
- Disable WooPay
- Add a product to the cart and go to the checkout page.
- `woocommerceanalytics_order_checkout_start` event should not appear in the Tracks Live view.

Check Admin events

- Optin to usage tracking in WooCommerce -> Settings -> Advanced -> WooCommerce.com -> Enable Tracking.
- Do the tests mentioned in this PR https://github.com/Automattic/woocommerce-payments/pull/3816. Note that setting `_wcpay_feature_platform_checkout` flag is no longer needed.
- Opt-out of tracking by disabling the checkbox set in step 1.
- Repeat the tests mentioned in https://github.com/Automattic/woocommerce-payments/pull/3816
- `wcadmin_platform_checkout_enabled` and `wcadmin_platform_checkout_disabled` events should still be registered.

Optional
- For good measure smoke test checkout with split upe enabled/disabled

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -35,6 +35,7 @@ use WCPay\Payment_Methods\UPE_Payment_Gateway;
 use WCPay\Payment_Methods\Link_Payment_Method;
 use WCPay\WooPay\WooPay_Order_Status_Sync;
 use WCPay\WooPay\WooPay_Utilities;
+use WCPay\WooPay_Tracker;
 use WCPay\Session_Rate_Limiter;
 use WCPay\Tracker;
 
@@ -183,6 +184,13 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	protected $woopay_util;
 
 	/**
+	 * WooPay Tracker.
+	 *
+	 * @var WooPay_Tracker
+	 */
+	protected $woopay_tracker;
+
+	/**
 	 * WC_Payment_Gateway_WCPay constructor.
 	 *
 	 * @param WC_Payments_API_Client               $payments_api_client             - WooCommerce Payments API client.
@@ -192,6 +200,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 * @param WC_Payments_Action_Scheduler_Service $action_scheduler_service        - Action Scheduler service instance.
 	 * @param Session_Rate_Limiter                 $failed_transaction_rate_limiter - Rate Limiter for failed transactions.
 	 * @param WC_Payments_Order_Service            $order_service                   - Order class instance.
+	 * @param WooPay_Tracker                       $woopay_Tracker                  - WooPay Tracker instance.
 	 */
 	public function __construct(
 		WC_Payments_API_Client $payments_api_client,
@@ -200,7 +209,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		WC_Payments_Token_Service $token_service,
 		WC_Payments_Action_Scheduler_Service $action_scheduler_service,
 		Session_Rate_Limiter $failed_transaction_rate_limiter = null,
-		WC_Payments_Order_Service $order_service
+		WC_Payments_Order_Service $order_service,
+		WooPay_Tracker $woopay_tracker
 	) {
 		$this->payments_api_client             = $payments_api_client;
 		$this->account                         = $account;
@@ -209,6 +219,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 		$this->action_scheduler_service        = $action_scheduler_service;
 		$this->failed_transaction_rate_limiter = $failed_transaction_rate_limiter;
 		$this->order_service                   = $order_service;
+		$this->woopay_tracker                  = $woopay_tracker;
 
 		$this->id                 = static::GATEWAY_ID;
 		$this->icon               = plugins_url( 'assets/images/payment-methods/cc.svg', WCPAY_PLUGIN_FILE );
@@ -1770,7 +1781,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	public function update_is_woopay_enabled( $is_woopay_enabled ) {
 		$current_is_woopay_enabled = 'yes' === $this->get_option( 'platform_checkout', 'no' );
 		if ( $is_woopay_enabled !== $current_is_woopay_enabled ) {
-			wc_admin_record_tracks_event(
+			$this->woopay_tracker->maybe_record_admin_event(
 				$is_woopay_enabled ? 'woopay_enabled' : 'woopay_disabled',
 				[ 'test_mode' => WC_Payments::mode()->is_test() ? 1 : 0 ]
 			);
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -200,7 +200,6 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 * @param WC_Payments_Action_Scheduler_Service $action_scheduler_service        - Action Scheduler service instance.
 	 * @param Session_Rate_Limiter                 $failed_transaction_rate_limiter - Rate Limiter for failed transactions.
 	 * @param WC_Payments_Order_Service            $order_service                   - Order class instance.
-	 * @param WooPay_Tracker                       $woopay_Tracker                  - WooPay Tracker instance.
 	 */
 	public function __construct(
 		WC_Payments_API_Client $payments_api_client,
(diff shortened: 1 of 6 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6175` or branch name `update/platform-checkout-admin-tracks` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/platform-checkout-admin-tracks) :rocket:

---

Build info: 

- Latest commit: 2c638f3a87a6b20f35ac75c13b839b192a51a587
- Build time: 2023-05-12 15:30:02 UTC

malithsen:
Hi @pierorocca 
@leonardola noted the following while reviewing the changes in this PR to update `platform_checkout_enabled` and `platform_checkout_disabled` events so that they are recorded regardless of the merchant's Usage Tracking opt-in status.

```
>  If you disable WooCommerce Usage Tracking we will not collect data from your site through our analytics tool, but we may still collect similar data if we need it to operate a service we provide to you, such as WooCommerce Payments.

To me that means no analytics but we will collect just enough data via other tools like WC Pay Server to make payments possible.
```

I know we've previously discussed that these two events would fall under product improvement data, but I could not find a good source to share with Leo to confirm it. Could you please confirm that we are okay to record these events for all WCPay merchants with the updated policy.

pierorocca:
@malithsen the main source is Neil's writeup in this P2 https://automattlock.wordpress.com/2023/04/18/making-our-wctracker-usage-tracking-page-more-accurate/

And to double check, I asked him to confirm the events in the spreadsheet against the usage tracking terms:
image

includes/class-platform-checkout-tracker.php
```diff
@@ -154,20 +165,22 @@ public function should_enable_tracking() {
...
+
+		$user = wp_get_current_user();
 
 		// We don't want to track user events during unit tests/CI runs.
 		if ( $user instanceof \WP_User && 'wptests_capabilities' === $user->cap_key ) {
 			return false;
 		}
 
-		if ( ! $this->should_enable_tracking() ) {
+		if ( ! $is_admin_event && ! $this->should_enable_tracking() ) {
```
leonardola: shouldn't this check actually happen in the `maybe_` functions?

includes/class-platform-checkout-tracker.php
```diff
@@ -197,9 +210,18 @@ public function tracks_record_event( $user, $event_name, $properties = [] ) {
...
 
+		//phpcs:ignore WordPress.Security.ValidatedSanitizedInput
+		$properties['_lg']       = isset( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) ? $_SERVER['HTTP_ACCEPT_LANGUAGE'] : '';
+		$properties['blog_url']  = $site_url;
+		$properties['blog_id']   = \Jetpack_Options::get_option( 'id' );
 		$properties['user_lang'] = $user->get( 'WPLANG' );
 
+		// Add event property for test mode vs. live mode events.
+		$properties['test_mode']     = WC_Payments::mode()->is_test() ? 1 : 0;
+		$properties['wcpay_version'] = get_option( 'woocommerce_woocommerce_payments_version' );
```
leonardola: Shouldn't we actually use the `WCPAY_VERSION_NUMBER` const instead of `get_option` as it will not change during downgrades and it will be calling the DB every time?

leonardola (commented):
Code generally looks good but when I added a debugger breakpoint to [this line](https://github.com/Automattic/woocommerce-payments/blob/update%2Fplatform-checkout-admin-tracks/includes/class-platform-checkout-tracker.php#L176) and visited the checkout page as a shopper I noticed `woocommerceanalytics_platform_checkout_express_button_offered` is also being recorded.

Also I left a couple of questions below

includes/class-platform-checkout-tracker.php
malithsen: `is_admin_event` identifies the events coming through the `maybe_record_admin_event`. This check makes sure that those requests bail out early without considering the conditions in `should_enable_tracking` method, as we want these events to be tracked always. 
But after you pointed it out, I realized it would make sense to consolidate all the validations in `should_enable_tracking`, so I moved this check inside that function in https://github.com/Automattic/woocommerce-payments/pull/6175/commits/36da0f3ddfe323c5e54646fd76c90f634d02fdee

includes/class-platform-checkout-tracker.php
malithsen: Good point. Thanks! I updated it in https://github.com/Automattic/woocommerce-payments/pull/6175/commits/390939be3d952a255ffbf960a38535dc7ebf2d4e

malithsen:
Thanks for the review @leonardola! I addressed your code comments above.

> Code generally looks good but when I added a debugger breakpoint to [this line](https://github.com/Automattic/woocommerce-payments/blob/update%2Fplatform-checkout-admin-tracks/includes/class-platform-checkout-tracker.php#L176) and visited the checkout page as a shopper I noticed woocommerceanalytics_platform_checkout_express_button_offered is also being recorded.

I misunderstood this when you first pointed it out on Slack. If you have WooPay enabled, what you see is the correct behavior. All Tracks events originating due to user behavior work the same way as before. That is, they are tracked when WooPay is enabled. 

I made some notes about existing tracking conditions [here](https://woopayp2.wordpress.com/woopay-data/) if it helps. Note that "Merchant Events" section should be updated following this PR.

Let me know if you have other questions.

includes/class-woopay-tracker.php
```diff
@@ -98,33 +105,54 @@ public function ajax_tracks() {
...
+			return true;
+		}
+
+		// For all other events ensure:
+		// 1. Only site pages are tracked.
+		// 2. Site Admin activity in site pages are not tracked.
+		// 3. Site page tracking is only enabled when WooPay is active.
+
 		// Track only site pages.
 		if ( is_admin() && ! wp_doing_ajax() ) {
```
leonardola, line 157: We can probably remove this check now

includes/class-platform-checkout-tracker.php
leonardola: What I meant is that it sounds strange to have checks in the function with a name that makes it look like it will execute something no matter what and have no checks in the functions with a name that indicate it may or may not execute. It's just a nitpicking and can be ignored.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1770,7 +1771,7 @@ public function update_option( $key, $value = '' ) {
 	public function update_is_woopay_enabled( $is_woopay_enabled ) {
 		$current_is_woopay_enabled = 'yes' === $this->get_option( 'platform_checkout', 'no' );
 		if ( $is_woopay_enabled !== $current_is_woopay_enabled ) {
-			wc_admin_record_tracks_event(
+			WC_Payments::woopay_tracker()->maybe_record_admin_event(
 				$is_woopay_enabled ? 'woopay_enabled' : 'woopay_disabled',
```
leonardola, line 1775: Just confirming that the events are now `wcadmin_woopay_enabled` and `wcadmin_woopay_disabled` and not `platform_checkout_enabled` and `platform_checkout_disabled`

leonardola (approved):
It looks good to me. Just added a comment about a check that does not seem necessary anymore and about the event name

includes/class-woopay-tracker.php
malithsen: This check is still needed. When an admin user goes through WooPay checkout, resulting events should not be tracked.

includes/class-wc-payment-gateway-wcpay.php
malithsen: Yes. This is correct. These events are renamed as a part of recently merged naming convention updates https://github.com/Automattic/woocommerce-payments/pull/6061

malithsen:
> What I meant is that it sounds strange to have checks in the function with a name that makes it look like it will execute something no matter what and have no checks in the functions with a name that indicate it may or may not execute. It's just a nitpicking and can be ignored.

Ah I see what you mean now. `maybe_` prefix was added to the top-level functions to indicate that they may not always result in actually tracking the event due to the validations that get triggered further down. 

I'll merge this as is but I acknowledge that it can be refactored to make it more clear.

pierorocca:
Nice work all!

Issue #2422, opened 2021-07-14
Refactor WC_Payments_API_Intention to support Setup Intents


FangedParakeet:
Currently, the function `WC_Payments_API_Client::get_intent` returns a `WC_Payments_API_Intention`, whereas the function `WC_Payments_API_Client::get_setup_intent` simply returns an array representing the `WP_REST_Response` sent back from the Payments Server. This creates a lot of unnecessary branching logic in sections such as this one: 

https://github.com/Automattic/woocommerce-payments/blob/ca0d60fdabdc9501c165fc8b423b23bc629043dd/includes/class-wc-payment-gateway-wcpay.php#L955-L980

The two have very similar fields, with Setup Intents differing by lacking any charge objects, so I think it should be possible to encapsulate and deserialize both of these REST responses into `WC_Payments_API_Intention` instances. I don't think this will require a massive effort, but will have some impact on both the UPE payment gateway, the existing WCPay card gateway, and all relevant tests in between.

RadoslavGeorgiev:
Just dropping a note here, as this is something we'll likely be changing soon.

---

IMO it would make sense to have two separate classes in order to be stricter. It'd suggest the following:

1. `WC_Payments_API_Intention` should become `WC_Payments_API_Payment_Intention`.
2. The old `WC_Payments_API_Intention` class (name) should be used as a base for both Payment and Setup intentions.
3. Introduce a new `WC_Payments_API_Setup_Intention` class.

```mermaid
classDiagram
    WC_Payments_API_Intention <|-- WC_Payments_API_Payment_Intention
    WC_Payments_API_Intention <|-- WC_Payments_API_Setup_Intention
    WC_Payments_API_Intention : +string id
    WC_Payments_API_Intention : +string status
    WC_Payments_API_Intention : ...
    WC_Payments_API_Intention: +get_id()
    WC_Payments_API_Intention: +get_status()
    WC_Payments_API_Intention: ....()
    class WC_Payments_API_Payment_Intention{
      +array charge
      +get_charge()
    }
    class WC_Payments_API_Setup_Intention{
      +array payment_method_options
      +get_payment_method_options()
    }
```

In addition, those two classes need to be added as response formats to the correct request classes.

RadoslavGeorgiev:
Another note: Let's add a proper request class once this is done.

naman03malhotra:
After a discussion with @htdat we have the following timelines for this issue:
[...]

PR #6234, opened 2023-05-05
Migration of simple requests to request classes

### LINKED ISSUE #6217: Migrate Simple and Straightforward Requests to a Generic Request Class
### Description
 As part of the request classes migration, we need to migrate a set of requests that are relatively simple and straightforward to a generic request class to reduce code duplication and improve code organization.

Here is the breakdown of the tasks:

- Analyze the simple and straightforward requests and determine the best approach for migrating them to a generic request class.
- Implement a generic request class that can handle the identified requests.
- Update the relevant code to use the new generic request class.
- Test the functionality of the requests after the migration.
### Acceptance criteria

- All simple and straightforward requests are successfully migrated to the new generic request class.
- The code is properly refactored to use the new generic request class, with no code duplication.

### Testing instructions

- Use manual testing to verify that all simple and straightforward requests are working as expected after the migration.
- Make sure that all unit tests are passing.

zmaglica:
Fixes #6217

This pull request introduces a new Get_Request class that simplifies the execution of straightforward requests towards the WCPay server. By utilizing this class, existing simple requests are migrated to dedicated request classes. 

#### Testing instructions

* Make sure that all tests are passing
* Make sure that all migrated requests are working as expected.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6234` or branch name `dev/6217-generic-get-request-class` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=dev/6217-generic-get-request-class) :rocket:

---

Build info: 

- Latest commit: f27c32865bd38d7e44e3591d58ae29f5a506b807
- Build time: 2023-07-20 09:50:59 UTC

includes/core/server/class-request.php
```diff
@@ -663,4 +678,64 @@ public function validate_user_name( string $user_name ) {
...
+	/**
+	 * Validates API endpoint or route.
+	 *
+	 * @param string $api_route API route to validate.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function validate_api_route( string $api_route ) {
+
+		// I will use array keys check validate is request to improve performance. Maybe we can move this outside of this function.
```
jessy-p: Can move this out as mentioned and remove this comment

zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/ad6dc07e9c994c0c3a97925a0350f54ab67b58e5

includes/core/server/request/class-get-request.php
```diff
@@ -0,0 +1,81 @@
...
+ * Generic GET WCPay Server Request.
+ */
+class Get_Request extends Request {
+	/**
+	 * The request's API.
+	 *
+	 * @var string Check WCPay\Core\Server\APIs.
+	 */
+	private $api = ''; // Just to be safe in case someone misses to add it. It will be validated by default.
+
```
RadoslavGeorgiev, line 23: There are a few too many empty lines here. Let's limit it to a single one.

includes/core/server/request/class-get-request.php
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Set api route.
+	 *
+	 * @param string $api Api route.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function set_api( string $api ) {
+		$this->validate_api_route( $api );
```
RadoslavGeorgiev, line 32: Is this required considering that `Request::apply_filters()` will run `validate_api_route` as well?

includes/core/server/request/class-get-request.php
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Set ID:
+	 *
+	 * @param mixed $id Set id.
+	 *
+	 * @return void
+	 */
+	public function set_id( $id ) {
+		// We won't validate it in detail. Just to be sure it is not empty.
+		if ( $id ) {
```
RadoslavGeorgiev: Should we remove this check, making it clear that if `set_id` is called, an ID is actually expected?

includes/core/server/request/class-get-request.md
```diff
@@ -0,0 +1,27 @@
...
+|-----------|--------------------------|:---------:|:--------:|:-------------:|
+| `id`      | `set_id( $id )`          |    Yes    |    No    |       -       |
+| `api`     | `set_api( string $api )` |     -     |   Yes    |       -       |
+
+
+
+## Example:
+
+```php
+$request = Get_Request::create( $id );
```
RadoslavGeorgiev, line 36: Let's make sure that we point other developers towards `Request::get()` instead of creating the `Get_Request` directly? I there is no strict reason to avoid it, but IMO this way we'd have higher consistency.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -745,7 +746,10 @@ public function process_redirect_payment( $order_id, $intent_id, $save_payment_m
 				$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
 				$error                  = $intent->get_last_payment_error();
 			} else {
-				$intent                 = $this->payments_api_client->get_setup_intent( $intent_id );
+				$setup_intent_request = Request::get( WC_Payments_API_Client::SETUP_INTENTS_API, $intent_id );
```
RadoslavGeorgiev, line 747: Feel free to ignore this comment, but intent requests should be all returning the right format. Maybe we can change that together with https://github.com/Automattic/woocommerce-payments/issues/2422.

RadoslavGeorgiev (approved):
Changes look great overall, I left a few small notes. Feel free to fix & 🚢 

Note: I have not looked into tests.

includes/core/server/request/class-get-request.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/2f8597db1a38058544e88271b1d953148d907b69

includes/core/server/request/class-get-request.php
zmaglica: The reason why I left is to prevent filter functions to set incorrect API method/route.

includes/core/server/request/class-get-request.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/commit/2f8597db1a38058544e88271b1d953148d907b69

includes/core/server/request/class-get-request.md
zmaglica: I added more examples with Request class.

includes/payment-methods/class-upe-payment-gateway.php
zmaglica: I think that this is probably a good option to move it out of this PR since this PR addresses the simple requests

htdat: I see that Rado left a note here https://github.com/Automattic/woocommerce-payments/issues/2422#issuecomment-1545394463 so I think it's fine to keep the current code in this PR.

includes/core/server/class-request.php
```diff
@@ -663,4 +754,24 @@ public function validate_user_name( string $user_name ) {
...
 	}
+
+	/**
+	 * Validates API endpoint or route.
+	 *
+	 * @param string $api_route API route to validate.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function validate_api_route( string $api_route ) {
```
htdat, line 766: IMO, this method needs to be tested a bit extensively. 

Can we have a dedicated unit test using [data provider](https://docs.phpunit.de/en/10.2/writing-tests-for-phpunit.html#data-providers) to test this method? 

It seems we're not having a specific test class/file for this abstract class `WCPay\Core\Server\Request`, and you're writing some unit tests for this specific method in `Get_Request_Test`. I think it's OK to have the test like that in `Get_Request_Test` for now.

includes/core/server/request/class-get-request.md
htdat: @zmaglica - I think Rado's point makes sense. Although it's possible to use `Get_Request::create`, I don't think we should make it as an example here. Can you remove it from this MD file, or at least move it to the bottom as the last example with a note like `Not recommended`.

includes/core/server/class-request.php
```diff
@@ -272,6 +342,24 @@ public function format_response( $response ) {
...
 
+	/**
+	 * Used to validate passed ID from the constructor. You can easily override it if you need custom id handling in your classes.
+	 *
+	 * @param mixed $id ID parameter.
+	 *
+	 * @return void
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function set_request_route_id_parameter( $id ) {
```
htdat: This is a pretty critical method to ensure the integrity of all request classes. I think we should not expose it as `public`, instead limit it to `protected`. 
```suggestion
	protected function set_request_route_id_parameter( $id ) {
```

tests/unit/core/server/request/test-class-get-request.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+	public function test_request_will_be_created_correctly_passed_api_endpoint_has_additional_route_arguments() {
+		$request = new Get_Request( $this->mock_api_client, $this->mock_wc_payments_http_client );
+		$request->set_api( WC_Payments_API_Client::DEPOSITS_API . '/foo' );
+		$this->assertSame( 'GET', $request->get_method() );
+		$this->assertSame( WC_Payments_API_Client::DEPOSITS_API . '/foo', $request->get_api() );
+	}
+
+	public function test_get_method_from_parent_request_class_will_create_get_request() {
+		$request = Request::get( WC_Payments_API_Client::DEPOSITS_API );
```
htdat: May add another similar test with the second parameter `$id` too 
```php
	public function test_get_method_with_id_from_parent_request_class_will_create_get_request() {
		$request = Request::get( WC_Payments_API_Client::DEPOSITS_API, '123' );
```

includes/admin/class-wc-payments-admin.php
```diff
@@ -1135,9 +1136,14 @@ public function add_transactions_notification_badge() {
 	 * @return int The number of disputes which need a response.
 	 */
 	private function get_disputes_awaiting_response_count() {
+		$request       = Request::get( WC_Payments_API_Client::DISPUTES_API . '/status_counts' );
+		$send_callback = function() use ( $request ) {
+			$request->send( 'wcpay_get_dispute_status_counts' );
+		};
+
```
htdat: ❌  I think you missed `return` for the callback? Otherwise, when the database cache is invalidated, the `$data` will always return to `null` - [codde ref](https://github.com/Automattic/woocommerce-payments/blob/f1ebc9063643776c772736cd5c83574bc8ad5052/includes/class-database-cache.php#L91-L92). You can test this by adding the fourth parameter `$force_refresh` to `true` around [this call](https://github.com/Automattic/woocommerce-payments/blob/f1ebc9063643776c772736cd5c83574bc8ad5052/includes/admin/class-wc-payments-admin.php#L1144-L1149), and will not see the indicator of current `Needs Response` disputes - [screenshot](https://github.com/Automattic/woocommerce-payments/assets/10045087/f3c14051-8009-472e-b0ae-9ec3dc530556).

Also, may not need to use `use` like your current code? 
```suggestion
		$send_callback = function() {
			$request = Request::get( WC_Payments_API_Client::DISPUTES_API . '/status_counts' );
			return $request->send( 'wcpay_get_dispute_status_counts' );
		};

```

includes/admin/class-wc-payments-admin.php
```diff
@@ -1159,9 +1165,14 @@ private function get_uncaptured_transactions_count() {
 		$test_mode = WC_Payments::mode()->is_test();
 		$cache_key = $test_mode ? DATABASE_CACHE::AUTHORIZATION_SUMMARY_KEY_TEST_MODE : DATABASE_CACHE::AUTHORIZATION_SUMMARY_KEY;
 
+		$request = Request::get( WC_Payments_API_Client::AUTHORIZATIONS_API . '/summary' );
+
+		$send_callback         = function() use ( $request ) {
+			$request->send( 'wc_pay_get_authorizations_summary' );
+		};
```
htdat: ❌  Similarly above: 

- Miss `return`
- May not need to use `use ( $request )`.

htdat (changes requested):
🎉 Great work on this PR. 

Besides inline comments, I have some other comments: 

- Should document or mention this new way to approach for similar simple GET requests in the general doc here `includes/core/server/README.md`? 
- This PR is consistently failing this test run: `E2E Tests - Pull Request / WC - latest | subscriptions` - may check it?

includes/core/server/class-request.php
zmaglica: Added in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/cd2703850d925d41ad7bf8e1de41c8e86f559bbf

includes/core/server/request/class-get-request.md
zmaglica: Addressed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/133739c69252d8698ca11e530c25b05dae0c2f6b

includes/core/server/class-request.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/133739c69252d8698ca11e530c25b05dae0c2f6b

includes/admin/class-wc-payments-admin.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/133739c69252d8698ca11e530c25b05dae0c2f6b

includes/admin/class-wc-payments-admin.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6234/commits/133739c69252d8698ca11e530c25b05dae0c2f6b

tests/unit/core/server/request/test-class-get-request.php
zmaglica: Fixed in https://github.com/Automattic/woocommerce-payments/commit/cd2703850d925d41ad7bf8e1de41c8e86f559bbf

changelog/dev-6217-generic-get-request-class
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Added generic GET request class
```
htdat: Nitpick: 

```suggestion
Add generic Get_Request class, and migrate current simple requests to use it
```

includes/admin/class-wc-rest-payments-vat-controller.php
```diff
@@ -66,7 +67,8 @@ public function register_routes() {
 	 */
 	public function validate_vat( $request ) {
 		$vat_number = $request->get_param( 'vat_number' );
-		return $this->forward_request( 'validate_vat', [ $vat_number ] );
+		$request    = Request::get( WC_Payments_API_Client::VAT_API, $vat_number );
+		return $request->send( 'wcpay_validate_vat_request' );
```
htdat: Nitpick: Not too important for this short function but would use a new variable name to avoid confusion for other devs reading this again in the future: 


```suggestion
		$server_request    = Request::get( WC_Payments_API_Client::VAT_API, $vat_number );
		return $server_request->send( 'wcpay_validate_vat_request' );
```

includes/core/server/class-request.php
```diff
@@ -84,6 +85,60 @@ abstract class Request {
...
+	 */
+
+	/**
+	 * Route list.
+	 *
+	 * @var string[]
+	 */
+	private $route_list = [
+		WC_Payments_API_Client::ACCOUNTS_API           => 'accounts',
+		WC_Payments_API_Client::CAPABILITIES_API       => 'accounts/capabilities',
```
htdat: Since `key` and `value` are the same, I don't think that brings much value if we use an associate array here.
```suggestion
		WC_Payments_API_Client::ACCOUNTS_API,
		WC_Payments_API_Client::CAPABILITIES_API,
```

includes/admin/class-wc-payments-admin.php
```diff
@@ -1135,9 +1136,14 @@ public function add_transactions_notification_badge() {
 	 * @return int The number of disputes which need a response.
 	 */
 	private function get_disputes_awaiting_response_count() {
+		$send_callback = function() {
+			$request = Request::get( WC_Payments_API_Client::DISPUTES_API . '/status_counts' );
+			return $request->send( 'wcpay_get_dispute_status_counts' );
```
htdat, line 1181: Sorry, I would have caught it up earlier. I've manaully tested this again, and it requires an array as the output instead of an object: 

```suggestion
			return $request->send( 'wcpay_get_dispute_status_counts' )->to_array();
```

includes/admin/class-wc-payments-admin.php
```diff
@@ -1159,9 +1165,13 @@ private function get_uncaptured_transactions_count() {
 		$test_mode = WC_Payments::mode()->is_test();
 		$cache_key = $test_mode ? DATABASE_CACHE::AUTHORIZATION_SUMMARY_KEY_TEST_MODE : DATABASE_CACHE::AUTHORIZATION_SUMMARY_KEY;
 
+		$send_callback         = function() {
+			$request = Request::get( WC_Payments_API_Client::AUTHORIZATIONS_API . '/summary' );
+			return $request->send( 'wc_pay_get_authorizations_summary' );
```
htdat, line 1210: Would be like this too? 

```suggestion
			return $request->send( 'wc_pay_get_authorizations_summary' )->to_array();
```

htdat (commented):
Thanks for addressing my feedback @zmaglica! I have some next comments. 

I will review more and focus on these things: 

- Run more manual tests for some endpoints. 
- Figure out why subscription E2E tests are failing.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2738,7 +2740,9 @@ public function update_order_status() {
 				$this->order_service->attach_transaction_fee_to_order( $order, $charge );
 			} else {
 				// For $0 orders, fetch the Setup Intent instead.
-				$intent    = $this->payments_api_client->get_setup_intent( $intent_id );
+				$setup_intent_request = Request::get( WC_Payments_API_Client::SETUP_INTENTS_API, $intent_id );
+
+				$intent    = $setup_intent_request->send( 'wcpay_get_setup_intent_request' );
```
htdat, line 2723: I think it must be like this to deal with subscription E2E test failures: 

```suggestion
				$intent    = $setup_intent_request->send( 'wcpay_get_setup_intent_request' )->to_array();
```

htdat (commented):
> Figure out why subscription E2E tests are failing.

Like this GH Action run or another one, this is the failing test. 

```
FAIL tests/e2e/specs/subscriptions/shopper/shopper-subscriptions-purchase-free-trial.spec.js (28.577 s)
  Subscriptions > Purchase subscription with free trial
    ✕ should be able to purchase a subscription with free trial (18112 ms)
    ✕ should create an order with "Setup Intent" (6720 ms)
``` 

I could follow up the E2E test spec, set up a free trial subscription, and use a 3DS card like `4000002760003184` then I can replicate the issue with this PHP error: 

<details>

<summary>PHP Error log</summary>

```
[13-Jun-2023 15:38:39 UTC] PHP Fatal error:  Uncaught Error: Call to undefined method WCPay\Core\Server\Response::get_charge() in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-order-service.php:1453
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-order-service.php(123): WC_Payments_Order_Service->get_intent_data(Object(WCPay\Core\Server\Response))
#1 /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php(2753): WC_Payments_Order_Service->update_order_status_from_intent(Object(Automattic\WooCommerce\Admin\Overrides\Order), Object(WCPay\Core\Server\Response))
#2 /var/www/html/wp-includes/class-wp-hook.php(308): WC_Payment_Gateway_WCPay->update_order_status('')
#3 /var/www/html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters('', Array)
#4 /var/www/html/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 /var/www/html/wp-admin/admin-ajax.php(188): do_action('wp_ajax_update_...')
#6 {main}
  thrown in /var/www/html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments-order-service.php on line 1453
```

</details>

See my inline comment to have the proposed solution. This makes me concerned, as some specific cases will need the output is an array rather than just an object providing `ArrayAccess` such as [`WCPay\Core\Server\Response`](https://github.com/Automattic/woocommerce-payments/blob/a45d86489dcdd7905be17a1030d5da3c828af536/includes/core/server/class-response.php#L16).

includes/core/server/class-request.php
zmaglica: This is like a micro optimization, but since the array is small, we could easily convert it to array.

htdat: @zmaglica - you mentioned this https://gist.github.com/alcaeus/536156663fac96744eba77b3e133e50a in DM so I agree that makes a lot sense.

Since we’re using this quite often, this optimization is really worth doing. If so, instead of repeating the same thing for key and value, maybe we just switch all values into a fixed value such as `true`.

```suggestion
		WC_Payments_API_Client::ACCOUNTS_API           => true,
		WC_Payments_API_Client::CAPABILITIES_API       => true,
```

htdat (changes requested):
As I mentioned in a previous review https://github.com/Automattic/woocommerce-payments/pull/6234#pullrequestreview-1477557865 regarding return types: 

> This makes me concerned, as some specific cases will need the output is an array rather than just an object providing `ArrayAccess` such as [`WCPay\Core\Server\Response`](https://github.com/Automattic/woocommerce-payments/blob/a45d86489dcdd7905be17a1030d5da3c828af536/includes/core/server/class-response.php#L16).

It appears that the current PHPUnit tests do not cover well for similar cases. 
For example, I found another case that basically, developers do not test the output of WCPay Server requests [WC_REST_Payments_Reader_Controller::fetch_readers](https://github.com/Automattic/woocommerce-payments/pull/6234/files#diff-caf095d72581da0c2fa09f946140ac52d19b1e58c3df617f0b7f1e841a6755ffR241): 
- The expected return type for this method is `array` but with your changes, it’s the ` WCPay\Core\Server\Response` object. The only test for this private method `fetch_readers` is done through testing another public method `get_all_readers`. 
- However, [the test itself ](https://github.com/Automattic/woocommerce-payments/blob/3246e80504859b99c09d7d800847f378bddacdeb/tests/unit/admin/test-class-wc-rest-payments-readers-controller.php#L156-L157) only tests the transient that will not trigger any WCPay Server Request.

**All in all, I am afraid of that these simple changes can cause issues for merchants. So, I think, we should:**

1. Double check all changes, and ensure they are returning the same return type, i.e. `array` rather than object `WCPay\Core\Server\Response`.

2. Ideally, if we can do (though I know that will take a lot of effort), review tests too, and ensure that all usages will trigger data returned from the server request.

includes/core/server/class-request.php
zmaglica: I would rather keep in case we decide to refactor it and move from the constant from client class.

zmaglica:
> As I mentioned in a previous review https://github.com/Automattic/woocommerce-payments/pull/6234#pullrequestreview-1477557865 regarding return types:

I have changed the request's response to return the server's default response. Now the array will be returned for all GET Requests.

htdat (approved):
🚢 it! Amazing meticulous work.

I was confused at first regarding the difference between `handle_rest_request` and `send` of class `WCPay\Core\Server\Request`. 

But DMing with @zmaglica, that's clear now. `handle_rest_request` is better for replacing endpoints that originally used `WC_Payments_REST_Controller::forward_request`. Both of them nicely ensure the expectation for output of REST Controllers, either an array or a WP_Error. 

That explains why some places use `handle_rest_request` while others use `send` in this PR.

PR #6244, opened 2023-05-08
Migrate chip component to TS

reykjalin:

- Migrate the `client/components/chip` component and unit tests to TS.

#### Testing instructions

* Make sure unit tests pass.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6244` or branch name `add/migrate-chip-component` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/migrate-chip-component) :rocket:

---

Build info: 

- Latest commit: 47b0f52a3f26a1f99f87984aea4d800219282a62
- Build time: 2023-05-16 04:11:11 UTC

client/components/chip/test/index.tsx
```diff
@@ -34,11 +47,6 @@ describe( 'Chip', () => {
 		expect( chip ).toMatchSnapshot();
 	} );
 
-	test( 'renders a primary chip by default', () => {
-		const { container: chip } = renderChip( undefined, 'Message' );
-		expect( chip ).toMatchSnapshot();
-	} );
-
```
reykjalin: Removing this test felt a little weird because there's still potential for JS to not include the `type` parameter, but I checked the codebase and there's no place currently using the `Chip` component without providing `type` (or providing the wrong value for `type`), so I think this is fine.

client/components/chip/test/index.tsx
```diff
@@ -34,24 +47,8 @@ describe( 'Chip', () => {
...
 	test( 'renders a warning chip', () => {
 		const { container: chip } = renderChip( 'warning', 'Alert message' );
 		expect( chip ).toMatchSnapshot();
 	} );
-
-	test( 'renders default if type is invalid', () => {
-		const { container: chip } = renderChip( 'invalidtype', 'Message' );
-		expect( chip ).toMatchSnapshot();
-	} );
```
reykjalin: The same applies to this test that I mentioned in https://github.com/Automattic/woocommerce-payments/pull/6244#discussion_r1187915074

client/components/chip/test/index.tsx
```diff
@@ -2,14 +2,27 @@
...
  */
+import React from 'react';
 import { render } from '@testing-library/react';
 
 /**
  * Internal dependencies
  */
 import Chip from '../';
+// eslint-disable-next-line no-duplicate-imports
+import type { ChipType } from '../';
```
ismaeldcom: If we're going to use `import type` will be better to update our ESLint config and replace the old `no-duplicate-imports` by [`import/no-duplicates`](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-duplicates.md) which supports it. And even the new inline type import from TS 4.5.

client/components/chip/test/index.tsx
ismaeldcom: Did you remove it to avoid TS errors? Because we can safely cast it here just for testing purposes. Something like `'invalid' as ChipType`. As long as we continue using JS, I think it's safer to add/keep those extra tests and conditions.

client/components/chip/index.tsx
```diff
@@ -1,14 +1,25 @@
...
  * Internal dependencies
  */
 import './style.scss';
 import { HoverTooltip } from 'wcpay/components/tooltip';
 
-const types = [ 'primary', 'light', 'warning', 'alert' ];
+const types = [ 'primary', 'light', 'warning', 'alert' ] as const;
 
-const Chip = ( props ) => {
+export type ChipType = typeof types[ number ];
```
ismaeldcom, line 12: I think we can simplify this a bit, making `ChipType` directly a template literal, and `type` optional. Then we don't need to use `find` later, just the provided value. Something like:

```tsx
export type ChipType = 'primary' | 'light' | 'warning' | 'alert';

interface Props {
    // ...
	type?: ChipType;
    // ...
}

// ...
	const classNames = [
		'chip',
		`chip-${ type || 'primary' }`,
// ...
```

The only downside I see, is not using `primary` style when an invalid type is provided from JS code.

client/components/chip/index.tsx
```diff
@@ -1,14 +1,25 @@
...
+interface Props {
+	message: string;
+	type: ChipType;
+	isCompact?: boolean;
+	className?: string;
+	tooltip?: React.ReactNode;
+}
+
+const Chip: React.FC< Props > = ( props ) => {
 	const { message, type, isCompact, className, tooltip } = props;
```
Jinksi, line 27: Nit/question: should we prefer to destructure props within the function param list or on a separate line (as is currently implemented)?

```suggestion
const Chip: React.FC< Props > = ( {
	message,
	type,
	isCompact,
	className,
	tooltip
} ) => {
```

ismaeldcom: I usually prefer to do it within the function param list, but don't like too much when it ends up taking so many lines 😄

reykjalin: I think it's situational 🤔 I generally favor destructuring the props, but in some cases I think destructuring in the function (or even using `prop.<x>` is better.

At the same time I'm sort of indefferent? 😅  Like here I'd be fine with either. I personally like that the destructuring is on just one line before the change though 👍

client/components/chip/test/index.tsx
reykjalin: Partly yes, but partly because TS is supposed to get rid of the possibility of even using the component wrong.

Your point about JS still stands though. I'll add the test back in and make sure it's passing 👍

client/components/chip/index.tsx
reykjalin: Like you said: the problem with changing

```ts
`chip-${ types.find( ( t ) => t === type ) || 'primary' }`,
```

to 

```ts
`chip-${ type ?? 'primary' }`,
```

Is that if someone now passes in an invalid string as the type (e.g. `invalid` which is one of the unit tests) you'll get a class that's `chip-<invalid_type>` instead of the expected `chip-primary`.

To address that we can either keep the `const types = [ /* ... */ ];` array and the `types.find( /* ... */ )` functionality, or add a type guard like

```ts
const isChipType = ( type?: string ): type is ChipType => {
	return types.includes( type as ChipType );
};

// ...

const classNames = [
  // ...
  `chip-${ isChipType( type ) ? type : 'primary' }`,
  // ...
];
```

The type guard may be more TypeScript-y, but I'm fine with keeping the `.find` approach as well 🤔 

What do you think? I've pushed be9a333ac9240d245a0559db8049c5d5dba83c35 as an example of the "more TS-y" approach.

client/components/chip/test/index.tsx
reykjalin: I've re-added the tests in b26d5b1419c8e195b70b8bae732eb3989135fc20 🙌

client/components/chip/test/index.tsx
reykjalin: Oooh, yes nice, fixed this in b66141b0365e36e8524d9eb4777a986b83c0cfe7 🎉

reykjalin: I think I fixed the failing linting task in 47b0f52a3f26a1f99f87984aea4d800219282a62 but I had to apply some of the errors reported by the new `import/no-duplicates` rule. I had ESLint auto-fix the issues, and you can see those in the commit.

client/components/chip/index.tsx
ismaeldcom: I like type guards, but you're right it could be too "TS-y" 😄 Maybe that's too much if we want to keep the TS migration barrier low enough.

reykjalin:
Closing since this was actually wrapped up in https://github.com/Automattic/woocommerce-payments/pull/6796 (or some other PR before that)

PR #6281, opened 2023-05-12
Track downloads on deposit details page.

### LINKED ISSUE #5885: Ability to measure view and user interaction with deposits detail page
We need to be able to:

- Count how many times user view the deposits details page.

- Count user's interaction with the search bar.

- Count user's click on the Download button.

- Count user's click on the column header to sort the table by.

- Count user's interaction with the column columns show / hide UI from the three dots button at the top right of the table.

Screenshot 2023-03-28 at 20 29 28

ps: I see slight difference when deposit is estimated and not.

brucealdridge:
Fixes #5885 

image

This PR updates tracking on the deposit details page to add clarity to event tracking of downloads.

Previously downloads from this page have the ambiguous event `wcadmin_wcpay_transactions_download` this is due to the transactions and deposit details page both using the same component with the only difference being a `depositId`. ~This PR updates the event when downloading a deposits transactions to use `wcpay_deposits_detail_download`~

~This change will break any tracking flows for anything that is currently setup as it is currently assumed that deposit details downloads will have the `wcadmin_wcpay_transactions_download` event firing. However, there do not appear to be any current flows/funnels setup for this event. A note should be added on this event to make sure that viewers are aware.~

This PR now adds a new event `wcpay_transactions_download_csv_click` which fires when the button is clicked. The previous event `wcpay_deposits_detail_download` only fires when the download is successful. The new event contains the following attributes

 * `location`: `transactions` or `deposit_details` to signify where the button was clicked
 * `download_type`: `endpoint` or `browser` 
 * `exported_transactions`: Amount of transactions included in the downloaded report.
 * `total_transactions`: Total amount of transactions available with the used filters.

#### Testing instructions

* Visit the WC Pay transactions page and confirm that the `wcpay_transactions_download_csv_click` still fires when the "Download CSV" button is clicked.
* Visit the deposits page, click through to a deposit, confirm that the `wcpay_transactions_download_csv_click` still fires when the "Download CSV" button is clicked.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6281` or branch name `add/5885-deposit-detail-tracking` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/5885-deposit-detail-tracking) :rocket:

---

Build info: 

- Latest commit: 6da2aa8ad0b818c662da67ee22eb8158435bf6c4
- Build time: 2023-05-18 22:27:21 UTC

client/transactions/list/index.tsx
```diff
@@ -580,11 +580,18 @@ export const TransactionsList = (
...
-						exported_transactions: exportedTransactions,
-						total_transactions: exportedTransactions,
-						download_type: downloadType,
-					} );
+					wcpayTracks.recordEvent(
+						props.depositId
+							? wcpayTracks.events
+									.DEPOSITS_DETAIL_DOWNLOAD_CSV_CLICK
+							: wcpayTracks.events
+									.TRANSACTIONS_DOWNLOAD_CSV_CLICK,
```
haszari: I'm not sure we need a different event here – the action is the same, downloading a list of transactions as a CSV file. Could add a property to distinguish downloads from the transactions list vs. for a specific deposit. If we want to know how many merchants are downloading CSVs of transactions overall, we'd need to consider both events. 

Though I'm fine with having a separate event (as you've done here). If we do go with a separate event, I suggest naming it more clearly, i.e. indicating that it's downloading transactions. e.g. DEPOSITS_DETAIL_DOWNLOAD_TRANSACTIONS_CSV_CLICK

Jinksi: > I'm not sure we need a different event here ... add a property to distinguish downloads from the transactions list

+1

It's a bit of an unknown, but there could be a situation in the future where this transaction list can be downloaded from other locations (e.g. overview page). If this occurs, it would be simpler to add an event prop for each new location, rather than a new event for each.

Are there any downsides to using an event prop rather than a separate event here?

brucealdridge: > Are there any downsides to using an event prop rather than a separate event here?

My main reasoning was that it's unlikely that someone wants to know when the transaction list CSV is downloaded. 
A product owner looking at these stats will be looking at them to make product decisions. Are these flows working? Are these features being used?
Downloading all transactions and downloading transactions that make up a deposit are two separate flows/features. 

>  I suggest naming it more clearly, i.e. indicating that it's downloading transactions. e.g. DEPOSITS_DETAIL_DOWNLOAD_TRANSACTIONS_CSV_CLICK

After considering this I opted to keep with the naming convention in line with other CSV download events.

haszari: > After considering this I opted to keep with the naming convention in line with other CSV download events.

Assuming you mean the name is `DEPOSITS_DETAIL_DOWNLOAD_CSV_CLICK `? I'd argue `DEPOSITS_DETAIL_TRANSACTIONS_DOWNLOAD_CSV_CLICK` or similar is _more_ consistent, since both buttons download a list of transactions; "deposits detail" is not getting downloaded, that's the event source. Not a big deal though.

Jinksi: This might sound like a minor issue, but I noticed that the placement of the call to `recordEvent` means that it may not be recorded if there is an error in downloading.

This can be confirmed by watching the network requests while clicking: the `t.gif` request doesn't doesn't fire on the button click, only once the fetch request has completed (under 10,000 transactions). If you block the `/download` request URL, the event will not be recorded (see screen recording).

Perhaps `DOWNLOAD_CSV_CLICK` is misleading? Although it seems minor, it may be a fundamental issue with the event name. E.g. in the future, we may want to track how many successful `click` events result in successful `download` events.

https://github.com/Automattic/woocommerce-payments/assets/3147296/c4a36af6-b40b-4fb5-9d90-80265738fdd6

haszari: Good spotting! The event should be fired as close to the user action as possible, i.e. `download_click`, in `onClick`. 

If we want to track actual successful downloads that's a different event (IMO we should just track the intent to download).

Jinksi (changes requested):
@brucealdridge I've marked this review a `request changes` in order to get your feedback on the issue with the event variable name including `CSV_CLICK` which may be misleading, since it doesn't occur on a click event: https://github.com/Automattic/woocommerce-payments/pull/6281#discussion_r1195798826.

client/transactions/list/index.tsx
brucealdridge: > If we want to track actual successful downloads that's a different event

As this is an event that is already in place, maybe that was the intention and the renaming of the event/constant is the error.

haszari: Exactly 😁 . TBH I think the existing event is problematic, we should implement a good clear `_click` event for both kinds of transactions download, and look to deprecate/remove the existing event. Sounds like you're open to this as you're investigating how much it's used!?

client/transactions/list/index.tsx
```diff
@@ -550,6 +550,18 @@ export const TransactionsList = (
...
+						wcpayTracks.events.TRANSACTIONS_DOWNLOAD_CSV_CLICK,
+						{
+							location: props.depositId
+								? 'deposit_details'
+								: 'transactions',
+							download_type: downloadType,
+							exported_transactions: rows.length,
+							total_transactions: transactionsSummary.count,
+						}
+					);
```
Jinksi: Since this call to `recordEvent` will occur for both `endpoint`/`browser` download types, it could be placed _before_ the if/else statement, to be more DRY. And it would occur regardless of whether the `window.confirm()` call was confirmed by the user (closer to the `_click` action).

I'm happy to go with your judgement here.

Jinksi (approved):
LGTM @brucealdridge 🙌 

I've tested clicking the **download** button from both the deposit details and transactions screens. Each request to `t.gif` contains correct event props (including a distinct `location` for each) and uses the `wcadmin_wcpay_transactions_download_csv_click` event name.

I've left a comment regarding the placement of the call to `recordEvent`, feel free to take or leave it.

client/transactions/list/index.tsx
brucealdridge: Thanks, Good spotting on this. I've moved this tracking and marked this as resolved.

haszari: Nice, now the event is fired very close to user action, almost first thing in the `onClick` handler 👌

PR #6305, opened 2023-05-15
Default WCPay gateway should always be present on WC Settings screen for Payments

### LINKED ISSUE #6279: When UPE is enabled, WCPay cannot be enabled/disabled via the "All payment methods" page
### Describe the bug
When UPE is enabled, WCPay cannot be enabled or disabled via **WooCommerce > Settings > Payments > All payment methods**. The option disappears until UPE is disabled.

### To Reproduce
1. Navigate to **WooCommerce > Settings > Payments > All payment methods** while UPE is disabled.
2. Note that WCPay is an option.
3. Navigate to **Payments > Settings**.
4. Enable UPE.
5. Repeat step 1.
6. Note that WCPay is not an option.

### Actual behavior
This isn't that big of an issue as WCPay can be enabled/disabled directly from the **Payments > Settings** page. However, this doesn't match the expectations for other payment gateways and can cause confusion.

Additionally, we found an instance in 6273194-zen where this issue _may_ prevent the credit card fields from displaying in the following scenario:

1. UPE is enabled on a production site.
2. A staging site is created.
3. The account is transferred to the staging site.
4. A new account is onboarded on the production site.
5. EPMs display properly.
6. CC fields do not display.

For this merchant, we're asking them to disable UPE to see if this forces the card fields to display properly.

### Expected behavior
When UPE is enabled, WCPay should be an option at **WooCommerce > Settings > Payments > All payment methods**.

FangedParakeet:
Fixes #6279 

# Default WCPay gateway should always be present on WC Settings screen for Payments

The WCPay payment gateway supports multiple different methods and when the split-UPE is enabled these methods appear as separate payment gateways. [We manually remove these additional payment methods/gateways from the general WC Settings screen for Payments](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L701), to ensure that only the main WCPay gateway is present. 

When the split-UPE gateway is enabled, currently [we remove all but the legacy card gateway](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L701) from this list on the settings screen. However, this is incorrect, because when WooPay is disabled, the UPE card gateway is the main WCPay gateway instead. Consequently, when WooPay is disabled, this section of code ruthlessly removes all WCPay gateways, leading to the situation described in #6279.

This PR amends this section of code to maintain the correct gateway in all scenarios.  

#### Testing instructions

To get to the main WC Settings screen, visit **Payments > Settings** and select "All payment methods". WooCommerce Payments should always be present as a payment method on this screen. This should be true in all of the following situations:
- All UPE disabled
- Legacy UPE enabled
- Split-UPE enabled
   -  WooPay enabled
   - WooPay disabled

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6305` or branch name `fix/6279-show-wcpay-on-main-settings-screen` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6279-show-wcpay-on-main-settings-screen) :rocket:

---

Build info: 

- Latest commit: b26db83748e2e7255bb6db03b497dcd6014bd501
- Build time: 2023-05-22 08:16:10 UTC

timur27 (commented):
Thanks @FangedParakeet for working on this! I've found out that with the WooPay enabled, **_WooCommerce Payments_** isn't present on the page for the split UPE. Do you think we need to introduce a more sophisticated check in order to support all the scenarios? 

I've checked the deferred intent as well and it behaves the same way the split UPE behaves, so we might want to address them both together. 

Please let me know if I misunderstood anything and if the gateway entry shouldn't be visible with the WooPay enabled.

FangedParakeet:
>I've found out that with the WooPay enabled, WooCommerce Payments isn't present on the page for the split UPE. Do you think we need to introduce a more sophisticated check in order to support all the scenarios?

Oops, fired a little too quickly on my first shot at this and didn't follow the logic of which card gateway is actually registered. I added a new variable in 6476a39299a4ab641f8ca5d62ffeaff6cd1cea10 to keep track of this very gateway. Please let me know if you think this solution is a little more effective and elegant or if it instead incenses your intuitions, @timur27!

timur27 (approved):
In terms of the functionality, the latest changes work perfectly as expected and as a user I'm fully satisfied since the WCPay entry is available in the Settings under any circumstances (WooPay + UPE settings). 

From the technical standpoint, I think we did our best by introducing the `$registered_card_gateway`. However, in order to understand the change, I went to the `WC_Payments::init()` method and I found myself trying to understand the logic there which took me some time. Looking at `$registered_card_gateway` in isolation, it looks like a rightful member of the class with a clear goal. At the same time, when zooming out a bit, it becomes clear immediately, that the logic in this class is not straightforward, since `$registered_card_gateway` now became yet another concept on top of already existing `$legacy_card_gateway` and `$card_gateway`. It took me some time to remind, which gateway does what and I'm afraid it will take even more time for people who are working with this code for their first time. 

I think it's not a fault of this issue that the code became more complex. What I'm trying to emphasize is that this issue is a perfect example of how complex the initialization logic became and how complex it is now to perform even basic modifications. I was trying to find an easier solution that would allow us to avoid introducing a new variable, but it's nearly impossible and the only way is to merge `$legacy_card_gateway` and `$card_gateway` as part of the broader refactor. 

For me it's a sign that it's a perfect moment to decide, when the refactor should occur, if it should wait until the payment process re-engineering project is completed which (in my opinion) could make sense since that refactor would simplify the gateways construction. 

I'm approving this PR because it did what it's supposed to do and thanks for addressing the issue the best way possible! 🚀  I'm looking forward to discussing the topic of refactor during the upcoming meeting.

FangedParakeet:
I agree completely, @timur27: the logic in both `WC_Payments::init()` and `WC_Payments::register_gateway()` is getting very complicated and it's definitely gaudy having to maintain three static variables for the card gateway--`$card_gateway`, `$legacy_card_gateway`, and now `$registered_card_gateway`. 

I added the third new variable simply because the branching feature flag logic in `register_gateway()` is also required in `hide_gateways_on_settings_page()` to summon the correct gateway for settings. I couldn't really see a simple solution to extract and reuse this logic, so to avoid duplicating it in both functions, I added a new variable to capture it instead.

It's not ideal, but a reluctant resolution nonetheless for quite a rueful issue. This entire situation is essentially necessitated, because the card gateway can currently be one of three classes--`WC_Payment_Gateway_WCPay`, `UPE_Payment_Gateway`, or `UPE_Split_Payment_Gateway`. Once our deferred intent gateway is complete and fully operational, we will hopefully be in a position to unify this logic with a single gateway and hopefully all of this extraneous code can be consolidated and eliminated.

PR #6313, opened 2023-05-16
Ensures manual capture hook fired only once when multiple gateways present

### LINKED ISSUE #5780: Cannot capture charge with UPE enabled
### Describe the bug
Technically, you can capture the charge, however, the `capture_charge` method is called around 11 times when you attempt to, which causes an error to be returned from the server after the first one is successful. The error causes the page not to load and you need to go back to see the charge was captured successfully.

The action below gets added for every UPE gateway there is, causing it to be actioned when the order action is done:

```
add_action( 'woocommerce_order_action_capture_charge', [ $this, 'capture_charge' ] );
```

### To Reproduce

1. Use the latest `develop` for the client and `trunk` for the server.
2. Go to Payments > Settings and enable _Issue an authorization on checkout, and capture later_
3. Add an item to your cart, and go through the checkout process.
4. Go to the order in the admin, and use the _Capture charge_ option under _Order actions_
5. The page should time out.
6. If you go back you will see the charge was captured.
7. If you find the intent in Stripe, you can see multiple attempts to capture the charge were attempted.

FangedParakeet:
Fixes #5780 

# Refactoring manual capture authorisation hooks 

> Technically, you can capture the charge, however, the capture_charge method is called around 11 times when you attempt to, which causes an error to be returned from the server after the first one is successful. The error causes the page not to load and you need to go back to see the charge was captured successfully.
>
> The action below gets added for every UPE gateway there is, causing it to be actioned when the order action is done:
>
> `add_action( 'woocommerce_order_action_capture_charge', [ $this, 'capture_charge' ] ); `

#### Testing instructions

1. Use this branch for the client and `trunk` for the server.
2. Go to Payments > Settings and enable _Issue an authorization on checkout, and capture later_
3. Add an item to your cart, and go through the checkout process.
4. Go to the order in the admin, and use the _Capture charge_ option under _Order actions_
5. The page should not time out.
6. The charge should be captured successfully.
7. Locate the corresponding intent for the transaction in Stripe's dashboard. There should only be a single charge associated with the intent.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6313` or branch name `fix/5780-multiple-manual-captures-with-upe` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/5780-multiple-manual-captures-with-upe) :rocket:

---

Build info: 

- Latest commit: 78aae3b1c20cc477139b584b2976d2b014b99c90
- Build time: 2023-05-24 17:29:16 UTC

mdmoore (approved):
Capturing the charge doesn't time out and only a single charge is associated with the intent in Stripe. LGTM! 🚢

includes/class-wc-payments-checkout.php
```diff
@@ -74,6 +74,8 @@ public function __construct(
 		$this->customer_service = $customer_service;
 
 		add_action( 'wc_payments_add_payment_fields', [ $this, 'payment_fields' ] );
+		add_action( 'woocommerce_order_action_capture_charge', [ $this->gateway, 'capture_charge' ] );
+		add_action( 'woocommerce_order_action_cancel_authorization', [ $this->gateway, 'cancel_authorization' ] );
```
marcinbot, line 78: While this works, it's not a proper fix.

The Checkout class appears to handle the checkout page-related stuff. Capturing the charge via the admin panel has nothing to do with the checkout.

I think a better fix could be to add empty `capture_charge`/`cancel_authorization` methods to the Split UPE gateway classes (with a comment explaining why, of course)

Let me know what you think @FangedParakeet

FangedParakeet: Totally agree that this isn't really a hook associated with checkout and doesn't really merit a mention in this class. In previous attempts to refactor these hooks, the `WC_Payments_UPE_Checkout` and `WC_Payments_Checkout` classes have been used as catch-all holders for payment gateway hooks, since--unlike the gateway constructors--we could be confidently assured that thsee checkout classes would only be initialised once.

However, your refactor in #6326 seems like a much more coherent and concrete solution to this problem. I would probably support the change here in #6313 as a temporary solution--in my mind preferable to retaining this in the class constructor, attaching redundant hooks, and having those empty methods fire unnecessarily from the Split UPE gateways; but if you were able to move this to the [`WC_Payment_Gateway_WCPay::init_hooks()` function](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payment-gateway-wcpay.php#L419) in #6326, I would readily support that as a much more appropriate long-term solution and would gladly invest all hopes and faith in that blessed PR instead. 🙏

marcinbot: > In previous attempts to refactor these hooks, the WC_Payments_UPE_Checkout and WC_Payments_Checkout classes have been used as catch-all holders for payment gateway hooks, since--unlike the gateway constructors--we could be confidently assured that thsee checkout classes would only be initialised once.

While I see how this would work, it's just not good engineering. Just because we can tie the hooks to the Checkout class and solve the immediate issue we're facing, it's not the purpose of the Checkout class. Such sloppiness only enables more bad practices and leads to us delivering a bad product. Although TBH the code was messy before - the checkout class shouldn't have to be aware of the gateway - it should be the other way around. It also shouldn't be called via the hooks either. Something to clean up in the future.

> I would probably support the change here in https://github.com/Automattic/woocommerce-payments/pull/6313 as a temporary solution

Fair enough, I updated #6313 to account for this, and let's get it shipped soon :)

Issue #5698, opened 2023-03-09
Move hooks from service constructors to a helper for them


htdat:
### Description

From this comment https://github.com/Automattic/woocommerce-payments/pull/5693#discussion_r1130553433

Move hooks outside of service constructors, and add a helper for those, ex. `WC_Payments::get_service( $class_name )`, which would instantiate a singleton service service, **and** call a specific method, ex. `register_hooks`.

### Acceptance criteria

* Move hooks outside of service constructors, and add a helper for those.
* Check and convert service classes into singleton.

### Dev notes

For singleton, as far as I see, there are two patterns. 

- One in our current code like this. We can abstract this into a base service class
https://github.com/Automattic/woocommerce-payments/blob/f694c5282c9c052c6e578427ac26de39e8784e28/includes/fraud-prevention/class-fraud-prevention-service.php#L59-L65

- Second approach is to use trait like WC Core is doing: declare a [SingletonTrait](https://github.com/woocommerce/woocommerce/blob/bec3ec18ac801875a830ddf097ba173069161c13/plugins/woocommerce/includes/rest-api/Utilities/SingletonTrait.php#L13-L13), then [use it for new clasees](https://github.com/woocommerce/woocommerce/blob/bec3ec18ac801875a830ddf097ba173069161c13/plugins/woocommerce/includes/rest-api/Server.php#L18-L19).
### Additional context

This may be good for the next phrase of platform project.

htdat:
https://github.com/Automattic/woocommerce-payments/issues/2769

marcinbot:
Closing in favor of:
https://github.com/Automattic/woocommerce-payments/pull/7249
https://github.com/Automattic/woocommerce-payments/issues/7258
https://github.com/Automattic/woocommerce-payments/issues/7259
https://github.com/Automattic/woocommerce-payments/issues/7266
https://github.com/Automattic/woocommerce-payments/issues/7260
https://github.com/Automattic/woocommerce-payments/issues/7261
https://github.com/Automattic/woocommerce-payments/issues/7262
https://github.com/Automattic/woocommerce-payments/issues/7263
https://github.com/Automattic/woocommerce-payments/issues/7264
https://github.com/Automattic/woocommerce-payments/issues/7265

PR #6326, opened 2023-05-18
Move hooks out of the payment processing class constructors

marcinbot:
Partial fix for https://github.com/Automattic/woocommerce-payments/issues/2769 and https://github.com/Automattic/woocommerce-payments/issues/5698

This takes the payment processing classes - Gateways (classic, UPE, and split), Checkouts, and Subs traits - and moves the hooks out of their constructors to a new `init_hooks` function.

Functionality-wise, not much should change since the new function is called right after the constructor. But a few attempts have been made to limit registering of certain hooks to happen only once.

#### Testing instructions

- For classic, UPE, and Split UPE payments
  - Light smoke testing covering cases like card payments, auth+capture, payments using different methods, saving payment methods for later
  - Light smoke testing of the settings pages
- For Subscriptions
  - Also a light smoke test for cases like subscription creation, renewal, and edge cases such as free trials etc. 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6326` or branch name `add/init-hooks` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/init-hooks) :rocket:

---

Build info: 

- Latest commit: a9c84fbbe9ae40a7049f5c0c3eb1a5b2a2f88489
- Build time: 2023-06-02 08:22:41 UTC

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -409,33 +409,45 @@ public function __construct(
...
 		}
+	}
 
+	/**
+	 * Initializes this class's WP hooks.
+	 *
+	 * @return void
+	 */
+	public function init_hooks() {
 		add_action( 'woocommerce_update_options_payment_gateways_' . $this->id, [ $this, 'process_admin_options' ] );
```
FangedParakeet: Should this one not also be included within the `if` block below for hooks that only need to be added once for the main gateway? I'm not very familiar with what's going on here, but it looks like an admin function that should probably only be fired once.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -409,33 +409,45 @@ public function __construct(
...
-		add_action( 'admin_notices', [ $this, 'display_errors' ], 9999 );
-		add_action( 'woocommerce_order_actions', [ $this, 'add_order_actions' ] );
-		add_action( 'woocommerce_order_action_capture_charge', [ $this, 'capture_charge' ] );
-		add_action( 'woocommerce_order_action_cancel_authorization', [ $this, 'cancel_authorization' ] );
+		add_action( 'woocommerce_update_order', [ $this, 'schedule_order_tracking' ], 10, 2 );
 
-		add_action( 'wp_ajax_update_order_status', [ $this, 'update_order_status' ] );
-		add_action( 'wp_ajax_nopriv_update_order_status', [ $this, 'update_order_status' ] );
+		// Only add certain actions/filter if this is the main gateway (i.e. not split UPE).
+		if ( self::GATEWAY_ID === $this->id ) {
```
FangedParakeet, line 421: Just FYI, I think when the split UPE is enabled, these hooks will still be added twice: once by the `UPE_Split_Payment_Gateway` card gateway [here](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L476) and then again by the same gateway below [here](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L498). I'm not sure if this is actually such a disaster, since in both incantations the CC gateway is being used--and not the respective gateways for other LPMs--but just worth mentioning, in case this is specifically unintended behaviour.

includes/class-wc-payments.php
FangedParakeet, line 1: When the split-UPE is enabled, [the `WC_Payments_Checkout` hooks inside `$card_payments_checkout`](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L482) never fire. We need both the `WC_Payments_Checkout` and `WC_Payments_UPE_Checkout` to be active in the split-UPE with WooPay enabled, because in this scenario the card gateway will serve the legacy card element (generated by `WC_Payments_Checkout`), whereas the other payment gateways will use UPE payment elements (generated by `WC_Payments_UPE_Checkout`).

Currently, when both the split-UPE and WooPay are enabled, the CC fields that should be represented by `WC_Payments_Checkout` do not load at all, because the `payment_fields` hook within this class is never fired.

CC fields fail to load with split-UPE and WooPay
_CC fields empty with WooPay enabled with the split-UPE_

I think if you just add `$card_payments_checkout->init_hooks();` below [this line](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L482) that should resolve this without too much drama and AFAICT no other ill effects.

FangedParakeet (changes requested):
Nice one! I like the approach here of explicitly casting out these hooks, as opposed to relying on the murky depths of the gateway class constructors to attach these hooks covertly instead. 

I left a few comments about minor concerns and I found one rather large breaking issue when the split-UPE is enabled (which is now the default experience for all new WCPay users). I've suggested a fix that I believe is pretty simple though, so nothing to fear yet. Let me know your thoughts on my suggestions and I'll have another look later on! ☮️

includes/class-wc-payment-gateway-wcpay.php
marcinbot: I figured it doesn't need to be because the hook ID contains the gateway ID so should fire only for the main gateway. Upon closer inspection, though, it looks like this hook is not used since we now have a JS UI for managing settings. 

Removed in [2ac6f1b](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/2ac6f1b9c80ecd6b0c15c1b776557c7ac18ef078)

includes/class-wc-payment-gateway-wcpay.php
marcinbot: I think that's one thing we're trying to avoid here. Added an ugly fix here: [c17850d](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/c17850dacb9c9cf1ae940fe2529be1561ff80524). But in the future we might want to consider a better code structure for Split UPE

includes/class-wc-payments.php
marcinbot: Good catch. I reproduced and tried to find a more elegant solution but there's a lot to untangle here that's not related to this PR or even Split UPE. So I went with your proposed solution in [766fcf1](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/766fcf1b3208162adc6e1ba58285d37f4a1ea4ac) after confirming it's safe.

The nice thing about this PR is that now it's clear why `WC_Payments_Checkout` is initialized - before it was just a dangling object unused anywhere else.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -409,31 +409,44 @@ public function __construct(
...
-		add_action( 'wp_ajax_nopriv_update_order_status', [ $this, 'update_order_status' ] );
+			add_action( 'wp_ajax_update_order_status', [ $this, 'update_order_status' ] );
+			add_action( 'wp_ajax_nopriv_update_order_status', [ $this, 'update_order_status' ] );
 
-		add_action( 'wp_enqueue_scripts', [ $this, 'register_scripts' ] );
-		add_action( 'wp_enqueue_scripts', [ $this, 'register_scripts_for_zero_order_total' ], 11 );
-		add_action( 'wp_ajax_create_setup_intent', [ $this, 'create_setup_intent_ajax' ] );
-		add_action( 'wp_ajax_nopriv_create_setup_intent', [ $this, 'create_setup_intent_ajax' ] );
+			add_action( 'wp_enqueue_scripts', [ $this, 'register_scripts' ] );
+			add_action( 'wp_enqueue_scripts', [ $this, 'register_scripts_for_zero_order_total' ], 11 );
```
FangedParakeet: I wonder if these should be moved to the `WC_Payments_Checkout` class. 

When WooPay is enabled with the split-UPE, we need to register a card gateway using the legacy card element and other gateways for additional LPMs using UPE elements: this means that we need to register/enqueue scripts for the `UPE_Split_Payment_Gateway` classes _and_ the `WC_Payment_Gateway_WCPay` class. You can test this by enabling Split UPE checkout and WooPay eligibility from the Dev Tools: now enable WooPay from WCPay settings and proceed into a checkout for an item. If you select the CC gateway and choose "Use a new payment method", you should find that the CC field fails to load, because the scripts have not been enqueued onto the page.

I initially thought that we could resolve this by adding `self::$legacy_card_gateway->init_hooks();` below [this line](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L496) in `WC_Payments::init()`, but this would have an unintended side-effect of calling all of these hooks twice for the [card gateway instance of `UPE_Split_Payment_Gateway` used for settings](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L493) and the [legacy card gateway instance of `WC_Payment_Gateway_WCPay` used for the CC checkout gateway](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L460). 

In another conversation you'd elaborated that the checkout classes (`WC_Payments_Checkout` and `WC_Payments_UPE_Checkout`) should control all checkout page-related stuff and I think these hooks would come underneath that umbrella...but let me know if you disagree or have any better ideas on how to free ourselves from this fiendish foible.

FangedParakeet (changes requested):
Almost everything looks good! Tested with a few different flavours of the UPE, while testing a few subscription purchase and renewal flows against the split-UPE and didn't encounter anything unexpected. ✅ 

...But I found that when WooPay and the split UPE are both enabled, the CC fields fail to display on the page, because I believe the scripts for the legacy card gateway are not enqueued onto the page. I left more of my thoughts and some better instructions to replicate in another comment, so let me know if you're able to see the same and if you have any ideas on how to solve this situation. 🙏

includes/class-wc-payment-gateway-wcpay.php
marcinbot: Fixed in [805640a](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/805640afedbc10cabd889e8229ce1c575a0f8a6a)

I tested with Standard, UPE, and Split UPE, and also tested the APay/GPay buttons still work :)

includes/class-wc-payments-checkout.php
```diff
@@ -86,6 +94,60 @@ public function enqueue_payment_scripts() {
...
+		) {
+			WC_Payments::get_gateway()->tokenization_script();
+			$script_handle = 'WCPAY_CHECKOUT';
+			$js_object     = 'wcpayConfig';
+			if ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+				$script_handle = 'wcpay-upe-checkout';
+				$js_object     = 'wcpay_upe_config';
+			} elseif ( WC_Payments_Features::is_upe_legacy_enabled() ) {
+				$script_handle = 'wcpay-upe-checkout';
+			}
```
FangedParakeet, line 145: I'm not very familiar with the purpose of this function, but as things stand right now, this section of code will never be called.

When `WC_Payments_Features::is_upe_legacy_enabled()` is `true`, we do not instantiate this `WC_Payments_Checkout` class at all, as [we only instantiate the `WC_Payments_UPE_Checkout` class](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments.php#L499-L507) instead. `WC_Payments_UPE_Checkout` is only used for split-UPE and non-UPE--not legacy UPE--so this `elseif` branch becomes redundant. When this hook was in the main `WC_Payment_Gateway_WCPay` gateway class this `elseif` was definitely called into action whenever the `UPE_Payment_Gateway` gateway class was used for the legacy UPE, so it's not a redundant and removable clause.

I think we need to divide this function into two and send [this part](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payments-checkout.php#L140-L145) to the `WC_Payments_UPE_Checkout` class, to ensure that this hook enqueues legacy, UPE, and split UPE scripts in all cases where they are required. 

Again I'm not very familiar with this function, but after glancing over the logic here, this seems appropriate to me, but please enlighten me otherwise if you disagree!

FangedParakeet: Ooh, actually ignore this comment--I see what you've done now: you duplicate the hook, just not the function. I suppose some parts of this conditional logic will be completely dependent depending on where this is called, but in the end, I believe all the right pieces will fit into place at the right to me. Cool beans, satisfies me: ignore me lol.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -409,31 +409,41 @@ public function __construct(
...
-		add_action( 'woocommerce_update_options_payment_gateways_' . $this->id, [ $this, 'process_admin_options' ] );
-		add_action( 'admin_notices', [ $this, 'display_errors' ], 9999 );
-		add_action( 'woocommerce_order_actions', [ $this, 'add_order_actions' ] );
+	/**
+	 * Initializes this class's WP hooks.
+	 *
+	 * @return void
+	 */
+	public function init_hooks() {
+		add_action( 'woocommerce_update_order', [ $this, 'schedule_order_tracking' ], 10, 2 );
```
FangedParakeet: I'm just thinking out aloud here, but I was wondering whether this hook should also be included amongst those below only by attached by the main gateway. Currently, the `schedule_order_tracking` function has no gateway specific logic...but I think this is actually a bug.

[This line](https://github.com/Automattic/woocommerce-payments/blob/add/init-hooks/includes/class-wc-payment-gateway-wcpay.php#L2888) will exit out of the hook, if the order's payment method is not equal to the gateway ID, which is fine now...because the only gateway that can be used with subscriptions is the card gateway. However, when SEPA and other PMs are enabled for recurring usage, we will need to add logic to check for their gateway IDs as well.

Basically, I think this hook is currently in the right place and you don't need to change anything, but when we sort out SEPA and subscriptions, we will need to return here and adapt the `schedule_order_tracking` function to work for all purposes. Anyways, apologies for leaving another comment that can be ignored, but this is more so that I remember to return to this in the future lol.

includes/class-wc-payments-upe-checkout.php
```diff
@@ -88,6 +95,43 @@ public function __construct(
...
+			$script_dependencies[] = 'woocommerce-tokenization-form';
+		}
+
+		if ( ! WC_Payments_Features::is_upe_split_enabled() ) {
+			$script = 'dist/upe_checkout';
+		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+			$script = 'dist/upe_with_deferred_intent_creation_checkout';
+		} else {
+			$script = 'dist/upe_split_checkout';
+		}
```
```suggestion
		if ( WC_Payments_Features::is_upe_split_enabled() ) {
			$script = 'dist/upe_split_checkout';
		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
			$script = 'dist/upe_with_deferred_intent_creation_checkout';
		} else {
			$script = 'dist/upe_checkout';
		}
```
I've just noticed that when the new UPE with deferred intent is enabled, the checkout scripts do not get enqueued. This is because when the UPE split is not enabled, we will enqueue the legacy UPE scripts and ignore everything else. 

The small change above resolves this and I believe keeps everything else chugging along too. Let me know if that works for you!

New UPE with deferred intent feature flag
_New UPE with deferred intent feature flag FYI_

includes/class-wc-payment-gateway-wcpay.php
marcinbot: Initially I was checking whether the hook action is checking the ID and then leaving it outside of the check in `init_hooks`. But since this is the only hook left and looks like there will be some dedicated work to add more IDs, I think it's fine to have it inside the `init_hooks` check too. Done in [a9c84fb](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/a9c84fbbe9ae40a7049f5c0c3eb1a5b2a2f88489)

includes/class-wc-payments-upe-checkout.php
marcinbot: Good catch, I think it will also depend on how the features are enabled - it's possible to have Split UPE and deferred one both enabled, and that should work as if only the latter was enabled.

Looks like Deferred Intent is a superset of Split UPE which is a superset of Legacy UPE and it makes sense to perform checks in that order. Done in [fd894f2](https://github.com/Automattic/woocommerce-payments/pull/6326/commits/fd894f2a86b21aec404f746c2ca66a0f2eaa841c)

FangedParakeet: Perfect, that's actually a much preferred order. Nice one!

FangedParakeet (approved):
Thanks for persisting on the wild ride of my scattergun feedback. Everything appears to be working for me now! From my vantage point on this side of the tracks, I'm happy to give this one the green light. LGTM.

a-danae (approved):
> For Subscriptions
Also a light smoke test for cases like subscription creation, renewal, and edge cases such as free trials etc.

I tested WC Subscriptions with these changes using the classic checkout, UPE, and split UPE, and everything worked as expected. No errors in the logs and no unexpected behaviors 👍

PR #6351, opened 2023-05-22
Set UPE order payment method title before email is sent

### LINKED ISSUE #3841: Payment method in the  'Processing Order' email  displayed as 'Popular payment methods' instead of 'Credit / Debit card'
### Describe the bug
The payment method in the 'Processing Order' email is displayed as Popular Payment methods instead of  'Credit / Debit card'.

### Actual behavior

Screen Shot 2022-02-23 at 14 05 03

### To Reproduce

1. Create a test site
2. Set the Store address to a Country different than US
3. Activate WC Pay and complete onboard, make sure the address is set to a country other than US
4. Under WP Admin > Payments > Settings enable UPE checkout
5. Activate Credit card / debit card + other popular payment methods (Bancontact, Giropay, iDeal...)
6. Place an order and proceed to checkout.
7. Make sure that the `Popular payment methods` widget is available as payment option and pay with credit card (note you might need to change customer address to be able to see the credit card):

Screen Shot 2022-02-21 at 14 05 50

8. Check the error in the email that the customer receives. Hint: you might need to install a plugin to catch the emails like [WP Mail Logging](https://wordpress.org/plugins/wp-mail-logging/)

### Expected behavior
The payment method is 'Credit / Debit Card' in the email.

mdmoore:
Fixes #3841 

We're currently [filtering the payment gateway title](https://github.com/Automattic/woocommerce-payments/blob/d4983d2f9a587ecbabaea2b1a12547ded1d3bf07/includes/payment-methods/class-upe-payment-gateway.php#L107) to change the heading displayed on checkout to Popular Payment Methods. We then set the correct payment method title on the order [in the `process_payment` function](https://github.com/Automattic/woocommerce-payments/blob/d4983d2f9a587ecbabaea2b1a12547ded1d3bf07/includes/payment-methods/class-upe-payment-gateway.php#L579). The problem is that emails are already sent before this function fires and the payment method title is still Popular Payment Methods at that time. This PR hooks into the email, setting the correct payment method title, which ensures that emails arrive with the correct payment method title. 

[This adjustment to the CC payment methods `get_title()` method](https://github.com/Automattic/woocommerce-payments/blob/d4983d2f9a587ecbabaea2b1a12547ded1d3bf07/includes/payment-methods/class-cc-payment-method.php#L53) was necessary because this function expects `payment_details` to come in the shape of a charge from the payment intent which contains a top level `network` key. When getting the payment method details via the api client, the network is nested inside a `networks` array. 

#### Testing instructions

1. Enable UPE checkout (legacy) using WCPay Dev Tools
2. Add a product to the cart and proceed to checkout
3. Ensure that Popular Payments Methods is the heading for WooCommerce Payments payment methods
4. Complete checkout using credit card
5. Check for the correct payment method title in the emails sent, such as the New Order and Processing Order emails. The [WP Mail Logging](https://wordpress.org/plugins/wp-mail-logging/) is helpful for catching emails. For example, *Visa credit card* or similar. 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6351` or branch name `fix/3841-email-payment-method-title` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/3841-email-payment-method-title) :rocket:

---

Build info: 

- Latest commit: 2894a9e413c65a4e010ceea4e42dcb137f8543db
- Build time: 2023-06-22 16:00:56 UTC

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1055,6 +1056,20 @@ public function maybe_filter_gateway_title( $title, $id ) {
...
+	 * Sets the payment method title on the order for emails.
+	 *
+	 * @param WC_Order $order   WC Order object.
+	 * @param bool     $sent_to_admin Whether the email is being sent to the admin or not.
+	 * @param bool     $plain_text Whether the email is being sent as plain text or not.
+	 */
+	public function set_payment_method_title_for_email( $order, $sent_to_admin, $plain_text ) {
+		$payment_method_id      = $this->order_service->get_payment_method_id_for_order( $order );
+		$payment_method_details = $this->payments_api_client->get_payment_method( $payment_method_id );
+		$payment_method_type    = $payment_method_details ? $payment_method_details['type'] : null;
```
timur27: This line now appears in three places ([#1](https://github.com/Automattic/woocommerce-payments/blob/7387d421bb0538e56dde54bfeaeb2443483e11ca/includes/payment-methods/class-upe-payment-gateway.php#L551), [#2](https://github.com/Automattic/woocommerce-payments/blob/7387d421bb0538e56dde54bfeaeb2443483e11ca/includes/payment-methods/class-upe-payment-gateway.php#L728), [#3](https://github.com/Automattic/woocommerce-payments/blob/7387d421bb0538e56dde54bfeaeb2443483e11ca/includes/payment-methods/class-upe-payment-gateway.php#L1069)) and I'm starting to think if it's worth it to move it to a separate method. If `$payment_method_id` and `$payment_method_details` fetching would also look identical in all three places, I would strongly consider this refactor. However, since only one field is duplicated, I have no strong opinion. Leaving it up to you, and since it doesn't really block the PR, I won't consider this comment as a blocker for PR approval.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1055,6 +1056,20 @@ public function maybe_filter_gateway_title( $title, $id ) {
...
 	}
 
+	/**
+	 * Sets the payment method title on the order for emails.
+	 *
+	 * @param WC_Order $order   WC Order object.
+	 * @param bool     $sent_to_admin Whether the email is being sent to the admin or not.
+	 * @param bool     $plain_text Whether the email is being sent as plain text or not.
+	 */
+	public function set_payment_method_title_for_email( $order, $sent_to_admin, $plain_text ) {
```
timur27: Is there any reason to keep the two last unused parameters in this method definition, or can we go with:
```suggestion
	public function set_payment_method_title_for_email( $order ) {
```
?

includes/payment-methods/class-cc-payment-method.php
```diff
@@ -50,10 +50,12 @@ public function get_title( $payment_details = false ) {
 			'unknown' => __( 'unknown', 'woocommerce-payments' ),
 		];
 
+		$network = $details['network'] ?? $details['networks']['available'][0];
```
timur27: It's really a nitpick and probably my problem, but if this var were called `$card_network` or similar, the code would be more readable for me. Obviously, we are within `*-cc-payment-method.php` file, but still, for the first few seconds I assumed we deal with some mystic computer network var or similar 😄

timur27 (commented):
Thanks @mdmoore - the PR looks good and tests really well! I've left a few minor comments and will be happy to quickly review it once more after they're resolved. 👍

includes/payment-methods/class-upe-payment-gateway.php
mdmoore: This is a good point! 

>If $payment_method_id and $payment_method_details fetching would also look identical in all three places

The crux of this problem is being unable to get data from the charge on the as is being done in #1 and #2 above *before* the email is generated and sent. So this new function gets the needed data from elsewhere. But the payment method type is still pulled from the payment method details in the same way, so there's still opportunity for some consolidation here. Done via [cba6d9e](https://github.com/Automattic/woocommerce-payments/pull/6351/commits/cba6d9e79464e36efa32d24d7069194121e9922c).

includes/payment-methods/class-upe-payment-gateway.php
mdmoore: I thought so, but now that you've mentioned it, I believe these unused variables can be removed. Took care of it in [cba6d9e](https://github.com/Automattic/woocommerce-payments/pull/6351/commits/cba6d9e79464e36efa32d24d7069194121e9922c).

includes/payment-methods/class-cc-payment-method.php
mdmoore: I like this nitpick! It also wasn't clear to me what network this was dealing with when I first encountered it. `$card_network` is much more clear. [e7976a6](https://github.com/Automattic/woocommerce-payments/pull/6351/commits/e7976a61c23a2adca809c0c2791c92d466c6d871)

mdmoore:
@timur27 - Thanks for the thoughtful comments. I think I've addressed everything if you'd like to give it another look.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1248,6 +1261,16 @@ public function get_payment_method() {
 		return $this->payment_methods['card'];
 	}
 
+	/**
+	 * Return the payment method type from the payment method details.
+	 *
+	 * @param array $payment_method_details Payment method details.
+	 * @return string|null Payment method type or nothing.
+	 */
+	public function get_payment_method_type_from_payment_details( $payment_method_details ) {
```
timur27: I would vote for the most restrictive access level modifier if a method is invoked from inside of the class only.

```suggestion
	private function get_payment_method_type_from_payment_details( $payment_method_details ) {
```

timur27 (approved):
Thank you for addressing my suggestions, Mike! I have retested the flows, including Visa, Mastercard, iDEAL, giropay, and Bancontact, and everything works as expected with the legacy UPE enabled. I also performed quick tests with split UPE and deferred intent creation UPE using cards, and the correct information was provided in emails. 

I've left one last suggestion on the access modifier, but it's so small, so I'm approving your PR anyway and feel free to merge after [the thread](https://github.com/Automattic/woocommerce-payments/pull/6351#discussion_r1238124251) is resolved. 👍

includes/payment-methods/class-upe-payment-gateway.php
mdmoore: Yes, nice catch! Thank again Timur 🚀

timur27:
I forgot to add that it would be great to include testing instructions for this fix in the release testing steps. [I've added them now](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-6.1.0#fix-for-the-legacy-upe---the-correct-payment-method-is-mentioned-in-the-order-email). Feel free to modify/extend 👍

mdmoore:
Thanks for adding those! The steps look great.

PR #6361, opened 2023-05-24
Refactor `AccountBalances` card to allow separate welcome / balances cards

### LINKED ISSUE #6360: Refactor `AccountBalances` card so "greeting" and  "balances" components can be optionally split into two cards
Part of #6353

As part of the the Highlighting active disputes project (pdjTHR-2Hv-p2), we are introducing a notice to the WCPay overview screen if the merchant has active disputes.

This issue covers any refactoring required to enable the [`AccountBalances` card](https://github.com/Automattic/woocommerce-payments/blob/3383af4246cd99b16bc98bd8aad806fce26906a2/client/components/account-balances/index.tsx) to render the [`AccountBalancesHeader` component](https://github.com/Automattic/woocommerce-payments/blob/3383af4246cd99b16bc98bd8aad806fce26906a2/client/components/account-balances/index.tsx#L22) in a separate card to the [`AccountBalancesTabPanel` component](https://github.com/Automattic/woocommerce-payments/blob/3383af4246cd99b16bc98bd8aad806fce26906a2/client/components/account-balances/index.tsx#L24).

Since the "Welcome" greeting will be separated from the `AccountBalances` component, we can consider renaming this component from `AccountBalancesCardHeader`, e.g. to `Welcome`.

**Note:** this issue does not include adding the disputes notice.

### Additional details

See pdjTHR-2F9-p2#comment-3409 for project exploration details related to this issue.
See https://github.com/Automattic/woocommerce-payments/compare/develop...dev/overview-balance-card-welcome-split for prototype.

Jinksi:
Fixes #6360

This PR allows the `AccountBalances` card to be split into 2 cards.

1. The `AccountBalancesHeader` component has been renamed to `Welcome`.
2. `Welcome` code is now located in a distinct component directory: `client/components/welcome/` since it is now decoupled from the `AccountBalances` card.
3. The `AccountBalances` component is now simpler: it no longer includes the `<Card>` wrap or logic related to the `Welcome` component. This is now handled in `overview/index.js`.
4. Unit tests for `AccountBalances` and `Welcome` have been improved or updated.

> **Note**
> Note for reviewer: we could achieve the outcome without change 1 and 2 (renaming & moving the Header component to a distinct component), but I think it makes sense to separate the Header (greeting) component from the Balances component in file structure.

**Before**

image

**After**

image

#### Testing instructions

* Visit Payments → Overview
* If your store has active disputes or not, you should see the existing combined Welcome/Balances card (see screenshot **After**)
* Ensure all tests pass
* Manually edit the Overview screen code ([code ref](https://github.com/Automattic/woocommerce-payments/blob/8c4c85fb08b1f26a3e206fde65b815abf6c15833/client/overview/index.js#L146-L153)) to test that the components work as separate cards:

```tsx
<>
	<Card>
		<Welcome />
	</Card>
	<Card>
		<AccountBalances />
	</Card>

	<DepositsOverview />
</>
```

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6361` or branch name `dev/overview-balance-card-welcome-split` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=dev/overview-balance-card-welcome-split) :rocket:

---

Build info: 

- Latest commit: 8c4c85fb08b1f26a3e206fde65b815abf6c15833
- Build time: 2023-05-31 01:21:47 UTC

changelog/dev-overview-balance-card-welcome-split
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: dev
+Comment: No changelog required: insignificant change, preparation for a future user-facing change.
```
haszari, line 3: Let's leave this blank and ensure we add one changelog per epic.

client/components/account-balances/index.tsx
```diff
@@ -7,19 +7,45 @@ import { Card } from '@wordpress/components';
...
 /**
  * Renders an overview of the account's balances.
  *
+ * @param {Object} props Component props.
  * @return {JSX.Element} Rendered element with the account balances card.
  */
-const AccountBalances: React.FC = () => {
+const AccountBalances: React.FC< Props > = ( {
+	numDisputesNeedingResponse = 0,
+} ) => {
```
haszari: This new prop knows too much – it could be dumber. However, perhaps we should move this logic out to the page so it's clearer …

client/components/account-balances/index.tsx
```diff
@@ -7,19 +7,45 @@ import { Card } from '@wordpress/components';
...
+		// to the balances tab panel.
+		return (
+			<>
+				<Card>
+					<AccountOverviewHeader />
+				</Card>
+				<Card className="wcpay-account-balances">
+					<AccountBalancesTabPanel />
+				</Card>
+			</>
```
haszari: I think it's cleaner and to move the `Card` wrangling out to `overview/index`. The page gets to decide what info it displays, and how to structure it. The components can be whole `Card`s or card content that `OverviewPage` renders in specific cards.

client/components/account-balances/index.tsx
```diff
@@ -7,19 +7,45 @@ import { Card } from '@wordpress/components';
...
+		);
+	}
+
 	return (
 		<Card className="wcpay-account-balances">
-			<AccountBalancesHeader />
+			<AccountOverviewHeader />
```
haszari: Good thinking to rename this. Could potentially name it `welcome` if the primary purpose is to greet merchant.

client/components/account-balances/style.scss
```diff
@@ -67,9 +50,3 @@
 	border-top: 1px solid $gray-200;
 	padding: 24px;
 }
-
-@media screen and ( max-width: $break-mobile ) {
-	.components-flex.wcpay-account-balances__header__flex {
-		flex-direction: column-reverse;
-	}
-}
```
haszari, line 75: Not sure what's going on here – are these no longer needed (or have they moved)? If so yay!

client/components/account-balances/test/__snapshots__/index.test.tsx.snap
```diff
@@ -11,17 +11,17 @@ exports[`AccountBalances renders 1`] = `
...
       >
         <div
-          class="components-flex wcpay-account-balances__header__flex css-1giw1wa-View-Flex-sx-Base-sx-Items-ItemsRow em57xhy0"
+          class="components-flex wcpay-account-overview-header__flex css-1giw1wa-View-Flex-sx-Base-sx-Items-ItemsRow em57xhy0"
           data-wp-c16t="true"
           data-wp-component="Flex"
         >
           <div
-            class="components-flex-item wcpay-account-balances__header__flex__greeting css-mw3lhz-View-Item-sx-Base em57xhy0"
+            class="components-flex-item wcpay-account-overview-header__flex__greeting css-mw3lhz-View-Item-sx-Base em57xhy0"
```
haszari: Assuming this is all consistent renaming.

client/components/account-balances/style.scss
haszari: Ah I see, refactor move => client/components/account-overview-header/style.scss

client/components/account-overview-header/test/index.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+	test( 'renders the correct greeting in the header', () => {
+		const expectedGreeting = 'Good afternoon, Tester 👋';
+		mockGetGreeting.mockReturnValue( expectedGreeting );
+		mockUseCurrentWpUser.mockReturnValue( {
+			user: mockUser,
+			isLoading: false,
+		} );
+		const { getByText } = render( <AccountOverviewHeader /> );
+		getByText( expectedGreeting );
+	} );
```
haszari: What regressions does this test protect us from? How would merchant be impacted if such an issue snuck through to a release?

client/components/account-overview-header/test/utils.test.ts
```diff
@@ -3,7 +3,7 @@
  */
 import { getTimeOfDayString, getGreeting } from '../utils';
 
-describe( 'Account Balances Utils', () => {
+describe( 'Account Overview Header Utils', () => {
```
haszari: Curious about the importance of testing that this API functions as intended.

client/components/account-overview-header/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @return {string} A greeting string.
+ */
+export const getGreeting = (
+	name?: string,
+	date: Date = new Date()
+): string => {
+	const timeOfDay = getTimeOfDayString( date );
+	let greeting = greetingStrings.withoutName[ timeOfDay ];
+	if ( name ) {
+		greeting = sprintf( greetingStrings.withName[ timeOfDay ], name );
```
haszari: I realise this code is not new, but it feels relatively convoluted. Can we not just have a flat component (single file!) which renders a greeting based on props? I don't see the benefits of factoring out the strings, and the logic for generating a greeting, etc.

client/overview/index.js
```diff
@@ -144,7 +144,11 @@ const OverviewPage = () => {
 			{ ! accountRejected && (
 				<ErrorBoundary>
 					<>
-						<AccountBalances />
+						<AccountBalances
+							numDisputesNeedingResponse={
+								numDisputesNeedingResponse
+							}
+						/>
```
haszari: As noted earlier, I think we need to make the decision about which cards to display out here in `client/overview`; the components we use to render the content should be as dumb as possible.

haszari (commented):
This is looking good. Other than naming, I have two suggestions:

1. Move the logic for wrangling cards out to the page. Make the components as simple and dumb as possible.
2. I don't think we should change the behaviour in this PR – i.e. we should invoke the new "split mode" when we implement the new notice. 

Expanding on point 2…

> After: 1+ active disputes

I was assuming this PR would allow developers to toggle "split greeting" mode for testing, e.g. by passing a flag (hacking code or browser state). And that the UI would be unchanged (i.e. for merchants or other WCPay devs).

This is not a big deal as we'll likely get the notice implemented soon (before next release), but I wanted to highlight it as I think it's important that every PR is shippable (unless using a feature flag). Aka `develop` should always be ready for release.

If it's easier to just do the notice and this in the same PR/merge I have no objection to that. I figure the refactor is a nice PR even if it requires hacking a prop to test :)

changelog/dev-overview-balance-card-welcome-split
Jinksi: Do you mean remove the `Comment:`?

Unless the workflow has changed, adding a comment is part of "not adding a changelog" as I understand it.

Part of the `npm run changelog` prompts with:
`You omitted the changelog entry, which is fine. But please comment as to why no entry is needed.`

client/components/account-balances/index.tsx
Jinksi: Thanks, I agree. I was tempted to do this but hesitated, thinking that this refactor was getting too heavy-handed for a relatively small UI change.

With your preference as a blessing, I'll give this a go! 😝

client/components/account-balances/index.tsx
Jinksi, line 1: I've renamed `balances-tab-panel.tsx` to replace `index.tsx`, since it is now the only component returned as part of the `AccountBalances` component.

changelog/dev-overview-balance-card-welcome-split
haszari: Yeah TBH the tool really is annoying – it allows an empty string, and that's the official way to "skip changelog" (as I understand it). Would be great if the tool was a closer match to our workflow.

client/components/account-balances/test/index.test.tsx
```diff
@@ -7,26 +7,11 @@ import { render, screen, fireEvent, within } from '@testing-library/react';
...
-import AccountBalancesTabPanel from '../balances-tab-panel';
-
-import { getGreeting, getCurrencyTabTitle } from '../utils';
-import { useCurrentWpUser } from '../hooks';
+import AccountBalances from '..';
+import { getCurrencyTabTitle } from '../utils';
 import { useAllDepositsOverviews, useInstantDeposit } from 'wcpay/data';
 import { useSelectedCurrency } from 'wcpay/overview/hooks';
 
-const mockUser = {
```
Jinksi, line 19: Tests related to the Welcome greeting component are not relevant to the AccountBalances component, now exist in `client/components/welcome` dir.

client/components/account-overview-header/test/utils.test.ts
Jinksi: For background: I wrote these util functions with TDD to ensure I met the requirements.

I think it's worth keeping: if this were to respond with "good undefined, undefined" for some unknown reason, that would be an ugly bug that this test would catch.

haszari: Your call :) if this is testing implementation details, we'll get pinged enough to revise it over time.

client/components/account-overview-header/utils.ts
Jinksi: Logic: I'd lean towards leaving this as a util function to ease testing.
Strings: can be moved to be incorporated in this `utils.ts` file, separation was deemed suitable when part of the larger `AccountBalances` component to keep strings in one location. I'll make this change.

haszari: I'm ok with this approach :)

> Logic: I'd lean towards leaving this as a util function to ease testing.

Referring to my earlier comments about testing implementation details. If e2e or TDD is pushing us to make our system more complex, we need to be sure it's worth it. 

Does this help us confirm the software is working for merchants? Or does it make it harder to iterate on or change how this works? How many clients rely on this API (dividing day into 3 named periods, and mapping that to some strings) – is an abstraction needed? Can we test this logic as part of a larger e2e test (e.g. `Ensure payments overview screen shows a welcoming message`)

client/overview/index.js
```diff
@@ -144,11 +145,25 @@ const OverviewPage = () => {
 			{ ! accountRejected && (
 				<ErrorBoundary>
 					<>
-						<AccountBalances
-							numDisputesNeedingResponse={
-								numDisputesNeedingResponse
-							}
-						/>
+						{ 0 < numDisputesNeedingResponse ? (
```
Jinksi: This is a strange way to order the statement IMO, but enforced by the "yoda condition" eslint rule.

client/overview/index.js
Jinksi: Addressed in c09d1e3

client/components/account-balances/test/__snapshots__/index.test.tsx.snap
Jinksi: This snapshot test has now been removed, since the header component is not tested as part of AccountBalances. 👍

client/components/account-balances/index.tsx
Jinksi: I've addressed this in c09d1e3

client/overview/index.js
haszari: Can always use a local variable to make this less awkward, e.g. 

```js
const someDisputesNeedAttention = ( 0 < numDisputesNeedingResponse );
```

client/components/account-balances/index.tsx
Jinksi: Thanks, `Welcome` is less of a mouthful. I'll give it a go.

Jinksi: Renamed to `Welcome` in e31fedc 🤩

client/components/account-overview-header/test/utils.test.ts
Jinksi: I _could_ delete this entire file and don't have a strong reason not to.

I've updated the component-level tests to catch any empty username bug – 09bec96.

client/overview/index.js
Jinksi: Thanks, I've used your suggestion in fe1d978

client/components/account-overview-header/utils.ts
Jinksi: > Can we test this logic as part of a larger e2e test (e.g. Ensure payments overview screen shows a welcoming message)

I've moved these tests to the component-level in 09bec96, so leaning towards moving the utils inside the component file and removing unit tests, given your advice.

I am concerned about how big this refactor PR is becoming and IMO I should have opened a new PR for unit test refactoring (different outcome to this PR's intended goal).

client/components/account-balances/index.tsx
```diff
@@ -1,28 +1,265 @@
...
+															// eslint-disable-next-line jsx-a11y/anchor-has-content
+															<a
+																rel="external noopener noreferrer"
+																target="_blank"
+																href={
+																	documentationUrls.depositSchedule
+																}
+															/>
+														),
+													},
```
haszari, line 185: Looks like tabs snuck in somewhere along the way :) (we use spaces right?)

client/components/welcome/index.tsx
```diff
@@ -13,27 +13,27 @@ import wooPaymentsLogo from 'assets/images/woopayments.svg?asset';
 import './style.scss';
 
 /**
- * Renders the header of the account's balances card.
+ * Renders a welcome card header with a greeting and the Woo Payments logo.
```
haszari, line 70: This will ultimately also display important notices, right?

client/components/account-balances/index.tsx
haszari: Wait … does WCPay code use tabs!? TIL

I must have an issue with the GitHub diff viewer – the tabs are much bigger than I expect.

Jinksi: Apparently, we use tabs! (based on `.prettierrc`)

I never think about this, prettier does all of the auto-formatting on save.

https://github.com/Automattic/woocommerce-payments/blob/99de2876c4bd5335d87136ddc4673db847235857/.prettierrc#L2-L3

client/overview/index.js
Jinksi: No longer relevant since we're not splitting card in this PR. 8b31b06

client/components/welcome/index.tsx
Jinksi: Thanks! I've added this detail in 41ae333

client/components/account-overview-header/test/utils.test.ts
Jinksi: 🙈 I've refactored the `Welcome` component and tests in 2af9591:
* Removed `utils.ts`
* Moved logic into component file `index.tsx`
* Removed redundant tests from `utils.test.ts` as the major edge-cases are now tested in `test/index.tsx`

client/components/account-overview-header/utils.ts
Jinksi: As I mentioned above (https://github.com/Automattic/woocommerce-payments/pull/6361#discussion_r1206184938):

🙈 I've refactored the `Welcome` component and tests in 2af9591:
* Removed `utils.ts`
* Moved logic into component file `index.tsx`
* Removed redundant tests from `utils.test.ts` as the major edge-cases are now tested in `test/index.tsx`

client/components/account-balances/index.tsx
Jinksi: > I must have an issue with the GitHub diff viewer – the tabs are much bigger than I expect.

I was thinking the same, has something in GH changed recently?

Jinksi: Changing this [GH appearance setting](https://github.com/settings/appearance) makes it a bit less... spacious.

image

client/components/welcome/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+	let greeting = greetingStrings.withoutName[ timeOfDay ];
+	if ( name ) {
+		greeting = sprintf( greetingStrings.withName[ timeOfDay ], name );
+	}
+	greeting += ' 👋';
+	return greeting;
+};
+
+/**
+ * Renders a welcome card header with a greeting, the Woo Payments logo and important notices.
```
Jinksi:
```suggestion
 * Renders a welcome card header with a greeting and the Woo Payments logo.
```

I'll change this back to clarify that this component only returns the card header and not the important notices.

client/components/welcome/index.tsx
Jinksi: **Update**

@haszari I've reverted this description to clarify that this component only returns the card header (greeting + logo). 7c66faf

changelog/dev-overview-balance-card-welcome-split
haszari: Still recommend clearing out this string – that's the official way to say "no changelog".

client/components/account-balances/index.tsx
```diff
@@ -1,28 +1,265 @@
...
- * @return {JSX.Element} Rendered element with the account balances card.
+ * @typedef {Object} BalanceTab
+ *
+ * @param {string} name           Name of the tab.
+ * @param {string} title          Title of the tab.
+ * @param {string} currencyCode   Currency code of the tab.
+ * @param {number} availableFunds Available funds of the tab.
+ * @param {number} pendingFunds   Pending funds of the tab.
+ * @param {number} delayDays	  The account's pending period in days.
+ */
```
haszari, line 33: I see this type is used in a bunch of places – nice use of an interface / struct type to make this consistent and give it a name! Might be even clearer to add a summary/title – e.g. `Props used for displaying an account balance in a specific currency`, and/or name `BalanceTabProps`. (nitpick / not a blocker, this is fine as is)

client/components/account-balances/index.tsx
```diff
@@ -1,28 +1,265 @@
...
+ * @return {JSX.Element} Rendered balances panel with tab navigation for each currency.
  */
 const AccountBalances: React.FC = () => {
-	return (
-		<Card className="wcpay-account-balances">
-			<AccountBalancesHeader />
+	const {
+		overviews,
+		isLoading,
+	} = useAllDepositsOverviews() as AccountOverview.OverviewsResponse;
```
haszari: What does the `as AccountOverview.OverviewsResponse` do? It looks like a cast, but `useAllDepositsOverviews()` returns that type anyway.

client/components/account-balances/index.tsx
```diff
@@ -1,28 +1,265 @@
...
+							className="wcpay-account-balances__instant-deposit"
+						>
+							<InstantDepositButton
+								instantBalance={ tab.instantBalance }
+							/>
+						</Flex>
+					) }
+				</>
+			) }
+		</TabPanel>
```
haszari, line 259: I skimmed this component – assuming it's relatively unchanged from previous `BalanceTabPanel`.

client/components/account-balances/utils.ts
```diff
@@ -6,49 +6,7 @@ import { sprintf } from '@wordpress/i18n';
...
-): string => {
-	const timeOfDay = getTimeOfDayString( date );
-	let greeting = greetingStrings.withoutName[ timeOfDay ];
-	if ( name ) {
-		greeting = sprintf( greetingStrings.withName[ timeOfDay ], name );
-	}
-	greeting += ' 👋';
-	return greeting;
-};
+import { currencyBalanceString } from './strings';
```
haszari: I realise this is not new, but noting another place where having the string inline would be more readable and reduce dependencies / complexity.

```js
export const getCurrencyTabTitle = ( currencyCode: string ): string => {
	return sprintf(
		// string format: {currency} balance
		//currencyBalanceString,
        // Inline strings are self documenting 🎉 
        __( '%s Balance', 'woocommerce-payments' ),
		currencyCode.toUpperCase()
	);
};
```

client/components/welcome/test/index.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+		};
+		const expectedGreeting = /Good (morning|afternoon|evening) 👋/;
+		mockUseCurrentWpUser.mockReturnValue( {
+			user: mockUser,
+			isLoading: false,
+		} );
+		const { getByText } = render( <Welcome /> );
+		getByText( expectedGreeting );
+	} );
+} );
```
haszari, line 61: Nice, testing a well-defined component – much less likely to cause "test churn" next time someone needs to iterate on this UI :)

haszari (approved):
Looking good! I added some comments about minor details (discussion not feedback) – nothing blocking, merge at will :) 

I tested by factoring out the top 1-2 cards as a single component. I suggest adding that info to the PR description so it's clear how to test the new behaviour.

```jsx
const WelcomeAndBalances = () => {
	return (
		<ErrorBoundary>
			<>
				<Card>
					<Welcome />
					<AccountBalances />
				</Card>

				<DepositsOverview />
			</>
		</ErrorBoundary>
	)
};

const WelcomeAndBalancesSplit = () => {
	return (
		<ErrorBoundary>
			<>
				<Card>
					<Welcome />
				</Card>
				<Card>
					<AccountBalances />
				</Card>

				<DepositsOverview />
			</>
		</ErrorBoundary>
	)
};
```

changelog/dev-overview-balance-card-welcome-split
Jinksi: In this case, what do you make of the `You omitted the changelog entry, which is fine. But please comment as to why no entry is needed.` prompt?

I interpret this as:
* leave an empty changelog entry (line 4)
* leave a comment when prompted (line 3)

Jinksi: In this case, what do you make of the `You omitted the changelog entry, which is fine. But please comment as to why no entry is needed.` prompt?

I interpret this as:
* leave an empty changelog entry (line 4)
* leave a comment when prompted (line 3)

image

haszari: Oh … does the `Comment` line show vs. `Changelog` for unskipped? If so that makes sense, I get it now!

client/components/account-balances/index.tsx
Jinksi: This is true, removed in 0f33c1a92a370d2fe01e497fb09515e905318a66.

Jinksi: This is true, removed in 0f33c1a92a370d2fe01e497fb09515e905318a66.

client/components/account-balances/utils.ts
Jinksi: Nice catch 🙏 I've simplified this in d17dabb48e139fd2a691feacb240e8b2afce4ec1

client/components/account-balances/index.tsx
Jinksi: Thanks, renaming this helps to clarify it's meaning. I've renamed this to `BalanceTabProps` in 8c4c85fb08b1f26a3e206fde65b815abf6c15833

PR #6367, opened 2023-05-24
Protect WooPay styles from being easily overridden by themes

### LINKED ISSUE #6327: Fix WooPay button theming and markup issues
### Describe the bug

The WooPay express checkout button has a few theming conflicts. Most of these are general to all the express checkout options, but this issue is focused on three things that are specific to WooPay that we should address quickly. 

- The `id=wcpay-payment-request-wrapper` is included twice when there is a WooPay and Google/Apple Pay button on the page. This should likely be a class instead to be syntactically correct.
- WooPay button color should not be overriden by theming for branding purposes. This will likely require a ! important on the colors for the WooPay button as we can't depend on the specificity of the selector.
- The WooPay sign up checkbox on checkout tooltip should have a text color defined so there is always contrast with the background.

### To Reproduce

For the ID issue:
1. Go to a merchant store with 2 express checkout options enabled, WooPay + Google / Apple Pay.
2. Go to a page where both would be displayed (product, cart, checkout).
3. Inspect the buttons in the browser and notice that the HTML id `wcpay-payment-request-wrapper` is used twice.

For the button color issue:
1. Go to a merchant store with WooPay enabled.
2. Go to a page where it would be displayed (product, cart, checkout).
3. Add some styling in the browser to override the background color for the button. For example on the cart: `.page .cart_totals button { background: black; }`. The background, border, and color attributes at a minimum should not get overriden by theme styles.

For the sign up checkbox issue:
1. Go to a merchant store with WooPay enabled.
2. Go to the checkout page where the WooPay sign up checkbox would be displayed.
3. Add some styling in the browser to set the font color to white for a broad set of options. For example, `body span { color: #fff; }`. This particular example will impact the label text as well, but we're only concerned with the tooltip since we are manually setting the background color there, we should be sure that we force contrast with the text.

### Actual behavior

For the ID issue: There are two HTML elements with the same ID of `wcpay-payment-request-wrapper`
For the button color issue: The WooPay button colors will pick up theme colors.
For the sign up checkbox issue: The theme may override the tooltip font color and make it unreadable due to no contrast.

### Screenshots

Screen Shot 2023-05-18 at 10 35 12 AM

Screen Shot 2023-05-18 at 10 41 07 AM

### Expected behavior

For the ID issue: The duplicate ID should instead be a class and any references to that ID should be replaced with a class reference (CSS or JS selectors).
For the button color issue: The WooPay button colors will remain consistent regardless of the theme.
For the sign up checkbox issue: The text in the tooltip will always be legible.

### Additional context

For more context reference this issue and its corresponding discussion: Automattic/woopay#1854

mdmoore:
Fixes #6327 

This fixes the 3 issues mentioned in #6327. The duplicate `id` attribute has been changed to a `class` instead. The styling issues have been addressed by forcing important styles via the `!important` tag.

#### Testing instructions

**For the ID issue:**
1. Go to a merchant store with 2 express checkout options enabled, WooPay + Google / Apple Pay.
2. Go to a page where both would be displayed (product, cart, checkout).
3. Inspect the buttons in the browser and notice that the `wcpay-payment-request-wrapper` class is used twice and no `id` is used.

**For the button color issue:**
1. Go to a merchant store with WooPay enabled.
2. Go to a page where it would be displayed (product, cart, checkout).
3. Add some styling in the browser to override the background color for the button. For example on the cart: `.page .cart_totals button { background: black; }`. The background, border, and color attributes should not be overridden by theme styles. Of course, it will still be possible with some very specific styling that also uses the `!important` tag.

**For the sign up checkbox issue:**
1. Go to a merchant store with WooPay enabled.
2. Go to the checkout page where the WooPay sign up checkbox would be displayed.
3. Add some styling in the browser to set the font color to white for a broad set of options. For example, `body span { color: #fff; }`. Check that the text in the tooltip flyout is not affected by the styling added and that the text has contrast against the background.

**Sign-up checkbox alignment issue:** 

1. Go to a merchant store with WooPay enabled.
2. Go to the checkout page where the WooPay sign up checkbox would be displayed.
3. Add some styling in the browser to set a `line-height` of something like 2 on `span` elements in an attempt to affect the the checkbox label text. The text should stay aligned with the checkbox. At narrow widths, when the label text wraps to the next line, the text should not wrap below the checkbox, instead remaining in it's own "column". 

**Note** - It's kind of tough to control what happens to the layout if some unwieldy line-height is added. One solution to this would be setting a line height for this header section which could potentially lead to visual disparity with other elements on the page. However, I think it's unrealistic for sites to use line-heights at a value that would cause problems as the rest of the page would also look broken. 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6367` or branch name `fix/6327-woopay-theming-and-markup` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6327-woopay-theming-and-markup) :rocket:

---

Build info: 

- Latest commit: 893769905ced1e28c04b3a05cef7ea84e4c997f2
- Build time: 2023-06-01 16:34:06 UTC

client/components/woopay/save-user/style.scss
```diff
@@ -88,6 +90,10 @@
 				display: flex;
 			}
 
+			span {
+				color: #333 !important;
```
bborman22: One thing to call out here is that the original issue is just that we only defined a background color on the flyout and not a text color on it. This led to the theme's text color being applied (white), but the background used our styles (also white). All that to say I'm not sure we need an important here or if we do, we should also have it on the background color as well to ensure contrast. Let me know your thoughts on getting those two things in sync?

bborman22 (approved):
All of these fixes worked for me. I also applied to stylesheet to several other stores to confirm they fix the specified issues and they did where I was able to test them. And finally I tested locally against several different themes and even with modifications attempting to break the listed issues and didn't have any issues.

I know the use of `!important` can be a red flag, but I believe in this instance that is the correct solution as we want to protect branding and make sure that any overriding is very intentional by the theme developer and not by accident.

I left one comment around something we may just want to get in sync with the style applications, but otherwise it LGTM!

client/components/woopay/save-user/style.scss
mdmoore: >All that to say I'm not sure we need an important here

Good callout! I'm happy removing important since it will now require intention to change the color.

bborman22: Perfect, thanks for addressing that!

LGTM!

bborman22:
Thanks for getting that one call out addressed, I think this is good to merge now!

Issue #6353, opened 2023-05-23
Epic: WCPay Disputes – Highlight active disputes in Payments → Overview screen


haszari:
### Description
The goal of this epic is to improve the visibility of active disputes in the payment overview screen, and ultimately make it easier for merchants to respond to disputes.

![payment-overview](https://github.com/Automattic/woocommerce-payments/assets/3147296/986f469e-0a5b-47b9-831b-8b8262569cc4)

![woo-home](https://github.com/Automattic/woocommerce-payments/assets/3147296/35de076e-a660-4832-896f-f438d52bb23c)

### Steps

- [x] #6360
- [x] #6382
- [x] #6370
- [x] #6404
- [x] #6403
- [x] #6485
- [x] #6505
- [x] Tracks event registration
	- [x] #6541
	- Note: Payments → overview disputes task tracking is already setup under the existing `wcadmin_wcpay_overview_task` event, although it has not been registered: https://github.com/Automattic/woocommerce-payments/blob/b39314fbbfa4bf18a21928fc02345a052a07f1bf/client/overview/task-list/dispute-tasks.ts#L33-L35. In #6371 this event name has been corrected to `wcadmin_wcpay_overview_task_click` and will be registered.
- [x] Add testing instructions to this description for impacted flows
	- [x] Active Disputes Overview Task
	- [x] Active Disputes WC Home Task
- [x] Check if there is any impact on critical flows testing instructions – **No impact**
- [x] Update the release testing instructions
- [x] Update WCPay release thread

The above is a high-level summary of the changes in the design. These can be split off as separate issues. Can add / adapt if details change.

### Additional context (internal links)
- Project thread: pdjTHR-2Hv-p2
- Exploration thread: pdjTHR-2F9-p2
- Design i1: paJDYF-8Cm-p2

# Testing flows

## Payments Overview Active Dispute Task

* View the Payments → Overview screen with the following configuration of disputes and confirm it matches the requirements:
	* 0 active disputes (task not visible)
	* 1+ active disputes due more than 7 days in the future (task not visible)
	* 1 active dispute due within 7 days
	* 1 active dispute due within 24h
	* 2+ active disputes of multiple currencies due within 7 days
	* 2+ active disputes of multiple currencies due within 24h
* Check that the task list looks correct with a non-dispute task item enabled alongside the dispute task item.
[...]

PR #6371, opened 2023-05-24
Improve the wording and style of the "Active Disputes" task list item on the Payments → Overview screen to highlight urgency

### LINKED ISSUE #6370: [Disputes task item] Update the active disputes task item wording in the Payments → Overview `Things to do` task list
Part of #6353

As part of the the Highlighting active disputes project (pdjTHR-2Hv-p2), ~we are introducing a notice to the WCPay overview screen if the merchant has active disputes~. 💡 **Update**: we are moving the `Things to do` task list to the `Welcome` component.

This issue updates the wording of the active disputes `Things to do` task list item. The dynamic colour and subtitle are handled by subsequent issues in #6353. 

payment-overview

As per the project's exploration thread (pdjTHR-2F9-p2#comment-3408) the wording for this notice should be the following:

* Single active dispute: `Respond to a dispute for $59.99`
* Single active dispute, <24 hours to respond: `Respond to a dispute for $59.99 – last day`
* Multiple active disputes: `Respond to 3 active disputes for a total of $259.97`

The `Respond now` button should click through to the disputes screen.

### LINKED ISSUE #6403: [Disputes task item] Show subtitle on active disputes task list item with info about most urgent dispute deadline
Part of #6353

As part of the the Highlighting active disputes project (pdjTHR-2Hv-p2) we are updating the active dispute `Things to do ` task to improve visibility and highlight urgency.

This issue updates the wording of the task subtitle.

image

### LINKED ISSUE #6404: [Disputes task item] Use colours and wording to highlight urgent disputes on active disputes task list item.
Part of #6353

As part of the the Highlighting active disputes project (pdjTHR-2Hv-p2) we are updating the active dispute `Things to do` task to improve visibility and highlight urgency.

This issue updates the colour of the active disputes task depending on the proximity of the dispute deadline.

* Within 24h: Red
* Within 7 days: Yellow
* Not displayed in list otherwise

image

### LINKED ISSUE #6485: Dismissed dispute task shouldn't stay hidden even though there is new dispute
### Describe the bug

When there is a dispute, it will appear as task on the overview page about the dispute. When merchant dismiss that task (despite whether they have resolved the dispute or not), the task will stay dismissed. Even though after that there is a new dispute, the task will stay hidden and it's a big risk because merchant might not get alerted about the new dispute from the task list.

As @Jinksi [mentioned](https://github.com/Automattic/woocommerce-payments/pull/6397#issuecomment-1576013516), the solution might be to come up with a new key for the dispute task. Part of the work for this issue is to make sure changing that key does not have negative impact, or find another solution if possible. Is the state of the task, whether dismissed or not, stored somewhere? Can we make it reset the state when there is new dispute?

### To Reproduce

1. Purchase a product using one of the [test cards](https://stripe.com/docs/testing#disputes) to result in a dispute.
2. Notice there is a dispute task on the Payments > Overview page.
3. Dismiss the task by clicking the three dots menu on the right of the task.
4. Make another dispute.

### Actual behavior

Dispute task will stay hidden.

### Expected behavior

Dispute task should appear.

### Additional context

Related to:
- https://github.com/Automattic/woocommerce-payments/pull/6397 which moves the task list to the greeting card at the top, and
- https://github.com/Automattic/woocommerce-payments/pull/6371 which makes task item not dismissible.

Jinksi:
* Fixes #6370
* Fixes #6404
* Fixes #6403
* Fixes #6485

This PR updates the title, subtitle and colours and action button label for the Payments → Overview `Things to do` active disputes task list item. It also makes the task dismissible and ensures the task will appear for stores who have previously dismissed this task.

payment-overview

**Single/multiple disputes with > 7 days until due**
* **Visibility**: Not visible

**Single dispute with < 7 days until due**
* **Visibility/colour**: 🟡 `Yellow`
* **Title**: `Respond to a dispute for $123.00`
* **Subtitle**: `By Jun 30, 2023 – 1 week left to respond`
* **Action label**: `Respond now`
* **Action redirect**: Single dispute details page

**Single dispute with < 24 hours until due**
* **Visibility/colour**: 🔴 `Red`
* **Title**: `Respond to a dispute for $123.00 – Last day`
* **Subtitle**: `Respond today by 11:59 PM`
* **Action label**: `Respond now`
* **Action redirect**: Single dispute details page

**Multiple disputes with at least one due within 7 days**
* **Visibility/colour**: 🟡 `Yellow`
* **Title**: `Respond to N active disputes for $123.00, €123.00`
* **Subtitle**: `Last week to respond to N of the disputes`
* **Action label**: `See disputes`
* **Action redirect**: Active disputes list screen

**Multiple disputes with at least one due within 24 hours**
* **Visibility/colour**: 🔴 `Red`
* **Title**: `Respond to N active disputes for $123.00, €123.00`
* **Subtitle**: `Final day to respond to N of the disputes`
* **Action label**: `See disputes`
* **Action redirect**: Active disputes list screen

**Tracks event**

The unregistered tracks event `wcpay_overview_task` has been renamed `wcpay_overview_task_click` to meet tracks event naming guidelines. This event was only in use for the task `dispute_resolution_task`. It now accepts the event property `task` and `active_dispute_count`. It will be registered once this PR has been merged: see issue #6541.

**Design/component library tradeoffs**

The task list items use the `TaskItem` component from `@woocommerce/experimental`.

This raises some issues/limitations:

1. The ❗ icon does not appear if the `TaskItem` `level` is `2` (yellow).
1. There is a tooltip that displays when hovering the icon/checkbox:
	* level 1 (red): `This task is required to keep your store running`
	* level 2 (yellow): `This task is required to set up your extension`

💡 I've solved this by using level 1 for both red/yellow and overriding colour via CSS.

---

### Testing instructions

* Tip: To create a disputed transaction, purchase a product with the Stripe test card `4000000000000259`.
* View the Payments → Overview screen with the following configuration of disputes and confirm it matches the requirements above:
	* 0 active disputes (task not visible)
	* 1+ active disputes due more than 7 days in the future (task not visible)
	* 1 active dispute due within 7 days
	* 1 active dispute due within 24h
	* 2+ active disputes of multiple currencies due within 7 days
	* 2+ active disputes of multiple currencies due within 24h
* Check that the task list looks correct with a non-dispute task item enabled alongside the dispute task item.
	* Mark these tasks as visible by returning `true` for `showUpdateDetailsTask` and `wpcomReconnectUrl` in [`WC_Payments_Admin::get_js_settings()`](https://github.com/Automattic/woocommerce-payments/blob/5a3909d4823d3794f53f73fed5d468ca905ca656/includes/admin/class-wc-payments-admin.php#L763-L764).

* Check that the currency amounts look correct after modifying the WC country:
	* Navigate to WooCommerce > Settings > General.
	* Change the Country / State to a region with non-decimal (Japan, Yemen) or three-decimal (Bahrain, Kuwait) display.

**Mocking dispute due_by when testing**

To mock dispute `due_by` dates testing purposes, add the following logic and adjust the 'days' value: https://github.com/Automattic/woocommerce-payments/blob/56dd736e4b2d335ffcdd4e4a6b676d7df0861c78/client/overview/index.js#L75
```diff
-		activeDisputes,
+		activeDisputes: activeDisputes.map( ( dispute ) => {
+			const days = 6;
+			dispute.due_by = new Date(
+				Date.now() + days * 24 * 60 * 60 * 1000
+			).toISOString();
+			return dispute;
+		} ),
```

**Screenshot with multiple currency disputes, a non-decimal country (Japan) and other tasks**
image

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6371` or branch name `add/overview-active-disputes-notice` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/overview-active-disputes-notice) :rocket:

---

Build info: 

- Latest commit: 3eaf1885d6473ef20d5de35872fcfd2163340817
- Build time: 2023-06-15 22:13:10 UTC

changelog/add-overview-active-disputes-notice
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Display a notice on the Payments → Overview screen when there are active disputes that require a response.
```
Jinksi: Adding this changelog entry with the intention that it will likely cover all issues in #6353

client/disputes/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+import { formatCurrency } from 'wcpay/utils/currency';
+
+interface GetDisputesNoticeStringArgs {
+	/**
+	 * Active disputes (awaiting a response) to generate the notice string for.
+	 */
+	activeDisputes: CachedDispute[];
+}
+
+export const getDisputesNoticeString = ( {
```
Jinksi: Creating this function in `disputes/utils`, allowing re-use in `active-disputes-notice` and "Things to do" task item (future PR).

Jinksi:
Note for reviewer: although implementation does not 100% match the designs, I am requesting a review to start a discussion. In particular, see **Design/component library tradeoffs** in the PR description.

client/disputes/utils.ts
haszari: Thanks for clarifying; my first question was going to be "is this going to be reused"!

If it was me I'd implement it inline and refactor for the second client (or later).

client/disputes/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+import type { CachedDispute } from 'wcpay/types/disputes';
+import { formatCurrency } from 'wcpay/utils/currency';
+
+interface GetDisputesNoticeStringArgs {
+	/**
+	 * Active disputes (awaiting a response) to generate the notice string for.
+	 */
+	activeDisputes: CachedDispute[];
+}
```
haszari: Is this a recommended pattern – defining a type for one function's args? Is it impractical to do this inline, since this "type" is tightly coupled to the function?

Alternatively, is this a generally useful struct type – if so, could rename it.

client/disputes/test/utils.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+	it( 'returns the correct string for multiple disputes', () => {
+		expect(
+			getDisputesNoticeString( {
+				activeDisputes: mockDisputes.slice( 0, 2 ),
+			} )
+		).toEqual( `Respond to 2 active disputes for a total of $22.34` );
+	} );
+
+	it( 'returns the correct string for multiple disputes with multiple currencies', () => {
```
haszari: What do you think about testing the components render the right strings etc, rather than testing this utility? IMO the util is an implementation detail that future engineers should be free to iterate on and adjust without tests getting in the way.

client/disputes/utils.ts
haszari: (aka premature generalisation)

client/disputes/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+		return null;
+	}
+
+	if ( disputeCount === 1 ) {
+		const dispute = activeDisputes[ 0 ];
+		const now = moment();
+		const dueBy = moment( dispute.due_by );
+
+		const dueWithin24h =
+			dueBy.diff( now, 'hours' ) > 0 && dueBy.diff( now, 'hours' ) <= 24;
```
haszari: This logic is going to be useful for the dispute urgency/colour scheme. So maybe it is better to implement this logic all inline in the component now, and factor out small functions for shared needs across multiple components.

haszari (commented):
Added some early comments.

TLDR: I don't think `getDisputesNoticeString` is a necessary abstraction. I prefer to factor out shared logic as it is needed (in small chunks), and implement (and test/verify) self-contained components.

client/disputes/utils.ts
Jinksi: Yeh I was borderline about this (PS I wrote this function before our discussion in the other PR #6361 so I unapologetically wrote the `utils` file before the component in a TDD fashion 😝).

I know we will have at least _some_ shared dispute logic, I leaned towards creating the `disputes/utils` file as a framework for this project, in a similar approach to the simplify deposits project.

Jinksi: Although still a WIP draft, I'm working on the "Things to do" wording PR #6383 in parallel and it looks like sharing this `getDisputesNoticeString` util function will be the simplest way to implement:

https://github.com/Automattic/woocommerce-payments/pull/6383/files#diff-062eb1e5dc64b4146f2536ba4987039bedb128b2d938de7217119a9beaa2fd87R185

client/disputes/utils.ts
Jinksi: Agree this is a bit heavy for a single function arg. I have no preference for a pattern here.

I've changed this to a single arg function in in 335587d.

client/disputes/utils.ts
Jinksi: I've now moved this logic within the `client/overview/task-list/dispute-tasks.ts` file, consistent with logic for other task list items.

client/disputes/test/utils.test.ts
Jinksi: I've updated the tests for the consumer of this function (`getTasks()`) in `client/overview/task-list/test/tasks.js`. This removes the need to test individual util functions, which have been removed.

client/disputes/utils.ts
Jinksi: I've now moved this logic within the `client/overview/task-list/dispute-tasks.ts` file, consistent with logic for other task list items.

shendy-a8c (commented):
~Any reason why we're moving away from using the number of disputes needing response (or active disputes) and use a list of disputes instead?~

~The `numDisputesNeedingResponse` variable sounds so familiar and turns out it's [something we worked on last year](https://github.com/Automattic/woocommerce-payments/pull/4388) to solve pagination issue, which now I wonder why didn't we just use `filter: 'awaiting_response'`.~

~Anyway, I see that in `getTasks()`, either we use `numDisputesNeedingResponse` or `activeDisputes`, it's eventually used to determine whether there is visible dispute only.~

Never mind, I just saw that dispute objects are needed by `getDisputeResolutionTask()`.

haszari:
> Any reason why we're moving away from using the number of disputes needing response (or active disputes) and use a list of disputes instead?

The other consideration here is that multiple active disputes is an edge case. The most common case is 0 or 1 dispute, so we're optimising the experience for the common case – 1 dispute to follow up.

Jinksi:
Note: after our design walkthrough I should mention that this change impacts the wording of the task. Other adjustments will be handled in other PRs from this Epic:
* Colours based on urgency (yellow/red)
* Subtitle
* Only display if < 7 day due_by

What **should** be included in this PR is the dynamic button label. I'll mark this as `in_progress` to address this and clarify the requirements.

shendy-a8c:
@Jinksi, great work especially the call out on the details like multi-currencies, `TaskItem` tooltip, etc.

I should have posted this yesterday. Here's my feedback:

1. Because `useDisputes()` will only return at most `perPage` `awaiting_response` disputes, I think when the number of returned disputes is `perPage` or more, the text should be
```
Respond to 25+ active disputes for a total of $259.97 or more.
```
where 25 is the value of `perPage`.
Maybe the text copy should be something different but you get my point.

However, per discussed yesterday, as @haszari said, it might not be too important to be too invested on this, so consider my feedback here as a `nit`. I think it's just a small effort to accommodate this ask though.

2. About your suggestion on `TaskItem`:
> 💡 Idea to solve 2 & 3: we could stick to using level 1 for both red/yellow and override colour via CSS.

IMO that's a great idea. I don't see it implemented in this PR, so please move forward with it.

The content of the tooltip

3. Another thing we found out yesterday from the design overview meeting, the task should be:
- Color-coded as red when there is at least one dispute to be resolved in 24 hours or less.
- Otherwise, color-coded as yellow when there is dispute to be resolved in 7 days or less.
- Otherwise, no need to display the task.
- Clicking the task should go to the individual dispute if there is only 1 dispute. Otherwise, to the disputes list page, filtered `awaiting_response`.

Jinksi:
> Another thing we found out yesterday from the design overview meeting, the task should be:

@shendy-a8c Thanks! I'll make these changes and incorporate all of these requirements as part of this PR.

After the changes, this PR will resolve:
* #6404
* #6403
* #6485

> where 25 is the value of perPage

Estimating that 99.99% of merchants will have <25 disputes due in any 7day period, 25 may be OK. Increasing `perPage` to 50 or 100 may be an easy and suitable adjustment though, so I'll take a look.

client/transactions/declarations.d.ts
```diff
@@ -103,6 +103,16 @@ declare module '@woocommerce/components' {
 }
 
 declare module '@woocommerce/navigation' {
+	import type { BrowserHistory, Location } from 'history';
+	/**
+	 * Extension of history.BrowserHistory but also adds { pathname: string } to the location object.
+	 */
+	interface WooBrowserHistory extends BrowserHistory {
```
Jinksi, line 110: Creating this interface allows us to use the `getHistory` function with a correct return type.

client/overview/style.scss
```diff
@@ -13,7 +13,17 @@
 	position: relative;
 
 	.woocommerce-experimental-list__item {
-		position: initial;
+		position: relative;
```
Jinksi, line 16: This fixed a styling bug where the border-left of one task was overflowing across multiple tasks.

client/overview/task-list/index.js
```diff
@@ -169,6 +169,9 @@ const TaskList = ( { overviewTasksVisibility, tasks } ) => {
 			{ visibleTasks.map( ( task ) => (
 				<TaskItem
 					key={ task.key }
+					data-key={ task.key }
```
Jinksi, line 172: Allows targeting `data-key="dispute-resolution-task"` as a CSS selector

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+		if ( disputeCount === 1 ) {
+			// Redirect to the dispute details page if there is only one dispute.
+			const disputeId = activeDisputes[ 0 ].dispute_id;
+			history.push(
+				getAdminUrl( {
+					page: 'wc-admin',
+					path: '/payments/disputes/details',
+					id: disputeId,
+				} )
+			);
```
Jinksi, line 85: Using `history.push()` client-side navigation for better UX

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+	const disputeCount = activeDisputes.length;
+
+	if ( disputeCount === 0 ) {
+		return null;
+	}
+
+	const handleClick = () => {
+		wcpayTracks.recordEvent( 'wcpay_overview_task', {
+			task: 'dispute-resolution-task',
+		} );
```
Jinksi: Tracks considerations:
* Should we change this name to an action:  e.g. `wcpay_overview_task_click`?
* Should we include additional props: e.g. `activeDisputeCount`?

Jinksi:
I have addressed the considerations raised so far and updated the PR description. Ready for re-review!

client/overview/index.js
```diff
@@ -59,16 +59,20 @@ const OverviewPage = () => {
 		wpcomReconnectUrl,
 		featureFlags: { accountOverviewTaskList },
 	} = wcpaySettings;
-	const numDisputesNeedingResponse =
-		parseInt( wcpaySettings.numDisputesNeedingResponse, 10 ) || 0;
+
 	const { isLoading: settingsIsLoading, settings } = useSettings();
 
+	const { disputes: activeDisputes } = useDisputes( {
+		filter: 'awaiting_response',
```
shendy-a8c, line 66: By default, we're fetching the disputes sorted by `created`. I think it's better if it's sorted by `due_by` since we're showing disputes by when they're due. What do you think?

Sorting by `created` is not bad though because very likely it aligns with `due_by`, ie the older the dispute was created, the sooner it's due.

I'm tempted to suggest adding a filter [here](https://github.com/Automattic/woocommerce-payments/blob/a0a1f9bd0f103b504508831d6994d347c148e06f/includes/core/server/request/class-list-disputes.php#L44-L53) so we can query disputes whose `due_by` within <= 7d but that's overengineering IMO.

```suggestion
		filter: 'awaiting_response',
		orderby: 'due_by',
```

changelog/add-overview-active-disputes-notice
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Improve the wording and style of the "Active Disputes" task list item on the Payments → Overview screen to better communicate the urgency of resolving these disputes.
```
shendy-a8c, line 4: Since we're now only showing disputes due within 7d, should we hint that in the changelog?

It's too crowded to add it in this changelog entry, so I'm suggesting to add a new changelog entry.

Feel free to come up with better wording but I'm thinking:
```
Show only disputes waiting for response that are due within 7 days on the Payments → Overview screen.
```

client/overview/index.js
shendy-a8c: I take back my suggestion because I see that the `due_by` field can be null and not indexed! I just checked the db migration in the server.

changelog/add-overview-active-disputes-notice
Jinksi: Thanks! Good idea, I've added another changelog entry in 3209294:

```
Display the "Active Disputes" task list item on the Payments → Overview screen only if there are disputes due within seven days.
```

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Returns true if the dispute is due within the specified number of days.
+ *
+ * @param {CachedDispute} dispute - The dispute to check.
+ * @param {number} [days] - The number of days to check. Defaults to 1.
+ *
+ * @return {boolean} True if the dispute is due within the specified number of days.
+ */
+const isDueWithin = ( dispute: CachedDispute, days = 1 ) => {
```
shendy-a8c: `nit`: I don't see the reason we need the 2nd argument, `days`, to have default value here. All usages pass the number of days.

```suggestion
const isDueWithin = ( dispute: CachedDispute, days ) => {
```

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+import type { CachedDispute } from 'wcpay/types/disputes';
+import { formatCurrency } from 'wcpay/utils/currency';
+import { getAdminUrl } from 'wcpay/utils';
+import wcpayTracks from 'wcpay/tracks';
+
+/**
+ * Returns true if the dispute is due within the specified number of days.
+ *
+ * @param {CachedDispute} dispute - The dispute to check.
+ * @param {number} [days] - The number of days to check. Defaults to 1.
```
shendy-a8c: If you agree that the 2nd argument does not need to be optional...

```suggestion
 * @param {number} days - The number of days to check.
```

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * @return {boolean} True if the dispute is due within the specified number of days.
+ */
+const isDueWithin = ( dispute: CachedDispute, days = 1 ) => {
+	const now = moment();
+	const dueBy = moment( dispute.due_by );
```
shendy-a8c, line 32: Just by the db table definition, `due_by` can be null (not sure what case that allows a dispute having no due though). Maybe safer to handle null / falsy value here. Since `CachedDispute.due_by`'s type is a string, we need to handle when its value is an empty string.

I see that when `due_by` is an empty string, `dueBy.diff( now, 'hours' )` is `NaN` so the return value will be `false`, which I expect what `isDueWithin()` should return, so maybe nothing needs to be done here.

client/overview/task-list/dispute-tasks.ts
shendy-a8c: Yes to all of your questions.

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+					path: '/payments/disputes',
+					filter: 'awaiting_response',
+				} )
+			);
+		}
+	};
+
+	// Create a new array and sort by `due_by` ascending.
+	activeDisputes = [ ...activeDisputes ].sort( ( a, b ) =>
+		moment( a.due_by ).diff( moment( b.due_by ) )
```
shendy-a8c: When `due_by` is empty string, `moment.diff()` will return `NaN` which might give a wrong return value.

For example:
```
moment( '' ).diff( moment( '2023-06-14' ) )
```
will return `NaN` and per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters
> NaN is treated as 0

so `a` will be considered equal to `b`, which is wrong.

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,221 @@
...
+		content: '', // Subtitle text defined below.
+		level: 1,
+		completed: false,
+		expanded: true,
+		isDismissable: false,
+		showActionButton: true,
+		actionLabel: __( 'Respond now', 'woocommerce-payments' ),
+		action: handleClick,
+		onClick: () => {
+			// Only handle clicks on the action button.
```
shendy-a8c, line 120: Do you think we need to handle CSS on hover? Currently, the background becomes grey as if the whole task is clickable. IMO it's OK to leave it as is though.

client/overview/style.scss
shendy-a8c: I don't know how to reproduce this. Do you have screenshots?

shendy-a8c (changes requested):
Very quality work, @Jinksi!

I have some minor feedbacks.

shendy-a8c:
The tooltip is daunting and untrue. Without responding to disputes, the store can still run =)

Screenshot 2023-06-14 at 13 39 27

What do you think of hiding that with CSS. I see the tooltip is injected within the task item component, so we can target it precisely only for dispute task item.

client/overview/index.js
```diff
@@ -59,16 +59,20 @@ const OverviewPage = () => {
 		wpcomReconnectUrl,
 		featureFlags: { accountOverviewTaskList },
 	} = wcpaySettings;
-	const numDisputesNeedingResponse =
-		parseInt( wcpaySettings.numDisputesNeedingResponse, 10 ) || 0;
```
shendy-a8c, line 63: Now that `numDisputesNeedingResponse` is not used, we should remove it completely. I see references:

https://github.com/Automattic/woocommerce-payments/blob/ff27f046cfa50032d1ce47369f82a37680f17a8e/includes/admin/class-wc-payments-admin.php#L783

https://github.com/Automattic/woocommerce-payments/blob/ff27f046cfa50032d1ce47369f82a37680f17a8e/client/globals.d.ts#L72

Jinksi: Good idea, thanks! I've removed this unused var from `wcpaySettings` in 082d69bbdcfd32bd024b6881dfce682d3bc4af7c

client/overview/task-list/dispute-tasks.ts
Jinksi: Good catch! Making this a required arg reduces complexity 👍

Updated in a7ef1e5

client/overview/task-list/dispute-tasks.ts
Jinksi: 🙏 Updated in 0f53fa85f3e4148eb31ede53278d91cea3b229e5

Jinksi: 🙏 Updated in 0f53fa85f3e4148eb31ede53278d91cea3b229e5

client/overview/style.scss
Jinksi: This can be reproduced by enabled other tasks in the list: force this by returning `true` for `showUpdateDetailsTask` and `wpcomReconnectUrl` in [WC_Payments_Admin::get_js_settings()](https://github.com/Automattic/woocommerce-payments/blob/5a3909d4823d3794f53f73fed5d468ca905ca656/includes/admin/class-wc-payments-admin.php#L763-L764).

**With `position: initial` (before this PR)**
image

**With `position: relative` (after this PR)**

image

client/overview/task-list/dispute-tasks.ts
Jinksi: 👍 I think it's good to explicitly handle this case. I've handled the case of an empty string `due_by` value in 35c573e.

```ts
if ( dispute.due_by === '' ) {
	return false;
}
```

client/overview/task-list/dispute-tasks.ts
Jinksi: Nice catch! I've filtered out empty string `due_by` values at the top of the `getDisputeResolutionTask()` function in ad001c7:

```ts
activeDisputes = [ ...activeDisputes ]
	.filter( ( dispute ) => dispute.due_by !== '' )
	.sort( ( a, b ) => moment( a.due_by ).diff( moment( b.due_by ) ) );
```

client/overview/task-list/dispute-tasks.ts
Jinksi: It is _sort of_ consistent with the other task list items:
* The collapsible tasks toggles collapsing on click
* The completed task does nothing on click, is still grey on hover
* The new dispute task does nothing on click, is still grey on hover

https://github.com/Automattic/woocommerce-payments/assets/3147296/7cb46993-74d7-41af-9005-294a379c1d5b

Jinksi:
> The tooltip is daunting and untrue. Without responding to disputes, the store can still run =)

Good point, @shendy-a8c. I've hidden this with CSS in 6ffcfc9.

Thanks for the thorough review, I think I've resolved all of your suggestions now, ready for re-review 🚀

client/overview/task-list/dispute-tasks.ts
shendy-a8c: Good point to be consistent with other tasks.

client/overview/style.scss
shendy-a8c: Thanks for explaining that.

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,227 @@
...
+		.sort( ( a, b ) => moment( a.due_by ).diff( moment( b.due_by ) ) );
+
+	const activeDisputeCount = activeDisputes.length;
+
+	if ( activeDisputeCount === 0 ) {
+		return null;
+	}
+
+	const handleClick = () => {
+		wcpayTracks.recordEvent( wcpayTracks.events.OVERVIEW_TASK_CLICK, {
```
shendy-a8c, line 71: Just a reminder that this new event needs to be registered.

client/overview/task-list/dispute-tasks.ts
```diff
@@ -0,0 +1,227 @@
...
+			task: 'dispute-resolution-task',
+			activeDisputeCount,
```
shendy-a8c: When I grep `eventprop_name` in `tracks-events-registration` repo, looks like the convention for prop name is all lowercase with underscore. I don't think it's a big deal but for consistency sake, maybe we should use `active_dispute_count` here.

```suggestion
			active_dispute_count: activeDisputeCount,
```

shendy-a8c (approved):
This looks good and ready to merge!

I have one last thing about tracking. Feel free to accept my suggestion or not.

client/overview/task-list/dispute-tasks.ts
shendy-a8c: I just saw that you have mentioned this in https://github.com/Automattic/woocommerce-payments/issues/6353.

client/overview/task-list/dispute-tasks.ts
Jinksi: Yes! Thanks, the FG mentions lowercase + underscores are required for event props.

client/overview/task-list/dispute-tasks.ts
Jinksi: Thanks! I've created #6541 to ensure we don't miss this step.

client/overview/task-list/dispute-tasks.ts
Jinksi: I've committed your suggestion in a41fb14

PR #6390, opened 2023-05-26
Implement the deferred intent creation UPE for the My account subpages

### LINKED ISSUE #5023: Enhance My Account section to render and fully function payment gateways on subpages

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/5023

This PR introduces the functionality for _Add payment method_, _Pay for order_ and _Change payment method for subscription_ pages which are all to be found under _My account_ page. The implementation mostly re-uses the existing mechanism of initializing and mounting Stripe payment elements without creating the intent, which was previously introduced in `createStripePaymentElement` function in `stripe-checkout.js` script. `stripe-checkout.js` script was re-named and is called `payment-processing.js` now since it has an implementation also for pages other than checkout.

For adding a new payment method, the ajax request was used to create and confirm a setup intent (see `createAndConfirmSetupIntent` in `payment-processing.js`), which returns the confirmed intent information to the client so that any next actions, such as 3DS confirmation, can be successfully handled. It seems to be the only possible way to accomplish our goals. 

If it goes to paying for an order and changing the payment method for a subscription, both areas are pretty straightforward on the client side. Ultimately, they attach the created payment method id to the hidden input field and submit the form. The whole processing is handled by the legacy WCPay gateway, similar to how the shortcode checkout was implemented in https://github.com/Automattic/woocommerce-payments/pull/5869.

Additionally, certain refinements were made to the `class-wc-payments.php` file, which is responsible for registering the payment gateways. Specifically, when modifying the subscription payment method, it was necessary to restrict the available gateways to only the card gateway. To achieve this, a check was implemented to verify if the user is changing the payment method for a subscription. In such cases, the registration process skips all non-card gateways.

#### _Payment methods -> Add payment method_
- Add a regular card
- Add a 3DSv2 card (e.g., `4000000000003220`)
  - Authenticate and confirm that the behavior is correct
  - Fail and confirm that the user sees the error notice* as expected and in line with the split UPE/legacy card
  - Fail and after seeing the error notice, click Add payment method once again and confirm that the 3DS modal pop-up appears correctly. Authenticate the card and confirm that the flow ended up successfully and the behavior is expected.
- Add a 3DSv1 card (e.g., `4000002760003184`)
  - Authenticate and confirm that the behavior is correct
  - Fail and confirm that the user sees the error notice* as expected and in line with the split UPE/legacy card
  - Fail and after seeing the error notice, click Add payment method once again and confirm that the 3DS modal pop-up appears correctly. Authenticate the card and confirm that the flow ended up successfully and the behavior is expected.
- Add a declined card (e.g., `4000000000000002`) and confirm the error notice is displayed as expected
- Provide the expiry year from the past
  - Confirm that the error appears right under the input field
  - Click Add payment method and confirm that the notice with the same text is displayed
- Perform the same steps but provide an incomplete (too short) card number and confirm the behavior is as expected
- After all the above tests, you should have a few cards saved on the merchant site. Check the My account -> Payment methods list and confirm that all the saved cards are displayed.
- Delete one card and ensure the deletion process runs smoothly and finishes successfully.
- Make any saved card of your choice a default one by clicking the Make default button. Go to the checkout page and ensure this card is chosen by default.
- Add a product to the cart, go to the checkout page, and pay with one of the previously saved cards.

> *Error notice = We are unable to authenticate your payment method. Please choose a different payment method and try again.

#### _Pay for order_
1. Go to the shortcode checkout, pay e.g. with giropay and fail the transaction.
2. Go to My account -> Orders, find the last order and click the Pay button.
3. Ensure all the enabled payment gateways are successfully loaded.
4. At this point, go to the Stripe dashboard and confirm that there are no payment intents created.
5. Pay with a default card:
   - Confirm that the order was paid successfully.
   - Check that the status is Processing in WooCommerce.
   - Verify that the payment status is Succeeded on the payments list in Stripe.
6. Repeat steps 1 and 2, but on the 5th step, pay with another payment method (e.g., EPS) and confirm that the payment ended up with success.
7. Repeat steps 1 and 2, but on the 5th step, pay with a saved payment method.

#### _Subscription payment method change_
1. Purchase a subscription.
2. Go to My account -> Subscriptions and find the subscription you just purchased. Click View.
3. On the opened details page, click Change payment.
4. Use a new card payment method and insert all the details of the default card.
5. Click Change payment method and confirm that you are successfully redirected back to the subscription’s details page.
6. Confirm that on the details page, in the Payment row, you see the details of the card you just added (e.g., Visa ending in 4242, expires 05/65).
7. Go to the admin panel, WooCommerce -> Subscriptions.
8. Open subscription details and on the right side, choose Process renewal.
9. After the renewal is processed, confirm that it's successful on the Stripe dashboard page.

1. Check [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and ensure that all the three above use-cases are reflected in the flows
2. Check [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows) and ensure all the three above use-cases are reflected in the instructions

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6390` or branch name `feature/deferred-upe-my-account-page` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=feature/deferred-upe-my-account-page) :rocket:

---

Build info: 

- Latest commit: 5143193e56fe86c24a9bfaecccdb9353e1f2225d
- Build time: 2023-06-09 11:13:00 UTC

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -172,13 +184,66 @@ export async function mountStripePaymentElement( api, domElement ) {
...
 	const upeElement =
 		gatewayUPEComponents[ paymentMethodType ].upeElement ||
 		( await createStripePaymentElement( api, paymentMethodType ) );
 	upeElement.mount( domElement );
 }
 
+export const createAndConfirmSetupIntent = ( $form, { id }, api ) => {
+	return api.setupIntent( id ).then( function ( confirmedSetupIntent ) {
+		appendSetupIntentToForm( $form, confirmedSetupIntent );
+		return confirmedSetupIntent;
```
FangedParakeet:
```suggestion
```
It's not really a huge deal, but this return value isn't used anywhere, no?

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -172,13 +184,66 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+	} );
+};
+
+/**
+ * Saves the payment method ID in a hidden input, and re-submits the form.
+ *
+ * @param {Object} $form         The jQuery object for the form.
+ * @param {Object} paymentMethod Payment method object.
+ */
+export const handleOrderPayment = ( $form, { id } ) => {
```
FangedParakeet:
```suggestion

 * @param {Object} paymentMethod Payment method object.
 */
export const handleOrderPayment = ( { id } ) => {
```
First I thought this `$form` variable was unused and the parameter could be removed, but I see now that to follow [the callback signature for the `additionalActionsHandler`](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/client/checkout/classic/upe-deferred-intent-creation/payment-processing.js#L287-L291), we need it after all...

...However, if you swapped the order of the parameters, so that the `paymentMethod` is used first, `$form` second, and `api` last, you could remove this unused parameter after all. `paymentMethod` is the only parameter required by both callback functions--`handleOrderPayment` & `createAndConfirmSetupIntent`--whereas the other parameters are technically not required, since they are only required by the latter. 

It's a minor complaint, but I think it's marginally cleaner, so let me know what you think!

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -206,16 +276,23 @@ export const checkout = ( api, jQueryForm, paymentMethodType ) => {
...
 			appendFingerprintInputToForm( jQueryForm, fingerprint );
 			appendPaymentMethodIdToForm(
 				jQueryForm,
 				paymentMethodObject.paymentMethod.id
 			);
+			await additionalActionsHandler(
+				jQueryForm,
+				paymentMethodObject.paymentMethod,
+				api
+			);
```
FangedParakeet:
```suggestion
			await additionalActionsHandler(
				paymentMethodObject.paymentMethod,
				jQueryForm,
				api
			);
```
I suggested swapping these two parameters in another comment, since only the payment method is required by all callback functions that can apply here. Just spelling it out here as a complementary comment, just in case my previous explanation was confusing beyond clarity. 😅

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -155,6 +157,16 @@ async function createStripePaymentElement( api, paymentMethodType ) {
 	return createdStripePaymentElement;
 }
 
+function appendSetupIntentToForm( form, confirmedIntent ) {
```
FangedParakeet, line 166: Again not really a huge deal, but there are a few function in here missing doc comments...not that I'm having difficulty understanding the code nor that the linter seems bothered at all, but just thought it's worth a mention. 🤷

client/checkout/classic/upe-deferred-intent-creation/test/payment-processing.test.js
```diff
@@ -342,3 +426,52 @@ describe( 'Checkout', () => {
 		document.body.appendChild( stateInput );
 	}
 } );
+
+describe( 'Setup intent creation and confirmation', () => {
+	test( 'Setup intent is creation and confirmation', async () => {
```
FangedParakeet:
```suggestion
	test( 'Setup intent is created and confirmed', async () => {
```
Grammar police. ⚠️

includes/class-wc-payments.php
```diff
@@ -668,7 +668,10 @@ public static function register_gateway( $gateways ) {
 			} else {
 				self::$registered_card_gateway = self::$card_gateway;
 			}
-			$gateways[]       = self::$registered_card_gateway;
+			$gateways[] = self::$registered_card_gateway;
+			if ( self::$registered_card_gateway->is_subscriptions_enabled() && self::$registered_card_gateway->is_changing_payment_method_for_subscription() ) {
+				return $gateways;
+			}
```
```suggestion
			$gateways[] = self::$registered_card_gateway;
```
Correct me if I'm mistaken, but is the idea here to only return the card gateway when processing some kind of subscription payment? I think the safer play would be to add this logic down [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/class-wc-payments.php#L692). This seems like a more appropriate place for it and means that we won't need to revisit this block of code when we enable SEPA as a reusable payment method that can be used with subscriptions. What say?

FangedParakeet (changes requested):
Top notch work as always! I've left a few small comments, but most are pretty negligible. 

I tested all the pay for order, saved payment, and subscription flows and was able to proceed unfettered through the transactions with glee in all cases. Hurrah!

I did notice one small UI issue on the Pay for Order page. When I was checking out for a regular product, I noticed that for some reason the terms that normally appear when saving a payment method were present underneath my payment information, even though I hadn't elected to save any payment information at this time.

Saved payment terms appearing when not required
_Saved payment terms appearing without an invite_

I was able to get rid of them by simply checking and unchecking the "Save payment information" checkbox, so I'm pretty sure they're not supposed to be there. Can you have a quick look into this one? I think we may need to ensure `renderTerms` is fired manually or initialised in some way.

I didn't really debug this too deeply, because while I was double-checking it, I just noticed a new issue. Sorry! 😁 

No checkbox to save payment method
_Checkbox to save payment information absent without notice_

For some reason, this time the terms appeared, but the checkbox to save my payment information did not, so I couldn't even save my payment method if I wanted to! I'm just checking out with a regular product in the screenshot above, so not sure what happened. Do we maybe need to add some logic for the tokenization display over [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/class-wc-payments-upe-checkout.php#L242)? 🤔 

I'm honestly not sure what I did here, but let me know if you are able to replicate any of the above or have any suggestions for what heretical blasphemy I'm obliviously committing instead. Cheers!

includes/class-wc-payments.php
timur27: Thanks for the catch - this totally makes sense. I've updated the code in cfa0f6c06b5eb684d7d0adb0ac42366c5d64ad0d.

timur27:
Thanks for the review, @FangedParakeet! I've addressed all the open threads and appreciate your input. Regarding the issues found and described in https://github.com/Automattic/woocommerce-payments/pull/6390#pullrequestreview-1454397731 - I've fixed the first issue, which was essentially a problem of deferred UPE implementation specifically. 

I think it makes sense to address the second issue as part of another issue mainly because the problem is valid for any UPE setup and the legacy card (non-UPE), which means we need to add a robust and UPE-agnostic fix. Please let me know your feelings on this, and I'm happy to discuss further. 

In the meantime, I'm sending a kind request to review the code after my latest updates. 👍 
To test the first issue described in https://github.com/Automattic/woocommerce-payments/pull/6390#pullrequestreview-1454397731: 

0. Confirm that HEAD refers to 5b8178fbcf0cb3c8b7fdc08782e6a223d44c6a86
1. Ensure you have two orders that can be paid on the My account page
2. Go to both of them, Use a new payment method and confirm that the terms are not displayed unless you check the Save payment information checkbox. In other words, confirm that things behave as expected at this stage.
3. Go to one of these two orders, pay for it with a default card, and perform a usual procedure without saving it.
4. After succeeding, go to another failed order and confirm that after choosing to Use a new payment method, terms are displayed even if you don’t check the Save payment information checkbox.
5. Pull the latest changes from `feature/deferred-upe-my-account-page`
6. Perform steps 1-4 and confirm, that on the 4th step terms are not appearing unexpectedly, unless you click the _Save payment information_ checkbox

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -66,4 +68,46 @@ jQuery( function ( $ ) {
...
+			$( '.wcpay-upe-element' ).length &&
+			! $( '.wcpay-upe-element' ).children().length &&
+			getUPEConfig( 'isUPEEnabled' )
+		) {
+			$( '.wcpay-upe-element' )
+				.toArray()
+				.forEach( ( domElement ) =>
+					mountStripePaymentElement( api, domElement )
+				);
+		}
```
FangedParakeet: This is almost exactly duplicated from up [here](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js#L38-L47): possible to extract it and reuse in its own function--`maybeMountStripePaymentElement()` or something like that?

The only difference between these two blocks of code is that we check the UPE config for the value of `isUPEEnabled` before continuing in the latter case. Is that a crucial difference that requires us to duplicate the rest of this code? 🤔 

Not a huge deal either way--just pondering on it!

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -172,13 +190,73 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+ * Saves the payment method ID in a hidden input, and re-submits the form.
+ *
+ * @param {Object} id Payment method object ID.
+ */
+export const handleOrderPayment = ( { id } ) => {
+	const paymentSelector = '#wcpay-payment-method';
+
+	// Populate form with the payment method.
+	document.querySelector( paymentSelector ).value = id;
+};
```
FangedParakeet: Just wondering to myself, as I do--is this step required? 🤔 

In [the step prior to this](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/client/checkout/classic/upe-deferred-intent-creation/payment-processing.js#L296-L299) in the `processPayment` function, we append the hidden payment method input to the page and [set its value to the payment method ID](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/client/checkout/utils/upe.js#L198). Then in this step, we repopulate the same input with the same payment method ID. 

Am I missing something here or is this a little redundant?

includes/class-wc-payments.php
```diff
@@ -686,7 +686,7 @@ public static function register_gateway( $gateways ) {
 
 			}
 
-			if ( is_add_payment_method_page() ) {
+			if ( is_add_payment_method_page() || ( self::$registered_card_gateway->is_subscriptions_enabled() && self::$registered_card_gateway->is_changing_payment_method_for_subscription() ) ) {
```
FangedParakeet: Apologies, but I've just had another thought (sorry lol). 

I was wondering how we filtered the payment gateways when simply checking out for a subscription product (i.e. not on the add payment method page nor changing payment method for a subscription) and it's basically because we check whether the potential payment method `is_enabled_at_checkout()` over [here](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/includes/payment-methods/class-upe-payment-gateway.php#L935). 

If we take a peek inside that function, we can see that [if there is currently a subscription item in the cart, we check whether the payment method is reusable](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/includes/payment-methods/class-upe-payment-method.php#L113-L118) and otherwise the payment method is included. If we simply moved this additional logic, highlighted here inside the `register_gateway()` function, into the `UPE_Payment_Method::is_enabled_at_checkout()` function instead, then we wouldn't need to handle any additional logic with the `$reusable_methods` array in this function, since these payment methods are already filtered. 

FWIW, `is_subscriptions_enabled()` and `is_changing_payment_method_for_subscription()` are both available to the `UPE_Payment_Method` class, thanks to [the inclusion of the `WC_Payments_Subscriptions_Utilities` trait](https://github.com/Automattic/woocommerce-payments/blob/4bbebe49e4747a200d65fccded7bcc865f145e25/includes/payment-methods/class-upe-payment-method.php#LL24C6-L24C41), so I don't think there would be any issues with asking these questions within this class instead of relying on the gateway. 

Does that make sense? Does this sound like a good idea? Am I just a rotten trouble-maker? Please let me know, I'm all ears. 😁

FangedParakeet (commented):
Retested and everything worked this time! 🎉 

Just have a few more small comments though: please let me know your thoughts and then we can look to wrap this one up (and I'll promise to respond to your feedback a lot quicker this time). Cheers!

includes/class-wc-payments.php
timur27: Thanks, I think it's a good point and worth moving to the `UPE_Payment_Method` - eventually the check ends up looking easier and in the right place. The main thing that concerned me with the previous solution was that `register_gateway` function seemed to be too high-level place for such filtering logic so I'm more than happy to change this.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -66,4 +58,50 @@ jQuery( function ( $ ) {
...
+
+	$( 'form#order_review' ).on( 'submit', function () {
+		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
+		if ( ! isUsingSavedPaymentMethod( paymentMethodType ) ) {
+			return processPayment(
+				api,
+				$( 'form#order_review' ),
+				paymentMethodType
+			);
+		}
```
FangedParakeet: This is now almost identical to this anonymous function up [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js#L41-L44) on the regular checkout events. We could extract this into its own function as well...but I'll leave it up to you. The logic is compact enough and it does make sense to keep these functions separated, so I don't mind if things stay as they are or not. Either way, I will cease creating complaints. 😁

FangedParakeet (commented):
@timur27, can you quickly double-check something for me, please? 

I'm finding that my Order Pay page is broken when non-card payment methods are present on the page.  I can replicate this with the following steps...

1. Enable all non-card LPMs--particularly giropay, EPS, or P24, for which Stripe does not support recurring payments.
2. Create a new order with currency compatible with above payment methods (e.g. EUR). 
3. Ensure that your cart is currently empty.
4. Proceed to Pay for Order page for above order and select one of the LPMs. 
5. Notice that payment element does not load for selected LPM.

EPS payment element fails to load
_EPS is kil_

Here's the error I see in the JS console that producees the above failure. 

JS console error
_Invalid `payment_method_types` provided_

Following the error to Stripe's logs (see [here](https://dashboard.stripe.com/acct_1MvrvGCA84dwgkbI/test/logs/req_rQfDHi77DgLLlM?t=1686260565), if the link works for the above error) I noticed that for some reason `setup_future_usage` is set to `off_session` and the mode of the intent is `setup`, as if we are trying to set up a payment method for recurring usage. 

Request query params from Stripe logs
_"mode": "setup"_

I _think_ this is happening, because the `mode` parameter is set in the element options [here](https://cdn-std.droplr.net/files/acc_1104206/3cbk4O), where if the `cartTotal` amount is 0, a setup intent will be requested. However, this `cartTotal` parameter that we pass to the client [is calculated using the WC cart total](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/class-wc-payments-upe-checkout.php#L116-L117), even though the cart object is not present on the Pay for Order page. 

I think we need to overwrite the `cartTotal`--and `currency`--value down [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/class-wc-payments-upe-checkout.php#L132-L164), if we're aware that we are on the Pay for Order page, to use the order total and order currency, so that we send the correct values to the client and request the correct intent type.

Let me know if you're able to replicate this or I'm doing something wrong. If the former, also let me know if my explanation sounds like it might make sense! Cheers.

timur27:
@FangedParakeet Many thanks for pointing out an issue with 3 non-recurring not loading on the pay for order page. I didn't experience this since, as it turned out, I always had something in my cart. To sum this up: 

* The issue described [here](https://github.com/Automattic/woocommerce-payments/pull/6390#pullrequestreview-1470776934) is reproducible
* It has been fixed in 7b639ccdadf3a0eb92ff8b6a2be84c49bcffaf6f

I've also addressed [this suggestion](https://github.com/Automattic/woocommerce-payments/pull/6390#discussion_r1223557728) in 6ac115393308c53633e47935fbc0fbbde9721ec0

FangedParakeet (approved):
Thanks for responding to my barrage of feedback! Every blemish I noticed before has been comprehensively wiped clean. 

I have identified two open issues that I'll describe forthwith--but they they are UPE-agnostic, so I think we can handle them in separate issues. They also both involve the Pay for Order page, which is both the bane of many a developer's existences and also slightly lower priority due to its limited usage.

### Pay for Order with Subscription product

1. Enable additional LPMs that do not support recurring payments (e.g. giropay, P24, EPS, etc.).
2. Create a new order ensuring that currency will support above payment methods (e.g. EUR).
3. Add a subscription product to the aforementioned order.
4. Proceed to the Pay for Order page for this order.
5. LPMs that cannot be used to complete the order (e.g. giropay, P24, EPS, etc.) will appear as payment methods on the order.
6. Oh no.

This is because [we filter payment methods when there is a subscription product in the cart](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/payment-methods/class-upe-payment-method.php#L114), but on the Pay for Order page there is no cart object and we need to check the order instead. 

### Pay for Order with Polluted Cart

1. Enable additional LPMs that do not support recurring payments (e.g. giropay, P24, EPS, etc.).
2. Create a new order ensuring that currency will support above payment methods (e.g. EUR).
3. Add a regular product to this order.
4. As a customer add a subscription product to your regular cart.
5. Now proceed to the Pay for Order page for the order created in the earlier step.
4. The LPMs that do not support recurring payments will now not be available to use, even though the product in the order will not incur recurring costs. 
5. Furthermore underneath the CC payment method, the save payment information checkbox will be automatically selected and hidden, so that it cannot be unselected. You can confirm this by checking the page source and also regard that the saved payment method terms will be present beneath the payment method.
6. Oh no.

The LPMs do not appear for the same reason as the previous issue, since we are filtering the methods [when there is a subscription item in the cart](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/payment-methods/class-upe-payment-method.php#L114) rather than in the relevant order.

The save payment information checkbox is automatically and immutably selected, because we do so when there is a subscription product present in the cart (see [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/class-wc-payments-upe-checkout.php#L313) -> [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/compat/subscriptions/trait-wc-payment-gateway-wcpay-subscriptions.php#L161) -> [here](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-my-account-page/includes/compat/subscriptions/trait-wc-payments-subscriptions-utilities.php#L72)), when we should again instead be checking the relevant order.

However, if you agree and are willing to do so, I am ready to leave these resolutions for a new issue, as they affect all types of the UPE, so aren't specifically related to the narrower goals of this issue. Also, I believe I have tried and tested your patience enough as it is in this PR. 😁 

Anyways, I think this PR achieves the ambitious mission it set out to achieve, as well as tightening a few loose nuts and bolts along the way. LGTM. :shipit:

timur27:
@FangedParakeet Thanks for the review and tests, Samir!

> they are UPE-agnostic, so I think we can handle them in separate issues

I agree with this - let's handle them separately since the issue is UPE agnostic. It seems to me that both problems can be addressed within a single issue since the root cause is the same for both at first glance.

PR #6408, opened 2023-05-31
Remove the redundant network requests in `capture_charge`

### LINKED ISSUE #6358: IPP: Speed up capture terminal payment endpoint
This is a follow-up to the spike #6239

After running a few tests on the `capture_terminal_payment` function, we've seen that it can be improved by passing information to the `capture_charge` function instead of resending a new HTTP function to Stripe.

This issue is to keep track of this useful change and to implement it in the future.

mordeth:
Fixes #6358

By removing the redundant network requests, we aim to make capturing an intent faster.

In `capture_charge`:

* Remove the Get Intent API request, which was earlier used to fetch intent metadata from Stripe. This method now receives metadata from the calling function.
* Remove the Update Intent API request, which was earlier used to update an intent's metadata with metadata from both the order and itself. We now send the merged metadata to Stripe in the Capture Intent request.

#### Testing instructions

Prerequisite: Server#3937

Do this first with `trunk` and then with this branch. Compare the response time in both case: it should be less with this branch.

> [!NOTE]
> On @anu-rock's MacBook Pro 16" with M2 Pro: Before: 9 secs, After: 6 secs.

* Create a new order in WooCommerce frontend with "Cash on delivery" as the payment method.

* Create a payment intent in Stripe Dashboard. Go to: Your test account → Payments → Create payment → Manual payment. Select the "capture funds later" option.

* On the intent details page, scroll down to the "Metadata" section. Create a new key `order_id` with value from the order created in the previous step. Create another key `reader_ID` with value "wisepad".

* Make a POST request to `/wp-json/wc/v3/payments/orders/<order_id>/capture_terminal_payment`.

* Verify that the payment succeeds.

* Also verify on the intent details page that the metadata contains `reader_ID` along with order-related key-value pairs.

##### BONUS: Testing with mobile app

Use this guide (PdfdoF-D-p2) to test with the Woo mobile app and verify that:
1. IPP payment goes through successfully.
2. The payment intent's metadata contains data from both order and app (screenshot 👇).

Screenshot 2023-09-27 at 4 14 10 PM

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6408` or branch name `fix/optimise-capture-charge-calls` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/optimise-capture-charge-calls) :rocket:

---

Build info: 

- Latest commit: 77ca53d31488846a2577efbe8b9393f86c39e498
- Build time: 2023-09-28 02:44:06 UTC

includes/core/server/request/class-capture-intention.php
```diff
@@ -76,6 +76,15 @@ public function set_level3( $level3 ) {
 		$this->set_param( 'level3', $this->fix_level3_data( $level3 ) );
 	}
 
+	/**
+	 * Setter for intent metadata.
+	 *
+	 * @param array $metadata Intent metadata that includes stuff like order details, card reader specifics, etc..
+	 */
+	public function set_metadata( $metadata ) {
```
dpaun1985: we can add typehint here.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2663,23 +2664,14 @@ public function capture_charge( $order, $include_level3 = true ) {
...
-
-			$metadata_from_intent = $intent->get_metadata(); // mobile app may have set metadata.
-			$metadata_from_order  = $this->get_metadata_from_order( $order, $payment_type );
-			$merged_metadata      = array_merge( (array) $metadata_from_order, (array) $metadata_from_intent ); // prioritize metadata from mobile app.
-
-			$wcpay_request = Update_Intention::create( $intent_id );
-			$wcpay_request->set_metadata( $merged_metadata );
-			$wcpay_request->send( 'wcpay_prepare_intention_for_capture', $order );
+			$intent_id           = $order->get_transaction_id();
+			$payment_type        = $this->is_payment_recurring( $order->get_id() ) ? Payment_Type::RECURRING() : Payment_Type::SINGLE();
```
dpaun1985, line 2678: `Payment_Type::RECURRING() : Payment_Type::SINGLE();` - we should not call the constants as functions. This will generate an error.
`Payment_Type::RECURRING : Payment_Type::SINGLE;`

dpaun1985 (commented):
Looks good overall. Left some comments.

kalessil:
Tracing steps back: IIRC (should be double-checked toward the PR which added it initially), getting intent, which we are dropping here, was introduced to pull original meta-data created by mobile apps via Stripe SDK directly. And extending the metas to WooPayments goodies. So both mobile/WooPayments can live easy lives and extend metas as needed.

This change might work if Strpe merges [meta](https://stripe.com/docs/api/payment_intents/capture#capture_payment_intent-metadata) with old ones. But [the documentation](https://stripe.com/docs/api/payment_intents/capture#capture_payment_intent-metadata) doesn't clarify if this merge/override handling of metadata data happens there. 

So, the questions:
- do we know which of merge/override handling of meta is in the capture intent endpoint on Stripe's end?
- did we test the change with the mobile APP?

anu-rock:
@kalessil 

I'm glad you brought up this point. Based on my testing, metadata merge works the same way as before.

> do we know which of merge/override handling of meta is in the capture intent endpoint on Stripe's end?

Merging of metadata happens in our code rather than Stripe:

https://github.com/Automattic/woocommerce-payments/blob/2fb4c9b662e180749c4f78d798e6ba0b7cf2a220/includes/class-wc-payment-gateway-wcpay.php#L2670

The `metadata` param exposed in Stripe's capture intent API merely allows us to set it to whatever value we'd like. In this case, we set it to our merged dataset.

> did we test the change with the mobile APP?

I haven't tested with the mobile app, but I believe my testing instructions cover this, especially this point:

_"Also verify on the intent details page that the metadata contains reader_ID along with order-related key-value pairs."_

P.S. I have added a prerequisite in testing instructions above for Server#3937.

includes/class-wc-payment-gateway-wcpay.php
anu-rock: Great catch! Fixed in https://github.com/Automattic/woocommerce-payments/pull/6408/commits/bc2f2a254c01d2a2ddcafa41e82e68ea752564c0.

includes/core/server/request/class-capture-intention.php
anu-rock: Good point. Fixed in https://github.com/Automattic/woocommerce-payments/pull/6408/commits/d7a8d3b695659c42ca56c0572b4c894bb318bd29.

kalessil:
@anu-rock: oki-doki. But the order meta-data, including many WooPayments internal fields - I'm not sure if we shall expose those blindly in intent and if the mobile app creates intents with the same set of data.

I'd cross-check an intent meta created by the mobile app vs complimentary order meta-data.

kalessil:
Okay ignore my previous comment - we did use order meta previously, so no big change there.
I have no further remarks/comments - waving out further reviews to fellow folks from Fusion =)

kalessil (approved):

anu-rock:
@kalessil Awesome, thanks 🙌

dpaun1985 (approved):
✅  LGTM

anu-rock:
@kalessil Just an FYI - I tested using the mobile app just to be double sure. All worked as expected (including metadata). I've updated the testing instructions accordingly 🙌

includes/class-wc-payment-gateway-wcpay.php
anu-rock: @dpaun1985 I had to revert this fix in https://github.com/Automattic/woocommerce-payments/pull/6408/commits/19184ee642603293f8bc58b064821df5d18598c5 because it was resulting in a psalm error in `WC_Payment_Gateway_WCPay::get_metadata_from_order` which expects payment type as `Payment_Type` instead of string. I cannot change the argument type in this method as it's being used in other places, like this:

https://github.com/Automattic/woocommerce-payments/blob/19184ee642603293f8bc58b064821df5d18598c5/includes/class-wc-payment-gateway-wcpay.php#L1125

PR #6419, opened 2023-06-01
Fix Pay for Order missing item count meta, wrong fraud box type, missing status codes on json errors, and confirming after error response

### LINKED ISSUE #6278: FRT meta not set on the order for non-upe failed card manual auth capture
### Describe the bug
Fraud meta not set when we expect it to be set.

### To Reproduce

**Payment started message - With manual auth capture (setting checked) - non-UPE**
Add an item to your cart and proceed to checkout, making sure not to do anything to trigger Your Rule.
Check out using card: `4000000000000002`
The transaction will be declined.
Go to WooCommerce > Orders in the admin and there should be a new Failed order, view/edit that order.

**Expected:**
The order should have a Fraud & Risk meta box, and Order notes similar to the one below.
The No action taken message is displayed because fraud filtering never took place.

Screenshot 2023-05-11 at 23 00 30

**Actual:**
The risk meta box shows: Risk filtering through WooCommerce Payments was not found on this order, it may have been created while filtering was not enabled.
Screenshot 2023-05-11 at 23 00 36

### LINKED ISSUE #6416: FRT meta is not set when Pay for Order is being used.
### Describe the bug
The orders which are going to be paid with a payment link (Pay for Order method) are missing the FRT meta values, which will get the rule engine skipped on payment.

### To Reproduce
1. On the settings page, enable advanced mode of the fraud protection, and configure it with having "Order items threshold" minimum item count set as 1.
2. Go to WooCommerce > Orders page, click "Add Order" button.
3. Add one item to the order.
4. Fill in the billing address.
5. Select "WooCommerce Payments" as the payment method.
6. Leave the user as "Guest". Save. 
7. Right click the "Customer payment page" and copy the URL, open it in an incognito window.
8. Pay it. 
 
### Actual behavior
The server won't run the rules because `Can't find the data source to compare: item_count` error will be thrown.

### Expected behavior
The payment should complete with all rules running.

### LINKED ISSUE #6423: Payment update response errors are ignored on UPE checkout
### Describe the bug
When a payment intent update fails, the result is not checked before attempting to confirm, and in all cases, the client sends a confirm request.

https://github.com/Automattic/woocommerce-payments/blob/242e60e334f7becea083de02dfabc3c3a5a2755b/client/checkout/classic/upe-split.js#L439-L462

and 

https://github.com/Automattic/woocommerce-payments/blob/242e60e334f7becea083de02dfabc3c3a5a2755b/client/checkout/classic/upe-split.js#L439-L458

needs to have a check before attempting to confirm. Also the status codes of the failing intent updates are sent as HTTP OK 200. Should be something different, indicating an error.

### To Reproduce
1. Enable order item threshold filter from fraud protection advanced settings, set it as having minimum 2 items.
2. Open dev console, switch to network tab, click "Preserve logs" to keep the logs visible after page changes.
3. Add an item to the cart and sell it with legacy UPE card element, with a valid card.
4. Sell it and see that the page is redirected to the payment received confirmation page, even if the update intention request failed.

### Actual behavior
The checkout is completed.

### Expected behavior
The checkout page will show an error.

tpaksu:
Fixes #6278, #6416, #6423

This PR fixes three issues that seem to be connected with each other:

- Pay for Order page was not sending the order items count meta, causing the "Order items threshold" filter always fail for orders, because the `WC()->cart` object was not available on the Pay for Order. Added a safety net by getting the items count directly from the order when cart isn't available at the moment. 
- On the same page with UPE enabled (separate update and confirm calls), the confirm always took place even if the update intent call returned an error. And this required two fixes:
    - Added a check to see if the update intent message contained an error.
    - Added HTTP status codes to `wp_send_json_error` calls, because they were sending the errors with HTTP 200 status code, which also allowed the confirmation after failing intent update run. Also, non-UPE card saving through My Account needed a fix because the new error code didn't work well with the current error handling on the client side.
- Also, when the fraud checks were passed, but the payment got declined, or failed with another reason, the order page was showing that the fraud checks weren't applied for the order, which was also wrong because before getting the payment result from Stripe, we always run the fraud rules in the update intention call for card payments. Replaced it with the ALLOW type meta box contents, which indicate the fraud rules are passed. (If fraud rules don't pass, the intention won't be able to get to the confirmation phase.)
- Added some local E2E test fixes to make them run locally.

### Testing instructions

* E2E and unit tests should pass.

#### Testing missing item count meta on Pay for Order message
* Checkout `develop`.
* Disable UPE. 
* Activate advanced fraud protection on Payments > Settings > Fraud protection block
* Configure the protection settings, enable the Order items threshold to force customers having at least one item in the cart, save the configuration
* Create an order manually on the WooCommerce > Orders page, targeted to Guest accounts and having one item. You can enter the address automatically by setting your user in the dropdown and removing it to enable Guest checkout back. 
* Don't forget to set the Payment method to "WooCommerce Payments."
* Save the order, and right click the `Customer payment page` to open it in an incognito window.
* Pay for the order with 4242 4242 4242 4242.
* The order should fail with a "Can't process this payment" like error, which is the issue. The items count can't be sent to the server because the cart isn't initialized on that page.
* On this branch, it'll get paid without any problems.

#### Testing orders still getting confirmed after a update failure
* Checkout `develop` again. 
* Enable UPE and do the same test above again, but now set the minimum order items count as 10, we expect the payment getting blocked on this test. 
* See that the page sends you to the order received page without showing any errors about it. 
* Checkout this branch, build client assets
* Test it again and see that it stops and shows the error message now.

#### Testing declined payments setting a wrong fraud meta box on the order details page 
* Disable UPE. 
* Add beanie and proceed to classic checkout. 
* Pay with 4000 0000 0000 0002. The payment should be declined.
* Go to WooCommerce > Orders page, check the last failed order on the top of the table. 
* Click to it and reach order details
* On the right sidebar, there's a "Fraud & Risk" meta box, it's showing "Risk filtering for WooCommerce Payments was not found.." which is wrong, because the risk filters are applied to the order before the decline.
* Checkout this branch and test it again with a new order, it should show "No action taken, the payment for this order passed your risk filtering." 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6419` or branch name `fix/6278-frt-meta-missing-on-manual-capture` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6278-frt-meta-missing-on-manual-capture) :rocket:

---

Build info: 

- Latest commit: 9f0e15d8fbdf61ffd7e1c7bb3daf23f7cd6c94cd
- Build time: 2023-06-21 12:42:46 UTC

tests/e2e/env/setup.sh
```diff
@@ -187,7 +190,7 @@ else
 fi
 
 echo "Installing basic auth plugin for interfacing with the API"
-cli wp plugin install https://github.com/WP-API/Basic-Auth/archive/master.zip --activate
+cli wp plugin install https://github.com/WP-API/Basic-Auth/archive/master.zip --activate --force
```
tpaksu, line 193: This line had problems with installing the _Basic-Auth_ plugin when the folder was already there. Forced an overwrite if it existed.

tests/e2e/env/setup.sh
```diff
@@ -310,7 +313,10 @@ echo "Creating screenshots directory"
 mkdir -p $WCP_ROOT/screenshots
 
 echo "Disabling rate limiter for card declined in E2E tests"
-cli wp option add wcpay_session_rate_limiter_disabled_wcpay_card_declined_registry yes
+cli wp option set wcpay_session_rate_limiter_disabled_wcpay_card_declined_registry yes
```
tpaksu, line 316: `wp option add` had problems when the option was already there, switched it with `wp option set` which doesn't fail when the option already exists.

tests/e2e/env/setup.sh
```diff
@@ -310,7 +313,10 @@ echo "Creating screenshots directory"
...
+
+echo "Removing all coupons ..."
+cli wp db query "DELETE p, m FROM wp_posts p LEFT JOIN wp_postmeta m ON p.ID = m.post_id WHERE p.post_type = 'shop_coupon'"
```
tpaksu, line 319: Ensured that the `shop_coupon` doesn't exist before running the next line. It fails when the coupon already exists. Those changes were required when running `npm run test:e2e-setup` multiple times.

tests/js/jest.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 	setupFilesAfterEnv: [
 		'<rootDir>/node_modules/@wordpress/jest-preset-default/scripts/setup-test-framework.js',
 		'<rootDir>/tests/js/jest-extensions-setup.js',
+		'expect-puppeteer',
```
tpaksu, line 25: This line makes `expect( xxx ).toClick( selector );` to work locally on the E2E tests.

client/checkout/classic/index.js
```diff
@@ -292,7 +292,11 @@ jQuery( function ( $ ) {
 			.catch( function ( error ) {
 				paymentMethodGenerated = null;
 				$form.removeClass( 'processing' ).unblock();
-				showError( error.message );
+				if ( error.responseJSON && ! error.responseJSON.success ) {
```
tpaksu, line 295: This needed a fix because the change on https://github.com/Automattic/woocommerce-payments/pull/6419/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R3071-R3072

tests/e2e/env/setup.sh
deepakpathania: Could we use `shop_coupon get` maybe to check and create if it doesn't exist? Maintaining raw queries could get trickier compared to standard CLI commands.

tests/e2e/env/setup.sh
```diff
@@ -234,17 +237,17 @@ fi
 echo "Setting up WooCommerce Payments..."
 if [[ "0" == "$(cli wp option list --search=woocommerce_woocommerce_payments_settings --format=count)" ]]; then
 	echo "Creating WooCommerce Payments settings"
-	cli wp option add woocommerce_woocommerce_payments_settings --format=json '{"enabled":"yes"}'
+	cli wp option set woocommerce_woocommerce_payments_settings --format=json '{"enabled":"yes"}'
 else
 	echo "Updating WooCommerce Payments settings"
-	cli wp option update woocommerce_woocommerce_payments_settings --format=json '{"enabled":"yes"}'
+	cli wp option set woocommerce_woocommerce_payments_settings --format=json '{"enabled":"yes"}'
```
deepakpathania, line 243: Just for my understanding, does `update` also fail similar to `add`?

deepakpathania (commented):
Left some comments in the context of E2E setup.

tests/e2e/env/setup.sh
tpaksu: I couldn't be sure about what happens if the meta row doesn't exist, read the source code but couldn't understand it, and didn't want to dig deeper, so I converted it to `set`, which the name suggests that it will both do `add` and `update` jobs work.

tests/e2e/env/setup.sh
tpaksu: `shop_coupon get` returns the DB output directly, which will be hard to parse the ID from to use with `shop_coupon delete`. I've already tried that but this way it'll ensure no coupons exists on the DB.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -62,12 +62,21 @@ public function hash_data_for_fraud_prevention( string $data ): string {
 	 * @return array An array of hashed data for an order.
 	 */
 	public function get_hashed_data_for_customer( $fingerprint ): array {
+		$order_items_count = WC()->cart ? intval( WC()->cart->get_cart_contents_count() ) : null;
+		$order_id          = isset( $_POST['wcpay_order_id'] ) ? intval( $_POST['wcpay_order_id'], 10 ) : 0; // phpcs:ignore WordPress.Security.NonceVerification
```
allie500: Why aren't we doing nonce verification here? Is it not possible in this context? If it isn't possible, is that why you are casting the value of `$_POST['wcpay_order_id']` to a base 10 integer as a kind of sanitization and validation, just in case?

includes/fraud-prevention/class-buyer-fingerprinting-service.php
```diff
@@ -62,12 +62,21 @@ public function hash_data_for_fraud_prevention( string $data ): string {
...
+			}
+		}
+
 		return array_filter(
 			[
 				'fraud_prevention_data_shopper_ip_hash' => $this->hash_data_for_fraud_prevention( WC_Geolocation::get_ip_address() ),
 				'fraud_prevention_data_shopper_ua_hash' => $fingerprint,
 				'fraud_prevention_data_ip_country'      => WC_Geolocation::geolocate_ip( '', true )['country'],
-				'fraud_prevention_data_cart_contents'   => WC()->cart ? intval( WC()->cart->get_cart_contents_count() ) : null,
+				'fraud_prevention_data_cart_contents'   => $order_items_count,
```
allie500, line 85: What happens if we get through all the checks above and `$order_items_count` still ends up being `null` here?

allie500 (commented):
Thanks for working on this. I have a couple inline questions about the changes to `get_hashed_data_for_customer` in the `class-buyer-fingerprinting-service`. I'm working on the manual testing now.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
tpaksu: The verification is done on the very most parent method that handles the AJAX request. Doesn't need a second round of validation here because it's not a direct event handler.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
tpaksu: This is the case I'm trying to solve here, as the Pay for Order pages don't have a cart object available. We're trying to get it from the order, but it looks like it doesn't exist either.

includes/fraud-prevention/class-buyer-fingerprinting-service.php
allie500: Got it. Thanks for explaining. I saw the PHPCS ignore line and it wasn't obvious to me at first glance.

allie500:
Thanks for your continued work on this PR. I ran back through the testing steps again but I found an issue with the second test. See my notes below.

**Testing missing item count meta on Pay for Order message**

non-UPE - works.

Split UPE - works.

Legacy UPE - works.

Split UPE checkout with deferred intent creation - works.


**Testing orders still getting confirmed after a update failure**

Tested with Split UPE. 

Got the expected behavior on `develop`: the order went through.

Got the expected behavior on the PR branch. Order didn't go through. It gave this error on checkout:

> Error: There's a problem with this payment. Please try again or use a different payment method.

But I got the following in the Fraud & Risk meta box:

> Risk filtering through WooCommerce Payments was not found on this order, it may have been created while filtering was not enabled. 

It also didn't show up on my Blocked Transactions screen. No additional order notes were added to the order either.

I would expect the Fraud & Risk meta box to indicate that it was blocked due to my rules. I'd also expect to see a similar note in the order notes section. Lastly, it should also be included in the blocked transactions table.


**Testing declined payments setting a wrong fraud meta box on the order details page**

Tested with non-UPE on `develop`. Card was declined. I got the expected, incorrect Fraud & Risk meta box text.

Tested with non-UPE on the PR branch. Card was declined. I got the expected, correct Fraud & Risk meta box text: 

> No action taken
> The payment for this order passed your risk filtering.

tpaksu:
@allie500 I [talked](https://a8c.slack.com/archives/C01B8KNUYSW/p1687346589485759) with @jessepearson about the issue you found out in the second one, and I think it deserves another issue to be created. It's not in the scope of this PR, and may need a big change. If you can add your review with the light of this info, I'll appreciate.

tpaksu:
@allie500 @jessepearson created the issue: https://github.com/Automattic/woocommerce-payments/issues/6565

includes/fraud-prevention/class-buyer-fingerprinting-service.php
allie500: Based on your changes and my testing yesterday, you've got this sorted out for all forms of UPE as well as non-UPE. Marking as resolved.

allie500 (approved):
Thanks again for working on this and tackling the changes in `get_hashed_data_for_customer` so that it works for non-UPE and UPE. The code changes make sense and my manual testing got the expected results.

LGTM. :shipit:

PR #6487, opened 2023-06-08
Prevent duplicate orders with a single payment intent

### LINKED ISSUE #5786: Avoid multiple orders for a single payment in the UPE flow
### Describe the bug

As a result of this spike 30c9c-pb/#plain, we'd like to ensure this:

- Before updating payment intent, we might check its current `metadata['order_id]` and verify its current value is the same with the current processing order_id in the gateway. Otherwise, just send a failure signal back to the front-end.

### To Reproduce

Copied from the meta link above.

1. Add this snippet into your site, which is to delay saving session data into DB. 
```php
// Delay saving session data during PHP shutdown. 
add_action(
	'shutdown',
	function() {
		if ( 'checkout' === ( $_GET['wc-ajax'] ?? null  ) ){
			sleep(10);
		}
	},
	19 // Just before priority 20 at https://github.com/woocommerce/woocommerce/blob/bec3ec18ac801875a830ddf097ba173069161c13/plugins/woocommerce/includes/class-wc-session-handler.php#L74
);
``` 
2. Use UPE, add a product, go to the checkout page.
3. Click on `Use a new payment method`, to see the input fields. (This triggers creating a new PaymentIntent and caching its ID into WC_Session)
4. Fill out this card `4000 0027 6000 3184` (that requires 3DS), but DO NOT SUBMIT. 
5. Open a new tab (tab 2), also go to the checkout page.
6. Click on `Use a new payment method`, and also fill out the 3DS card `3184`.
7. Go back to tab 1, submit with the "Place order" button
8. Wait for 4-5 seconds. This will ensure that [the intention is updated](https://github.com/Automattic/woocommerce-payments/blob/4ec527e374f2b72fc8f86b4a3fed43b2428e0193/includes/payment-methods/class-upe-payment-gateway.php#L464-L464) with this order_id. 
9. Go to tab 2, submit with the "Place order" button.
10. Close tab 2, after 4-5 seconds, which also ensures that the intention is update with another order_id.
11. Go back to tab 1, finish 3DS.  
12. Wait for a bit, check wp-admin/edit.php?post_type=shop_order. 
13. Expectation: See two new paid/processing orders with the same payment intent. 

### Expected behavior

Prevent two paid orders for the same payment intent.

### Acceptance criteria

- Prevent this issue happening.
- Check both legacy and split UPE flows.
- Consider the trade-off as mentioned by @dechov - pbUcTB-1PE-p2#comment-6110

> Checking for an existing order ID in intent metadata makes sense to me, but that’s if we can do it at a point when we’re already having to retrieve the intent – do we do that (prior to confirmation)?
> 
> I’m not sure we could justify the performance hit of an extra intent retrieval request just for this purpose, and we don’t store the intent ID on our server – at least not in a way that we could relate it to the order ID. Perhaps we could check for an existing order with that payment ID on the plugin side, though I imagine this could be a performance issue for large stores as well since even with HPOS I’d be surprised if there was an index on the column where transaction IDs are stored (though I’m not certain on that).

### TODOs before and after the merge 

- [ ] Gather data for MFF, see more in meta issue 50.

htdat:
Fixes #5786

- Add a new check to compare the current processing order vs the metadata order_id value. In case, they're not matched, the process should be stopped right away. 
- There is another approach here https://github.com/Automattic/woocommerce-payments/pull/6493, which was originally proposed in #5786 . However, I discarded that approach because of the following reasons: 
1. It requires an extra GET request before updating intentions. We're trying to solve this issue for a small percentage of total payment requests (I estimate that it's less than 0.01%) so it's not worth this kind of trade-off. 
2. There are more than one code flow to place this check and ensure it's working properly. 

#### Testing instructions

Follow test steps in #5786, and confirm this issue does not happen.

**Note**: 

- You might check the browser dev network tab, then refresh the check out page a few times. You should start opening two tabs only until there is no AJAX / XHR request to create a new payment intent. That is, the checkout process will use the cache intent saved in the session.
- AJAX requests for creating payment intent: 
    - Split UPE http://localhost:8082/?wc-ajax=wcpay_create_payment_intent_card
    - Legacy UPE: http://localhost:8082/?wc-ajax=wcpay_create_payment_intent

**More expectations from the testing:** 

- Please test for both legacy UPE and split UPE (but no need for deferred intent UPE).
- Check the latest WC Log, and confirm seeing a message like this: 
```
2023-06-08T10:51:11+00:00 ERROR UPE Process Redirect Payment - Order ID mismatched. Received: 999. Intent Metadata Value: 1000
2023-06-08T10:51:11+00:00 INFO Error: We're not able to process this payment. Please try again later.
``` 
- In the example above, order 999 will not be marked as paid (through the sync process in the check-out page), while order 1000 will be marked as paid (through the webhook async processing)

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6487` or branch name `fix/5786-avoid-multiple-order-upe-flow` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/5786-avoid-multiple-order-upe-flow) :rocket:

---

Build info: 

- Latest commit: 20a124fc0ca14ccedfe2bdfde82fe077f1a61d99
- Build time: 2023-07-19 12:33:09 UTC

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -795,13 +802,19 @@ public function process_redirect_payment( $order_id, $intent_id, $save_payment_m
...
-			$message = sprintf( __( 'UPE payment failed: %s', 'woocommerce-payments' ), $e->getMessage() );
-			$this->order_service->mark_payment_failed( $order, $intent_id, $status, $charge_id, $message );
+			if ( ! $is_order_id_mismatched_exception ) {
+				// Confirm our needed variables are set before using them due to there could be a server issue during the get_intent process.
+				$status    = $status ?? null;
+				$charge_id = $charge_id ?? null;
+
+				/* translators: localized exception message */
+				$message = sprintf( __( 'UPE payment failed: %s', 'woocommerce-payments' ), $e->getMessage() );
+				$this->order_service->mark_payment_failed( $order, $intent_id, $status, $charge_id, $message );
```
RadoslavGeorgiev, line 827: Any reason to skip this? A meaningful note could provide context here, and AFAICT there are no downsides to calling `mark_payment_failed()`.

Was there a specific reason this was added?

RadoslavGeorgiev (approved):
The code works as advertised! 🙇  :shipit: 

I have a single nitpick, could be ignored.

To confirm: The issue here is that the webhook comes in, and marks the order as completed, based on the outdated order ID within the payment intent? If yes, we might really need to rethink what we're doing with webhooks.

This doesn't prevent the PR from being shipped, but even testing this change is complicated. I don't imagine anyone going through this scenario during normal testing. We might want to open another issue, to address this differently (in addition to this check, which cannot do any harm).

includes/payment-methods/class-upe-payment-gateway.php
htdat: I've added this commit to explain why that's the case 824dad021cef1716b53172af3225ec4729c72e14

Basically, when we detect the order ID mismatch here, the provided intent ID is no longer relevant for the current processing order.

htdat:
> To confirm: The issue here is that the webhook comes in, and marks the order as completed, based on the outdated order ID within the payment intent? If yes, we might really need to rethink what we're doing with webhooks.

@RadoslavGeorgiev - In the reversed way, actually, the order ID created in tab 2 per steps in this issue https://github.com/Automattic/woocommerce-payments/issues/5786 will be marked as paid. Webhook works correctly to do that. 

It's just that the sync process, i.e. method `UPE_Payment_Gateway::process_redirect_payment` is not working as expected, and this PR is to fix it. 

--- 

Also, I've added two other commits to improve this PR a bit too: a241b75a31a38e1e46fbc3d3e18d961e4cb91ca8 and 26ac73a9c07337f9260635128c13d228114ddbf5

It's ready to review again if necessary.

RadoslavGeorgiev (approved):
Great work, I understood things better. It's ready to 🚢  now :)

naman03malhotra:
Blocked on #6575

naman03malhotra:
The target is to merge it by 6.2, 19th July

htdat:
Ready for another review round @Automattic/gamma 

Changes include:

- ee48d85 - Based on the discussion in https://github.com/Automattic/woocommerce-payments/issues/6575#issuecomment-1614194510, I've modified to apply the new check method `validate_order_id_received_vs_intent_meta_order_id` for payment intents only. That's not ideal but it's acceptable as all the recognized issues so far are related to payment intents (see https://github.com/Automattic/woocommerce-payments/issues/5786). 
- 208411b - add new PHPUnit tests for the new method.

htdat:
This CI failure [E2E Tests - Pull Request / WC - latest | wcpay - merchant (pull_request)](https://github.com/Automattic/woocommerce-payments/actions/runs/5540262945/jobs/10112461112?pr=6487) **seems not** related to this PR. It's related to the dispute process instead:

> FAIL tests/e2e/specs/wcpay/merchant/merchant-disputes-submit-winning.spec.js (123.065 s)
>   Disputes > Submit winning dispute
>     ✕ should process and confirm a winning dispute (100010 ms)
> 
>   ● Disputes > Submit winning dispute › should process and confirm a winning dispute

All recent CI runs from other PRs in https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-pull-request.yml have failed with the same test suite too. 

_Update_: there is another PHP test failure [`WC_Payments_Test::test_ajax_init_woopay_sends_correct_customer_id`](https://github.com/Automattic/woocommerce-payments/actions/runs/5550007902/jobs/10134697555?pr=6487) but that's because of another PR - see p1689305403447819-slack-CGGCLBN58

jessy-p (approved):

jessy-p:
Reviewed the latest code changes and LGTM!

PR #5321, opened 2022-12-21
WCPay as a Platform: Phase 1

RadoslavGeorgiev:
Included PR(s):

- [x] https://github.com/Automattic/woocommerce-payments/pull/4986
- [x] https://github.com/Automattic/woocommerce-payments/pull/5162
- [x] https://github.com/Automattic/woocommerce-payments/pull/5094
- [ ] https://github.com/Automattic/woocommerce-payments/pull/5136 which includes
     - [x] https://github.com/Automattic/woocommerce-payments/pull/5235
     - [x] https://github.com/Automattic/woocommerce-payments/pull/5281

PR #6491, opened 2023-06-09
Allow `card_` prefix when validating payment method IDs to fix failing subscription renewals

### LINKED ISSUE #6490: Subscription renewals fail if payment method uses `card_` prefix
### Describe the bug

WC Subscription renewals fail with the error `card_xxx is not a valid Stripe identifier` if the payment method id is saved with a `card_` prefix. The `card_` IDs were issued under a legacy Stripe API that we no longer use. However, these are accepted as a valid payment method when renewing a subscription.

We currently apply validation for stripe via the [`WCPay\Core\Server\Request::validate_stripe_id()`](https://github.com/Automattic/woocommerce-payments/blob/1356f2f6101e49497f06eb9e0eff1eafb37236d5/includes/core/server/class-request.php#L494) method.

2 calls to `validate_stripe_response()` that enforce `'pm', 'src'`:
* [`WCPay\Core\Server\Request\Create_Intention::set_payment_method`](https://github.com/Automattic/woocommerce-payments/blob/1356f2f6101e49497f06eb9e0eff1eafb37236d5/includes/core/server/request/class-create-intention.php#L50)
* [`WCPay\Core\Server\Request\Create_Setup_Intention::set_payment_method`](https://github.com/Automattic/woocommerce-payments/blob/1356f2f6101e49497f06eb9e0eff1eafb37236d5/includes/core/server/request/class-create-and-confirm-setup-intention.php#L81)

Calling these with `[ 'pm', 'src', 'card' ]` has fixed the immediate issue on WCCOM: p1686268201180419-slack-CGGCLBN58

### To Reproduce

This is not straightforward to reproduce, since Stripe no longer issues `card_` payment method IDs.

### Actual behavior

Subscription renewals fail for subscriptions with an associated `card_` payment method.

### Expected behavior

Subscription renewals should process successfully for subscriptions with an associated `card_` payment method.

### Additional context

* Slack discussion: p1686268201180419-slack-CGGCLBN58
* We're reasonably confident that WCCOM is the only store impacted by this bug, since it only applies to stores that had payment methods migrated from a non-WCPay source

Jinksi:
Fixes #6490

This PR allows the `card_` prefix when validating payment method IDs. This fixes failing subscription renewals for stores that have stored legacy `card_` payment method IDs: namely WCCOM.

2 calls to `validate_stripe_response()` that enforce `'pm', 'src'` now allow the prefix `'card'`:
* [`WCPay\Core\Server\Request\Create_Intention::set_payment_method`](https://github.com/Automattic/woocommerce-payments/blob/1356f2f6101e49497f06eb9e0eff1eafb37236d5/includes/core/server/request/class-create-intention.php#L50)
* [`WCPay\Core\Server\Request\Create_Setup_Intention::set_payment_method`](https://github.com/Automattic/woocommerce-payments/blob/1356f2f6101e49497f06eb9e0eff1eafb37236d5/includes/core/server/request/class-create-and-confirm-setup-intention.php#L81)

#### Testing instructions

**Note:** This original issue described in #6490 is not straightforward to reproduce, since Stripe no longer issues `card_` payment method IDs.

**Regression testing**
1. Install WooCommerce Subscriptions alongside WCPay.
1. Purchase a subscription product. Note the subscription ID as shopper.
2. Navigate to this subscription via **WC→ Subscriptions** as merchant.
3. Process a WC Subscription renewal using the Actions dropdown. You may have to run the scheduled action `woocommerce_scheduled_subscription_payment` in WC → Status → Scheduled Actions.
4. Confirm the renewal order is processed successfully.

-------------------

- [x] Tested on mobile (or does not apply) **N/A**

**Post merge**

### DIFF
--- a/includes/core/server/request/class-create-and-confirm-setup-intention.php
+++ b/includes/core/server/request/class-create-and-confirm-setup-intention.php
@@ -79,6 +79,7 @@ class Create_And_Confirm_Setup_Intention extends Request {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
+		// Including the 'card' prefix to support subscription renewals using legacy payment method IDs.
 		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
 		$this->set_param( 'payment_method', $payment_method_id );
 	}
--- a/includes/core/server/request/class-create-and-confirm-setup-intention.php
+++ b/includes/core/server/request/class-create-and-confirm-setup-intention.php
@@ -79,7 +79,7 @@ class Create_And_Confirm_Setup_Intention extends Request {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
-		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src' ] );
+		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
 		$this->set_param( 'payment_method', $payment_method_id );
 	}
 }
--- a/includes/core/server/request/class-create-intention.php
+++ b/includes/core/server/request/class-create-intention.php
@@ -47,7 +47,7 @@ class Create_Intention extends Request {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
-		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src' ] );
+		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
 		$this->set_param( 'payment_method', $payment_method_id );
 	}
 
--- a/includes/core/server/request/class-create-intention.php
+++ b/includes/core/server/request/class-create-intention.php
@@ -47,6 +47,7 @@ class Create_Intention extends Request {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
+		// Including the 'card' prefix to support subscription renewals using legacy payment method IDs.
 		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
 		$this->set_param( 'payment_method', $payment_method_id );
 	}
--- a/changelog/fix-6490-allow-payment-method-card-prefix
+++ b/changelog/fix-6490-allow-payment-method-card-prefix
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Allow `card_` prefix when validating payment method IDs to fix failing subscription renewals
--- a/tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
+++ b/tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
@@ -166,18 +166,27 @@ class Create_And_Confirm_Intention_Test extends WCPAY_UnitTestCase {
 		$this->assertSame( WC_Payments_API_Client::INTENTIONS_API, $request->get_api() );
 	}
 
-	public function test_woopay_create_intent_request_will_be_created() {
+	public function create_intent_request_provider() {
+		return [
+			[ 'pm_1', 'cus_1' ],
+			[ 'src_2', 'cus_3' ],
+			[ 'card_1', 'cus_1' ],
+		];
+	}
+
+	/**
+	 *   * @dataProvider create_intent_request_provider
+	 */
+	public function test_woopay_create_intent_request_will_be_created( $payment_method, $customer_id ) {
 		$amount     = 1;
 		$currency   = 'usd';
-		$pm         = 'pm_1';
-		$cs         = 'cus_1';
 		$cvc        = 'cvc';
 		$return_url = 'localhost/order-received/';
 		$request    = new WooPay_Create_And_Confirm_Intention( $this->mock_api_client, $this->mock_wc_payments_http_client );
 		$request->set_amount( 1 );
 		$request->set_currency_code( 'usd' );
-		$request->set_payment_method( 'pm_1' );
-		$request->set_customer( 'cus_1' );
+		$request->set_payment_method( $payment_method );
+		$request->set_customer( $customer_id );
 		$request->set_capture_method( true );
 		$request->setup_future_usage();
 		$request->set_metadata( [ 'order_number' => 1 ] );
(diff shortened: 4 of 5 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6491` or branch name `fix/6490-allow-payment-method-card-prefix` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6490-allow-payment-method-card-prefix) :rocket:

---

Build info: 

- Latest commit: 80b764120391e68b1a28608574f44c3de2ebdfe9
- Build time: 2023-06-14 09:08:18 UTC

hesahesa (approved):
we verified the same approach in WCCOM and it works

Jinksi:
@Automattic/gamma I've requested your review on this due to your expertise around the `validate_stripe_id()` introduced in [WCPay as a Platform: Phase 1](https://github.com/Automattic/woocommerce-payments/pull/5321).

Is there any risk introduced here if we allow `card_` prefix for payment methods?

includes/core/server/request/class-create-intention.php
```diff
@@ -47,7 +47,7 @@ public function get_method(): string {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
-		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src' ] );
+		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
```
htdat, line 51: Not urgent but I think we might want to add more tests to cover for this change: 

https://github.com/Automattic/woocommerce-payments/blob/96a1efd7b9e94491a34c0f929661d9c8f636e353/tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php#L79-L83

Also, this method is being used in the child class `Create_And_Confirm_Intention` so we might lift it up to this class?

includes/core/server/request/class-create-and-confirm-setup-intention.php
```diff
@@ -79,7 +79,7 @@ public function set_metadata( array $metadata ) {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
-		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src' ] );
+		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
```
htdat, line 83: @Automattic/gamma - it sounds like this request class `Create_And_Confirm_Setup_Intention` does not have tests under `tests/unit/core/server/request`. I think we will need to add tests for it.

htdat (commented):
> Is there any risk introduced here if we allow `card_` prefix for payment methods?

AFAIK, there is no significant risk. If needed for an urgent patch, yes, we can use it. 

Other than that, I have some inline comments, mostly in regard to PHP Unit tests. 

**_Update_: it would be great if we add comments to explain why the `card` prefix is needed here. It's an edge-case knowledge for almost every WCPay devs.**

Jinksi:
> Update: it would be great if we add comments to explain why the card prefix is needed here. It's an edge-case knowledge for almost every WCPay devs.

Thanks @htdat! I've added the following comments in 8469698:
```
// Including the 'card' prefix to support subscription renewals using legacy payment method IDs.
```

jessy-p:
Next step for this PR: Gamma to add tests before merging.

naman03malhotra:
Adding 1sp for this effort. Jessy, please increase if required.

includes/core/server/request/class-create-intention.php
```diff
@@ -47,7 +47,8 @@ public function get_method(): string {
 	 * @throws Invalid_Request_Parameter_Exception
 	 */
 	public function set_payment_method( string $payment_method_id ) {
-		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src' ] );
+		// Including the 'card' prefix to support subscription renewals using legacy payment method IDs.
+		$this->validate_stripe_id( $payment_method_id, [ 'pm', 'src', 'card' ] );
```
RadoslavGeorgiev, line 51: This is a different discussion, but I'm thinking of isolating those prefixes into a separate constants class. This way we'd have all definitions in one place, and a central explanation why `card` might be included in the mix.

Something like this:
```php
namespace WCPay\Core\Server;

abstract class Stripe_Prefix {
	/**
	 * Prefix for PaymentMethod objects.
	 *
	 * @var string
	 */
	const PAYMENT_METHOD = 'pm';

	/**
	 * Prefix for Source objects.
	 *
	 * Sources preceeded payment methods, and were never implemented in the
	 * gateway, but legacy stores should support them.
	 *
	 * @var string
	 */
	const SOURCE = 'src';

	/**
	 * Prefix for Card objects.
	 *
	 * Cards preceeded sources, and were never implemented in
	 * the gateway, but legacy stores should support them.
	 *
	 * @var string
	 */
	const CARD = 'card';
}
```

How does that sound?

RadoslavGeorgiev (commented):
Thanks for taking this up @Jinksi! 🙌 

I left one comment, I'm curious whether it makes sense to emphasize on prefixes a bit more.

By the way, it's interesting. I seem to have added `src` [here](https://github.com/Automattic/woocommerce-payments/commit/ef10ff85f66961312a7c1ba7bdfd93dc46dd7668#diff-6a9840ad4b37c00ddd84090575c6cdbf2f6c1d6aa36ffa2a27e3413bb58b3e92R100), but never knew of needing to add `card` as well. We need to check whether that's needed for refunds as well in [this PR](https://github.com/Automattic/woocommerce-payments/pull/4663).

includes/core/server/request/class-create-intention.php
htdat: > This is a different discussion, but I'm thinking of isolating those prefixes into a separate constants class. 

Good idea! But I think we don't often to use these values separately, so we just need to have an array aggregating those values like this:

```php
	const VALID_PAYMENT_METHODS = [ 'pm', 'src', 'card' ]; 
```

RadoslavGeorgiev: That would probably be even more useful, as we would not create a new request later, and miss any of them.  And with a PHPDoc comment, describing everything, that would be ideal.

RadoslavGeorgiev:
@jessy-p the tests look great, thank you for adding them! 🙇  🟢

htdat (approved):
@jessy-p Pre-approve PR. Thanks for adding the new test file. 

Can we also add tests for two prefixes `src` and `card` in `tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php` too? Currently, it's only testing `pm`.

htdat (approved):
@jessy-p - LGTM! 🚢 Thanks for adding more tests and addressing my feedback.

PR #6509, opened 2023-06-13
Fix condition to enable WooPay in the checkout page

asumaran:
Restores a missing condition required to enable WooPay in the checkout page. Spotted by @bborman22.

It was removed by mistake in https://github.com/Automattic/woocommerce-payments/pull/6061 as part of the naming convention update.

#### Testing instructions
- Make sure you are logged out from the merchant store.
- Login to WooPay.
- Go to the merchant store and add a simple subscription to the cart.
- Go to the checkout page.
- Enter the user's WooPay email
- WooPay popup **should not** be displayed.
- Remove the subscription and add another product.
- Go to the checkout page.
- Enter the user's WooPay email.
- WooPay popup should be displayed.

### DIFF
--- a/tests/unit/test-class-wc-payment-gateway-wcpay.php
+++ b/tests/unit/test-class-wc-payment-gateway-wcpay.php
@@ -2145,8 +2145,8 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 		$this->wcpay_gateway->update_option( 'platform_checkout', 'yes' );
 		$this->assertTrue( $this->woopay_utilities->should_enable_woopay( $this->wcpay_gateway ) );
 
-		// This will return TRUE because woopay_utilities->should_enable_woopay() will return true.
-		$this->assertTrue( $this->payments_checkout->get_payment_fields_js_config()['isWooPayEnabled'] );
+		// This will return false because woopay_utilities->should_enable_woopay_on_cart_or_checkout() will return false.
+		$this->assertFalse( $this->payments_checkout->get_payment_fields_js_config()['isWooPayEnabled'] );
 	}
 
 	public function test_should_use_stripe_platform_on_checkout_page_not_woopay_eligible() {
--- a/tests/unit/test-class-wc-payment-gateway-wcpay.php
+++ b/tests/unit/test-class-wc-payment-gateway-wcpay.php
@@ -80,13 +80,6 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	private $mock_wcpay_account;
 
 	/**
-	 * WooPay_Utilities instance
-	 *
-	 * @var WooPay_Utilities
-	 */
-	private $woopay_utilities;
-
-	/**
 	 * Session_Rate_Limiter instance.
 	 *
 	 * @var Session_Rate_Limiter|PHPUnit_Framework_MockObject_MockObject
@@ -99,6 +92,13 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	 * @var WC_Payments_Order_Service
 	 */
 	private $order_service;
+
+	/**
+	 * WooPay_Utilities instance.
+	 *
+	 * @var WooPay_Utilities
+	 */
+	private $woopay_utilities;
 
 	/**
 	 * WC_Payments_Checkout instance.
--- a/tests/unit/test-class-wc-payment-gateway-wcpay.php
+++ b/tests/unit/test-class-wc-payment-gateway-wcpay.php
@@ -80,6 +80,14 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	private $mock_wcpay_account;
 
 	/**
+	 * WooPay_Utilities instance
+	 *
+	 * @var WooPay_Utilities|PHPUnit_Framework_MockObject_MockObject
+	 *
+	 */
+	private $mock_woopay_utilities;
+
+	/**
 	 * Session_Rate_Limiter instance.
 	 *
 	 * @var Session_Rate_Limiter|PHPUnit_Framework_MockObject_MockObject
@@ -92,13 +100,6 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	 * @var WC_Payments_Order_Service
 	 */
 	private $order_service;
-
-	/**
-	 * WooPay_Utilities instance.
-	 *
-	 * @var WooPay_Utilities
-	 */
-	private $woopay_utilities;
 
 	/**
 	 * WC_Payments_Checkout instance.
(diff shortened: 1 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6509` or branch name `as-fix-subs-woopay` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=as-fix-subs-woopay) :rocket:

---

Build info: 

- Latest commit: a110666ddc7acac0161b8c4b11c9db17d8b07c74
- Build time: 2023-07-07 16:58:00 UTC

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -2135,8 +2135,8 @@ public function test_is_woopay_enabled_returns_true() {
 		$this->wcpay_gateway->update_option( 'platform_checkout', 'yes' );
 		$this->assertTrue( $this->woopay_utilities->should_enable_woopay( $this->wcpay_gateway ) );
 
-		// This will return TRUE because woopay_utilities->should_enable_woopay() will return true.
-		$this->assertTrue( $this->payments_checkout->get_payment_fields_js_config()['isWooPayEnabled'] );
+		// This will return false because woopay_utilities->should_enable_woopay_on_cart_or_checkout() will return false.
+		$this->assertFalse( $this->payments_checkout->get_payment_fields_js_config()['isWooPayEnabled'] );
```
reykjalin: I think we need to update this test to work a little differently. It's called `test_is_woopay_enabled_returns_true()` but the assertion is checking that it's actually false 🤔 

Not sure what the best way to do that is, but maybe there's a way to force `should_enable_woopay_on_cart_or_checkout()` to return `true`?

reykjalin (changes requested):
Instructions are missing the step to make sure you log out of the merchant store, but otherwise this tests well.

I think we need to change the unit test a bit, but functionally this is looking great 👍

bborman22 (commented):
I was just taking a look at this PR earlier and was able to confirm everything worked as expected!

I also left a note about that one odd test case which we should clean up, but otherwise the changes looked good!

tests/unit/test-class-wc-payment-gateway-wcpay.php
bborman22: This is definitely an odd test case now with those conflicts on true / false. I was thinking one way to handle it is if we set a current user (`wp_set_current_user( 1 )`) and can set the `is_checkout` function to be true (I couldn't figure this one out quickly), then I think we would get our desired result. I wouldn't be a fan of it, but I suppose we could also mock `should_enable_woopay_on_cart_or_checkout` as we're primarily checking the JS config result?

asumaran: Tweaked the test to make sure we are checking the expected value here ~3a21a49c0967e8e00ce0f72a7edfe3a4123d9e73~ bc1bef4ea0a28b6e71eb14360aa4359e39b0845d

bborman22: Thanks so much for taking a look at that @asumaran I was still a bit hesitant to use a mock there, especially since that utility function is used in other tests. I came up with the following diff that seems to get the test passing without a mock. Basically I just set some conditions that would allow the result to be true (on the checkout page, logged in user). Let me know what you think!

```diff
diff --git a/tests/unit/test-class-wc-payment-gateway-wcpay.php b/tests/unit/test-class-wc-payment-gateway-wcpay.php
index 59549b79..aa7ece07 100644
--- a/tests/unit/test-class-wc-payment-gateway-wcpay.php
+++ b/tests/unit/test-class-wc-payment-gateway-wcpay.php
@@ -82,10 +82,9 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	/**
 	 * WooPay_Utilities instance
 	 *
-	 * @var WooPay_Utilities|PHPUnit_Framework_MockObject_MockObject
-	 *
+	 * @var WooPay_Utilities
 	 */
-	private $mock_woopay_utilities;
+	private $woopay_utilities;
 
 	/**
 	 * Session_Rate_Limiter instance.
@@ -179,11 +178,11 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 			$this->mock_dpps
 		);
 
-		$this->mock_woopay_utilities = $this->createMock( WooPay_Utilities::class );
+		$this->woopay_utilities = new WooPay_Utilities();
 
 		$this->payments_checkout = new WC_Payments_Checkout(
 			$this->wcpay_gateway,
-			$this->mock_woopay_utilities,
+			$this->woopay_utilities,
 			$this->mock_wcpay_account,
 			$this->mock_customer_service
 		);
@@ -2142,18 +2141,15 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 	}
 
 	public function test_is_woopay_enabled_returns_true() {
-		$this->mock_woopay_utilities
-			->method( 'should_enable_woopay_on_cart_or_checkout' )
-			->willReturn( true );
-		$this->mock_woopay_utilities
-			->method( 'should_enable_woopay' )
-			->willReturn( true );
-
 		$this->mock_cache->method( 'get' )->willReturn( [ 'platform_checkout_eligible' => true ] );
 		$this->wcpay_gateway->update_option( 'platform_checkout', 'yes' );
+		wp_set_current_user( 1 );
+		add_filter( 'woocommerce_is_checkout', '__return_true' );
 
-		$this->assertTrue( $this->mock_woopay_utilities->should_enable_woopay( $this->wcpay_gateway ) );
+		$this->assertTrue( $this->woopay_utilities->should_enable_woopay( $this->wcpay_gateway ) );
 		$this->assertTrue( $this->payments_checkout->get_payment_fields_js_config()['isWooPayEnabled'] );
+
+		remove_filter( 'woocommerce_is_checkout', '__return_true' );
 	}
 
 	public function test_should_use_stripe_platform_on_checkout_page_not_woopay_eligible() {
@@ -2178,7 +2174,7 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 
 		$payments_checkout = new WC_Payments_Checkout(
 			$mock_wcpay_gateway,
-			$this->mock_woopay_utilities,
+			$this->woopay_utilities,
 			$this->mock_wcpay_account,
 			$this->mock_customer_service
 		);
@@ -2323,7 +2319,7 @@ class WC_Payment_Gateway_WCPay_Test extends WCPAY_UnitTestCase {
 
 		$this->payments_checkout = new WC_Payments_Checkout(
 			$this->wcpay_gateway,
-			$this->mock_woopay_utilities,
+			$this->woopay_utilities,
 			$this->mock_wcpay_account,
 			$this->mock_customer_service
 		);
```

asumaran: @bborman22 Cool! That's a much better approach I think. Can you commit the changes or can I do it?

bborman22: Sure I can commit that now!

bborman22 (approved):
With the test updated and the functionality testing well and the code changes being straightforward, I think this is good to go! 

LGTM!

asumaran:
@bborman22 E2E tests keep failing for reasons unrelated to this PR. Can we merge anyway?

bborman22:
As you mentioned and I double checked as well, these test failures seem unrelated and are occurring on other PRs in the same way. I would say we are good to merge this PR now.

PR #6537, opened 2023-06-15
Deferred intent creation for UPE in Blocks checkout

### LINKED ISSUE #5020: Enhanced UPE - Blocks integration

timur27:
Closes https://github.com/Automattic/woocommerce-payments/issues/5020

This PR adds support for deferred intent creation while using split UPE on the Blocks checkout page, thus being a continuation of deferred intent creation UPE for _shortcode_ checkout page (https://github.com/Automattic/woocommerce-payments/pull/5869) and _My account_ page (https://github.com/Automattic/woocommerce-payments/pull/6390).

There are two new scripts introduced with this PR. `payment-elements.js` exports the React component responsible for rolling out, or mounting, Stripe Payment Element. It adds the `PaymentProcessor` component defined in `payment-processor.js`, which implements payment processing and error handling logic. This process of rendering `<Elements>` and `<PaymentElement>` Stripe React components  is well-described in [Stripe docs](https://stripe.com/docs/payments/finalize-payments-on-the-server?platform=web&type=payment#add-and-configure-the-elements-provider-to-your-checkout-page), in case anyone is interested.

`PaymentElement` component consists of the logic for [generating the element appearance options and checking if there are any problems with a fingerprint](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-deferred-intent-creation/payment-elements.js#L22-L43) - these things are well-known from previous implementations and don't differ from them. The main change this component introduces is [the way `<Elements>` component is rendered](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-deferred-intent-creation/payment-elements.js#L49-L56), which is done exactly the same way as in previous deferred intent implementation e.g. for shortcode checkout. 

`PaymentProcessor` also has some already known steps. When the _Place Order_ button is clicked on Blocks checkout page, it: 

1. [Runs input validation](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js#L81-L117) of different kinds and shows an error in case anything is in an unexpected state
2. [Creates payment method object](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js#L119-L128)
3. Returns [the result](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js#L130-L141) which is then being processed by WC Core and eventually, by [`class-wc-payment-gateway-wcpay.php`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L710)

At first glance, it looks like there are some new functions introduced [here](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/utils/upe.js#L223-L302). The thing is, that those functions were just moved from other implementations e.g. `upe-split-fields.js`, so that they can be re-used by deferred intent creation implementation for Blocks checkout. Having that said, some refactoring was done, but I tried to keep it balanced and avoid moving possibly _all_ functions to `utils/upe.js` - the reasoning here is the fact, that if at some point we'll decide to remove split UPE implementation altogether, it will be more expensive to move all the functions from utils back to the deferred intent creation implementation. 

In terms of rolling out scripts, I went with deciding on which UPE scripts to enqueue on [the front-end](https://github.com/Automattic/woocommerce-payments/blob/318d3353b10b68bfe9721ee2ffc343c5ba73d7c2/client/checkout/blocks/upe-split.js#L75-L87). Split UPE and deferred intent UPE rollout implementations are identical in this area, and it wasn't worth it to create a separate script which contains the exact same logic. 

1. In WCPay Dev Tools, uncheck the **_Enable UPE checkout (legacy) and Enable Split UPE checkout_** feature flags. At the same time, check the _Enable Split UPE checkout with deferred intent creation_ flag.
5. Set the currency to EUR on the merchant store
6. Add a simple product to the cart, checkout using Blocks checkout and card, and confirm that only one payment intent is created and processed successfully in the Stripe Dashboard. The intent should be created _**after**_ the _Place order_ button is clicked, opposite to the split UPE implementation where the intent is created before Stripe Payment Element rendering.
7. Repeat the above step. Instead of the card, pay with any other payment gateway (e.g., `giropay`, `eps` etc.) and confirm a single intent approach works as expected when a payment is accepted on the test third-party authentication page.
8. Repeat the above step but reject the payment on the test third-party authentication page and confirm that you are redirected to the checkout page and that the payment intent is not confirmed on the Stripe side. The order status should be `Failed`. (Note: after failing authentication, users are redirected to the shortcode checkout page, and not to the Blocks. I observe it in every UPE and non-UPE implementation. In case this should be considered a bug, it would make sense to address it in a separate issue). 
9. Checkout with 3DS (`4000002760003184`) and 3DSv2 (`4000000000003220`) card. Confirm that the bank authentication pop-up modal appears on the screen.
10. Test the transaction failure for any of the above 3DS cards - confirm that the error notice is successfully shown to the user on the checkout page after the transaction failure, and confirm that the Order status is `Failed`. 
11. Repeat the attempt but authorize the transaction this time. Confirm that the order status has changed from `Failed` to `Processing` in the `WooCommerce -> Orders`. Confirm that the order status/history are in line with the payment status/history in Stripe Dashboard.
12. Save the default card and confirm that it's saved successfully. After clicking the save checkbox, confirm that the Terms are appearing within the Payment Element.
13. Purchase a subscription and pay with a default card. After purchase, ensure the subscription was purchased successfully, and the card is saved.
14. Using [the WooCommerce Subscriptions plugin](https://github.com/woocommerce/woocommerce-subscriptions/releases), renew the previously bought subscription. Confirm that the status of the subscription is active and the Stripe dashboard has the payment entry for the renewal, which is not in status `Succeeded`

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6537` or branch name `feature/deferred-upe-blocks` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=feature/deferred-upe-blocks) :rocket:

---

Build info: 

- Latest commit: 3e1b6684b54b6a1946274a5bc82119623486b072
- Build time: 2023-06-20 16:06:54 UTC

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -0,0 +1,204 @@
...
+import {
+	getStripeElementOptions,
+	useCustomerData,
+} from 'wcpay/checkout/utils/upe';
+import { getUPEConfig } from 'wcpay/utils/checkout';
+import { validateElements } from 'wcpay/checkout/classic/upe-deferred-intent-creation/payment-processing';
+
+const getBillingDetails = ( billingData ) => {
+	return {
+		name: ( billingData.first_name + ' ' + billingData.last_name ).trim(),
```
FangedParakeet:
```suggestion
		name: `${ billingData.first_name } ${ billingData.last_name }`.trim(),
```
🤷 Little tidier maybe? Up 2 u.

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -0,0 +1,204 @@
...
+	const upeOnChange = ( event ) => {
+		// Update WC Blocks gateway config based on selected UPE payment method.
+		const paymentType =
+			'link' !== event.value.type ? event.value.type : 'card';
+		gatewayConfig.supports.showSaveOption =
+			paymentMethodsConfig[ paymentType ].showSaveOption;
+
+		setIsUPEComplete( event.complete );
+		setSelectedUPEPaymentType( paymentType );
+	};
```
FangedParakeet:
```suggestion
	// Checks whether there are errors within a field, and saves them for later reporting.
	const upeOnChange = ( event ) => {
		setIsUPEComplete( event.complete );
	};
```
Do we need all this? With the [original legacy UPE](https://github.com/Automattic/woocommerce-payments/blob/feature/deferred-upe-blocks/client/checkout/blocks/upe-fields.js#L337), where the payment element contained multiple payment methods, we used the `onChange` event to keep track of which specific payment method was actually selected. This would also control whether the save payment checkbox and terms would appear or not, depending on the compatibility of the selected payment method.

With the split UPE, we will know the payment method type for each mounted payment element when we mount the element. I think the only relevant value that may change during checkout is the `complete` value. I think all the other values can be initialised once and then left that way, AFAICT. 🤔 

Though, looks like I was the one who originally implemented this `upeOnChange` handler like this inside `upe-split-fields.js`, so maybe I'm to blame here. 😅 If there was a logical reason why I did this before, I can't recall it now, but it appears to me that this could be improved now.

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -0,0 +1,204 @@
...
+						return {
+							type: 'error',
+							message: errorMessage,
+						};
+					}
+
+					if (
+						gatewayConfig.supports.showSaveOption &&
+						shouldSavePayment &&
+						! paymentMethodsConfig[ selectedUPEPaymentType ]
```
FangedParakeet:
```suggestion
						! gatewayConfig.isReusable
```
Minor mention, discussed this in another comment, but I think `selectedUPEPaymentType` will be static in the split UPE, so I think we can use `gatewayConfig` here.

client/checkout/blocks/upe-split.js
```diff
@@ -71,22 +72,32 @@ Object.entries( enabledPaymentMethodsConfig )
...
+						upeMethods,
+						api,
+						upeConfig.testingInstructions
+				  )
+				: getDeferredIntentCreationUPEFields(
+						upeName,
+						upeMethods,
+						api,
+						upeConfig.testingInstructions
+				  ),
```
FangedParakeet:
```suggestion
			content: getUPEConfig( 'isUPEDeferredEnabled' )
				? getDeferredIntentCreationUPEFields(
						upeName,
						upeMethods,
						api,
						upeConfig.testingInstructions
				  )
				: getSplitUPEFields(
						upeName,
						upeMethods,
						api,
						upeConfig.testingInstructions
				  ),
```
Thoughts on checking the deferred intent feature flag here instead? The advantage is we can rely on the most relevant feature flag alone and can ensure that on rollout we only have to switch on the new flag and not worry about the previous flag. This would theoretically allow us to distinguish between users who have upgraded to deferred intent UPE from split UPE and those upgrading from previous non-UPE sites (permitting to fall back to the correct feature flag, if we disabled deferred intent UPE for whatever reason).

It's a minor issue and there are good arguments in either direction, so I can leave this one to your discretion. 🤷

FangedParakeet (commented):
Tried as many checkout combinations as I could think of, but the blocks checkout is bouncing along mightily well for me--nice one! Just left a few suggestions for some minor adjustments that might be made, but otherwise this one's looking in great shape. 🙌 🎉

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
timur27: I've checked all the values, and here's the result: 
1. `gatewayConfig.supports.showSaveOption` doesn't need to be set here since [the gateway config already comes with the relevant data for this setting](https://github.com/Automattic/woocommerce-payments/blob/7c5b120eaa82d82c58c0e883a7e93673ef92920a/client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js#L74) ✔️ 
2. `selectedUPEPaymentType` indeed is not a must, mainly because [the `isReusable` setting can be easily fetched here by using `paymentMethodId` instead of `selectedUPEPaymentType`](https://github.com/Automattic/woocommerce-payments/blob/7c5b120eaa82d82c58c0e883a7e93673ef92920a/client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js#L107-L108) ✔️ 

Having said that, I agree only `isUPEComplete` can be set within this function. I've adjusted the code in cc99e6a35aa468b84df4fd6cb3582bc7939e9696, where I also changed the function names to reflect the actual use case more precisely now. The code is shorter and easier to read now - thanks!

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
timur27: Currently, the `isReusable` flag isn't included in `gatewayConfig`. This could be potentially changed [by extending this `supports` array](https://github.com/Automattic/woocommerce-payments/blob/7c5b120eaa82d82c58c0e883a7e93673ef92920a/client/checkout/blocks/upe-split.js#L114-L118). However, taking the setting from `paymentMethodsConfig[ paymentMethodId ].isReusable` seems to be a good candidate as well which also requires fewer changes so I went with this approach as part of cc99e6a35aa468b84df4fd6cb3582bc7939e9696. Let me know if this change looks good to you or not 👍

client/checkout/blocks/upe-split.js
timur27: This topic is an excellent area to refine further, and I'm happy you started the discussion. This script will be executed in three scenarios: 
1. split UPE FF enabled
2. deferred intent UPE FF enabled
3. both FFs enabled

The first two points are irrelevant in our discussion since, regardless of the final approach we'll go with, the proper implementation will be chosen. For the third point, we'd need to decide which implementation takes precedence. 

> The advantage is we can rely on **the most relevant feature flag** alone

From your comment and all the details you mentioned above, I understand it's deferred intent UPE that needs to take precedence. Also, I could consider [the shortcode implementation where deferred is a higher priority](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L126-L132). 

Thus I'll change the implementation shortly to leverage the `isUPEDeferredEnabled` flag instead. 

UPDATE: changed in a5e8c3af52fed2e2a81e600befd00a64b8e95758

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
FangedParakeet: Aye aye, cap. I think I misread this one and misunderstood that `gatewayConfig` was returned by blocks' own `getPaymentMethods()` function rather than our own payment methods config. That being said, I'm glad that the redundant `selectedUPEPaymentType` variable has been exterminated, so thanks for making that change!

FangedParakeet (approved):
Thanks for responding to all of my feedback. LTGM now and it working as wonderfully as it was the first time! :shipit:

PR #6559, opened 2023-06-20
Allow WooPay verified email requests

alefesouza:

Allow authenticated store API requests if the customer has their e-mail address verified on WooPay.

#### Testing instructions

* Follow 1921-gh-Automattic/woopay testing instructions.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6559` or branch name `add/allow-woopay-verified-email-requests` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/allow-woopay-verified-email-requests) :rocket:

---

Build info: 

- Latest commit: d29ec4a3bd76fca167ea0f0f845c5e60f417fbc3
- Build time: 2023-07-17 18:32:14 UTC

alefesouza:
I'd like some opinions about the approach before adding tests.

includes/woopay/class-woopay-utilities.php
```diff
@@ -23,6 +23,7 @@ class WooPay_Utilities {
 	const AVAILABLE_COUNTRIES_KEY            = 'woocommerce_woocommerce_payments_woopay_available_countries';
 	const AVAILABLE_COUNTRIES_LAST_CHECK_KEY = 'woocommerce_woocommerce_payments_woopay_available_countries_last_check';
 	const AVAILABLE_COUNTRIES_DEFAULT        = '["US"]';
+	const SUPPORTED_EXTENSIONS               = [ 'woocommerce-points-and-rewards/woocommerce-points-and-rewards.php', 'woocommerce-gift-cards/woocommerce-gift-cards.php' ];
```
alefesouza: This list will be updated on WooPay after https://github.com/Automattic/woocommerce-payments/pull/6562.

leonardola (approved):
Tests ok.
Code looks good.
We'll need more eyes on this one as it messes with the session

includes/woopay/class-woopay-store-api-token.php
```diff
@@ -57,5 +58,25 @@ public function get_args() {
...
+		 *
+		 * @param string $id The user ID.
+		 *
+		 * @return string The cart token.
+		 * @psalm-suppress UndefinedMethod
+		 */
+		public function get_store_api_token_for_user_id( $id ) {
+			return JsonWebToken::create(
+				[
+					'user_id' => $id,
```
marcinbot: If I understand correctly that's what the `get_cart_token` did: gets the WP user ID if logged in, otherwise a non-numeric ID. [[source 1](https://github.com/woocommerce/woocommerce-blocks/blob/d59afbce01e374d207d4bf47292ba97d2888d4a3/src/StoreApi/Routes/V1/AbstractCartRoute.php#L181C33-L181C48)] [[source 2](https://github.com/woocommerce/woocommerce/blob/040f1387170202bf99bf9a1c0ecaf6bf91ac75a1/plugins/woocommerce/includes/class-wc-session-handler.php#L208C4-L208C17)] So why not stick to that and check if the tokens ID is numeric?

alefesouza: On this PR if the current user is guest but have a matching e-mail address on the merchant store we send two tokens to WooPay, one from `get_cart_token` for the current user session (the same way we use today) and a [second one](https://github.com/Automattic/woocommerce-payments/blob/764d829187b15412b1d79d9e26504af8c234d4c5/includes/class-wc-payments.php#L1489) from this function for the user who matches the e-mail address registered on the merchant store, `get_cart_token` does not allow us to choose a different user.

We use the last token only if the user has the e-mail verified on WooPay to get and apply this user extension data (ex. Points and Rewards and Gift Cards balance) to the cart on WooPay itself, this way the user doesn't need to be logged in into the merchant store if the e-mail is verified on WooPay.

marcinbot: Ok, I see that now. It should work as long as we verify the address in WooPay.

But we still duplicate the mechanism for the token generation, whereas the entire point of the API Token class was not to do that in case the logic ever changes in core. We also send two tokens which is a bit messy.

Also the name of `supported_extensions` can become a bit confusing in the future. In this case, the "support" is validating the email on guest checkout. But in the future, it could mean something completely different for another extension.

Would we be open to modifying the extensions themselves to, instead of expecting an authenticated user, return the gift cards/points for the user with the validated email? I think that might be the most appropriate approach rather than hacking WCPay to trick them to work and risking introducing side effects. But it would mean that the store would have to use the latest version of them.

Or, at the very least, we might want to rethink the name of supported extensions to break down what support means.

alefesouza: > Also the name of supported_extensions can become a bit confusing in the future

On https://github.com/Automattic/woocommerce-payments/pull/6562 it's renamed to "adapted extensions", I've been using this name on the WooPay PRs.

> Would we be open to modifying the extensions themselves to, instead of expecting an authenticated user, return the gift cards/points for the user with the validated email?

The extensions themselves need the user to be authenticated to make sure the discount is being applied to the right user, by validated e-mail you mean a flag sent by WooPay that makes the extension identify if the e-mail is verified? I believe we still would need another token for it as it cannot be a single flag because the extension data is a private data (ex anyone could see someone points balance and apply it with this flag).

I don't believe there will be side effects with this PR approach as we authenticate the user as expected for the extension.

asumaran: How long does the token live? Do we have to update or renew the token in WooPay? 

For the regular Cart-Token we update it on every request.

asumaran (commented):
Test as expected, I just left a comment for you @alefesouza

includes/woopay/class-woopay-store-api-token.php
alefesouza: The token was without expiration, good catch, on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/fa8d6dc2846dcfddbbcc556594fbeb6be27497eb we add a condition to check the token expiration, it's valid for two days [as the cart token](https://github.com/woocommerce/woocommerce-blocks/blob/1b4985ad9a86b083d7f6fa6e6bd5dde432203478/src/StoreApi/Routes/V1/AbstractCartRoute.php#L211).

includes/class-wc-payments.php
```diff
@@ -1468,6 +1468,8 @@ public static function ajax_init_woopay() {
 
 		$store_logo = self::get_gateway()->get_option( 'platform_checkout_store_logo' );
 
+		$store_api_token = self::init_store_api_token();
```
ricardo: Since we've refactored `init_store_api_token` to return `WooPay_Store_Api_Token::init()` directly now, I'd vote for removing the extra step.

```php
$store_api_token = WooPay_Store_Api_Token::init();
```

Mostly to reduce noise and because `init_store_api_token` doesn't add any logic anymore.

includes/woopay/class-woopay-utilities.php
```diff
@@ -282,4 +283,19 @@ public static function get_woopay_rest_url( $endpoint ) {
...
+	 */
+	public function has_compatible_extension_installed() {
+		foreach ( self::SUPPORTED_EXTENSIONS as $supported_extension ) {
+			if ( in_array( $supported_extension, apply_filters( 'active_plugins', get_option( 'active_plugins' ) ), true ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
ricardo: I think we should aim for consistency with the terms "supported" vs. "compatible" and "installed" vs. "active".

From the function name, it seems we're checking if it has "compatible" and "installed" plugins, but from the function body, we're checking for active plugins (not just installed).

I understand these can be read as synonyms but just more of a consistency thing 🤏.

includes/class-wc-payments.php
```diff
@@ -1499,6 +1501,18 @@ public static function ajax_init_woopay() {
 			],
 			'user_session'    => isset( $_REQUEST['user_session'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['user_session'] ) ) : null,
 		];
+
+		$has_compatible_extension_installed = self::$woopay_util->has_compatible_extension_installed();
+		if ( $has_compatible_extension_installed && ! empty( $email ) && ! is_user_logged_in() ) {
```
ricardo: I agree with Marcin on the term "supported" becoming confusing in the future. Maybe "adapted" would be a better term and we could make it consistent everywhere.

We might also benefit from adding a comment explaining that the extra token is necessary for certain adapted extensions.

includes/woopay/class-woopay-session.php
```diff
@@ -123,27 +123,54 @@ private static function get_user_id_from_cart_token() {
...
+		if ( null === $user_id ) {
+			return null;
+		}
 
-			if ( empty( $payload ) ) {
-				return null;
-			}
+		$session_handler = new SessionHandler();
+		$session_data    = $session_handler->get_session( $user_id );
+		$customer        = maybe_unserialize( $session_data['customer'] );
```
ricardo: @alefesouza Have we considered the feasibility of keeping the verified user ID as an extra session data for the current session instead of introducing a different cart token?

I'm not sure if this is possible, or any better than the current approach, but I guess we could introduce a new key into the WooPay session handler similar to what you've done in `setup_session` in the tests, and then retrieve it here as the user ID in case it exists.

Alternatively, have we considered replacing the current session's user ID instead of introducing a new cart token?

Just dumping some thoughts before I tested anything thoughroughlly.

cc @asumaran as you might have insights into these questions as well.

includes/class-wc-payments.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/6283f0a4e4826ed92922a62b3de826e9596310c7.

includes/woopay/class-woopay-utilities.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/6283f0a4e4826ed92922a62b3de826e9596310c7.

includes/class-wc-payments.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/6283f0a4e4826ed92922a62b3de826e9596310c7.

includes/woopay/class-woopay-session.php
alefesouza: > the feasibility of keeping the verified user ID as an extra session data for the current session instead of introducing a different cart token?

You mean using [this line](https://github.com/Automattic/woocommerce-payments/blob/e45fc8418c3dadfa67e9c780419e25c0497f96cf/includes/class-wc-payments.php#L1513) and [this line](https://github.com/Automattic/woocommerce-payments/blob/add/allow-woopay-verified-email-requests/includes/woopay/class-woopay-session.php#L142) without the token verification? I'd prefer using the token to prevent someone from changing the session data (using some plugin maybe?) and then be logged in with the desired customer id.

includes/woopay/class-woopay-utilities.php
ricardo: @alefesouza do you think it's appropriate to use "active" instead of "installed" as an installed extension may not be active?

includes/woopay/class-woopay-store-api-token.php
marcinbot: > The token was without expiration, good catch, on https://github.com/Automattic/woocommerce-payments/commit/fa8d6dc2846dcfddbbcc556594fbeb6be27497eb we add a condition to check the token expiration, it's valid for two days [as the cart token](https://github.com/woocommerce/woocommerce-blocks/blob/1b4985ad9a86b083d7f6fa6e6bd5dde432203478/src/StoreApi/Routes/V1/AbstractCartRoute.php#L211).

Maybe we should add such a check for the other token as well, if it's not there already?

> The extensions themselves need the user to be authenticated to make sure the discount is being applied to the right user, by validated e-mail you mean a flag sent by WooPay that makes the extension identify if the e-mail is verified?

I meant that all of this logic should live in the affected extensions since it could be argued that they are responsible for handling such integrations. IMO it could go either way, so let's leave it aside.

> I believe we still would need another token for it as it cannot be a single flag because the extension data is a private data (ex anyone could see someone points balance and apply it with this flag).

Yes, but once we get to this point, we will already have authenticated with the regular token and confirmed that the request came from WooPay. The request cannot be spoofed without proper authentication and intercepting the network traffic between WooPay and the site.

So we might as well use just a string field (`email_address`) and a bool flag (`is_email_validated`) to indicate that we recognise the user and log them in here (after confirming that the email is the same as in the session). 

So looks like there's no need for two tokens, which would greatly simplify the code here.

alefesouza: > Maybe we should add such a check for the other token as well, if it's not there already?

The session returned by the token has the expiration less than the token expiration [here](https://github.com/Automattic/woocommerce-payments/blob/38c63a1c7bec594b9f6a51e3d6e1ed15275a6761/includes/woopay/class-woopay-store-api-session-handler.php#L120).

> we will already have authenticated with the regular token and confirmed that the request came from WooPay

Got it, I was wondering about guest users who has a merchant site account but we already do a connection test to ensure the request comes from WooPay/WPCOM, I will update the code to be like that, the only difference is if it's a guest user, we will need to save the user email to the session during the redirection as it's not saved today.

alefesouza: Updated the PR to use a WooPay email flag instead of another cart token on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/ce78575f582990cde6d788be348d2834ce29d920.

includes/woopay/class-woopay-utilities.php
alefesouza: This PR will be merged with #6562 that uses the term `enabled_adapted_extensions`.

includes/woopay/class-woopay-session.php
alefesouza: > the feasibility of keeping the verified user ID as an extra session data for the current session instead of introducing a different cart token?

@ricardo This PR was updated to use a verified email address header from WooPay on https://github.com/Automattic/woocommerce-payments/commit/ce78575f582990cde6d788be348d2834ce29d920, similar to this approach.

includes/woopay/class-woopay-session.php
```diff
@@ -147,11 +147,11 @@ public static function get_user_id_from_cart_token() {
 
 			// If the email is verified on WooPay and matches session email,
 			// return the user to get extension data without authentication.
-			if ( ! empty( $_SERVER['HTTP_X_WOOPAY_VERIFIED_EMAIL_ADDRESS'] ) && ! empty( $session_data['woopay_email'] ) ) {
+			if ( ! empty( $_SERVER['HTTP_X_WOOPAY_VERIFIED_EMAIL_ADDRESS'] ) && ! empty( $customer['email'] ) ) {
```
asumaran: @alefesouza according to the comment above, which one is the "session email"? `$customer['email']` doesn't come from the WooPay session but the WC session right? Can we update the comment to clarify the intent of the code?

alefesouza: We are settings the session customer email before redirection to WooPay [here](https://github.com/Automattic/woocommerce-payments/pull/6559/commits/300714bd9617fd8f407b56604c5ed52bc281b1e8#diff-57ee486c4a3ba4bae478eb31255c9a968b5b4eccd04b27f512e4003440d6f821R1510), with this we can do the the email verification and also fix 1951-gh-Automattic/woopay, I agree the comment should be updated.

alefesouza:
The [latest WooCommerce version](https://github.com/woocommerce/woocommerce/pull/38983) always ask the user to login on the TYP if the order has a customer ID, on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/b82dac5e28ceee49df8fe60676aef0d59b640874 I've added a hook to remove the customer ID from the order if it's a verified email address request to skip the login screen, by default it's set to the current user ID that would be the merchant store user authenticated by the WooPay verified email request, before:

Screenshot 2023-07-13 at 11 06 46 PM

After:

Screenshot 2023-07-13 at 11 17 12 PM

As WooCommerce core checks [here](https://github.com/woocommerce/woocommerce/blob/6d597cc4cd1c9c7fcc081047a99c47b781ab0ff5/plugins/woocommerce/includes/shortcodes/class-wc-shortcode-checkout.php#L373), the guest user can view this page while the email is on the session (right after the order is successful) or if they login with the order billing email (their merchant site account).

includes/woopay/class-woopay-store-api-session-handler.php
```diff
@@ -74,6 +74,15 @@ protected function init_session_from_token() {
 		$this->_data              = (array) $this->get_session( $this->_customer_id, [] );
 	}
 
+	/**
+	 * Return true if the current user has an active session,.
+	 *
+	 * @return bool
+	 */
+	public function has_session() {
```
marcinbot: This looks like it's not used anywhere now.

alefesouza: I'm also working on Gift Cards extension compatibility on WooPay, currently when this extension is installed on the merchant site it breaks WooPay because it looks for the `has_session` method on the `WC()->session` that's not implemented on the WooPay session handler, I've preferred to implement it on this PR (the first one of the project) to stop breaking WooPay in case this PR is merged earlier.

Screenshot 2023-07-14 at 12 23 19

Screenshot 2023-07-14 at 12 23 41

marcinbot: Wouldn't it make sense to bundle it with the Gift Card compatibility PR? Otherwise we're shipping dead code. Not a big deal, but also not a great practice.

Git clients, especially the visual ones, should allow you to selectively exclude certain changes from commits if this is breaking your dev site and you don't want to toggle the breaking plugin.

alefesouza: Actually it's breaking the production site too, if the merchant store has the Gift Cards installed (and possibly other extensions that check for the `has_session` method), WooPay doesn't work because it's a method expected to be implemented in the WC session handler, should I continue and remove it to implement only on the Gift Card compatibility PR?

includes/woopay/class-woopay-session.php
ricardo: I like that we removed the extra cart token. 💯

I agree with Marcin's comment [here](https://github.com/Automattic/woopay/pull/1921#discussion_r1263653327). I think we could pass a boolean from WooPay if the email is verified or not, and handle the logic in the client.

includes/woopay/class-woopay-store-api-session-handler.php
marcinbot: > should I continue and remove it to implement only on the Gift Card compatibility PR?

I think that might be best, especially since the new PR has a chance of being deployed faster if it will contain only that change.

includes/woopay/class-woopay-session.php
```diff
@@ -113,7 +114,77 @@ public static function determine_current_user_for_woopay( $user ) {
...
+		$session_handler = new SessionHandler();
+		$session_data    = $session_handler->get_session( $payload->user_id );
+		$customer        = maybe_unserialize( $session_data['customer'] );
+
+			// If the token is already authenticated, return the customer ID.
+		if ( is_numeric( $customer['id'] ) && intval( $customer['id'] ) > 0 ) {
+			return intval( $customer['id'] );
+		}
+
+		$woopay_verified_email_address   = self::get_woopay_verified_email_address();
```
marcinbot, line 133: If we returned early here, with the check for `null === $woopay_verified_email_address` and then below with `$has_adapted_extension_installed` it would simplify the condition below and make the code easier to understand.

includes/woopay/class-woopay-store-api-session-handler.php
alefesouza: Removed on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/bfe862601c7ac6b788b4809200c1e4c1b5a83d33.

includes/woopay/class-woopay-session.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-payments/pull/6559/commits/111a0bda9911615e41d42930267811cf52ca0974.

PR #6571, opened 2023-06-22
Add connect page incentive

### LINKED ISSUE #6587: Add Action Incentive to Connect Page
### Description

As part of our Action Incentive project, we want to include merchants that just installed WCPay and offer them an incentive to fully get on board. We need to integrate the Action Incentive that was included in [the WC core](https://github.com/Automattic/woocommerce-payments-server/issues/3480) with the Connect page.

### Acceptance criteria

* The Action Incentive is displayed on the Connect page
* The Connect Page design is updated according to the new designs (Figma [EJyQdHyExlXuiniHADVXEC-fi-166%3A12793](https://href.li/?https://www.figma.com/file/EJyQdHyExlXuiniHADVXEC/?node-id=166%3A12793))

### Additional context

- paJDYF-96z-p2

ismaeldcom:
Fixes #6587 

- Implement `Incentives_Service` to request and cache eligible incentives from the server – key features:
  - Cache based on provided TTL if it exists.
  - Skip fetch if there is a connected account or the country is not supported.
  - Validates data structure before returning it.
  - Load incentive data into `wcpaySettings`.
  - Filters on `woocommerce_admin_allowed_promo_notes` to add the incentive ID to the allowed promo notes.
- Update SCSS mixins and `<TipBox>` component.
- Apply the new connect page design (Figma EJyQdHyExlXuiniHADVXEC-fi-166%3A12793)
- Implement `sanitizeHTML` util using `dompurify` from WC, to safely render the HTML from the server.
- Implement the incentive card on connect account page, rendering it only when there is one available. Using `description` and `tc_url` from the server.
- Update `handleSetup` to request the activation of the incentive promo ID before redirecting. And ensure every scenario (not supported country, incentive, and redirection) are properly covered. Catching any incentive error, to display an error message, but proceed with the redirection to avoid breaking onboarding if it fails.
- Add incentive ID to `CONNECT_ACCOUNT_VIEW` and `CONNECT_ACCOUNT_CLICKED` Track events.

##### Set up
- Install WC 7.9 or a build with https://github.com/woocommerce/woocommerce/pull/38689 included.
- Update the server to return type `connect_page` instead of `welcome_page` for the action incentive. And any other condition to be able to test the incentive implementation.
- Your store **should not** have an onboarded account.
- It's recommended to enable progressive onboarding, to avoid being redirected to Stripe KYC and require deleting the account on each attempt.

##### No incentive available
 - Go to **Payments**, the sidebar menu entry **should not** have a red badge.
 - The page should match the new design, except for the promotion card which shouldn't be there.
 - Clicking on **Finish setup** should proceed with the onboarding as usual.

##### Incentive available
- Got to **Payments (1)**, the sidebar menu entry **should** have a red badge.
- The page should match the new design.
- Clicking on **Finish setup** should proceed with the onboarding.
- DB `wc_admin_notes` table should contain a note with name `wcpay-promo-2023-action-discount`.

##### Tracks
- While having an incentive available.
- Got to **Payments (1)**
- Confirm that `page_view` event with `path: payments_connect_v2` contains `incentive_id: <id>`. 
- Confirm that `wcpay_connect_account_clicked` event contains `incentive_id: <id>`.

##### Cache
Confirm that the incentive data is cache for the given TTL or a day after a successful request, falling back to 6 hours on error.

#### Implementation screenshot
Screenshot 2023-06-23 at 09 08 26

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _QA Testing Not Applicable_ – Requires a future server implementation.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6571` or branch name `add/connect-incentive` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/connect-incentive) :rocket:

---

Build info: 

- Latest commit: 23a2c0f8e8d825c3f176a385a419b6d80c6c613f
- Build time: 2023-06-23 10:36:44 UTC

client/connect-account-page/strings.tsx
```diff
@@ -3,18 +3,35 @@
...
+		sprintf(
+			/* translators: %s: first name of the merchant, if it exists. */
+			__(
+				'Hi%s, run your business and manage your payments all in one place, with no setup costs or monthly fees.',
+				'woocommerce'
+			),
+			firstName ? ` ${ firstName }` : ''
+		),
+	usp1: __(
+		'Offer card payments,Apple Pay, Sofort, iDeal, Affirm, Afterpay, and accept in-person payments with the Woo mobile app.',
```
dmallory42: Typo: Should there be a space between `card payments,` and `Apple Pay`?

e.g. 
```
Offer card payments, Apple Pay, Sofort, iDeal, Affirm, Afterpay, and accept in-person payments with the Woo mobile app.
```

client/connect-account-page/utils.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
dmallory42: I noticed we have a top-level `/utils` folder,  I wonder if it would make sense to have this code live there, rather than in the `/connect-account-page` folder? 

I can see the logic of having it here if it only would be used in this context, so I leave it up to you. 🙂

client/globals.d.ts
```diff
@@ -85,6 +85,11 @@ declare global {
 		errorMessage: string;
 		onBoardingDisabled: boolean;
 		isBnplAffirmAfterpayEnabled: boolean;
+		connectIncentive?: {
+			id: string;
+			description: string;
+			tc_url: string;
```
dmallory42, line 91: What is `tc_url` here? Also, it looks like we use a mix of camel-case and snakeCase in this file, it's not important to this PR but should we consider standardising it?

dmallory42:
Minor nitpick: it looks like the font in the `Learn more about how to receive deposits` hyperlink is slightly larger than the rest of the text, and also IIRC this text should be highlighted/underlined to make it clear it's a link. I think we used purple previously, but I think blue like the other hyperlinks on the page would also make sense.

Screenshot 2023-06-22 at 16 12 48

dmallory42:
Just checking - the linked page for the TOS for the action incentive is currently not live: https://woocommerce.com/terms-conditions/woocommerce-payments-action-promotion-2023/

I know it's probably on your radar but just mentioning it to ensure it is updated before launch. 🙂

dmallory42 (commented):
The code is looking really good so far! I left a few nitpicks and questions, but nothing serious 🙂 I started testing too but encountered an error (mentioned on Slack), I will dig into it a bit more later/tomorrow morning but will leave half a review for now. 😄

elizaan36:
hey @ismaeldcom, The implementation screenshot looks great! 

Could you share what this layout would look like on mobile? Here's a screenshot for the mobile layout - the illustration is removed and everything wraps nicely.

Frame 1000004628

ismaeldcom:
Hi @elizaan36! I took the mobile screenshot but forgot to attach it too 😅 Here it is. BTW do we have a mobile version of them in Figma, I tried to look for them in this and previous designs but weren't able to. Maybe I'm missing them.

Could you also answer [this](https://github.com/Automattic/woocommerce-payments/pull/6571#issuecomment-1602819729)? I initially used the already implemented design with purple links. But decided to fully respect the design later, would be good to get some input from you 🙂 

Screenshot 2023-06-22 at 18 05 13

elizaan36:
Mobile layout looks great. I need to clean up the Figmas and include the mobile design alongside for reference. Will do asap.

> Minor nitpick: it looks like the font in the Learn more about how to receive deposits hyperlink is slightly larger than the rest of the text, and also IIRC this text should be highlighted/underlined to make it clear it's a link. I think we used purple previously, but I think blue like the other hyperlinks on the page would also make sense.

We have used purple for that link in the past but since we're making the update I think it should be the consistent blue with other links, however on the background color we need to darken the blue by about 10% to pass accessibility. It's in the design library as `Gutenberg Blue -10%` and the value is `0063A1`.

image

client/globals.d.ts
vladolaru: It is the promo/incentive dedicated Terms and Conditions (aka TC) page URL.

includes/class-wc-payments-incentives-service.php
```diff
@@ -0,0 +1,208 @@
...
+				// Whether the store has at least one payment gateway enabled.
+				'has_payments' => ! empty( WC()->payment_gateways()->get_available_payment_gateways() ),
+			],
+			'https://public-api.wordpress.com/wpcom/v2/wcpay/incentives',
+		);
+
+		$response = wp_remote_get(
+			$url,
+			[
+				'user-agent' => 'WCPay/' . WCPAY_VERSION_NUMBER . '; ' . get_bloginfo( 'url' ),
```
vladolaru, line 143: Excellent! This way, we can discern between WooCommerce and WCPay on the server.

includes/class-wc-payments-incentives-service.php
```diff
@@ -0,0 +1,208 @@
...
+						[
+							'status'       => [ 'wc-completed', 'wc-processing' ],
+							'date_created' => '>=' . strtotime( '-90 days' ),
+							'return'       => 'ids',
+							'limit'        => 1,
+						]
+					)
+				),
+				// Whether the store has at least one payment gateway enabled.
+				'has_payments' => ! empty( WC()->payment_gateways()->get_available_payment_gateways() ),
```
vladolaru, line 135: I wonder if we should be sending `has_wcpay` also with a trimmed-down logic since the plugin is obviously active. I am especially looking at the orders processed with WCPay signal. I guess we could deduce that on the server via transactions, but not if the merchant stopped using WCPay at some point.

dmallory42 (approved):
Tested and works well, great job @ismaeldcom! 

I know the tests aren't finished yet and there are a few small details, I'll pre-approve but can also re-test when it's ready. 🙂

includes/class-wc-payments-incentives-service.php
vladolaru: We should bear in mind that coming from WCPay, `has_payments` means that there is another payments gateway enabled, not WCPay, since if WCPay is not connected or payments enabled on the account, the payment gateway `is_available` [method](https://github.com/Automattic/woocommerce-payments/blob/0f03a5f3c5e497b6bf68a7201dd0e830f6881ff5/includes/class-wc-payment-gateway-wcpay.php#L488-L488) will return `false` and WCPay will not be returned by `get_available_payment_gateways`.

Just something for us to keep in mind so as not to assume this will always be `true` since WCPay is obviously active (that was my initial thought).

includes/class-wc-payments-incentives-service.php
```diff
@@ -0,0 +1,208 @@
...
+	public function fetch_connect_incentive(): ?array {
+		// Request incentive from WCPAY API.
+		$url = add_query_arg(
+			[
+				// Store ISO-2 country code, e.g. `US`.
+				'country'      => WC()->countries->get_base_country(),
+				// Store locale, e.g. `en_US`.
+				'locale'       => get_locale(),
+				// WCPay active for duration in seconds.
+				'active_for'   => time() - get_option( 'wcpay_activation_timestamp', time() ),
```
vladolaru: We should keep the meaning of `active_for` to be store lifetime (the same way as in WC core), not since WCPay activation. We would complicate the server logic with different meanings behind these signals.

If we want to know how long WCPay has been active we can add a new query arg (like `wcpay_active_for`). The `/incentives` endpoint will decide if it uses it or not.

includes/class-wc-payments-incentives-service.php
```diff
@@ -0,0 +1,208 @@
...
+
+	/**
+	 * Class constructor
+	 *
+	 * @param Database_Cache $database_cache      Database cache util.
+	 */
+	public function __construct( Database_Cache $database_cache ) {
+		$this->database_cache = $database_cache;
+
+		add_action( 'wcpay_js_settings', [ $this, 'add_js_settings' ] );
```
vladolaru: This should be `add_filter`. It works because `add_action` is actually `add_filter` under the hood, but it's best to keep patterns for readability and clarity of intent.

client/connect-account-page/strings.tsx
ismaeldcom: Good catch! Addressed in 36f9c84.

client/connect-account-page/utils.ts
ismaeldcom: Yes, I put it there because it's only used by connect account page. But we could definitely use it in other places too. So I'll move it 🙂 Addressed in f8ae56a.

client/globals.d.ts
ismaeldcom: We were mapping snake case to camel case before, but decided to update ESLint rules to allow the former, because that's what's coming from the server.

I prefer to be consistent, but I think it will need to automatically map every prop we receive and send to the server to ensure we don't mix them by mistake.

includes/class-wc-payments-incentives-service.php
ismaeldcom: 💯 Not sure why I used `add_action` here, probably was just a copy/paste mistake. Good catch! Although as we mentioned in Slack, we'll not add the filter in this PR.

includes/class-wc-payments-incentives-service.php
ismaeldcom: For this case we don't actually need that information. But as the name is just `active_for` and we will act differently based on the `caller`, I don't see a need for another param.

A similar case to `has_wcpay`, we can treat them differently based on the `caller` or add new params. They sound generic enough to me to use the same.

includes/class-wc-payments-incentives-service.php
ismaeldcom: >I wonder if we should be sending has_wcpay also with a trimmed-down logic since the plugin is obviously active.

This is a similar case to `active_for`, I didn't want to change the meaning of `has_wcpay`. I think we have pretty clear that it's true by just having WCPay installed.

>I am especially looking at the orders processed with WCPay signal. I guess we could deduce that on the server via transactions.

I don't see how it can happen, given the client conditions, every call to the endpoint will be without a WCPay account, so no way for use to check anything about a previous account.

>We should bear in mind that coming from WCPay, has_payments...

Yep, I'm aware of it, that's why I kept it as it is. Because we want to know if there is any other payment method, except for WCPay.

client/globals.d.ts
dmallory42: Ah yes, I understand the mapping makes it difficult. Thanks for explaining!

includes/class-wc-payments-incentives-service.php
ismaeldcom: To be able to merge it already, I reverted `active_for` to WC time. As we don't need it for the eligibility in this case, we can decide later about how to treat it 🙂 And pretty much the same applies to `has_wcpay`.

vladolaru: Excellent! We may never know when we want to take into account the store age, especially with the shifting requirements.

includes/class-wc-payments-incentives-service.php
```diff
@@ -153,15 +153,17 @@ public function fetch_connect_incentive(): ?array {
 
 		if ( 200 === wp_remote_retrieve_response_code( $response ) ) {
 			// Decode the results, falling back to an empty array.
-			$results = json_decode( wp_remote_retrieve_body( $response ), true ) ?? [];
+			$results = json_decode( wp_remote_retrieve_body( $response ), true );
```
dmallory42, line 156: I noticed we added the new `if( ! empty( $results ) )` check, but what gets returned by the function in that case? It may be worth explicitly returning early or something similar in the case that the result isn't what we expect.

dmallory42 (approved):
LGTM :shipit: 

One small nitpick but this is good to go!

includes/class-wc-payments-incentives-service.php
ismaeldcom: We will return the initial value `$incentive = []` plus the `ttl` logic below, so it can be cached for the given amount of time. Am I missing something? 🤔

dmallory42: Chatted about it over Slack, just wanted to confirm that response was expected. 🙂

PR #6582, opened 2023-06-23
Tool to update Multi-Currency meta data

jessepearson:
Fixes https://github.com/Automattic/woocommerce-payments-server/issues/3431

This PR adds a tool that, when enabled, will allow a store admin to modify/add Multi-Currency exchange rate meta data in an order. This helps to fix issues where we've seen this data missing on orders, which causes incorrect values in Analytics reports.

* Add new `MultiCurrency/Helpers` directory for this and any additional helpers.
* Add new `OrderMetaHelper` class for the tool.
* Add `OrderMetaHelper.md` to describe usage of the tool.
* Added new flag and method `WC_Payments_Features` to allow for enabling and disabling the helper.

The helper itself adds a new meta box to the order. In that meta box it displays information about the store, order, and intent:

* Store currency
* Order currency
* Order default currency (store currency at the time of order)
* Payment method ID
* Payment method title
* Intent ID
* Intent currency (from order meta)
* Multi-Currency exchange rate
* Stripe exchange rate (from order meta)
* Charge exchange rate (from intent)
* Charge currency (from intent)

The list is extensive to try to provide a full picture to the merchant, but it can be narrowed down easily.

A single field is added in the meta box to allow the merchant to enter a new exchange rate for the order. Once the field is populated and the order is saved, the meta data is added/updated. An order note is created under the user making the change, and new meta data is added in order to log the changes.

The meta box will not show on orders in the default currency, however, it will show for any order that is _not_ through WooCommerce Payments. This is due to Multi-Currency works for WooCommerce and only relies on WCPay for getting the automatic rates. Customers are able to purchase items with any gateway if the feature is enabled. For this reason, we only update one of the two meta values that hold exchange rates.

#### Testing instructions

Setup:
* Go to WooCommerce > Settings > Multi-Currency and enable at least one other currency in your store.

Testing:
* Purchase a product in another currency other than the default currency in your store.
* Go to the order under WooCommerce > Orders in the admin.
* Confirm there is not a meta box named _Multi-Currency Meta Helper_ showing.
* Go to WooCommerce > Status > Tools
* Locate the _Enable/Disable Multi-Currency Meta Helper_ row and click the _Enable/Disable_ button to enable the tool.
* Confirm the "Multi-Currency Meta Helper has been enabled" message shows.
* Return to your new order and confirm that the meta box named _Multi-Currency Meta Helper_ now shows.
* Delete both the `_wcpay_multi_currency_stripe_exchange_rate` and `_wcpay_multi_currency_order_exchange_rate` meta items under the order.
  * This can be done by adding this snippet somewhere, which enables viewing hidden meta data: `add_filter( 'is_protected_meta', '__return_false', 999 );`, or
  * This can be done in the `wp_postmeta` table by searching for your order number under `post_id`.
* Refresh the order.
* The meta box should now show:
  *  _Multi-Currency Exchange Rate_ is _Not found_.
  *  _Stripe Exchange Rate_ is _Not found_.
  *  If the intent is valid, there should be a _Suggested exchange rate_ with the _Charge exchange rate_ value shown.
* Enter text into the field and hit enter/return or click the _Update_ button for the order.
* An error should show that non-numeric information was passed: _Error: Exchange rate value not numeric: xxx_
* Copy and paste the suggest rate into the box and hit enter/return or click the _Update_ button for the order.
* A new note should show in the order notes under your user ID.
* The note should state the exchange rate was updated and include the from and to rates.
* The _Multi-Currency Exchange Rate_ value should now be populated in the meta box, the _Stripe Exchange Rate_ value will not be.
* A new meta entry under the order with the `meta_key` will be added. It will be a multidimensional array the key being the timestamp, and the value being an array with your user ID, the old rate, and the new rate. Each update to the exchange rate will add another entry in this log.
  * This log will need to be viewed directly in the database, or with a plugin like [Post Meta Inspector](https://wordpress.org/plugins/post-meta-inspector/).
* Place an order for a product using the default store currency.
* View the order under WooCommerce > Orders in the admin.
* The meta box should not show.
* Go back to WooCommerce > Status > Tools.
* Locate the _Enable/Disable Multi-Currency Meta Helper_ row and click the _Enable/Disable_ button to disable the tool.
* Confirm the "Multi-Currency Meta Helper has been disabled" message shows.
* Return to the order where you updated the exchange rate.
* Confirm the meta box no longer shows.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6582` or branch name `add/server-3431-mc-tool-for-meta-conversion-correction` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/server-3431-mc-tool-for-meta-conversion-correction) :rocket:

---

Build info: 

- Latest commit: 1fe6f67b196fdbd5c155635e590894efcd6d19c4
- Build time: 2023-07-05 11:31:11 UTC

allie500 (approved):
Thanks for working on this tool. I ran through the manual testing and everything passed.

| **Test** | **Result** |
| ---- | ------ |
| No meta box showing if tool is disabled | ✅ |
| Enabled message shows when enabling tool | ✅ |
| Meta box shows if tool is enabled | ✅ |
| Exchange rates show Not found if meta is missing | ✅ |
| Error shown if non-numeric text used | ✅ |
| New order note added when adding exchange rate | ✅ |
| MC exchange rate updated after adding exchange rate | ✅ |
| Stripe exchange rate not updated after adding exchange rate | ✅ |
| MC exchange rate log added to order post meta in DB after adding exchange rate | ✅ |
| Meta box doesn't show for orders in store currency | ✅ |
| Disabled message shows when disabling tool | ✅ |
| Meta box doesn't show on previous non-store currency orders after tool disabled | ✅ |

I have two non-blocking items I wanted to get your feedback on. 

First, in the `OrderMetaHelper.md` file it is mentioned that the merchant would need to run the _Import historical data_ tool under Analytics > Settings to correct data related to any orders that have been adjusted. Are we notifying the merchant about this anywhere? I didn't catch it when I ran through the tests. Perhaps we should update the notice that is shown when they update the exchange rate using the tool in the order. What do you think?

Second, is there an edge case where the order could be missing all three exchange rate data points: `mc_exchange_rate`, `charge_exchange_rate`, and `stripe_exchange_rate`? Looking at the code if that scenario happened, we'd have `$suggested` still equal to `false`. This would result in the suggested row not showing in the box at all but the exchange rates would both say 'Not found'. I think it makes perfect sense to not show the suggested row when we have a value for `mc_exchange_rate` since they don't need to update the exchange rate. But in a case where they have no exchange rate data and we can't even provide them with a suggested rate due to the charge and Stripe exchange rates being unavailable, should we put a short notice to this effect in the suggested rate row in this case? Or is it pretty much impossible for this scenario to occur? 

I don't think either of these are blockers. Tests pass so I think we can merge this, but I'd be interested to here your thoughts on it. If you think one or both are easily actionable feel free to add to this PR and I can re-review later. Or we could just open separate issue(s) if needed.

includes/multi-currency/Helpers/OrderMetaHelper.php
```diff
@@ -0,0 +1,443 @@
...
+	/**
+	 * Updates the exchange rate meta data.
+	 *
+	 * @param int $order_id The order we are working with.
+	 *
+	 * @return void
+	 */
+	public function maybe_update_exchange_rate( $order_id ) {
+		// Verify nonce.
+		$nonce_value = ! empty( $_POST['wcpay_multi_currency_exchange_rate_nonce'] ) ? sanitize_text_field( wp_unslash( $_POST['wcpay_multi_currency_exchange_rate_nonce'] ) ) : false;
```
tpaksu, line 108: Maybe we should also add a role/permission check here? To prevent unauthorized users to change rates?

includes/multi-currency/Helpers/OrderMetaHelper.php
```diff
@@ -0,0 +1,443 @@
...
+	 *
+	 * @return array The modified tools array.
+	 */
+	public function add_tool_button( $tools ) {
+		/**
+		 * We are not able to look at the current status of the tool to dynamically change these labels. The tool runs after the labels are created.
+		 */
+		$tools['wcpay_mc_meta_helper_toggle'] = [
+			'name'     => __( 'Enable/Disable Multi-Currency Meta Helper', 'woocommerce-payments' ),
+			'button'   => __( 'Enable/Disable', 'woocommerce-payments' ),
```
tpaksu, line 57: I think we can conditionally change the button text to show Enable and Disable depending on the tool status. This way it's not clear if the tool is enabled or not.

includes/multi-currency/Helpers/OrderMetaHelper.php
```diff
@@ -0,0 +1,443 @@
...
+				'label' => __( 'Payment Method Title', 'woocommerce-payments' ),
+				'value' => $order->get_payment_method_title(),
+			],
+			'intent_id'              => [
+				'label' => __( 'Intent ID', 'woocommerce-payments' ),
+				'value' => $intent_id,
+			],
+			'intent_currency'        => [
+				'label' => __( 'Intent Currency', 'woocommerce-payments' ),
+				'value' => $order->get_meta( '_wcpay_intent_currency' ),
```
tpaksu: Can we convert this to uppercase to make it look like the rest of the currencies?

tpaksu (approved):
All items in the testing instructions worked as expected, left a couple of cosmetic issues, otherwise it's good to 🚢 . Do you think it'll affect analytics or COT when the rate is changed on the order's post meta? Also, I've entered "0" to the input and it didn't produce an error, but it also didn't save it.

jessepearson:
> First, in the OrderMetaHelper.md file it is mentioned that the merchant would need to run the Import historical data tool under Analytics > Settings to correct data related to any orders that have been adjusted. Are we notifying the merchant about this anywhere? I didn't catch it when I ran through the tests. Perhaps we should update the notice that is shown when they update the exchange rate using the tool in the order. What do you think?

I thought about that, as well. I didn't add a notice in order to keep it as simple as possible. It's likely that merchants will only be using this _after_ contacting support about the issue, so support can let them know what to do. Also, although it's not likely to change, it's possible that WC Admin could change where that tool is, how it operates, etc, and that would make anything said in the success message obsolete.

> Second, is there an edge case where the order could be missing all three exchange rate data points: mc_exchange_rate, charge_exchange_rate, and stripe_exchange_rate? Looking at the code if that scenario happened, we'd have $suggested still equal to false. This would result in the suggested row not showing in the box at all but the exchange rates would both say 'Not found'. I think it makes perfect sense to not show the suggested row when we have a value for mc_exchange_rate since they don't need to update the exchange rate. But in a case where they have no exchange rate data and we can't even provide them with a suggested rate due to the charge and Stripe exchange rates being unavailable, should we put a short notice to this effect in the suggested rate row in this case? Or is it pretty much impossible for this scenario to occur?

MC itself works for all gateways, so it's possible to have nothing for the tool to suggest from Stripe. We also are exploring adding other gateways into WCPay, so that data wouldn't be available for say PayPal if it were added. Again, trying to keep it more simple to help with any future issues.

Thank you for the review and comments :D

allie500:
> Thank you for the review and comments :D

You're welcome! Thanks for the feedback on my comments and questions. That makes sense and all sounds good to me.

includes/multi-currency/Helpers/OrderMetaHelper.php
jessepearson: Good catch on that. I didn't think to add it because the form and actions only work inside of the order page editing context, but better safe than sorry. Added in: https://github.com/Automattic/woocommerce-payments/pull/6582/commits/710beecd504906ed5ff896a55012d58cda775092

includes/multi-currency/Helpers/OrderMetaHelper.php
jessepearson: It's possible to check for the the status of the feature and to change the message and button, but it doesn't work immediately. The tool array containing the information is built before the action runs to run the tool itself. This means that if you click the button to enable the tool, it will be enabled, but the description and button will say it's disabled. If you manually refresh the page after that, it will show correctly.

I initially created it to work as you're talking about, but was disappointed when it didn't work as expected.

includes/multi-currency/Helpers/OrderMetaHelper.php
jessepearson: I went through and set them all to be uppercase, just to be sure. This is in https://github.com/Automattic/woocommerce-payments/pull/6582/commits/710beecd504906ed5ff896a55012d58cda775092

In my testing I never noticed any of them being lowercase, but then I did a database search and found that subscription renewals have the `_wcpay_intent_currency` as lowercase. I wonder if this affects anything else elsewhere.

jessepearson:
> Do you think it'll affect analytics or COT when the rate is changed on the order's post meta? Also, I've entered "0" to the input and it didn't produce an error, but it also didn't save it.

Analytics is affected, yes, but the merchant would need to re-run the task that builds that data. I am not sure what would happen with COT. AFIAK this has been an edge case with the current standard, so I didn't do any testing for the new setup.

0 shouldn't be allowed, I have it set to anything greater than 0.

jessepearson:
Also @tpaksu I know you said this was good to go, but I wanted your input on everything I added before merging, so I requested a review again.

includes/multi-currency/Helpers/OrderMetaHelper.php
tpaksu: Is it cached somehow? Maybe we can add a status indicator on the name prop?

tpaksu:
All changes in [710beec](https://github.com/Automattic/woocommerce-payments/pull/6582/commits/710beecd504906ed5ff896a55012d58cda775092) makes sense. Thanks for handling those. Still good to go.

jessepearson:
@dwainm Any input before this is merged?

Issue #6576, opened 2023-06-23
Legacy and Split UPE flows: setup intent ID is not attached to order meta during process_payment


htdat:
### Describe the bug

Similar to #6575, but intent_id order meta is not attached to order during process_payment. I am just separating the issue. 

### To Reproduce

Similar to #6575, purchase a subscription with a free trial to see this issue.

0. Enable legacy or split UPE flow.
1. Use WCPay Subscription and create a subscription product with free trial.
2. Purchase this product, and check out with this 3DS card `4000 0027 6000 3184`.
3. DO NOT authorize this payment in the pop-up iframe.
4. Check order meta of the newly created order in WP Admin > WooCommerce > Orders.
5. Issue: this setup intent ID is not saved in the order meta `_intent_id`.

### Actual behavior

intent_id for setup intents would be attached in the `process_payment`, that's similar to payment intents in these UPE flows, as well as setup intents in the traditional (card) flow.

### Additional context

On-going discussion: p1687513417167319-slack-C05AW31EKFF

timur27:
I think we need to treat this issue blocked by https://github.com/Automattic/woocommerce-payments/issues/6575. Once the blocker is completed, we'll have the updated setup intent and need to attach the updated data to the order.

[This code](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L557-L563) should accomplish all the goals of this issue. However, it's placed inside [the payment intent processing flow](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L475) and thus won't cover set up intent processing. It might be sufficient to move it outside of 

```php
if ( $payment_needed ) {
...
} else {
... 
} 
```
block. Once placed after this block, the order should be updated with all the latest intent details, regardless of intent type.

FangedParakeet:
[...]

PR #6586, opened 2023-06-26
Refactor: introduce WC_Payments_API_Setup_Intention for setup intents

### LINKED ISSUE #2422: Refactor WC_Payments_API_Intention to support Setup Intents

htdat:
Fixes #2422

- Create a new abstraction class `WC_Payments_API_Abstract_Intention`, and `WC_Payments_API_Setup_Intention` for setup intents to extend on this abstract class.
- Update the old payment intent class to `WC_Payments_API_Payment_Intention`, that now extends the abstract class. Move all shared getters and properties from this payment intent class to the abstract class.
- Fix test failures.
- For `WC_Payments_API_Abstract_Intention`, I also introduced the `get_payment_method_type` to assist the process of extracting the payment method type (`card`, `sofort`, `card_present`, etc) easily.

#### Testing instructions

Basically, this refactor will touch to multiple flows, and ideally, reviewers/testers would test [all critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows) regarding the payment process in the checkout or pay-for-order pages. That would include: 

- Flows: traditional (card), UPE, split-UPE, deferred UPE.
- Subscription products: free-trial (zero-amount orders), and normal (total > 0) orders.
- Checkout pages: classic and block.

*

-------------------

- [x] NON-RELEVANT Tested on mobile (or does not apply)

**Post merge**

- [X] NO NEED - USING THE CURRENT CRITICAL FLOWS IS ENOUGH - Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [X] NO NEED - USING THE CURRENT CRITICAL FLOWS IS ENOUGH _ Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [X] NO NEED Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.
- [ ] Monitor GlobalStep testings, especially check PHP error log wp-content/debug.log from their test sites.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6586` or branch name `refactor/2422-object-support-setup-intents` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=refactor/2422-object-support-setup-intents) :rocket:

---

Build info: 

- Latest commit: 473037d9ac172f3a1776d02bc427ed48c7c124d7
- Build time: 2023-07-21 15:39:09 UTC

htdat:
Ready to review!

includes/constants/class-setup-intent-status.php
```diff
@@ -0,0 +1,27 @@
...
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * This class gives a list of all Setup Intent status name constants.
+ * ref: https://stripe.com/docs/api/setup_intents/object#setup_intent_object-status
+ *
+ * @psalm-immutable
+ */
+class Setup_Intent_Status extends Base_Constant {
```
RadoslavGeorgiev: AFAICT the only _payment_-intent-specific status is `requires_capture`, which has very specific handling. With that in mind, would you consider using a single `Intent_Status` ENUM throughout the gateway? This way, most code would be agnostic to the type of intent being used. 

Also, right now, the gateway uses `WC_Payment_Gateway_WCPay::SUCCESSFUL_INTENT_STATUS` regardless.

bin/run-tests.sh
```diff
@@ -13,10 +13,10 @@ while getopts ':w' OPTION; do
...
 done
 
-echo "Installing the test environment..."
-
-docker-compose exec -u www-data wordpress \
-	/var/www/html/wp-content/plugins/woocommerce-payments/bin/install-wp-tests.sh
+#echo "Installing the test environment..."
+#
+#docker-compose exec -u www-data wordpress \
+#	/var/www/html/wp-content/plugins/woocommerce-payments/bin/install-wp-tests.sh
```
RadoslavGeorgiev: Is this intentionally left commented?

includes/core/server/request/class-create-and-confirm-setup-intention.php
```diff
@@ -83,4 +84,14 @@ public function set_payment_method( string $payment_method_id ) {
...
 	}
+
+	/**
+	 * Formats the response from the server.
+	 *
+	 * @param  mixed $response The response from `WC_Payments_API_Client::request`.
+	 * @return mixed           Either the same response, or the correct object.
+	 */
+	public function format_response( $response ) {
+		return WC_Payments::get_payments_api_client()->deserialize_setup_intention_object_from_array( $response );
```
RadoslavGeorgiev: I'm not sure how this managed to slip through the cracks previously, but the dependency on `WC_Payments::get_payments_api_client()` is not needed. The `Request` class already contains the API client, so a simple call to the dependency should be better.

Could you please change that in all requests? If not within this PR, let's open an issue for next sprint to fix that.

includes/class-wc-payments-order-service.php
```diff
@@ -1438,32 +1438,25 @@ private function update_order_status( $order, $order_status, $intent_id = '' ) {
...
+
+		$intent_data = [
+			'intent_id'           => $intent->get_id(),
+			'intent_status'       => $intent->get_status(),
+			'charge_id'           => '',
+			'fraud_outcome'       => $intent->get_metadata()['fraud_outcome'] ?? '',
+			'payment_method_type' => $intent->get_payment_method_type(),
+		];
+
+		if ( is_a( $intent, WC_Payments_API_Payment_Intention::class ) ) {
```
RadoslavGeorgiev:
```suggestion
		if ( $intent instanceof WC_Payments_API_Payment_Intention ) ) {
```

Let's use `instanceof` instead of `is_a()`. It's functionally the same, but VS Code understands it a bit better. 
image
image

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2121,7 +2122,7 @@ private function add_order_info_to_charge_object( $charge_id, $object ) {
 	private function add_order_info_to_object( $order, $object ) {
 		// Add order information to the `$transaction`.
 		// If the order couldn't be retrieved, return an empty order.
-		$object['order'] = null;
+		$object['order'] = [];
```
RadoslavGeorgiev, line 2021: Could you provide some reasoning here? Why is an empty order needed instead of just `null`? I [see the linked issue](https://github.com/Automattic/woocommerce-payments/pull/6586/commits/941bf3970c66c8080f2621080f3df2e28c002b5a), but I don't understand how having an array is better.

RadoslavGeorgiev (dismissed):
The changes look great, @htdat! 👏 

I've got a couple of minor comments, but the rest of the code looks excellent. I haven't performed thorough manual testing yet, as there are quite a few scenarios to test, but nothing looks dangerous right now.

includes/core/server/request/class-create-and-confirm-setup-intention.php
htdat: Good point! Updated b0130ee2cdf0e2a6aeddd587f4d5007ed58d5e94

bin/run-tests.sh
htdat: Good catch! Revert 3db3b12ce504a25e7efa4d8a391c15185fa2e2e6

includes/class-wc-payments-order-service.php
htdat: PHPStorm seems smarter :D 

Markup on 2023-07-14 at 14:21:39

Anyway, I've made the change: c9c09c1461876f51f81a4a54925b11b9455f47a2

includes/constants/class-setup-intent-status.php
htdat: Good point. Did that here 1f10437ea85e15ee4cabcbb590a7ca6b635f8819 and created this issue to refactor to `Intent_Status` later https://github.com/Automattic/woocommerce-payments/issues/6773

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: @RadoslavGeorgiev  - Good point, I would be more verbose for this simple change. 

First, if testing this PR with replicate steps in #6576, here is the error:

<details>

<summary>Click to see the full error with Stack trace</summary>

```
[14-Jul-2023 08:25:03 UTC] PHP Fatal error:  Uncaught TypeError: Argument 12 passed to WC_Payments_API_Setup_Intention::__construct() must be of the type array, null given, called in /var/www/html/wp-content/plugins/woocommerce-payments/includes/wc-payment-api/class-wc-payments-api-client.php on line 2313 and defined in /var/www/html/wp-content/plugins/woocommerce-payments/includes/wc-payment-api/models/class-wc-payments-api-setup-intention.php:38
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-payments/includes/wc-payment-api/class-wc-payments-api-client.php(2313): WC_Payments_API_Setup_Intention->__construct('seti_1NThJDR67t...', 'cus_NczxswVwvaT...', NULL, Object(DateTime), 'requires_paymen...', 'seti_1NThJDR67t...', Array, Array, Array, Array, Array, NULL)
#1 /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/request/class-create-setup-intention.php(76): WC_Payments_API_Client->deserialize_setup_intention_object_from_array(Array)
#2 /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/class- in /var/www/html/wp-content/plugins/woocommerce-payments/includes/wc-payment-api/models/class-wc-payments-api-setup-intention.php on line 38
```

</details>

It's a chain of actions to see an error like that. I also it's not an intention to give a `null` here so changing it to an array makes a lot of sense with the comment too: 

1, Method `deserialize_payment_intention_object_from_array` calls this https://github.com/Automattic/woocommerce-payments/blob/e3f1820ac05b9932a94ce1900d070363078c7eff/includes/wc-payment-api/class-wc-payments-api-client.php#L2310

2, Due to the bug #6576, `$order` here is returned `false`, i.e. no order with the provided setup intent ID. https://github.com/Automattic/woocommerce-payments/blob/e3f1820ac05b9932a94ce1900d070363078c7eff/includes/wc-payment-api/class-wc-payments-api-client.php#L2092-L2096

3, In the next iteration, `$object['order']` will be `null` in `add_order_info_to_object`. 

Then, finally the `$order` is used in step 1 above is `null`, not an array as expected.

htdat:
@Automattic/gamma - ready to be reviewed again after addressing feedback from @RadoslavGeorgiev.

At the moment, `E2E test failure regarding dispute` and `PHP test failure regarding woopay request` are known issues p1689305403447819-slack-CGGCLBN58

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1182,7 +1183,8 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 				// If the setup intent is included in the request use that intent.
 				$intent = $this->payments_api_client->get_setup_intent( $woopay_intent_id );
 
-				$intent_meta_order_id_raw = ! empty( $intent['metadata'] ) ? $intent['metadata']['order_id'] ?? '' : '';
+				$intent_metadata          = $intent->get_metadata();
+				$intent_meta_order_id_raw = ! empty( $intent_metadata ) ? $intent_metadata['order_id'] ?? '' : '';
```
zmaglica: Nitpick. You could simplify this line by doing only $intent_metadata['order_id'] ?? ''. You don't have to do an empty check since the null coalescing operator will do that for you and you are already doing it.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2229,31 +2230,37 @@ private function deserialize_charge_object_from_array( array $charge_array ) {
...
-		$processing           = $intention_array[ Payment_Intent_Status::PROCESSING ] ?? [];
-		$payment_method_types = $intention_array['payment_method_types'] ?? [];
+		// Metadata can be an empty stdClass object, so we need to check array type too.
+		// See https://github.com/Automattic/woocommerce-payments/pull/419/commits/c2c8438c3ed7be6d604435e059209fb87fb6d0c4.
+		$raw_metadata           = $intention_array['metadata'];
+		$metadata               = is_array( $raw_metadata ) && ! empty( $raw_metadata )
+			? $raw_metadata
+			: [];
+		$charge_array           = 0 < $intention_array['charges']['total_count'] ? end( $intention_array['charges']['data'] ) : null;
+		$next_action            = ! empty( $intention_array['next_action'] ) ? $intention_array['next_action'] : [];
```
zmaglica: Can you use `$intention_array['next_action'] ?? []` here? It is a small nitpick but makes the code a bit easier to read.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -2268,6 +2275,52 @@ public function deserialize_intention_object_from_array( array $intention_array
...
+		$created = new DateTime();
+		$created->setTimestamp( $intention_array['created'] );
+
+		// Metadata can be an empty stdClass object, so we need to check array type too.
+		// See https://github.com/Automattic/woocommerce-payments/pull/419/commits/c2c8438c3ed7be6d604435e059209fb87fb6d0c4.
+		$raw_metadata           = $intention_array['metadata'];
+		$metadata               = is_array( $raw_metadata ) && ! empty( $raw_metadata )
+			? $raw_metadata
+			: [];
+		$next_action            = ! empty( $intention_array['next_action'] ) ? $intention_array['next_action'] : [];
```
zmaglica: Same comment as above regarding the null coalescing operator for this type of checks

zmaglica (changes requested):
Left a couple of nitpicks. The overall code looks good.

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: Addressed fe80244

includes/wc-payment-api/class-wc-payments-api-client.php
htdat: Addressed fe80244

includes/class-wc-payment-gateway-wcpay.php
htdat: Addressed fe80244

htdat:
👋 @zmaglica Ready to review again. Major changes from your last review: 

- fe8024495 Address your feedback, use null coalescing
- f98a63183 This merge includes your PR 6f71c55cb, which also touches `get_setup_intent`. Therefore, I needed to implement the following changes. 
- 1103024c6 Add Get_Setup_Intention request class, and take advantage of the new model class `WC_Payments_API_Setup_Intention`. 
- 05956b9f3 Replace Request::get for get_setup_intent with Get_Setup_Intention
- 161608654 Fixing tests

includes/core/server/request/class-get-setup-intention.php
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Class constructor.
+	 *
+	 * @param WC_Payments_API_Client     $api_client Api client.
+	 * @param WC_Payments_Http_Interface $http_interface Http interface.
+	 * @param string                     $setup_intent_id Setup intent id.
+	 *
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function __construct( WC_Payments_API_Client $api_client, WC_Payments_Http_Interface $http_interface, string $setup_intent_id ) {
```
zmaglica: We already support id argument via the default create method:

```php
public static function create( $id = null ) {
		return WC_Payments::create_request( static::class, $id );
}
```

and default constructor looks like this:

```php
	public function __construct( WC_Payments_API_Client $api_client, WC_Payments_Http_Interface $http_interface, $id = null ) {
		$this->api_client     = $api_client;
		$this->http_interface = $http_interface;

		if ( method_exists( $this, 'set_id' ) ) {
			if ( null !== $id ) {
				$this->set_id( $id );
			} else {
				throw new Invalid_Request_Parameter_Exception( 'This request requires an item ID.', 'wcpay_core_invalid_request_parameter_missing_id' );
			}
		}
	}
```

In other words, your request class only needs to define `set_id` function where you can validate id by calling `validate_stripe_id`. You don't have to override the default constructor.

includes/core/server/request/class-get-setup-intention.php
```diff
@@ -0,0 +1,67 @@
...
+	 */
+	public function get_method(): string {
+		return 'GET';
+	}
+
+	/**
+	 * Formats the response from the server.
+	 *
+	 * @param  mixed $response The response from `WC_Payments_API_Client::request`.
+	 * @return mixed           Either the same response, or the correct object.
```
zmaglica: You can say that the return type will be `WC_Payments_API_Setup_Intention` and get all typrhints and other goodies.

htdat:
@zmaglica - Thanks for the last feedback. Addressed in 54835b1a6c7ece6ed05dfa3036e324923b5d48b5

Ready to review again.

zmaglica (approved):
LGTM! Good job on converting some of the requests to the standalone request classes.

PR #6596, opened 2023-06-27
Add - Tooltip and ARIA label to PM logos in Translation list and Payment Details page

### LINKED ISSUE #6589: Add tooltip and ARIA labels to PM logos in transaction list page
### Description

The Transactions list page currently shows only a logo of the Payment Method. For some of the Payment methods the icon itself may make it difficult to identify the PM name. It is good to have a tooltip show the name of the Payment method. Adding ARIA label tags will also improve accessibility / usage over screenreaders. 

### Dev notes
Identified in https://github.com/Automattic/woocommerce-payments/issues/6563

### Context

The issue was [discussed here](p1687401900469979/1685696069.251879-slack-C04VCT34PCP) and was supposed to be added along with the issue above. It could not be merged on time due to code freeze and team meetup. Hence created a separate issue for this.

nagpai:
Fixes #6589 

Add a tooltip indicating the brand name / payment method title to the small logo ( usually only symbol ) that appears on the Transactions list page ( Payments > Transactions ). 

The PR also adds an ARIA label for better accessibility. 

<details><summary>Desktop Screenshots</summary>

image

image

</details> 

----

<details><summary>Mobile Screenshot - Tooltips not shown </summary>

image
</details> 

#### Testing instructions
* Link your local site to an account that preferably has transactions from a variety of test payment methods. 
* Browse to `Payments > Transactions` and make sure you see the `Source` column. 
* Move your mouse over the source logo, you should see a black tooltip with the name of the payment method, similar to the screenshot above. 

#### Mobile view testing
* Open the transactions page in Chrome and use the mobile view option in browser tools: 

image

* Browse to the `Payments > Transactions` page to see the transaction list. "touch" (move the mouse over, you will see a circle indicating a simulated mobile touch. click and drag that) and scroll right to see the logos column. When you touch the logo, it should not show a tooltip, rather should open up the payment details page
* On the payment details page, "touching" the logo should not bring up the tooltip. 

-------------------

- [x] Tested on mobile (or does not apply) - tested on browser simulation

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6596` or branch name `update/6589-tooltip-transaction-logos` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/6589-tooltip-transaction-logos) :rocket:

---

Build info: 

- Latest commit: b4b74a669f9279c3d5759812197ce9c52c7a1344
- Build time: 2023-07-05 10:12:06 UTC

client/components/payment-method-details/index.js
```diff
@@ -6,6 +6,8 @@
 import { Fragment } from 'react';
 import './style.scss';
 import p24BankList from '../../payment-details/payment-method/p24/bank-list';
+import { HoverTooltip } from '../tooltip';
+import { PAYMENT_METHOD_TITLES } from 'payment-methods/constants';
```
dmvrtx, line 10: ~~We already have `PaymentMethodInformationObject` which contains translated title, accessible by payment method id. Can it be used here instead of a new constant?~~

I see that here we also use a split by payment method brands, which makes things more complex. Anyways, can we at least use translations inside the `PAYMENT_METHOD_TITLES` constant?

dmvrtx: On a separate thought: can brands be a part of the `PaymentMethodInformationObject` records? It would allow to have all the payment method properties in one place. And, it would clearly indicate which brand belongs to which payment method, for devs at least.

nagpai: Thanks for the review and feedback, @dmvrtx 🙌 

> Anyways, can we at least use translations inside the PAYMENT_METHOD_TITLES constant?

I was a bit divided about adding translations for brand names, since for eg. `Affirm` will be known as same in all languages, since it is a brand. Is it still recommended we add translation tags? 

Maybe good to add because some of them eg `Card Payments` and `In-person card payments` will need transation.  The translators will leave the proprietory brand names as it is. Am I understanding this correctly?

nagpai: >On a separate thought: can brands be a part of the PaymentMethodInformationObject records? It would allow to have all the payment method properties in one place. And, it would clearly indicate which brand belongs to which payment method, for devs at least.

The `PaymentMethodInformationObject` contains a list of PM configs. While most of them have a single brand associated with them, a few have multiple brands in them.. specifically so - Cards - ( Visa, Master, JCB, Amex, Discover etc ) . To help clarity for devs, do we want to add one more array property of `brands` or say `brandTitles` with all possible brands for that PM within it  ( or a single value for those that do not have multiple brands ) ?

dmvrtx: I like the `brandTitles` approach. It also made me think if we can somehow use it for "Afterpay/Clearpay" branding.

**Update:** Just checked, there is no `brand` property for BNPL payment methods, so it won't help us with this matter.

nagpai: Added translation tags here - 748cfb7

Added brandTitles property to the `PaymentMethodInformationObject`  here 49d11cc

mgascam: I may miss something, but it seems that the `brandTitles` are not being used?

mgascam (commented):
While testing in mobile I noticed that the tooltip is not displayed in the Transactions list. Since the tooltip is wrapped in a `clickable` element, when you tap on the icon it takes you to the payment details screen (in mobile you need to tap to trigger the tooltip, since there's no concept of hover).

A workaround could be to remove the `clickable` from this column, so tapping in the icon would display the tooltip in mobile. The drawback is that they would need to tap/click on other colum to get to the payment details.

On the other hand, if we don't remove the clickable, then merchants will be able to trigger the tooltip from the payment details screen.

I'm not sure what would be the best approach, so it would be great to hear your thoughts.

client/components/payment-method-details/index.js
nagpai: Hi Miguel. The `brandTitles` is useful to show a developer exploring `PaymentMethodInformationObject`, what are the different brand names that could be associated with the Payment method. Eg. there are many brand names that are associated with the `card` method. 

The current PR merely adds it since it came up in a discussion. ( probably not the best way to make additions to code, but this one was a small change ) .. 

The PR use case here requires the brand id to be changed to a proper humanly readable brand name with correct capitalization.. Eg. `visa` should get converted to `Visa` , `affirm` to `Affirm` .  Hence, `brandTitles` will probably help future devs find out what brands are associated with the PM, however it does not help us here to resolve a brand ID to a brand name. I hope this is making sense.

nagpai:
> While testing in mobile I noticed that the tooltip is not displayed in the Transactions list. Since the tooltip is wrapped in a `clickable` element, when you tap on the icon it takes you to the payment details screen (in mobile you need to tap to trigger the tooltip, since there's no concept of hover).

Thanks for delving into the mobile view of the transactions list page @mgascam 

From what I see, the page itself is not optimized for mobile. It has an enormous side-scroll that too only for the table part. 
I also checked that the tooltip showed up on the transactions details page, but I was not able to switch it off. This would be a problem on the transactions list page if all tooltips were turned on and it becomes annoying. With this context should we consult with the design team and probably reconsider using text `Affirm` or `Afterpay` beside the logo? ( what we had thought earlier ) ? IMO that would be a good option. 

Let me know your thoughts.

mgascam:
>> With this context should we consult with the design team and probably reconsider using text `Affirm` or `Afterpay` beside the logo? ( what we had thought earlier ) ? IMO that would be a good option.
> 
> Let me know your thoughts.
@nagpai Yes, I think we should reach out to the design team for advice. Thanks!

nagpai:
Hi @elizaan36 , we need some advice here regarding the tooltips on mobile view. I have managed to make them appear correctly on desktop view. I am using the HoverTip component already present in WCPay components. 

For mobile. It looks like the Transactions List page is anyway not responsive and involves a side-scroll.  In absence of mouseover in mobile,  the tooltip on the logo does not appear on touch, since it would open up the payment details page ( the whole row is clickable ) .  Even if we were to remove the link from the logo somehow, It seems that the tooltip shows up when we touch, but does not go away. If a user will click multiple logos, I think it will lead too too many tooltips open. 

Given this, what would be a good way to proceed? : 
1. Keep things as it is, since the transactions list page is currently optimized only for desktop 
2. Use text to the right of the logo 
Any other better way of doing this?  

Thanks!

elizaan36:
hi @nagpai, Looks good but I agree for mobile, the tooltip shouldn't block the merchant from opening the payment details. The tooltips are only useful on hover interactions for desktop, which are the majority of users. For mobile breakpoints (tappable interfaces) can we remove the the tooltip? This is a small percentage of users so I don't think it's needed and the impact of turning off the tooltip on mobile is quite low.

nagpai:
Based on guidance from design ( Elizabeth ) above, I have hidden the tooltip from the Payment Details page.

On the transactions list page, the tooltips anyway do not render in mobile since they are all clickable and lead to the payment details page.

kalessil (approved):
LGTM!

PR #6604, opened 2023-06-28
Hide Affirm from UPE if country isn't supported

### LINKED ISSUE #6515: UPE doesn't use dynamic rendering based on buyers location
### Describe the bug
UPE support dynamically rendering payment method as option based on customer location but the plugin doesn't integrate this functionality. As a result, payment methods that are restricted to specific locations (countries) are available for customers outside those locations.

So far this hasn't resulted in failed payments for existing payment methods but for BNPL Affirm payment method Stripe fails the payment if country is outside the allowed US or CAN. (it should be hidden as option if buyers is outside these countries)

### To Reproduce

1. Go to payments->settings
2. Enable any country-specific additional payment method, for example Giropay is available only in Germany
3. Go to checkout and set billing country not Germany
4. Observe that UPE is not re-rendered and still displays Giropay as option
5. Observe that changing currency re-renders UPE and Giropay is not available if currency is not EUR (example where we properly integrate UPE)

### Expected behavior
Additional payment method should be removed from UPE options in checkout when billing country changes

KarlisJ:
Fixes #6515 

Hide Affirm from UPE if the country isn't US or CA.

Unlike existing payment methods Stripe will fail the capture call for Affirm if the billing country is outside the supported US/CA country. This hasn't been a problem with existing payment methods. 

A proper solution would require exchanging WC billing fields with [Stripe's `address` element](https://stripe.com/docs/elements/address-element) which is a major refactor and can not be addressed on short notice. 

The proposed changes handle the hide/show of Affirm UPE container manually by listening to the changes in the billing country field. 

#### Testing instructions

1. Enable Affirm PM ([steps on server](https://github.com/Automattic/woocommerce-payments-server/pull/3405))
2. Add more than 50 USD in cart (Affirm won't show in checkout if less than 50)
(For both shortcode and blocks checkout)
3. Change the billing country to US -> Affirm should be available
4. Change the billing country to CA -> Affirm should be available
5. Change the billing country to anything else -> Affirm should **not** be available

#### Dev notes

~~This is currently a draft/PoC~~
- ~~Doesn't handle legacy UPE~~
- ~~Doesn't handle initial checkout load (for example, for existing customer billing country is pre-filled and could end up showing Affirm outside US/CA)~~

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6604` or branch name `fix/6515-hide-affirm-based-on-location` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6515-hide-affirm-based-on-location) :rocket:

---

Build info: 

- Latest commit: 1339c5ae10b0c8bfa8921b0fbd736c2254f623cd
- Build time: 2023-07-06 07:10:12 UTC

KarlisJ:
Commit [37685ae31b1d4a60ab17abe34a51242c1f04d952](https://github.com/Automattic/woocommerce-payments/pull/6604/commits/37685ae31b1d4a60ab17abe34a51242c1f04d952) handles initial load, but in case the billing country is not US/CA on initial load (existing customer, page refresh), we see Affirm for a second until it's hidden.

client/checkout/classic/upe-split.js
```diff
@@ -329,6 +329,18 @@ jQuery( function ( $ ) {
...
+	function maybeToggleAffirm() {
+		const billingCountry = $( '#billing_country' ).val();
+		const affirmContainer = $(
+			'#payment_method_woocommerce_payments_affirm'
+		).parent();
+		if ( 'US' === billingCountry || 'CA' === billingCountry ) {
+			affirmContainer.show();
+		} else {
+			affirmContainer.hide();
+		}
```
FangedParakeet: I'm having some issues enabling Affirm as a payment method, but thought I'd leave a comment in the meantime while I flail around in search for the right migration to run on my local Payments Server instance to end the pain. I'm pretty sure this will work and I don't have too many issues with the implementation. However, it seems like we could implement this in a way that is perhaps slightly more robust and extensible. 

Firstly, it seems like these countries should be both stored somewhere as constants instead of living alone inside this function. For example, we declare [compatible currencies](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-affirm-payment-method.php#L31) or whether the payment method [supports saved payments](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-affirm-payment-method.php#L29) as properties of the relevant `UPE_Payment_Method` class. These are then passed to the client via the [`paymentMethodsConfig` global parameter](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L152), whose contents are filled [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L229). 

This is one potential approach, but alternatively--since these values currently have no utility on the back-end of the plugin (unless we decide to toy with geolocation or anything like that)--perhaps these settings could live somewhere in a client-side constant somewhere like [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/constants.js)? This would basically make it easier to mirror similar functionality in the future, should we need to enable a new payment method with similar geographical restrictions.

Secondly, this implementation will successfully hide Affirm on the shortcode checkout for the split UPE. We will soon need to ensure that this integration also supports the blocks checkout and pay for order page, as well as all three checkouts when the deferred intent UPE is enabled. This definitely seems like it could be better placed inside a utility function ([here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/utils/upe.js) or [here](https://github.com/Automattic/woocommerce-payments/blob/develop/client/utils/checkout.js) maybe), so that we wouldn't have to duplicate client-side code in the plugin.

Finally, concerning hiding the gateways at checkout, I just want to confirm that this is simply to prevent an error being returned by Stripe. Obviously, since the gateways are still present in the DOM, an end-user could still revive these hidden payment methods and attempt to use them to complete a transaction. Hopefully, in this case, they would only be met by an error and nothing else harmful or unintended could happen from our perspective. 

Sorry, this is quite a large comment with some pretty scattered feedback, so hopefully it's not unwelcome! If there is an urgency to add this integration, many of the above suggestions and refactors could easily be handled in separate issues, so let me know if you'd prefer to proceed in that way instead of handling everything here. Also, it's likely I've made many miscomprehensions along the way--especially since I haven't actually been able to run the code yet 😅--so please don't hesitate to correct any incorrect assertions I may have made along the way and apologies in advance for the holes in my swiss-cheese understanding. Cheers!

KarlisJ: > Sorry, this is quite a large comment with some pretty scattered feedback, so hopefully it's not unwelcome! 

Your feedback is most welcomed here. Is overlaps with most of my own thoughts giving me a good green light for further improvements. 

> Finally, concerning hiding the gateways at checkout, I just want to confirm that this is simply to prevent an error being returned by Stripe.[..]Hopefully, in this case, they would only be met by an error and nothing else harmful or unintended could happen from our perspective.

Exactly, customer can not in any way tamper checkout, only expose them selves to flow that will always fail.

KarlisJ:
@FangedParakeet 
> need to ensure that this integration also supports the blocks checkout and pay for order page

I'm aware of blocks checkout but unaware of "pay for order page". Could you point me to the code handling that?

client/checkout/utils/upe.js
```diff
@@ -300,3 +300,18 @@ export const getStripeElementOptions = (
 
 	return options;
 };
+
+export const maybeTogglePaymentMethod = (
```
KarlisJ: I'm open for suggestions for better naming

client/checkout/classic/upe-split.js
```diff
@@ -354,6 +355,21 @@ jQuery( function ( $ ) {
...
+						paymentMethodType,
+						paymentMethodsConfig[ paymentMethodType ].countries
+					);
+					$( '#billing_country' ).on( 'change', function () {
+						maybeTogglePaymentMethod(
+							paymentMethodType,
+							paymentMethodsConfig[ paymentMethodType ].countries
+						);
+					} );
+				}
```
KarlisJ: 1. Should this block be moved to a separate function?

KarlisJ: 2. Is it ok to add multiple listeners, or should map all PMs with restricted countries and handle them in a single listener?

FangedParakeet:
> I'm aware of blocks checkout but unaware of "pay for order page". Could you point me to the code handling that?

For complete context, @KarlisJ, if you have an order that is pending payment, you can send customers a link to the [Pay for Order/Order Pay page](https://woocommerce.com/document/managing-orders/#paying-for-an-order), which will allow the customer to pay for the order. The difference between this and the regular checkout is that there is no cart object, but in its place there is an order that has already been created and we use this order to collect all the checkout payment information that we require. 

One popular way to set this up is to use a [3DS payment method](https://stripe.com/docs/testing#regulatory-cards) at checkout to complete an order and manually fail the authentication; this will create a new order with pending payment status and you can visit this new order in WP Admin to collect the Order Pay link. Alternatively (what I do more often) is to just manually create an order from WP Admin, add a product, a customer, and set the status to payment pending. 

On the Pay for Order page, when we mount the payment element, we first [collect the order ID from the config](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L207-L212), then [we use this to create a payment intent](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L228-L232), so that it uses the order--rather than absent cart--to generate cart totals, currency, etc. 

When the customer submits the order, instead of listening for the regular checkout form, [we instead add a listener to the `#order_review` form](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L730-L744). Finally, instead of [processing the checkout via AJAX to create the order](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L559), we just [update the payment intent](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L467) with payment method metadata, etc., then [pass the return URL to the Order Received page to Stripe and confirm the payment](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js#L485-L491) as usual. 

Those are the basics, but the tragedy of the Order Pay page is quite unorthodox and definitely not a story most Jedi would tell you about, so if anything else is still unclear (which I'd be surprised if it were not the case), please don't hesitate to question more. Cheers!

includes/payment-methods/class-upe-payment-method.php
```diff
@@ -74,6 +74,14 @@ abstract class UPE_Payment_Method {
 	 */
 	protected $icon_url;
 
+	/**
+	 * Supported customer locations for which charges for a payment method can be processed
+	 * Empty if all customer locations are supported
+	 *
+	 * @var string[]
+	 */
+	protected $countries = [];
```
KarlisJ, line 83: I'm open for discussion if we should be more specific in naming, like `supported_countries`

kalessil: Current naming looks good.

includes/payment-methods/class-affirm-payment-method.php
```diff
@@ -39,6 +39,7 @@ public function __construct( $token_service ) {
 				'max' => 3000000,
 			], // Represents USD 50 - 30,000 USD.
 		];
+		$this->countries           = [ 'US', 'CA' ];
```
kalessil, line 42: IIRC, it could be not a case for Afterpay, but would it make sense to specify countries there as well? Or add a comment explaining that here is a must-have as Affirm has stricter requirements toward the country?

client/checkout/utils/upe.js
```diff
@@ -300,3 +300,18 @@ export const getStripeElementOptions = (
...
+	paymentMethodType,
+	supportedCountries
+) => {
+	const billingCountry = document.querySelector( '#billing_country' ).value;
```
kalessil: Can we do a bit of optimization and ensure supportedCountries are not empty in the first place?

client/checkout/utils/upe.js
kalessil: togglePaymentMethodForSupportedCountries or togglePaymentMethodForCountry

kalessil: Would it make sense to relocate this into `client/checkout/classic/upe-split.js`?

client/checkout/classic/upe-split.js
```diff
@@ -354,6 +368,12 @@ jQuery( function ( $ ) {
 				} else {
 					mountUPEElement( paymentMethodType, upeDOMElement );
 				}
+
+				if (
```
kalessil: With the comment to check countries are not empty inside `maybeTogglePaymentMethod` we might omit the if-part and always call `restrictPaymentMethodToLocation`. To me, it'd be a more readable approach.

client/checkout/blocks/upe-split.js
```diff
@@ -99,7 +99,17 @@ Object.entries( enabledPaymentMethodsConfig )
 						upeConfig.testingInstructions
 				  ),
 			savedTokenComponent: <SavedTokenHandler api={ api } />,
-			canMakePayment: () => !! api.getStripe(),
+			canMakePayment: ( cartData ) => {
+				if (
```
kalessil: Find this a bit difficult to read, perhaps this approach can be adopted?
```javascript
const billingCountry                 = cartData.billingAddress.country;
const isAvailableInTheCountry = ! upeConfig.countries.length || upeConfig.countries.includes( billingCountry );
return isAvailableInTheCountry && !! api.getStripe();
```

kalessil (changes requested):
That's a great approach! Leaving some comments, which hopefully allow a bit of simplification and improve readability.

includes/payment-methods/class-affirm-payment-method.php
KarlisJ: Two things here.

**From the code perspective.** 
I followed same approach we have for currencies. Empty array means that there are no restrictions. Otherwise we'd have to list all countries that we support and update that list every time we add new supported country to WCpay. 

You also had comment 
> Can we do a bit of optimization and ensure supportedCountries are not empty in the first place?

which is related to this. We need a way to figure out, if any restriction logic should be applied. 

I think the approach is good, but we should ensure that it's clear why we have to check for supportedCountries. Either wrap in proper method name or adding a comment would work. 

**From business perspective** 
We need to raise the question if we want to apply this strict logic to existing LPMs. Existing LPMs don't end up in hard failure, so technically, not restricting them to how Stripe has listed them in docs increases the conversation rate by a very small amount.

I'd like to first merge this PR, so there's a better context for discussion.

client/checkout/utils/upe.js
KarlisJ: See [this comment](https://github.com/Automattic/woocommerce-payments/pull/6604/files/abb861e13364918c726666b5b49e137c4ee336a0#r1251650419)

client/checkout/classic/upe-split.js
KarlisJ: See [this comment](https://github.com/Automattic/woocommerce-payments/pull/6604/files/abb861e13364918c726666b5b49e137c4ee336a0#r1251650419)

client/checkout/utils/upe.js
KarlisJ: Yes, I'll do that. The initial idea was to reuse same method for blocks checkout but it's clear it's not reusable as the mechanics are completely different

includes/payment-methods/class-affirm-payment-method.php
KarlisJ: Here's a simple example of what I meant about "Either wrap in proper method name or adding a comment would work."
```JS
const isRestrictedInAnyCountry = !! upeConfig.countries.length;
```

Now it's clear that if there are no countries specified, then we allow PM in all countries.

kalessil: Makes sense, `const isRestrictedInAnyCountry = !! upeConfig.countries.length;` will work as bring the needed clarity. Go for it =)

kalessil: Before merging, lets polish as per remaining comments.

client/checkout/classic/upe-split.js
```diff
@@ -329,6 +329,41 @@ jQuery( function ( $ ) {
 		gatewayUPEComponents[ paymentMethodType ].upeElement = upeElement;
 	};
 
+	function togglePaymentMethodForCountry(
+		paymentMethodType,
+		supportedCountries
+	) {
+		const billingCountry = document.querySelector( '#billing_country' )
+			.value;
+		const affirmContainer = document.querySelector(
```
kalessil: And if affirmContainer is not there (Affirm disabled)?

kalessil (approved):
Love it :heart: Pre-approving: see also my comment + a couple of test are failing now.

client/checkout/classic/upe-split.js
kalessil: Also `affirmContainer` name is misleading at this point.

KarlisJ: I missed the naming, will update.

As for the case when it's not there. It's not possible. 
```JS
const upeDOMElements = $( '.wcpay-upe-element' );
for ( let i = 0; i < upeDOMElements.length; i++ ) {
	const upeDOMElement = upeDOMElements[ i ];
	const paymentMethodType = $( upeDOMElement ).attr(
		'data-payment-method-type'
	);
				```
				
If it's not on the page, then we won't try to restrict it. Simple as that.

kalessil: Great, thanks!

PR #6606, opened 2023-06-28
Draft - hide affirm based on location

nagpai:
related to #6515 

Hides Affirm if the country on initial load is not US or CA.  Also updates if the country is changed in the dropdown ( borrowed from #6604 by @KarlisJ )

### DIFF
--- a/client/checkout/classic/upe-split.js
+++ b/client/checkout/classic/upe-split.js
@@ -329,6 +329,19 @@ jQuery( function ( $ ) {
 		gatewayUPEComponents[ paymentMethodType ].upeElement = upeElement;
 	};
 
+	const maybeShowAffirm = () => {
+		const billingCountry = $( '#billing_country' ).val();
+		const affirmContainer = $(
+			'#payment_method_woocommerce_payments_affirm'
+		).parent();
+
+		if ( 'US' === billingCountry || 'CA' === billingCountry ) {
+			affirmContainer.show();
+		} else {
+			affirmContainer.hide();
+		}
+	};
+
 	// Only attempt to mount the card element once that section of the page has loaded. We can use the updated_checkout
 	// event for this. This part of the page can also reload based on changes to checkout details, so we call unmount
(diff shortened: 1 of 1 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6606` or branch name `fix/v1-6515-hide-affirm-based-on-location` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/v1-6515-hide-affirm-based-on-location) :rocket:

---

Build info: 

- Latest commit: a2c29595ad49d8a9c90f2313c50b51d942ceb48f
- Build time: 2023-06-28 08:38:33 UTC

kalessil:
In general, the solution is straightforward and the idea is clear.
What are the options to generalize this approach (so all available PMs are checked) and not hard-code countries in JS?

KarlisJ:
@kalessil 
> What are the options to generalize this approach (so all available PMs are checked) and not hard-code countries in JS?

JS side we create the listener for the change event but generalise the handler function. Ideally, we should add something like `$this->customer_countries = [..]` to `UPE_Payment_Method` and pass that data to frontend. Then the JS handler can read that config and act accordingly. 

That would include changes in 
- `abstract class UPE_Payment_Method`
- any implementation of that class that have such restrictions
- `display_gateway_html` method on split UPE gateway which is where we can pass data as html attribute (we do that now for payment type)
- JS changes in the same place they are now

But we should be careful with applying these changes to existing PM as it could be considered unwanted behaviour for those methods.

kalessil:
@nagpai: is this PR still relevant?

nagpai:
@kalessil  No longer relevant. Closing this since the issue has been resolved by PR #6604 , merged yesterday. Thanks.

PR #6708, opened 2023-07-07
Restrict PM to domestic country

### LINKED ISSUE #6517: UPE doesn't use dynamic rendering based on allowed presentment currency
### Describe the bug
Affirm/Afterpay payment method is supported for USD and CAN currency but Stripe limits the currency that Affirm/Afterpay can accept payments in based on account's registered location. So a US based account can not receive payments in CAD though a CAN based account can. 

Plugin only lists both currencies for payment method resulting in both being valid options for payment method in the UI but only one currency can actually be used.

### To Reproduce

1. Checkout `add/6119-shortcode-checkout-for-bnpl` branch
2. Ensure your Stripe account is US based
3. Ensure you use multicurrency and have USD and CAD as options
4. Enable Affirm payment method (Might need [server setup](https://github.com/Automattic/woocommerce-payments-server/pull/3405))
5. Visit checkout and select USD
6. Observe Affirm as option in UPE
7. Observe that payment is successful
8. Repeat steps but select CAD as currency
9. Observe that Affirm is still available
10. Observe that payment fails

### Actual behavior
UI gets error message
>We're not able to process this request. Please refresh the page and try again.

In the logs
> ERROR The currency provided (cad) is invalid. Payments with affirm support the following currencies: usd. (invalid_request_error)

### Expected behavior
Affirm payment method should not be available as option if currency is set to CAD

KarlisJ:
Fixes #6517 

Add property to `UPE_Payment_Method` class indicating that currency should be restricted to Stripe's connected account currency. This change is required for newly introduced BNPL methods - Affirm and Affterpay/Clearpay. 

See associated issue for more details about the problem. 

#### Testing instructions

1. Make sure merchant in US based account and multicurrecy turned on with at least CAD as option
2. Enable Affirm and Afterpay 
3. Add at least 50 USD to cart (Affirm is not available below that amount)
4. Go to checkout, default currency should be USD and both methods available in UPE element
5. Switch currency to CAD
6. Payment method should not be available anymore.

If same steps are repeated without proposed changes, we observe error in step 6.

**Turn off multi-currency** 
- the solution shouldn't brake existing checkout
- we expect that PMs are not available if currency is outside merchants currency (for example, EUR merchant still doesn't have Affirm)

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6708` or branch name `fix/6517-restrict-bnpl-to-domestic-country` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/6517-restrict-bnpl-to-domestic-country) :rocket:

---

Build info: 

- Latest commit: 1e6f7517de3becb4a77c9524e228b74729adeb9b
- Build time: 2023-07-25 10:51:14 UTC

kalessil:
Thanks for creating the draft - it gives great context!
I propose to simplify the approach a bit and do something like this for BNPLs:

```php
    public function __construct( $token_service ) {
        ...
        $this->currencies = [ strtoupper( \WC_Payments::get_account_service()->get_account_default_currency() ) ]
        ...
    }
```

KarlisJ:
@kalessil At first, I though of the same approach but it's a bit more tricky.

Here's the end rules we need
- empty array -> no restriction
- array -> restrict to values in array
- array + BNPL -> on top of previous, check against default currency

In your example, we'll get `$this->currencies = [ 'EUR' ]` for all EU based merchants which is incorrect. However, we can't add check like 
```php
$this->currencties = in_array($default_currency, ['USD', 'CAD'] ? [ $default_currency ] : [];
```
because then for EU merchants we get empty array which is interpreted already as "no currency restrictions".

kalessil:
@KarlisJ, go for it! And please add the comment about the case for EUR in Affirm code, so future people understand that we simplified as much as possible.

includes/payment-methods/class-affirm-payment-method.php
```diff
@@ -24,12 +25,13 @@ class Affirm_Payment_Method extends UPE_Payment_Method {
...
-		$this->icon_url            = plugins_url( 'assets/images/payment-methods/affirm.svg', WCPAY_PLUGIN_FILE );
-		$this->currencies          = [ 'USD', 'CAD' ];
-		$this->limits_per_currency = [
+		$this->stripe_id                     = self::PAYMENT_METHOD_STRIPE_ID;
+		$this->title                         = __( 'Affirm', 'woocommerce-payments' );
+		$this->is_reusable                   = false;
+		$this->icon_url                      = plugins_url( 'assets/images/payment-methods/affirm.svg', WCPAY_PLUGIN_FILE );
+		$this->currencies                    = [ 'USD', 'CAD' ];
+		$this->accept_only_domestic_currency = true;
+		$this->limits_per_currency           = [
```
kalessil: I added some comments asking to tweak naming `limits_per_currency` -> `limits_per_buyer_currency` and `countries` -> `buyer_countries`, but don't see them. Duplicating the request once more.

KarlisJ:
@kalessil 
> go for it!

I think you misunderstood me. We shouldn't limit the currencies in the constructor per your suggestion because that alters how we think about what that property is used for.

includes/payment-methods/class-affirm-payment-method.php
KarlisJ: > I added some comments asking to tweak naming limits_per_currency -> limits_per_buyer_currency and countries -> buyer_countries, but don't see them.

Sorry, I can't find any comment about this request. 

Can you elaborate the reasoning for this change? I wonder why not also `currencies` -> `buyer_currencies` then?

kalessil: Sure: it's to align with naming we have on the server side (see transactions cache tables, parameters naming for transactions endpoints). Having the naming aligned makes it easier to draw connections.

kalessil:
@karlis: I'm not following. 

My understanding is the following:
- We want to limit BNPLs so they are available on checkout when the currency matches the merchant's default currency.
- `currencies` is what's designed to implement this limitation, even if populating for BNPLs is not as straightforward as for other payment methods.

The rules for calculating currency as I see them (incorporates your insights):
- order currency shall be in `[ 'USD', 'CAD' ]` for Affirm, in `[ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ]` for Afterpay
- if so, we populate `currencies` with default merchant currency, otherwise, with `[ 'USD', 'CAD' ]`/`[ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ]`

Regarding EUR-zone:
- we rolled out Affirm for non-EU countries only: `const COUNTRIES_WITH_AFTERPAY    = [ Country::US, Country::GB, Country::CA, Country::AU, Country::NZ ];` so we should remove EUR from the allowed currencies for Afterpay
- it still a valid case, which invalidates my comment 😢  (I was hoping we can simplify this)

I think we are aligned now, so feel free to mark this thread as resolved - I'll do a new review round.

includes/payment-methods/class-afterpay-payment-method.php
```diff
@@ -24,12 +24,13 @@ class Afterpay_Payment_Method extends UPE_Payment_Method {
...
-		$this->title               = __( 'Afterpay', 'woocommerce-payments' );
-		$this->is_reusable         = false;
-		$this->icon_url            = plugins_url( 'assets/images/payment-methods/afterpay.svg', WCPAY_PLUGIN_FILE );
-		$this->currencies          = [ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ];
-		$this->limits_per_currency = [
+		$this->stripe_id                     = self::PAYMENT_METHOD_STRIPE_ID;
+		$this->title                         = __( 'Afterpay', 'woocommerce-payments' );
+		$this->is_reusable                   = false;
+		$this->icon_url                      = plugins_url( 'assets/images/payment-methods/afterpay.svg', WCPAY_PLUGIN_FILE );
+		$this->currencies                    = [ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ];
```
kalessil: We an drop 'EUR' as the BNPLs rollout didn't include EU-countries.

includes/payment-methods/class-upe-payment-method.php
```diff
@@ -167,7 +176,16 @@ public function is_reusable() {
 	 * @return bool
 	 */
 	public function is_currency_valid( $order_id = null ) {
-		return empty( $this->currencies ) || in_array( $this->get_currency( $order_id ), $this->currencies, true );
+		$current_store_currency = $this->get_currency( $order_id );
+
+		if ( $this->accept_only_domestic_currency ) {
+			$default_currency = \WC_Payments::get_account_service()->get_account_default_currency();
```
kalessil: Can we tweak naming `default_currency` -> `default_merchant_currency`, please?

includes/payment-methods/class-upe-payment-method.php
```diff
@@ -60,6 +61,14 @@ abstract class UPE_Payment_Method {
 	 */
 	protected $currencies;
 
+	/**
+	 * Should payment method be restricted to only domestic payments.
+	 * E.g. only to Stripe's connected account currency.
+	 *
+	 * @var boolean
+	 */
+	protected $accept_only_domestic_currency = false;
```
kalessil: `accept_only_domestic_payments` would better explain the intention here.

kalessil (changes requested):

karlis:
Wrong tagged KārlisOn 14 Jul 2023, at 10:45, Vladimir Reznichenko ***@***.***> wrote:﻿
@kalessil requested changes on this pull request.



In includes/payment-methods/class-afterpay-payment-method.php:
> @@ -24,12 +24,13 @@ class Afterpay_Payment_Method extends UPE_Payment_Method {
 	 */
 	public function __construct( $token_service ) {
 		parent::__construct( $token_service );
-		$this->stripe_id           = self::PAYMENT_METHOD_STRIPE_ID;
-		$this->title               = __( 'Afterpay', 'woocommerce-payments' );
-		$this->is_reusable         = false;
-		$this->icon_url            = plugins_url( 'assets/images/payment-methods/afterpay.svg', WCPAY_PLUGIN_FILE );
-		$this->currencies          = [ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ];
-		$this->limits_per_currency = [
+		$this->stripe_id                     = self::PAYMENT_METHOD_STRIPE_ID;
+		$this->title                         = __( 'Afterpay', 'woocommerce-payments' );
+		$this->is_reusable                   = false;
+		$this->icon_url                      = plugins_url( 'assets/images/payment-methods/afterpay.svg', WCPAY_PLUGIN_FILE );
+		$this->currencies                    = [ 'USD', 'CAD', 'AUD', 'NZD', 'GBP', 'EUR' ];

We an drop 'EUR' as the BNPLs rollout didn't include EU-countries.



In includes/payment-methods/class-upe-payment-method.php:
> @@ -167,7 +176,16 @@ public function is_reusable() {
 	 * @return bool
 	 */
 	public function is_currency_valid( $order_id = null ) {
-		return empty( $this->currencies ) || in_array( $this->get_currency( $order_id ), $this->currencies, true );
+		$current_store_currency = $this->get_currency( $order_id );
+
+		if ( $this->accept_only_domestic_currency ) {
+			$default_currency = \WC_Payments::get_account_service()->get_account_default_currency();

Can we tweak naming default_currency -> default_merchant_currency, please?



In includes/payment-methods/class-upe-payment-method.php:
> @@ -60,6 +61,14 @@ abstract class UPE_Payment_Method {
 	 */
 	protected $currencies;
 
+	/**
+	 * Should payment method be restricted to only domestic payments.
+	 * E.g. only to Stripe's connected account currency.
+	 *
+	 * @var boolean
+	 */
+	protected $accept_only_domestic_currency = false;

accept_only_domestic_payments would better explain the intention here.

KarlisJ:
@kalessil My thought process here is that I don't like having hidden logic in the constructor. We have `get_currencies` method which in future can be used with expectation to get currencies where PM can be used. If we have this hidden dependency in the constructor then this method will return different results depending on the account executing the method. What will the outcome look for internal analytics, for example? 

That's why I'm keen on adding this new logic in the `is_currency_valid` because that's a method dedicated to such check.

includes/payment-methods/class-affirm-payment-method.php
KarlisJ: Sorry, I can't see how it makes things better. I don't see any reference of `buyer_...` either in associated cache class or endpoints. We have `customer_currency` and `customer_country` in the cache, for example.

Also, I don't see how it's related to this PR. Can we maybe create a separate PR and move any related discussion there? I'm happy to create the PR and take it on so that it's not lost in prioritisation.

includes/payment-methods/class-afterpay-payment-method.php
KarlisJ: Given that it's wrongly introduced with the BNPL merge, I'd feel better merging simple one line PR for this so that the reasoning of removing EUR is fully clear (not hidden inside this bug fix).

kalessil: Ok, we can address this in a separate PR.

includes/payment-methods/class-affirm-payment-method.php
kalessil: You are right about naming, it's `customer_` in DB.

includes/payment-methods/class-upe-payment-method.php
kalessil: That's works for PoC, but for the prod-grade implementation, the currency shall be dispatched from outside.

The gateway implementation has `\WC_Payment_Gateway_WCPay::get_account_statement_descriptor`, something similar can be introduced for accessing the currency and passing it by in here.

That's a clearer way towards testing and not coupling different layers of components.

kalessil (changes requested):
Thanks for answering my questions and exploring the simplification opportunities.
Let's stick to the current approach, and do final stretch toward prod-grade implementation - see my comment.

includes/payment-methods/class-upe-payment-method.php
KarlisJ: I tried implementing this and not sure it's necessary.

In the `UPE_Payment_Method` we currently don't have the gateway injected. Do we really want to update all instances of UPE PMs with that dependency if it's necessary only when `if ( $this->accept_only_domestic_payment ) {` ?

kalessil: I was thinking instead about `public function is_currency_valid( $order_id = null, $domestic_currency = null )`.

kalessil: Or `public function is_currency_valid( $order_id = null, Cached_Account_Data_Holder $account = null )`

KarlisJ: oh, I see. I'll explore

kalessil (approved):
LGTM (code review only)!

PR #6656, opened 2023-07-03
WooPay compatibility with deferred intent creation UPE

timur27:
Closes https://github.com/Automattic/woocommerce-payments/issues/6291

#### Changes proposed in this Pull Request
Currently, split UPE with the deferred intent creation on the server is implemented in all flows, including shortcode & Blocks checkout and _My account_ subpages, such as _Payment methods_, _Subscriptions_, _Pay for order_. The time has come for the WooPay to be compatible with this UPE type as well, thus this PR ensures that WooPay's functionality previously used only in pair with the legacy card gateway, is now entirely usable with the deferred intent creation UPE as well.

#### Testing instructions

> 💡 Ensure your WCPay Server listens to Stripe webhooks.

> Legacy card/legacy UPE/split UPE all disabled in WCPay Dev Tools. Only deferred intent creation UPE is enabled.
WooPay is enabled in Payments -> Settings. 
Only card payment method is enabled in Payments -> Settings. Act as a logged-in user. 

1️⃣ 

1. Add a product to the cart and navigate the shortcode checkout. 
2. On shortcode checkout, input default card details, billing details, and the email which isn’t yet registered with WooPay
3. Confirm that the _Save my information for a faster checkout_ checkbox is displayed under the card gateway. Don’t check this checkbox yet. 
4. Place the order and confirm that the order status is _Processing_, and on the Stripe side - payment in status _Succeeded_
5. Repeat steps 1-2, then check the  _Save my information for a faster checkout_ checkbox and place the order. 
6. Confirm you were redirected to an order placed page.
7. On WooPay's site admin panel, navigate to _Users_ and confirm that the user with the email you used in step 5 is now visible on the list. 
8. On the Stripe Customers page (e.g., `https://dashboard.stripe.com/test/customers`), confirm that the user is also listed.

Now, since you have the user registered in WooPay, check the following:

2️⃣  
1. Add a product to the cart and proceed to the shortcode checkout page again.
2. The email should be already provided but if not - provide the email of the previously registered user. 
3. Confirm that WooPay's pop-up modal appears on the page asking you to provide the OTP code for verification. 
4. After entering `000000` make sure you're redirected successfully to the checkout page with all the cart details.
5. After adding a new payment method/new shipping address if needed, on the WooPay checkout page, click _Place Order_ and confirm you're redirected to the merchant's order summary page stating the order was received successfully, and the payment method is a card with WooPay icon on the left. 
[...]

PR #6784, opened 2023-07-15
Implement Stripe Link with deferred intent UPE

### LINKED ISSUE #6261: Stripe Link is Incompatible with Subscription Renewals
### Describe the bug
If a subscription is purchased via Stripe Link, subsequent renewals will not process. They will fail with the following error:

```
2023-05-09T01:38:14+00:00 ERROR There is no saved payment token for order # [...]
```

In my tests, the `pm_` ID is stored correctly in Stripe, but that information isn't being passed to Woo or stored on the site to retrieve that information for renewals.

### To Reproduce

1. Enable UPE and enable Stripe Link.
2. Purchase a subscription product.
3. Edit the subscription at **WooCommerce > Subscriptions** and attempt to change the payment method—you will be unable to.
4. Go through the `Renew now` flow and the subscription can be renewed as the customer has to go through the checkout flow again.
5. Process a renewal via the `Subscription actions` metabox or by manually running the associated `woocommerce_scheduled_subscription_payment` scheduled action.
6. The renewal order will fail.

### Actual behavior
If a customer purchases a subscription via Stripe Link, the renewals fail, and there is no way for the subscription tokens to be added manually. 

### Screenshots
Markup on 2023-05-09 at 17:42:14

Markup on 2023-05-09 at 17:43:15

Markup on 2023-05-09 at 17:43:53

Markup on 2023-05-09 at 17:44:20

### Expected behavior
Either of the following:
1. Subscriptions purchased via Stripe Link renew properly.
2. Stripe Link is not offered for carts that include a subscription product.

### Additional context
p1683573649730159-slack-C7U3Y3VMY
6271366-zen

### LINKED ISSUE #6391: Implement Stripe Link with deferred intent UPE
### Description

In the current landscape of WCPay, Stripe Link is incompatible with WooPay and dependent on the legacy UPE card element. While due to UI conflicts, we may want to persist the incompatibility of WooPay and Link, we will need to add client scripts to implement Link functionality within the deferred intent gateway, so that we no longer need to use the insecure, legacy UPE.

### Acceptance criteria

* With deferred intent split UPE and Stripe Link enabled, card payment gateway uses new deferred intent UPE and still supports Link.
* Sign-up and checkout flows are functional for Stripe Link, using the deferred intent UPE.
* Link and WooPay cannot be enabled together.

### LINKED ISSUE #6709: Temporarily disable Link for subscription products
### Describe the bug

Related to https://github.com/Automattic/woocommerce-payments/issues/6261
As a temporary measure (discussed in Slack p1688658199754469/1688575009.270539-slack-C043JN7AJHY ), we're going to temporarily disable Link as a payment method when the cart contains a subscription product.

Additional investigation will be done https://github.com/Automattic/woocommerce-payments/issues/6261

### To Reproduce

1. Enable UPE and enable Stripe Link.
2. Purchase a subscription product.

### Actual behavior

1. Edit the subscription at **WooCommerce > Subscriptions** and attempt to change the payment method—you will be unable to.
2. Go through the `Renew now` flow and the subscription can be renewed as the customer has to go through the checkout flow again.
3. Process a renewal via the `Subscription actions` metabox or by manually running the associated `woocommerce_scheduled_subscription_payment` scheduled action.
4. The renewal order will fail.

### Expected behavior

1. Enable UPE and enable Stripe Link.
2. Add a subscription product to the cart
3. Stripe Link should not be an available payment method displayed at checkout

FangedParakeet:
Fixes #6391 
Fixes https://github.com/Automattic/woocommerce-payments/issues/6709
Fixes https://github.com/Automattic/woocommerce-payments/issues/6261

# Linking up with Stripe Link

This implements Stripe Link in both shortcode and blocks checkouts, while using the deferred intent split UPE. The changes involved are somewhat less dramatic than they might appear. The client-side integration remains mostly the same, using the existing exportable integration with the deferred intent checkout scripts. On the server-side, we need to ensure that, when Link is enabled, `link` is added alongside `card` in the `payment_method_types` value sent to Stripe and we must also add a new `mandate_data` value, which is fortunately already an available parameter on the Payments Server.

Scattered in amongst this implementation are a few fixes to the blocks checkout parameters and some fixes to how session intent data is stored and cleared, just FYI.

## Steps to enable Stripe Link

Stripe Link is only available to US stores, so you will need to onboard your test site with Stripe by using a US-based address. Other than please enable the deferred intent UPE via the dev tools and you should be able to enable Link from the Payments settings, when WooPay is disabled. If Link does still not appear on your Payments settings screen, you may need to visit the Stripe dashboard for your site's account and ensure that the Link capability is requested and active. After requesting the capability, clear your Payments Server cache and hopefully Link will appear before your eyes.

## How to checkout with Stripe Link

Link is only available to US stores and consequently is also only compatible with USD at checkout. When these conditions are met, you should find that the email input rises to the top of the checkout and has a Link icon attached to it. This icon is a button that will trigger the Link checkout experience. 

However, if you do not immediately see this button, you should be aware that this button is linked to the payment element and only present once this element has been mounted to the DOM. In the shortcode checkout, we mount all the payment elements immediately on page load, so this button should appear immediately without any encouragement. On the blocks checkout, however, we only mount payment elements when the relevant payment gateway is selected, so you may need to select the CC payment gateway radio button first, in order to mount the element and summon the Link button from of its slumber.

If your email address is not identified, you should trigger the registration "Save your info" checkbox to be added to the payment element; whereas if your email address is identified, you will instead trigger the OTP authentication modal to appear on top of the checkout. Note that in test mode, when you register a new Link account **it will only be usable with the Stripe account where this Link account is registered** and cannot be used across different sites connected to the same platform. Additionally, while registration requires a phone number, in test mode **Link will never send actual SMS codes to your device**: you can instead use the top-secret code `000000` every time to surpass the OTP challenge. 	

#### Testing instructions

Stripe Link should be implemented on both blocks and shortcode checkout pages, so these will need to be tested. Here are some combinations of features you should probably ensure are working.

- Regular product checkout.
- Checkout with subscription products.
   - Checkout with simple subscription.
   - Checkout with free-trial subscription.
   - Manual renewal for subscription.
   - Automatic renewal for subscription.
 - Check payment intent metadata. 
 - Saving Link payment methods as WC Payment Tokens (apparently this is possible now).
 - Link on the Order Pay page (I think this should be possible).
 - And anything else you think might be particularly dicey.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6784` or branch name `add/6391-implement-stripe-link-with-deferred-intent` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/6391-implement-stripe-link-with-deferred-intent) :rocket:

---

Build info: 

- Latest commit: 732799127da0493072d0e8bbce1969e982e4f79d
- Build time: 2023-08-04 12:42:35 UTC

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -129,7 +198,7 @@ const PaymentProcessor = ( {
 						type: 'success',
 						meta: {
 							paymentMethodData: {
-								payment_method: paymentMethodId,
+								payment_method: upeMethods[ paymentMethodId ],
```
FangedParakeet, line 199: I think this is more consistent with other gateways: essentially we send the gateway ID (e.g. `woocommerce_payments`) instead of the Stripe ID (e.g. `card`) in the `$_POST['payment_method']` field. This is how I believe this parameter is intended to be used and have added a few other accommodations in the plugin to account for this.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1117,6 +1126,7 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
 
 				// Afterpay expects the shipping address to be sent in the request. This is not required for other payment methods.
+				// TODO: Check if this ever works...
 				if ( Payment_Method::AFTERPAY === $payment_information->get_payment_method() ) {
 					$request->set_shipping( $this->get_shipping_data_from_order( $order ) );
```
FangedParakeet: I forgot I left this TODO in the code. 😂 But FWIW, I reached out to the team responsible and it appears this indeed never worked and is apparently not needed anyways and soon to be removed (see p1689657981302729/1689060282.807879-slack-C02C00HCJ8J).

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1371,13 +1400,35 @@ private function get_payment_methods_from_request() {
...
+	 *
+	 * @return array mandate_data values to use in request.
+	 */
+	private function get_mandate_data() {
+		return [
+			'customer_acceptance' => [
+				'type'   => 'online',
+				'online' => [
+					'ip_address' => WC_Geolocation::get_ip_address(),
+					'user_agent' => 'WooCommerce Payments/' . WCPAY_VERSION_NUMBER . '; ' . get_bloginfo( 'url' ),
```
FangedParakeet, line 1430: Our usage of user agent parameters is pretty inconsistent. Our API client uses `"WooCommerce Payments/" + WCPAY_VERSION_NUMBER`, but there are a few other places where the more specific blog URL is also added. I'd think either that or the Stripe account ID would probably be most relevant here, so included the former. Also, not sure if there's a better way to collect the IP address without the WC dependency. 🤷

includes/core/server/request/class-create-and-confirm-setup-intention.php
```diff
@@ -83,4 +83,34 @@ public function set_payment_method( string $payment_method_id ) {
...
+		// Hard to validate without hardcoding a list here.
+		if ( empty( $payment_methods ) ) {
+			throw new Invalid_Request_Parameter_Exception(
+				__( 'Intentions require at least one payment method', 'woocommerce-payments' ),
+				'wcpay_core_invalid_request_parameter_missing_payment_method_types'
+			);
+		}
+
+		$this->set_param( 'payment_method_types', $payment_methods );
+	}
```
FangedParakeet: I wish there was a better place to define this function, so that it could be shared between `Create_And_Confirm_Intention` and `Create_And_Confirm_Setup_Intention` without duplicating code, but sadly they are in separate branches of the class hierarchy for some reason. Perhaps this could be cause for a better refactor in the future.

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -778,7 +778,7 @@ public function process_redirect_payment( $order_id, $intent_id, $save_payment_m
 				$this->set_payment_method_title_for_order( $order, $payment_method_type, $payment_method_details );
 				$this->order_service->attach_transaction_fee_to_order( $order, $charge );
 
-				self::remove_upe_payment_intent_from_session();
+				static::remove_upe_payment_intent_from_session();
```
FangedParakeet, line 819: I think these should all be `static` calls and this was actually a dormant bug. When placed here, `self` will always call `UPE_Payment_Gateway::remove_upe_payment_intent_from_session()` instead of `UPE_Split_Payment_Gateway::remove_upe_payment_intent_from_session()` when the calling class is instead an instance of `UPE_Split_Payment_Gateway`.

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -164,7 +164,7 @@ public function set_payment_method_title_for_order( $order, $payment_method_type
 
 		$payment_method_title = $payment_method->get_title( $payment_method_details );
 
-		$payment_gateway = 'card' === $payment_method->get_id() ? self::GATEWAY_ID : self::GATEWAY_ID . '_' . $payment_method_type;
+		$payment_gateway = in_array( $payment_method->get_id(), [ Payment_Method::CARD, Payment_Method::LINK ], true ) ? self::GATEWAY_ID : self::GATEWAY_ID . '_' . $payment_method_type;
```
FangedParakeet, line 160: I believe this will fix #6261 in both original and deferred intent split UPE.

tests/unit/payment-methods/test-class-upe-split-payment-gateway.php
```diff
@@ -2028,7 +2024,7 @@ public function test_link_payment_method_if_card_enabled() {
 					$this->mock_customer_service,
 					$this->mock_token_service,
 					$this->mock_action_scheduler_service,
-					$this->mock_payment_methods[ Payment_Method::LINK ],
+					$this->mock_payment_methods[ Payment_Method::CARD ],
```
FangedParakeet, line 2060: `WC_Payments_UPE_Checkout` will always be initialised with an instance of the card gateway.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: (Just removed the TODO in 077df16a39ef3b335c7739a6b7cadbb16512856b, BTW, because this is basically TO-DONE and will soon be TO-GONE.)

client/checkout/blocks/upe-deferred-intent-creation/payment-elements.js
```diff
@@ -19,6 +19,14 @@ const PaymentElements = ( { api, ...props } ) => {
...
 	const currency = getUPEConfig( 'currency' ).toLowerCase();
 
+	const isStripeLinkEnabled =
+		getUPEConfig( 'paymentMethodsConfig' ).link !== undefined &&
+		getUPEConfig( 'paymentMethodsConfig' ).card !== undefined;
+	const paymentMethodTypes = [ props.paymentMethodId ];
+	if ( 'card' === props.paymentMethodId && isStripeLinkEnabled ) {
+		paymentMethodTypes.push( 'link' );
+	}
+
```
timur27: Would it be easier to read if we moved this code to the function e.g. `getPaymentMethodTypes` and call it on line `62`?

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -71,7 +78,69 @@ const PaymentProcessor = ( {
...
+					);
+				},
+				shipping_fields: BLOCKS_SHIPPING_ADDRESS_FIELDS,
+				billing_fields: BLOCKS_BILLING_ADDRESS_FIELDS,
+				complete_billing: () => {
+					return (
+						null !== document.getElementById( 'billing-address_1' )
+					);
+				},
+			} );
```
timur27: This code seems to be identical to the one under `client/checkout/blocks/upe-split-fields.js`. So my question is if we want to merge them or, given the single gateway experience in the future, do we want to leave them duplicated? I'm fine with any of the options which is why am not suggesting anything at the moment and want to hear your opinion on this. Thanks!

timur27 (commented):
I really appreciate the work you've done with Stripe Link and also with polishing other things which fairly seemed to not be optimal!

While I have a few files yet to review which I'd like to complete tomorrow morning, I'd like to submit a few pending items and also share some testing results with you! 

The first thing that I'd like to discuss is the behavior, when for a user registered with Link not all input fields are filled out on the Blocks checkout: 

https://github.com/Automattic/woocommerce-payments/assets/22319444/15258072-9799-4d9a-9aec-3fd4cb8753c4

Could you please let me know if you experience the same? I created this user in the incognito mode but I'm unsure if this is important. 

Then for the other user I've registered in a non-incognito mode the fields are populated, but when trying to place an order, I'm facing an error: 


https://github.com/Automattic/woocommerce-payments/assets/22319444/a2fdf7a6-7fd9-4ecc-9b87-6984711b21ec


Another smallish thing that caught my eye was the _Save payment method for future checkouts_ checkbox on the shortcode checkout, which appears with the Stripe Link being enabled. My gut feeling is it's unexpected, but I'd appreciate if you could confirm it or decline. And yeah, this is something I discovered on the split UPE as well, so I'm fine if we'll keep it out of the current scope. 

image

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -184,6 +189,49 @@ function appendSetupIntentToForm( form, confirmedIntent ) {
 	form.append( input );
 }
 
+function maybeEnableStripeLink( api, paymentMethodType ) {
+	if ( 'card' === paymentMethodType && 'link' in gatewayUPEComponents ) {
```
timur27: Can we move this check in a separate function given that it's also used in line `146`?

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -218,6 +266,7 @@ export async function mountStripePaymentElement( api, domElement ) {
 	const upeElement =
 		gatewayUPEComponents[ paymentMethodType ].upeElement ||
 		( await createStripePaymentElement( api, paymentMethodType ) );
+	maybeEnableStripeLink( api, paymentMethodType );
```
timur27: Since we're mounting multiple elements, would it make sense to invoke this function from `event-handlers.js::maybeMountStripePaymentElement` but outside of the `foreach`? 


```js
	function maybeMountStripePaymentElement() {
		if (
			$( '.wcpay-upe-element' ).length &&
			! $( '.wcpay-upe-element' ).children().length
		) {
			$( '.wcpay-upe-element' )
				.toArray()
				.forEach( ( domElement ) =>
					mountStripePaymentElement( api, domElement )
				);
                         maybeEnableStripeLink( api ); 
		}
	}
```

We could then remove the check for the card payment method within the `maybeEnableStripeLink` as the initialization attempt will take place only once. We would then also replace `paymentMethodType` to `card` within than function.

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
```diff
@@ -184,6 +189,49 @@ function appendSetupIntentToForm( form, confirmedIntent ) {
 	form.append( input );
 }
 
+function maybeEnableStripeLink( api, paymentMethodType ) {
```
timur27: I'm wondering if it would make sense to move the function which is initializing Stripe Link from the script mainly responsible for payment processing. At the same time, I would agree if someone says this script isn't about payment processing anymore anyway, since it also covers payment elements initialization. Maybe we should then re-name `payment-processing.js` so that it reflects better what its responsibilities?

client/checkout/utils/upe.js
```diff
@@ -300,3 +301,35 @@ export const getStripeElementOptions = (
...
+
+	//Handle StripeLink button click.
+	jQuery( '.wcpay-stripelink-modal-trigger' ).on( 'click', ( event ) => {
+		event.preventDefault();
+		// Trigger modal.
+		linkAutofill.launch( {
+			email: jQuery( '#email' ).val(),
+		} );
+	} );
+};
```
timur27: This code seems to be easily migratable from jQuery to vanilla JS. If we would re-write it in JS, the testing will be way easier. Given that this script has a pretty high test coverage at the moment (`client/checkout/utils/test/upe.test.js`), it would be nice to test this function as well.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -754,7 +755,11 @@ public function process_payment( $order_id ) {
 				];
 			}
 
-			UPE_Payment_Gateway::remove_upe_payment_intent_from_session();
+			if ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
```
timur27: For the deferred intent creation UPE, no intent will be stored in the session - thus should we consider this UPE type in this logic? Same for lines `864-868`.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1371,13 +1399,35 @@ private function get_payment_methods_from_request() {
 			$payment_methods = [ str_replace( 'woocommerce_payments_', '', $upe_payment_method ) ];
 		} elseif ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 			$payment_methods = [ 'card' ];
+			if ( WC_Payments_Features::is_upe_deferred_intent_enabled() &&
+				in_array( Payment_Method::CARD, $payment_methods, true ) &&
```
timur27: This line will be always `true` due to the line `1401` - should we exclude this check from the `if` statement?

includes/core/server/request/class-create-and-confirm-setup-intention.php
```diff
@@ -83,4 +83,33 @@ public function set_payment_method( string $payment_method_id ) {
...
 		$this->set_param( 'payment_method', $payment_method_id );
 	}
+
+	/**
+	 * Payment methods setter.
+	 *
+	 * @param  array $payment_methods               An array of payment methods that might be used for the payment.
+	 * @throws Invalid_Request_Parameter_Exception  When there are no payment methods provided.
+	 */
+	public function set_payment_method_types( array $payment_methods ) {
```
timur27, line 94: As part of https://github.com/Automattic/woocommerce-payments/pull/6656, I extended `Payment_Information` class with the `$payment_method_stripe_id` variable, which has the same value type in it. It seems to me that we might want to keep these consistent and either rename `$payment_method_stripe_id` in the `Payment_Information` class to e.g. `$payment_method_type` or re-name `payment_method_types` to `payment_method_stripe_ids` here. What do you think?

client/checkout/blocks/upe-split.js
```diff
@@ -66,37 +66,26 @@ const api = new WCPayAPI(
...
+				upeMethods,
+				api,
+				upeConfig.testingInstructions
+			),
+			edit: getUPEFields(
+				upeName,
+				upeMethods,
+				api,
+				upeConfig.testingInstructions
+			),
```
timur27: Nice one, I love it and really appreciate the improvement you made here! 🥇

client/checkout/blocks/upe-split-fields.js
```diff
@@ -123,49 +106,24 @@ const WCPayUPEFields = ( {
...
-						( event ) => {
-							event.preventDefault();
-							// Trigger modal.
-							linkAutofill.launch( {
-								email: jQuery( '#email' ).val(),
-							} );
-						}
-					);
-				},
+				show_button: blocksShowLinkButtonHandler,
```
timur27, line 118: Thanks for removing the duplicates in this place! Could we do the same for the `upe-fields.js` lines `119-140`?

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -31,6 +31,9 @@ jQuery( function ( $ ) {
 			forceNetworkSavedCards: getUPEConfig( 'forceNetworkSavedCards' ),
 			locale: getUPEConfig( 'locale' ),
 			isUPEEnabled: getUPEConfig( 'isUPEEnabled' ),
+			isStripeLinkEnabled:
+				getUPEConfig( 'paymentMethodsConfig' ).link !== undefined &&
+				getUPEConfig( 'paymentMethodsConfig' ).card !== undefined,
```
timur27: It's a good candidate for wrapping it into a boolean function within `client/checkout/utils/upe.js`, to use [here](https://github.com/Automattic/woocommerce-payments/pull/6784/files#diff-d9e0d3ccf656775c42c68e2f0b68e0ecbe0ca1a5728f321d843298600b450d58R22-R24) as well.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1213,6 +1231,16 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
 				$request->set_metadata( $metadata );
+
+				if (
+					WC_Payments_Features::is_upe_deferred_intent_enabled() &&
+					Payment_Method::CARD === $this->get_selected_stripe_payment_type_id() &&
+					in_array( Payment_Method::LINK, $this->get_upe_enabled_payment_method_ids(), true )
+					) {
+					$request->set_payment_method_types( $this->get_payment_methods_from_request() );
+					$request->set_mandate_data( $this->get_mandate_data() );
+				}
```
timur27, line 1250: Could we add some tests to this code where we'd expect `set_mandate_data` and `set_payment_method_types` to be called under certain circumstances?

timur27 (commented):
I went through the code and a bit more testing. Thanks for the refactor you did for other UPEs while working on this issue! I've left a couple of comments to the code. 

Besides that, subscription payments work both with automatic and manual renewals. I'm now able to pay with the Stripe Link account on the _Pay for order_ page, but there's only email listed and no Link button - I think it's expected though, because there's no email address input field on that page. 

One thing I've noticed is when you pay by using your Stripe Link account for a subscription, your token is saved and added to the list of saved payment methods even if it already exists there, eventually bringing us to the following state on the checkout/pay for order pages.

image


I'd expect only one entry to be on the list, since I'm paying with the same Link email multiple times. Could you please double-check and let me know if I'm missing or misunderstanding something? I'm happy to get back to the review after we'll align on the pending threads. Thank you!

client/checkout/blocks/upe-deferred-intent-creation/payment-elements.js
FangedParakeet: Created a new function for this purpose in 2ee6b6ebea00cc109bfa8c8c52069e2c2eb633a3. 👍

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
FangedParakeet: Refactored this in 2ee6b6ebea00cc109bfa8c8c52069e2c2eb633a3. 👍

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
FangedParakeet: `maybeEnableStripeLink` will have to stay inside `payment-processing.js`, because it requires access to the Stripe Elements object that is only available therein. However, I've refactored the rest tailored to your suggestion and moved the declaration of `maybeEnableStripeLink` to `event-handlers.js`, as you've suggested. See 2ee6b6ebea00cc109bfa8c8c52069e2c2eb633a3.

client/checkout/classic/upe-deferred-intent-creation/payment-processing.js
FangedParakeet: > Maybe we should then re-name payment-processing.js so that it reflects better what its responsibilities?

The definition of `maybeEnableStripeLink` has to stay here; you could argue it's still pretty relevant to payment processing. I've moved it's declaration to `event-handlers.js`, where it is actually initialised. See 2ee6b6ebea00cc109bfa8c8c52069e2c2eb633a3.

client/checkout/utils/upe.js
FangedParakeet: You're right, I was just being lazy, lawl. Amended in 1bea7e74f96d1a4e95704f32a1e2bb695ad09803.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Amended in 2616a6febeb6e3fcbfdc2ce78f1a94bd8b6226c7...Though, having another think about it, in the previous implementation we cleared the legacy UPE intent session regardless of which feature flag is enabled. Perhaps now we should just nuke all session data--legacy UPE and split UPE--to be safer rather than sorry? Let me know which option you'd prefer.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Makes sense. Amended in 8534408f91efd4db7fd3faf82d06600a2d3c8733.

client/checkout/blocks/upe-split-fields.js
FangedParakeet: Whoops, forgot to do this before. Amended in d782bdf2a7e3c7dca8fd4c0312275dc12993deff.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
FangedParakeet: Good idea. Extracted this code into a util in f7aa8c7ed0e1e7d9cdd70a4d94ccede31ba2e211 and got a LOT of mileage out of it.

includes/core/server/request/class-create-and-confirm-setup-intention.php
FangedParakeet: Personally I concede that `payment_method_stripe_id` might be more accurate, but `payment_method_types` remains consistent with the `payment_method_types` [Payments Server parameter](https://github.com/Automattic/woocommerce-payments-server/blob/trunk/server/wp-content/rest-api-plugins/endpoints/wcpay/class-setup-intents-controller.php#L50) on the setup intentions endpoint where this request is sent to. 

Using `payment_method_types` also tries to keep this `Create_And_Confirm_Setup_Intention` class consistent with its sibling `Create_and_Confirm_Intention` request class, which uses two setters--`set_payment_methods` and `set_payment_method_types`--(for some reason) to set the same field. Of the two, I felt `set_payment_method_types` was the less confusing method name to duplicate.

client/checkout/blocks/upe-deferred-intent-creation/payment-processor.js
FangedParakeet: This implementation already used a lot of copied code between gateway implementations. I tried to reduce this as much as possible, but sadly React would not let me remove the `fill_field_method` to another function, because `customerData` manipulates state data, so must be defined within the `useEffect` block. Consequently, `enableStripeLinkPaymentMethod` must be declared like this in all its hideous glory. 

It made me sad, but once we eventually have a single gateway, there won't be any more duplicates and that makes me a little more glad. 🤷

FangedParakeet:
>Another smallish thing that caught my eye was the Save payment method for future checkouts checkbox on the shortcode checkout, which appears with the Stripe Link being enabled.

This is basically a symptom of Link being made available for use with WC Subscriptions in #4781. The only way (or at least the way that is currently convention) to make a payment method compatible with subscriptions is to save it as a WC Token and then use this saved payment token as the payment method when processing the renewal. This has the side-effect of allowing payment methods to be saved as payment tokens.

Link complicates this further, because while the payment element will provide a completed event, there is no way of knowing whether the payment method within is a Link payment method or a regular card payment method--the payment method type on the event will always be `card` either way. This means that we cannot hide the saved payment checkbox for Link payment method without hiding it for the card payment gateway entirely, since AFAICT there is no way to distinguish between these two payment methods on the client while processing the payment. (With setup intents, most irritatingly, there is no way to distinguish between Link and card payment methods even _after_ we have processed the payment, as the setup intent has no charge and the payment method type of a Link setup intent will also be `card` 😵 ).

>One thing I've noticed is when you pay by using your Stripe Link account for a subscription, your token is saved and added to the list of saved payment methods even if it already exists there, eventually bringing us to the following state on the checkout/pay for order pages.

This is a mixture of our own fault and Stripe's. If Stripe stores a payment method as a new separate saved payment method, then we will treat it the same as well. In this case, even though the Link payment method is the same, Stripe is still saving it as a new payment method. You can repeat this trick with saved cards as well: if you purchase multiple subscriptions using the same credit card details (card number, expiry, and CVV), you will still find it saved in WC as a new payment method and will find yourself with many duplicates of what appears to be the same payment token. 

I don't think there is a way to resolve this without making changes to WC Core, but if you like I can create a new issue to explore this further?

client/checkout/classic/upe-split.js
```diff
@@ -44,6 +45,10 @@ import {
 	appendFingerprintInputToForm,
 } from '../utils/fingerprint';
 import PAYMENT_METHOD_IDS from 'wcpay/payment-methods/constants';
+import {
+	SHORTCODE_SHIPPING_ADDRESS_FIELDS,
+	SHORTCODE_BILLING_ADDRESS_FIELDS,
+} from '../../constants';
```
timur27: Shouldn't it be? 

```suggestion
} from '../constants';
```

Does it compile with the current version?

client/checkout/classic/upe.js
```diff
@@ -33,6 +34,10 @@ import {
 	appendFingerprintInputToForm,
 } from '../utils/fingerprint';
 import PAYMENT_METHOD_IDS from 'wcpay/payment-methods/constants';
+import {
+	SHORTCODE_SHIPPING_ADDRESS_FIELDS,
+	SHORTCODE_BILLING_ADDRESS_FIELDS,
+} from '../../constants';
```
timur27:
```suggestion
} from '../constants';
```

client/checkout/classic/upe-split.js
FangedParakeet: Fixed in 174e72c75c94d482af17642e99c98dea7de0db28.

client/checkout/classic/upe.js
FangedParakeet: See 174e72c75c94d482af17642e99c98dea7de0db28.

includes/core/server/request/class-create-and-confirm-setup-intention.php
timur27: Fully agree with everything you just said. We then need to align `Payment_Information` so it also has `payment_method_types` and not `payment_method_stripe_ids`. We can do it separately, but I'd like us to align this code after your changes.

includes/class-wc-payment-gateway-wcpay.php
timur27: I fully agree - it just that the FF check for the deferred UPE wasn't needed. Other than that - this enhancement is more than acceptable and thanks for improvements here!

includes/core/server/request/class-create-and-confirm-setup-intention.php
timur27: We decided f2f to leave the current implementation to avoid making it less readable.

S-code652 (approved):

FangedParakeet:
Created #6881 to address issues with blocks checkout and #6882 to address improvements to integration tests.

includes/class-wc-payment-gateway-wcpay.php
timur27: To be addressed in https://github.com/Automattic/woocommerce-payments/issues/6882

timur27 (commented):
All the previous threads are now successfully closed 👍 Thanks for going through them, Samir! I've found two more things not fully working as expected: 

1️⃣ Stripe Link isn't initilized with more than one payment method enabled. I'd wait for https://github.com/Automattic/woocommerce-payments/issues/6790 to be resolved and my hunch is the issue with Stripe Link should be resolved as well.

2️⃣ When using Stripe Link to pay for a subscription and thus saving the payment method for the future use, we get an error on the _My account_ page while listing all the available payment methods.
 
<details>
  <summary>Here's how it looks</summary>
  image
</details>

I'll try to get back to the 2️⃣nd issue after I finish https://github.com/Automattic/woocommerce-payments/issues/6790, since you're AFK this week.

timur27:
### Latest changes 
1. The first problem described [here](https://github.com/Automattic/woocommerce-payments/pull/6784#pullrequestreview-1556461260) was fixed in https://github.com/Automattic/woocommerce-payments/pull/6784/commits/de49868c2bd72b2efa0dcd76e5e541cd69afe944.
2. The second problem described [here](https://github.com/Automattic/woocommerce-payments/pull/6784#pullrequestreview-1556461260) was fixed in https://github.com/Automattic/woocommerce-payments/pull/6784/commits/cbcbef205af542affbb7ffd680a547e7319d51a9.

### Testing outcome
1. Regular product checkout (Blocks & shortcode) ✅ 
3. Checkout with subscription 
   * Simple subscription purchase ✅ 
   * Simple subscription renewal from the _My account -> Subscriptions_ page ✅
   * Simple subscription manual renewal from the admin panel ✅
   * Simple subscription renewal with the `woocommerce_scheduled_subscription_payment` action ✅ 
   * Free-trial subscription purchase, manual renewal from the admin panel and with `woocommerce_scheduled_subscription_payment` action ✅ 
4. `mandate_data` is attached to the intent creation request ✅ 
5.  Able to pay for an order on the _Pay for order_  page ✅ 
6. List Stripe Link saved payment method on the _My account -> Payment methods_ page ✅ 


### Follow-ups
1. https://github.com/Automattic/woocommerce-payments/issues/6881
2. https://github.com/Automattic/woocommerce-payments/issues/6882
3. Create an issue for _Payment gateways that are not supported by a currency are still not filtered out on the shortcode checkout page_
4. Create an issue/investigate _Saved Stripe Link tokens are listed as **Saved token for woocommerce_payments** on Blocks checkout page_
5. Update release testing instructions or critical flows testing instructions with new functionality for Stripe Link
6. With Stripe Link enabled, its saved payment method is visible on the checkout page even with a non-supported currency (e.g., EUR). They can also be used to perform a successful payment.

includes/class-wc-payments-token-service.php
```diff
@@ -264,6 +265,21 @@ public function get_account_saved_payment_methods_list_item_sepa( $item, $paymen
...
+	 * Controls the output for Stripe Link on the My account page.
+	 *
+	 * @param  array                                        $item          Individual list item from woocommerce_saved_payment_methods_list.
+	 * @param  WC_Payment_Token|WC_Payment_Token_WCPay_Link $payment_token The payment token associated with this method entry.
+	 * @return array                                        Filtered item
+	 */
+	public function get_account_saved_payment_methods_list_item_link( $item, $payment_token ) {
+		if ( WC_Payment_Token_WCPay_Link::TYPE === strtolower( $payment_token->get_type() ) ) {
+			/* translators: %s is a registered Stripe Link email. */
+			$item['method']['brand'] = sprintf( esc_html__( 'Stripe Link email %s', 'woocommerce-payments' ), esc_html( $payment_token->get_email() ) );
```
mdmoore, line 278: I wonder if [`sanitize_email()`](https://developer.wordpress.org/reference/functions/sanitize_email/) would be a better choice for the email address here. I don't think `esc_html()` will have much of an effect on email addresses whereas `sanitize_email()` is designed specifically to validate and sanitize email addresses. 

```suggestion
			$item['method']['brand'] = sprintf( esc_html__( 'Stripe Link email %s', 'woocommerce-payments' ), sanitize_email( $payment_token->get_email() ) );
```

timur27: That's a good point, I didn't think of sanitizing the output because the email value should have been sanitized on the input ~~([here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L1411) I believe)~~. [Here's more details on where the input is sanitized.](https://github.com/Automattic/woocommerce-payments/pull/6784#discussion_r1284413654)

So on one side, at first glance, I wouldn't expect sanitization here as sanitization is supposed to be done while saving the value. On the other side, while being redundant, sanitization still probably shouldn't bring any data loss/corruption since the value was already sanitized before and thus can be used? But if used, maybe still apply `esc_html` on top of it, to be sure that eventual HTML tags are considered plain text.

Let me know what you think about all this! 👍

timur27: @mdmoore Another thing I started to wonder right now is if with `esc_html` I'm not escaping it too early? Eventually, this value will be still escaped [in WooCommerce Core](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/templates/myaccount/payment-methods.php#L51), so my escaping here might be redundant. I think it's usually suggested that the best way is to escape late (which, in our case is WC Core which does the latest escape possible), to avoid the returned value updates somewhere in between. Still, I'd personally leave it since this escaping won't hurt.

timur27: Speaking input sanitization: billing email is being saved upon order placement by the `/customer` API call, the data for which is [being prepared by the `WC_Payments_Customer_Service::map_customer_data`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-customer-service.php#L311). At this stage, the email coming from the order is already sanitized and below is why.

The order is initially created by WooCommerce Core as [one of the first checkout processing steps.](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-checkout.php#L1256C5-L1256C5). 

The email field is being set in the order [here](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-checkout.php#L415). Then the `class-wc-order.php` has [the setter](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-order.php#L1265-L1270) which [sanitizes the email value](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-order.php#L1269C48-L1269C72).

mdmoore: Yeah, it make sense that this email is already sanitized when it's set. I think you're safe to use the email as is here. 

>Eventually, this value will be still escaped [in WooCommerce Core](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/templates/myaccount/payment-methods.php#L51), so my escaping here might be redundant

I agree here as well. It would be escaped twice which isn't a problem, but even without escaping here, we're following the escape late principle and would still be safe IMO. So I'd say it's up to you whether you want to escape here or just pass the email through.

timur27: Yep, glad we're aligned now. I'll keep the escaping here and thus close the thread without new commits. Thanks for the help!

timur27 (approved):

PR #6791, opened 2023-07-17
Endpoint for creating a payment intent

### LINKED ISSUE #6705: Extensibility | Extend Payments acceptance via REST - Plugin changes
Check if the  POST `payments/payment_intents` can be added to the plugin.

jessy-p:
Fixes #6705 

- Introduced endpoint for create payment_intent
- This is for use in  PoC pcreKM-1U2-p2
- If approved, it can be merged to develop and become part of the plugin OR this branch may be used for the PoC

#### Testing instructions
- The endpoint `payments/payment_intents` can be used for creating a `payment_intent`

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6791` or branch name `update/payment-intent-create-endpoint` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/payment-intent-create-endpoint) :rocket:

---

Build info: 

- Latest commit: 3144fe9046c651eb6b598a35eb6247fe9b0870e4
- Build time: 2023-10-04 19:08:22 UTC

marcinbot (approved):
Looks straightforward enough :shipit:

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -224,6 +224,23 @@ public function get_intent( $intent_id ) {
...
+	 *
+	 * @throws API_Exception Error creating intent.
+	 */
+	public function create_intent( $request ) {
+		return $this->request(
+			$request->get_params(),
+			self::INTENTIONS_API,
+			self::POST
+		);
+	}
```
htdat: IMHO, we have already had `Create_Intention` core request class for the same purpose, we would not introduce this method? 

https://github.com/Automattic/woocommerce-payments/blob/08bd8a0d5a60e027ee155148e7f64b966234705b/includes/core/server/request/class-create-intention.php#L18

jessy-p:
@htdat  Changed to use `Create_Intention` and added tests. Please have a look.

changelog/update-payment-intent-create-endpoint
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Added payment intent creation endpoint
```
htdat: Nitpick: 
```suggestion
Add payment intent creation endpoint
```

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +56,35 @@ public function get_payment_intent( $request ) {
...
 	}
+
+	/**
+	 * Create a payment intent.
+	 *
+	 * @param WP_REST_Request $request data about the request.
+	 */
+	public function post_payment_intent( $request ) {
+		try {
+			$metadata = $request->get_param( 'metadata' );
```
htdat: It seems that you're using the structure of PaymentIntent itself for this code. However, here I think we're going to receive a request to create a new payment intent so I think we just need to have `order_id` for requests. 

Note: `order_id` is really the post ID (int) while `order_number` can be a string in Woo. 

`metadata` and `level3` params are not necessary for this request. We should build up them from what we have already had (order_id and order object): 

https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/class-wc-payment-gateway-wcpay.php#L1115-L1116

https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/class-wc-payment-gateway-wcpay.php#L2918-L2921

Also, in terms of security, we should not allow to update the whole `metadata` of payment intent directly from REST. We have some critical functions (fraud, integrity checks before processing) rely on it.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +56,35 @@ public function get_payment_intent( $request ) {
...
+			$order    = wc_get_order( $order_id );
+
+			$wcpay_server_request = Create_Intention::create();
+			$currency             = strtolower( $order->get_currency() );
+			$amount               = WC_Payments_Utils::prepare_amount( $order->get_total(), $currency );
+			$wcpay_server_request->set_currency_code( $currency );
+			$wcpay_server_request->set_amount( $amount );
+			$wcpay_server_request->set_metadata( $metadata );
+			$wcpay_server_request->set_customer( $request->get_param( 'customer' ) );
+			$wcpay_server_request->set_level3( $request->get_param( 'level3' ) );
```
htdat: `metadata` and `level3` params are not necessary. We should build up them from what we have already had (order_id and order object): 

https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/class-wc-payment-gateway-wcpay.php#L1115-L1116

https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/class-wc-payment-gateway-wcpay.php#L2918-L2921

htdat (dismissed):
I tried to understand the final purpose of this new endpoint. I think it's equally to [`UPE_Payment_Gateway::create_payment_intent`](https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/payment-methods/class-upe-payment-gateway.php#L310-L310)? 

If this is just for testing, it's OK to keep it like now. However, if we decided to add this into `develop` branch, we might carefully review it in terms of security too. 

And finally, I'd like to have a shape of REST request, can be a `curl` 
command to test this endpoint. The reason is that this is not a simple GET request, it's a POST requests with a few parameters.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Thanks for pointing me to building metadata and level3 from order data. It reduces the code to call the API on the PoC side as well 👍

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Fixed.

jessy-p:
> I tried to understand the final purpose of this new endpoint. I think it's equally to [UPE_Payment_Gateway::create_payment_intent](https://github.com/Automattic/woocommerce-payments/blob/232acdfa2ce16abe1bee8efbd86e31e143909c5e/includes/payment-methods/class-upe-payment-gateway.php#L310-L310)?

This is for use in  PoC pcreKM-1U2-p2. 

>If this is just for testing, it's OK to keep it like now. However, if we decided to add this into develop branch, we might carefully review it in terms of security too.

@marcinbot  what do you think about merging it and security aspects?
I will check with CS team as well, who will be using this PoC, if they are OK with leaving it un-merged, will do so.

jessy-p:
> And finally, I'd like to have a shape of REST request, can be a curl
command to test this endpoint. The reason is that this is not a simple GET request, it's a POST requests with a few parameters.

```
curl -X POST https://<site_host>/wp-json/wc/v3/payments/payment_intents -u <customer_key>:<customer_secret> -H "Content-Type: application/json" -d '{"confirm":"true","payment_method":"<card_id/pm_id>","customer":"<cus_id>","order_id":"<order_id>","capture_method":"automatic","description":"Online Payment for Order <order_id> for <site_id>","payment_method_types":["card"],"is_platform_payment_method":false}'
```

You can create an order on the store and a payment method on Stripe before using the above request.

jessy-p:
@htdat Addressing the suggestions around parameter simplification. (Edit by @htdat: suggestions are here p1690350125782319/1690349903.234759-slack-C01BPL3ALGP). 

**customer** --> Infer from order_id. Currently I am using the deprecated `/create_customer` API endpoint which uses the similar code as `manage_customer_details_for_order` which is a protected method of the gateway. As it is for a PoC, I am not changing the method to public at this time. For now the PoC can use `/create_customer` endpoint and can pass the customer to the `payment_intents` endpoint.

**capture_method** -> I'll avoid this.

As suggested, reducing the parameters passed to bare minimum, an example curl command would be:
`curl -X POST https://<site_host>/wp-json/wc/v3/payments/payment_intents -u <customer_key>:<customer_secret> -H "Content-Type: application/json" -d '{"payment_method":"<card_id/pm_id>","customer":"<cus_id>","order_id":"<order_id>"}`

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +74,47 @@ public function get_payment_intent( $request ) {
...
+			$metadata = [
+				'customer_name'  => $name,
+				'customer_email' => $email,
+				'site_url'       => esc_url( get_site_url() ),
+				'order_id'       => $order->get_id(),
+				'order_number'   => $order->get_order_number(),
+				'order_key'      => $order->get_order_key(),
+				'payment_type'   => 'single',
+			];
+
```
htdat: It might be that you missed this comment in my previous review 
p1690350125782319/1690349903.234759-slack-C01BPL3ALGP

> `metadata` are built for `$wcpay_server_request` right here. However, we have already have a method for that purpose `WC_Payment_Gateway_WCPay::get_metadata_from_order` - which is currently `protected` however we can change it to `public`. And ideally, all of these would become Util methods, but this is not a focus of this PR.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +74,47 @@ public function get_payment_intent( $request ) {
...
+			$wcpay_server_request->set_amount( $amount );
+			$wcpay_server_request->set_metadata( $metadata );
+			$wcpay_server_request->set_customer( $request->get_param( 'customer' ) );
+			$wcpay_server_request->set_level3( $this->gateway->get_level3_data_from_order( $order ) );
+			$wcpay_server_request->set_payment_method( $request->get_param( 'payment_method' ) );
+			$wcpay_server_request->set_payment_method_types( [ 'card' ] );
+			$wcpay_server_request->set_capture_method( WC_Payments::get_gateway()->get_option( 'manual_capture' ) && ( 'yes' === WC_Payments::get_gateway()->get_option( 'manual_capture' ) ) );
+
+			$intent = $wcpay_server_request->send( 'wcpay_create_intent_request', $order );
+			return rest_ensure_response( $intent );
```
htdat: I tested the `curl` provided command, it's working well. However, in Woo store, the order is not marked as `processing`.

**_Update: It turns out that I enabled manual_capture on my testing site. Without it, the testing order is marked as processing, which is handled via webhook._** 

It seems that after sending serve requests, like  `WC_Payment_Gateway_WCPay::process_payment`, this would handle multiple outcomes such as:

- `succeeded`: mark the Woo order as `processing` or `completed`. 
- `requires_action`: 3DS card is used, lay out the info and ask shoppers to verify again. 
- another status for manual capture, etc. 

This is not a blocker for PoC but it's worth considering for next steps.

htdat (commented):
Looks much better, and thank you for addressing my feedback. However, at the moment, the order_id is not marked as `processing` or `completed` when the PaymentIntent is `succeeded`. See more in my inline comment.

**_Update: It turns out that I enabled `manual_capture` on my testing site. Without it, the testing order is marked as `processing`, which is handled via webhook._**

naman03malhotra:
Since this PR was part of the POC, let's be extra careful while merging it into production. @jessy-p , do you plan a 2nd review from Architects before merging?

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +74,47 @@ public function get_payment_intent( $request ) {
 
 		return $this->forward_request( 'get_intent', [ $payment_intent_id ] );
 	}
+
+	/**
+	 * Create a payment intent.
+	 *
+	 * @param WP_REST_Request $request data about the request.
+	 */
+	public function post_payment_intent( $request ) {
```
zmaglica: Nitpick. I would rename this function to create_payment_intent, but feel free to ignore my comment.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +74,47 @@ public function get_payment_intent( $request ) {
...
+				'customer_name'  => $name,
+				'customer_email' => $email,
+				'site_url'       => esc_url( get_site_url() ),
+				'order_id'       => $order->get_id(),
+				'order_number'   => $order->get_order_number(),
+				'order_key'      => $order->get_order_key(),
+				'payment_type'   => 'single',
+			];
+
+			$wcpay_server_request = Create_And_Confirm_Intention::create();
```
zmaglica, line 97: Have you considered using `SetupIntents` instead of 'regular' intents, since we might charge this intent later on? I am not saying that this is wrong, I am just curious would `SetupIntents` be a better option here.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -34,6 +42,26 @@ public function register_routes() {
 				'permission_callback' => [ $this, 'check_permission' ],
 			]
 		);
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			[
+				'methods'             => WP_REST_Server::CREATABLE,
+				'callback'            => [ $this, 'post_payment_intent' ],
+				'permission_callback' => [ $this, 'check_permission' ],
```
zmaglica, line 53: I do know that this is POC, but if we are planning to add it to the production, could we use WordPress REST API Schemas and perhaps add validation like we do in reports API endpoints?

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Discussed here - pcreKM-20q-p2#comment-3005

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: That would require to either change all instances (including the current payment processing) into that format, or create a separate code for this instance, both of which does not seem advisable considering the sensitive area (payment processing)

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Done

jessy-p:
Modified to accomodate changes to Request classes, and addressed the review comments.
Please review and approve if ready to merge -  @marcinbot  @zmaglica

includes/admin/class-wc-rest-payments-payment-intents-controller.php
zmaglica: I think I picked the wrong line for this comment. I was referring to usage of JSON schema like we [did](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/reports/class-wc-rest-payments-reports-transactions-controller.php#L257-L386) with reports API where we define what properties API can return and what type will it be.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +76,49 @@ public function get_payment_intent( $request ) {
...
+	/**
+	 * Create a payment intent.
+	 *
+	 * @param WP_REST_Request $request data about the request.
+	 */
+	public function create_payment_intent( $request ) {
+		try {
+
+			$order_id = $request->get_param( 'order_id' );
+			$order    = wc_get_order( $order_id );
```
zmaglica, line 92: We need to handle a situation when order doesn't exist.

marcinbot (commented):
Looks good, but I agree with Dat and Zvonimir regarding the comments here: https://github.com/Automattic/woocommerce-payments/pull/6791/files#r1275945608 and here: https://github.com/Automattic/woocommerce-payments/pull/6791/files#r1344029038

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Done

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Fixed

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +77,286 @@ public function get_payment_intent( $request ) {
...
+			'type'       => 'object',
+			'properties' => [
+				'id'       => [
+					'description' => __( 'ID for the payment intent.', 'woocommerce-payments' ),
+					'type'        => 'string',
+					'context'     => [ 'view' ],
+				],
+				'amount'   => [
+					'description' => __( 'The amount of the transaction.', 'woocommerce-payments' ),
+					'type'        => 'string',
```
zmaglica: Nitpick: This should be integer.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +77,286 @@ public function get_payment_intent( $request ) {
...
+					'context'     => [ 'view' ],
+				],
+				'currency' => [
+					'description' => __( 'The currency of the transaction.', 'woocommerce-payments' ),
+					'type'        => 'string',
+					'context'     => [ 'view' ],
+				],
+				'created'  => [
+					'description' => __( 'Timestamp for when the payment intent was created.', 'woocommerce-payments' ),
+					'type'        => 'string',
```
zmaglica, line 155: Nitpick. You can add` 'format'  => 'date-time'`

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +77,286 @@ public function get_payment_intent( $request ) {
...
+							],
+						],
+						'payment_method'         => [
+							'description' => 'The payment method associated with this charge.',
+							'type'        => 'string',
+							'context'     => [ 'view' ],
+						],
+						'application_fee_amount' => [
+							'description' => 'The application fee amount.',
+							'type'        => 'string',
```
zmaglica: Nitpick: Maybe this could be integer.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
```diff
@@ -46,4 +77,286 @@ public function get_payment_intent( $request ) {
...
+							'context'     => [ 'view' ],
+						],
+						'status'                 => [
+							'description' => 'The application fee amount.',
+							'type'        => 'string',
```
zmaglica, line 299: Nitpick: Maybe this could be integer.

zmaglica (approved):
🚢  Looks good to me. Left a couple of nitpick, but they can be ignored.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Its actually a timestamp, which I'll try and fix later.

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Done

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: https://github.com/Automattic/woocommerce-payments-meta/issues/286

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: Done

includes/admin/class-wc-rest-payments-payment-intents-controller.php
jessy-p: MIstake, it was not an amount, corrected the description. It is a status string

PR #6805, opened 2023-07-18
RPP: Create a payment service class

### LINKED ISSUE #6674: Payment Process: Create a payment service class
### Description

Create a new payment service class, which will be used for processing payments.

This service will be used as a proxy between the payment process, and gateways classes. In the end gateway should just be able to call `$payment_service->process_payment()` instead.

### Acceptance criteria

- The class is in a new `src` directory.
- `npm run release` copies the directory into the release.
- Add an empty `index.php` to the new directory, preventing listing.
- Use the `WCPay` or `WCPay\Service` namespace for the class.

### Testing instructions

None. The class is not functional.

### Dev notes

Consider whether to keep the class internal or external.

### Additional context

Project: paJDYF-9hL-p2

htdat:
Note: RPP = Re-engineering Payment Process. 

Fixes #6674

Relevant links: 

- https://betadownload.jetpack.me/data/woocommerce-payments/rpp_6674-create-payment-service-class/woocommerce-payments-dev.zip
- https://jurassic.ninja/wp-json/jurassic.ninja/jetpack-beta/plugins/woocommerce-payments/branches

- Add a new `src` folder, and the new `PaymentService` class.

#### Testing instructions

- Ensure no CI tests fail. 
- Check the build that can be found on https://jurassic.ninja/wp-json/jurassic.ninja/jetpack-beta/plugins/woocommerce-payments/branches - searching for the branch of this PR `rpp/6674-create-payment-service-class`. Ensure that the `commit` for this file is the same for the last commit in this PR. At the moment, it is `a026e9870e061fdaf8412f1c7755c9072b4abe8b`
- In the build, ensure the files in the new `src` folder are included. 

The build can be found on https://jurassic.ninja/wp-json/jurassic.ninja/jetpack-beta/plugins/woocommerce-payments/branches - searching for the branch of this PR `rpp/6674-create-payment-service-class`. 

*

-------------------

- [x] NO NEED Covered with tests (or have a good reason not to test in description ☝️)

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6805` or branch name `rpp/6674-create-payment-service-class` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=rpp/6674-create-payment-service-class) :rocket:

---

Build info: 

- Latest commit: e5a6b94c6d239635cb54d4bfe9907a1b731606eb
- Build time: 2023-08-01 14:30:32 UTC

src/Internal/Services/PaymentService.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Class PaymentService
+ *
+ * @package WooPayments
+ */
+
+namespace WooPayments\Internal\Services;
```
htdat: Would love feedback on this.
I suggest using “WooPayments” namespace for all the things for this project as in general, everything WCPay or WooCommerce Payments will be changed to this new name paJDYF-7Sw-p2

IMO, this namespace is just easier to know that classes using this new namespace are structured in the new modern way, rather than use the current prefix `WCPay`.

RadoslavGeorgiev: Pinging @Automattic/transact-architecture for an opinion here. On one hand, this could be a nice way to separate `src` from `includes`, and introduce the new branding. On the other hand, it's yet another inconsistency in our codebase.

@htdat let's follow [the server](https://github.com/Automattic/woocommerce-payments-server/tree/trunk/server/wp-content/rest-api-plugins/endpoints/wcpay) here, and use singular directories (`Service` in this case) rather than plural.

src/Internal/Services/PaymentService.php
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Processing payment service.
+ */
+class PaymentService {
```
RadoslavGeorgiev: I was about to ask what you'd think about `PaymentProcessingService` rather than `PaymentService`, then I saw the doc block. Let's make the service specific to _processing_ payments.

htdat: 👍 Addressed 6192259

src/Internal/Services/PaymentService.php
htdat: > @htdat let's follow [the server](https://github.com/Automattic/woocommerce-payments-server/tree/trunk/server/wp-content/rest-api-plugins/endpoints/wcpay) here, and use singular directories (Service in this case) rather than plural.

✅ Update both namespace and directory to use the singular form in commit 6192259.

RadoslavGeorgiev: > IMO, this namespace is just easier to know that classes using this new namespace are structured in the new modern way, rather than use the current prefix WCPay.

Based on a conversation (p1690470053281149-slack-CGGCLBN58), I was also thinking something else: Since we never explicitly defined anything as internal, it means that we're essentially stuck with the existing classes and the `WCPay` namespace that's already within `includes`. We cannot change the namespace of existing classes.

By introducing `WooPayments` as a namespace here, we're essentially preventing existing classes from being autoloaded and managed through DI (a.k.a. moved to `src`). Are we sure we want to do that? While working on both features, I'm realizing that if we have any hopes of existing code to be moved into `src`, we'd need to stick to `WCPay`. WDYT @htdat ?

htdat: > By introducing `WooPayments` as a namespace here, we're essentially preventing existing classes from being autoloaded and managed through DI (a.k.a. moved to `src`). Are we sure we want to do that?

Eventually, IMO, we'd like to migrate existing classes into `src` when possible and reasonable. 

However, we also should not keep the philosophy of "class names or namespaces will be never changed", as that will just hold us back. Instead, is there any way to prevent any error happens? 

For example, when moving a class from `includes` to `src`, can we provide a depreciated warning message for the class in `includes`, but at the same time, we inherit this old class from the new migrated class in `src`.

Even we're still using namespace `WCPay` for new files in `src`, a lot of classes must change their name and namespaces, such as this one [WC_Payments_Admin](https://github.com/Automattic/woocommerce-payments/blob/20f22cc1495f9842eff24b75392b5f6480d37349/includes/admin/class-wc-payments-admin.php#L19-L19).

RadoslavGeorgiev: > Eventually, IMO, we'd like to migrate existing classes into src when possible and reasonable.

And to make the process as painless as possible.

> However, we also should not keep the philosophy of "class names or namespaces will be never changed", as that will just hold us back.

I wouldn't call it a philosophy, for me it's more of a fact 🤔 

> Instead, is there any way to prevent any error happens?
> For example, when moving a class from includes to src, can we provide a depreciated warning message for the class in includes, but at the same time, we inherit this old class from the new migrated class in src.

We could figure out various ways here, but we'd be introducing a problem that needs resolving. We've got 146 files, which could be moved by `src`. If we keep fall-backs, that would be 146 classes that we'd need to add and connect, instead of just renaming the original ones.

> Even we're still using namespace WCPay for new files in src, a lot of classes must change their name and namespaces, such as this one [WC_Payments_Admin](https://github.com/Automattic/woocommerce-payments/blob/20f22cc1495f9842eff24b75392b5f6480d37349/includes/admin/class-wc-payments-admin.php#L19-L19).

When the time comes for those, we'll definitely need to do something to allow migration. And we've got a lot of them. However, as it is now, when finishing the project, we will be left with 3 separate naming/placing techniques, without a plan (or possibility) to combine them whenever needed.

For those files, I imagine we'd have to move the main class to `src`, rename it, and then use a facade with the name of the old file.

Let's try it the other way: What's the practical/effort-wise upside of having the new namespace?

htdat: > Let's try it the other way: What's the practical/effort-wise upside of having the new namespace?

Mostly to easily differentiate between old way and new way, and of course, if it's not the burden for the future. Otherwise, using a `WCPay` prefix is definitely a way to go.

htdat:
Ready to review again. I am switching back to the original idea of using `WCPay` namespace.

RadoslavGeorgiev (approved):
LGTM :shipit: 

The failing E2E test cannot be related to the changes here.

marcinbot (approved):
Missed most of the discussion here, but happy with the direction. The namespace, folder structure, and naming look good to me :shipit:

PR #6826, opened 2023-07-20
Move email title hook to the card gateway

### LINKED ISSUE #6731: Excessive amount of redundant attempts to fetch payment method during payment processing
### Describe the bug
The scope of this issue is UPE - the split UPE and the UPE with the deferred intent creation. As a precautionary measure, other settings should be tested as part of this issue. 

### To Reproduce
1. Enable deferred intent creation UPE 
2. Go to the checkout and place an order
3. Note that the checkout lasts longer than expected (first screenshot), and during this time, you are not redirected
4. Open the server's `access.log` (second screenshot) and find the significant number of requests to get the payment method

image

image

The same symptoms are to be observed for the split UPE, but there's no visual regression from the user perspective happening, most probably because the redirect occurs before the hook's callback is executed. Still, the payment method is fetched more times than expected. This is also why the issue remained invisible/hidden for some time.

### Expected behavior
As part of setting the email title, only one payment method should be fetched equal to the one used during the checkout.

### Additional context
Per my research, this is happening because of the hook that sets the method title for the email during order processing. The [payment method is fetched multiple times in this place](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L1073), as result of this callback being attached to a hook multiple times [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L120).

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/6731

This PR moves the callback's attachment for `woocommerce_email_before_order_table` hook from the UPE class to the parent legacy gateway class. This was done because the legacy gateway class allows registering hooks for the card gateway only. The problem we're trying to fix appears because previously for the split UPE, 

```php
add_action( 'woocommerce_email_before_order_table', [ $this, 'set_payment_method_title_for_email' ], 10, 3 );
```

was added N times where N = number of available UPE gateways, which led to a visible performance downgrade. The hook should be initialized only once and thus was moved to the parent class.

#### Testing instructions

Perform the below instructions sequentially for: **legacy card, legacy UPE, split UPE, deferred intent creation UPE**.

1. Perform steps 2-5 from the testing instructions [here](https://github.com/Automattic/woocommerce-payments/pull/6351) and confirm they are successful
2. Open `access.log` on your server and confirm that the `GET /wp-json/wpcom/v2/sites/${xyz}/wcpay/payment_methods` request is sent only twice and not more.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6826` or branch name `fix/multiple-payment-methods-fetching` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/multiple-payment-methods-fetching) :rocket:

---

Build info: 

- Latest commit: d896d905fc7207a28a9bfb4860172a9b06f59a86
- Build time: 2023-07-21 07:49:13 UTC

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1390,6 +1391,15 @@ private function get_payment_methods_from_request() {
 	public function set_payment_method_title_for_order( $order, $payment_method_type, $payment_method_details ) {
 	}
```
FangedParakeet: Do we still need this base definition of the function that is never overriden with this signature?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -439,6 +439,7 @@ public function init_hooks() {
 			add_action( 'woocommerce_update_order', [ $this, 'schedule_order_tracking' ], 10, 2 );
 
 			add_filter( 'rest_request_before_callbacks', [ $this, 'remove_all_actions_on_preflight_check' ], 10, 3 );
+			add_action( 'woocommerce_email_before_order_table', [ $this, 'set_payment_method_title_for_email' ], 10, 3 );
```
FangedParakeet: I'd maybe vote towards moving this into the [`init_hooks()`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-upe-checkout.php#L81) function inside `WC_Payments_UPE_Checkout`. This hook is only intended to be fired once and only required for the UPE. Putting it here inside `WC_Payment_Gateway_WCPay` satisfies the first goal, but this hook will also be added when the legacy card is enabled and UPE is disabled after [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L542).

Attaching this hook with UPE disabled probably won't break anything and likely won't be a long-term issue, as we move away from the gateway, but seems redundant nonetheless. 🤷

timur27: Agree with the fact that although there is no negative impact while initialized in the `WC_Payment_Gateway_WCPay`, still `WC_Payments_UPE_Checkout` seems to be a more appropriate place for it. Done in https://github.com/Automattic/woocommerce-payments/pull/6826/commits/11ac6ed5014ee3bf6efffa8d5b20223df7d2a96a

includes/class-wc-payment-gateway-wcpay.php
timur27: Removed in https://github.com/Automattic/woocommerce-payments/pull/6826/commits/11ac6ed5014ee3bf6efffa8d5b20223df7d2a96a

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -118,10 +118,10 @@ public function __construct(
 	 * @return void
 	 */
 	public function init_hooks() {
+		// Initializing a hook within this function increases the probability of multiple calls for each split UPE gateway. Consider adding the hook in the parent hook initialization.
```
FangedParakeet, line 121: Nice!

FangedParakeet (approved):
LGTM! Cheers for the quick turn on both the initial PR and response to feedback. :shipit:

PR #6828, opened 2023-07-20
Determine the proper setting for split UPE to work with WooPay

### LINKED ISSUE #6730: Card payments on the shortcode checkout fail for the split UPE with WooPay
### Describe the bug
The scope of this issue is the shortcode checkout, Blocks checkout isn't affected but should be tested more as a precautionary measure. [The wrong instance of Stripe object](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/api/index.js#L74-L95) is acquired instead of [the proper one](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/api/index.js#L65-L71) for the card gateway, [due to the `forceNetworkSavedCards`](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/api/index.js#L61) field being `undefined`.

### To Reproduce
1. Enable split UPE
2. Enable WooPay
3. Enable any number of additional payment methods other than the card
4. Attempt to purchase a product on the shortcode checkout while paying with card
5. Instead of payment processing, note `Error: No such PaymentMethod: 'pm_${ID}'`

### Screenshots
<p align="center">
image
</p>

### Expected behavior

Payment processing should behave correctly, regardless of the UPE/WooPay/enabled payment methods settings.

### Additional context
* In the split UPE, not only [the `upe-split.js` is enrolled](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/upe-split.js), but also [the legacy card ](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/classic/index.js) scripts. The problem will probably be found and fixed in the latter, not in the `upe-split.js`. 
* Similar problem for the deferred intent creation UPE was fixed as part of [the WooPay integration PR](https://github.com/Automattic/woocommerce-payments/pull/6656). 
* Blocks checkout isn't affected, most probably because of its nature of rendering each component separately, i.e., if the card payment method is chosen, then only the card is being rendered, and other gateways aren't negatively affecting the `forceNetworkSavedCards` value.

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/6730

This PR ensures that for the legacy card gateway in the split UPE, the correct flag is provided to the client side so that the proper Stripe account is used, depending on whether the WooPay is enabled or disabled. Due to the nature of rolling out scripts, we ended up overriding the proper `forceNetworkSavedCards` value with the UPE gateway, which was rolled out as the second and could never use WooPay, leading the flag having the value of `false` always.

#### Testing instructions

1. Enable split UPE
2. Enable WooPay
3. Enable any number of additional payment methods other than the card
4. Attempt to purchase a product on the shortcode checkout while paying with card
5. Confirm that the payment went successfully and you were redirected to the _Order received_ page
6. Purchase a product with any other payment method like `giropay` or similar and confirm the successful purchase

Checkout to `develop`, perform steps 1-4 from above and confirm that instead of the fifth successful step, a user is seeing `Error: No such PaymentMethod: 'pm_${ID}'`.

Blocks checkout isn't affected, but making a single purchase and confirming it works will be appreciated.

Perform steps 1-5 for the legacy card gateway and steps 1-6 for the deferred intent creation UPE gateway. Also, sanity checks the legacy UPE checkout by performing steps 1-6 excluding step 2.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-split-upe-with-woopay-and-multiple-pms
+++ b/changelog/fix-split-upe-with-woopay-and-multiple-pms
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Improve split UPE support of WooPay with multiple payment methods enabled.
--- a/includes/class-wc-payments-checkout.php
+++ b/includes/class-wc-payments-checkout.php
@@ -176,7 +176,7 @@ class WC_Payments_Checkout {
 			'genericErrorMessage'            => __( 'There was a problem processing the payment. Please check your email inbox and refresh the page to try again.', 'woocommerce-payments' ),
 			'fraudServices'                  => $this->account->get_fraud_services_config(),
 			'features'                       => $this->gateway->supports,
-			'forceNetworkSavedCards'         => WC_Payments::is_network_saved_cards_enabled() || $this->gateway->should_use_stripe_platform_on_checkout_page(),
+			'forceNetworkSavedCards'         => WC_Payments::is_network_saved_cards_enabled() || WC_Payments::get_registered_card_gateway()->should_use_stripe_platform_on_checkout_page(),
 			'locale'                         => WC_Payments_Utils::convert_to_stripe_locale( get_locale() ),
 			'isPreview'                      => is_preview(),
 			'isUPEEnabled'                   => WC_Payments_Features::is_upe_enabled(),
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -770,6 +770,15 @@ class WC_Payments {
 	}
 
 	/**
+	 * Sets registered card gateway instance.
+	 *
+	 * @param WC_Payment_Gateway_WCPay|UPE_Payment_Gateway|UPE_Split_Payment_Gateway $gateway Gateway instance.
+	 */
+	public static function set_registered_card_gateway( $gateway ) {
+		self::$registered_card_gateway = $gateway;
+	}
+
+	/**
 	 * Called on Payments setting page.
 	 *
 	 * Remove all WCPay gateways except CC one.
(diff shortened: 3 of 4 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6828` or branch name `fix/split-upe-with-woopay-and-multiple-pms` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/split-upe-with-woopay-and-multiple-pms) :rocket:

---

Build info: 

- Latest commit: 16b50bbb57900c94368a14a84ddbce6973028238
- Build time: 2023-07-20 10:52:38 UTC

FangedParakeet (approved):
Can confirm the original issue and that this is indeed resolving it. ✅ 

Was the root cause of this because the JS config was being overwritten when rendering the payment fields for UPE payment methods? 🤔 If so, maybe we should think about refactoring this in the future, so that like the hooks these global variables are only appended to the page once, to ensure there aren't any more unintended side-effects in the future.

That being said, for now, LGTM. 👍

timur27:
> Was the root cause of this because the JS config was being overwritten when rendering the payment fields for UPE payment methods?

Yes. 

This field is now supposed to base only on the card gateway since the flag for non-card gateways was separately added in https://github.com/Automattic/woocommerce-payments/pull/6656. I agree the refactor is needed in this area at some point.

PR #6831, opened 2023-07-20
Combine Session Initialization with User Authentication on WooPay

rafaelzaleski:
[1968-gh-Automattic/woopay
](1968-gh-Automattic/woopay)

- Session Initialization and User Authentication are combined into one request to WooPay to minimize unnecessary round-trips to/from the merchant server.
- Refactor code to move WooPay session functions to the class `WooPay_Session`

#### Testing instructions

2000-gh-Automattic/woopay

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6831` or branch name `perf/woopay-combine-verify-init-requests` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=perf/woopay-combine-verify-init-requests) :rocket:

---

Build info: 

- Latest commit: f8e1a6f9705cb54dd8d23c9aa83e8522f613749e
- Build time: 2023-08-14 22:48:12 UTC

includes/woopay/class-woopay-session.php
```diff
@@ -46,6 +52,44 @@ public static function init() {
...
+	/**
+	 * Initializes the session data for WooPay to be combined with the /verify request.
+	 * This is only used for the express button outside of the checkout page.
+	 *
+	 * @return void
+	 */
+	public static function init_woopay_session_data_button() {
+
+		// the checkout is excluded and handled in init_woopay_session_data_checkout.
+		if ( ! WC_Payments_Features::is_woopay_enabled() || is_checkout() ) {
```
lovo-h: Should we only load the session data for the button when the `woopay_button` is loaded? If so, we may want to reuse or replicate the [should_show_woopay_button()](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-woopay-button-handler.php#L478C32-L478C32) logic.

lovo-h (dismissed):
The code changes are looking good. Thanks for adding this optimization. There was just one concern I wanted to address before moving forward (see failing test).

**Testing Instructions**
* ✅ Unit tests passed.
* ✅ E2E tests passed.
* ❌ Tests from 2000-gh-Automattic/woopay.

includes/woopay/class-woopay-session.php
```diff
@@ -216,6 +260,70 @@ private static function get_payload_from_cart_token() {
...
+		add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+		$preloaded_checkout_data = rest_preload_api_request( [], '/wc/store/v1/checkout' );
+		remove_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+		$checkout_data = isset( $preloaded_checkout_data['/wc/store/v1/checkout'] ) ? $preloaded_checkout_data['/wc/store/v1/checkout']['body'] : '';
+
+		$request = [
+			'wcpay_version'      => WCPAY_VERSION_NUMBER,
+			'user_id'            => $user->ID,
+			'customer_id'        => $customer_id,
+			'session_nonce'      => wp_create_nonce( 'wc_store_api' ),
```
gpressutto5: Heads up! This was updated in #6868!

includes/woopay/class-woopay-session.php
```diff
@@ -216,6 +260,70 @@ private static function get_payload_from_cart_token() {
...
+
+		include_once WCPAY_ABSPATH . 'includes/compat/blocks/class-blocks-data-extractor.php';
+		$blocks_data_extractor = new Blocks_Data_Extractor();
+
+		// This uses the same logic as the Checkout block in hydrate_from_api to get the cart and checkout data.
+		$cart_data = rest_preload_api_request( [], '/wc/store/v1/cart' )['/wc/store/v1/cart']['body'];
+		add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+		$preloaded_checkout_data = rest_preload_api_request( [], '/wc/store/v1/checkout' );
+		remove_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
+		$checkout_data = isset( $preloaded_checkout_data['/wc/store/v1/checkout'] ) ? $preloaded_checkout_data['/wc/store/v1/checkout']['body'] : '';
```
gpressutto5, line 353: These 2 requests are already being preloaded and exposed to the frontend by the Checkout Block. We needed this because this used to be a separate endpoint, but now we are doing this at html render time, which is the same time the checkout Block does it. Do you know if this method is using the already existing preloaded response instead of ignoring it and getting a new response? If not, can we get what is already in the frontend and use it instead of reloading?

includes/woopay/class-woopay-session.php
```diff
@@ -46,6 +52,44 @@ public static function init() {
...
+	 */
+	public static function init_woopay_session_data_checkout() {
+
+		if ( ! WC_Payments_Features::is_woopay_enabled() || ! is_checkout() ) {
+			return;
+		}
+
+		$session_data = self::get_init_session_request();
+
+		echo '<script type="text/javascript">sessionDataWooPay=' . wp_json_encode( $session_data ) . '</script>';
```
gpressutto5: Why do we need two different methods to add the data to the frontend?

tests/unit/test-class-wc-payments.php
```diff
@@ -117,51 +119,6 @@ public function test_rest_endpoints_validate_nonce() {
...
-			}
-
-			return $result;
-		};
-
-		$pre_http_request_cb = function ( $preempt, $parsed_args, $url ) use ( $customer_id ) {
-			$body = json_decode( $parsed_args['body'] );
-			$this->assertEquals( $customer_id, $body->customer_id );
-			return [ 'body' => wp_json_encode( [] ) ];
-		};
```
rafaelzaleski, line 138: This assertion is never called because the request on `Automattic/Jetpack/Connection/Client::remote_request` will always fail (it cannot be signed inside the test). This is the reason for the removal.

includes/woopay/class-woopay-session.php
rafaelzaleski: The logic was changed on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/1afcf02fdb1f67f21b94fbd1289b2d49ac6e1989 and the session data is not printed in the page anymore.

includes/woopay/class-woopay-session.php
rafaelzaleski: The logic was changed on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/1afcf02fdb1f67f21b94fbd1289b2d49ac6e1989 and the session data is not printed in the page anymore.

includes/woopay/class-woopay-session.php
rafaelzaleski: The logic was changed on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/1afcf02fdb1f67f21b94fbd1289b2d49ac6e1989 and the session data is not printed in the page anymore.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -510,6 +527,14 @@ export const handleWooPayEmailInput = async (
 				hasCheckedLoginSession = true;
 				closeLoginSessionIframe();
 				break;
+			case 'redirect_to_platform_checkout_skip_session_init':
```
ricardo: We should aim to use `woopay` instead of `platform_checkout` from now on as per https://github.com/Automattic/woocommerce-payments/pull/6061.

Same for `express-checkout-iframe.js`.

client/checkout/woopay/email-input-iframe.js
```diff
@@ -510,6 +527,14 @@ export const handleWooPayEmailInput = async (
...
+				wcpayTracks.recordUserEvent(
+					wcpayTracks.events.WOOPAY_OTP_COMPLETE
```
ricardo: I see in other instances of `recordUserEvent`, we're setting `isLegacy = true`. Should we do the same here as well?

```JavaScript
wcpayTracks.recordUserEvent(
	wcpayTracks.events.WOOPAY_OTP_COMPLETE,
	[],
	true
);
```

Same for `express-checkout-iframe.js`.

includes/woopay/class-woopay-session.php
```diff
@@ -277,6 +285,201 @@ private static function get_payload_from_cart_token() {
...
+
+		$store_blog_token = Jetpack_Options::get_option( 'blog_token' );
+
+		$message = wp_json_encode( $session );
+
+		// Generate an initialization vector (IV) for encryption.
+		$iv = openssl_random_pseudo_bytes( openssl_cipher_iv_length( 'aes-256-cbc' ) );
+
+		// Encrypt the JSON session.
+		$session_encrypted = openssl_encrypt( $message, 'aes-256-cbc', WC_Payments::mode()->is_dev() ? 'dev_mode' : $store_blog_token, OPENSSL_RAW_DATA, $iv );
```
ricardo: We might want to check if the `openssl` module is loaded to prevent errors, there's an utility function for this: https://github.com/Automattic/woocommerce-payments/blob/a728962115459efd739d07bf99a1d11d939ee8f5/includes/class-wc-payments-features.php#L152-L154

There's also an [existing function](https://github.com/Automattic/woocommerce-payments/blob/a728962115459efd739d07bf99a1d11d939ee8f5/includes/class-wc-payments-utils.php#L869) used for this type of encryption, but the problem is that it falls back to an unencrypted string.

includes/woopay/class-woopay-session.php
```diff
@@ -277,6 +285,201 @@ private static function get_payload_from_cart_token() {
...
+			Logger::error( 'HTTP_REQUEST_ERROR ' . var_export( $response, true ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_var_export
+			// phpcs:ignore
+			/**
+			 * @psalm-suppress UndefinedDocblockClass
+			 */
+			$message = sprintf(
+				// translators: %1: original error message.
+				__( 'Http request failed. Reason: %1$s', 'woocommerce-payments' ),
+				$response->get_error_message()
+			);
```
ricardo: This variable is never used.

includes/woopay/class-woopay-session.php
```diff
@@ -277,6 +285,201 @@ private static function get_payload_from_cart_token() {
...
+		$response = [
+			'blog_id' => Jetpack_Options::get_option( 'id' ),
+			'data'    => array_map( 'base64_encode', $data ),
+		];
+
+		return $response;
+	}
+
+	/**
+	 * Summary of get_init_session_request
```
ricardo: This looks like a placeholder text, should we add some description like "Returns the initial session request data."?

ricardo (changes requested):
Changes look good and test well. 💯 

I've left a couple minor comments.

client/checkout/woopay/email-input-iframe.js
rafaelzaleski: Done in https://github.com/Automattic/woocommerce-payments/pull/6831/commits/71f5f636ab1969cbfd9abb90dfbba950455b1407

includes/woopay/class-woopay-session.php
rafaelzaleski: Removed in https://github.com/Automattic/woocommerce-payments/pull/6831/commits/2e81fb05c6c201e6988267e9f4a9e6e14f18af2b.

includes/woopay/class-woopay-session.php
rafaelzaleski: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6831/commits/792c6e335d7f4c4efc98f65e05bac4c3496eb42b

includes/woopay/class-woopay-session.php
rafaelzaleski: Good call! Added on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/ea5cafbd173ad3a9162d0e08ec1f6c45c69dd1cd.

client/checkout/woopay/email-input-iframe.js
rafaelzaleski: Good find! They added these parameters after I open this PR.
Added on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/88cbec43436bb706bb5095e1317a8f7a36e6f7b3

includes/woopay/class-woopay-session.php
ricardo: Just to clarify, if the encryption doesn't work in this case, will it fallback to a separate request to `wc_ajax_wcpay_init_woopay`?

Edit: Ok, I just tested it and it does. Nice.

ricardo (commented):
Recent code changes LGTM, however, see my comment in the WooPay repo regarding the sandbox testing.

includes/woopay/class-woopay-session.php
```diff
@@ -277,6 +285,196 @@ private static function get_payload_from_cart_token() {
...
+	 * @return array The encrypted session request or an empty array if the server is not eligible for encryption.
+	 */
+	public static function get_frontend_init_session_request() {
+		if ( ! WC_Payments_Features::is_client_secret_encryption_eligible() ) {
+			return [];
+		}
+
+		$session = self::get_init_session_request();
+
+		$store_blog_token = Jetpack_Options::get_option( 'blog_token' );
```
ricardo: Maybe we can change this line to:

```php
$store_blog_token = WC_Payments::mode()->is_dev() ? 'dev_mode' : Jetpack_Options::get_option( 'blog_token' );
```

To avoid repeating the same condition on line `308` and `311`.

ricardo (approved):
LGTM and tests well. 👍

rafaelzaleski:
@ricardo, I changed the condition to select the passphrase on https://github.com/Automattic/woocommerce-payments/pull/6831/commits/4a441a38b98a139850ea133218f336ec764511cb. Would you like to review?

ricardo (approved):
LGTM.

PR #6421, opened 2023-06-01
Duplicate payment prevention service

RadoslavGeorgiev:
Fixes #6285.

#### Changes proposed in this Pull Request

Moves the duplicate payment prevention functionality into a new `Duplicate_Payment_Prevention_Service`.

This helps clean up the gateway class, and instead of relying on parent functionality, UPE classes will now use the service.

- [x] Create the new class and add it as a dependency of gateways.
- [x] Move functionality into the new class.
- [x] Introduce an `init` method, which is called by `WC_Payments`, preventing issues with circular dependencies (the gateway relies on the service, and the service relies on the gateway).
- [x] Fix existing tests by removing mocks, and using the real service.
- [x] Add new unit tests (focusing on the service), and consider whether the existing tests should be more lightweight afterwards.

#### Testing instructions

- Same as the previous PRs (linked in the issue).
- Make sure tests work.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)

**Post merge**

- [x] No additional testing instructions
- [x] Critical flows already include duplicate payments.
- [x] No screenshot needed for What's New, this is just refactoring.

PR #6862, opened 2023-07-26
RPP: Dependency Container

### LINKED ISSUE #6678: Payment Process: Introduce a dependency container
### Description

Introduce a dependency container for the new `src` directory, created in https://github.com/Automattic/woocommerce-payments/issues/6675

Being a WooCommerce extension, we should follow the [Core example](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/src#the-container). The WooCommerce container is not designed to be used outside of WooCommerce, so we will need to implement something similar.

I already spiked the container in this PR: https://github.com/Automattic/woocommerce-payments/pull/6593. My findings:

- The [League container](https://container.thephpleague.com/) seems like a suitable option. We don't need to stray further from core.
- The container supports [Auto-Wiring](https://container.thephpleague.com/4.x/auto-wiring/), meaning that we would not necessarily need to manually define all available classes, and their dependencies. I asked @Konamiman why WooCommerce is not using the feature, and the answer was simple: Performance. Writing the definition of a class and its dependencies is a one-time job, it's easy, and quick. Let's _not_ use this feature.
- [Delegate containers](https://container.thephpleague.com/4.x/delegate-containers/) allow a container to delegate to another one in case the primary container does not contain the definition of a class. If we want to utilize this, we cannot use auto-wiring. If we use it, we can delegate down to the WooCommerce container, which would be nice.
The Core container doesn't implement [the PSR-11 interface](https://github.com/php-fig/container/blob/master/src/ContainerInterface.php), required for a delegate, but we can still create our own class, which would delegate to the core container.
```php
class WCPay_WooCommerce_Container_Delegate implements Psr\Container\ContainerInterface {
    public function get( string $id ) {
        return wc_get_container()->get( $id );
    }

    public function has( string $id ) {
        return wc_get_container()->has( $id );
    }
}

$wcpay_container->delegate( new WCPay_WooCommerce_Container_Delegate() );
```
- We might implement a [simple base service provider class](https://github.com/Automattic/woocommerce-payments/pull/6593/files#diff-402fb89a8da8140fea6d7659b90bc9fc88fa4c925d98ef6f4ecad2d9f6eaef55), which would allow defining a `$provides` property to have our providers look similar to Core's.
- It will probably make sense to wrap the League container in a class of our own, limiting what can be done outside of it. [Example](https://github.com/Automattic/woocommerce-payments/pull/6593/files#diff-55c5d44a849e66cfaefd2086077dfb27d261c84a9736a17c9ff35d285739a6ee)
- Use a tool like [Mozart](https://github.com/coenjacobs/mozart) to pull the container into our namespace, avoiding conflicts with other extensions. Otherwise, upgrading WCPay might end up causing fatal errors on merchant sites.

### Acceptance criteria

- The container can be accessed externally through `wcpay_get_container()`.
- We are delegating at least `Automattic\WooCommerce` classes to the WooCommerce container.
- We are utilizing [Service Providers](https://container.thephpleague.com/4.x/service-providers/) for class definitions.
- The container dependency is prefixed by `WCPay\Vendor` through Mozart, and remains so after Composer operations.
- The container is wired up in a way that works in unit tests, allowing classes to be replaced.

### Follow-up tasks

See https://github.com/Automattic/woocommerce-payments/issues/6707

### Testing instructions

- Make sure that whatever custom container code we have, is tested.

### Additional context

Project: paJDYF-9hL-p2

RadoslavGeorgiev:
Fixes #6678.

Introduce a dependency container for the new `src` directory

To achieve that, within this PR:

- The [League container](https://container.thephpleague.com/) is used, as in Core.
- [Auto-Wiring](https://container.thephpleague.com/4.x/auto-wiring/) is intentionally omitted.
- A [Delegate container](https://container.thephpleague.com/4.x/delegate-containers/) allows our container to load WooCommerce core classes.
- A simple base service provider class is implemented, as well as the first service provider for the `PaymentProcessingService`.
- The League Container is extended by `WooPayments\Internal\DependencyManagement\ExtendedContainer` (which adds all necessary additional functionality), while the actual `WooPayments\Container` contains it, and only allows access to selected methods (`get` and `has`).
- [Mozart](https://github.com/coenjacobs/mozart) is used to wrap the League container in our own namespace, through a separate Composer instance.
- The PSR-11 container interface is still included in the root, as it should be useful for compatibility with other plugins.
- The container can be accessed through `wcpay_get_container()`.
- The container is wired up in a way that works in unit tests, allowing classes to be replaced.

#### Docs

The additions in this PR are described in two docs:

1. The readme within `src` ([ref](https://github.com/Automattic/woocommerce-payments/tree/rpp/6678-dependency-container/src)) contains general developed docs.
2. The one in `lib` ([ref](https://github.com/Automattic/woocommerce-payments/tree/rpp/6678-dependency-container/lib)) describes how to manage the container package.
 
#### Testing instructions

Everything is covered in automated tests. Please inspect `tests/unit/src/ContainerTest.php`.

-------------------

**Post merge**

No applicable post-merge actions.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6862` or branch name `rpp/6678-dependency-container` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=rpp/6678-dependency-container) :rocket:

---

Build info: 

- Latest commit: b8be247f88b35235f2eca2f95b94bc74d6db67de
- Build time: 2023-08-04 14:13:49 UTC

woocommerce-payments.php
```diff
@@ -12,7 +12,7 @@
  * WC tested up to: 7.8.0
  * Requires at least: 6.0
  * Requires PHP: 7.3
- * Version: 6.2.0
+ * Version: 6.4.0
```
RadoslavGeorgiev: This was commited by mistake, and will be fixed soon.

woocommerce-payments.php
```diff
@@ -150,6 +149,7 @@ function () {
  */
 function wcpay_init() {
 	require_once WCPAY_ABSPATH . '/includes/class-wc-payments.php';
+	require_once WCPAY_ABSPATH . '/src/wcpay-get-container.php';
```
htdat: May config `autoload` in composer?  https://getcomposer.org/doc/04-schema.md#files

Something like this https://gist.github.com/htdat/191438fed64b6d7fefac5e3ff274960d

With this, we do not need to manually load this file here, pslam-loader.php, and tests bootstrap.php

lib/composer.json
```diff
@@ -0,0 +1,33 @@
+{
+		"name": "wcpay/wcpay-lib",
+		"description": "A package for hiding re-nameed dependencies.",
```
marcinbot: Typo: `re-nameed`

Also, it might be good to fix the tabs vs spaces and some other such inconsistency in this and the root `composer.json` file.

lib/README.md
```diff
@@ -0,0 +1,25 @@
+# WooPayments Lib Directory
+
+Similarly to how Composer loads packages into the `vendor` directory, this directory will contain WooPayments-specific libraries. There are a couple of differences here:
+
+- All packages have their namespaces prefixed by `WooPayments\Vendor\` to prevent conflicts with other plugins. This is automated through the `coenjacobs/mozart` package.
+- Even though `vendor` packages are not commited to git, the packages within `lib/packages` will be commited.
```
marcinbot, line 6: I wonder if this could be automated somehow during the root `composer install` (maybe using the `post-install/update-cmd` hook). I.e. run the install here every time `composer install` is run in the root and make sure the result is included on release. This could save us a lot of confusion when it comes to dev process around this.

At the same time, I know that Core does the same, and it's an age-old question of how to control dependencies, so I'm not overly fussed about it.

src/Container.php
```diff
@@ -0,0 +1,82 @@
...
+	 * Dependencies should not be provided during runtime, but will allow
+	 * mocking during tests. This is only needed for the container.
+	 *
+	 * @param WooContainer $woo_container The delegate container for WooCommerce (Optional).
+	 */
+	public function __construct( WooContainer $woo_container = null ) {
+		$this->container = new ExtendedContainer();
+
+		// Allow the container to be used as a dependency.
+		$this->container->addShared( static::class, $this );
```
marcinbot, line 47: Is this necessary if we already have `wcpay_get_container`? I would rather the container be used only once at the flow entry point and resolve all possible dependencies, in keeping with inversion of control (the individual classes should not be concerned about how they or others are created, only business logic).

src/Internal/DependencyManagement/ExtendedContainer.php
```diff
@@ -0,0 +1,102 @@
...
+
+use WCPay\Vendor\League\Container\Container;
+use WCPay\Vendor\League\Container\Definition\Definition;
+
+/**
+ * Extends the League container to allow WCPay customizations.
+ *
+ * During tests, `wcpay_get_test_container()` will return an instance of this class.
+ */
+class ExtendedContainer extends Container {
```
marcinbot, line 18: Maybe this could belong in the tests directory, so we don't expose it to the devs or ship it with the plugin?

src/README.md
```diff
@@ -0,0 +1,180 @@
...
+	/**
+	 * Registers all provided classes.
+	 */
+	public function register(): void {
+		$container = $this->getContainer();
+
+		$container->addShared( SimpleService::class );
+
+		$container->addShared( SimpleServiceWithDependencies::class )
+			->addArgument( ServiceWithDependencies::class);
```
marcinbot: Perhaps `SimpleService` would be better in this line given the rest of the function?

src/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+		$container->add( Request::class )
+			->addArgument( WC_Payments_API_Client::class );
```
marcinbot, line 132: It might be confusing where `WC_Payments_API_Client` is added to the container. I think this highlights an interesting case where the new `src` relies on the old `includes` code. Maybe a comment here would help?

src/wcpay-get-container.php
```diff
@@ -0,0 +1,19 @@
...
+$GLOBALS['wcpay_container'] = new Container();
+
+/**
+ * Returns the WCPay DI container.
+ *
+ * @return Container
+ */
+function wcpay_get_container() {
+	return $GLOBALS['wcpay_container'];
+}
```
marcinbot: Maybe we could initialize the container within the function body, like:
```
function wcpay_get_container() {
        if ( ! isset( $GLOBALS['wcpay_container'] ) ) {
            $GLOBALS['wcpay_container'] = new Container();
        }
	return $GLOBALS['wcpay_container'];
}
```

Just trying to avoid side effects when loading this file. I know there are other files that do it in this and other plugins, but IMO we should start with a bit better standards.

marcinbot (commented):
Looks good. Some minor comments and questions below

lib/composer.json
RadoslavGeorgiev: > Typo: re-nameed

Thanks, that's fixed now.

> Also, it might be good to fix the tabs vs spaces and some other such inconsistency in this and the root composer.json file.

I was actually battling with this in earlier commits, but it got out of focus. Now I switched to 4 spaces for indentation in both files (https://github.com/Automattic/woocommerce-payments/pull/6862/commits/5bf37999d183523864fb9878350b6e4002c5ecde), as Composer uses spaces. A bit more changes in the PR, but at least we'll finally have those files well indented.

lib/README.md
RadoslavGeorgiev: It's a weird experience, trying to wrap a public package within your own namespace.

We need Mozart to achieve this on `composer install` (or `update`), and some builds (ex. live) do not include dev dependencies.

To allow the `lib` package to be installed and updated automatically, we'll need to alter our workflows to first install Composer with all dependencies, then again, but skipping dev dependencies. I was considering the same, but I think it's a complicated and tricky effort, without a significant outcome.

Woo Core has commands for in their root `package.json`, but even those would be just some additional files, which only achieve the same goal. That's why my opinion is that even that's not worth it.

Thoughts?

src/Container.php
RadoslavGeorgiev: > Is this necessary if we already have `wcpay_get_container`?

There's a bit of a mental difference: The container should be used as a dependency within `src`, while `wcpay_get_container()` should be used within `includes`.

>  I would rather the container be used only once at the flow entry point and resolve all possible dependencies, in keeping with inversion of control (the individual classes should not be concerned about how they or others are created, only business logic).

TBH, I was mostly following Woo's example here, and thought it made sense.

In general, I'm 100% 👍 for IoC. However, I don't consider one class being aware of ways to instantiate another one as opposing the principle. Even though DI and IoC go hand in hand, there's a reason why dependencies are not necessarily always shared.

`WCPay\Container` class exists for a reason, and that's to limit the methods, accessible later. Did you notice that only `has` and `get` are available, but others like `add` and `extend` are not? This prevents "mid-level" classes from altering the behavior of the application, so the control remains on top.

There are plenty of cases where it makes sense for a non-shared class to have dependencies, and for those to be provided directly by the container. I've got payment states in my mind right now, wonder who suggested that pattern? 😝 

Anticipating the next question: Technically, we could have payment states as services, and pass the `$payment` object around to all methods. Yet, it also makes sense to instantiate states specifically for a given payment, and using that payment for context, together with all other needed dependencies.

In any case, consider the following: We need the container to be accessible publicly (from `includes`), and as long as `wcpay_get_container()` is accessible publicly, developers will use it even in `src` when needed.

I would rather have the container available as a dependency because that's still better to mock, similarly to a factory:

```php
class Payment {
	private $container;
	private $state;

	public function __construct( Container $container ) {
		$this->container = $container;
	}

	public function process() {
		$state = $this->container->get( SomeState::class );
		$state->set_context( $this );
		$this->state = $state; // I don't 100% like the recursion here by the way.
	}
}

class Payment_Test extends WCPAY_UnitTestCase {
	public function test_payment_transitions_to_state() {
		$mock_container = $this->createMock( Container::class );
		$mock_state     = $this->createMock( SomeState::class );
		
		$mock_container->expects( $this->once() )
			->method( 'get' )
			->with( SomeState::class )
			->willReturn( $mock_state );
		
		$sut = new Payment( $mock_container );

		$mock_state->expects( $this->once() )
			->method( 'set_context' )
			->with( $sut );

		$sut->process();
	}
}
```

---

Another note on a similar topic: Once we migrate server requests to `src`, the case would be very similar for them. Instead of using `WC_Payments::create_request()`, a class would be able to use the container as a dependency, and instantiate requests through it. Those would be mockable in the same way during tests. Of course, requests would need not to be shared.

```php
class SomeService {
	private $container;

	public function __construct( Container $container ) {
		$this->container = $container;
	}

	public function some_method() {
		$request = $this->container->get( SomeRequest::class );
		$request->set_xyz( 123 );
		$request->send();
	}
}
```

I know this is a different can of worms, but it's already open anyway.

src/Internal/DependencyManagement/ExtendedContainer.php
RadoslavGeorgiev: It is not really exposed to devs because the extended container is a private property of `WCPay\Container`.  `WCPay\Container` does not, and it should never expose those methods of the internal container.

I thought about having a separate container for tests, to avoid shipping those few methods with the plugin, but achieving that would involve hacks. Seems like the easiest and most straight-forward method is to use reflection during tests, and access the private property.

By the way, the main `WCPay\Container` class is still the one that's supplied to classes, and also the one responsible for loading service providers. I don't want to include any code, which allows access to/replacement of the internal `WCPay\Internal\DependencyManagement\ExtendedContainer`.

src/README.md
RadoslavGeorgiev: Oh, my mistake, thanks for noticing this! Fixed in https://github.com/Automattic/woocommerce-payments/pull/6862/commits/4183481bba1c4739a0ed76e8d68a87114a6ce66f

Note: It will make sense to update this example once `src` has some more real classes.

src/README.md
RadoslavGeorgiev: The readme will be updated in https://github.com/Automattic/woocommerce-payments/pull/6876, which introduces a `LegacyContainer` that has hard-coded ways to access legacy classes. If the approach there is approved, this example will be valid.

Is there any chance that you could take a quick glance at https://github.com/Automattic/woocommerce-payments/pull/6876/files#diff-97c044d71e0d3b85870e78635d1f2de33e6ba482799e5d61ac21485432ab6f93, and comment on the approach before the PR is polished? Note that the idea is similar to Woo's `LegacyContainer` ([ref](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Proxies/LegacyProxy.php#L69)), but instead of accessing legacy instances through `LegacyContainer`, they would be available as direct dependencies. Once moved to `src`, the classes that depend on them will not require modifications.

src/wcpay-get-container.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/6862/commits/daab07084a284c3ed469e5962efabbedda74a5b8

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -718,7 +718,7 @@ public function payment_fields() {
 	public function process_payment( $order_id ) {
 
 		if ( defined( 'WCPAY_NEW_PROCESS' ) && true === WCPAY_NEW_PROCESS ) {
-			$new_process = new PaymentProcessingService();
+			$new_process = wcpay_get_container()->get( PaymentProcessingService::class );
```
RadoslavGeorgiev, line 721: Note: The service should be loaded within the constructor of the class, and used later. However, that would require fixing a ton of tests ([ref](https://github.com/Automattic/woocommerce-payments/pull/6421/files)), and this PR already has too many changes.

lib/README.md
marcinbot: > We need Mozart to achieve this on composer install (or update), and some builds (ex. live) do not include dev dependencies.

Yeah, I was thinking that even if the build is live we'd pull dev dependencies for this folder only. And then my understanding is that Mozart would take care of replacing the namespace and copying the "dev" container dependency to this folder, which would be included in the build.

I guess the problem could be excluding the "dev" dependencies from the `vendor` folder after the pull.

No easy solution - if Core does things the same way feel free to keep it as it is now :) I think it will become less jarring after this initial commit anyway.

src/Container.php
marcinbot: Yeah, the States pattern is a bit of an exception here, and it will be interesting to see how it works. I do agree including the container as a dependency might be more elegant than calling it in a static way.

> Once we migrate server requests to src, the case would be very similar for them. 

Unrelated, but some comments were left on Zvonimir's recent post about the Requests. Maybe not as a priority, but we might want to revisit them briefly, especially Request instantiation bit.

src/Internal/DependencyManagement/ExtendedContainer.php
marcinbot: Ok, that makes sense. The labyrinth of namespaces and containers is a bit confusing, but once it's in place then it won't need to be changed much. Let's go with this approach.

src/README.md
marcinbot: > Is there any chance that you could take a quick glance at https://github.com/Automattic/woocommerce-payments/pull/6876/files#diff-97c044d71e0d3b85870e78635d1f2de33e6ba482799e5d61ac21485432ab6f93, and comment on the approach before the PR is polished? 

Yep, the approach looks good and consistent with what's here, nothing stands out so far :) I can leave a proper review once it is ready.

marcinbot (approved):
Looks good to me now, thanks for answering my questions.

lib/README.md
RadoslavGeorgiev: > I guess the problem could be excluding the "dev" dependencies from the vendor folder after the pull.

Exactly. It's mainly Mozart itself that will need to be excluded, or deleted. Anyway, closing this thread, I think we have a conclusion.

src/Container.php
RadoslavGeorgiev: > Unrelated, but some comments were left on Zvonimir's recent post about the Requests. Maybe not as a priority, but we might want to revisit them briefly, especially Request instantiation bit.

I have them in mind for some time this week. :)

src/Container.php
```diff
@@ -0,0 +1,82 @@
...
+
+		// Add shared services.
+		$this->load_providers();
+
+		// Allow delegating unresolved queries to the WooCommerce container.
+		$this->container->delegate( $woo_container ?? new WooContainer() );
```
htdat, line 53: Is there any benefit to having a dependency like this `WooContainer $woo_container = null` and then still check if it is `null` and then call itself? 

So we can remove the dependency and make it simple like this one: 
```suggestion
		$this->container->delegate( new WooContainer() );
```

src/Container.php
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	/**
+	 * Retrieves an instance of a given class.
+	 *
+	 * @param string $id The ID of the class to retrieve.
+	 * @return mixed
+	 */
```
htdat: May use this dynamic return type? That can assist IDEs better. I got that idea from here https://phpstan.org/developing-extensions/dynamic-return-type-extensions and it seems to work well in PhpStorm when I check the change of `class-wc-payment-gateway-wcpay.php` in this PR. 

```suggestion
	 * @template T
	 * @param class-string<T> $id The ID of the class to retrieve.
	 * @return T
	 */
```

htdat (approved):
Pre-approve PR. I also have the similar concern with `ExtendedContainer` as I also think it should be wrapped in the `tests` folder. But the answer from Rado makes sense. 

Instead of that, there are some minor inline comments. 

Pre-approve the PR.

src/Container.php
htdat: It turns out this template annotation was also mentioned in Psalm https://psalm.dev/docs/annotating_code/templated_annotations/#param-class-stringt

And League Container itself has a commit to improve this one https://github.com/thephpleague/container/commit/388e992884ab75632d93dbaf4d29ec024d64b964

src/Container.php
RadoslavGeorgiev: @htdat I did this to allow mocking the container in tests: https://github.com/Automattic/woocommerce-payments/blob/rpp/6678-dependency-container/tests/unit/src/ContainerTest.php#L56-L57

> Is there any benefit to having a dependency like this WooContainer $woo_container = null and then still check if it is null and then call itself?

The line in question simply creates the instance if `$woo_container` isn't already set.

src/Container.php
RadoslavGeorgiev: Amazing suggestion, @htdat! Implemented in https://github.com/Automattic/woocommerce-payments/pull/6862/commits/0766124aecb621f6d781d49e9e3ab76415010dd1

By the way, I never knew this existed in PHP.

woocommerce-payments.php
RadoslavGeorgiev: Interesting suggestion. I added it in https://github.com/Automattic/woocommerce-payments/pull/6862/commits/e7e7842549f6b8bb6478deb312f93e17cb52c34b, let's see how it works.

What concerns me a bit is that this way we have less control over the order of what's loaded. On the other hand, nothing should happen there, besides registering a function, so we should be good to go.

It will be interesting what happens once more stuff gets added to `src`, and some of it needs to be loaded directly, instead of through a method in `includes`. In the worst case we'll revert this change, but I have a suspicion that it might end up being helpful instead, forcing us into a more elegant solution.

src/Container.php
RadoslavGeorgiev: I'm marking this as resolved, and merging the PR, since it's pre-approved.

htdat: Ah, gotcha! If that is for tests, so it makes sense.

PR #6876, opened 2023-07-28
RPP: Legacy Container

### LINKED ISSUE #6707: Payment Process: Dependency container additions
### Description

I've added 2 extra SPs to this issue, which should be used to:

- Copy/imitate the `LegacyProxy` and `ActionsProxy` from Core, making them work in unit tests.
- Figure out a way to bring `includes` classes into the container. Hard-coding methods [like this](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Proxies/LegacyProxy.php#L69-L71) seems suitable.

RadoslavGeorgiev:
Partially fixes #6707, specifically this part:

> Figure out a way to bring `includes` classes into the container. Hard-coding methods [like this](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Proxies/LegacyProxy.php#L69-L71) seems suitable.

Building on top of https://github.com/Automattic/woocommerce-payments/pull/6862, this PR tries to allow classes from the legacy `includes` directory to be loaded as dependencies by `src` classes. This is achieved through `WCPay\Internal\DependencyManagement\DelegateContainer\LegacyContainer`.

`LegacyContainer` implements the PSR-11 interface, and is then used as a delegate container within `WCPay\Container`. It works as a hard-coded bridge between `src` and `includes`.

Essentially, it:

1. Transforms the name of the needed class to a method. For example `WCPay\Mode` becomes `get_wcpay_core_mode_instance`.
2. Checks if the class has such a method, and uses this check within the public `has()` method, which indicates whether the container can provide a certain class.
3. Uses the method to provide the instance through `get()`.

Right now `WCPay\Mode` is the only example here, but it can be used for all sorts of classes, ex. `WC_Payments_API_Client` from [this question](https://github.com/Automattic/woocommerce-payments/pull/6862#discussion_r1281826677):

```php
private function get_wc_payments_api_client_instance() {
	return \WC_Payments::get_payments_api_client();
}
```

#### Potential classes

Assuming this concept gets approved, before merging this PR, the container should include all classes, available through `WC_Payments::get_[CLASS_NAME]`.

#### Alternatives

What I've considered:

1. Using Woo's `LegacyProxy` ([ref](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Proxies/LegacyProxy.php)): We could use `LegacyProxy` as a dependency, and then `$legacy_proxy->get_instance_of()`, together with hard-coded methods for specific classes ([example](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Proxies/LegacyProxy.php#L69)), similarly to the `LegacyContainer` here. The upside of this approach is that `src` code does not depend on `includes` code, the downside is the extra layer. I don't see a reason for introducing it as an obstacle.
2. The legacy `WC_Payments` class could feed dependencies into the container, but this would mean allowing external modifications of the League container. That's also something that should be avoided IMO.

### Testing instructions

There is not much to manually test, as we've got few `src` classes and instances where they have dependencies.

You can add this snippet somewhere in your code, and open the front-end:

```php
add_action( 'template_redirect', function() {
	$service = wcpay_get_container()->get( ExampleServiceWithDependencies::class );
	var_dump( $service );
	exit;
} );
```

If instantiation succeeds, this means that `LegacyContainer` is working, and the WooPayments container is delegating to it.

I also recommend looking into the additional container tests.

-------------------

- [x] Covered with tests


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6876` or branch name `add/6707-container-additions` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/6707-container-additions) :rocket:

---

Build info: 

- Latest commit: 11aa13bbad4fc461b912967c6e16d8acaa66ad24
- Build time: 2023-08-08 13:39:48 UTC

src/Internal/Service/ExampleService.php
```diff
@@ -0,0 +1,19 @@
...
+namespace WCPay\Internal\Service;
+
+use WCPay\Core\Mode;
+
+/**
+ * This is a service, which will be used for developing
+ * DI-related functionality until there are better services
+ * to test with.
+ */
+class ExampleService {
```
marcinbot, line 17: I hope this is going to be removed in the future?

marcinbot (approved):
Looks good to me, aside from one question below.

src/Internal/Service/ExampleService.php
RadoslavGeorgiev: Yes, I have it in mind. As soon as there are a couple of real-world classes to use, they will replace the example service. I'm confident that it will happen within the payment process re-engineering project.

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
```diff
@@ -0,0 +1,151 @@
...
+
+namespace WCPay\Internal\DependencyManagement\DelegateContainer;
+
+use Psr\Container\ContainerInterface;
+use WC_Payments;
+
+/**
+ * WooPayments Legacy Container Delegate.
+ *
+ * This class is a proxy between `src` and `includes`, allwing
```
htdat:
```suggestion
 * This class is a proxy between `src` and `includes`, allowing
```

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
```diff
@@ -0,0 +1,151 @@
...
+ * `includes` classes to be used as dependencies.
+ */
+class LegacyContainer implements ContainerInterface {
+	/**
+	 * Finds an entry of the container by its identifier and returns it.
+	 *
+	 * @param string $id Identifier of the entry to look for.
+	 * @return mixed Entry.
```
htdat: May use dynamic return type here too? 
See more https://github.com/Automattic/woocommerce-payments/pull/6862#discussion_r1284425974

changelog/add-6707-container-additions
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Adding a LegacyContainer to `src` to allow loading classes from `includes`.
```
htdat:
```suggestion
Add LegacyContainer to `src` to allow loading classes from `includes`.
```

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
```diff
@@ -0,0 +1,151 @@
...
+	 */
+	public function get( $id ) {
+		$method = $this->transform_class_to_method( $id );
+		return $this->$method();
```
htdat: I don't think we would blindly call `$this-> method()` here. 

Instead, use `$this->has($id)` to check before calling it, and then throw an exception `WCPay\Internal\DependencyManagement\ContainerException` in case `has` method return `false`.

htdat (commented):
Left some comments. Other than that, everything looks great.

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/6876/commits/31fb9dff15f5d320564c7b5914b0b7d4cb87e135

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
RadoslavGeorgiev: Fixed in in https://github.com/Automattic/woocommerce-payments/pull/6876/commits/31fb9dff15f5d320564c7b5914b0b7d4cb87e135

changelog/add-6707-container-additions
RadoslavGeorgiev: Fixed in in https://github.com/Automattic/woocommerce-payments/pull/6876/commits/31fb9dff15f5d320564c7b5914b0b7d4cb87e135

src/Internal/DependencyManagement/DelegateContainer/LegacyContainer.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/6876/commits/11aa13bbad4fc461b912967c6e16d8acaa66ad24. FWIW, this container would only be called by the main container, and that one already knows not to use `get` before checking `has`, but additional safety could never hurt.

PR #6880, opened 2023-07-31
BNPLs: updated methods copy on settings page

### LINKED ISSUE #6799: Modify BNPL methods copy in Settings page.
### Description

The BNPL methods are currently appearing with generic copy in the settings, “Expand your business with…”.
image-37

This copy can be enhanced to indicate what the payment methods are to merchants who aren’t familiar.

### Acceptance criteria

* Modify current copy. The exact copy is TBD, please check this thread paJDYF-7XP-p2#comment-19185
* Add a link to relevant section in the public documentation

### Designs

TBD

### Additional context

paJDYF-7XP-p2#comment-19185

mgascam:
Fixes #6799

- Updated Affirm/Afterpay PMs description on the setting page.

#### Testing instructions

* Check settings page to ensure the copy is updated
* Check JS console to ensure no JS errors are being introduced (incl. when checking/unchecking boxes)

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6880` or branch name `update/6799-bnpl-methods-copy-in-settings-page` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=update/6799-bnpl-methods-copy-in-settings-page) :rocket:

---

Build info: 

- Latest commit: 04c3813caae56a0d5967ba691ae1d9dea022eb2f
- Build time: 2023-08-03 08:52:45 UTC

client/payment-methods/index.js
```diff
@@ -222,6 +225,11 @@ const PaymentMethods = () => {
 	const { isUpeEnabled, status, upeType } = useContext( WcPayUpeContext );
 	const [ openModalIdentifier, setOpenModalIdentifier ] = useState( '' );
 
+	const PAYMENT_METHODS_WITH_DYNAMIC_DESCRIPTION = [
```
kalessil: This naming is misleading, can we rename it e.g. `BNPL_PAYMENT_METHODS`. Or even better (if feasible): taking advantage of `allows_pay_later` property available in `client/payment-methods-map.tsx`.

kalessil (changes requested):

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
```diff
@@ -22,10 +22,37 @@ import PaymentMethodsMap from '../../payment-methods-map';
...
+	name: string;
+};
+
+const getDescription = ( name: string, currency: string ) => {
+	const paymentMethod = PaymentMethodsMap[ name ];
+
+	if ( ! paymentMethod ) return null;
+
+	let description = paymentMethod.description;
+	const isAllowingPayLater = paymentMethod.allows_pay_later;
```
kalessil: Looks like we don't need a local variable here.

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
```diff
@@ -22,10 +22,37 @@ import PaymentMethodsMap from '../../payment-methods-map';
...
+
```
kalessil, line 39: No need for this empty line.

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
```diff
@@ -22,10 +22,37 @@ import PaymentMethodsMap from '../../payment-methods-map';
...
+	}
+
+	return description;
+};
+
+const PaymentMethodDescription: React.FC< PaymentMethodProps > = ( {
+	name,
+} ) => {
+	const [ stripeAccountDefaultCurrency ] = useAccountDefaultCurrency();
+	const description = useMemo(
```
kalessil: Nice! Though all descriptions will be unique, and cashing here has no practical value.

kalessil (approved):
That's a clean and well-readable solution. Thanks for working on it @mgascam!
Leaving some minor comments and pre-approving (did manual testing).

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
mgascam: Went for a different approach here [e4f8971](https://github.com/Automattic/woocommerce-payments/pull/6880/commits/e4f8971d33dcf8b99c726edb2115244362072666), hope you find it fine

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
mgascam: Indeed, after a more careful read of the docs, the recommendation is to only use memo when there are expensive calculations involved. Since this is not the case, I'm removing it in this commit [e4f8971](https://github.com/Automattic/woocommerce-payments/pull/6880/commits/e4f8971d33dcf8b99c726edb2115244362072666)

Thanks!

client/components/payment-methods-checkboxes/payment-method-checkbox.tsx
mgascam: Removed here [e4f8971](https://github.com/Automattic/woocommerce-payments/pull/6880/commits/e4f8971d33dcf8b99c726edb2115244362072666)

aheckler:
@mgascam Forgive me if this is a dumb question, but will this PR account for merchants whose Stripe country allows for BNPL, but where their account default currency _does not_?

For example, here's a JN test site where the Stripe account is in the UK (so they are eligible to use Afterpay), but their default currency is EUR:

Screenshot taken on 2023-08-03 at 13 59 39 UTC@2x

If I'm reading your PR right, the copy under Afterpay would state:

> Allow customers to pay over time with Afterpay. Available to all customers paying in EUR.

However, per Stripe's documentation [here](https://stripe.com/docs/payments/afterpay-clearpay), EUR can never be a presentment currency for Afterpay/Clearpay.

Screenshot taken on 2023-08-03 at 14 07 55 UTC@2x

I tried to explain this [here](https://github.com/Automattic/woocommerce-payments/issues/6799#issuecomment-1643448914), but perhaps did not do a good job. :/

mgascam:
Thanks for checking it @aheckler! I'm afraid that I was assuming that the default store currency would match the country's currency. But seeing your example, I was clearly wrong. I'm going to re-open the issue.

> I tried to explain this https://github.com/Automattic/woocommerce-payments/issues/6799#issuecomment-1643448914, but perhaps did not do a good job. :/

I believe it was clear enough, but just for confirmation, we should use the currency of the Stripe account country isn't it?

aheckler:
> we should use the currency of the Stripe account country isn't it?

Yep. The chart in [my comment](https://github.com/Automattic/woocommerce-payments/issues/6799#issuecomment-1643448914) shows the account country to currency mapping.

KarlisJ:
@aheckler @mgascam I want to jump into this discussion.

Similar to Miguel, I was under the assumption that the default store currency would match the country's currency. With the example that Adam showcased, I have some new concerns.

In [Stripe docs for Afterpay/Clearpay](https://stripe.com/docs/payments/afterpay-clearpay/accept-a-payment?platform=web&ui=checkout#compatibility) they have this line under "Determine Compatibility"
> You must express [Prices](https://stripe.com/docs/api/prices) in your domestic currency.

@aheckler, in your example, what is considered "domestic currency"? Is it `GBP` or `EUR`? I worry that in this case, the merchant is actually NOT eligible for Afterpay. In [Restrict PM to domestic country](https://github.com/Automattic/woocommerce-payments/pull/6708#top) we use store default currency as "domestic" currency check. The testing was straightforward - any payments using other presentment currencies failed. 

If in Stripe's mind "domestic" is the same as "default" then this particular example merchant shouldn't be able to receive any payments with Afterpay, because the default currency is EUR which is not supported for PM. 

Have we done testing for the example merchant setup?

aheckler:
> in your example, what is considered "domestic currency"? Is it GBP or EUR?

I would think it's GBP for UK accounts. I will do some testing around this today.

aheckler:
Yeah @KarlisJ, after looking at that PR, it seems like it's using `get_account_default_currency()` to help determine whether to display a BNPL method or not. However, that function pulls the value of `['store_currencies']['default']`, which is the default **deposit** currency for the merchant. The default deposit currency doesn't necessarily match the "domestic"/"local" currency of a given account though, and it's not really relevant for BNPL either.

The two currencies that are relevant to whether to display BNPL or not are:

- The presentment/customer currency, i.e. the currency that shoppers see on the frontend.
- The domestic/local currency of merchant's Stripe account. Those are shown in this chart:


| Account Country | Currency |
|--------|--------|
| Australia | AUD |
| Canada | CAD |
| New Zealand | NZD |
| UK | GBP |
| U.S. | USD | 

That is why [the transaction limits chart](https://stripe.com/docs/payments/afterpay-clearpay#collection-schedule) in Stripe's docs don't contain other currencies.

Screenshot taken on 2023-08-04 at 15 57 03 UTC@2x


As such, if I'm a customer browsing a store that offers BNPL, but I'm looking at prices in EUR, I should never be offered BNPL methods, regardless of the value of `['store_currencies']['default']` for the merchant.

> If in Stripe's mind "domestic" is the same as "default" then this particular example merchant shouldn't be able to receive any payments with Afterpay, because the default currency is EUR which is not supported for PM.

I think this is where the misunderstanding lies. As noted above, the domestic currency and the default deposit currency (which we store as `['store_currencies']['default']`) aren't necessarily the same. 

I'll leave solving this up to y'all, but we probably should start adding some metadata or something to accounts that saves their domestic/local currency, or have some sort of function that can return it from an array or whatever.

aheckler:
Another thing to keep in mind: WooPayments has a multi-currency feature. :) So the presentment/customer currency won't always match the `woocommerce_currency` value that the merchant configured in the WC core settings.

KarlisJ:
Thank you, @aheckler, this is most helpful!

> domestic currency and the default deposit currency (which we store as ['store_currencies']['default']) aren't necessarily the same.

My assumption is that, in most cases, they happen to match. This is why we have incorrectly implemented the check-in [Restrict PM to the domestic country](https://github.com/Automattic/woocommerce-payments/pull/6708#top). We should, similarly, check the country's local currency instead of the default currency. 
(I'll use JN site you provided to validate this assumption)

> we probably should start adding some metadata or something to accounts that saves their domestic/local currency, or have some sort of function that can return it from an array or whatever

If the above assumption is correct, we'll have to revisit the condition in `is_currency_valid` and, as part of that, will implement such meta or function. 

I think we'll be able to pick this up from here, thanks again.

aheckler:
> in most cases, they happen to match

For BNPL, yes, they do match in most cases:

- Australian merchants can only add AUD bank accounts, so their domestic/local currency will always match `['store_currencies']['default']`.
- New Zealand merchants can only add NZD bank accounts, so their domestic/local currency will always match `['store_currencies']['default']`.
- U.S. merchants can only add USD bank accounts, so their domestic/local currency will always match `['store_currencies']['default']`.

However, in 2 cases they might not match:

- Canadian merchants can add bank accounts in CAD or USD, so their domestic/local currency may not match `['store_currencies']['default']`.
- UK merchants can add bank accounts in CHF, DKK, EUR, GBP, NOK, SEK, or USD, so their domestic/local currency may not match `['store_currencies']['default']`.

PR #6897, opened 2023-08-01
Add RTL support script for webpack

elazzabi:
Fixes #

#### Add support for RTL CSS generation using a webpack config

This adds a new webpack config to generate rtl files to be used for RTL merchants. For every `style.css` file, a new one `style-rtl.css` will be generated. **This will raise our bundle size**.

Those files will be automatically registered and used when a merchant switches their store to an RTL language.

#### Testing instructions

Manual testing steps
* Without this PR, go to Settings > General in WP-Admin
* Switch the site language to Arabic, it should look like this:
Screenshot 2023-08-02 at 15 07 59

* Then go back to Payments > Settings, you should see some problems like wrong padding, margins, etc. Something like this:

slack-imgs

* Now, check this PR locally.
* Run `npm run dev`
* Refresh your page and double-check that all those margins and padding problems were fixed.
* Make a quick tour around the plugin pages and double-check if everything looks good.

-------------------

**Post merge**

### DIFF
--- a/includes/multi-currency/MultiCurrency.php
+++ b/includes/multi-currency/MultiCurrency.php
@@ -351,7 +351,6 @@ class MultiCurrency {
 		$this->register_admin_scripts();
 
 		wp_enqueue_script( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
-		wp_style_add_data( 'WCPAY_MULTI_CURRENCY_SETTINGS', 'rtl', 'replace' );
 		wp_enqueue_style( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
 	}
 
@@ -1248,8 +1247,6 @@ class MultiCurrency {
 			[ 'wc-components', 'WCPAY_ADMIN_SETTINGS' ],
 			\WC_Payments::get_file_version( 'dist/multi-currency.css' )
 		);
-
-		wp_style_add_data( 'WCPAY_MULTI_CURRENCY_SETTINGS', 'rtl', 'replace' );
 	}
 
 	/**
--- a/includes/multi-currency/MultiCurrency.php
+++ b/includes/multi-currency/MultiCurrency.php
@@ -351,7 +351,7 @@ class MultiCurrency {
 		$this->register_admin_scripts();
 
 		wp_enqueue_script( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
-		wp_enqueue_style( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
+		WC_Payments_Utils::register_and_enqueue_style( 'WCPAY_MULTI_CURRENCY_SETTINGS', '', [], WC_VERSION, 'all', true );
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6897` or branch name `add/rtl-support` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/rtl-support) :rocket:

---

Build info: 

- Latest commit: 4da1a3aec8b9627494a728bb944d5438c862987e
- Build time: 2023-08-04 09:16:09 UTC

includes/multi-currency/MultiCurrency.php
```diff
@@ -351,7 +351,7 @@ public function enqueue_admin_scripts() {
 		$this->register_admin_scripts();
 
 		wp_enqueue_script( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
-		wp_enqueue_style( 'WCPAY_MULTI_CURRENCY_SETTINGS' );
+		WC_Payments_Utils::register_and_enqueue_style( 'WCPAY_MULTI_CURRENCY_SETTINGS', '', [], WC_VERSION, 'all', true );
```
ismaeldcom: We are registering the same style twice, with different params here. We could probably do something similar to [the WC `enqueue_style` method](https://github.com/woocommerce/woocommerce/blob/fd9e92e7bd4ff598a3fd8a8857ab4fd6cb2da397/plugins/woocommerce/includes/class-wc-frontend-scripts.php#L176) where it is only registered if the required params are provided, otherwise just enqueued.

ismaeldcom (commented):
Thanks for working on RTL support. It works as expected, and the code looks good, except for duplicated style registrations.

includes/multi-currency/MultiCurrency.php
elazzabi: Thank you @ismaeldcom for the review! I agree and I updated the implementation.

The only thing I did differently from the linked WC method is explicitly checking for `''` inside the if: `if ( '' !== $path )` before registering the style.

ismaeldcom (approved):
Approving it after the latest changes 🙂

I wonder if we should set `$has_rtl` to `true` by default, knowing RTL style sheets are automatically generated. Removing those extra params on every call. Anyway is just a nitpick 🙂

mordeth (approved):
LGTM! ✅

elazzabi:
Thank you for the reviews!

@ismaeldcom I like the idea of making RTL generation the default behavior, so I made that in [be3953f](https://github.com/Automattic/woocommerce-payments/pull/6897/commits/be3953f6af75ddecae3d4854dcacc7ceb969bc89)

elazzabi:
Also, as discussed in Slack, I made `RTL=false` for CSS files in `/assets`, those aren't being transpiled to their `rtl` version, and registering/enqueuing them results in errors in styles applied to the page. https://github.com/Automattic/woocommerce-payments/pull/6897/commits/4da1a3aec8b9627494a728bb944d5438c862987e

PR #6901, opened 2023-08-02
Resolve legacy card initialization issue for cart with subscription products in split UPE

### LINKED ISSUE #6790: Card gateway fails to load when purchasing a subscription with split UPE and WooPay
### Describe the bug
Card gateway appears greyed out and fails to load during the subscription purchase, affecting both mixed carts with subscription scenarios and standalone subscriptions. The issue occurs with split UPE and WooPay enabled on the shortcode checkout page

### To Reproduce

1. Enable split UPE with more than one payment gateway and WooPay
2. Add a subscription product to the cart
3. Proceed to the shortcode checkout
4. Confirm that the card payment gateway isn't loaded and is greyed out

### Screenshots
<p align="center">
    
</p>

### Expected behavior
* The legacy card gateway should load properly and allow users to checkout successfully.
* A subscription bought with the legacy card gateway with split UPE can be renewed without errors.

### Additional context
💡 Blocks checkout page functions properly, purchasing and renewing a subscription.

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/6790

The main problem this PR fixes is the problem of the legacy card gateway not loading properly in the split UPE when users have a subscription product in their cart. As we know, the legacy card gateway initializes for the split UPE only in case of WooPay being enabled, so for this PR, we'd need to enable WooPay so that we're able to reproduce the problem and confirm the fix.

#### Background
In the split UPE, we roll out multiple separate payment gateways, one under another. We load gateways enabled in _Payments -> Settings_ in the admin panel. However, the subscription-based products, only reusable gateways can be used, which means that if a subscription is in the card, only e.g. card should be loaded and gateways such as Bancontact, EPS etc, should be filtered out. 

For the split UPE and split UPE with the deferred intent creation, we [fetch all the payment methods enabled at the checkout](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L734). For any gateway of the type UPE, [there's additional filter logic](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L1002) which should filter out every non-reusable gateway [in case there's subscription product in the cart](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-method.php#L140-L142). At first glance it's clear, that the implementation on `develop` should cover that case and there should be no problem.

However, it turns out, that [`is_subscription_item_in_cart`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/compat/subscriptions/trait-wc-payments-subscriptions-utilities.php#L85) doesn't return the right value. With the subscription in the cart, it returns `false`. Eventually, it was found out, that [`$cart_contains_subscription`](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/helpers/class-wc-helper-subscriptions-cart.php#L21) is uninitialized. It turns out, that the cart is initialized [right here](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-cart-session.php#L84-L224) and this method is a callback [run on the `wp_loaded` hook](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-cart-session.php#L61). `wp_loaded` hook, in turn, is executed after [gateway registration](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L731). **Effectively, for the split UPE we load non-reusable gateways while they should be filtered. This process ends up with an error in the JS scripts, which, in turn, blocks the card gateway from loading.**

#### Solution

There's validation in place, which [denies any attempt to load the cart before `wp_loaded`](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-cart.php#L598-L600), so I don't think it's a good idea to change the execution order and try to initialize the cart before registering gateways. That means that there's no way to ensure that WooCommerce [loads](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-payment-gateways.php#L88) the right list of gateways. 

Luckily, WooCommerce Core gives us a bit of space to inject some filtering logic after gateways are registered. When [returning the gateways list for initialization](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-payment-gateways.php#L150-L164), WC Core [checks if a gateway `is_available`](https://github.com/Automattic/woocommerce.com/blob/trunk/plugins/woocommerce/includes/class-wc-payment-gateways.php#L154). This exact method was introduced in the split UPE gateway implementation, which provides the right information about the cart at the point when it's invoked.

### Before & After 
<details>
  <summary>Split UPE</summary>

| Before  | After |
| image  |   image  |

</details>

<details>
  <summary>Split UPE with deferred intent creation</summary>

| Before  | After |
| image  |   image  |

</details>

##### Split UPE
1. Navigate to WCPay Dev Tools and enable split UPE
2. Enable WooPay and some payment method/s besides the card
3. Add a subscription product to the cart and proceed to the shortcode checkout page
4. Confirm that the legacy card gateway is loaded properly and that non-reusable gateways, in turn, are filtered out. (see _**Before/After**_ section for the split UPE)
5. Pay with a card and confirm the payment is successful
6. Perform steps 1-4 but then pay using WooPay. Although WooPay isn't affected at all, it'd be a good sanity check.
7. Add physical products only to the cart and, confirm that non-reusable gateways are loaded, confirm a successful payment with one of the non-reusable payment method.

##### Split UPE with deferred intent creation
1. Navigate to WCPay Dev Tools and enable split UPE with deferred intent creation
2. Enable WooPay and some payment method/s besides the card
3. Add a subscription product to the cart and proceed to the shortcode checkout page
4. Confirm that the card gateway is loaded properly and that non-reusable gateways, in turn, are filtered out. (see _**Before/After**_ section for the split UPE with deferred intent creation)
5. Pay with a card and confirm the payment is successful
6. Perform steps 1-4 but then pay using WooPay. Although WooPay isn't affected, it'd be a good sanity check.
7. Add physical product/s only to the cart and, confirm that non-reusable gateways are loaded, confirm a successful payment with one of the non-reusable payment method. 
-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6901` or branch name `fix/card-gateway-with-woopay-and-subscriptions` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/card-gateway-with-woopay-and-subscriptions) :rocket:

---

Build info: 

- Latest commit: 8d00503697c71dafaf7e8988129a2ee33b46a5cd
- Build time: 2023-08-04 16:10:52 UTC

mdmoore (approved):
Thanks for all of your research on this @timur27! This tests well in both Split UPE and Split UPE with Deferred Intent Creation with the correct payment gateway types becoming available depending on the type of products in the cart. No errors encountered along the way. 🚢

Issue #6566, opened 2023-06-21
Possibility for uncaught exceptions in `WC_Payments_API_Client`


jessepearson:
### Describe the bug
The `get_intent` and `get_timeline` methods in `WC_Payments_API_Client` do not have a try/catch to account for the possibility of exceptions being thrown in the `request` method. 

https://github.com/Automattic/woocommerce-payments/blob/6.0.0/includes/wc-payment-api/class-wc-payments-api-client.php#L220-L224

https://github.com/Automattic/woocommerce-payments/blob/6.0.0/includes/wc-payment-api/class-wc-payments-api-client.php#L734-L737

### Additional context

Came across this when trying to add code that used `get_intent` and purposely passed a bad intent ID to confirm what would be returned so I could handle accordingly.

RadoslavGeorgiev:
@jessepearson that's expected. Most methods in the client (as well as request classes) don't have any error handling built-in. Do you see nay reason to treat those methods differently?

jessepearson:
@RadoslavGeorgiev I haven't tried, but what if an intent ID was under a test account and `get_intent()` was called on it? If there's an exception thrown, the site will just error out rather than displaying something helpful.

I was working on a helper tool to fix Multi-Currency meta data in orders when it's missing and that's when I found this. I added a character to the end of my `_intent_id` and when using `get_intent()` I get an uncaught exception error.

Can you confirm why it would be expected to have uncaught exceptions in our code?

RadoslavGeorgiev:
> Can you confirm why it would be expected to have uncaught exceptions in our code?

@jessepearson of course not :D

However, it does not make much sense to catch those exceptions in the API client or request classes, as neither of them are aware how to handle those errors. That's why the invoking code should be able to handle them. For example, the payment process has all of its code wrapped within a try-catch block, to simply be able to return an error message.

Similarly, the helper tool knows that it's performing a server request, and should be able to handle its failure.

That's all to say that it's not the API client or request classes' responsibility to handle errors. Instead, they throw exceptions to communicate what happened to the rest of the code.

jessepearson:
@RadoslavGeorgiev If I am reading your response correctly, you are stating that code calling `get_intent()`, for example, should handle catching the exception thrown from the call to `request()` in the method?

RadoslavGeorgiev:
@jessepearson exactly.

[...]

PR #6908, opened 2023-08-03
RPP: Proxies

### LINKED ISSUE #6707: Payment Process: Dependency container additions

RadoslavGeorgiev:
Fixes #6707, specifically this part:

> Copy/imitate the LegacyProxy and ActionsProxy from Core, making them work in unit tests.

Built on top of https://github.com/Automattic/woocommerce-payments/pull/6876, which got too big itself.

This description is simple, as the classes are simple. There are further details in the [README.md](https://github.com/Automattic/woocommerce-payments/blob/2abeae390605b4469ba1031a618c06296e035863/src/README.md#proxies).

This PR introduces the following classes and methods:

- [`HooksProxy`](https://github.com/Automattic/woocommerce-payments/blob/2abeae390605b4469ba1031a618c06296e035863/src/Internal/Proxy/HooksProxy.php)
    - `add_filter`
    - `add_action`
- [`LegacyProxy`](https://github.com/Automattic/woocommerce-payments/blob/2abeae390605b4469ba1031a618c06296e035863/src/Internal/Proxy/LegacyProxy.php)
    - `call_function`
    - `call_static`
    - `has_global`
    - `get_global`

All of those methods are pretty simple, essentially one-liners, which allow access to the outside world if a way, which can be easily verified during tests.

#### Testing instructions

Everything's covered with tests. Not much context to test right now.

-------------------

- [x] Covered with tests
- [x] Docs


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6908` or branch name `add/6707-proxies` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/6707-proxies) :rocket:

---

Build info: 

- Latest commit: 8922ad1ebe99dd11cd736ea7ac77cfa25556ab47
- Build time: 2023-08-10 14:42:23 UTC

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * Returns a global variable.
+	 *
+	 * @param string $name Name of the variable.
+	 * @throws Exception   In case the variable is not set.
+	 * @return mixed
+	 */
+	public function get_global( string $name ) {
+		if ( ! isset( $GLOBALS[ $name ] ) ) {
```
marcinbot: Probably worth reusing the `has_global` function here?

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,72 @@
...
+	 *
+	 * Use this for WP, WC, and other generic non-native PHP functions.
+	 *
+	 * @param string $name          Name of the function.
+	 * @param mixed  ...$parameters Parameters to pass to the function.
+	 *
+	 * @return mixed The response from the function.
+	 */
+	public function call_function( string $name, ...$parameters ) {
+		return call_user_func_array( $name, $parameters );
```
marcinbot, line 28: Should we make a check for whether the function `is_callable`, similar to how we check the globals?

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	/**
+	 * Calls the static method of a class outside of `src`.
+	 *
+	 * Use this for non-`src` classes. `src` classes should only have pure static methods.
```
marcinbot: Maybe let's reword this to something like:
> Static methods on `src` classes should be pure (with no side effects).

ATM it sounds like `src` classes should only be static and pure 😅

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * Returns a global variable.
+	 *
+	 * @param string $name Name of the variable.
+	 * @throws Exception   In case the variable is not set.
+	 * @return mixed
+	 */
+	public function get_global( string $name ) {
+		if ( ! isset( $GLOBALS[ $name ] ) ) {
+			throw new Exception( sprintf( 'The global "%s" is not set.', $name ) );
```
marcinbot: So the tricky bit about exceptions in WP is that there's nothing to catch them. Meaning that we will have to be careful about how we write our entry points to always include `try/catch`.

Since this here is such a low-level framework thing it's worth a consideration. But not sure if there is much we can do TBH other than enforcing practices during PR reviews and educating the devs.

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,72 @@
...
+	 * Use this for non-`src` classes. `src` classes should only have pure static methods.
+	 *
+	 * @param string $class_name    Name of the class.
+	 * @param string $method_name   Name of the method.
+	 * @param mixed  ...$parameters Parameters to pass to the method.
+	 *
+	 * @return mixed The response from the method.
+	 */
+	public function call_static( string $class_name, string $method_name, ...$parameters ) {
+		return call_user_func_array( [ $class_name, $method_name ], $parameters );
```
marcinbot, line 44: Let's also check for `is_callable` here

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/90c785affb6c4c2ea1cd1088aa71b857c793dbd9

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: LOL, good point 🤣 

Fixed in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/187181ac4ea2bc0b214739a6be8cdd1e99c47c62.

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: My thinking is that trying to access an undefined global variable would only generate a notice, but functions and methods kind of should just be there when we're trying to access them, throwing a fatal error otherwise.

With that in mind, what are your thoughts on having a separate `is_callable()` method within the proxy?

I know this is a small detail, but we use many functions to connect with the outside world, and I don't want to make it seem like each of them requires a `try/catch`. WDYT? I just wanted to discuss it a bit first, but I'm fine either way.

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: ... Pending the outcome of https://github.com/Automattic/woocommerce-payments/pull/6908#discussion_r1284079263

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: > So the tricky bit about exceptions in WP is that there's nothing to catch them.

I started a thread in the architecture team Slack channel, maybe that could help with uncaught exceptions.

> But not sure if there is much we can do TBH other than enforcing practices during PR reviews and educating the devs.

That's one of the ways, but there's something else that might help.

Looking at https://github.com/Automattic/woocommerce-payments/issues/6566 (different issue, similar topic), I seem to recall writing code (could not even have been PHP) and being forced by the IDE/coding standards to either catch an exception or add a `@throws` tag. 

image

Something similar to what we have right now ☝️, but a bit more intelligent, detecting not only `throw` statements, but also called methods, which might throw an exception.

Do you know of such a solution that we could maybe apply here?

marcinbot: > Looking at https://github.com/Automattic/woocommerce-payments/issues/6566 (different issue, similar topic), I seem to recall writing code (could not even have been PHP) and being forced by the IDE/coding standards to either catch an exception or add a @throws tag.

Yeah, that's a PHPCS warning, but it's quite simple in its implementation. PHPCS does not have anything like a call stack or usage tracking, so it's not going to check if a function call somewhere deeper also throws an exception.

Looks like [Psalm has some `throws` tracking](https://psalm.dev/docs/running_psalm/configuration/#checkforthrowsdocblock) and it might go a bit deeper than just the current function. Worth checking, although I expect that would start flagging up issues in a lot of already-existing places. Maybe it's possible to configure it so it only enables that option for `src`?

src/Internal/Proxy/LegacyProxy.php
marcinbot: > I know this is a small detail, but we use many functions to connect with the outside world, and I don't want to make it seem like each of them requires a try/catch.

I think the same argument could be applied to throwing a fatal - it could be caught, but either way shouldn't happen because we should be writing code calling functions that are there.

We could leave it as is then, now that you explained why you check and throw for globals but not here:
> My thinking is that trying to access an undefined global variable would only generate a notice, but functions and methods kind of should just be there when we're trying to access them, throwing a fatal error otherwise.

src/Internal/Proxy/HooksProxy.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Class HooksProxy
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Proxy;
+
+/**
+ * Actions Proxy
```
htdat:
```suggestion
 * Hooks Proxy
```

changelog/add-6707-proxies
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: dev
+
+Adding proxies to src, used to access code outside of it (hooks, globals, functions, and static methods).
```
htdat:
```suggestion
Adding HooksProxy and LegacyProxy to src, used to access code outside of it (hooks, globals, functions, and static methods).
```

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Class LegacyProxy
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Proxy;
+
+use Exception; // ToDo: Replace this with another exception.
```
htdat: May better use `ProxyException` now? I don't think we will visit this class often after this PR is merged.

htdat (commented):
Just some nitpick comments from me. After answering Marcin's questions, I think we can merge this PR.

changelog/add-6707-proxies
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/c57fdca57b9394a3ea564c475aee875b7b32d4ba.

src/Internal/Proxy/HooksProxy.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/c57fdca57b9394a3ea564c475aee875b7b32d4ba.

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: Good point, done in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/0af396445bd32401d7c36eab6eb899db26f84087.

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: Quick update: I'm trying this, and I think we should be able to suppress the check for `includes`. However, adding the setting throws a Psalm exception for me. I'll need to dig deeper, before I can understand whether Psalm goes deeper than PHPCS in this regard.

Good news is that I realized Psalm isn't running within `src`, so I fixed that in https://github.com/Automattic/woocommerce-payments/pull/6908/commits/93fc6ac00c8189658cdb3668090d992d42690b75, even though this PR might not be the perfect place for it.

RadoslavGeorgiev: Small update: it works!
image

Only issue is that it worked with checks for `includes` disabled. I'll work on understanding and fixing the exception.

@marcinbot would you be comfortable with me merging this PR, and opening another issue (that I'll start working on immediately) to enable `checkForThrowsDocblock` for `src`? It doesn't feel like something that should be addressed here.

marcinbot: > would you be comfortable with me merging this PR, and opening another issue (that I'll start working on immediately) to enable checkForThrowsDocblock for src? It doesn't feel like something that should be addressed here.

Agreed. This doesn't need to be addressed here. Thanks for digging into it!

RadoslavGeorgiev: Opened https://github.com/Automattic/woocommerce-payments/issues/6953, merging this PR.

RadoslavGeorgiev:
@marcinbot @htdat Could either of you approve the PR, so I can merge it. Only now I realized that both of you left comments, not reviews.

marcinbot (approved):
:shipit:

PR #6930, opened 2023-08-04
Multicurrency: Do not modify deposit percentage for display.

peterwilsoncc:
Fixes: WooCommerce Deposits #\506

As reported in  WooCommerce Deposits #\506, enabling multi-currency modifies the display of percentage deposits.

The issue comes from `WooCommerceDeposits::modify_cart_item_deposit_amount_meta()` attempting to convert a percentage number although it were currency.

A store with a USD base selling a product with a 20% deposit results in the following displaying for select currencies:

AUD: 31%
CAD: 27%
INR: 1656%

I haven't written any tests and the new change relies on detecting a backtrace to the Deposits extension. 

**Before**
Screen Shot 2023-08-04 at 2 58 01 pm

**After**
Screen Shot 2023-08-04 at 2 58 52 pm

#### Testing instructions

1. Configure Multi-currency Payments: using USD (default) and INR will best demonstrate the bug
2. Enable the WooCommerce Deposits extension
3. Configure and publish product to accept deposits:
    * Enable Deposits: Yes - Deposits are optional
    * Deposit Type:: percentage
    * Deposit Amount: 20
    * Default Deposit Selected Type: Pay Deposit
4. View the product on the front end
5. Switch to USD if required, observe the deposit amount is displayed as 20%
6. Switch the currency from USD to INR
7. Observe the deposit amount is displayed as significantly over 100% (the exact amount will vary according to currency fluctuations)

*

-------------------

**Post merge**

### DIFF
--- a/changelog/update-6974-transaction-details-breakdown-labels-and-fees
+++ b/changelog/update-6974-transaction-details-breakdown-labels-and-fees
@@ -1,0 +1,4 @@
+Significance: minor
+Type: update
+
+Updated Transaction Details summary with added fee breakdown tooltip for disputed transactions.
--- a/client/components/tooltip/index.tsx
+++ b/client/components/tooltip/index.tsx
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import React, { useState, useRef } from 'react';
+import classNames from 'classnames';
 import { noop } from 'lodash';
 import { Icon } from '@wordpress/components';
 
@@ -112,6 +113,7 @@ export const ClickTooltip: React.FC< TooltipProps > = ( {
 	buttonLabel,
 	buttonSize = 16,
 	children,
+	className,
 	...props
 } ) => {
 	const [ isClicked, setIsClicked ] = useState( false );
@@ -144,7 +146,10 @@ export const ClickTooltip: React.FC< TooltipProps > = ( {
 				parentElement={ tooltipParentRef.current || undefined }
 				onHide={ handleHide }
 				isVisible={ isVisible || isClicked }
-				className="wcpay-tooltip--click__tooltip"
+				className={ classNames(
+					'wcpay-tooltip--click__tooltip',
+					className
+				) }
 			>
 				{ buttonIcon ? (
 					<div
--- a/client/disputes/utils.ts
+++ b/client/disputes/utils.ts
@@ -18,7 +18,7 @@ import {
 	disputeAwaitingResponseStatuses,
 	disputeUnderReviewStatuses,
 } from 'wcpay/disputes/filters/config';
-import { formatExplicitCurrency } from 'wcpay/utils/currency';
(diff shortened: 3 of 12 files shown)


Mayisha (changes requested):
Thanks @peterwilsoncc for working on this. The deposit percentage on the individual product page is showing up correctly now. However, the charged amount is still incorrect resulting in a negative future payment after checkout.

Here is the flow I tested and the purchase with incorrectly charged amount with currency INR.
deposits

peterwilsoncc:
Thanks for catching that @Mayisha, it looks like that string comes from the deposits extension and requires a sibling branch in the other repo.

Once I have that set up, I'll let you know which branch to check out.

iamdharmesh:
Hi @peterwilsoncc, 

I have checked on this and figured out that the issue happening due to a double conversion of the deposit amount.

Deposits extension use [`$product->get_price()`](https://github.com/woocommerce/woocommerce-deposits/blob/trunk/includes/class-wc-deposits-product-manager.php#L251C49-L251C70) function to get deposit amount. Since it uses the product's get_price function in the view context, currency conversion is already handled by default (via the `woocommerce_product_get_price` filter [here](https://github.com/Automattic/woocommerce-payments/blob/635e38668f37d49143703f2bbb4a76d644e48cdd/includes/multi-currency/FrontendPrices.php#L44)). However, a [filter](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/Compatibility/WooCommerceDeposits.php#L27) added for Deposits' multicurrency compatibility is converting the currency again, which is causing problems with the cart and order totals.

**Solution:**
Since the Deposits extension already calculates deposit amounts using the product's get_price function, currency conversion will work fine without any additional compatibility code for `percentage` and `payment plan` based deposit products. However, for the `fixed_amount` deposit product, we need to utilize the `woocommerce_deposits_fixed_deposit_amount` filter to convert the deposit amount to a specific currency.

Therefore, replacing the entire current compatibility code with the following code will resolve the issue.

```php
add_filter( 'woocommerce_deposits_fixed_deposit_amount', function( $amount ) {
    return $this->multi_currency->get_price( $amount, 'product' );
}, 10 );
```

Thanks.

jeffpaul:
@iamdharmesh so is that a change still needed here in WooCommerce Payments or merely in Deposits?

iamdharmesh:
@jeffpaul Changes are needed here in WooCommerce Payments. However, we can move this fixed_amount filter to the deposits side but in that case as well, we need to remove legacy compatibility functions from the WooCommerce Payments

peterwilsoncc:
@Mayisha @iamdharmesh @jeffpaul I've replaced this PR with https://github.com/Automattic/woocommerce-payments/pull/7247 as I'd pushed to the wrong branch in my fork of the repository.

PR #6954, opened 2023-08-09
RPP: Add Psalm exception handling to src

### LINKED ISSUE #6953: Payment Process: Psalm and Exceptions
Addition to: https://github.com/Automattic/woocommerce-payments/issues/6707
Context: https://github.com/Automattic/woocommerce-payments/pull/6908#discussion_r1283427553

Allow Psalm to deep-track exceptions within `src`. I already tried this in the other PR, and it works:

image

However, that was achieved by disabling `includes` coverage altogether, which is not an option. This issue should try to make both work.

RadoslavGeorgiev:
Fixes #6953

https://github.com/Automattic/woocommerce-payments/pull/6908 introduced exceptions within the new `src` directory, which triggered a discussion around exceptions and catching them.

@marcinbot suggested Psalm's `checkForThrowsDocblock` configuration, which turned out to use full analysis, and report uncaught exceptions from nested methods, not just one thrown in the current method. This PR implements it.

Details:

- Configuration is added to `psalm.xml` to check for exceptions.
- `includes` is excluded from the new check, as there are 287 issues to fix there (I was actually expecting more).
- Even though `includes` is excluded, it was preventing Psalm from working. Turns out that `WCPay\Exceptions\Order_Not_Found_Exception` was used in a few places without `use` statements for it. Those are fixed here.
- The `ExampleService` and `ExampleServiceWithDependencies` classes have exceptions and handlers now, to test the configuration.

#### Testing instructions

1. All checks here should be green.
2. Remove the `@throws` tag and the `try - catch` statement from `src/Internal/Service/ExampleServiceWithDependencies.php`.
3. Execute `npm run psalm` locally, and observe the errors.

-------------------

- [x] Covered with tests (no tests really, this is not PHP)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6954` or branch name `add/psalm-exceptions-to-src` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=add/psalm-exceptions-to-src) :rocket:

---

Build info: 

- Latest commit: ba2a06fa23ed9b6b858ddb0a48b11671182133cb
- Build time: 2023-08-11 04:43:48 UTC

src/Internal/Service/ExampleService.php
```diff
@@ -15,5 +16,12 @@
...
  */
 class ExampleService {
-
+	/**
+	 * Does something.
+	 *
+	 * @throws Exception Whenever Huell is not happy.
+	 */
+	public function do_something_with_exception() {
+		throw new Exception( 'Huell is not happy.' );
```
marcinbot, line 25: 😆

marcinbot (approved):
Looks good! :shipit:

htdat (approved):
Pre-approve PR. Just a minor question but you are welcome to ignore it, and merge PR.

src/Internal/Service/ExampleServiceWithDependencies.php
```diff
@@ -57,9 +58,17 @@ public function __construct(
...
 	 * @return bool
 	 */
 	public function is_in_test_mode() {
+		try {
+			$this->example_service->do_something_with_exception();
+		} catch ( Exception $e ) {
+			// Do nothing, this is fine. We're just testing Psalm.
+			return false;
+		}
+
```
htdat, line 71: I understand this is just for testing Psalm, but for which purpose? 

- With it, `return $this->mode->is_test();` is never reached. 
- And without it, Psalm catches up correctly that this method needs to declare `@throw`.

RadoslavGeorgiev: It's just an example, which ensures that Psalm reads both `src` and `includes`, and understands where exceptions might be thrown. The purpose here isn't valid code, I just wanted to make sure that there are a couple of lines, which when commented out, would yield something like this:

image

Of course, I cannot commit without the `@throws` and `try-catch` because all checks would be 🔴 .

Anyway, those examples will be removed once we have better code to test, so I'm merging.

Issue #5517, opened 2023-02-03
SEPA doesn't support subscription products in split UPE


timur27:
Split UPE doesn't support checking out subscription products with SEPA.

With an old UPE in `develop`:

* Users can use SEPA to buy a subscription product
* Users can re-use SEPA, which are already on the saved payment methods list
* Users cannot save SEPA they use when checking out a subscription.

On the split UPE, however:
* Users cannot use SEPA to buy a subscription product
* Users see SEPA methods that were saved previously, but when using them, there’s an error

Acceptance criteria: 
* Users can use SEPA to buy a subscription product
* Users can re-use SEPA, which are already on the saved payment methods list
* Users cannot save SEPA they use when checking out a subscription.

timur27:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5517?planning-poker) with Zenhub @FangedParakeet @mdmoore

frosso:
Moving to ToDo - de-prioritized SEPA compatibility for the Split UPE.

zmaglica:
@frosso Are you or perhaps your team planning to work on this issue? The reason why I am asking is that I have to triage it and put it in prioritization column, and if you are planning to work on it, I can move it to different queue.

frosso:
@zmaglica yep, we're planning to work on this as a low priority item.

thenbrent:
> Users cannot save SEPA they use when checking out a subscription.

@timur27 are you aware of this ever working? I'm investigating and AFAICS, SEPA has never worked for processing of automatic recurring payments even though it was presented as an option during checkout.

timur27:
Hey @thenbrent - sorry for the late response! I think there are two things and I might have provided a bit unclear description. 

1️⃣ 
> Users cannot save SEPA they use when checking out a subscription.

What I had in mind here was that there is no _Save payment information to my account for future purchases._ checkbox provided for users, when they check out using SEPA with a subscription. This is expected, and I probably need to rephrase this point in the description to make it more clear. 

------------

2️⃣ 
> SEPA has never worked for processing of automatic recurring payments even though it was presented as an option during checkout

[...]

PR #6062, opened 2023-04-18
SEPA: remove declaration of subscriptions support

kalessil:
Fixes #3776

#### Changes proposed in this Pull Request

- Remove subscriptions support declaration for SEPA payment method.
- Note the saved token will reveal, and we keep the underlying SEPA+subs integrations which need more work. Hence our changes are kinda cosmetic.

#### Testing instructions

* Enable SEPA in WooCommerce Payments settings
* Create a subscription product
* Add a subscription to cart and proceed to checkout
* Ensure SEPA is NOT available as payment method there
* Clear cart, add a regular product in there and proceed to checkout
* Ensure SEPA is available there
* Complete the purchase and ensure transaction details are available

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

Issue #5531, opened 2023-02-08
Split UPE: migrate saved SEPA tokens


frosso:
### Describe the bug

If a merchant switches from the "Legacy" UPE to the "Split UPE", saved SEPA tokens will not appear anymore.
We need to find a way to migrate the SEPA tokens to the new SEPA gateway.

We will also need to check if SEPA subscriptions are still working ~(but it should be working fine)~.

### Steps to reproduce
- Enable the "legacy" UPE
- Add SEPA tokens
- Disable the UPE
- Enable the "Split" UPE
- Ensure that SEPA tokens are still available

frosso:
Team Fractal will take care of this - thanks @diegocurbelo and team!

FangedParakeet:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/5531?planning-poker) with Zenhub @mdmoore @timur27

frosso:
Looks like this never worked ( https://github.com/Automattic/woocommerce-payments/issues/3776 ) - closing.

PR #6965, opened 2023-08-09
Resolve renewal failures for subscription orders with the UPE

### LINKED ISSUE #4492: Payment fails if not compatible with all payment types included in Payment intent (invalid_request_error)
### Describe the bug
When a payment intent is sent to Stripe via WooCommerce Payments, all enabled payment types are included in the array of payment types on the payment intent. Because of this, if the payment is not compatible with **all** enabled payment types, it will fail.

In the instance where this issue surfaced 5397304-zen, the user had customers which had initiated subscriptions in $USD via Credit/Debit Card payments, and then when the user changed to a third party currency switcher tool, disabling the built-in Multi-currency functionality, recurring $USD payments began to fail, because the `Bancontact` method, which was enabled, was not compatible with $USD payments, despite these payments using the `Credit/Debit Card` type 

I should note that since my WCPay test environment is US based I have been unable to test/reproduce this behavior.

### To Reproduce

1. In a EUR denominated shop, enable WCPay multi-currency functionality, as well as the `Bancontact` payment method.
2. Initiate a $USD denominated Subscription paid Via `Credit/Debit Card`
3. Deactivate Multi-currency Functionality
4. Attempt to process renewal for $USD denominated subscription
5. See error message: `The currency provided (usd) is invalid for one or more payment method types on this PaymentIntent. The payment method type bancontact only supports the following currencies: eur. Either correct the provided currency, or use a different PaymentIntent to accept this payment method type. (invalid_request_error)`
6. Disable the `Bancontact` payment type
7. See that the payment functions, despite having multi-currency disabled

### Actual behavior

The payment fails, despite $USD being valid with the type of payment being requested (`Credit/Debit Card`)

### Expected behavior

I would expect a $USD denominated payment to function regardless of whether an incompatible payment type is enabled. Or at least for the error message to indicate that they payment has failed due to multi-currency functionality being disabled if that is the reason.

Further I would expect that since this payment is of the `Credit/Debit Card` type, only that type would be included in the array of payment types in the Payment Intent.

* OS: [macOS Monterey v 12.4]
* Browser [Chrome]
* Version [Version 103.0.5060.134 (Official Build) (arm64)]

### Additional context

From their logs:

```
'payment_method_types' =>
array (
0 => 'card',
1 => 'bancontact',
2 => 'eps',
3 => 'giropay',
4 => 'ideal',
5 => 'sofort',
6 => 'p24',
),
'off_session' => 'true',
)
2022-07-19T20:14:49+00:00 ERROR The currency provided (usd) is invalid for one or more payment method types on this PaymentIntent. The payment method type bancontact only supports the following currencies: eur. Either correct the provided currency, or use a different PaymentIntent to accept this payment method type. (invalid_request_error)
```

### LINKED ISSUE #6291: New deferred intent UPE compatibility with WooPay
### Description

Currently, the split-UPE achieves compatibility with WooPay by falling back to the legacy card element for the CC gateway when WooPay is enabled. With the deferred intent UPE, this is no longer a necessity, since we will create the payment method prior to confirming the intent, allowing us to perform the regular WooPay cloning procedures in time.

This issue is to add WooPay compatibility to the deferred intent UPE card payment method. At this point, when the deferred intent UPE is enabled, we will no longer need to include any legacy Stripe scripts and we can present a UPE payment method in all available payment gateways. Consequently, we will need to amend the code where gateways and scripts are registered to reflect this.

As discussed in our roundtable, we would also like this issue to cover all available WooPay flows, so this should support an implementation in both the shortcode and blocks checkouts. 

### Acceptance criteria

* When WooPay is enabled, UPE card payment method remains.
* WooPay registration and checkout are both functional.
* WooPay is implemented on both blocks and shortcode checkouts.

### Dev notes

We should to reuse as much as possible from the existing legacy card gateway implementation, but from this vantage point I cannot yet see how much will be reusable. 

### Additional context

Blocked by #5020

FangedParakeet:
Fixes #4492 
Fixes #6291 

# Fixing dysfunctional subscriptions with split and legacy UPE

This PR simplifies and clarifies the logic used to select an array of payment method type IDs to attach to a payment intent. We first check whether `$_POST` parameters are present--in which case, the request is from a standard checkout form. If the `$_POST` parameters are missing, this may be a subscription renewal--in which case, we use the attached payment token to ascertain the payment gateway and ensure we use the currency on the order to filter payment methods. If the scenario fits neither case, we return the payment methods enabled by the current gateway.

Hopefully this should fix...
- Subscription renewals using UPE where the request context (`$_POST[ 'payment_method' ]`) is not present.
- Subscription renewals where store currency differs from order currency.
- Subscription renewals with Link as a payment method.

## Testing subscription renewals

This is the main feature that needs to be tested in this PR and can generally be performed in two ways.

First ensure your store contains at least one subscription product, add it to cart, and complete a checkout for this product. 

### Trigger subscription renewal from order page

1. Find and open the order that you just completed the checkout for.
2. Scroll down and select the subscription order related to this order.
3. From the dropdown options in the top-right of the subscription order detail page, select "Process renewal".
4. Confirm that renewal completes and order status and notes update as expected.

### Trigger subscription renewal from scheduled actions

1. On the order received page, note the order ID of the subscription order related to your checkout order.
2. Visit the scheduled actions tab (**WooCommerce -> Status -> Scheduled Actions**).
3. Search for a 'Pending' hook named 'woocommerce_scheduled_subscription_payment' and find the corresponding entry that matches your subscription order ID.
5. Select "Run" to trigger the renewal for this subscription.
6. Confirm that renewal completes and order status and notes on order detail update as expected.

Ideally, we should test subscription renewals with all UPEs disabled, legacy UPE enabled, split UPE enabled, and deferred intent UPE enabled. You can toggle these features using the dev tools. 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6965` or branch name `fix/upe-subscriptions-schisms` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/upe-subscriptions-schisms) :rocket:

---

Build info: 

- Latest commit: 4061bb79707221324ce4f4c912b4c0105a0ae628
- Build time: 2023-08-14 18:37:57 UTC

csmcneill:
Once this is reviewed/tested/merged, how would we feel about sharing a "beta" build with 6430073-zen? They are greatly affected by #4492:

> We are losing lots of revenue (we are talking about thousands of euros per month) just because people cannot use their local currency.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
 	/**
-	 * Get the payment methods used in the request.
+	 * Get payment method types to attach to intention request.
 	 *
+	 * @param Payment_Information $payment_information Payment information object for transaction.
 	 * @return array List of payment methods.
 	 */
-	private function get_payment_methods_from_request() {
-		$upe_payment_method = sanitize_text_field( wp_unslash( $_POST['payment_method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification
+	public function get_payment_method_types( $payment_information ) {
```
kalessil: I guess for the supported PHP versions we can harden this method a bit: `public function get_payment_method_types( Payment_Information $payment_information ): array`

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
+		$request_payment_method = sanitize_text_field( wp_unslash( $_POST['payment_method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification
```
kalessil: Naming proposal: `$requested_payment_method`

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
+		$token                  = $payment_information->get_payment_token();
+
+		if ( ! empty( $request_payment_method ) ) {
```
kalessil: Why do we prioritize posted PM over token? Can we add a comment explaining the order in which are trying to pick PMs?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
+			$payment_methods = $this->get_payment_methods_from_gateway_id( $request_payment_method );
+		} elseif ( ! is_null( $token ) ) {
+			$order           = $payment_information->get_order();
+			$order_id        = is_a( $order, 'WC_Order' ) ? $order->get_id() : null;
```
kalessil: Let's use `$order instanceof WC_Order` - a better practice.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
-			$payment_methods = [ str_replace( 'woocommerce_payments_', '', $upe_payment_method ) ];
+	/**
+	 * Get the payment methods used in the request.
+	 *
+	 * @param string $gateway_id ID of processing payment gateway.
+	 * @param int    $order_id ID of related order, if applicable.
+	 * @return array List of payment methods.
+	 */
+	public function get_payment_methods_from_gateway_id( $gateway_id, $order_id = null ) {
+		if ( 'woocommerce_payments' !== $gateway_id ) {
```
kalessil: I believe we should check here is the id start with `woocommerce_payments_` as well. WDYT?

kalessil: Also, I lack a comment here explaining which gateways have this sort of IDs. Can you add it, please?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
+			$payment_methods = [ str_replace( 'woocommerce_payments_', '', $gateway_id ) ];
 		} elseif ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
```
kalessil: That's difficult to understand can we introduce separate ifs for `WC_Payments_Features::is_upe_split_enabled()` and `WC_Payments_Features::is_upe_deferred_intent_enabled() )`?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
 			$payment_methods = [ 'card' ];
```
kalessil: There is a constant for card PM.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1375,28 +1375,52 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
 			if ( WC_Payments_Features::is_upe_deferred_intent_enabled() &&
 				in_array( Payment_Method::LINK, $this->get_upe_enabled_payment_method_ids(), true ) ) {
 				$payment_methods[] = Payment_Method::LINK;
 			}
 		} else {
-			$payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( null, true );
+			$payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( $order_id, true );
```
kalessil, line 1444: Can we add a comment explaining when we get into this branch please?

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Sure thing--added in 178d0760d108485e4ffa4debf73c0fa62ad35969.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Makes sense--added in 178d0760d108485e4ffa4debf73c0fa62ad35969.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: I'm basically prioritising checkout transactions over subscription renewals, since I believe the former will be more common: explained over [here](https://github.com/Automattic/woocommerce-payments/blob/fix/upe-subscriptions-schisms/includes/class-wc-payment-gateway-wcpay.php#L1404).

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Good call--updated in 178d0760d108485e4ffa4debf73c0fa62ad35969.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: At this point, we know that the ID will start with `woocommerce_payments`; if the ID is anything other than `woocommerce_payments`, we presume that it will begin with `woocommerce_payments_` as this should be the only remaining available option. 

I simplified this (with a few explanatory comments) in 178d0760d108485e4ffa4debf73c0fa62ad35969 by simply checking whether the ID begins with `woocommerce_payments_` first instead. However, I don't really feel a need to also confirm that the ID begins with `woocommerce_payments`, as I don't believe that it's possible for neither of those cases to be `true`.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Separate the clauses in 178d0760d108485e4ffa4debf73c0fa62ad35969.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: Good shout--amended in 178d0760d108485e4ffa4debf73c0fa62ad35969.

includes/class-wc-payment-gateway-wcpay.php
FangedParakeet: I've added a comment explaining the logic a bit better in 178d0760d108485e4ffa4debf73c0fa62ad35969 and 08e0daa2e492c230b060d6e8ffe983cca7949b99. We will get here if the `$gateway_id` is `woocommerce_payments` and neither split nor deferred intent UPE is enabled--so this is either the legacy UPE or legacy card gateway.

FangedParakeet:
@kalessil, just seen your comment over [here](https://github.com/Automattic/woocommerce-payments/pull/6062#issuecomment-1674447646). I tested this locally and I think this does enable us to process renewals for subscriptions using SEPA as a payment method...but this wasn't particularly intended by me when I authored this PR. 😅 

I was actually under the impression that we would continue proceeding to keep SEPA subscriptions disabled--consequently we have closed #5517 and #5531. Can you confirm whether this is still the case or whether we intend to give SEPA subscriptions another chance? In the latter case, I would have to reassess whether SEPA subscriptions integration should be included in our upcoming deferred intent UPE feature release.

gpressutto5 (approved):
It's working great!

PR #6975, opened 2023-08-11
Fix AutomateWoo error on WooPay redirection

alefesouza:
Info p1691695202653879-slack-C022WMN88KG

Fix error when redirecting to WooPay checkout with AutomateWoo installed.

#### Testing instructions

* Install AutomateWoo.
* Checkout with WooPay.
* The redirection should occur as expected.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-automatewoo-error-woopay-redirection
+++ b/changelog/fix-automatewoo-error-woopay-redirection
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix AutomateWoo error on WooPay redirection.
--- a/includes/woopay/class-woopay-adapted-extensions.php
+++ b/includes/woopay/class-woopay-adapted-extensions.php
@@ -7,23 +7,16 @@
 
 namespace WCPay\WooPay;
 
-use Automattic\WooCommerce\Blocks\Integrations\IntegrationInterface;
+use Automattic\WooCommerce\Blocks\Integrations\IntegrationRegistry;
 
 /**
  * WooPay
  */
-class WooPay_Adapted_Extensions {
+class WooPay_Adapted_Extensions extends IntegrationRegistry {
 	const POINTS_AND_REWARDS_PLUGIN = 'woocommerce-points-and-rewards';
 	const POINTS_AND_REWARDS_API    = 'points-and-rewards';
 	const GIFT_CARDS_API            = 'woocommerce-gift-cards';
 	const GIFT_CARDS_BLOCKS         = 'wc-gift-cards-blocks';
-
-	/**
-	 * WC Blocks registered integrations, as `$name => $instance` pairs.
-	 *
-	 * @var IntegrationInterface[]
-	 */
-	protected $registered_integrations = [];
 
 	/**
 	 * Initializa WC Blocks regitered integrations.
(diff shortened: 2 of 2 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6975` or branch name `fix/automatewoo-error-woopay-redirection` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&branches.woocommerce-payments=fix/automatewoo-error-woopay-redirection) :rocket:

---

Build info: 

- Latest commit: 0d9a5d8e755a164f991e57c81d571e0d59c61335
- Build time: 2023-08-12 04:13:56 UTC

malithsen (approved):
Thanks @alefesouza can confirm this fixes the issue with AutomateWoo.

I am a bit uneasy about having to extend the `IntegrationRegistry`, mostly due to my unfamiliarity with Blocks integrations. We are calling the `woocommerce_blocks_checkout_block_registration` hook which will make many other extensions that have integrated with Blocks to react to the hook, not just the "WooPay Adapted Extensions". Is there a way we could only attempt registering the extensions we want? 

If that's not possible, I am okay with merging this as is. 
Would also like to get a second pair of eyes for the review if possible.

hsingyuc (approved):
Test as expected and LGTM!

hsingyuc (changes requested):
>I am a bit uneasy about having to extend the IntegrationRegistry, mostly due to my unfamiliarity with Blocks integrations. We are calling the woocommerce_blocks_checkout_block_registration hook 

I just saw this now and agree with @malithsen.

@alefesouza could you explain why we need this hook?  I see that it was being used before this PR, but I'm also curious how we're using it in wc pay.  It seems like it's only supposed to be used inside of a block.

If we do use this, it also seems like we're missing a `registry_identifier` property that would [make this work in IntegrationRegistry](https://github.com/woocommerce/woocommerce-blocks/blob/5f20ebb65a8f420ece6a71714cda1c049b8a6281/src/Integrations/IntegrationRegistry.php#L50).

alefesouza:
@hsingyuc by default the plugins integrations use this hook to be integrated into WC Blocks, I've preferred this approach instead of instantiate the plugin integration class (see `WC_Points_Rewards_Manager` [here](https://github.com/Automattic/woocommerce-payments/blob/6e607c0404c825a0f62f91d65c697ed257aedaca/includes/woopay/class-woopay-adapted-extensions.php#L36) and `WC_GC_Checkout_Blocks_Integration` [here](https://github.com/Automattic/woocommerce-payments/blob/6e607c0404c825a0f62f91d65c697ed257aedaca/includes/woopay/class-woopay-adapted-extensions.php#L68)) to prevent errors in case of the plugin rename this class name, as they always will need to run this hook to be added to blocks.

> this, it also seems like we're missing a registry_identifier property

We only need the "checkout" registry identifier to integrate the plugins on WooPay checkout.

I can follow Malith suggestion and filter the `register` method only to the adapted extensions.

hsingyuc (approved):
Going to approve this since the hook was previously used before this PR. 

It might be good later to figure out a better way of doing this since using this hook seems like it might have unexpected behavior.

Issue #6263, opened 2023-05-10
WooPay Button Logo cut off in certain views


bborman22:
### Describe the bug

This issue was noticed while viewing the WooPay Express Checkout button on the blocks cart with the base storefront theme when there is an additional express payment option (Google Pay or Apple Pay). It appears that the logo will scale down so that the words are still readable and in some cases (when the CTA is Donate with) the logo wasn't displayed at all. I believe this will require the three column layout on the block cart page that comes with out of the box Storefront theme.

### To Reproduce

1. Have a merchant store with WooPay and Google or Apple Pay enabled and displayed on the cart page.
2. For the Express checkout buttons turn on a CTA text (Buy with, Donate with, or Book with).
3. Have a merchant store with the Storefront theme that has three columns on the cart page.
4. Have a merchant store with a cart block page.
5. As a shopper add a product to the cart and visit the cart block page.
6. Observe that the WooPay logo in the button is either scaled down very small or not visible at all in the case of Donate with.

### Actual behavior

The WooPay logo displayed within the button is scaled down and barely visible in the above scenario.

### Screenshots

![Screen Shot 2023-05-02 at 1 59 54 PM](https://github.com/Automattic/woocommerce-payments/assets/68524302/3544abf9-c511-48c9-9f3a-08cb66b43ec4)

### Expected behavior

The WooPay logo should stay scaled up and visible similar to how the Google Pay logo works. This appears to be achieved by making margins and text size smaller while making the logo only slightly smaller.

PR #6985, opened 2023-08-11
Fix Express Checkout Button Design Issues

bborman22:
Fixes https://github.com/Automattic/woopay/issues/1854.

This changes the design around the express checkout buttons and relevant buttons at the product and cart pages:

-  Buttons are vertically aligned.
- "Add to Cart" and "Proceed to Checkout" buttons get precedence and are listed on top of other buttons.
- The OR separator is removed from the product and cart pages (remains on checkout page).

#### Testing instructions

Check out the express checkout buttons +
-  "Add to cart" button on the product page.
- "Proceed to checkout" button in the cart.
- "Place order" button at checkout.

Confirm all of the aforementioned buttons look OK and similar in those pages. Our base reference is the Storefront theme, but tests on other themes are appreciated. The requirements are:

-  Buttons are vertically aligned.
- "Add to Cart" and "Proceed to Checkout" buttons get precedence and are listed on top of other buttons.
- The OR separator is removed from the product and cart pages (remains on checkout page).

Note: if you find buttons placed weirdly in other themes and that's caused by the theme's CSS, that's alright, as we're not looking to override the theme's styles.

#### Storefront
- Product page
image

- Cart
image

- Checkout
image

#### Astra
- Product page (we have a theme conflict here – element with `float: left`)
image

- Cart
image

- Checkout
image

#### Botiga
- Product page
image

- Cart
image

- Checkout
image

#### Minimalist eCommerce
- Product page (we have a WooCommerce conflict here – element with `float: left`)
image

- Cart
image

- Checkout
image

#### Neve
- Product page
image

- Cart (we have a theme conflict here – element with `display: flex`)
image

- Checkout
image

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `6985` or branch name `fix/express-button-design` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/express-button-design) :rocket:

---

Build info: 

- Latest commit: ce3b77ab07179611042bc87f114b7f115408b435
- Build time: 2023-08-23 13:07:15 UTC

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -761,6 +761,13 @@ public function scripts() {
 
 		WC_Payments::register_script_with_dependencies( 'WCPAY_PAYMENT_REQUEST', 'dist/payment-request', [ 'jquery', 'stripe' ] );
 
+		wp_enqueue_style(
+			'WCPAY_PAYMENT_REQUEST',
+			plugins_url( 'dist/payment-request.css', WCPAY_PLUGIN_FILE ),
+			[],
+			'1.0'
+		);
```
rafaelzaleski: To maintain consistency with the rest of the plugin, I believe we should use:

```suggestion
		WC_Payments_Utils::register_style(
			'WCPAY_PAYMENT_REQUEST',
			plugins_url( 'dist/payment-request.css', WCPAY_PLUGIN_FILE ),
			[],
			WC_Payments::get_file_version( 'dist/payment-request.css' )
		);
```

rafaelzaleski (approved):
The code looks good and all the tests are passing; The buttons are displayed as described.

I've made a minor comment regarding the way we handle the CSS enqueue, but I'll still leave the PR pre-approved since it's not critical.

bborman22:
@pierorocca and @elizaan36 We dropped a few screenshots of the new button layouts include a few different themes and calling out some places where there were conflicts including the style causing the conflict. Let me know if these all look as expected with the exception of the conflicts!

pierorocca:
Thanks for the preview Brian!! Much appreciated especially the due diligence of testing against multiple themes.

> @pierorocca and @elizaan36 We dropped a few screenshots of the new button layouts include a few different themes and calling out some places where there were conflicts including the style causing the conflict. Let me know if these all look as expected with the exception of the conflicts!

Thanks @bborman22. In almost all of those examples the medium sized button at 48px feels undersized. That makes me ask @elizaan36 if 52px height is a better default medium size?

**General comments**
- The spacing between buttons looks a lot larger than the designs and considerably larger than Shopify's design. Please confirm.
- While the WooPay button follows the theme Case, the PRBs are in Sentence case. I don't believe we can change the PRB case unless we move to implement custom buttons. @elizaan36 should WooPay align with the theme case on the PRB Sentence case?
- What does this all look like on mobile?

**Astra**
- The product page conflict, are we in general unable to override the theme styling?

**Botiga**
- The spacing between Add to cart and the WooPay button looks inconsistent to and larger than on the other themes.

**Minimalist eCommerce**
- Similar comment to Astra, we're unable to override the theme styling? 

**Neve**
- As above about overriding. On the few Shopify sites I could find enabled with GPay on the cart (the trend - which I like - seems to be moving away from PRBs except on the checkout page), the cart and PRB buttons are in different containers while one site wrapped both into a `form_cta_fix` class. https://rothys.com/products/the-driver-spruce-camo 
- Are there other ways of avoiding these conflicts?

cesarcosta99:
### Mobile Screenshots

#### Storefront

<table>
<tr>
<th>Product page</th>
<th>Cart</th>
<th>Checkout</th>
</tr>
<tr>
<td>
image
</td>
<td> 
image
</td>
<td>
image
</td>
</tr>
</table>

#### Astra

<table>
<tr>
<th>Product page *</th>
<th>Cart</th>
<th>Checkout</th>
</tr>
<tr>
<td>
image
</td>
<td> 
image
</td>
<td> 
image
</td>
</tr>
</table>

* Same conflict from desktop: we have a theme conflict – element with `float: left`.

#### Botiga

<table>
<tr>
<th>Product page</th>
<th>Cart</th>
<th>Checkout</th>
</tr>
<tr>
<td>
image
</td>
<td> 
image
</td>
<td> 
image
</td>
</tr>
</table>


#### Minimalist eCommerce

<table>
<tr>
<th>Product page *</th>
<th>Cart</th>
<th>Checkout</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
<td> 
image
</td>
</tr>
</table>

* Same conflict from desktop: we have a theme conflict – element with `float: left`.

#### Neve

<table>
<tr>
<th>Product page</th>
<th>Cart *</th>
<th>Checkout</th>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
<td> 
image
</td>
</tr>
</table>

* Same conflict from desktop: we have a theme conflict – element with `display: flex`.

pierorocca:
Thanks @cesarcosta99. @elizaan36 @stephendharmadi could use your expertise here before this is merged.

bborman22:
Also to address the conflict questions (including the spacing) we were finding that any fixes would require overriding the theme styles (which we were trying to avoid) and would more times than not break something in another theme, leading to a whack a mole situation. 

> the cart and PRB buttons are in different containers while one site wrapped both into a form_cta_fix class.

This actually touches closely to one piece of this PR I consider important in that we are now [wrapping both buttons in a container](https://github.com/Automattic/woocommerce-payments/pull/6985/files#diff-3e2029df35f2c6fe2b71f0f467c377316c92e31c43c2cb2c32b44f4b0715a41aR95.). This gives theme developers an easy way to move and adjust this container as needed, whereas previously they may have had to apply styles to individual buttons (or both buttons at the same time), but didn't have a wrapper to move the whole area around. This doesn't address the conflicts as seen in some of the screenshots (as the container itself is still subject to it's surrounding content and styles), but it should make it easier to fix by the themes themselves.

elizaan36:
> The spacing between buttons looks a lot larger than the designs and considerably larger than Shopify's design. Please confirm.

I agree the spacing looks wide but sounds like we can't change that without theme conflicts.

> @elizaan36 should WooPay align with the theme case on the PRB Sentence case?

All caps on the WooPay button looks awkward so I'd lean towards Sentence case to match the other payment buttons.

> Button height

Yes, this is a challenge. I see in the other issue that we're concerned about overriding merchant settings if they've specifically chosen a particular button size. I agree with this. We're also not taking into account 3PD payment buttons such as PayPal that appear with their own height and styles currently, so there's a long term unification that needs to be explored further. For the time being I think this default height is fine.

pierorocca:
@bborman22 can you double check the spacing please. Top image is the design and bottom is what has been implemented.

image
image

bborman22:
Sure thing, I've tightened up the spacing from 1.5em to 1em now:

**Storefront**


| Product | Cart | Checkout 
| :---: | :---: | :---:
| Screenshot 2023-08-18 at 8 54 37 AM | Screenshot 2023-08-18 at 8 51 25 AM | Screenshot 2023-08-18 at 8 51 32 AM

As an example of where conflicts between themes could occur here is the spacing on the Astra theme cart page, notice the gap between proceed to checkout in Storefront is larger than the gap between proceed to checkout on Astra, if we tighten Storefront, Astra will be too close and vice versa.

Screenshot 2023-08-18 at 8 56 08 AM

@pierorocca Let me know if this looks better and I can push these changes up!

Edit: I've also made it so that the button will always use the sentence case instead of going uppercase in certain themes.

pierorocca:
Looks good Brian. On mobile the Astra example looks OK. Relative to Shopify's spacing we're still larger so unless I'm missing something it looks better all around.

bummytime:
Nice work on this @bborman22 :shipit:

bborman22:
Just an update that we're waiting on a question to some of our internal theme developers to explore any best practices we could implement here before merging.

bborman22:
So after some experimentation, I found that we might get more broad compatibility with the buttons if we used the `woocommerce_after_add_to_cart_form` action instead of the `woocommerce_after_add_to_cart_button` action. This allows us to break our button container outside of the themes quantity and add to cart container. The only drawback is that now the Add to Cart button won't be uniform with the other express checkout buttons, but the big pro is the wider compatibility.

This change only impacts the product page, so I'm only including the product page screenshots. @elizaan36 and @pierorocca let me know what you think of this update. This will require some additional functional testing, but the initial testing all passed. I'd just want to see if this is the right path forward for now on the product pages.

**Storefront**

| Desktop | Mobile
| :--: | :--:
|  Screenshot 2023-08-21 at 4 13 22 PM | Screenshot 2023-08-21 at 4 31 26 PM

**Astra**

| Desktop | Mobile
| :--: | :--:
|  Astra-Desktop |  Astra-Mobile

**Botiga**

| Desktop | Mobile
| :--: | :--:
|  Botiga-Desktop |  Botega-Mobile

**Minimalist eCommerce**

| Desktop | Mobile
| :--: | :--:
|  MinimalisteCommerce-Desktop | MinimalisteCommerce-Mobile

**Neve**

| Desktop | Mobile
| :--: | :--:
| Neve-Desktop | Neve-Mobile

pierorocca:
@bborman22 thanks again for entertaining my many questions and for finding another potential option. @elizaan36 would appreciate your eyes on this, what you feel are the pros/cons, and your overall recommendation for how we proceed on the product page.

pierorocca:
cc @stephendharmadi

elizaan36:
This feels way better to me because it creates enough visual distinction between the "Add to cart" and express buttons. Stephen and I were chatting the other day about creating a better default for the Add to cart button since it was hard to visually separate and understand it's purpose compared to the express buttons. This helps.

LGTM!

elizaan36:
All that being said, I really want to make sure we're not implementing anything that will break other themes. This should go through comprehensive testing.

pierorocca:
@elizaan36 following this now long thread, you're in favor of the new approach over the original approach because 1) less theme breakage and 2) better visual distinction between 'Add to cart' and express buttons?

elizaan36:
Yes, and it seems like a much more versatile technical approach that would cause fewer conflicts. Separating out the Quantity + Add to cart from the buttons makes sense.

pierorocca:
@elizaan36 makes sense. Assuming "Buy with" or other button labels are configured, the above screenshots would show even more distinction and clarity between the two.

@bborman22 is "Buy now" the default setting when enabling WooPay or Google Pay/Apple Pay for the first time? If not, "Buy with" would be a good default.

LGTM.

bborman22:
Thanks everyone for taking a look! I'm going to have this get another round of testing/review before merging and that will give us a week of the change being in the `develop` branch where others could spot issues as well as another round of GS testing. @cesarcosta99 would you mind giving these latest changes another round of testing and code review? I want to confirm that styling as well as functionality are working as expected since we changed the hook being used (I tested against all different product types to confirm: simple, variable, digital, and subscription and against all the themes I've screenshot.).

I've also confirmed that "Buy with" is the default option out of the box.

includes/class-wc-payments-woopay-button-handler.php
```diff
@@ -146,6 +146,12 @@ public function scripts() {
 		}
 
 		WC_Payments::register_script_with_dependencies( 'WCPAY_WOOPAY_EXPRESS_BUTTON', 'dist/woopay-express-button' );
+		wp_enqueue_style(
+			'WCPAY_WOOPAY_EXPRESS_BUTTON',
+			plugins_url( 'dist/woopay-express-button.css', WCPAY_PLUGIN_FILE ),
+			[],
+			'1.0'
+		);
```
cesarcosta99: 🤏 Nit: to keep consistency in this PR, as we applied https://github.com/Automattic/woocommerce-payments/pull/6985#discussion_r1294082482, we should update this as well:

```suggestion
		WC_Payments_Utils::register_style(
			'WCPAY_WOOPAY_EXPRESS_BUTTON',
			plugins_url( 'dist/woopay-express-button.css', WCPAY_PLUGIN_FILE ),
			[],
			WC_Payments::get_file_version( 'dist/woopay-express-button.css' )
		);
```

cesarcosta99 (approved):
Awesome work, @bborman22! Code looks good, everything worked as expected and styling-wise it looked better by inserting the express checkout buttons in new action indeed 🎉  I tested with multiple types of products, multiple themes and against mobile and desktop viewports.

Regarding the action change in particular, I looked into the WC core code and I don't have any concerns and haven't seen any drawbacks. On the contrary, semantically the usage of the `woocommerce_after_add_to_cart_form` action in this case makes more sense as the express checkout buttons are not part of the add to cart form per se.

I left only a small nitpick, but I'm pre-approving it since it's not a blocker.

includes/class-wc-payments-woopay-button-handler.php
bborman22: Thanks for this! I updated this, but then found that the CSS wasn't being loaded then. I ended up using `WC_Payments_Utils::enqueue_style` and it puts the styles on the page. I was wondering if you could take a look at that commit and make sure what I did use the desired pattern there?

pierorocca:
> ion change in particular, I looked into the WC core code and I don't have any concerns and haven't seen any drawbacks. On the contrary, semantically the usage of the woocommerce_after_add_to_cart_form action in this case makes more sense as the express checkout buttons are not part of the add to cart form per se.

Following your semantic logic @cesarcosta99 what does Neve and Minimalist ecommerce cart pages look like if we hook into woocommerce_after_cart_totals instead? Does it fix the conflict but then makes the proceed to checkout button less than container width? Is it possible to do both, force container width size for Proceed to checkout button and insert the PRBs after the cart_totals?

includes/class-wc-payments-woopay-button-handler.php
cesarcosta99: I tested it and it happened the same to me 🤔 I find it weird that it worked in the aforementioned case. Do you think it'd be safer to replace `register_style` with `enqueue_style` in the other case too? I noted that the enqueue helper registers the style as well.

bborman22: As we discussed on our call, I updated the other `register_style` to also use `enqueue-style` as we discovered the discrepancy was between having different configurations of WooPay and Payment Request options enabled. I think we should do some additional testing around those different configurations now as well to confirm.

pierorocca:
@aheckler here's the enhancement to how express checkout buttons will be styled in WooPayments in the upcoming release (shortcode only - though the product page is not blockified yet). The - OR - divider is removed, Add to cart is moved back into the top position where it should be, and overall we're improving the styling and making it easier for merchants to fix styling issues.

Given the breadth of this change and potential to break styling on sites, this could benefit from extra documentation in the next release to a) inform merchants on why we're doing this b) what to do if something breaks as a result. Thoughts?

pierorocca:
@elizaan36 @nerrad @pmcpinto when these changes go live on WooPayments, it only affects shortcode sites and not yet blockified product pages. 

Are there changes to the Cart block that could be made in order to keep the experiences aligned? The two changes would be to remove the OR and to place Add to cart at the top of the stack. The attached example looks decent as is, though it's not necessarily representative of what's live on real stores. Ok to leave as is due to lower blocks theme usage and the issues likely aren't as severe as on shortcode. Wanted to get your thoughts as this issue winds down.

image

includes/class-wc-payments-woopay-button-handler.php
cesarcosta99: Thanks, Brian! I played with all scenarios we can have with WooPay and Apple/Google Pay enabled/disabled and everything is looking as it should. Marking this as resolved 👍

nerrad:
> Are there changes to the Cart block that could be made in order to keep the experiences aligned? The two changes would be to remove the OR and to place Add to cart at the top of the stack.

Will you explain more about placing the "Add to cart at the top of the stack"? In the context of the Cart block there is no "Add to Cart", are you referring to the "Proceed to Checkout" button?

bborman22:
@pierorocca As I'll be merging this PR today, it might be best to move the conversation to a new issue just so it doesn't get lost on a closed PR.

aheckler:
> Given the breadth of this change and potential to break styling on sites, this could benefit from extra documentation in the next release to a) inform merchants on why we're doing this b) what to do if something breaks as a result. Thoughts?

Yeah, I can check the changes against our express checkout docs and see what might need updating. That raises the question though: what *should* merchants do if this breaks on their site? I assume we can't really help write custom CSS adjustments for every theme under the sun?

pierorocca:
>  I assume we can't really help write custom CSS adjustments for every theme under the sun?

Correct. Bare minimum would be to describe where custom CSS could be applied to affect the button styling and point to any existing WooCommerce articles we may have on button styling, or display: box, display: flex.

pierorocca:
@bborman22 will do an open a new slack thread to determine if we need an issue for Blocks. @nerrad sorry about that I meant the "Proceed to Checkout" button for the cart page.

stephendharmadi:
Hey, I just wonder when I see the WooPay button label looks smaller than Google Pay. Could we update the WooPay button with [the new specifications of the centralized payment button](https://www.figma.com/file/qNVPHJXgRkftiSR7j2XOlj/Centralized-Payment-Button-Guidelines?type=design&node-id=953-39224&mode=design&t=lDMD8CtDdkXdMuEI-4)?

It will be good first step to have the WooPay button to reflect this latest update.

pierorocca:
> Hey, I just wonder when I see the WooPay button label looks smaller than Google Pay. Could we update the WooPay button with [the new specifications of the centralized payment button]

I think this was fixed in a different issue, which I can't find. @bborman22 I think your team was working on this but I can't find the issue it was attached to after searching high and low.

bborman22:
Hmm, I don't recall any work on that. The closest thing I recall is forcing the button color and font colors in #6367

pierorocca:
https://github.com/Automattic/woocommerce-payments/issues/6263 this one I think? That covered off only the logo whereas the new guidance appears to increase the font height. Looking at the previous label sizes, did they happen to shrink recently?

pierorocca:
Looking at the previous label sizes, did they happen to shrink recently?

bborman22:
Ahh you're right, that could definitely be what's causing it as the trade off was to shrink text size to keep the logo size big, but it shouldn't have been happening here. I'll double check to confirm this.

pmcpinto:
> Are there changes to the Cart block that could be made in order to keep the experiences aligned? The two changes would be to remove the OR and to place Add to cart at the top of the stack. The attached example looks decent as is, though it's not necessarily representative of what's live on real stores. Ok to leave as is due to lower blocks theme usage and the issues likely aren't as severe as on shortcode. Wanted to get your thoughts as this issue winds down.

Thanks for the heads up. I'd leave it as is for now, and revisit this change as part of a broader initiative to centralize the payments buttons or improve the shopper experience. 

Btw, has this change been tested in cases where elements such as payment badges are displayed below the "Proceed to Checkout" button? I remember seeing stores using this approach, but I don't have any evidence that this is a common use case.

pierorocca:
> Btw, has this change been tested in cases where elements such as payment badges are displayed below the "Proceed to Checkout" button? I remember seeing stores using this approach, but I don't have any evidence that this is a common use case.

@bborman22 the main pattern to test here is PaPal below the button stack. I doubt it's impacted since Paypal is an iframe but worth testing.

bborman22:
They seem to work well together from my testing, this is the default PayPal button settings in Storefront:


Screenshot 2023-08-29 at 9 48 20 AM

Screenshot 2023-08-29 at 9 48 29 AM

PR #7004, opened 2023-08-15
Fix/5188 duplicate charges delay webhook processing

### LINKED ISSUE #5188: Fix Duplicate Charges: Delay webhook processing, query scheduled before failure

zmaglica:
Fixes #5188

#### Testing instructions

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7004` or branch name `fix/5188-duplicate-charges-delay-webhook-processing` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/5188-duplicate-charges-delay-webhook-processing) :rocket:

---

Build info: 

- Latest commit: 15b1311b5f2ce4b8c47fd0b4a2d9bf980068f52b
- Build time: 2023-09-01 08:24:22 UTC

includes/class-wc-payments.php
```diff
@@ -555,7 +555,7 @@ public static function init() {
 
 		self::$webhook_processing_service  = new WC_Payments_Webhook_Processing_Service( self::$api_client, self::$db_helper, self::$account, self::$remote_note_service, self::$order_service, self::$in_person_payments_receipts_service, self::get_gateway(), self::$customer_service, self::$database_cache );
 		self::$webhook_reliability_service = new WC_Payments_Webhook_Reliability_Service( self::$api_client, self::$action_scheduler_service, self::$webhook_processing_service );
-
+		add_action( 'init', [ self::$webhook_reliability_service, 'init' ] ); // Add custom post type and taxonomies.
```
RadoslavGeorgiev, line 550: To be consistent with the rest of this method, let's move this `add_action()` call to an `init_hooks` method within the service, and call that method here instead.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -12,16 +12,22 @@
...
+use WCPay\Payment_Methods\Sepa_Payment_Method;
 
 /**
  * Improve webhook reliability by fetching failed events from the server,
  * then process them with ActionScheduler
  */
 class WC_Payments_Webhook_Reliability_Service {
+	const POST_TYPE                          = 'wcpay-webhook';
+	const TYPE_TAXONOMY                      = 'wcpay-webhook-type';
+	const LIVEMODE_TAXONOMY                  = 'wcpay-webhook-livemode';
```
RadoslavGeorgiev: We have many types of webhooks, which warranted a taxonomy for them. Do you think something with two options deserves a whole taxonomy? Keep in mind that `get_posts()` will need to perform an additional `INNER JOIN` with a potentially huge table.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -12,16 +12,22 @@
...
 	const CONTINUOUS_FETCH_FLAG_EVENTS_LIST  = 'has_more';
 	const CONTINUOUS_FETCH_FLAG_ACCOUNT_DATA = 'has_more_failed_events';
 	const WEBHOOK_FETCH_EVENTS_ACTION        = 'wcpay_webhook_fetch_events';
 	const WEBHOOK_PROCESS_EVENT_ACTION       = 'wcpay_webhook_process_event';
+	const WEBHOOK_PROCESS_EVENTS_ACTION      = 'wcpay_webhook_process_events';
+	const PAYMENT_INTENT_SUCCEEDED_EVENT     = 'payment_intent.succeeded';
```
RadoslavGeorgiev, line 27: Have you considered another place for this constant? Right now we have the single event type here, but we'll likely have more and more in the future.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -60,9 +66,16 @@ public function __construct(
 		$this->action_scheduler_service   = $action_scheduler_service;
 		$this->webhook_processing_service = $webhook_processing_service;
 
-		add_action( 'woocommerce_payments_account_refreshed', [ $this, 'maybe_schedule_fetch_events' ] );
-		add_action( self::WEBHOOK_FETCH_EVENTS_ACTION, [ $this, 'fetch_events_and_schedule_processing_jobs' ] );
+		add_action( 'woocommerce_payments_account_refreshed', [ $this, 'maybe_schedule_failed_webhook_events' ] );
```
RadoslavGeorgiev: I don't think this change makes sense. We're scheduling a simple fetch from the server here.

Context: To avoid overloading the account details endpoint (and cache), we simply serve a flag, which indicates whether there are events to fetch. We are not scheduling their processing within the linked method.

With that in mind, let's restore its name?

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+			self::LIVEMODE_TAXONOMY,
+			self::POST_TYPE,
+			[
+				'public'       => false,
+				'show_in_rest' => false,
+			]
+		);
+
+		if ( ! as_next_scheduled_action( self::WEBHOOK_PROCESS_EVENTS_ACTION ) ) {
+			as_schedule_recurring_action( time(), 60, self::WEBHOOK_PROCESS_EVENTS_ACTION );
```
RadoslavGeorgiev: Let's use the `MINUTE_IN_SECONDS` constant here.

Also, do you think that one minute is a meaningful interval? On most sites, this will likely mean booting too many PHP processes, which wouldn't really do much. IMO the interval could be a few minutes, if not even an hour here.

Also, maybe we should consider only scheduling the action whenever an event is stored? This way it will never run until it's actually needed.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+	 * @param array $event The event array.
+	 * @return bool        Whether to delay the processing of the event.
 	 */
-	private function schedule_fetch_events() {
-		$this->action_scheduler_service->schedule_job( time(), self::WEBHOOK_FETCH_EVENTS_ACTION );
-		Logger::info( 'Successfully schedule a job to fetch failed events from the server.' );
+	public function should_delay_event( array $event ) {
+
+		// If the payment method is SEPA, return false immediately.
+		if ( isset( $event['data']['object']['payment_method_details']['type'] ) && Sepa_Payment_Method::PAYMENT_METHOD_STRIPE_ID === $event['data']['object']['payment_method_details']['type'] ) {
```
RadoslavGeorgiev: Shouldn't we check for `data`, `object`, `payment_method_details`, and `type` individually?

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+			'post_date'    => $this->get_event_processing_time_from_event_type( $event['type'] ?? '' ),
+		];
+
+		$post_id = wp_insert_post( $post_arr );
+
+		wp_set_post_terms( $post_id, $event['type'], self::TYPE_TAXONOMY );
+		wp_set_post_terms( $post_id, $livemode, self::LIVEMODE_TAXONOMY );
+
+		if ( is_wp_error( $post_id ) || 0 === $post_id ) {
+			Logger::error( 'Could not store event in CPT. Event data: ' . var_export( $event, true ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_var_export
```
RadoslavGeorgiev: Instead of just logging this, should we maybe force the event to be processed immediately?

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+		// Retry mechanisms might try to store an existing webhook.
+		$args     = [ 'id' => $event['id'] ];
+		$existing = $this->get_events( $args );
+		if ( ! empty( $existing ) ) {
+			return;
+		}
+		if ( isset( $event['livemode'] ) ) {
+			$livemode = $event['livemode'] ? 1 : 0;
+		} else {
+			$livemode = WC_Payments::mode()->is_live() ? 1 : 0;
```
RadoslavGeorgiev: If we decide to use the taxonomy, let's use unique strings here. I imagine a discrepancy between `0`, `false`, `null` and etc.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+		if ( isset( $extra['type'] ) ) {
+			$args['tax_query'][] = [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+				[
+					'taxonomy' => self::TYPE_TAXONOMY,
+					'terms'    => $extra['type'],
+					'field'    => 'name',
+				],
+			];
+
+			$args['tax_query']['relation'] = 'AND';
```
RadoslavGeorgiev: `AND` is and has always been the default relation for tax queries. We can remove this bit.

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -141,74 +159,266 @@ public function process_event( string $event_id ) {
...
+			$livemode = WC_Payments::mode()->is_live() ? 1 : 0;
+		}
+
+		$post_arr = [
+			'post_title'   => $event['id'] . ' (' . $event['type'] . ')',
+			'post_name'    => $event['id'],
+			'post_type'    => self::POST_TYPE,
+			'post_parent'  => $event['data']['object']['metadata']['order_id'] ?? 0,
+			'post_content' => maybe_serialize( $event['data'] ),
+			'post_date'    => $this->get_event_processing_time_from_event_type( $event['type'] ?? '' ),
```
RadoslavGeorgiev: I'm not sure how WP would handle post dates in the future, as they are used for post status and whatnot. Can't we achieve the same query with normal dates, but while changing the query to retrieve events?

includes/class-wc-payments-webhook-reliability-service.php
```diff
@@ -60,9 +66,16 @@ public function __construct(
...
 		$this->webhook_processing_service = $webhook_processing_service;
 
-		add_action( 'woocommerce_payments_account_refreshed', [ $this, 'maybe_schedule_fetch_events' ] );
-		add_action( self::WEBHOOK_FETCH_EVENTS_ACTION, [ $this, 'fetch_events_and_schedule_processing_jobs' ] );
+		add_action( 'woocommerce_payments_account_refreshed', [ $this, 'maybe_schedule_failed_webhook_events' ] );
+		add_action( self::WEBHOOK_FETCH_EVENTS_ACTION, [ $this, 'schedule_processing_failed_webhook_events' ] );
 		add_action( self::WEBHOOK_PROCESS_EVENT_ACTION, [ $this, 'process_event' ] );
+
+		// Hook the scheduled action to the method.
+		add_action( self::WEBHOOK_PROCESS_EVENTS_ACTION, [ $this, 'load_and_process_events' ] );
```
RadoslavGeorgiev: Hooking the event this way is dangerous. All parameters of the hook would be passed to `get_events`, leading to potential side effects.

RadoslavGeorgiev (changes requested):
Thank you for moving this forward @zmaglica! I left a couple of comments, but I want to make sure we're on the same page regarding the general approach before we focus on them.

One big piece that is missing here is checking for stored webhooks at the end of the payment process. Successful or failed, we should load related events immediately, and either delete them or process them. There are two reasons for this:

1. Even though we will be storing all webhooks of a given type, we want to do it in the shortest possible timeframe. High-traffic sites might receive plenty of webhooks, and those should be processed as quickly as possible / immediately to keep the DB lightweight.
2. Simply delaying webhooks will likely open the gate for a flood of duplicate charges because if something fails during checkout, the same order could be paid again before the webhook gets processed. We want to avoid the concurrency of webhooks by delaying them, but delaying them too much could have more negative than positive effects. Even if we remove the emphasis away from duplicate orders, this would simply degrade customers' experiences.

This needs to be added, and tested in his PR, before we start delaying webhooks. For now, we could simply try processing the webhook within the `catch` block of process payment. Once we have states working well (payment process re-engineering), we could plug in the intent in the middle of the process as well.

---

Another significant issue is that currently all stored events will be processed at once. Webhooks are only fetched from the server each couple of hours, and will be stored at once. This could mean hundreds of orders being processed in one go. Keeping in mind that completing an order is an expensive operation, I can all but guarantee that there will be timed-out processes, or ones exceeding the allowed memory limit.

Do you have any plans regarding how to handle this issue?

dmvrtx:
Thanks a lot for the work on this one — there’s clearly been a lot of thoughtful effort here, especially around storing webhook data more reliably and trying to recover successful payments from the failure path. That context is still useful, and the PR helped clarify the shape of the problem.

At this point, though, I think it makes more sense to close this draft and start over with a fresh PR instead of trying to bring it up to date. The branch is quite old, develop has moved on in the duplicate-payment area, and this change touches core webhook and payment-processing behavior deeply enough that resolving conflicts would likely turn into a partial redesign anyway.

Starting fresh should make it easier to:

1. Re-evaluate whether delaying payment_intent.succeeded is still the right remedy for the problem we want to solve today.
2. Build on top of the current process_payment() and duplicate-payment-prevention flow instead of reconciling 2023 assumptions with 2026 code.
3. Keep the new implementation narrower, easier to review, and better covered by current tests.

Closing this PR is not a reflection on the quality of the work here — it’s mainly about the age of the branch and how much the surrounding code has changed since it was opened.

PR #7013, opened 2023-08-16
Add `is_user_connected()` and `get_connected_user_data()` methods to `WC_Payments_Http_Interface`

### LINKED ISSUE #6997: public methods of WC_Payments_Http class being used without being a part of the WC_Payments_Http_Interface
### Description

The default `WC_Payments_Http_Interface` implementation, `WC_Payments_Http`, has public methods that are used but not declared in the interface.

The following methods have been added without having the interface updated:

`::is_user_connected()`
`::get_connected_user_data()`

This makes custom implementations need to wait until fatal runtime errors are created rather than allowing for static analysis to be used to test the incompatibility.

### Acceptance criteria

Any new public methods added to `WC_Payments_Http` should either be added to the interface or the calling methods should check if the `method_exists()` prior to calling.

### Additional context

We use a custom `WC_Payments_Http_Interface` implementation on WordPress.com to avoid external requests.  Making sure the interface is kept up to date will help us catch the compatibility issues when upgrading sooner.

prettyboymp:
Fixes #6997 

Add `is_user_connected()` and `get_connected_user_data()` methods to `WC_Payments_Http_Interface`.  

These methods are used by `WC_Payments::woopay_tracker()->tracks_get_identity()`, but because the interface doesn't require these methods to be implemented, a custom implementation may not have them, causing a fatal error at runtime.

##### How can this code break?

If a custom implementation already exists, this code will break the site immediately as static analysis of the PHP will result in an immediate error rather than one that only occurs during checkout.

An alternative approach would be for the calling code to always check if the `WC_Payments_Http_Interface` implementation has the methods needed.  However, I think it is pretty unlikely that a site outside of WordPress.com would have a custom `WC_Payments_Http_Interface` implementation.

### DIFF
--- a/changelog/6997-public-methods-of-wc_payments_http-class-being-used-without-being-a-part-of-the-wc_payments_http_interface
+++ b/changelog/6997-public-methods-of-wc_payments_http-class-being-used-without-being-a-part-of-the-wc_payments_http_interface
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Add `is_user_connected()` and `get_connected_user_data()` methods to `WC_Payments_Http_Interface`
--- a/includes/wc-payment-api/class-wc-payments-http-interface.php
+++ b/includes/wc-payment-api/class-wc-payments-http-interface.php
@@ -46,6 +46,20 @@ interface WC_Payments_Http_Interface {
 	public function has_connection_owner();
 
 	/**
+	 * Checks if the current user is connected to WordPress.com.
+	 *
+	 * @return bool true if the current user is connected.
+	 */
+	public function is_user_connected();
+
+	/**
+	 * Get the wpcom user data of the current connected user.
+	 *
+	 * @return bool|array An array with the WPCOM user data on success, false otherwise.
+	 */
+	public function get_connected_user_data();
+
+	/**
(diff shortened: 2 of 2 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7013` or branch name `6997-public-methods-of-wc_payments_http-class-being-used-without-being-a-part-of-the-wc_payments_http_interface` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=6997-public-methods-of-wc_payments_http-class-being-used-without-being-a-part-of-the-wc_payments_http_interface) :rocket:

---

Build info: 

- Latest commit: 4e8513d998390a8b858fada59b68d426dd580ccb
- Build time: 2023-08-16 12:57:21 UTC

marcinbot (approved):
Looks good! Thank you for catching this. I think the only implementation for this is the `WC_Payments_Internal_REST_Request` in wpcom

:shipit:

PR #7021, opened 2023-08-17
RPP: Full test coverage for src

RadoslavGeorgiev:

Add the necessary tests to have 100% tests coverage within `src`.

#### Testing instructions

1. Check out this branch.
2. Execute `npm run test:php-coverage`
3. Check http://localhost:8082/php-test-coverage/index.html and make sure that `src` has 100% coverage:

image

I wanted to have full coverage here, there might be another PR for a CI check.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7021` or branch name `rpp/src-100-percent-coverage` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/src-100-percent-coverage) :rocket:

---

Build info: 

- Latest commit: 18112398a88007ff1f635fd1ec5a88b4025a12aa
- Build time: 2023-08-21 21:24:51 UTC

marcinbot (approved):
Do you think it might be worth setting a minimum coverage for this directory?

Probably not a 100%, but maybe something like 90%

Otherwise, looks good, :shipit:

RadoslavGeorgiev:
> Do you think it might be worth setting a minimum coverage for this directory?

Definitely, which is why I mentioned that "_I wanted to have full coverage here, there might be another PR for a CI check_" in the description. Right now it seems like there are two very separate scripts that check code coverage:

- `npm run test:php-coverage`, which generates the `php-test-coverage` directory locally, and has been adjusted to show the `src` directory as well.
- The `.github/workflows/coverage.yml` workflow and `bin/run-ci-tests-check-coverage.bash` script, which is where we require the 60% for `includes`. This one would need some additional work, as it might require a separate `phpunit.xml`.

So not impossible, but a bit of an additional scope. I think I'll open another issue, and handle it further in the project, when we have more dedicated devs for the project. In the interim, manually monitoring coverage is also an option.

> Probably not a 100%, but maybe something like 90%

There might have been some discussions regarding why not 100%, but I don't remember them. Any particular reason not to at least start with 100%?

htdat (approved):

htdat:
> I think I'll open another issue, and handle it further in the project, when we have more dedicated devs for the project.

Furthermore, agreed with this.

PR #7026, opened 2023-08-17
Stripe Link and SEPA Debit tokens always stored to correct payment gateway

### LINKED ISSUE #6261: Stripe Link is Incompatible with Subscription Renewals

FangedParakeet:
Fixes #6261

# Stripe Link and SEPA Debit payment token trauma terminated at last

## Stripe Link

When the split UPE is enabled--not legacy UPE and not deferred intent UPE--is enabled, Stripe Link saved payment tokens are bestowed with an incorrect gateway ID. [Here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L83) we can see that they will be stored with the gateway ID of `woocommerce_payments_link`. This is incorrect, as they should always be associated with the card gateway and have a gateway ID of `woocommerce_payments`. 

Peculiarly if you save a Stripe Link payment method with the split UPE enabled, it will appear on the My accounts dashboard (*My account > Payment methods*), but not at checkout. This is because when payment methods are [recalled by the `woocommerce_get_customer_payment_tokens` function](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L139), `is_valid_payment_method_type_for_gateway` will always return `false` in [this conditional](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L196C72-L196C112), since the gateway ID indicated [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L24) is incorrect. However, when this function is called on the My accounts Payment methods page, `$gateway_id` is `null`, so the check is ignored. 

Consequently, Stripe Link tokens are not presented at checkout, but are present on the My accounts dashboard (and are otherwise fully functional in renewals).

## SEPA Debit

When the deferred intent UPE is enabled--and split UPE is disabled--SEPA Debit payment methods can be saved, but the tokens will appear in the CC gateway instead of the SEPA Debit gateway. 

This is because the logic [here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L74) is incorrect, since this conditional does not include deferred intent UPE. So when the deferred intent UPE is enabled, SEPA Debit tokens will erroneously be saved with the CC gateway's ID.

In short, this PR addresses all of the above, sets the score straight, and ensures that gateway IDs and validation for Stripe Link and SEPA Debit tokens is now correct for whichever feature flag is enabled. 

## Stripe Link

1. Enable *only* the split UPE, remove SEPA Debit from your store (so that the legacy UPE is not used as a fallback), and make sure you are on the `develop` branch first.
2. Save a Link payment token, by either opting to do so during a regular checkout or purchasing a subscription product using Stripe Link.
3. Visit a new checkout and notice that your saved Link token is not visible.
4. Visit the My accounts dashboard (*My account > Payment methods*) and notice that your Link token *is* visible.
5. Switch to this PR branch, repeat the above steps, and you should find that your Link token is present on any checkout page (shortcode/blocks) and the My accounts dashboard, as expected.
6. Purchase a subscription product using Link. You should be able to change the payment method from Link and process a renewal payment on the subscription using Link.
7. Lucky number seven.

## SEPA Debit

1. Enable *only* the deferred intent UPE, add SEPA Debit to your store, and make sure you are on the `develop`  branch first. Edit [the `WC_Payments_Features::is_upe_split_eligible()` function](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-features.php#L87) to always return `true`.
2. Save a SEPA Debit payment token, by either opting to do so during a regular checkout or purchasing a subscription product using Stripe Link.
3. Visit any new checkout and notice that your SEPA Debit token appears under the CC gateway, instead of the SEPA Debit gateway.
4. Switch to this PR branch, repeat the above steps, and you should find your SEPA Debit token stored under the correct gateway this time.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-token-service.php
+++ b/includes/class-wc-payments-token-service.php
@@ -21,7 +21,7 @@ class WC_Payments_Token_Service {
 	const REUSABLE_GATEWAYS_BY_PAYMENT_METHOD = [
 		Payment_Method::CARD => WC_Payment_Gateway_WCPay::GATEWAY_ID,
 		Payment_Method::SEPA => WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA,
-		Payment_Method::LINK => WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::LINK,
+		Payment_Method::LINK => WC_Payment_Gateway_WCPay::GATEWAY_ID,
 	];
 
 	/**
@@ -71,7 +71,7 @@ class WC_Payments_Token_Service {
 		switch ( $payment_method['type'] ) {
 			case Payment_Method::SEPA:
 				$token      = new WC_Payment_Token_WCPay_SEPA();
-				$gateway_id = WC_Payments_Features::is_upe_split_enabled() ?
+				$gateway_id = WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ?
 					WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA :
 					CC_Payment_Gateway::GATEWAY_ID;
 				$token->set_gateway_id( $gateway_id );
@@ -79,9 +79,7 @@ class WC_Payments_Token_Service {
 				break;
 			case Payment_Method::LINK:
 				$token      = new WC_Payment_Token_WCPay_Link();
-				$gateway_id = WC_Payments_Features::is_upe_split_enabled() ?
-					WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::LINK :
-					CC_Payment_Gateway::GATEWAY_ID;
+				$gateway_id = CC_Payment_Gateway::GATEWAY_ID;
 				$token->set_gateway_id( $gateway_id );
 				$token->set_email( $payment_method[ Payment_Method::LINK ]['email'] );
 				break;
@@ -121,7 +119,7 @@ class WC_Payments_Token_Service {
 	 * @return bool                       True, if payment method type matches gateway, false if otherwise.
 	 */
 	public function is_valid_payment_method_type_for_gateway( $payment_method_type, $gateway_id ) {
-		if ( WC_Payments_Features::is_upe_split_enabled() ) {
+		if ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 			return self::REUSABLE_GATEWAYS_BY_PAYMENT_METHOD[ $payment_method_type ] === $gateway_id;
 		} else {
 			return WC_Payments::get_gateway()->id === $gateway_id;
--- a/changelog/fix-get-stripe-link-payment-tokens-split-upe
+++ b/changelog/fix-get-stripe-link-payment-tokens-split-upe
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Ensures that Stripe Link and SEPA Debit saved payment tokens are stored and validated with correct gateway IDs for relevant feature flags enabled.
--- a/tests/unit/test-class-wc-payments-token-service.php
+++ b/tests/unit/test-class-wc-payments-token-service.php
@@ -111,6 +111,54 @@ class WC_Payments_Token_Service_Test extends WCPAY_UnitTestCase {
 	}
 
 	/**
+	 * Test add SEPA token to user with split UPE.
+	 */
+	public function test_add_token_to_user_for_sepa_split_upe() {
+		update_option( WC_Payments_Features::UPE_SPLIT_FLAG_NAME, '1' );
+		$mock_payment_method = [
+			'id'         => 'pm_mock',
+			'sepa_debit' => [
+				'last4' => '3000',
+			],
+			'type'       => Payment_Method::SEPA,
+		];
+
+		$token = $this->token_service->add_token_to_user( $mock_payment_method, wp_get_current_user() );
+
+		$this->assertEquals( 'woocommerce_payments_sepa_debit', $token->get_gateway_id() );
+		$this->assertEquals( 1, $token->get_user_id() );
+		$this->assertEquals( 'pm_mock', $token->get_token() );
+		$this->assertEquals( '3000', $token->get_last4() );
+		$this->assertInstanceOf( WC_Payment_Token_WCPay_SEPA::class, $token );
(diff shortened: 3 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7026` or branch name `fix/get-stripe-link-payment-tokens-split-upe` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/get-stripe-link-payment-tokens-split-upe) :rocket:

---

Build info: 

- Latest commit: bbf521320333c86d6fc74bbe84254cde0be00afc
- Build time: 2023-08-21 14:44:57 UTC

includes/class-wc-payments-token-service.php
```diff
@@ -71,17 +71,15 @@ public function add_token_to_user( $payment_method, $user ) {
...
 					WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::SEPA :
 					CC_Payment_Gateway::GATEWAY_ID;
 				$token->set_gateway_id( $gateway_id );
 				$token->set_last4( $payment_method[ Payment_Method::SEPA ]['last4'] );
 				break;
 			case Payment_Method::LINK:
 				$token      = new WC_Payment_Token_WCPay_Link();
-				$gateway_id = WC_Payments_Features::is_upe_split_enabled() ?
-					WC_Payment_Gateway_WCPay::GATEWAY_ID . '_' . Payment_Method::LINK :
-					CC_Payment_Gateway::GATEWAY_ID;
```
frosso, line 84: (non-blocking) Does this change have any repercussions on Stripe Link tokens that have already been saved?

FangedParakeet: Good question!

For payment tokens at checkout, these will automatically be replaced, since whenever we list out saved payment tokens, we collect a list from Stripe, compare this list to what we have stored, and [create new tokens for anything that is missing](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-token-service.php#L197). This means that we create new valid tokens to take the place of the original invalid tokens. 

You can confirm by this by adding a saved payment method before Stripe on `develop`, where it will be hidden: upon switching to this branch it will magically reappear.

But what does this mean for existing saved payment tokens used with subscriptions? I tested this out by paying for a subscription product using Link and then switching to this branch and trying to renew the subscription and a few peculiar things happened. 

As I mentioned in [this comment](https://github.com/Automattic/woocommerce-payments/issues/6261#issuecomment-1682899459), the renewal _did_ actually process successfully. A new renewal order was created with accompanying payment intent--all of which were successful. I think this might be because the payment token still has a valid, reusable payment method attached to it, so WC Subscriptions somehow manages to make it work. 

However, there was an error notice at the top of the subscription order after I processed the renewal successfully, stating that the payment method on the subscription was missing. Following the prompt, I was this time able to update the saved payment method on the subscription to the proper saved Stripe token using the dropdown in the billing information, since this token now appears under the correct gateway. 

Updating subscription payment method dropdown

Compare the image above to the second screenshot in Chris's original issue description over [here](https://github.com/Automattic/woocommerce-payments/issues/6261#issue-1702849296), where the store owner is unable to update the saved payment method to the desired method.

TL;DR: subscription renewals appear to process successfully with a non-breaking error that prompts store owner to manually fix the subscription, thanks to the fix in this PR. It's not really working by intention...but is working without breaking and suggests an intuitive workaround to permanently resolve itself. 🤷 

If this is deemed acceptable, maybe we don't need to worry about manually migrating payment tokens--SEPA tokens included. 🤔

frosso: Thank you for the investigation and explanation! The workaround following the error message sounds acceptable as well! 👍

gpressutto5 (commented):
The saved link is displayed after switching to this branch, but I can't checkout using it.

image

image

```
2023-08-21T17:59:17+00:00 ERROR The PaymentMethod provided (link) is not allowed for this PaymentIntent. Please attach a PaymentMethod of one of the following types: card. Alternatively, update the allowed payment_method_types for this PaymentIntent to include "link". (invalid_request_error)
2023-08-21T17:59:17+00:00 ERROR Level3 data error: 
Error: The PaymentMethod provided (link) is not allowed for this PaymentIntent. Please attach a PaymentMethod of one of the following types: card. Alternatively, update the allowed payment_method_types for this PaymentIntent to include "link".
Level 3 data sent: 
Array
(
    [merchant_reference] => 473
    [customer_reference] => 473
    [shipping_amount] => 0
    [line_items] => Array
        (
            [0] => stdClass Object
                (
                    [product_code] => 240
                    [product_description] => Subscription
                    [unit_cost] => 1000
                    [quantity] => 1
                    [tax_amount] => 0
                    [discount_amount] => 0
                )

        )

    [shipping_address_zip] => 94110
    [shipping_from_zip] => 94110
)
```

gpressutto5 (approved):
As discussed in slack, the error above was related to a missing step of enabling deferred intent after switching to this branch. Everything works and the code looks great!

PR #7028, opened 2023-08-18
Add `is_test_mode` to every JS track event

ismaeldcom:
Add `is_test_mode` to every JS track event, updating `recordEvent` method to include the property, getting its value from `wcpaySettings.testMode`.

This will allow us to accurately measure live accounts on Tracks funnels.

#### Testing instructions
Go to any WooPayments page and check that the registered `*_wcpay_*` events have `is_test_mode` property defined.

>[!Note]
> Remember to enable `Allow usage of WooCommerce to be Tracked` in **WooCommerce > Settings > Advanced > WooCommerce.com**.

To check that Tracks events are properly registered, you can use one of the following methods:
- Run `localStorage.setItem( 'debug', 'wc-admin:tracks' );` on your site browser console. And look for `verbose` logs.
- Install [Tracks Vigilante](https://github.com/Automattic/tracks-chrome-extension) browser extension and use its log of events.
- Go to [Tracks Live view](https://mc.a8c.com/tracks/live/), filter by any of the allowed methods, and wait patiently (~10min) for events to be registered there.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7028` or branch name `add/track-events-test-mode` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/track-events-test-mode) :rocket:

---

Build info: 

- Latest commit: cea3128ad6672da2aace6df1ca04feb8774670eb
- Build time: 2023-08-24 07:14:53 UTC

client/tracks/index.js
```diff
@@ -21,6 +21,9 @@ function isEnabled() {
  * @param {Object} [eventProperties] Event properties (optional).
  */
 function recordEvent( eventName, eventProperties ) {
+	// Add `is_test_mode` property to every event.
+	eventProperties.is_test_mode = wcpaySettings?.testMode;
```
dmallory42: Would there be any cases where `wcpaySettings` isn't properly populated? Just wondering if we should have a default value of `false` for that case.

dmallory42 (approved):
LGTM! Tested and works as expected. I had one question but I'm pre-approving!

client/tracks/index.js
ismaeldcom: Yes, it could be not populated, that's why I added the optional chaining operator. And I thought about falling back to false, but I think it's better to just keep it as undefined, to make it clear that we don't know it at that time. WDYT?

dmallory42: Ah, of course! I forget that the chaining operator will evaluate to false.

Yes, I think what you said makes sense - setting it to `undefined` if we don't know seems the best option. I guess we'd need to check if passing undefined causes any error with the tracking (I'm assuming it will just accept the value, but I'm also paranoid 😂).

Happy to go ahead with whichever you decide!

ismaeldcom: `undefined` should end being an empty value on Tracks, so when we build a funnel we will be able to choose between "empty", true, or false.

The _good_ thing about `undefined` is that it works as a way to clear/delete definitions. If you run `JSON.stringify` against `{x: 1, y: undefined}` it will evaluate to `{"x": 1}` 🙂

dmallory42: Perfect! That sounds like the way to go then. 🙂

PR #7035, opened 2023-08-21
RPP: Factor flags

### LINKED ISSUE #6679: Payment Process: Define factor flags
### Description

Make sure that all current and future (in terms of developing the payment process) factors are included as features/flags within the plugin.

Since those are not feature flags, let’s only include them within the payment service, or an adjacent class. 

### Acceptance criteria

All possible factors are listed next to each other within the client.

### Testing instructions

Those are just flags, no testing instructions needed.

### Dev notes

All flags can be found in the _Routing and factors_ section of the project thread.

### Additional context

Project: paJDYF-9hL-p2

### LINKED ISSUE #6680: Payment Process: Detect Factors
### Description

Work on detecting each of the factors, defined in https://github.com/Automattic/woocommerce-payments/issues/6679.

Generally, this requires a single and simple check per factor. Alternatively, we could just use the existing payment information object, but we should avoid it. 

The new method (ex. `is_enabled()`) should replace `WCPAY_NEW_PROCESS`.

At this point, devs should be able to manually force any flag, leading to entering the payment process upon certain conditions.

Example:

```php
<?php
if (
    $payment_service->should_use_new_payment_process(
        $order,
        [
            'USE_SAVED_PM' => isset( $_POST['wc-' . WC_Payment_Gateway_WCPay::GATEWAY_ID . '-payment-token'] ),
            'SAVE_PM'      => $payment_information->should_save_payment_method_to_store(),
            // ... all other factors we can detect in the current context
            // if not set here, default to false.
        ]
    )
) {
    return $payment_service->process_payment( $order );
}
```

As you can see above, those factors should be provided by the service caller (ex. gateway's `process_payment`), and should be added all at once.

### Acceptance criteria

All factors are checked, and provided to a temporary method of the service.

### Testing instructions

TBD. Generally, this should allow defined constants to enable entering the new payment process. By default, the old one should always be used.

### Additional context

Project: paJDYF-9hL-p2

RadoslavGeorgiev:
Fixes #6679. Fixes #6680.

## Related PRs

- https://github.com/Automattic/woocommerce-payments-server/pull/3828
- https://github.com/Automattic/woocommerce-payments-dev-tools/pull/98

This PR allows the client to use the factor flags, provided by the server (see the related PR) to conditionally route customers into the new payment process.

#### New request class

There's a simple `Get_Payment_Process_Factors` request class, which loads the `payment-process-config/factors` route. It's not a top-level route because the payment process might require further configuration in the future.

#### Factors definition (https://github.com/Automattic/woocommerce-payments/issues/6679)

The new `WCPay\Internal\Payment\Factor` class contains constants with all possible payment factors, as well as their descriptions. Compared to the P2, there's an additional `NEW_PAYMENT_PROCESS` factor, which is used to indicate whether even the basic payment process can be entered at all, even if all other detected factors are `false`.

#### Router

The new `WCPay\Internal\Payment\Router` class:

- Is used to load and cache factors (`get_cached_factors()`) through the `Database_Cache` class.
- Allow factors to be modified by dev tools through the `wcpay_new_payment_process_enabled_factors` filter.
- Contains the `should_use_new_payment_process()` method, which compares the provided factors by the gateway to the ones in configuration. If any of the current (gateway-provided) factors is not present in the list of allowed factors, it will return `false`.

Ideally, all present factors would also be checked within this class; however, some checks require functionality that's only present within the gateway class.

#### Gateway changes

Things got slightly changed within the base gateway class. Instead of checking for the `WCPAY_NEW_PROCESS` constant within `process_payment()`, all routing and checks are now extracted into the `new_process_payment` method, which would either return null (proceeding with the legacy process), or return the payment response.

All factors are also loaded within `new_process_payment`.  Some of the checks there do not look too beautiful, but the `Payment_Information` instance is not available that early in `process_payment`, and cannot be passed to the method.

An important note here is that routing (factors) and the actual new payment process are intentionally separate. The `process_payment` method of the payment service should not be aware of any of that.

### Unrelated changes

The PR contains some changes that are not immediately related to its main purpose:

1. There is a new `AJAX_ALLOWED_KEYS` constant in the `Database_Cache` class, which allows the cache to be refreshed during AJAX calls for specific keys. That's needed because the payment process generally happens in AJAX calls.
2. There's a small change to the DI container tests. There were some leaks from `ContainerTest` that are now prevented.

## Testing instructions

1. Check out this PR. Tinker with the payment process, and make sure that because of the current version, the new payment process is never entered.
2. Use dev tools to enable certain (or all) factors for the new payment process.
3. Try checking out, and seeing the expected new payment process exception.
4. Repeat 2 and 3 with random combinations of factors.

-------------------

- [x] Covered with tests (100% coverage 😎 )


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7035` or branch name `rpp/6679-factor-flags` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/6679-factor-flags) :rocket:

---

Build info: 

- Latest commit: cbcd40978ca62ec6b03c523e5746facd055afc01
- Build time: 2023-09-01 05:49:39 UTC

src/Internal/Payment/Factor.php
```diff
@@ -0,0 +1,131 @@
...
+
+namespace WCPay\Internal\Payment;
+
+/**
+ * A class for payment factors.
+ *
+ * These factors will be used to determine whether to enter the new
+ * payment process until it is fully developed and this class gets removed.
+ */
+class Factor {
```
htdat: May consider inheriting `WCPay\Constants\Base_Constant`

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @param bool $force_refresh Forces data to be fetched from the server, rather than using the cache.
+	 * @return array Factors, or an empty array.
+	 */
+	private function get_cached_factors( bool $force_refresh = false ) {
+		$factors = $this->database_cache->get_or_add(
+			Database_Cache::PAYMENT_PROCESS_FACTORS_KEY,
+			function () {
+				try {
+					$request  = Get_Payment_Process_Factors::create();
+					$response = $request->send( 'wcpay_Get_Payment_Process_Factors' );
```
htdat: The hook name should be all lower-cases?

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+
+		return true;
+	}
+
+	/**
+	 * Returns all factors, which can be handled by the new payment process.
+	 * If any factor, not enabled in the returned array, is present while paying,
+	 * the payment should fall back to the legacy process.
```
htdat: These line 68 and 69 should not be here?

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+		return $factors;
+	}
+
+	/**
+	 * Checks if cached data is valid.
+	 *
+	 * @param mixed $cache The cached data.
+	 * @return bool
+	 */
+	public function is_valid_cache( $cache ): bool {
```
htdat, line 86: Should this method be private?

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+		return is_array( $cache ) && isset( $cache[ Factor::NEW_PAYMENT_PROCESS ] );
+	}
+
+	/**
+	 * Refetches payment factors and returns the fresh data.
+	 *
+	 * @return array Either the new arrayh of factors, or false if unavailable.
+	 */
+	public function refresh_factors() {
```
htdat:
```suggestion
	public function get_refreshed_factors() {
```

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @return bool
+	 */
+	public function is_valid_cache( $cache ): bool {
+		return is_array( $cache ) && isset( $cache[ Factor::NEW_PAYMENT_PROCESS ] );
+	}
+
+	/**
+	 * Refetches payment factors and returns the fresh data.
+	 *
+	 * @return array Either the new arrayh of factors, or false if unavailable.
```
htdat: This can not be false? Though it can be an empty array.

`arrayh` is mistypo too.

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	/**
+	 * Checks whether a given payment should use the new payment process.
+	 *
+	 * @param array $factors Factors, describing the type and conditions of the payment.
+	 * @return bool
+	 */
+	public function should_use_new_payment_process( array $factors ): bool {
+		$allowed = $this->get_enabled_factors();
```
htdat: I think `allowed` and `enabled` are mixing here. Should we use only one term?

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @return array Either the new arrayh of factors, or false if unavailable.
+	 */
+	public function refresh_factors() {
+		return $this->get_cached_factors( true );
+	}
+
+	/**
+	 * Gets and chaches all factors, which can be handled by the new payment process.
+	 * If any factor, not enabled in the returned array, is present while paying,
+	 * the payment should fall back to the legacy process.
```
htdat: Similarly, these lines 100 and 101 should not be here? They do not describe anything meaningful for this method.

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+	 *
+	 * @return array
+	 */
+	public function get_enabled_factors() {
+		$factors = $this->get_cached_factors() ?? [];
+		$factors = apply_filters( 'wcpay_new_payment_process_enabled_factors', $factors );
+		return $factors;
+	}
```
htdat: With the `get_enabaled_factors`, I thought it would loop through all factors fetching from Database_Cache, and filter out all enabled ones, i.e. ones with `true`? But it's NOT doing that here.

The `wcpay_new_payment_process_enabled_factors` filter helps WooPayments Dev Tools force enabling desired factors.

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,125 @@
...
+		$this->database_cache = $database_cache;
+	}
+
+	/**
+	 * Checks whether a given payment should use the new payment process.
+	 *
+	 * @param array $factors Factors, describing the type and conditions of the payment.
+	 * @return bool
+	 */
+	public function should_use_new_payment_process( array $factors ): bool {
```
htdat, line 46: The shape of parameter array $factors is `array <string, bool>`, something like this: 
```php
[
	Factor::NO_PAYMENT          => false,
	Factor::USE_SAVED_PM        => true,
]
```

However, I think we can simplify that by just providing required enabling factors as array values, i.e. `array<string>` like: 

```php
[
	Factor::NO_PAYMENT,
	Factor::USE_SAVED_PM,
]
```

includes/class-database-cache.php
```diff
@@ -13,11 +13,22 @@
...
+
+	/**
+	 * Refresh during AJAX calls is avoided, but white-listing
+	 * a key here will allow the refresh to happen.
+	 *
+	 * @var string[]
+	 */
+	const AJAX_ALLOWED_KEYS = [
+		self::PAYMENT_PROCESS_FACTORS_KEY,
+	];
```
htdat, line 31: I saw the explanation of this new constant in the PR description, but I am still not clear about its purpose? 

I assume you want to make it to retrieve factors from the server ASAP in case something wrong. If so, I don't think this trade-off is worth doing and helps much in that case. Instead, may we build up a server webhook to trigger the fetching factors? That's more efficient in managing crisis, IMO.

htdat (commented):
Initial feedback.

src/Internal/Payment/Router.php
marcinbot: I would add to this that this doesn't seem to be used anywhere outside of tests. I think we should remove it. We could add a refreshing mechanism in the future (or in the dev tools) by calling `Database_Cache::delete`

src/Internal/Payment/Router.php
marcinbot: It's called by the `Database_Cache` in `get_cached_factors`. I think it would be a bit cleaner if it was inline, but as it is is fine too.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -713,14 +782,14 @@ public function payment_fields() {
...
 	public function process_payment( $order_id ) {
+		$order = wc_get_order( $order_id );
 
-		if ( defined( 'WCPAY_NEW_PROCESS' ) && true === WCPAY_NEW_PROCESS ) {
-			$new_process = wcpay_get_container()->get( PaymentProcessingService::class );
-			return $new_process->process_payment( $order_id );
+		// If the new payment process was entered, use the result.
+		$new_process_result = $this->new_process_payment( $order );
+		if ( ! is_null( $new_process_result ) ) {
+			return $new_process_result;
```
marcinbot: I think it will be cleaner to have something like this:
```
if ( $this->should_use_new_process( $order ) ) {
    return $this->new_process_payment( $order );
}
```
The `maybe_` functions are a bit anti-patterny and not pure.

src/Internal/Payment/Router.php
marcinbot: I think `enabled` should be replaced with `present` because that's what it is - it's a factor that's part of the ongoing payment process.

Otherwise agree: enabled and allowed are synonymous.

RadoslavGeorgiev: I'll switch to `allowed`, I think it's better for this context.

> I think enabled should be replaced with present because that's what it is - it's a factor that's part of the ongoing payment process.

Actually, this particular line is exactly for the allowed factors. `$factors` (the parameter) is about the present factors.

includes/class-database-cache.php
marcinbot: Nothing to do with webhooks. The code below in `Database_Cache` disabled re-fetching if the current request is AJAX. This worked for the Accounts and limited the number of requests, but the payments are all processed as AJAX.

src/Internal/Payment/Router.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/6e571121da0fdc31574be617115d575ee672b996. It used to be lowercase, but I couldn't settle on some names, and I overused mass-selection in VSCode.

src/Internal/Payment/Router.php
RadoslavGeorgiev: Well, the comments were not wrong per se, but too much. Removed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/6185a2953d46421f4b5b36634661a3aaf108c5dc

src/Internal/Payment/Router.php
RadoslavGeorgiev: The method is now removed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/b1e46e1d244d1e0a4f744f56c4d7bbf82a7c3e49, based on this comment: https://github.com/Automattic/woocommerce-payments/pull/7035#discussion_r1304167844

src/Internal/Payment/Router.php
RadoslavGeorgiev: I agree with @marcinbot. Removed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/b1e46e1d244d1e0a4f744f56c4d7bbf82a7c3e49

src/Internal/Payment/Router.php
RadoslavGeorgiev: Switched to `allowed` in this commit: https://github.com/Automattic/woocommerce-payments/pull/7035/commits/d3d5f2077d3d020c2e2ffc4b8fa5e7d109b95a27

src/Internal/Payment/Router.php
RadoslavGeorgiev: Removed, similarly to the other comment.

src/Internal/Payment/Router.php
RadoslavGeorgiev: Not sure how I missed this. I switched to `string[]` in:

- https://github.com/Automattic/woocommerce-payments/pull/7035/commits/e9743599ca247ee790214097e43164640d271cc4 for `get_allowed_factors()`
- https://github.com/Automattic/woocommerce-payments/pull/7035/commits/e8fbfd3c05824aa4d85754d296ce2428a03b8376 for the gateway

Now the server does return the full `string => bool` combinations, but they are not used past `get_cached_factors()`.

includes/class-database-cache.php
RadoslavGeorgiev: Yeah, this was not about doing it ASAP. There is still a 2-hour TTL for the cache, but without this change, it would never get populated. Marking this as resolved.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Good catch, I'm not certain why I did it this way in the first place. Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/ca8c12d214bb68167d7c58cb8b0ddedb553896cc, now `should_use_new_process` and `new_process_payment` are separate methods.

src/Internal/Payment/Router.php
RadoslavGeorgiev: See https://github.com/Automattic/woocommerce-payments/pull/7035#discussion_r1304400965. Now it's just the _allowed_ factors that are returned.

src/Internal/Payment/Router.php
RadoslavGeorgiev: Yes, I wanted a testable method, but because it is a callback, it needed to be public.

src/Internal/Payment/Factor.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/8c6af100520b8d0c90fde15b2e34e33c41f50a69, ~however I don't really like it, to be honest:~

1. ~Requires suppressing Psalm `throws` tags in a few places, just because of trying to use `Factor::SOME_FACTOR()`.~
2. ~Strict comparison is impossible with `in_array()`.~
3. ~I'm still accessing constants directly in tests, to avoid a ton of `()->get_value()` just to have correct indexes.~

~@marcinbot @htdat please check the commit and tell me whether to keep it. I'm completely alright removing it :)~

NVM, looking at it with fresh eyes, it looks alright.

src/Internal/Payment/Router.php
htdat: > Yes, I wanted a testable method, but because it is a callback, it needed to be public.

@RadoslavGeorgiev - WooPayments has util [`PHPUnit_Utils::call_method`](https://github.com/Automattic/woocommerce-payments/blob/b9d4a92463953a5e494e9800c0642520248363ef/tests/unit/helpers/class-phpunit-utils.php#L23-L23) to overcome this PHPUnit shortcoming.

includes/class-database-cache.php
htdat: >  There is still a 2-hour TTL for the cache, but without this change, it would never get populated.

Make sense, thanks for the explanation.

> Nothing to do with webhooks.

Yeah, I understand the code around here is not about webhooks. Instead, my idea is that in case of something going wrong urgently, we can send a webhook event like `payment_process_factors.updated`, which is similar to `account.updated`, and force update payment factors cache: 

https://github.com/Automattic/woocommerce-payments/blob/8c6af100520b8d0c90fde15b2e34e33c41f50a69/includes/class-wc-payments-webhook-processing-service.php#L170-L171

src/Internal/Payment/Router.php
RadoslavGeorgiev: Testability is a bonus. However, because the method is provided as a callback to `get_or_add`, it needs to be callable, and only public methods are callable.

includes/class-database-cache.php
RadoslavGeorgiev: To be fair, mass-sending webhooks to all sites with a given version will probably not be much quicker than just waiting for the cache to expire.

I'm open to creating another issue, and discussing this there, but I think it's out of the planned scope for this router. FWIW, I'd prefer a universal hook, which accepts DB cache keys, and deletes all of them directly, instead of a specific one for factors.

RadoslavGeorgiev:
The PR is ready for another review round.

src/Internal/Payment/Factor.php
htdat: @RadoslavGeorgiev  

1. Agreed that his is annoying. I am not certain how to avoid this. May remove the `@throw` tag in the `Base_Constant::__callStatic` method, but not ideal anyway. 
2. I looked at `Base_Constant` implementation, and it appears that we would improve it so that each `__callStatic()` would not create a new object. Instead, we cache initialized objects in a static property, and return the same object for the same static call function, which is similar to the singleton pattern. Something like this https://github.com/Automattic/woocommerce-payments/pull/7063 (and more in this commit  https://github.com/Automattic/woocommerce-payments/pull/7063/commits/ae6f3318589d840dc4ed48228a7d70ccaac45a06)
3. IMO, this is acceptable for array keys (indexes).

includes/class-database-cache.php
htdat: yeah, no need to do that here, creating an issue for consideration is enough. Originally, I did not understand the purpose of `AJAX_ALLOWED_KEYS`.

src/Internal/Payment/Factor.php
RadoslavGeorgiev: I like the idea there, especially the fact that we can use `===` to compare instances because there would not be more than one instance of the same constant. I think it's worth proceeding with that PR.

htdat: Moving that one to a new PR https://github.com/Automattic/woocommerce-payments/pull/7064

RadoslavGeorgiev: https://github.com/Automattic/woocommerce-payments/pull/7064 is approved, I'll implement the changes here once it gets merged.

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,119 @@
...
+	 *
+	 * @param Factor[] $factors Factors, describing the type and conditions of the payment.
+	 * @return bool
+	 * @psalm-suppress MissingThrowsDocblock
+	 */
+	public function should_use_new_payment_process( array $factors ): bool {
+		$allowed_factors = $this->get_allowed_factors();
+
+		// This would make sure that the payment process is a factor as well.
+		$factors[] = Factor::NEW_PAYMENT_PROCESS();
```
marcinbot: I think I understand why this is like that (it's a global flag for controlling whether or not the new process is enabled, regardless of other factors, right?). But I think it belongs in `should_use_new_process` in the gateway, together with other factors, and probably with a better comment explaining why it's there.

It could also be specified as the first item in the list.

marcinbot (commented):
Looks good to me, aside from one small note below.

src/Internal/Payment/Router.php
RadoslavGeorgiev: It’s not controlling whether the new process gets entered regardless of the other factors, rather when they are missing. The point is that if no other factor is present, the `foreach` loop would never be entered, and the method would immediately return `true` for the simplest payments.

Even if this method is coded differently, we still need the factor to be controlled by the server, and that’s actually going to be the first factor that we enable, rather than the last (global) one. 

I see how that could be misleading, I’ll update the comment.

marcinbot: That's more or less how I understood it. I still think it should be added to the rest of the factors in `should_use_new_process`, maybe even more so than before :)

RadoslavGeorgiev: My thinking was that this factor must always be present, which is why I added it to this method, to have it centralised and impossible to miss. However, adding it to the gateway too will not hurt. I’ll do so on Monday.

RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/b029106a7bf3160b08a9ccf27b93cd9653bfb2b7

src/Internal/Payment/Router.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @psalm-suppress MissingThrowsDocblock
+	 */
+	public function should_use_new_payment_process( array $factors ): bool {
+		$allowed_factors = $this->get_allowed_factors();
+
+		// This would make sure that the payment process is a factor as well.
+		// phpcs:ignore WordPress.PHP.StrictInArray.FoundNonStrictFalse
+		if ( ! in_array( Factor::NEW_PAYMENT_PROCESS(), $factors, false ) ) {
+			$factors[] = Factor::NEW_PAYMENT_PROCESS();
+		}
```
marcinbot: I think this is redundant - if the caller didn't pass the `NEW_PAYMENT_PROCESS` then it feels like the caller is doing things wrong. And besides, this will be called in only one place and removed altogether in the future.

RadoslavGeorgiev: > this will be called in only one place and removed altogether in the future.

Actually, this routing will be used in a few places before being removed:

- The post-3DS-authentication part of the flow
- Subscription renewal
- I'd like IPP payments to also start sharing the payment process. [Some methods](https://github.com/Automattic/woocommerce-payments/blob/rpp/6679-factor-flags/includes/admin/class-wc-rest-payments-orders-controller.php#L138) are very similar to `process_payment`, and have basically copy-pasted code right now.
- Similarly, Link payments could also be linked there.

With that clarified, I realize that I'm listing specific flows, which have their own factor flag. Their entry points will always have `Factor::SOME_FLAG() => true` hard-coded. I'm removing this condition...

RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/75bfaa28805cee399aea2e1405aa53b881c708a9

marcinbot (commented):
LGTM

But would it make sense to have a client-side flag making extra sure that we don't ship code that accidentally enters the not-implemented new flow? Something like an early return in the Router so the request for flags is not even sent.

We could later remove it or move it somewhere else in the code as we implement the factors - so that a client will never enter an unimplemented flow without the flag even if the server tells it to.

Trying to think of all the scenarios where things could go wrong given that we'll ship this out there.

RadoslavGeorgiev:
> Something like an early return in the Router so the request for flags is not even sent.

Based on the discussion in p1693221168889909-slack-C05997J5PCG, this is now implemented in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/f0ab089f4352701d9852e1706365578abe002fa4 and https://github.com/Automattic/woocommerce-payments/pull/7035/commits/4d437c736172a77d6fe2d62193d64a01f63b2cb2

marcinbot (approved):

tests/unit/src/Internal/Payment/RouterTest.php
```diff
@@ -0,0 +1,274 @@
...
+		$result = $this->sut->get_allowed_factors();
+
+		$this->assertIsArray( $result );
+		$this->assertEquals( $replaced_factors, $result );
+
+		remove_filter( 'wcpay_new_payment_process_enabled_factors', $filter_cb );
+	}
+
+	/**
+	 * Verify that `is_valid_cache` returns false win a non-array.
```
htdat: Nitpick: 
```suggestion
	 * Verify that `is_valid_cache` returns false with a non-array.
```

tests/unit/src/Internal/Payment/RouterTest.php
```diff
@@ -0,0 +1,274 @@
...
+			Factor::SUBSCRIPTION_SIGNUP => false,
+		];
+		$processed_factors = [ Factor::SAVE_PM() ];
+
+		$this->mock_db_cache_factors( $cached_factors, false );
+
+		$result = $this->sut->get_allowed_factors();
+
+		$this->assertIsArray( $result );
+		$this->assertEquals( $processed_factors, $result );
```
htdat:
```suggestion
		$this->assertSame( $processed_factors, $result );
```

tests/unit/src/Internal/Payment/RouterTest.php
```diff
@@ -0,0 +1,274 @@
...
+						return $request_response === $cb();
+					}
+				),
+				[ $this->sut, 'is_valid_cache' ],
+				false
+			)
+			->willReturn( $request_response );
+
+		$result = $this->sut->get_allowed_factors();
+		$this->assertEquals( $processed_factors, $result );
```
htdat:
```suggestion
		$this->assertSame( $processed_factors, $result );
```

tests/unit/src/Internal/Payment/RouterTest.php
```diff
@@ -0,0 +1,274 @@
...
+
+		$filter_cb = function() use ( $replaced_factors ) {
+			return $replaced_factors;
+		};
+		add_filter( 'wcpay_new_payment_process_enabled_factors', $filter_cb );
+
+		$result = $this->sut->get_allowed_factors();
+
+		$this->assertIsArray( $result );
+		$this->assertEquals( $replaced_factors, $result );
```
htdat:
```suggestion
		$this->assertSame( $replaced_factors, $result );
```

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -2338,6 +2355,287 @@ public function test_no_payment_is_processed_for_woopay_preflight_check_request(
...
+		$order = WC_Helper_Order::create_order();
+
+		// Assert: The router is never called.
+		$mock_router->expects( $this->never() )
+			->method( 'should_use_new_payment_process' );
+
+		$this->assertFalse( $this->wcpay_gateway->should_use_new_process( $order ) );
+	}
+
+	public function test_should_use_new_process_returns_null_if_feature_unavailable() {
```
htdat:
```suggestion
	public function test_should_use_new_process_returns_false_if_feature_unavailable() {
```

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -2338,6 +2355,287 @@ public function test_no_payment_is_processed_for_woopay_preflight_check_request(
...
+
+		$result = $this->wcpay_gateway->process_payment( $order->get_id() );
+		$this->assertSame( $mock_response, $result );
+	}
+
+	/**
+	 * Sets up the expectation for a certain factor for the new payment
+	 * process to be either set or unset.
+	 *
+	 * @param string $factor_name Name of the factor.
```
htdat:
```suggestion
	 * @param Factor $factor_name Name of the factor.
```

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -2338,6 +2355,287 @@ public function test_no_payment_is_processed_for_woopay_preflight_check_request(
...
+	 * @param bool   $value       Expected value.
+	 */
+	private function expect_new_payment_process_factor( $factor_name, $value ) {
```
htdat: This took me a while to understand the purpose of tests calling out this private method. That's because it's a bit confusing between two similar method names in two classes: 

- WC_Payment_Gateway_WCPay::should_use_new_process
- Router::should_use_new_payment_process

Hence I suggest this name to describe better what it is going to do: 
 
```suggestion
	private function expect_router_should_use_new_payment_process_called_with_factor( $factor_name, $value ) {
```

htdat (approved):
Thanks for addressing feedback. 

Pre-approve this PR. I am having only nitpick comments.

tests/unit/src/Internal/Payment/RouterTest.php
RadoslavGeorgiev: Thanks, fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/5b1b08b5ad10cb2fe2f7d884fd8274d044b082f5

tests/unit/src/Internal/Payment/RouterTest.php
RadoslavGeorgiev: There are a few cases where because of separate constant instances, I could not use `assertSame` before. Thanks for the suggestion now though, fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/5b1b08b5ad10cb2fe2f7d884fd8274d044b082f5

tests/unit/src/Internal/Payment/RouterTest.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/5b1b08b5ad10cb2fe2f7d884fd8274d044b082f5

tests/unit/src/Internal/Payment/RouterTest.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/5b1b08b5ad10cb2fe2f7d884fd8274d044b082f5

tests/unit/test-class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/2a1d19f9e8a10dcb97fbf0625d10fe574c6e5204

tests/unit/test-class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7035/commits/2a1d19f9e8a10dcb97fbf0625d10fe574c6e5204

tests/unit/test-class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: I understand the confusion, so I renamed the method. However, `expect_router_should_use_new_payment_process_called_with_factor` seemed needlessly long, especially considering the router has a single significant public method, which is why I switched to `expect_router_factor`. Hope it helps :)

PR #7046, opened 2023-08-23
Add typescript guidelines

reykjalin:

Add in-repo documentation for TS development guidelines for WooPayments.

#### Testing instructions

N/A

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7046` or branch name `add/typescript-guidelines` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/typescript-guidelines) :rocket:

---

Build info: 

- Latest commit: 951e319d5e9de7cd1669c9913717ddd3b4d24169
- Build time: 2023-08-24 20:42:44 UTC

ismaeldcom (approved):
Thanks for taking care of the PR! 🚀 

I think the guidelines are in good shape and ready to be merged, we can all keep adding new entries as we do TS migrations.

jbordonado (approved):

docs/typescript/interface-vs-type.md
```diff
@@ -0,0 +1,39 @@
...
+The biggest reason to prefer `interface` is that they're more flexible.
+They can be extended to add more properties when needed whereas types cannot.
+The TypeScript Handbook itself [recommends](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces) using `interface` unless there's a specific feature you need from `type`, in which case the TypeScript compiler will likely tell you about it:
+
+> For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type.
+
+
+## What might be a case where I need to use `type` instead of `interface`?
+
+The one useful case for this is renaming a primitive type such as `string` or `number`.
```
sirreal:
```suggestion
The one useful case for this is aliasing a primitive type such as `string` or `number`.
```

docs/typescript/react-components.md
```diff
@@ -0,0 +1,139 @@
...
+   * The last name to display.
+   */
+  lastName: string;
+}
+```
+
+You will then use `React.PropsWithChildren` like so:
+
+```ts
+export const DisplayName: React.FunctionComponent< React.PropsWithChildren< Props > > = ( { isLoading, numRows, numColumns, children } ) => {
```
sirreal: It seems preferable to create the `Props` interface with `children` instead of adding it later.

````suggestion
interface Props extends React.PropsWithChildren {
  /**
   * Set the loading state of the component. Provide `true` if the component should show a loading state.
   */
  isLoading: boolean;
  
  /**
   * The first name to display.
   */
  firstName: string;
  
  /**
   * The last name to display.
   */
  lastName: string;
}
```

You will then use `React.PropsWithChildren` like so:

```ts
export const DisplayName: React.FunctionComponent< Props > = ( { isLoading, numRows, numColumns, children } ) => {
````

File: docs/typescript/utility-and-manipulation.md [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+### `Pick<Type>` – Picks a set of properties.
+```ts
+type TodoPreview = Pick<Todo, "title" | "completed">;
+// Same as
+interface TodoPreview {
+  title: string;
+  completed: string;
+}
+```
+### `Return<Type>` – Constructs a type consisting of the return type of function `Type`.
```
sirreal: Typo here, the mapped type is `ReturnType`:

```suggestion
### `ReturnType<Type>` – Constructs a type consisting of the return type of function `Type`.
```

docs/typescript/utility-and-manipulation.md
```diff
@@ -0,0 +1,89 @@
...
+
+### [Typeof Type Operator](https://www.typescriptlang.org/docs/handbook/2/typeof-types.html) – Refers to the type of a variable or property.
+```ts
+typeof updateTodo
+// Equals to
+declare function updateTodo(todo: Todo, fieldsToUpdate: Partial<Todo>): {
+    title: string;
+    description: string;
+    completed: boolean;
+};
```
sirreal: This example isn't correct. Here, `typeof` is the JavaScript operator, equal to `"function"` in this case. And it wouldn't be equal to declaring a function.

The example should be reworked, `typeof` is for when you want to access the type of a value.

I'm not sure what example would be better here, probably pick something from the linked documentation page.

docs/typescript/utility-and-manipulation.md
```diff
@@ -0,0 +1,89 @@
...
+// Record key access
+type Title = Todo['title'];
+// Equals to
+type Title = string;
+
+// Array number access
+type TodoList = Todo[];
+type ListItem = TodoList[number];
+// Equals to
+type ListItem = Todo;
```
sirreal: This example reads confusing to me. Maybe it would be better to use `Array<T>` instead of `T[]`. I thought it was a typo. It's correct, but creating an array type below the heading "Array number access" confused me.

```suggestion
type TodoList = Array<Todo>;
type ListItem = TodoList[number];
// Equals to
type ListItem = Todo;
```

sirreal (commented):
I've reviewed this and left some comments and suggestions.

A few things that would be nice to see here that I've found powerful when working with TypeScript are readonly types (`ReadonlyArray<T>`), tuples, and const assertions. I don't see mention of those here but they're common solutions to difficulties folks face.

docs/typescript/utility-and-manipulation.md
ismaeldcom: You're right, the best example for its use case with TS could be actually the ReturnType<Type> one. Where you can do something like `ReturnType<typeof updateTodo>`. _Addressed in 610e3ad_

docs/typescript/utility-and-manipulation.md
ismaeldcom: Not sure what's the confusion here, maybe you're more used to the `Array` syntax instead of the shorter one `[]` 🤔 Anyway, I agree about using more explicit code 🙂

docs/typescript/utility-and-manipulation.md
```diff
@@ -0,0 +1,84 @@
...
+
+### `Partial<Type>` – Makes all properties `optional`
+```ts
+interface Todo {
+  title: string;
+  description: string;
+  completed: boolean;
+}
+
+type TodoUpdate = Partial<Todo>
```
sirreal: Is there autoformatting applied to markdown? I think WordPress flavored `prettier` would include spaces inside the angle brackets. We're not consistent across files here, let's make sure we are.

(This is just one example, there are many in this PR)

```suggestion
type TodoUpdate = Partial< Todo >;
```

docs/typescript/utility-and-manipulation.md
sirreal: I think it was the fact that the comment says "array number access", but the following line creates an array type:

```ts
// Array number access
type TodoList = Todo[];
type ListItem = TodoList[number];
```

Something more verbose would've helped me process:

````md
The type of elements in an array can be accessed via `ArrayType[number]`:

```ts
// Make an array type
type ArrayOfStrings = string[];
// The type of array elements
type ElementOfArrayOfStrings = ArrayOfStrings[number]; // This is equivalent to `string`
```
[…]
````

docs/typescript/utility-and-manipulation.md
ismaeldcom: Good point, I actually have the formatter disabled in VS Code for Markdown because I have prettier enabled, but we don't have anything configured for it in the repo. Will be good to choose a formatting style.

Jinksi: I'm finding that VSCode prettier will use the repo's `.prettierrc` for auto-formatting `.md`.

`singleQuote: true` and `semi: true` rules should be applied to this file in addition to `parenSpacing` to be consistent with the rest of the codebase.

Jinksi (approved):
LGTM 🏆 This is a great starting point for TS/FE guidelines to build upon.

There are some minor formatting issues that could be addressed to keep the code style consistent with the codebase, but not a blocker. (https://github.com/Automattic/woocommerce-payments/pull/7046#discussion_r1303674745)

docs/typescript/utility-and-manipulation.md
reykjalin: I ran prettier on the markdown files in 30ec8452a56a410d0e39f486ecf99c956ea900a4 so the formatting _should_ be consistent now.

sirreal (approved):
This looks like a good starting point, thanks!

We have some TypeScript guidelines written in a private repository that may be helpful. If you'd like, I can follow-up with a pull request suggesting those additions. They're things like:

- Avoid `any`
- Avoid type assertions
- Rely on type inference
- Use readonly
- Use const assertions
- Avoid enums

reykjalin:
> If you'd like, I can follow-up with a pull request suggesting those additions. They're things like:

That'd be great! Thank you @sirreal!

PR #7054, opened 2023-08-23
RPP - Payment request class for loading, sanitizing, and escaping data

### LINKED ISSUE #6684: Payment Process: Request Class
### Description

Implement a request utility class, which is used as a proxy to $_POST: Loading, sanitizing, escaping, etc. We wouldn’t want to dilute the payment class with `// phpcs:ignore.`

### Acceptance criteria

All possible parameters of the current `process_payment()` are covered, except for payment methods.

### Testing instructions

Cover all methods with unit tests. Tests should ensure that data is sanitized, so try intentionally providing invalid values.

### Dev notes

Previously implemented here: https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-85bf8375d34971ad7aa5df10a35771c1af4b8bc8d6952f83592ff17c2731fe2c

### Additional context

Project: paJDYF-9hL-p2

htdat:
Note: RPP = Re-engineering Payment Process.

Fixes #6684

- Create Payment request class for loading, sanitizing, and escaping data form $_POST or params from WP_REST_Request.

#### Testing instructions

* All tests passed.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7054` or branch name `rpp/6684-request-class` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/6684-request-class) :rocket:

---

Build info: 

- Latest commit: 61fef48b7b00e3d468e350abceb6010bb7878b92
- Build time: 2023-08-26 09:11:44 UTC

src/Util/PaymentRequestUtil.php
```diff
@@ -0,0 +1,37 @@
...
+ * Util class for loading, sanitizing, and escaping data from payment requests.
+ */
+class PaymentRequestUtil {
+
+	/**
+	 * Get the fraud prevention token from the request.
+	 *
+	 * @return string|null
+	 */
+	public static function get_fraud_prevention_token(): ?string {
```
RadoslavGeorgiev: Context: a comment from [this doc](https://github.com/Automattic/woocommerce-payments/tree/develop/src#legacy-proxy):

>  Use this for non-`src` classes. `src` classes should only have pure static methods.

Let's not make this class a utility, or keep its methods static. Static methods should be pure, and accessing `$_POST` here is not pure. It might be easy to test this class now, but we'll be making this a bit weirder when testing services, which use this utility because static methods cannot be mocked. We'd be forcing devs to check the static method to understand what `$_POST` parameters to modify/mock.

I also don't like that we need to access/manipulate (for tests) $_POST here. When I think about it, I don't really like [my implementation](https://github.com/Automattic/woocommerce-payments/blob/spike/payment-process-states/includes/payment/class-post-request.php) either. Let's use `$_POST` as a constructor parameter for the class, and then use something like `$this->request['wcpay-fraud-prevention-token']`. WDYT?

htdat: All good points. Going to change the approach now.

htdat:
@RadoslavGeorgiev  - I've changed the approach. It would be great if you can have a look, and see if the new direction looks ok? I am still working on other utils for now.

src/Internal/Payment/Request.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+class Request {
+	/**
+	 * The request array.
+	 *
+	 * @param  array|null $request  Request data, this can be $_POST, or WP_REST_Request::get_params().
+	 */
+	public function __construct( array $request = null ) {
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$this->request = $request ?? $_POST;
```
RadoslavGeorgiev: I love defaulting to `$_POST` here, while allowing the request to be overwritten. Very minor, but let's define the class property.

RadoslavGeorgiev:
It looks great now! Looking forward to seeing the rest of the class :)

htdat:
Ready for the first round of review. I checked $_POST in `payment_process` and `payment_process_for_order` of `WC_Payment_Gateway_WCPay` to create the new Reqest class.

src/Internal/Payment/Request.php
```diff
@@ -0,0 +1,92 @@
...
+			: null;
+	}
+
+	/**
+	 * Check if the request is a WooPay preflight check.
+	 *
+	 * @return bool
+	 */
+	public function is_woopay_preflight_check(): bool {
+		return ! empty( $this->request['is-woopay-preflight-check'] );
```
RadoslavGeorgiev: I think the presence of the key alone is enough for the preflight check. There is an `isset` needed anyway.

src/Internal/Payment/Request.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * Gets intent ID if any.
+	 *
+	 * @return string|null
+	 */
+	public function get_intent_id(): ?string {
+		return isset( $this->request['intent_id'] )
+			? sanitize_text_field( $this->request['intent_id'] )
```
RadoslavGeorgiev: `sanitize_text_field` is used in conjuction with `wp_unslash`. Maybe worth adding here as well?

src/Internal/Payment/Request.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * Gets the ID of the provided payment method.
+	 *
+	 * @return string|null
+	 */
+	public function get_payment_method_id(): ?string {
+		return isset( $this->request['payment_method_id'] )
+			? sanitize_text_field( $this->request['payment_method_id'] )
```
RadoslavGeorgiev: Same `wp_unslash` comment here.

tests/unit/src/Internal/Payment/RequestTest.php
```diff
@@ -0,0 +1,132 @@
+<?php
+/**
+ * Class RequestTest
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Tests\Util;
+
+use WCPay\Internal\Payment\Request as PaymentRequest;
```
RadoslavGeorgiev: I think simply using `Request` here might cause confusion with server requests. Maybe just use `PaymentRequest`?

RadoslavGeorgiev (changes requested):
Great work! just a couple of minor nitpicks on my side.

I love seeing how we're trying to hack validation in tests :)

tests/unit/src/Internal/Payment/RequestTest.php
htdat: Make sense, addressed 1f7b895bb6f95bb44cddcd57ef1c87325a5728d7

src/Internal/Payment/Request.php
htdat: Good point! I used `empty` because this is how the same parameter is inferred in `payment_process`. Addressed here 13e49c3a06e685406cacba965ab3df1ccac42631

src/Internal/Payment/Request.php
htdat: Addressed here 61fef48b7b00e3d468e350abceb6010bb7878b92.

htdat:
@RadoslavGeorgiev - Thanks for all great feedback. I've addressed all of them. Ready for another review round.

RadoslavGeorgiev (approved):
LGTM, thank you for fixing my nitpicks! :shipit:

PR #7005, opened 2023-08-15
Fix empty cart WooPay error on WoA (WordPress.com on Atomic) environments

ricardo:
Context: p1692118875495869-slack-C022WMN88KG

#### Changes proposed in this Pull Request

Before #6331, we had [this condition](https://github.com/Automattic/woocommerce-payments/commit/5f74b01c692567ebda122802b8fcefb99116125b#diff-b92b064836d1d92a6b3e17be46f62e1faeb72b6ac6c4debdb87d56414adfc415L298) which unexpectedly prevented a race condition problem in Atomic environments.

This PR adds a similar, but more focused condition under `is_woopay_enabled`.

#### Testing instructions

1. Since this can only be tested on Atomic environments, upload a build from this branch into https://woopaytestmerchant.wpcomstaging.com.
   a. Switch to this branch and run `npm run build`.
   b. Create a zip file from the generated build under the `release` folder.
   c. Upload the zip into an Atomic WooPay test site.
   d. When prompted, replace the existing plugin with the uploaded one.
2. Test the WooPay checkout. Ensure everything works as expected.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #6331, opened 2023-05-19
Fix Multi-Currency switcher support in WooPay

ricardo:
Context: p1684449926045609/1684445728.311959-slack-C022WMN88KG

#### Changes proposed in this Pull Request

* Remove the requirement for `REST_REQUEST` constant in Store API remote requests.
    * Note: Oddly, not all Store API REST requests have that constant set to `true`.
* Set WooPay session handler directly in the `woocommerce_session_handler` hook.
    * Similar to the `REST_REQUEST` constant, some store API requests didn't call the `rest_request_before_callbacks` hook, and this was preventing the Multi-currency feature from retrieving the currency from an existing guest session.

#### Testing instructions

1. Ensure WooPay is eligible and enabled.
2. Enable Multi-Currency + the currency switcher from **WC > Settings > Multi-Currency**.
3. Enable multiple currencies, for example: USD, JPY, EUR.
4. As a guest shopper use the currency switcher to change the currency to JPY.
5. Navigate to the checkout page and enter a valid WooPay email + OTP.
6. From the WooPay checkout, ensure the currency is displayed correctly.
7. In a separate tab, go to the merchant store and change the currency to, "USD", for example.
8. Refresh the WooPay checkout tab and ensure the currency, totals and currency symbol are set to USD.
9. Checkout with WooPay.
10. Ensure the charge is correctly processed for the chosen currency.
11. Logout from WooPay and repeat steps 4-10 as a logged in shopper.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7055, opened 2023-08-23
Session compatible fix for WooPay multi-currency

bborman22:
Fixes https://github.com/Automattic/woopay/issues/2076

This PR fixes the issue when a shopper selects a different currency from the store currency and then checks out using WooPay. 

This PR re-implements the fixes in #6331 that were reverted. In addition it takes a different approach to solving the issues discovered as done in #7005. In #7005 it was noted that the fix in #6331 exposed an unexpected error condition in certain environments when `WooPay_Session::determine_current_user_for_woopay` was run before `WC_Payments` was initialized. Rather than trying to filter out those conditions, this PR looks to solve that problem by moving the `WooPay_Session::init` call to be after `WC_Payments` is initialized.

The risk of this move would be if we re-introduce the issue in #4759. In my testing I was not encountering that issue when following the testing instructions provided in that PR. 

Due to the issues mentioned above it will be important to test three things:
- The multi-currency fix for WooPay
- That the multi-currency fix for WooPay doesn't re-introduce the Atomic issues we encountered
- That the "Draft Order fix" issue wasn't re-introduced by moving that WooPay_Session::init call

#### Testing instructions

These testing instructions are copied from their original PRs. This should be considered a starting point for testing this PR, but additional testing (including in Atomic environments) would be best given the long history of some of these changes.

**Multi-currency Fix**

1. Ensure WooPay is eligible and enabled.
2. Enable Multi-Currency + the currency switcher from WC > Settings > Multi-Currency.
3. Enable multiple currencies, for example: USD, JPY, EUR.
4. As a guest shopper use the currency switcher to change the currency to JPY.
5. Navigate to the checkout page and enter a valid WooPay email + OTP.
6. From the WooPay checkout, ensure the currency is displayed correctly.
7. In a separate tab, go to the merchant store and change the currency to, "USD", for example.
8. Refresh the WooPay checkout tab and ensure the currency, totals and currency symbol are set to USD.
9. Checkout with WooPay.
10. Ensure the charge is correctly processed for the chosen currency.
11. Logout from WooPay and repeat steps 4-10 as a logged in shopper.

**Atomic site fix**

1. Since this can only be tested on Atomic environments, upload a build from this branch into https://woopaytestmerchant.wpcomstaging.com/.
   a. Switch to this branch and run npm run build.
   b. Create a zip file from the generated build under the release folder.
   c. Upload the zip into an Atomic WooPay test site.
   d. When prompted, replace the existing plugin with the uploaded one.
2. Test the WooPay checkout. Ensure everything works as expected.

**Draft Order fix**

**Setup**

1. Enable WooCommerce > Settings > Accounts & Privacy > Allow customer to place orders without an account
2. Enable WooCommerce > Settings > Accounts & Privacy > Allow customers to create an account during checkout
3. This is just to make it easier to create or skip the account creation
4. Customer with site account

**Customer with site account**

1. As a non logged shopper add a product to the cart
2. Go to checkout
3. Create a new WooPay user by checking the option Save my data for faster checkouts
4. Check the option to create the user in the merchant site
5. Try to buy a product using WooPay
6. No errors should be generated and you should be redirected to the success page
7. No new draft order should be created
8. Guest customer

**Guest customer**

1. As a non logged shopper add a product to the cart
2. Go to checkout
3. Create a new WooPay user by checking the option Save my data for faster checkouts
4. Don't check the option to create the user in the merchant site
5. Try to buy a product using WooPay
6. No errors should be generated and you should be redirected to the success page
7. No new draft order should be created

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7055` or branch name `fix/woopay-multi-currency` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/woopay-multi-currency) :rocket:

---

Build info: 

- Latest commit: af6b53c78e88300801c924651ba34e1d3a3b9bf0
- Build time: 2023-09-08 12:41:34 UTC

woocommerce-payments.php
```diff
@@ -150,6 +144,11 @@ function () {
 function wcpay_init() {
 	require_once WCPAY_ABSPATH . '/includes/class-wc-payments.php';
 	WC_Payments::init();
+	/**
+	 * Needs to be loaded as soon as possible
+	 * Check https://github.com/Automattic/woocommerce-payments/issues/4759
+	 */
+	\WCPay\WooPay\WooPay_Session::init();
```
reykjalin, line 151: I'm a little bit suspicious on whether this will work or not. Before we moved the `init` call to the top of the file it was effectively called in `WC_Payments::init()`, so moving it here actually puts it _later_ in the execution cycle 🤔

https://github.com/Automattic/woocommerce-payments/blob/4b78576848769d30aadbe0c68ac100828c681706/includes/class-wc-payments.php#L337

(`maybe_register_platform_hooks` was the call that registered the action before https://github.com/Automattic/woocommerce-payments/pull/4762/)

I wonder if maybe we've changed how things actually work significantly enough for this no longer to be an issue? I'll make sure to test this part as thoroughly as I can 👍

includes/woopay/class-woopay-session.php
```diff
@@ -493,10 +486,6 @@ private static function get_woopay_verified_email_address() {
 	 * @return bool True if request is a Store API request, false otherwise.
 	 */
 	private static function is_store_api_request(): bool {
-		if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) {
-			return false;
-		}
```
reykjalin, line 505: Is removing this necessary? I guess maybe some of the checks are happening early enough that `REST_REQUEST` hasn't been defined yet, but it should only be possible to interact with the Store API through REST requests so I'm a bit surprised to see that we need to remove this condition 🤔

reykjalin (commented):
Code looks good, I just have some questions and concerns on a couple points.

I haven't tested the code yet and will be focusing on that tomorrow, but wanted to leave this comment just to start a discussion on some of the code.

woocommerce-payments.php
bborman22: Yeah I was wondering the same about if things have changed enough for this to not be an issue, but it is a concern of mine. In my initial testing it didn't appear to be an issue, but we may have to go the route of decoupling any reliance on `WC_Payments` being initialized.

includes/woopay/class-woopay-session.php
bborman22: > I guess maybe some of the checks are happening early enough that REST_REQUEST hasn't been defined yet

From my discussions with @ricardo that's exactly what was happening, the constant wasn't defined yet. I agree, the condition itself would make sense to keep, but again from my understanding the constant not being defined yet made some checks fail here and then prevented getting the correct shopper set to get their currency.

reykjalin:
I couldn't test in staging because I kept running into an authentication issue that seems unrelated to this PR. I'll come back to that tomorrow, in the meantime I'll do what testing I can locally. More context on the issue in p1692940108463869-slack-C022WMN88KG.

woocommerce-payments.php
bborman22: I've now reverted this part of the PR and instead removed the `WC_Payments` dependency from the `is_woopay_enabled` check.

includes/woopay/class-woopay-session.php
```diff
@@ -583,6 +578,28 @@ private static function create_woopay_nonce( int $uid ) {
...
+		}
+
+		// read directly from options to avoid loading WC_Payments.
+		$account      = get_option( 'wcpay_account_data', [] );
+		$account_data = [];
+		if ( is_array( $account ) && array_key_exists( 'data', $account ) ) {
+			$account_data = $account['data'];
+		}
+
+		return is_array( $account_data ) && ( $account_data['platform_checkout_eligible'] ?? false );
```
reykjalin: Minor nitpick; I think we can just use the null coalescing operator `??` here since [it doesn't throw errors on variables that don't exist](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce):

```sh
$ php --interactive
Interactive shell

php > echo $this_does_not_exist['foo']['bar'] ?? 'baz';
baz
php > $text = 'foo';
php > echo $text['foo']['bar'] ?? 'baz';
baz
php > $array = [ 'foo' => [] ];
php > echo $array['foo']['bar'] ?? 'baz';
baz
php > $array = [ 'foo' => [ 'bar' => 'foobar' ] ];
php > echo $array['foo']['bar'] ?? 'baz';
foobar
```

```suggestion
		return get_option( 'wcpay_account_data', [] )['data']['platform_checkout_eligible'] ?? false;
```

includes/woopay/class-woopay-session.php
```diff
@@ -535,7 +524,13 @@ private static function has_valid_request_signature() {
 	 * @return bool True if WooPay is enabled, false otherwise.
 	 */
 	private static function is_woopay_enabled(): bool {
-		return WC_Payments_Features::is_woopay_eligible() && 'yes' === WC_Payments::get_gateway()->get_option( 'platform_checkout', 'no' );
+		$payment_settings = get_option( 'woocommerce_woocommerce_payments_settings', [] );
+		$woopay_enabled   = 'no';
+		if ( is_array( $payment_settings ) && ! empty( $payment_settings['platform_checkout'] ) ) {
+			$woopay_enabled = $payment_settings['platform_checkout'];
+		}
```
reykjalin: Minor nitpick; I think we can just use the null coalescing operator `??` here since [it doesn't throw errors on variables that don't exist](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce):

```sh
$ php --interactive
Interactive shell

php > echo $this_does_not_exist['foo']['bar'] ?? 'baz';
baz
php > $text = 'foo';
php > echo $text['foo']['bar'] ?? 'baz';
baz
php > $array = [ 'foo' => [] ];
php > echo $array['foo']['bar'] ?? 'baz';
baz
php > $array = [ 'foo' => [ 'bar' => 'foobar' ] ];
php > echo $array['foo']['bar'] ?? 'baz';
foobar
```

```suggestion
		$woopay_enabled   = get_option( 'woocommerce_woocommerce_payments_settings', [] )['platform_checkout'] ?? 'no';
```

reykjalin (changes requested):

includes/woopay/class-woopay-session.php
bborman22: Thanks so much for this suggestion, was blanking on the best way to simplify this and didn't realize null coalescing was so forgiving.

woocommerce-payments.php
reykjalin: Awesome, thank you Brian! 🙌

reykjalin:
I seem to be running into an issue when I upload the build to the staging store:


https://github.com/Automattic/woocommerce-payments/assets/13835680/1d195ee8-dacc-4e6a-a783-c24c19efe0dc

It seems there's a problem on WooPay? Or maybe the store? I'm getting a `500 internal server error` on the `batch` Store API request (which is pretty much sending the `GET /cart` request).

I'm digging into this further but haven't been able to figure out what's going on yet. FWIW this tests well locally, but the staging environment is giving me some issues :)

**EDIT:** I have not been able to figure out what the issue is and I'm calling it a day for now. Maybe we can pair up on this on Monday.

bborman22:
After several days of trial and error, I was able to determine the key state that triggered our early 500 errors and why we needed the WooPay Session init to be so early. The short answer is if in Jetpack `Terms_Of_Service::has_agreed` returns false, it tries to look up the current user which hits our `determine_current_user` function really early (before even WC is initialized). I've now moved the WooPay Session init to be after `WC_Payments` is initialized as described in the original PR description. 

In my testing I was not able to reproduce the issue in #4759, though there are some issues with draft orders when third party cookies are NOT blocked and the shopper is already logged into WooPay (basically if the OTP modal triggers an auto-redirect). These issues are not related to what was happening in #4759 and my understanding is that #4759 issues are being primarily resolved by the use of the cart token when processing the request though I wasn't able to 100% confirm that. But I wanted to keep this PR moving along and so I'm requesting another review especially since we are also now able to reproduce the unique conditions locally.

@reykjalin let me know there are any issues you run into during testing and perhaps we could walk through them on a call since this PR seems susceptible to different environment factors.

includes/woopay/class-woopay-session.php
```diff
@@ -535,7 +534,9 @@ private static function has_valid_request_signature() {
 	 * @return bool True if WooPay is enabled, false otherwise.
 	 */
 	private static function is_woopay_enabled(): bool {
-		return WC_Payments_Features::is_woopay_eligible() && 'yes' === WC_Payments::get_gateway()->get_option( 'platform_checkout', 'no' );
+		$woopay_enabled = get_option( 'woocommerce_woocommerce_payments_settings', [] )['platform_checkout'] ?? 'no';
+
+		return self::is_woopay_eligible() && 'yes' === $woopay_enabled;
```
reykjalin: Since we delayed the initialization, can we go back to relying on the `WC_Payments_Features` and `WC_Payments` classes here? Or should we still decouple the code this way? 🤔 

I'm _leaning_ towards going back to using the WooPayments classes instead of decoupling here since we have guaranteed the other classes have been initialized.

includes/woopay/class-woopay-session.php
```diff
@@ -583,6 +584,22 @@ private static function create_woopay_nonce( int $uid ) {
...
+	private static function is_woopay_eligible() {
+		// Checks for the dependency on Store API AbstractCartRoute.
+		if ( ! class_exists( 'Automattic\WooCommerce\StoreApi\Routes\V1\AbstractCartRoute' ) ) {
+			return false;
+		}
+
+		// read directly from options to avoid loading WC_Payments.
+		return get_option( 'wcpay_account_data', [] )['data']['platform_checkout_eligible'] ?? false;
+	}
+
```
reykjalin: On the same note as the previous comment: I think we can remove this function in favor of `WC_Payments_Feautres::is_woopay_eligibile()`.

tests/unit/woopay/test-class-woopay-session.php
```diff
@@ -204,6 +199,6 @@ private function setup_adapted_extensions() {
 	}
 
 	private function set_is_woopay_eligible( $is_woopay_eligible ) {
-		$this->mock_cache->method( 'get' )->willReturn( [ 'platform_checkout_eligible' => $is_woopay_eligible ] );
+		update_option( 'wcpay_account_data', [ 'data' => [ 'platform_checkout_eligible' => $is_woopay_eligible ] ] );
```
reykjalin: If we go back to using the WooPayments classes I think we need to revert these tests?

reykjalin (changes requested):
This tests really well locally, but when I uploaded this build to the staging store the multi-currency doesn't seem to be fixed 🤔 

I didn't look too hard given that I won't be able to continue the review until late next week so it's probably better for someone else to investigate.

I just generated the build via `npm run build` on this branch after testing locally and uploaded that to the staging store, that doesn't seem to have fixed the problem like it does locally. I tried both by logging in via OTP and being redirected when already logged in, neither resulted in the right currency being used. It's still showing USD as the currency but with the converted amount, so a crazy high price when select 0-decimal currencies.

includes/woopay/class-woopay-session.php
bborman22: Yeah, if this is testing well with the delayed initialization, I would want to go back to the original implementation here to avoid some duplicated code.

bborman22:
> but when I uploaded this build to the staging store the multi-currency doesn't seem to be fixed

I was curious what your test looked like. I was able to see multi-currency working for: guest shopper, logged in shopper, and the create new account shopper. I wasn't able to test the one scenario we discussed where a shopper creates a cart in one currency, then logs in with a shopper with a different currency, because for me when I logged in, it cleared my cart which is different behavior from what I get locally.

I also gave this a test run locally with the Points and Rewards and Mailchimp Newsletter and had success with both. I'll be reverting those dependency decoupling changes now and will request another review!

reykjalin:
> **Warning**
> I've left the testing build I used up on the staging merchant site instead of reverting back to the Atomic managed version.

@bborman22 I actually didn't have to do anything special to get the multi-currency issue I was running into, just add a product to the cart, change the currency, and go to WooPay:

> **Note**
> Please ignore the confusion on the `my-account` page, I realized I don't have an account to log into via the `my-account` page 🙃

https://github.com/Automattic/woocommerce-payments/assets/13835680/ee89a2d4-a3c8-4566-80f9-008a1b96494c

### Repro steps

1. `npm run build` on this branch.
2. Upload the zip to https://woopay-test-merchant.blog (https://woopaytestmerchant.wordpress.com)
3. Open a private window.
4. Add any product to the cart.
5. Change the currency to a 0-decimal currency.
6. Go to the **Checkout** page and start a WooPay session
7. Once you're in WooPay note that the price is the correct amount, but wrong currency.

bborman22:
> I've left the testing build I used up on the staging merchant site instead of reverting back to the Atomic managed version.

Thanks for this warning. I actually think this is a good idea in general and that when not testing a specific version of the plugin on staging we should just upload the current stable version. This way we don't have to bother with switching from managed Atomic and back again, we can just always stay self managed and just upload the stable version to achieve the same result.

Also with that note, I noticed that for some reason what was currently on the staging site didn't have my latest changes applied to it. I checked this by going to the Plugin File Editor and checking out the `WooPay_Session::get_init_session_request` function for the `$currency` check / set. I think the version numbers being out of sync was causing it to still use the most recent. So I merged in `develop` to this branch and rebuilt and uploaded again and I can see the currency conversion working now.

Let me know if you get a chance to check it again and if it's not working, double check the plugin file editor has the changes showing.

Just to confirm to my future self that it definitely was working:


https://github.com/Automattic/woocommerce-payments/assets/68524302/07967976-b817-4da0-a36e-95a3584aa290

includes/woopay/class-woopay-session.php
```diff
@@ -338,6 +332,15 @@ private static function get_init_session_request() {
...
 		}
 
+		if ( 0 !== $user->ID ) {
+			$currency      = get_user_meta( $user->ID, MultiCurrency::CURRENCY_META_KEY, true );
+			$currency_code = strtoupper( $currency );
+
+			if ( ! empty( $currency_code ) && WC()->session ) {
+				WC()->session->set( MultiCurrency::CURRENCY_SESSION_KEY, $currency_code );
+			}
+		}
```
reykjalin: Can we add a comment here explaining why this is necessary? I could see this being an area where someone updating this code in the future has no idea why this is here 🤔

includes/woopay/class-woopay-session.php
```diff
@@ -500,10 +503,6 @@ private static function get_woopay_verified_email_address() {
 	 * @return bool True if request is a Store API request, false otherwise.
 	 */
 	private static function is_store_api_request(): bool {
-		if ( ! defined( 'REST_REQUEST' ) || ! REST_REQUEST ) {
-			return false;
-		}
```
reykjalin, line 505: Now that we've moved the initialization to a later point in the init process should we keep this check?

includes/woopay/class-woopay-session.php
```diff
@@ -542,6 +541,11 @@ private static function has_valid_request_signature() {
 	 * @return bool True if WooPay is enabled, false otherwise.
 	 */
 	private static function is_woopay_enabled(): bool {
+		// There were previously instances of this function being called too early. While those should be resolved, adding this defensive check as well.
+		if ( ! class_exists( WC_Payments_Features::class ) || ! class_exists( WC_Payments::class ) || is_null( WC_Payments::get_gateway() ) ) {
+			return false;
+		}
```
reykjalin, line 549: Love this addition 👏 ❤️

tests/unit/woopay/test-class-woopay-session.php
```diff
@@ -14,6 +14,11 @@
  * WooPay_Session unit tests.
  */
 class WooPay_Session_Test extends WCPAY_UnitTestCase {
+	/**
+	 * @var Database_Cache|MockObject
+	 */
+	protected $mock_cache;
+
```
reykjalin, line 21: I'm assuming we should be removing this as well? It's not used anywhere in the test file.

reykjalin (changes requested):
This tests great and looks great, I'm just requesting changes for some minor nitpicks and 1 instance of added code that's not doing anything in the tests. Once that's resolved I think we're good.

Awesome work here Brian! 🎉 👏

tests/unit/woopay/test-class-woopay-session.php
bborman22: I actually had just added it as part of reverting the tests as I was getting some lint complaints in the editor. It's used in the https://github.com/Automattic/woocommerce-payments/pull/7055/files/a63d301cb29eb550ae77f352d5d85c46309a9670#diff-e2ba3588802b8c413a54ea81192a62c55e1db5aaa904edf9119c08b4cfa62417R212 in the test.

includes/woopay/class-woopay-session.php
bborman22: Thanks, but credit here goes to @ricardo for his proposed fix in #7005. 👏 I loved the defensive programming here as well!

tests/unit/woopay/test-class-woopay-session.php
reykjalin: Ah, let's keep it then!

includes/woopay/class-woopay-session.php
bborman22: This one I'm less sure on how to approach. This was also from @ricardo's fix so was curious if you think adding this conditional back in is adding protection or not given that we've moved the init call to now be later in the process when the constant should be defined?

reykjalin: I think it's a valuable check it include, so I'd love to see us reintroduce this check if it doesn't break anything.

includes/woopay/class-woopay-session.php
bborman22: Great callout! Added a comment now in latest commit.

reykjalin: Looks good!

includes/woopay/class-woopay-session.php
bborman22: I'm able to confirm that adding it back in prevents the currency lookup on WooPay from working. I believe this makes sense as we're still binding the `determine_current_user` hook during the `plugins_loaded` hook and `REST_REQUEST` isn't defined until `parse_request` which I believe comes sometime after `init`.

reykjalin: Thank you for confirming, in that case I'll go ahead and approve ☺️

reykjalin (approved):

Issue #6023, opened 2023-04-13
New UPE rollout recipe


FangedParakeet:
### Description

The New UPE is essentially the split-UPE with deferred, server-side payment intent confirmation. Visually and functionally it appears basically the same as the existing split-UPE; however, the client-side is bereft of payment intent details, which can be confirmed securely on the server instead, and since we can perform this confirmation on the server with the UPE, we are now able to replace the card element with a new UPE payment element and still support WooPay.

We are working on a PoC that is actually an MVP in #5869; this feature will remain behind a client-side feature flag. This issue is to devise and decide how we will eventually release this feature to our users.

For users who are already using the split-UPE, this change should be relatively seamless and not really affect any of their interactions with the plugin. In theory, it appears as if would be relatively straight-forward to upgrade this subset of users automatically. However, for users who are still using the legacy UPE, we have not yet developed a natural way to migrate them to the split-UPE, without disabling and re-enabling the UPE. Similarly, there is no natural path to deliver the new UPE to them either.

Questions posed by this issue:
- In which scenarios should the new UPE feature flag be enabled for users?
- How can we migrate legacy UPE users?
- What kind of compatibility will we need to persist for legacy versions of the UPE, especially considering that legacy UPE is by its nature insecure? 

### Acceptance criteria

We will need to include many relevant WCPay stakeholders to ensure that we have a unified plan to release this new feature. This could culminate in a P2 post or alternatively a brief discussion and declaration of intent on this issue.

### Additional context

#5770

FangedParakeet:
Seeking the wisdom of @c-shultz & @pierorocca here for guidance towards our eventual [new UPE](https://github.com/automattic/woocommerce-payments/issues/5770) rollout plan. 

As of now the original split-UPE is the default payment gateway offered to new WCPay users; existing WCPay users are able to enable the split-UPE via the "new checkout experience"; whereas legacy UPE users currently do not have a direct pathway to the split-UPE without manually disabling and re-enabling the "new checkout experience". 

[...]

PR #7058, opened 2023-08-24
Enables deferred intent UPE for existing split UPE stores and newly onboarded stores

### LINKED ISSUE #7003: Enable Deferred Intent UPE for newly onboarded merchants
## Description

Merchants that are newly onboarded will have the Deferred Intent UPE feature flag enabled by default.

FangedParakeet:
Fixes #7003 

This PR will initiate the first phase of our deferred intent rollout. Please see #6023 for more details on this rollout plan.

With this PR stores that currently have the split UPE enabled will effectively be upgraded to the deferred intent UPE; additionally newly onboarded stores will also begin operating with the deferred intent UPE enabled. 

Effectively the split UPE flag (`_wcpay_feature_upe_split` or `WC_Payments_Features::UPE_SPLIT_FLAG_NAME`) is now made redundant and there should be no way for stores to use this feature: whenever this flag is enabled, stores will be using the deferred intent UPE. (This flag can probably be removed once our rollout is complete.) Consequently, in situations where the split UPE would be enabled, the deferred intent UPE will now be served in its place. If non-UPE users enabled the "new checkout experience", previously this would enable the split UPE; now this will enable the deferred intent UPE instead. If legacy UPE users enabled WooPay--which is incompatible with legacy UPE--previously the split UPE would also be enabled; now the deferred intent UPE will also be enabled.

This PR contains many changes to our unit tests to ensure they are now compatible with these scenarios. There are also a few fixes to unreliable unit tests to ensure that they are completely atomic when setting feature flags or setting global class constants (e.g. functions like `WC_Payments::set_gateway()`) to ensure that tests function as expected without affecting other tests. Additionally, there are some amends to our Tracks implementation to ensure that our analytics also reflect these new scenarios as well.

#### Testing instructions

We need to ensure that the deferred intent is enabled when it should be. In most cases you can confirm this easily by checking whether the deferred intent flag (`_wcpay_feature_upe_deferred_intent`) is enabled using the dev tools. However, when the split UPE flag (`_wcpay_feature_upe_split`) is enabled, the deferred intent will be served, though it's flag will not automatically be updated. The deferred intent UPE differs from the split UPE by having a new preloading state and also by presenting the CC gateway with the UPE instead of the legacy card element. These two GIFs should show the difference between the split and deferred intent UPE CC gateways--also, note the preloading state on the deferred intent UPE element that will not be present on any of the split UPE LPM gateways either. 

Split UPE CC card element
_Split UPE CC card element at checkout_

Deferred intent UPE CC card element
_Deferred intent UPE CC card element at checkout_

Here are the scenarios that we will need to confirm...

1. Newly onboarded store will use deferred intent UPE. _(Deferred intent UPE flag should be enabled.)_
2. Store with deferred intent UPE flag enabled will use deferred intent UPE.
3. Store without deferred intent UPE flag, but with split UPE flag will use deferred intent UPE.
4. With all flags disabled (legacy card), enable the "new checkout experience" from WooPayments settings. Store should now use deferred intent UPE. _(Deferred intent UPE flag should be enabled.)_
5. With _only_ legacy UPE enabled (`_wcpay_feature_upe`), enable WooPay from WooPayments settings. Store should now use deferred intent UPE. _(Deferred intent UPE flag should be enabled.)_
6. With _only_ legacy UPE enabled, disable "new checkout experience" from WooPayments settings. Now re-enable "new checkout experience" from WooPayments settings. Store should now use deferred intent UPE. _(Deferred intent UPE flag should be enabled.)_
7. With deferred intent UPE enabled, disabling "new checkout experience" should return you to the basic legacy card gateway.

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -1296,7 +1296,7 @@ class WC_Payments_Account {
 		// user might not have agreed to TOS yet.
 		update_option( '_wcpay_onboarding_stripe_connected', [ 'is_existing_stripe_account' => false ] );
 
-		// Automatically enable split UPE for new stores.
+		// Automatically enable deferred intent UPE for new stores.
 		update_option( WC_Payments_Features::UPE_DEFERRED_INTENT_FLAG_NAME, '1' );
 
 		// Track account connection finish.
--- a/includes/class-wc-payments-account.php
+++ b/includes/class-wc-payments-account.php
@@ -1265,7 +1265,7 @@ class WC_Payments_Account {
 		update_option( '_wcpay_onboarding_stripe_connected', [ 'is_existing_stripe_account' => false ] );
 
 		// Automatically enable split UPE for new stores.
-		update_option( WC_Payments_Features::UPE_SPLIT_FLAG_NAME, '1' );
+		update_option( WC_Payments_Features::UPE_DEFERRED_INTENT_FLAG_NAME, '1' );
 
 		// Track account connection finish.
 		$incentive        = ! empty( $_GET['promo'] ) ? sanitize_text_field( wp_unslash( $_GET['promo'] ) ) : '';
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -524,11 +524,7 @@ class WC_Payments {
 				self::$upe_payment_gateway_map[ $payment_method->get_id() ] = $split_gateway;
 			}
 
-			self::$card_gateway = self::get_payment_gateway_by_id( 'card' );
-
-			$card_payments_checkout = new WC_Payments_Checkout( self::$legacy_card_gateway, self::$woopay_util, self::$account, self::$customer_service );
-			$card_payments_checkout->init_hooks();
-
+			self::$card_gateway         = self::get_payment_gateway_by_id( 'card' );
 			self::$wc_payments_checkout = new WC_Payments_UPE_Checkout( self::get_gateway(), self::$woopay_util, self::$account, self::$customer_service );
 		} elseif ( WC_Payments_Features::is_upe_legacy_enabled() ) {
 			$payment_methods = [];
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -505,7 +505,7 @@ class WC_Payments {
 			Afterpay_Payment_Method::class,
 			JCB_Payment_Method::class,
 		];
-		if ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+		if ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 			$payment_methods = [];
 			foreach ( $payment_method_classes as $payment_method_class ) {
 				$payment_method                               = new $payment_method_class( self::$token_service );
@@ -728,7 +728,7 @@ class WC_Payments {
 	 * @return array The list of payment gateways that will be available, including WooPayments' Gateway class.
 	 */
 	public static function register_gateway( $gateways ) {
-		if ( WC_Payments_Features::is_upe_split_enabled() || WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+		if ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 
 			$payment_methods = self::$card_gateway->get_payment_method_ids_enabled_at_checkout();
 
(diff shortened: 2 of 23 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7058` or branch name `add/7003-deferred-intent-rollout-phase-1` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/7003-deferred-intent-rollout-phase-1) :rocket:

---

Build info: 

- Latest commit: c9022a3dcd3358c00a22d7f8057a7ee158245247
- Build time: 2023-08-25 17:21:59 UTC

includes/class-wc-payments.php
```diff
@@ -524,11 +524,7 @@ public static function init() {
 				self::$upe_payment_gateway_map[ $payment_method->get_id() ] = $split_gateway;
 			}
 
-			self::$card_gateway = self::get_payment_gateway_by_id( 'card' );
-
-			$card_payments_checkout = new WC_Payments_Checkout( self::$legacy_card_gateway, self::$woopay_util, self::$account, self::$customer_service );
-			$card_payments_checkout->init_hooks();
```
FangedParakeet, line 530: I'm a little nervous about removing these pieces and hooks...but all they do AFAICT is enqueue the scripts/DOM elements for the legacy card gateway, which should no longer be required, since the deferred intent UPE uses UPE elements for all gateways now.

includes/class-wc-payments-account.php
```diff
@@ -1297,7 +1297,7 @@ private function finalize_connection( $state, $mode ) {
 		update_option( '_wcpay_onboarding_stripe_connected', [ 'is_existing_stripe_account' => false ] );
 
 		// Automatically enable split UPE for new stores.
```
mdmoore, line 1299: Tiny nit just to update the comment here.

includes/class-wc-payments-account.php
```diff
@@ -1297,7 +1297,7 @@ private function finalize_connection( $state, $mode ) {
...
-		update_option( WC_Payments_Features::UPE_SPLIT_FLAG_NAME, '1' );
+		update_option( WC_Payments_Features::UPE_DEFERRED_INTENT_FLAG_NAME, '1' );
```
mdmoore, line 1300: Tiny nit just to update the comment here.

includes/class-wc-payments.php
mdmoore: This makes sense to me given that legacy card gateway will no longer be used here under any condition. Looks like we're getting all the scripts we need for the deferred intent in the `init_hooks` call [here](https://github.com/Automattic/woocommerce-payments/pull/7058/files#diff-57ee486c4a3ba4bae478eb31255c9a968b5b4eccd04b27f512e4003440d6f821R543). I think it's safe 👍.

mdmoore (approved):
All scenarios are testing well. Orders are processing as expected under each scenario, including WooPay orders. 

✅ Newly onboarded store will use deferred intent UPE. (Deferred intent UPE flag should be enabled.)
✅ Store with deferred intent UPE flag enabled will use deferred intent UPE.
✅ Store without deferred intent UPE flag, but with split UPE flag will use deferred intent UPE.
✅ With all flags disabled (legacy card), enable the "new checkout experience" from WooPayments settings. Store should now use deferred intent UPE. (Deferred intent UPE flag should be enabled.)
✅ With only legacy UPE enabled (_wcpay_feature_upe), enable WooPay from WooPayments settings. Store should now use deferred intent UPE. (Deferred intent UPE flag should be enabled.)
✅ With only legacy UPE enabled, disable "new checkout experience" from WooPayments settings. Now re-enable "new checkout experience" from WooPayments settings. Store should now use deferred intent UPE. (Deferred intent UPE flag should be enabled.)
✅ With deferred intent UPE enabled, disabling "new checkout experience" should return you to the basic legacy card gateway.

I left a couple comments but no real actions to take. Thanks for all the test updates and improvements! 🚢 🚢 🚢

includes/class-wc-payments-account.php
FangedParakeet: Nice catch! Updated in ad974d4ecda01124a46d74226a25e07cc28c9b3e. 👍

PR #7060, opened 2023-08-24
Add refund controls when transaction is missing order

### LINKED ISSUE #7001: Refunds: transaction details page, introduce refund controls when order is not connected
### Description

When an order is not linked (WooPayments incidents or store migrated without migration order tables), the following controls need to be available:

Screenshot from 2023-08-15 14-26-52
Screenshot from 2023-08-15 14-28-04

After the refund is performed, it's expected to see a confirmation (a black box showing up and disappearing later in the bottom left of the screen).

### Notes

- it's expected several smaller PRs for this feature, or we can split the ticket into several smaller issue
- only transaction status to be considered for refund availability here (not disputed, charged). Open disputes are not blocking refunds. Only paid transactions can be refunded (Auth&cpature requiring capture, Fraud prevention requiring approval/decline transaction can not be refunded). Only fully refundable transactions can be refunded with the controls.
- design review session pointed to improving the wording inside the yellow box
- as the order is not linked, a refund have to be performed directly via server APIs

KarlisJ:
Fixes #7001 

Add notice under transaction details card when missing order that has "refund" action button

#### Testing instructions
I haven't found a way to simulate payment without order (generally it shouldn't happen, and if we knew when it happens, we would fix the root cause) so best to alter the condition in the `client/payment-details/summary/index.tsx` file:
```JS
{ Object.keys( charge ).length > 0 && ! charge.order && (
	<MissingOrderNotice
		isLoading={ isLoading }
	         amount={ formattedAmount }
	/>
) }
```

			

-------------------

**Post merge**


client/components/card-notice/index.tsx
```diff
@@ -0,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import './styles.scss';
+import { Button, CardFooter } from '@wordpress/components';
```
mgascam: Button does not seem to be used

client/components/confirmation-modal/styles.scss
```diff
@@ -12,8 +12,8 @@
 	}
 
 	.components-modal__header {
-		margin: 0 -#{$grid-unit-30} $grid-unit-30;
```
mgascam: It seems that you need to fix a spacing issue? A friendly reminder that we should test that we are not affecting other usages of the component

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+	const handleOnButtonClick = () => {
+		setIsModalOpen( true );
+	};
+
+	const handleModalCancel = () => {
+		setIsModalOpen( false );
+	};
+
+	const handleModalConfirmation = () => {
+		// TODO: Handle the refund.
```
mgascam, line 44: Do you know if we plan to add Tracks events?

client/payment-details/summary/style.scss
```diff
@@ -71,18 +71,3 @@
 .woocommerce-list--horizontal {
 	margin-top: 0 !important;
 }
-
-.payment-details-capture-notice {
```
mgascam, line 97: Nice clean-up after refactor

mgascam (commented):
I like the approach so far, left some minor comments.

client/components/confirmation-modal/styles.scss
KarlisJ: I think this is change from other PR I'm working on that has crept into this one. Thanks for pointing out, I'll remove it from here

client/payment-details/summary/missing-order-notice/index.tsx
KarlisJ: I don't. And for the scope of PR I think we'll stop with empty handler until we have clarity on how the backed will work. I'll put this under some flag before merging.

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7060` or branch name `add/7001-refund-control-when-no-order` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/7001-refund-control-when-no-order) :rocket:

---

Build info: 

- Latest commit: 7019a8b3ccc00279ee023dfbe4d2b2008b94438a
- Build time: 2023-11-02 19:51:54 UTC

client/payment-details/summary/index.tsx
```diff
@@ -378,6 +384,14 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 			{ isDisputeOnTransactionPageEnabled && charge.dispute && (
 				<DisputeDetails dispute={ charge.dispute } />
 			) }
+			{ isRefundControlsEnabled &&
+				Object.keys( charge ).length > 0 &&
```
kalessil: Is there more elegant alternative to `Object.keys( charge ).length > 0`?

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Internal dependencies.
+ */
+
+import './style.scss';
+import CardNotice from 'wcpay/components/card-notice';
+
+interface MissingOrderNoticeProps {
+	isLoading: boolean;
+	amount: string;
```
kalessil: Can we use a better naming perhaps to reflect it's pre-formatted amount?

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+								),
+								formattedAmount
+							),
+							components: {
+								strong: <strong />,
+							},
+						} ) }
+					</p>
+					<RadioControl
+						className="missing-order-notice__modal__reason"
```
mgascam: This class name deviates a bit from the BEM convention. Usually you'll only see "__" used once to denote an Element that belongs to a Block, but in this case we have two. I'd say that the modal is Block, and then we have reason element. Something like:

```suggestion
						className="missing-order-notice-modal__reason"
```

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+					}
+				>
+					{ __(
+						'This transaction is not connected to order. Investigate this purchase and refund the transaction as needed.',
+						'woocommerce-payments'
+					) }
+				</CardNotice>
+			</Loadable>
+			{ isModalOpen && (
+				<ConfirmationModal
```
mgascam, line 68: This component supports passing class names. I'd take advantage to pass `missing-order-notice-modal`. In terms of BEM, this would be the Block.

```suggestion
				<ConfirmationModal
				    className="missing-order-notice-modal"
```

client/payment-details/summary/missing-order-notice/style.scss
```diff
@@ -0,0 +1,9 @@
+.missing-order-notice__modal__reason {
```
mgascam: This selector should be amended following my other comments around BEM convention.

client/payment-details/summary/missing-order-notice/style.scss
```diff
@@ -0,0 +1,9 @@
...
+	.components-base-control__field {
```
mgascam: I believe we can reduce nesting by removing` components-base-control__field` here and still target the `label`

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+							components: {
+								strong: <strong />,
+							},
+						} ) }
+					</p>
+					<RadioControl
+						className="missing-order-notice__modal__reason"
+						label={ interpolateComponents( {
+							mixedString: __(
+								'{{strong}}Select a reason (Optional){{/strong}}',
```
mgascam: I'd prefer to implement this via CSS so we don't need to use  `interpolateComponents`

mgascam (changes requested):
Like the direction but I'm requesting some changes related to CSS and BEM naming convention.

client/payment-details/summary/missing-order-notice/style.scss
```diff
@@ -0,0 +1,10 @@
+.missing-order-notice-modal__reason {
+	.components-base-control__label {
```
mgascam, line 2: Thanks for styling the label with CSS. It seems we can remove the usage of interpolateComponents [here](https://github.com/Automattic/woocommerce-payments/blob/6f2d772a9389348d27f55b464fc1b6741517b7c1/client/payment-details/summary/missing-order-notice/index.tsx#L104)?

mgascam (changes requested):
Thanks for addressing my comments, I requested another small change.

client/payment-details/summary/missing-order-notice/style.scss
KarlisJ: I think you've referenced an old commit. This has already been done in https://github.com/Automattic/woocommerce-payments/pull/7060/commits/d54edf09577ccd820191d310d0254a014d46b517

client/components/card-notice/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+import React from 'react';
+
+/**
+ * Internal dependencies
+ */
+import './styles.scss';
+import { CardFooter } from '@wordpress/components';
+
+interface CardNoticeProps {
```
mgascam, line 12: We are missing the `children` type here, is it intentional?

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+		// TODO: Handle the refund.
+	};
+
+	return (
+		<>
+			<Loadable isLoading={ isLoading } placeholder="">
+				<CardNotice
+					actions={
+						<Button
+							isPrimary={ true }
```
mgascam: VSCode shows a notice saying `isPrimary` prop is marked as deprecated.

Screenshot 2023-10-20 at 19 53 33

Can we use the suggested approach `variant="primary"` instead of `isPrimary`?

```suggestion
							variant="primary"
```

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+					) }
+				</CardNotice>
+			</Loadable>
+			{ isModalOpen && (
+				<ConfirmationModal
+					className="missing-order-notice-modal"
+					title={ __( 'Refund Transaction', 'woocommerce-payments' ) }
+					actions={
+						<>
+							<Button onClick={ handleModalCancel } isSecondary>
```
mgascam:
```suggestion
							<Button onClick={ handleModalCancel } variant="secondary">
```

client/payment-details/summary/index.tsx
```diff
@@ -50,6 +50,8 @@ import WCPaySettingsContext from '../../settings/wcpay-settings-context';
 import { FraudOutcome } from '../../types/fraud-outcome';
 import CancelAuthorizationButton from '../../components/cancel-authorization-button';
 import { PaymentIntent } from '../../types/payment-intents';
+import MissingOrderNotice from 'wcpay/payment-details/summary/missing-order-notice';
+import _ from 'lodash';
```
mgascam: Lodash is an external dependency, can you move it to the top of the file please?

client/payment-details/summary/index.tsx
```diff
@@ -487,7 +494,14 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 					) }
 				</ErrorBoundary>
 			) }
-
+			{ isRefundControlsEnabled &&
+				! _.isEmpty( charge ) &&
+				! charge.order && (
+					<MissingOrderNotice
+						isLoading={ isLoading }
+						formattedAmount={ formattedAmount }
```
mgascam, line 499: While testing I noticed that the `formattedAmount` that will be refunded includes the shipping costs. I believe it is expected, but wanted to double check.

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Internal dependencies.
+ */
+
+import './style.scss';
+import CardNotice from 'wcpay/components/card-notice';
+
+interface MissingOrderNoticeProps {
+	isLoading: boolean;
+	formattedAmount: string;
```
mgascam, line 24: Non blocking: I'd suggest a different name for this prop since I found `formattedAmount` a bit generic. Maybe `amountToBeRefunded`?

mgascam (commented):
Finished a new review round and it tests well. I left some comments, mostly minor, for your consideration. Let's go through them before final approval, thanks!

client/payment-details/summary/missing-order-notice/index.tsx
KarlisJ: We have the amount available from `charge.amount` but we pass it through a method `formatAmount` to generate display value, thus the naming. 

Same value is then used to display the amount of order in the UI and in the modal before refunding, thus putting `refunded` in naming would be weird since it's used outside the "refund".

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -0,0 +1,145 @@
+/** @format **/
+
+/**
+ * External dependencies
+ */
+
+import React from 'react';
```
kalessil, line 7: There are some internal deps listed here, which should be in another block of imports.

kalessil (approved):
LGTM (code review only; pre-appropving)! Leaving it to the team to do more review and testing rounds here.

client/payment-details/summary/missing-order-notice/index.tsx
mgascam: Thanks for explanation. For clarity's sake,  I was referring to the name of the `formattedAmount` property of the `MissingOrderNoticeProps` component, not to the name of the variable.

mgascam (approved):
LGTM! Did a final test round and works as intended.

PR #7061, opened 2023-08-24
Fixed issue in extending requests class when request class was already extended.

### LINKED ISSUE #7007: Issue in Extending requests class during runtime
### Describe the bug

I wanted to extend request during runtime following the [official guideline](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/core/server/CONTRIBUTING.md#extending-requests-during-runtime).

I created a new class named `My_Namespace\Prefix_Create_And_Confirm_Intention` to extend the `Create_And_Confirm_Intention class`. My purpose was to include additional information in the request during the checkout process. To achieve this, I utilized a callback function bound to the `wcpay_create_and_confirm_intent_request` hook with a priority of 20, to run it after a [similar callback is run by WooPay](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/woopay/services/class-checkout-service.php).

Sample code:
```
namespace My_Namespace;

use WCPay\Core\Server\Request\Create_And_Confirm_Intention;

class Prefix_Create_And_Confirm_Intention extends Create_And_Confirm_Intention {
  // some custom function
}
```

```
add_filter( 'wcpay_create_and_confirm_intent_request', [ __CLASS__, 'add_dynamic_statement_descriptor' ], 20, 2 );
...
public static function add_dynamic_statement_descriptor( Request $base_request, Payment_Information $payment_information ) {
		$request = Prefix_Create_And_Confirm_Intention::extend( $base_request );
		// some customization
		return $request;
}
```

However, when attempting to complete the checkout, an error emerged: 
```
"Failed to extend request. My_Namespace\Prefix_Create_And_Confirm_Intention is not a subclass of WCPay\Core\Server\Request\WooPay_Create_And_Confirm_Intention."
```

Afterwards, I tried adjusting the priority from 20 to 5 (to run it before WooPay), which led to a similar error: 
```
"Failed to extend request. `WCPay\Core\Server\Request\WooPay_Create_And_Confirm_Intention` is not a subclass of `My_Namespace\Prefix_Create_And_Confirm_Intention`."
```

Upon examining the codebase, I discovered the root cause: as the `wcpay_create_and_confirm_intent_request` is applied, if another callback extends the base request and send the new request down the chain, the next callback can not extend the base request anymore. Because the base request of the second filter is the one filtered by the first filter and no parent-child relationship exists between the two.

In this context, when I attempted to extend the base class with `Prefix_Create_And_Confirm_Intention`, it appears that the base class had already been altered and no longer corresponded to `Create_And_Confirm_Intention`.
 
Could you kindly review this situation and provide some guidance?

zmaglica:
Fixes #7007 

#### Fixed issue when request class is extended from the request class that was already extended.

In our codebase, we've enhanced the `Create_And_Confirm` request class with filters. During payment processing, we employ a [filter](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/woopay/services/class-checkout-service.php#L30) to modify certain parameters of the original `Create_And_Confirm` request class. 

If you try to create a new request class using the `extend` or `apply_filters` function based on a class that has already been extended similarly with those functions, you'll be met with an error. Refer to the https://github.com/Automattic/woocommerce-payments/issues/7007 for more details. 

This PR addresses the issue by verifying if the extended request class is a subclass of the class it's extended from or if they share a common parent. Simply put, the class you pass in the `extend()` function needs to be the parent class or grandfather class (parent of parent, but not base default Request abstract class) of the new class you wish to create.

#### Testing instructions

* Make sure to run all cases described in the GitHub issue description.

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-7007-extend-request-class-bugfix
+++ b/changelog/fix-7007-extend-request-class-bugfix
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed issue with validation of extended request class
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -675,17 +675,27 @@ abstract class Request {
 	 */
 	public function validate_extended_class( $child_class, string $parent_class ) {
 
-		if ( ! is_subclass_of( $child_class, $parent_class ) ) {
-			throw new Extend_Request_Exception(
-				sprintf(
-					'Failed to extend request. %s is not a subclass of %s',
-					is_string( $child_class ) ? $child_class : get_class( $child_class ),
-					$parent_class
-				),
-				'wcpay_core_extend_class_not_subclass'
-			);
+		// Check if the child class directly extends the parent class.
+		if ( is_subclass_of( $child_class, $parent_class ) ) {
+			return;
 		}
 
+		// Retrieve the immediate parent (grandparent) of the specified parent class.
+		$grandparent_class = get_parent_class( $parent_class );
+
+		// Check if the child class extends the grandparent class, ensuring that the grandparent class exists
+		// and is not the same as the class where this method is defined.
+		if ( $grandparent_class && self::class !== $grandparent_class && is_subclass_of( $child_class, $grandparent_class ) ) {
+			return;
+		}
+		throw new Extend_Request_Exception(
+			sprintf(
+				'Failed to extend request. %s is not a subclass of %s',
+				is_string( $child_class ) ? $child_class : get_class( $child_class ),
+				$parent_class
+			),
+			'wcpay_core_extend_class_not_subclass'
+		);
 	}
 
 	/**
--- a/tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
+++ b/tests/unit/core/server/request/test-class-core-create-and-confirm-intention-request.php
@@ -6,6 +6,7 @@
  */
 
 use PHPUnit\Framework\MockObject\MockObject;
+use WCPay\Core\Exceptions\Server\Request\Extend_Request_Exception;
 use WCPay\Core\Exceptions\Server\Request\Immutable_Parameter_Exception;
 use WCPay\Core\Exceptions\Server\Request\Invalid_Request_Parameter_Exception;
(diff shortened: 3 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7061` or branch name `fix/7007-extend-request-class-bugfix` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7007-extend-request-class-bugfix) :rocket:

---

Build info: 

- Latest commit: 6053bbc5b46e9f1a04247f3178381150af56c933
- Build time: 2023-08-25 12:53:45 UTC

RadoslavGeorgiev (approved):
The changes look good, and tests are working well! 🚢 

As an additional idea, maybe we can instantiate a private `$base_request_class` property when `apply_filters` is called at first, and do a `while ( $parent_class = get_parent_class( $class_name ) && $parent_class !== $base_request_class )` loop? This way we wouldn't be focusing on parents, grandparents, etc, but just looping. WDYT?

zmaglica:
> The changes look good, and tests are working well! 🚢
> 
> As an additional idea, maybe we can instantiate a private `$base_request_class` property when `apply_filters` is called at first, and do a `while ( $parent_class = get_parent_class( $class_name ) && $parent_class !== $base_request_class )` loop? This way we wouldn't be focusing on parents, grandparents, etc, but just looping. WDYT?

The problem with the code you provided is that it won't function correctly in scenarios similar to the following:

```php
class Parent {
    public static function extend(Request $base_request) {
        //
    }
}

class ChildOne extends Parent {}
class ChildTwo extends Parent {}
ChildTwo::extend(new ChildOne());
```

This almost closely mirrors the issue we're discussing.

In order to make your example both effective and bulletproof, we must retrieve all parent classes for the two given classes. We then need to verify if they share the same parent by comparing their extended classes.

PR #7092, opened 2023-08-29
Refactor Woo Subscriptions compatibility to fix currency being able to be updated during renewals, resubscribes, or switches

### LINKED ISSUE #5507: Currency symbol on cart and checkout is wrong when switching subscription
### Describe the bug

When switching subscription, the currency symbol and decimal format follows the customer currency where it should follow the currency when the subscription was initially purchased.

### To Reproduce

1. Enable switching in subscription settings (wp-admin > WooCommerce > Settings > Subscriptions)

Screenshot 2023-02-02 at 17 20 27

2. Enable [multi-currency](https://woocommerce.com/document/woocommerce-payments/currencies/multi-currency-setup). Add some currencies, eg `INR` and `JPY`.
3. Create 2 simple subscriptions, one is cheaper than the other.
4. Create a grouped product and add the 2 subscription products under the 'Linked products' option.
5. In USD, purchase the cheaper subscription product.
6. Go to My Accounts > Subscriptions > View the newly purchased subscription.
7. Click on the 'Upgrade or Downgrade' link.
8. Choose the more expensive subscription.
9. Open a new tab and change the customer currency to something else, eg `JPY`.
10. Notice the currency symbol and decimal format (eg `JPY` is a zero-decimal currency and the price is displayed without decimal points) follows the customer currency and not the subscription's currency which is `USD`.

### Actual behavior

The currency symbol and decimal format (eg `JPY` is a zero-decimal currency and the price is displayed without decimal points) follows the customer currency.

### Screenshots

https://user-images.githubusercontent.com/73803630/216297786-a694a853-adb4-4429-820b-0ccfefabe85c.mov

### Expected behavior

The currency symbol and decimal format should follow the subscription's currency which is `USD`.

```[tasklist]
### Tasks
- [x] Replicate the issue.
- [x] Determine why the currency is able to be changed.
- [x] Stop the currency from being able to be changed during a switch or renewal.
- [x] See if tests can be added to make sure this cannot happen again.
```

jessepearson:
Fixes #5507

This is a refactor to hopefully finally resolve the issues we've seen with subscription renewals, resubscribes, and switches where the currency can be changed during one of those processes. The largest change is the addition of the `get_subscription_type_from_cart` method, which will check the `WC()-cart` array and the `WC->session->cart` array for the specified subscription type being in the cart. Both are needed due to at different points in time they may not be in sync. 

* Updated checks where `count()` was used to confirm if an array was empty or not to use `empty()`, as the check is faster.
* The `override_selected_currency` method was simplified by being able to utilize new `get_subscription_type_from_cart` method for checking for items in the cart.
* Replaces `cart_contains_renewal`, `get_subscription_switch_cart_items`, and `cart_contains_resubscribe` methods with the new `get_subscription_type_from_cart` method.
* Update `get_subscription_switch_id_from_superglobal` method to `get_subscription_from_superglobal_switch_id` and have it return the subscription itself instead of the ID of the subscription. I found that the subscription was queried again after the ID was returned, so it works better if the subscription itself is returned.
* Added additional methods and properties into `WC_Subscription` helper class for tests.
* With the refactoring, nearly every test needed to be updated. When starting on the update of the tests, there were around 49 tests with 54 assertions. There are now 50 tests with 100 assertions. The tests and their assertions try to cover all possible instances of happy/sad paths, and try to account for any changes to the methods themselves, meaning they try to fail if any of the code is changed.

#### Setup for Multi-Currency

* Make sure you have Multi-Currency enabled under Payments > Settings.
* Under WooCommerce > Settings > Multi-Currency, make sure you have JPY as an additional currency along with another currency, such as CAD.
* If you have not added the Multi-Currency widget, this will need to be added.
  * Under Appearance > Widgets add the Currency Switcher Block to a sidebar on your site.

#### Setup for WooCommerce Subscriptions

* You will need WooCommerce Subscriptions installed. This can be found in the `all-plugins` repo. p1693496664612309-slack-C01B8KNUYSW
* Create a simple subscription product with a signup fee, make sure it bills monthly.
* Create a second simple subscription product with a higher cost and a higher signup fee, make sure it bills monthly.
* Create a grouped product with both simple subscriptions under it.
* Create a variable subscription product with a signup fee, make sure that it has at least 2 variations with different prices, such as 10/month and 20/month.
* Create a _Recurring Product Discount_ coupon that is less than the amount of the products you're going to use for testing.
* Create a _Recurring Product % Discount_ coupon.
* Make sure you have _Early Renewals_ enabled under WooCommerce > Settings > Subscriptions.
* Go to WooCommerce > Settings > Subscriptions:
  * Enable switching for both variable and grouped products, also set to prorate recurring payment for upgrades and downgrades.
  * Enable _ Accept Early Renewal Payments_ under _Early renewals_.

**_Note: All of the below testing should be done with a currency that is NOT your default store currency. Each test will need to be done twice, first while adding your _Recurring Product Discount_ coupon to your cart, and then again while adding your _Recurring Product % Discount_ coupon to your cart. The amounts should convert correctly in each step._**

#### Manual renewal order testing

* Take note that you have a currency other than your default store currency selected.
* View one of your simple subscriptions in your shop, add it to cart. 
* Apply the respective subscription coupon, confirm the amounts are converted correctly, then check out.
* Navigate to WooCommerce > Subscriptions, and click into this new subscription.
* Use _Subscription actions_ at the top right to _Create pending renewal order_ and click the arrow button.
* Navigate to the new order through the link in the _Order notes_.
* The order will have a _Customer payment page_ link above the Status, click that link.
* Currency switcher widget should be hidden.
* Add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Proceed to renew the subscription.

#### Early renewal testing

* You should still have the confirmation page up from your last renewal. If not, go to My Account > Subscriptions.
* Click to view the Subscription.
* Click to _Renew Now_.
* Currency switcher widget should be hidden.
* Add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Confirm the coupon amount is correct.
* Proceed to renew the subscription.

#### Resubscribing testing

* Navigate to WooCommerce > Subscriptions in the admin.
* Hover over your newest subscription and click Cancel, once the page refreshes hover again and choose Cancel Now.
* Navigate to My Account > Subscriptions in the front end, and then click View for the cancelled subscription.
* Click the Resubscribe button.
* Currency switcher widget should be hidden.
* Add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Apply the respective subscription coupon, confirm the amounts are converted correctly.
* Proceed to purchase the subscription.

#### Subscription switching testing - variable product

* Clear your cart if not empty.
* View your variable subscription product you created on the front end.
* Select the cheapest variation of the subscription, add to cart.
* Apply the respective subscription coupon, confirm the amounts are converted correctly, then check out.
* Once order is complete, scroll down to use the _View_ option to view the subscription.
* Use the _Upgrade or Downgrade_ button on the subscription to start the process.
* Currency switcher widget should be hidden.
* Add `&currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Choose the higher priced variation on the product and add to cart.
* On the cart page apply the respective coupon again and confirm the amounts are converted correctly.
* Add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* On the checkout page add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Confirm the coupon amount is correct.

#### Subscription switching testing - grouped product

* Clear your cart if not empty.
* View your grouped subscription product you created on the front end.
* Select the cheapest option of the subscription, add to cart.
* Apply the respective subscription coupon, confirm the amounts are converted correctly, then check out.
* Once order is complete, scroll down to use the _View_ option to view the subscription.
* Use the _Upgrade or Downgrade_ button on the subscription to start the process.
* Currency switcher widget should be hidden.
* Add `&currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* Choose the higher priced product in the group and add to cart.
* On the cart page apply the respective coupon again and confirm the amounts are converted correctly.
* Add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.
* On the checkout page add `?currency=JPY` to the url in the address bar.
* The costs on the page should still be in the currency of the subscription itself, including the currency code and format.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7092` or branch name `fix/5507-block-currency-update-on-sub-switch` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/5507-block-currency-update-on-sub-switch) :rocket:

---

Build info: 

- Latest commit: 7a9d4af42b40a7991888a4197304841c3a91d97e
- Build time: 2023-09-06 12:45:13 UTC

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -143,56 +145,36 @@ public function maybe_disable_mixed_cart( $value ) {
...
 			return $return;
 		}
 
-		$subscription_renewal = $this->cart_contains_renewal();
-		if ( $subscription_renewal ) {
-			$order = wc_get_order( $subscription_renewal['subscription_renewal']['renewal_order_id'] );
-			return $order ? $order->get_currency() : $return;
-		}
+		// Loop through subscription types and check for cart items.
+		foreach ( self::SUBSCRIPTION_TYPES as $type ) {
```
jessepearson, line 161: This loop now handles checking for a renewal, resubscribe, or switch in the cart. This includes `order_awaiting_payment`, which there was previously a check for.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -143,56 +145,36 @@ public function maybe_disable_mixed_cart( $value ) {
...
-			return $subscription ? $subscription->get_currency() : $return;
+				$this->running_override_selected_currency_filters = false;
+				return $subscription ? $subscription->get_currency() : $return;
+			}
 		}
 
-		return $return;
+		// This instance is for when the customer lands on the product page to choose a new subscription tier.
+		$switch_subscription = $this->get_subscription_from_superglobal_switch_id();
+		return $switch_subscription ? $switch_subscription->get_currency() : $return;
```
jessepearson, line 177: If this needs to be written more clearly, let me know. Basically, it tries to get a sub, and if the sub exists, get its currency, otherwise return what was initially passed.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
```diff
@@ -284,61 +266,64 @@ public function should_hide_widgets( bool $return ): bool {
...
 	/**
 	 * Getter for subscription objects.
 	 *
 	 * @param  mixed $the_subscription Post object or post ID of the order.
+	 *
 	 * @return mixed The subscription object, or false if it cannot be found.
-	 *               Note: this is WC_Subscription|bool in normal use, but in tests
-	 *               we use WC_Order to simulate a subscription (hence `mixed`).
+	 *               Note: This should be WC_Subscription|bool, but Psalm throws errors like:
+	 *                     Docblock-defined class, interface or enum named WC_Subscription does not exist (see https://psalm.dev/200)
```
jessepearson, line 326: If there are any thoughts on how to fix this, I am open to it. It may be better to leave it like this since Subs is going to be removed from WCPay.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
allie500: It looks like previously the `order_awaiting_payment` check would then check if that order contained a renewal. Is it correct for me to assume that the three types in the `SUBSCRIPTION_TYPES` would return true if checked for `( isset( WC()->session ) && WC()->session->get( 'order_awaiting_payment' ) )`? 

I just want to make sure I correctly understand how this loop includes an `order_awaiting_payment` check similar to the previous code.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
allie500: It makes sense to me.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
jessepearson: The new check has the same result, previously we did this:

```
$order = wc_get_order( WC()->session->get( 'order_awaiting_payment' ) );
			$this->running_override_selected_currency_filters = false;
			if ( $order && $this->order_contains_renewal( $order ) ) {
				return $order->get_currency();
			}
```

This would get the `order_awaiting_payment` from the session, then check that order to see if it had a subscription renewal in it.

Now we just check either the cart itself, or the cart in the session with `$cart_item = $this->get_subscription_type_from_cart( $type )` to see if there's a renewal present. If there is an `order_awaiting_payment` this will catch it because the renewal item will be set in the cart or session cart.

This was available previously, it just wasn't being looked for. It's likely I wrote the initial code and just didn't realize that session could be checked this way. It's cleaner and more definitive.

Hope that helps.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
allie500: I would say it may be best to leave it since subscriptions will be removed from WCPay.

includes/multi-currency/Compatibility/WooCommerceSubscriptions.php
allie500: Yes, this helps. Thanks for clarifying!

allie500 (approved):
Thanks for all the work you've done for this refactor, Jesse. All the tests passed for me. Thanks for the excellent PR description explaining what was changed and why. The test instructions were also easy to follow.

I spent a good bit of time reviewing the unit tests file in particular. The tests look good to me. I'm pretty confident that you've covered all the scenarios with those tests. The changes in the subscriptions compatibility file are easy to understand.

LGTM! :shipit:

PR #7095, opened 2023-08-30
Fix the way request constants are traversed

RadoslavGeorgiev:

`Request::traverse_class_constants` is used to combine the constants of multiple classes, which inherit each-other, like `DEFAULT_PARAMS`. The way the method was initially coded, it started with the current class, and while traversing through parent classes, it added to the array. However, that turns out to be an issue, as properties of parent classes end up overwriting the properties of child classes, instead of the other way around.

Example:

```php
class Request {
	const DEFAULT_PARAMS = [];
}

class Paginated extends Request {
	const DEFAULT_PARAMS = [
		'sort' => 'created',
	];
}

class List_Transactions extends Paginated {
	const DEFAULT_PARAMS = [
		'sort' => 'date',
	];
}

$defaults = List_Transactions::traverse_class_constants( 'DEFAULT_PARAMS' );
var_dump( $defaults );

array(1) {
	["sort"]=>
	string(1) "created"
}
```

As you can see, instead of `date`, `sort` remains `created`.

**With this PR, the order of traversing classes is reversed,** and the result above would be `date`.

#### Testing instructions

You could tinker with loading transactions in the admin, but the unit test is the best way to make sure this works.

-------------------

### DIFF
--- a/changelog/fix-request-constant-traversing
+++ b/changelog/fix-request-constant-traversing
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fix the way request params are loaded between parent and child classes.
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -537,24 +537,18 @@ abstract class Request {
 	 * @return string[] The unique combined values from the arrays.
 	 */
 	public static function traverse_class_constants( string $constant_name, bool $unique = false ) {
-		// Generate a class tree first, in order to reverse it later.
-		$class_tree = [];
+		$keys       = [];
 		$class_name = static::class;
-		do {
-			$class_tree[] = $class_name;
-			$class_name   = get_parent_class( $class_name );
-		} while ( $class_name );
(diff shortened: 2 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7095` or branch name `fix/request-constant-traversing` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/request-constant-traversing) :rocket:

---

Build info: 

- Latest commit: dfb4278e8ff75ae560eb1882fcdaeb5503dcc5ec
- Build time: 2023-08-30 08:54:24 UTC

zmaglica (approved):

zmaglica:
:shipit:

PR #7112, opened 2023-09-01
Pass pay for order param to init WooPay session

hsingyuc:
Fixes 2135-gh-Automattic/woopay

This PR passes `pay_for_order` param to init WooPay session and pass a fake `order_id` to checkout data as a workaround for the [checkout order error](https://github.com/woocommerce/woocommerce-blocks/blob/04f36065b34977f02079e6c2c8cb955200a783ff/assets/js/blocks/checkout/block.tsx#L81-L83.).  This is a quick fix and we can optimize it more later.

#### GIFs
|Before|After|
|--|--|
|ezgif com-optimize|Screen Shot 2023-09-01 at 11 48 20 AM|
|Screen Shot 2023-09-01 at 10 00 10 AM|

#### Testing instructions
1. Clone the latest WooCommerce Blocks `11.0.0-dev` and install WooCommerce (>=8.0.2) on your merchant and WooPay site. It should include the two [endpoints](https://github.com/woocommerce/woocommerce-blocks/blob/75c115e84262fee379e0a301fd0f58e1fc7772b8/src/StoreApi/RoutesController.php#L65-L68).
3. Test with this WooPay branch 2136-gh-Automattic/woopay
4. Create a pay-for-order flow order (ex. WC Bookings or Pre-orders)
5. Go through WooPay checkout process

-------------------

### DIFF
--- a/includes/woopay/class-woopay-session.php
+++ b/includes/woopay/class-woopay-session.php
@@ -381,8 +381,8 @@ class WooPay_Session {
 			'email'                => '',
 			'store_data'           => [
 				'store_name'                     => get_bloginfo( 'name' ),
-				'store_logo'                     => ! empty( $store_logo ) ? get_rest_url( null, 'wc/v3/payments/file/' . $store_logo ) : '',
-				'custom_message'                 => WC_Payments::get_gateway()->get_option( 'platform_checkout_custom_message' ),
+				'store_logo'                     => $store_logo,
+				'custom_message'                 => self::get_formatted_custom_message(),
 				'blog_id'                        => Jetpack_Options::get_option( 'id' ),
 				'blog_url'                       => get_site_url(),
 				'blog_checkout_url'              => ! $is_pay_for_order ? wc_get_checkout_url() : $order->get_checkout_payment_url(),
--- a/includes/woopay/class-woopay-session.php
+++ b/includes/woopay/class-woopay-session.php
@@ -358,7 +358,14 @@ class WooPay_Session {
 
 		$account_id = WC_Payments::get_account_service()->get_stripe_account_id();
 
-		$store_logo = WC_Payments::get_gateway()->get_option( 'platform_checkout_store_logo' );
+		$site_logo_id      = get_theme_mod( 'custom_logo' );
+		$site_logo_url     = $site_logo_id ? ( wp_get_attachment_image_src( $site_logo_id, 'full' )[0] ?? '' ) : '';
+		$woopay_store_logo = WC_Payments::get_gateway()->get_option( 'platform_checkout_store_logo' );
+
+		$store_logo = $site_logo_url;
+		if ( ! empty( $woopay_store_logo ) ) {
+			$store_logo = get_rest_url( null, 'wc/v3/payments/file/' . $woopay_store_logo );
+		}
 
 		include_once WCPAY_ABSPATH . 'includes/compat/blocks/class-blocks-data-extractor.php';
 		$blocks_data_extractor = new Blocks_Data_Extractor();
--- a/includes/class-wc-payments-checkout.php
+++ b/includes/class-wc-payments-checkout.php
@@ -209,6 +209,27 @@ class WC_Payments_Checkout {
 	}
 
 	/**
+	 * Add the Pay for order params to the JS config.
+	 *
+	 * @param WC_Order $order The pay-for-order order.
+	 */
+	public function add_pay_for_order_params_to_js_config( $order ) {
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		if ( isset( $_GET['pay_for_order'] ) ) {
+			add_filter(
+				'wcpay_payment_fields_js_config',
+				function( $js_config ) use ( $order ) {
+					$js_config['order_id']      = $order->get_id();
+					$js_config['pay_for_order'] = sanitize_text_field( wp_unslash( $_GET['pay_for_order'] ) );
+
+					return $js_config;
+				}
+			);
+		}
+		// phpcs:enable WordPress.Security.NonceVerification.Recommended
+	}
+
+	/**
 	 * Renders the credit card input fields needed to get the user's payment information on the checkout page.
 	 *
 	 * We also add the JavaScript which drives the UI.
--- a/includes/class-wc-payments-checkout.php
+++ b/includes/class-wc-payments-checkout.php
@@ -209,28 +209,6 @@ class WC_Payments_Checkout {
 	}
 
 	/**
-	 * Add the Pay for order params to the JS config.
-	 *
-	 * @param WC_Order $order The pay-for-order order.
-	 */
-	public function add_pay_for_order_params_to_js_config( $order ) {
-		// phpcs:disable WordPress.Security.NonceVerification.Recommended
-		if ( isset( $_GET['pay_for_order'] ) && isset( $_GET['key'] ) ) {
-			add_filter(
-				'wcpay_payment_fields_js_config',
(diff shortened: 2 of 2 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7112` or branch name `fix/init-woopay-error` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/init-woopay-error) :rocket:

---

Build info: 

- Latest commit: 77f66b2d0ff50f41b2e5424c93460da6f272c8b1
- Build time: 2023-09-21 19:27:52 UTC

client/checkout/woopay/express-button/express-checkout-iframe.js
```diff
@@ -250,7 +254,9 @@ export const expressCheckoutIframe = async ( api, context, emailSelector ) => {
 					true
 				);
 				if ( e.data.redirectUrl ) {
-					window.location = e.data.redirectUrl;
+					window.location = appendRedirectionParams(
```
malithsen: This flow does not call the `api.initWooPay` method. So the `order_id` is not passed into the session creation flow. Is that intentional?

malithsen (changes requested):
@hsingyuc Thanks for your work here. A couple of questions.

For users going through the OTP model flow (without automatic redirection), WooPay will emit the `redirect_to_woopay_skip_session_init` post message. That flow doesn't call the `initWooPay` function and updates to the session initiation request will not be made, which seems like an omission to me.

I am curious what happens when a user sets the value of `window.wcpayConfig.order_id` to someone else's order id? I haven't been able to test this because I keep running into the issue I mentioned above. Are there safeguards to prevent potential PII leakage?

client/checkout/woopay/express-button/express-checkout-iframe.js
hsingyuc: Do you mean we should remove `appendRedirectionParams` here?

hsingyuc: Fixed here [6d0cd3c](https://github.com/Automattic/woocommerce-payments/pull/7112/commits/6d0cd3c39dfbf681b7af6e06a07a36ac23205293)

hsingyuc:
>I am curious what happens when a user sets the value of window.wcpayConfig.order_id to someone else's order id? I haven't been able to test this because I keep running into the issue I mentioned above. Are there safeguards to prevent potential PII leakage?

Yes.  The `key`, and the `billint_email` are still required.

hsingyuc:
Do you know how the session nonce works?  I logged in and the first session nonce seemed incorrect, so I could not place the order but placed the order the second time the session nonce was recreated correctly, so I could successfully place the order.

hsingyuc:
Test with both user order with different billing email and guest order.


https://github.com/Automattic/woopay/assets/56378160/ec7da9e1-ec8b-4a86-818c-3ecfad59c137

includes/woopay/class-woopay-session.php
```diff
@@ -368,8 +375,8 @@ private static function get_init_session_request() {
 			'email'                => '',
 			'store_data'           => [
 				'store_name'                     => get_bloginfo( 'name' ),
-				'store_logo'                     => $store_logo,
-				'custom_message'                 => self::get_formatted_custom_message(),
+				'store_logo'                     => ! empty( $store_logo ) ? get_rest_url( null, 'wc/v3/payments/file/' . $store_logo ) : '',
+				'custom_message'                 => WC_Payments::get_gateway()->get_option( 'platform_checkout_custom_message' ),
```
malithsen: Just want to confirm the passing of the option value directly without any formatting in `get_formatted_custom_message` is intentional as it doesn't appear to be directly relevant to the PR.

malithsen (commented):
Thanks for working on this @hsingyuc 
Changes make sense. Following the testing instructions, I can proceed to to WooPay in the pay-for-order. I attempted completing an order with the latest WooPay changes, however I am unable to complete an order due to WooPay's pre-flight check failures. 

I'd like to get another pair of eyes on this cc @alefesouza

malithsen (commented):
Thanks for working on this @hsingyuc 
Changes make sense. Following the testing instructions, I can proceed to to WooPay in the pay-for-order. I attempted completing an order with the latest WooPay changes, however I am unable to complete an order due to WooPay's pre-flight check failures. 

I'd like to get another pair of eyes on this cc @alefesouza

alefesouza:
@malithsen is this WooPay pre-flight check failure an error which you click the `Place Order` button, it changes to loading state and then turn back to normal state in less than 1 second?

alefesouza (changes requested):
I had WooCommerce Product Bundles installed to work on https://github.com/Automattic/woocommerce-payments/pull/7158, this PR has an incompatibility with this extension, if we start the WooPay Pay-for-order flow with Product Bundles installed, we get a `Your cart is currently empty!` error message and this error is logged on the merchant WCPay logs:

`PHP Fatal error:  Uncaught Error: Call to a member function get_cart() on null in /var/www/html/wp-content/plugins/woocommerce-product-bundles/includes/api/class-wc-pb-store-api.php:517`

Disabling the Product Bundles extension fixes the problem.

hsingyuc:
@alefesouza Can you confirm without the Product Bundles plugin activated, you can redirect to the WooPay checkout page and get the pay-for-order order without an error?

includes/woopay/class-woopay-session.php
hsingyuc: Thank you for catching that!  Fixed in [132344b](https://github.com/Automattic/woocommerce-payments/pull/7112/commits/132344bf2a9cb76ea06c761dbe98d53c7f002122).

alefesouza:
@hsingyuc without the Product Bundles plugin installed it works as expected.

hsingyuc:
>@hsingyuc without the Product Bundles plugin installed it works as expected.

@malithsen Do you think we can merge this PR as it is?

malithsen:
>> @hsingyuc without the Product Bundles plugin installed it works as expected.

> @malithsen Do you think we can merge this PR as it is?

@hsingyuc Do we know the underlying issue that causes the empty cart? As in, is it limited to having Product Bundles and using WooPay's pay for order flow? Or could this happen with other extensions? 

If the issue is only affecting the WooPay pay for order flow, I feel like it's okay to merge and later solve the incompatibility issue. But I'll defer to @LCmry @pierorocca to confirm whether we are okay with the limitation.

Let's also fix the failing [test case](https://github.com/Automattic/woocommerce-payments/actions/runs/6189221702/job/16802870755?pr=7112).

hsingyuc:
>Do we know the underlying issue that causes the empty cart? As in, is it limited to having Product Bundles and using WooPay's pay for order flow? Or could this happen with other extensions?

When we initialize WooPay, we pass cart and checkout data as preloaded endpoints.  If the cart is empty, WooPay sees an empty cart.  Pay for orders preload this endpoint with pay for order data.

>If the issue is only affecting the WooPay pay for order flow, I feel like it's okay to merge and later solve the incompatibility issue. But I'll defer to @LCmry @pierorocca to confirm whether we are okay with the limitation.

As far as I know, the preload request only affects the WooPay pay-for-order flow.

hsingyuc:
Moving the changes to this [PR](https://github.com/Automattic/woocommerce-payments/pull/7253) and closing the one

PR #7129, opened 2023-09-05
Add New Task to Enable Payment Methods

### LINKED ISSUE #6779: New task should be added to set up additional payment methods after enabling payouts in PO
### Description
Currently after completing the progressive onboarding flow, merchants are redirected to the Payments Overview with a success message however there's no prompt to install APMs. This is part of the old onboarding flow, which redirects to the "Set up payments" task after KYC completion, and required as part of partner onboarding requirements.

### Designs
After KYC is complete and account is connected in full (bank account details included), a task appears on the Payments overview with the title,

> Add more ways for buyers to pay
> Enable payment methods that work seamlessly with WooPayments

The task is dismissible.

Payments Overview
Overview

Woo Home
The same task will appear on Woo Home, however without the subtext due to restrictions.
Home - Set up WooCommerce Payments

cc @dmallory42 @brandonwanA8C 
Related to [#37542](https://github.com/woocommerce/woocommerce/issues/37542)

dmallory42:
Fixes #6779

* Adds a new task that displays on both the `Overview` and main WooCommerce pages when a user has completed setup, but has not yet enabled any additional payment methods.
* Fixes a bug with the APM setup flow where a notice would appear in some unintentional circumstances (it was intended to only display to PO incomplete users but was displaying to other users too).

#### Testing instructions

* Verify task matches the design in #6779.
* Run `npm run listen` on your server.
* Ensure you have no APMs enabled on your store (you can do this by disabling them in `Payments > Settings`, so that only `Card` is enabled.
* Onboard a new account using Progressive Onboarding.
* You will be taken to the `Overview` page after partial onboarding to reach "PO enabled" status.
* Verify that you cannot see the new task. (Because you aren't yet `PO Complete`, you cannot yet add APMs.
* Complete full onboarding to enable deposits.
* Verify that you see the new task on the `Overview` page. 
* Click on the new task, you should be redirected to the APM setup.
* Enable at least one new Payment Method and complete the APM setup.
* Go back to the `Overview` page and verify the task is now gone.

## The task on the Overview page:

Screenshot 2023-09-05 at 12 30 14

## The task on WooCommerce Home:

Screenshot 2023-09-05 at 12 30 30

## Implementation Notes

- The task should display in the following conditions:
      - User has completed KYC (submitted their details)
      - User is not in the `PO Enabled` state (they either need to be `PO Complete`, or a normal, non-PO account - we show the task for normal users because they will also be redirected to the Overview page after onboarding.)
      - User has exactly one payment method enabled. 
- Currently, we display the task if exactly 1 payment method is enabled (this would be card only, since card is required). Ideally, I'd like to hide the task after the user has visited the APM setup, but I'm not sure on the best way to handle it. And maybe it is desired the task displays again if the user disables their APMs? (cc. @elizaan36 for thoughts on this 🙂)

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7129` or branch name `dev/6779-po-new-task` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=dev/6779-po-new-task) :rocket:

---

Build info: 

- Latest commit: cdf5cc235e45a2531940c52ae64cca0884c0ccad
- Build time: 2023-09-11 09:48:38 UTC

ismaeldcom (approved):
Looking good so far, testing instructions work as expected. Thanks for fixing the notice issue too 🙂 

I only see a tiny difference, according to the design reference, the task on WC Home doesn't show the `additionalInfo`. Not sure if that was intended in the design.

>Ideally, I'd like to hide the task after the user has visited the APM setup, but I'm not sure on the best way to handle it.

From the technical side, probably the simpler way is to create an option on APM setup visit, to avoid displaying it again.

Preapproving it! 🚀

dmallory42:
Thanks for the review, @ismaeldcom!

> I only see a tiny difference, according to the design reference, the task on WC Home doesn't show the additionalInfo. Not sure if that was intended in the design.

Yes - I noticed that too, I chose to include it because I saw it is actually also included in other similar tasks on the WC home page. But I can remove it if we think it's better that way. Also - it references `WooPayments` in the additional info - it could be confusing without that because we already have other tasks in WC Core relating to adding new payment methods (meaning Stripe/PayPal/etc)

I'll also make an update to include your suggestion of using an option when the APM setup is complete. Good idea!

Will wait to make sure everything looks good from a design POV before merging.

ismaeldcom:
>I'll also make an update to include your suggestion of using an option when the APM setup is complete. Good idea!

Let me know if you'd like a new review on this part 🙂

elizaan36:
hey, The original thinking is that this task would replace the "Set up additional payment methods" task.

> Click on the new task, you should be redirected to the APM setup.
> Enable at least one new Payment Method and complete the APM setup.

Could you share the APM setup screen as well? Is this the vertical stepper? Because the two tasks need to be combined into one flow rather than repeated.

Sorry that this wasn't explicitly called out, but it will be confusing for the merchant to have two tasks encouraging them to set up APMs.

dmallory42:
Hi @elizaan36!

> hey, The original thinking is that this task would replace the "Set up additional payment methods" task.

The `Set up additional payment methods` task that we have in WooCommerce Core actually relates to adding additional payment gateways (not additional payment methods) - the name of the task is slightly confusing so to make it easier to talk about I've attached screenshots below of where both flows lead:

"Set up additional payment methods" flow:
Screenshot 2023-09-06 at 12 51 43

The APM flow (triggered by this new task):
Screenshot 2023-09-06 at 12 51 51

The issue we have is that the "set up additional payment methods" flow is tied to the WC Core code (not the WCPay code), so what we'd probably need to do is check if the task is present, remove it if it is and replace it with this task. (which we can do).

> Could you share the APM setup screen as well? Is this the vertical stepper? Because the two tasks need to be combined into one flow rather than repeated.

Shared screenshot above of the vertical stepper. 🙂 This would probably need to be a new task because as these two tasks live in different code bases, we'd need to see how much work it would be to combine them into one flow and it may make sense for us to start from a design of what we'd want the combined flow to look like and then build it.

In terms of moving this task forward, would it be okay to try and replace the "Set up additional payment methods" task from WC Core with this new task to set up APMs, or would you prefer to have both flows combined before shipping this? (as I mentioned, this would be a much bigger task so would affect when it could be delivered). 

Thanks! 🙂

elizaan36:
> In terms of moving this task forward, would it be okay to try and replace the "Set up additional payment methods" task from WC Core with this new task to set up APMs, or would you prefer to have both flows combined before shipping this? (as I mentioned, this would be a much bigger task so would affect when it could be delivered).

I think we'll need to combine them because hiding the "Set up additional..." task would conflict with some of our contractual commitments that state certain payment gateways/methods "need to be part of onboarding".

I know this has been discussed (somewhere :D ) before, but from the core perspective we can add a section to enable the WooPayments APMs to the "Set up additional..." task. I had a look and I have two possible directions for this. Both options would direct the merchant through the vertical stepper guided flow to set up APMs, then when setup is complete they would be directed back either to Woo Home or Payments overview.

Option 1 - New section in "Set up additional..." task
Payments _ WooCommerce Payments-1

Option 2 - One-time directed flow from "Set up additional..." task. Announcement would dismiss once setup is complete.
Payments _ WooCommerce Payments

Let me know your thoughts on these options. This needs more work from a design perspective, so I'll likely share with Rik next week and he can hit the ground running on this.

elizaan36:
Since the "Set up additional..." task is in the core codebase does that mean we can't have identical tasks on both Woo Home and the Overview? (Maybe a dumb question.)

I was thinking that for the purpose of this issue we could ship only the "Add more ways..." task on the Payments Overview for now, but ideally the tasks would be consistent and contain the same options to avoid confusion.

daquinons:
> > In terms of moving this task forward, would it be okay to try and replace the "Set up additional payment methods" task from WC Core with this new task to set up APMs, or would you prefer to have both flows combined before shipping this? (as I mentioned, this would be a much bigger task so would affect when it could be delivered).
> 
> I think we'll need to combine them because hiding the "Set up additional..." task would conflict with some of our contractual commitments that state certain payment gateways/methods "need to be part of onboarding".
> 
> I know this has been discussed (somewhere :D ) before, but from the core perspective we can add a section to enable the WooPayments APMs to the "Set up additional..." task. I had a look and I have two possible directions for this. Both options would direct the merchant through the vertical stepper guided flow to set up APMs, then when setup is complete they would be directed back either to Woo Home or Payments overview.
> 
> Option 1 - New section in "Set up additional..." task Payments _ WooCommerce Payments-1
> 
> Option 2 - One-time directed flow from "Set up additional..." task. Announcement would dismiss once setup is complete. Payments _ WooCommerce Payments
> 
> Let me know your thoughts on these options. This needs more work from a design perspective, so I'll likely share with Rik next week and he can hit the ground running on this.

Hi @elizaan36! If we combine them in any of the two proposed options, how would the completion of the task be set? 
- If a local payment from WCPay is enabled, do we mark the task as done? Wouldn't that interfere with the mentioned commitments with third-party gateways?
- If a third-party gateway is enabled but not a local payment method from WCPay, do we mark it as done?
- Should it be marked as done only if a third party gateway + local payment method are enabled?

In my opinion, I think combining both tasks in one, even if it can sound logical (hey, they are both for adding payment methods at the end!) is not the best option here: one task is for adding new extensions and is dependant of Core, the other task is an internal WooPayments task to encourage the user to set up a setting in WooPayments. Both enable payment options, yes, but the fact that is not clear how to consider an action as completed should give us an indication that they might not be, in essence, doing the same action.

I think our  ultimate goal would be to present both actions to the user in a way that is impossible to get confused by having both at the same time :) And also provide options for discard them if the user is not interested in either installing other payment gateways or enabling local payment methods.

elizaan36:
### Task on payments overview 

I think the main problem to solve is that Progressive Onboarding lands the user on the Payments overview with no next step. There are currently no recommendations to add APMs/LPMs at all. For the purpose of this issue we could ship only the "Add more ways..." task on the Payments Overview to give them some guidance in this regard.

@frosso I'm looping you in here for your guidance on the changes to the UPE. We're planning to implement a new task that would direct new WP users through the LPM setup. (Screenshot)
image

Will this setup step be necessary once everyone is migrated to the split UPE? Or will these methods be enabled by default?

### Core payments setup

The core payments setup flow deserves more discussion, but I'll add some answers below.

> how would the completion of the task be set?

@daquinons The core payments task would be completed upon setup of any APM, whether within WooPayments or not. The store setup flow is meant to be a jumpstart to reach success with your store. If that encourages the user to setup PayPal, for example, or go through and enable the LPMs on checkout, they'll be that one step closer to setup and the task can be marked as complete.

> the fact that is not clear how to consider an action as completed should give us an indication that they might not be, in essence, doing the same action.

I don't follow this logic. From the user perspective these things are not different. They only know "payments are enabled on my store" and "here are some additional options". Most merchants have no comprehension of the concept of WooPayments as a separate plugin, they just want to enable the payment methods that will get them more sales.

> one task is for adding new extensions and is dependant of Core, the other task is an internal WooPayments task to encourage the user to set up a setting in WooPayments

I think we need to move away from the thinking that WooPayments is so separate from other payments options that it would require an entirely new task on the home screen. All of the payment options listed in the "Set up additional..." task once WP is set up are complementary to WP. An additional task creates unnecessary fragmentation of the experience, which is already noisy and complex, and cognitive load. 

The user needs more guidance upon successful completion of the KYC, so I'd recommend focusing on the Payments Overview for the scope of this update.

frosso:
> Will this setup step be necessary once everyone is migrated to the split UPE? Or will these methods be enabled by default?

@elizaan36 the UPE migration won't be making changes to the current behavior - no method besides Credit Card will be enabled by default. Enabling new methods means that new currencies would potentially need to be enabled as well (e.g.: US-based store adding Bancontact - Euro currency would need to be enabled).

dmallory42:
Thanks everyone for the input. I will ship the new task on the Payments screen only to address the issue of users not knowing what to do next after completing onboarding. 

I've created a new issue: https://github.com/Automattic/woocommerce-payments/issues/7166 to continue the discussion around unifying the payment tasks!

client/index.js
```diff
@@ -296,8 +296,8 @@ addFilter(
 		const { showUpdateDetailsTask, wpcomReconnectUrl } = wcpaySettings;
 
 		const wcPayTasks = getTasks( {
-			showUpdateDetailsTask,
-			wpcomReconnectUrl,
+			showUpdateDetailsTask: showUpdateDetailsTask,
+			wpcomReconnectUrl: wpcomReconnectUrl,
```
ismaeldcom, line 300: _Nitpick_: we can keep the old way, because `{x: x}` equals `{x}` 🙂

ismaeldcom (approved):
Latest changes make sense, and keeps working as expected.

PR #7134, opened 2023-09-05
Bump WC and PHP versions

ricardo:
The changes and this PR were made via the [wcpay:bump-versions](https://github.com/woocommerce/woorelease-extensions/blob/trunk/class-woocommerce-payments-bump-versions.php) command.

- Bump WC 'tested up to' version to 7.9 and minimum version to 7.7.
- Bump PHP version to 7.4 as per [WC 8.1](https://github.com/woocommerce/woocommerce/blame/8.1.0-a.4/plugins/woocommerce/woocommerce.php).


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7134` or branch name `dev/bump-min-wc-8-1-php-7-4` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=dev/bump-min-wc-8-1-php-7-4) :rocket:

---

Build info: 

- Latest commit: 4b6ec589af3d96056c835614a66982b7fc6b0d89
- Build time: 2023-09-13 21:35:27 UTC

ricardo:
@Automattic/harmony WooCommerce 8.1 will require PHP 7.4 ([src](https://github.com/woocommerce/woocommerce/blame/8.1.0-a.4/plugins/woocommerce/woocommerce.php)) so I ran `wcpay:bump-versions` with `--php_min_version="7.4"`, however, in the past it seems like we didn't catch up with WC consistently and we've updated PHP based on our own dependencies requirements. What's the best approach here?

The `wcpay:bump-versions` works great, but had to tweak a few things manually:
- Changed the `composer.json` PHP requirement to 7.4 as in WC ([src](https://github.com/woocommerce/woocommerce/commit/4876ab35b8b30b1c8c81e545d9a35c6d9b43799e)).
- Changed the PHP version matrix [here](https://github.com/Automattic/woocommerce-payments/blob/72dc9d284402be41b20994927e03f55c61d8d636/.github/workflows/php-lint-test.yml#L47) to remove PHP 7.3 and add 2 newer versions.

cc @cesarcosta99

cesarcosta99:
@Automattic/harmony, any input on the approach taken here? See comment above.

---

I reintroduced the environment variable `GUTENBERG_VERSION_FOR_WP_MIN` in `.github/workflows/compatibility.yml`. That was being used when this PR was opened. It got removed in #7142 as part of a revert and its absence was causing the `Compatibility - WC, WP and PHP / WC compatibility (7.9.0, 6.1, 15.7.0, 7.4)` check to fail ([using version 13.6.0](https://github.com/Automattic/woocommerce-payments/actions/runs/6152176518/job/16693817520)). @shendy-a8c, since you've opened that revert PR, any objections or concerns with this, or any change, in this PR? I see the revert was meant to resolve part of an incident so your insight is much appreciated here.

jbordonado:
I think we should not bump our dependency requirements until #7147 is done, given that doing it recently triggered a patch release 6.4.1, as you rightly noticed

Also, we're bumping the minimum WC version by 3, which could be too much for merchants.. Maybe let's increment by one at each release until we caught up again with the latest?

shendy-a8c:
> @shendy-a8c, since you've opened that revert PR, any objections or concerns with this, or any change, in this PR? I see the revert was meant to resolve part of an incident so your insight is much appreciated here.

No concern around `GUTENBERG_VERSION_FOR_WP_MIN` environment variable introduction.

https://github.com/Automattic/woocommerce-payments/pull/7142 was solely a revert PR to move forward with our recent 6.4.1 patch release.

rafaelzaleski (approved):
LGTM 👍 
Should `PHP testing (7.3)` be removed from the list of required checks for trunk merging?

cesarcosta99:
>Should PHP testing (7.3) be removed from the list of required checks for trunk merging?

@rafaelzaleski, I think that check is in the limbo because we removed 7.3. Looks like in other PRs we don't see it that way, and in fact, the 7.3 one is duplicated – that is being fixed in this PR, though, and as soon this PR is merged we'll have 7.4, 8.0 and 8.1 being tested.

image

cesarcosta99:
>Also, we're bumping the minimum WC version by 3, which could be too much for merchants.. Maybe let's increment by one at each release until we caught up again with the latest?

@eduardoumpierre, just FYI, I only bumped 1 minimum WC version, which means we need to bump it again in the following two releases, at least.

jbordonado:
Was there any progress on the incident fix? I thought we agreed to wait for a proper fix before bumping again, but maybe I missed something

cesarcosta99:
@jbordonado if we should've waited, that was unclear. The only call out so far was to not bump the dependency requirements, which I reverted.

PR #7140, opened 2023-09-06
HACK week: Display server error messages on Settings save

### LINKED ISSUE #7090: Saving Settings page should display an error message in case of server error
### Describe the bug
With the current implementation of saving the WooPayments Settings page, success is always returned from the backend, which is causing a vast amount of issues with validating the data on the backend. 

So far we successfully bypassed that issue with a client validation. However, sometimes it is not possible to correctly validate the input on the client side, and in case of an error, we shouldn't display a successful message.

### To Reproduce
*Case 1*
1. Connect a Japanese account.
1. Go to the `WooCommerce` settings page.
2. Set the country to Japan and save.
3. Navigate to `WooPayments` -> `Settings` and scroll down to `Support phone number`
4. Enter a non-Japanese phone number.
5. Save the settings and reload the page.
6. Verify the phone number is not updated.

### Expected behavior
Whenever a server error is returned on the Settings page save it should be displayed.

mordeth:
Fixes #7090

This PR improves the handling and displays the server errors on the WooPayments settings page. It acts like a second layer of error handling for account update errors that are not covered by front-end or back-end validation. The server errors are rarely a case, but when they happen, the save settings action is displayed as successful, however, the settings are not preserved.

Screenshot 2023-09-07 at 11 29 11

#### Testing instructions
Testing the server errors is tricky as we validate the account inputs in both the front-end and back-end. However, with minor code modifications, we could bypass the validations.

*Case 1*
- Open the [following file](https://github.com/Automattic/woocommerce-payments/blob/12ac12d4cfacabc8edc9a45ba362c670fa3a9397/includes/class-wc-payment-gateway-wcpay.php#L2600) and on line 2600 add the following code `return $value;`
- Navigate to `Payments` -> `Settings`.
- Enter `Test` in the `Customer bank statement`.
- Save settings.
- Verify the `Error: The statement descriptor matches a common term or website URL, and can't be used.` error is displayed.

*Case 2*
- Open the [following file](https://github.com/Automattic/woocommerce-payments/blob/12ac12d4cfacabc8edc9a45ba362c670fa3a9397/includes/class-wc-payment-gateway-wcpay.php#L2600) and on line 2600 add the following code `return $value;`
- Navigate to `Payments` -> `Settings`.
- Enter `Проба` in the `Customer bank statement`.
- Save settings.
- Verify the `Error: The statement descriptor must contain at least one Latin character.` error is displayed.

*Case 3*
- Remove all modifications on the [file](https://github.com/Automattic/woocommerce-payments/blob/12ac12d4cfacabc8edc9a45ba362c670fa3a9397/includes/class-wc-payment-gateway-wcpay.php#L2600) 
- Navigate to `Payments` -> `Settings`.
- Enter `Проба` in the `Customer bank statement`.
- Save settings.
- Verify no server error is displayed and inline error is displayed.

Screenshot 2023-09-07 at 11 37 16

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7140` or branch name `add/settings-save-error-notice` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/settings-save-error-notice) :rocket:

---

Build info: 

- Latest commit: b15fddb008e0a17b72b91e0e679b1439643a97ca
- Build time: 2023-10-04 10:05:04 UTC

includes/admin/class-wc-rest-payments-settings-controller.php
```diff
@@ -709,7 +716,11 @@ private function update_account( WP_REST_Request $request ) {
 			$updated_fields['deposit_schedule_interval'] = $this->wcpay_gateway->get_option( 'deposit_schedule_interval' );
 		}
 
-		$this->wcpay_gateway->update_account_settings( $updated_fields );
+		try {
+			$this->wcpay_gateway->update_account_settings( $updated_fields );
+		} catch ( Exception $e ) {
+			throw new Exception( $e->getMessage() );
```
anu-rock: I'm confused as to why we are rethrowing the same exception here.

Also, looks like this way we'll lose access to other useful information about the original exception such as stack trace.

anu-rock: The same comment applies to other places in the below code.

anu-rock (commented):
Thanks for taking the time to make our Settings page better 🎉 

While the current approach of displaying upstream server errors is functional (black error notice in bottom left), it's different from how we display existing errors (like the one in Case 3 in PR's testing instructions). IMO, this difference may result in a disjointed user experience.

Ideally, we should display upstream errors in a red box near the field that caused it. At the same time, I also understand that this fix was done as part of HACK Week and there may not be enough time for my suggested change.

From my perspective, it's okay to go ahead with the current approach. But we should check with Design once to see if this won't be a blocker.

___

#### Review checklist

| Parameter  | Result |
| Approach  | 👍  Almost good (see my comments above) |
| Implementation  | ✅ All good |
| Automated tests  | ✅ All good |
| Naming  | ✅ All good |
| Typing  | ✅ All good |
| Code comments  | ✅ All good |
| Changelog  | ✅ All good |
| Testing instructions  | ✅ All good |
| Manual testing  | ✅ Works as expected |

mordeth:
> Ideally, we should display upstream errors in a red box near the field that caused it.

Unfortunately, we do not get any information on the server side for which field the error is returned. The server errors would be rarely displayed, which won't break the UX.

includes/admin/class-wc-rest-payments-settings-controller.php
anu-rock: @mordeth Any thoughts on this?

anu-rock (approved):
Pre-approving with one open comment.

includes/admin/class-wc-rest-payments-settings-controller.php
mordeth: Updated, I have used a `WP_Error` instead.

anu-rock: Looks like the change is not pushed yet?

mordeth: Sorry, it should be pushed now!

dpaun1985 (approved):
✅ LGTM

anu-rock:
Looks good after the latest commit https://github.com/Automattic/woocommerce-payments/pull/7140/commits/6a749fc4e653dfcafcde7c17ef9f571bd3db381a. I recommend also getting it reviewed with someone from Harmony who have been doing a lot of work on error handling.

Also, a bunch of GH checks are failing.

includes/admin/class-wc-rest-payments-settings-controller.php
anu-rock: https://github.com/Automattic/woocommerce-payments/pull/7140#issuecomment-1734071600

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2580,18 +2586,24 @@ public function get_icon_url() {
...
+	 *
+	 * @throws Exception
 	 */
 	public function update_account( $account_settings ) {
 		if ( empty( $account_settings ) ) {
 			return;
 		}
 
-		$error_message = $this->account->update_stripe_account( $account_settings );
+		$request = $this->account->update_stripe_account( $account_settings );
```
deepakpathania: Just a suggestion but I feel since we're anyways catching the exception in the function below, we could potentially return a consistent response in both success and failure scenarios. Something like:
```
{
    "updated": "false",
    "error_message": "<// $e->getMessage() >" 
}
```
And then in the calling function, check for the `boolean` to decide what to log. Also makes the naming convention a lot simpler so instead of `request`, it can be `stripe_account_updation_response` which can then handle both cases.

In either case the naming does seem off for this, so that can be improved regardless.

deepakpathania (commented):
Left a suggestion but you don't necessarily have to accept it, take what you may from it.

PR #7143, opened 2023-09-06
Store the customer currencies data as an option

### LINKED ISSUE #7000: Investigate and improve the performance of `get_all_customer_currencies`
From https://github.com/Automattic/woocommerce-payments/pull/6902#discussion_r1294363972

Two options to approach this:

Option 1:
- Check if this function is even needed. It appears that the only place it's used in, `Analytics::register_customer_currencies`, could just use the currency setting. If there is a currency with no orders, then it would just show empty graphs. Worth investigating further in the context of analytics, I might be wrong here.

Option 2:
- Store a list of currencies as an option (`Database_Cache` might not be needed)
  - Need to calculate it first for the pre-existing orders, might be a matter of a one-off run of the existing SQL if the option is not there
- When a new order is placed, check if its currency is in the list
- If not, then add to the list and save
  - There's a small possibility of multiple orders placed in parallel overwriting each other, but once the currency is in the list, it won't go away, so other order currencies can be added. In time, the list will contain all currencies and be updated very rarely if at all.

Also a couple of notes regarding the existing implementation:
- A comment in the method mentions caching for 1h, but the code `time() - ( 5 * MINUTE_IN_SECONDS )` appears to set it to 5 minutes.
- The TTL can be set for this key inside the `Database_Cache` class and we wouldn't have to have the TTL check in the method or return the `updated` key array. If we decide to remove the `updated` key and flatten the array, we will need to consider backward compatibility.

cc @Automattic/sigma

eduardoumpierre:
Fixes #7000

This PR aims to improve the `register_customer_currencies` performance by replacing the expensive calculation with an option that is updated on every new order. The expensive `get_all_customer_currencies` call will be called only the first time, and then its value will be used to set an option.

#### Benchmark

The benchmark was done using `microtime()` to calculate the start and end times of the function execution.

##### Benchmark with 205 orders

<table>
    <thead>
        <tr>
            <th>#</th>
            <th>Previous</th>
            <th>Current</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>1</td>
            <td><code>0.001592874526977539μs</code></td>
            <td><code>0.000027179718017578125μs</code></td>
        </tr>
        <tr>
            <td>2</td>
            <td><code>0.0019979476928710938μs</code></td>
            <td><code>0.000025987625122070312μs</code></td>
        </tr>
       <tr>
            <td>3</td>
            <td><code>0.0030279159545898438μs</code></td>
            <td><code>0.000030040740966796875μs</code></td>
        </tr>
       <tr>
            <td>4</td>
            <td><code>0.0017621517181396484μs</code></td>
            <td><code>0.000024080276489257812μs</code></td>
        </tr>
       <tr>
            <td>5</td>
            <td><code>0.00202178955078125μs</code></td>
            <td><code>0.000032901763916015625μs</code></td>
        </tr>
        <tr>
            <td>Avg</td>
            <td><code>0.002080535888672μs</code></td>
            <td><code>0.000028038024902μs</code></td>
        </tr>
        <tr>
            <td>Diff</td>
            <td colspan="2">~74x speedup</td>
        </tr>
    </tbody>
</table>

##### Benchmark with 5723 orders

<table>
    <thead>
        <tr>
            <th>#</th>
            <th>Previous</th>
            <th>Current</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>1</td>
            <td><code>0.5904631614685059μs</code></td>
            <td><code>0.00003600120544433594μs</code></td>
        </tr>
        <tr>
            <td>2</td>
            <td><code>0.6487610340118408μs</code></td>
            <td><code>0.00003600120544433594μs</code></td>
        </tr>
       <tr>
            <td>3</td>
            <td><code>0.8029019832611084μs</code></td>
            <td><code>0.000032901763916015625μs</code></td>
        </tr>
       <tr>
            <td>4</td>
            <td><code>0.6125071048736572μs</code></td>
            <td><code>0.00003409385681152344μs</code></td>
        </tr>
       <tr>
            <td>5</td>
            <td><code>0.5736191272735596μs</code></td>
            <td><code>0.000031948089599609375μs</code></td>
        </tr>
        <tr>
            <td>Avg</td>
            <td><code>0.642</code></td>
            <td><code>0.00003416</code></td>
        </tr>
        <tr>
            <td>Diff</td>
            <td colspan="2">~18,793x speedup</td>
        </tr>
    </tbody>
</table>

#### Testing instructions

* Checkout to this branch
* Go to the analytics page `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Forders`
  * The new option will be set on this page load
* After the page loads, open a new tab on PhpMyAdmin `http://localhost:8083/` and run the following query:
```sql
SELECT * FROM `wp_options` WHERE `option_name` LIKE "wcpay_multi_currency_stored_customer_currencies";
```
* Go to the Multi-currency settings page `/wp-admin/admin.php?page=wc-settings&tab=wcpay_multi_currency` and enable a new currency
* Go to your shop and perform the checkout using the currency you've just added
* Go to the PhpMyAdmin tab and refresh the query
* Check the array on `option_value`, it should contain the currency code
* Go to the analytics page `/wp-admin/admin.php?page=wc-admin&path=%2Fanalytics%2Forders`
* Check the "Customer currency" select – it should also contain the currency code option

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7143` or branch name `update/7000-improve-analytics-call-performance` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/7000-improve-analytics-call-performance) :rocket:

---

Build info: 

- Latest commit: 3f37c4dc5f578523b39ed8827e9231d635b73d0c
- Build time: 2023-09-28 18:32:24 UTC

includes/core/server/request/class-create-intention.php
```diff
@@ -194,6 +194,8 @@ public function set_shipping( array $shipping ) {
 	 * @return mixed           Either the same response, or the correct object.
 	 */
 	public function format_response( $response ) {
+		$this->api_client->update_customer_currencies( $response );
```
eduardoumpierre: Should we hook this into a `woocommerce_thankyou` action, for example?

eduardoumpierre:
I'll add additional unit tests for the `update_customer_currencies` method, but it would be good to receive some feedback while I work on it, since I'll be back next Monday.

includes/core/server/request/class-create-intention.php
marcinbot: We should hook somewhere else, because that function definitely doesn't belong in the API client. We should update the currency list once order is placed, maybe something like `woocommerce_checkout_order_created` action? But that only seems to fire on the checkout. So maybe `woocommerce_payment_complete`?

includes/multi-currency/Analytics.php
```diff
@@ -115,11 +114,18 @@ public function register_admin_scripts() {
 	 * @return void
 	 */
 	public function register_customer_currencies() {
-		$currencies           = $this->multi_currency->get_all_customer_currencies();
+		$currencies = get_option( MultiCurrency::CUSTOMER_CURRENCIES_KEY, false );
+
+		if ( ! $currencies ) {
+			$currencies = $this->multi_currency->get_all_customer_currencies();
+			update_option( MultiCurrency::CUSTOMER_CURRENCIES_KEY, $currencies );
+		}
```
marcinbot: I think it might be better to handle this logic inside `get_all_customer_currencies()`. If the option exists, return it, otherwise backfill it with all currencies.

eduardoumpierre: It does make sense. I updated it on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/5dd5cd8b65a407c94d65a17edc989be87eee8f0f. Thanks for the suggestion!

includes/core/server/request/class-create-intention.php
eduardoumpierre: I tested this action with different flows and found that it only works when the payment is completed (ha!), despite the documentation stating that it should also be triggered by the `on hold`, `pending`, `failed`, and `cancelled` statuses. I tested the capture later flow and the action was not triggered.

Since there is an advanced filter on the analytics page that allows the merchant to filter by any status, I believe we should update the customer currencies option even if the order is created manually.

<table>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
<tr>
<td>
<i>Advanced filters</i>
</td>
<td>
<i>Available statuses</i>
</td>
</tr>
</table>

After testing several available actions, I found that `woocommerce_order_status_changed` made the most sense to me. This is because there are multiple entry points to the payment flow, and the action also supports manually created orders.

<table>
<tr>
<td>

</td>
</tr>
<tr>
<td>
<a href="https://woocommerce.com/document/managing-orders/#visual-diagram-to-illustrate-order-statuses"><i>Visual Diagram to Illustrate Order Statuses</i></a>
</td>
</tr>
</table>

It was updated on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/1b5bf554a57bf08b336d0fb62168143fbb15a207. Thanks for the suggestion, and let me know your thoughts on this!

includes/multi-currency/MultiCurrency.php
```diff
@@ -544,6 +548,39 @@ public function update_single_currency_settings( string $currency_code, string $
...
+		if ( ! $order ) {
+			return;
+		}
+
+		$currency   = $order->get_currency();
+		$currencies = get_option( self::CUSTOMER_CURRENCIES_KEY );
+
+		// Skip if the currency is not a string or if the option is invalid.
+		if ( ! is_string( $currency ) || ! is_array( $currencies ) ) {
+			return;
```
tpaksu: I'm thinking about the initial migration, let's assume I upgraded to the version that contains this change for the first time, I won't have the option set. `get_option` will return `false` here, right? So the customer currency won't be updated unless I go to the Analytics section to have it updated. What do you think about splitting the condition into two parts, one checking and bailing if the order currency isn't valid, and the second part resetting the currency option when the option doesn't exist, or isn't valid?

tpaksu: Ok, missed the line in `register_customer_currencies` method on the MC class, the initialization will fill in that data. It was folded by GH, needed to unfold it :)

tpaksu: Nope, it was on the `Analytics` initialization. So the question stays valid :)

marcinbot: Yeah, a few things can go wrong here. I think if we used `self::get_all_customer_currencies` here instead of `get_option`, it might be slow on the very first order, but "cached" forever later. Sounds like a better way I think. It would also reduce code duplication.

includes/multi-currency/MultiCurrency.php
```diff
@@ -1451,32 +1488,24 @@ public function is_multi_currency_settings_page(): bool {
...
-			return $data['currencies'];
+		if ( self::is_customer_currencies_data_valid( $currencies ) ) {
+			return $currencies;
+		}
+
+		if ( class_exists( 'Automattic\WooCommerce\Utilities\OrderUtil' ) &&
+				\Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled() ) {
+			$currencies = $wpdb->get_col( "SELECT DISTINCT(currency) FROM {$wpdb->prefix}wc_orders" );
+		} else {
+			$currencies = $wpdb->get_col( "SELECT DISTINCT(meta_value) FROM {$wpdb->postmeta} WHERE meta_key = '_order_currency'" );
```
marcinbot: Might be worth including the fix from #6902. Or Sigma could take charge and merge that PR and then resolve the conflicts, since it's been approved for a while. I think it's worth shipping the two performance improvements together.

includes/multi-currency/MultiCurrency.php
```diff
@@ -544,6 +548,39 @@ public function update_single_currency_settings( string $currency_code, string $
...
+
+		$currency   = $order->get_currency();
+		$currencies = get_option( self::CUSTOMER_CURRENCIES_KEY );
+
+		// Skip if the currency is not a string or if the option is invalid.
+		if ( ! is_string( $currency ) || ! is_array( $currencies ) ) {
+			return;
+		}
+
+		$currency = strtoupper( $currency );
```
marcinbot: Another small issue here: we have no guarantee that the contents of `currencies` will be all uppercase. So if we were to use the getter, we could ensure that by casting it all to uppercase before returning, so it all follows the same standard.

includes/multi-currency/MultiCurrency.php
tpaksu: Added 6902 to merge queue now.

includes/multi-currency/MultiCurrency.php
eduardoumpierre: > So the customer currency won't be updated unless I go to the Analytics section to have it updated.

The migration will happen on any page of the admin dashboard if MultiCurrency is enabled, not necessarily on the Analytics page, and most probably will happen right after the plugin update process. This is the path that will lead to this behavior:

1. The Init method is called on the [MultiCurrency constructor](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/MultiCurrency.php#L231)
2. The Analytics class [gets instantiated](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/MultiCurrency.php#L272)
3. The init method is called on the [Analytics constructor](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/Analytics.php#L53)
4. The `register_customer_currencies` method is called [on the admin dashboard](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/multi-currency/Analytics.php#L64)

This can be tested by removing the `wcpay_multi_currency_customer_currencies` option from the local database and then accessing any page on the admin dashboard – it should be generated again.

includes/multi-currency/MultiCurrency.php
eduardoumpierre: Nice call! I updated the code to guarantee that the content of `currencies` is always uppercase on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/419ff2cbf599d1dd26eea966f9b18469f7a95a41.

includes/multi-currency/MultiCurrency.php
eduardoumpierre: As Marcin commented, after having a slow first setup (for big merchants), it will be faster on the following page interactions.

Do you still have concerns after my comment above?

tpaksu: That makes sense, thanks for the explanation! The customer currencies will be updated whenever the page loads with initializing the MultiCurrency class then. Gotcha.

tpaksu: Oh, sorry, it is to be updated when the user is admin and has the rights to `manage_woocommerce`. Since we only use this data for the Analytics section, it won't matter, but when a buyer purchases something on the frontend, until the admin is logged in, the option will not get populated. And it will be a problem when someone tries to use that in the frontend on a fresh site. Has a low chance to happen though.

https://github.com/Automattic/woocommerce-payments/blob/5f75b8fd87ad8d34d350a7d7c0c4774fb60d6efe/includes/multi-currency/Analytics.php#L62-L65

Do you think we should cover all possibilities by replacing the `get_option` call with `self::get_all_customer_currencies` in https://github.com/Automattic/woocommerce-payments/blob/419ff2cbf599d1dd26eea966f9b18469f7a95a41/includes/multi-currency/MultiCurrency.php#L566 ? I think Marcin was also suggesting that.

eduardoumpierre: That makes sense. I updated the code on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/b49072944d048424c7f34c236471096540fcfb3f. Thanks for the suggestion @tpaksu and @marcinbot!

includes/multi-currency/MultiCurrency.php
```diff
@@ -1446,11 +1484,11 @@ public function is_multi_currency_settings_page(): bool {
 	}
 
 	/**
-	 * Function used to compute the customer used currencies, used as internal callable for get_all_customer_currencies function.
+	 * Computes the customer used currencies.
 	 *
 	 * @return array
 	 */
-	public function callable_get_customer_currencies() {
+	public function get_customer_currencies() {
```
marcinbot: This could be a leftover from the merge, but is this function used anywhere outside of `get_all_customer_currencies`? If not, then it should perhaps become public. Also, we probably shouldn't expose a way to bypass the cache as public.

Also, if `get_all_customer_currencies` is the only place it is used in, then maybe we should roll the two functions into one. Right now the names are very similar and this function still gets all customer currencies.

includes/multi-currency/MultiCurrency.php
```diff
@@ -544,6 +548,40 @@ public function update_single_currency_settings( string $currency_code, string $
...
+		$currency   = $order->get_currency();
+		$currencies = self::get_all_customer_currencies();
+
+		// Skip if the currency is not a string or if the option is invalid.
+		if ( ! is_string( $currency ) || ! is_array( $currencies ) ) {
+			return;
+		}
+
+		$currency   = strtoupper( $currency );
+		$currencies = array_map( 'strtoupper', $currencies );
```
marcinbot: I think this map call would belong more in `get_all_customer_currencies` for consistency in future usages.

includes/multi-currency/MultiCurrency.php
eduardoumpierre: Great point, I also wasn't happy with the naming of those two functions. I merged both on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/872060c6f4aa096fc0f8fa6ae7270c95c07c0124. Thanks for the suggestion!

includes/multi-currency/MultiCurrency.php
eduardoumpierre: It does make sense. I updated the code to transform the array before saving the option on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/c947e6ce5bcd008aa3da77f2cdf3d322acac10af. Thanks for the suggestion!

includes/multi-currency/MultiCurrency.php
```diff
@@ -544,6 +548,39 @@ public function update_single_currency_settings( string $currency_code, string $
...
+
+		if ( ! $order ) {
+			return;
+		}
+
+		$currency   = $order->get_currency();
+		$currencies = self::get_all_customer_currencies();
+
+		// Skip if the currency is not a string or if the option is invalid.
+		if ( ! is_string( $currency ) || ! is_array( $currencies ) ) {
```
tpaksu: The `$currencies` variable doesn't have the chance to be anything else than an array now after switching to `self::get_all_customer_currencies()`, right? Maybe we should remove it. It being an empty array is also fine for the script to continue as far as I can see.

eduardoumpierre: I agree with you. It also appears unnecessary to include the check for `! is_string( $currency )` since orders cannot be created with a non-string currency. Perhaps we should remove the check entirely?

eduardoumpierre: I updated the code on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/71ce3fd515e45c3d8617138c916ec6d79f875c51. Thanks for the suggestion and let me know your thoughts on removing both checks.

includes/multi-currency/MultiCurrency.php
marcinbot: I meant more casting to upper before returning from `get_all_customer_currencies`, after a read, rather than before save 😅 So that we're backward-compatible in case there are lower-case currencies already saved.

includes/multi-currency/MultiCurrency.php
```diff
@@ -27,9 +27,10 @@
  */
 class MultiCurrency {
 
-	const CURRENCY_SESSION_KEY = 'wcpay_currency';
-	const CURRENCY_META_KEY    = 'wcpay_currency';
-	const FILTER_PREFIX        = 'wcpay_multi_currency_';
+	const CURRENCY_SESSION_KEY    = 'wcpay_currency';
+	const CURRENCY_META_KEY       = 'wcpay_currency';
+	const FILTER_PREFIX           = 'wcpay_multi_currency_';
+	const CUSTOMER_CURRENCIES_KEY = self::FILTER_PREFIX . 'customer_currencies';
```
marcinbot: Just realized that this could be a problem. DB cache also uses options. So there is a high chance that there will be an existing option under the same key with a different format (currencies array being under `data`). So I think it might be safe to slightly rename this key,

includes/multi-currency/MultiCurrency.php
eduardoumpierre: I thought about saving it directly as uppercase because the option wouldn't exist prior to this code execution, so it would always get the uppercase values. I do agree that it's better to guarantee that we're always dealing with uppercase currencies. The code was updated on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/862dcbeacd61d2ef2e46e506e6241a3f725d36fe.

includes/multi-currency/MultiCurrency.php
eduardoumpierre: Nice catch! I updated the key to be `wcpay_multi_currency_stored_customer_currencies` on https://github.com/Automattic/woocommerce-payments/pull/7143/commits/3d101abf27e55361640ffa933e4909f25b4e9112.

tpaksu (approved):
Code looks good, is understandable and has no problems, I had a small problem when running the SQL to get the updated field, then realized the option name was `wcpay_multi_currency_stored_customer_currencies`, not the one written in the query, now it's all good :) Thanks for working on this. LGTM!

marcinbot (approved):
:shipit:

PR #7146, opened 2023-09-07
Fix: Open two checkout tabs, and use 3DS cards with non-UPE flow 

### LINKED ISSUE #5678: Duplicate Payments: Open two checkout tabs, and use 3DS cards with non-UPE flow
### Describe the bug

Duplicate payments for a single order happens when shoppers open two checkout tabs, use 3DS cards, and confirm 3DS modals in both tabs in between.

### To Reproduce

- Use non-UPE flow or a saved card in UPE flows (as it eventually routes to non-UPE flow)
- Add one or some products into cart.
- Go to the checkout page (tab 1), and make a payment with a 3DS card - 4000 0027 6000 3184, which always triggers authentication (See more https://woocommerce.com/document/woocommerce-payments/testing-and-troubleshooting/testing/#3ds-test-cards)
- See the modal to confirm the payment in tab 1. **BUT STOP, do not confirm**.
- Open the checkout page on another tab (tab 2), and use the same card for the checkout.
- See the modal to confirm the payment in tab 2, and confirm it. 
- Go to tab 1, confirm the modal. 

### Actual behavior

- Check the latest order, there will be two payments (payment intents) in this order. 
- Confirm this in the Stripe Dashboard too.

### Expected behavior

Only one payment should be processed, the other one should be void. 

### Additional context

See `Order 1564354 - date: 2023-02-07` in this spike with a real merchant site - repo: woocommerce-payments-meta/issues/75#issuecomment-1455511181

htdat:
Fixes #5678

In this PR, I am approaching to cancel the attached payment intent (if it exists) rather than re-use it as originally planned in the PR. 

The main benefit of this approach is that it's simple, and does not need to involve any change in the server. 

**For the approach of re-using the attached payment intent**, it requires to two requests: one to update intent, and another one to [confirm intent](https://stripe.com/docs/api/payment_intents/confirm). To approach this, there are two ways: 

- Send two requests on the client: update, and then confirm it. This requires a new endpoint for confirming intent. 
- Send on request on the client to a new endpoint such as `update_and_confirm` on the server. On the server, it will update, and then confirm the attached intent.

**Anyway, re-using the attached payment intent might make code on both server and client more complicated** as it requires (1) more branching out as the `confirm` and `update` intent input data are not the same, (2) a change on the server, at least one new endpoint. So I'd like to approach the one in this PR. 

### Testing instructions

- Follow up instructions on #5678. 
- Confirm: only one payment intent is successful, the other one is cancelled. 
- Confirm: on the order under WP Admin, a note like this exists: `Payment cancelled, and removed from order`.

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7146` or branch name `fix/5678-two-checkout-tabs-3DS-card` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/5678-two-checkout-tabs-3DS-card) :rocket:

---

Build info: 

- Latest commit: 179e9a2b1c469ad8475ae10f505fee14a64a12e4
- Build time: 2023-10-03 06:30:34 UTC

includes/class-duplicate-payment-prevention-service.php
```diff
@@ -226,4 +228,42 @@ public function clear_session_processing_order_after_landing_order_received_page
...
+	 * @param  WC_Order $order Order object.
+	 *
+	 * @return void
+	 */
+	public function remove_and_cancel_attached_intent_id( WC_Order $order ): void {
+		$intent_id_attached_to_order = $this->order_service->get_intent_id_for_order( $order );
+		if ( ! empty( $intent_id_attached_to_order ) ) {
+			try {
+				$request = Cancel_Intention::create( $intent_id_attached_to_order );
+				$request->send( 'wcpay_cancel_intent_request', $order );
```
zmaglica, line 244: I think we changed the signature here. Look into `apply_hook` and `set_hook_args`

includes/class-wc-payments-order-service.php
```diff
@@ -458,6 +458,18 @@ public function set_intent_id_for_order( $order, $intent_id ) {
...
 	}
 
+	/**
+	 * Remove intent_id for $order.
+	 *
+	 * @param WC_Order $order The order object.
+	 *
+	 * @return void
+	 */
+	public function remove_intent_id_for_order( WC_Order $order ) {
```
zmaglica, line 468: Can we maybe add tests for this function?

includes/class-duplicate-payment-prevention-service.php
```diff
@@ -226,4 +228,42 @@ public function clear_session_processing_order_after_landing_order_received_page
...
 	}
+
+	/**
+	 * Remove and cancel attached intent ID to the order if exists.
+	 *
+	 * @param  WC_Order $order Order object.
+	 *
+	 * @return void
+	 */
+	public function remove_and_cancel_attached_intent_id( WC_Order $order ): void {
```
zmaglica, line 239: Nitpick: It would be nice to cover this function with tests.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1224,6 +1224,8 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 			}
 
 			if ( empty( $intent ) ) {
+				$this->duplicate_payment_prevention_service->remove_and_cancel_attached_intent_id( $order );
```
zmaglica, line 1227: Since this is the part of the checkout flow, I think it would be wise to cover this flow with the tests. 
On another hand, did we check if attached payment intent is already captured?  In that case, wouldn't this cause an exception?

htdat:
@zmaglica thanks for your feedback. 

> @RadoslavGeorgiev and I discussed yesterday and Rado proposed another idea:
> 
> > We already have the intent ID, and can load it. Within the DPPS, could we load it, and if the intent requires action, immediately show the confirmation modal for the existing intent, without creating another one? This way, it can be confirmed in either tab, and the second confirmation would fail.
> 
> My initial testing to force displaying the confirmation modal should work.
> 
> However, then Rado and I also discussed and both agreed that: whatever my approach or Rado’s suggestion, both feels hacky and not what we’d like to achieve ultimately. I also add another point that: any additional logic we’re adding to current gateway classes (or their dependencies) creates more work for us to port it to the current RPP project. Last but not least, any more effort here will reduce our focus on the RPP project.
> 
> With all of that, Rado and I have the consensus that we put a hold on this issue, and leave it to handle in the later phase of RPP. 
> cc @naman03malhotra 
> 
> -- Ref: p1697170318055929/1697094668.359039-slack-C01BPL3ALGP

naman03malhotra:
Thanks @htdat for adding the context and spending time on this issue. I agree with the consensus you folks have reached to, can you please add the right label to it and maybe close the PR and issue for now?

naman03malhotra (commented):
Thanks @htdat for adding the context and spending time on this issue. I agree with the consensus you folks have reached to, can you please add the right label to it and maybe close the PR and issue for now?

Issue #7110, opened 2023-09-01
Call to a member function delete_notes() on null


PrinceOfAbyss:
### Describe the bug
I tried to deactivate the plugin due to another issue, but during de-activation, the e-shop goes down with `PHP Fatal error:  Uncaught Error: Call to a member function delete_notes() on null in /wp-content/plugins/woocommerce-payments/includes/class-wc-payments.php:1386`

I had to manually rename the plugin's folder to bring back the e-shop online.

### To Reproduce
I suppose you just have to try de-activating the eshop?

### Actual behavior
N/A

### Screenshots
N/A

### Expected behavior
It should de-activate the plugin normally I suppose...

### Desktop (please complete the following information):
Not relevant

### Smartphone (please complete the following information):
Not relevant

### Additional context
N/A

zmaglica:
Greetings,

Thank you for reporting the issue. I've reviewed the function name in question and its call locations. Based on the current version of the plugin, the error you encountered seems unusual, as the file where this service is instantiated is already included. Could you kindly specify which version of the WooCommerce Payments plugin you are using?

For additional context, I also conducted tests by deactivating the plugin on my end and was unable to reproduce the error you mentioned.

PrinceOfAbyss:
Thanks a lot for your prompt reply. [Here](https://www.youtube.com/watch?v=zBThzMHESyQ) is a short video showing the error in action. About the e-shops versions:

WooCommerce version: 		7.7.0
WordPress version: 		6.2.2
PHP version: 		8.0.28

The error log that gets produced when the error occurs reads the following:

```
[04-Sep-2023 09:30:20 UTC] PHP Fatal error:  Uncaught Error: Call to a member function delete_notes() on null in /public_html/wp-content/plugins/woocommerce-payments/includes/class-wc-payments.php:1386
Stack trace:
#0 /public_html/wp-content/plugins/woocommerce-payments/woocommerce-payments.php(57): WC_Payments::remove_woo_admin_notes()
#1 /public_html/wp-includes/class-wp-hook.php(308): wcpay_deactivated()
#2 /public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()
#3 /public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
#4 /public_html/wp-admin/includes/plugin.php(814): do_action()
#5 /public_html/wp-admin/plugins.php(209): deactivate_plugins()
#6 {main}
[...]

PR #7148, opened 2023-09-07
Added valid dependencies check on plugin update hook

### LINKED ISSUE #7147: Prevent clearing cache during the plugin update when the plugin doesn't meet required dependencies.
### Describe the bug

Every time there is a plugin update, the minimum required versions for WooCommerce and WordPress are raised in line with the L-2 policy. This policy can occasionally lead to the plugin being disabled if the minimum requirements are not met. However, if the account cache remains undeleted, the plugin will not be disabled immediately. Users instead see error notifications prompting them to update the dependent plugins to the required versions.

The issue arises after the plugin update triggers this [hook](https://github.com/Automattic/woocommerce-payments/blob/12ac12d4cfacabc8edc9a45ba362c670fa3a9397/includes/class-wc-payments.php#L1293). This hook updates the plugin version in the options table and deletes the account cache, resulting in the plugin being disabled if the minimum requirements are not fulfilled, affecting other hooks and actions reliant on WCPay services that fail to initialize, thereby causing functionality issues.

### To Reproduce

1. Downgrade woocommerce to the unsupported version. You can use a rollback `[plugin](https://wordpress.org/plugins/wp-rollback/)`.
2. Downgrade plugin version in database by setting this option name `woocommerce_woocommerce_payments_version` to` 1.0.0` for example (`update_option( 'woocommerce_woocommerce_payments_version', '1.0.0' );`)
3. Refresh the page and make sure that the `woocommerce_woocommerce_payments_version` has the latest plugin version as a value
4. Disable WCPayments plugin
5. See error about certain variable or service is null.

### Actual behavior
When plugin is updated, the cache should be deleted ONLY if minimum requirements are set.

zmaglica:
Fixes #7147 #7110 

This PR enhances security in the post-plugin update hook by preventing the deletion or purging of the cache, as well as avoiding the increment of the plugin version in the options table, if the minimum dependencies are not satisfied.

#### Testing instructions

1. Downgrade WooCommerce to the unsupported version. You can use a rollback `[plugin](https://wordpress.org/plugins/wp-rollback/)`.
2. Downgrade plugin version in database by setting this option name `woocommerce_woocommerce_payments_version` to` 1.0.0` for example (`update_option( 'woocommerce_woocommerce_payments_version', '1.0.0' );`)
3. Refresh the page and make sure that the `woocommerce_woocommerce_payments_version` has the latest plugin version as a value
4. Disable WCPayments plugin
5. No errors should be shown.

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -1291,7 +1291,15 @@ class WC_Payments {
 	 * Handles upgrade routines.
 	 */
 	public static function install_actions() {
-		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) ) {
+		$dependency_service = self::$dependency_service;
+
+		// Make sure we have correct class.
+		if ( ! $dependency_service instanceof WC_Payments_Dependency_Service ) {
+			$dependency_service = new WC_Payments_Dependency_Service();
+		}
+
+		// Don't trigger any update plugin hook and don't bump current plugin version if there is a dependencies issues.
+		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) && true === $dependency_service->has_valid_dependencies() ) {
 			do_action( 'woocommerce_woocommerce_payments_updated' );
 			self::update_plugin_version();
 		}
@@ -1378,7 +1386,9 @@ class WC_Payments {
 	 */
 	public static function remove_woo_admin_notes() {
 		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '4.4.0', '>=' ) ) {
-			self::$remote_note_service->delete_notes();
+			if ( class_exists( 'WC_Payments_Remote_Note_Service' ) && self::$remote_note_service instanceof WC_Payments_Remote_Note_Service ) {
+				self::$remote_note_service->delete_notes();
+			}
 			require_once WCPAY_ABSPATH . 'includes/notes/class-wc-payments-notes-set-up-refund-policy.php';
 			require_once WCPAY_ABSPATH . 'includes/notes/class-wc-payments-notes-qualitative-feedback.php';
 			WC_Payments_Notes_Qualitative_Feedback::possibly_delete_note();
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -1299,7 +1299,7 @@ class WC_Payments {
 		}
 
 		// Don't trigger any update plugin hook and don't bump current plugin version if there is a dependencies issues.
-		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) && true === $dependency_service->has_valid_dependencies() ) {
+		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) && empty( $dependency_service->get_invalid_dependencies() ) ) {
 			do_action( 'woocommerce_woocommerce_payments_updated' );
 			self::update_plugin_version();
 		}
--- a/changelog/fix-add-dependencies-check-on-plugin-update-hook
+++ b/changelog/fix-add-dependencies-check-on-plugin-update-hook
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Prevent cache deletion during plugin update if minimum dependencies are not met.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7148` or branch name `fix/add-dependencies-check-on-plugin-update-hook` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/add-dependencies-check-on-plugin-update-hook) :rocket:

---

Build info: 

- Latest commit: 0bf403f8d261947caed7a7bf9db4a583048f1a15
- Build time: 2023-09-08 09:33:20 UTC

includes/class-wc-payments.php
```diff
@@ -1291,7 +1291,15 @@ public static function register_checkout_gateway( $payment_method_registry ) {
 	 * Handles upgrade routines.
 	 */
 	public static function install_actions() {
-		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) ) {
+		$dependency_service = self::$dependency_service;
+
+		// Make sure we have correct class.
+		if ( ! $dependency_service instanceof WC_Payments_Dependency_Service ) {
+			$dependency_service = new WC_Payments_Dependency_Service();
+		}
```
jrodger, line 1299: I know we're just trying to be defensive here, but I don't think it's possible for this to be null here?

What do you think of the trade-off between the complexity here defending against something that might change in the future?

includes/class-wc-payments.php
```diff
@@ -1378,7 +1386,9 @@ function wcpay_show_old_woocommerce_for_norway_notice() {
 	 */
 	public static function remove_woo_admin_notes() {
 		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '4.4.0', '>=' ) ) {
-			self::$remote_note_service->delete_notes();
+			if ( class_exists( 'WC_Payments_Remote_Note_Service' ) && self::$remote_note_service instanceof WC_Payments_Remote_Note_Service ) {
```
jrodger, line 1389: We don't want developers thinking they need to do this everywhere they call a method. A comment here explaining why this is a special case would be useful.

I know it's small change, but it's technically unrelated to the other change as well right? Would this be better in its own PR?

includes/class-wc-payments.php
```diff
@@ -1291,7 +1291,15 @@ public static function register_checkout_gateway( $payment_method_registry ) {
...
+		$dependency_service = self::$dependency_service;
+
+		// Make sure we have correct class.
+		if ( ! $dependency_service instanceof WC_Payments_Dependency_Service ) {
+			$dependency_service = new WC_Payments_Dependency_Service();
+		}
+
+		// Don't trigger any update plugin hook and don't bump current plugin version if there is a dependencies issues.
+		if ( version_compare( WCPAY_VERSION_NUMBER, get_option( 'woocommerce_woocommerce_payments_version' ), '>' ) && true === $dependency_service->has_valid_dependencies() ) {
 			do_action( 'woocommerce_woocommerce_payments_updated' );
```
jrodger, line 1303: Updating the comment to explain why we don't want to run the update plugin hook would be good (i.e. it clears the cache and messes with our ability to detect active merchants in our dependency checks)

Do we think it's safe to skip all the updated actions? I count 10 different things hooked into this at the moment. Are there going to be any side effects to not updating the plugin version here?

I wonder if there's a way for us to be more specific and just skip flushing the cache, but even then I wonder if that might break upgrades for some merchants? Might there be a better / different way of detecting active merchants we could look into instead of relying on the cache being there perhaps?

jrodger (commented):
I haven't tested the code live, I might have to leave that to someone else.

Some comments inline, I'm a bit uneasy with skipping all the upgrade actions and potentially leaving the plugin in a strange state. But then again, attempting to run the plugin on "unsupported" WP and WC versions is also a bit of a strange state. Is it worth considering other approaches if we can't come up with something solid in this PR?

includes/class-wc-payments.php
htdat: I wonder if the additional check `true === $dependency_service->has_valid_dependencies()` really helps here. 

`install_actions` is only called for the `init` hook in `WC_Payments::init`. However, as you see below, to run `install_actions`, `true === $dependency_service->has_valid_dependencies()` has already passed.

https://github.com/Automattic/woocommerce-payments/blob/83f6905673a9891821bac870367a46560a7ebdc3/includes/class-wc-payments.php#L303-L310

includes/class-wc-payments.php
zmaglica: Upon this point, I was scared and afraid of random nulls 😄  But you are correct. The reason why I choose this approach comes from this [talk](https://www.youtube.com/watch?v=HwkhpRUYIdc)

includes/class-wc-payments.php
zmaglica: > I wonder if the additional check `true === $dependency_service->has_valid_dependencies()` really helps here.
> 
> `install_actions` is only called for the `init` hook in `WC_Payments::init`. However, as you see below, to run `install_actions`, `true === $dependency_service->has_valid_dependencies()` has already passed.
> 
> https://github.com/Automattic/woocommerce-payments/blob/83f6905673a9891821bac870367a46560a7ebdc3/includes/class-wc-payments.php#L303-L310

Before the plugin update, this will be 'passed' `false === self::$dependency_service->has_valid_dependencies()` since the account cache is already there and `install_actions` will be executed so this will be executed right after the plugin update which we want to prevent, but I made an error here where I am doing the same check above. I have to check minimum dependencies without account cache flag.

zmaglica: > Updating the comment to explain why we don't want to run the update plugin hook would be good (i.e. it clears the cache and messes with our ability to detect active merchants in our dependency checks)
> 
> Do we think it's safe to skip all the updated actions? I count 10 different things hooked into this at the moment. Are there going to be any side effects to not updating the plugin version here?
> 
> I wonder if there's a way for us to be more specific and just skip flushing the cache, but even then I wonder if that might break upgrades for some merchants? Might there be a better / different way of detecting active merchants we could look into instead of relying on the cache being there perhaps?

A doable strategy is to meticulously inspect all actions, and perform null checks on important variables and classes, focusing specifically on actions we anticipate might be executed outside of the WooCommerce Payments context. Currently, we can pinpoint these actions, but projecting forward, this task may become less feasible.

The goal of this pull request is to avoid triggering plugin update-related hooks in the event of missing dependencies, which will keep the plugin active. Once these dependencies are reinstated, the plugin update hook will become operational during the subsequent request, thereby executing the necessary procedures.

From a different angle, we could think about turning off the update or auto-update features to avoid problems like this if some dependencies are missing. Basically, if something is missing, don't allow the plugin to update.

zmaglica:
I will close this PR and initiate a P2 post to discuss how we should address such situations. This issue demands a more in-depth analysis and collaboration from multiple teams.

PR #7153, opened 2023-09-07
Fix invalid value error for paymentMethodMessaging in product details page

### LINKED ISSUE #6819: [6.2.0] Invalid value for paymentMethodMessaging.update() when attempting to configure a variable product with the Flatsome theme active
### Describe the bug
When using [the Flatsome theme](https://flatsome3.uxthemes.com/) while WooPayments 6.2.0 is active, variable products cannot be added to the cart. Doing so produces an error notice when adding the product to the cart:

```
Please select some product options before adding this product to your cart.
```

After selecting all available variables—but before attempting to add the product to the cart—a JavaScript error is thrown in the console:

```
Uncaught IntegrationError: Invalid value for paymentMethodMessaging.update(): amount should be a positive amount in the currency's subunit. You specified: NaN.
    at ne (?ver=3.0:1:144524)
    at re (?ver=3.0:1:144596)
    at ?ver=3.0:1:186667
    at et (?ver=3.0:1:186693)
    at Object.update (?ver=3.0:1:217483)
    at t.<anonymous> (?ver=3.0:1:277609)
    at t.update (?ver=3.0:1:64872)
    at HTMLDivElement.<anonymous> (??-eJyNkMFqAzEMRH+oWqeHJumh9FOCaqmOUtvaWNqY/fvupgQCIVDQRWjmMaM+QtTqXD2MeUpSLcziR+iqUUvhFhm62DGLeUAzdgunZc4Tt3n4k8aehyJ1ONlLf8Tdke4A3+0qo4BE4AoRm8MFm6CL1v/hYMS5LDcLtKYbm9IUHYgdJdsTP/6IFfZwkBpvC9zSPPEk1ZQZ1h9ITQZYCZCuPb4myRSSYwK+rFkWxGf5eN3u33e7zeZt+wu4d4ZI:3:1530)
    at HTMLDivElement.dispatch (jquery.min.js?ver=3.6.4:2:43184)
    at y.handle (jquery.min.js?ver=3.6.4:2:41168)
```

This error is thrown on selection, and the selection event is not firing as a result.

Reverting to 6.1.1 resolves this issue. Given that reverting resolves the issue and the popularity of Flatsome, I'm opening this issue :)

**Note**: Reverting to Woo 7.8.0 or disabling multi-currency did not resolve the issue. Changing to a different theme does resolve the issue.

### To Reproduce

1. Install/activate Flatsome.
2. Install/activate WooPayments 6.2.0.
3. Create a variable product.
4. Attempt to add the product created in step 3 to the cart.
5. See errors.
6. Revert to WooPayments 6.1.1.
7. Repeat step 4.
8. The product will be successfully added to the cart.

### Actual behavior
While WooPayments 6.2.0 and the Flatsome theme are active, customers cannot add variable products to the cart because the selection event is not firing.

### Screenshots
Markup on 2023-07-19 at 19:56:43

### Expected behavior
WooPayments 6.2.0 + Flatsome allows variable products to be added to the cart without issue.

### Additional context
Discussion: p1689800687243289-slack-C3NCP7ZJ6

Initial reports: 
6556641-zen
6556849-zen
6557129-zen

mgascam:
Fixes #6819 (reopened)

This issue has been reopened. We received reports of errors in the plugin in the product details page after the previous fix was released. This time, errors seem to happen due to merchants customizing this page by removing the quantity input.

- Improved the method of fetching the product quantity and handling cases where the input element might be missing.
- Introduced safeguards against potential undefined values for productVariations and related fields.
- Established 'USD' as the fallback currency in case it's not defined in productVariations.
- Added more strict checks for potentially missing DOM elements to prevent script errors.

#### Testing instructions

To replicate the issue we need to remove the quantity input from the product details page. This can be achieved by several means. We got reports of merchants using a page builder. I was able to replicate the issue locally, using this hook `woocommerce_is_sold_individually`. In the functions.php file of the Storefront theme add the following:

```
function custom_woocommerce_remove_quantity_input() {
    return true;
}
add_filter( 'woocommerce_is_sold_individually', 'custom_woocommerce_remove_quantity_input' );
```
* Enable Affirm or Afterpay in your local store.
* Navigate to the details page of a product with variations.
* Check the Affirm and Afterpay messaging is displayed when the page loads.
* Check there are no errors in the browser console related to `paymentMethodMessaging`.
* Check you are able to select variations with no errors happening.
* Check you are able to add the product to the cart without issues.
* Navigate to the details page of a product **without** variations.
* Check the same steps as before.
* Enable the quantity selector back (you can return false in the `custom_woocommerce_remove_quantity_input`)
* Repeat the same steps to check you are able to add the product to the cart.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7153` or branch name `fix/6819-payment-method-messaging-invalid-value` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/6819-payment-method-messaging-invalid-value) :rocket:

---

Build info: 

- Latest commit: 3df98365f83b00c3a6b434878933adf4f042de7b
- Build time: 2023-09-28 07:55:31 UTC

client/product-details/index.js
```diff
@@ -37,13 +49,23 @@ jQuery( function ( $ ) {
 			event,
 			variation
 		) {
-			const quantity = $( '.quantity input[type=number]' ).val();
-			const variationPrice =
-				productVariations[ variation.variation_id ].amount;
+			const quantityInput = $( '.quantity input[type=number]' );
```
kalessil: I believe we shouldn't engage in updating the element if the input is missing or variation price and currency are not accessible.

client/product-details/index.js
```diff
@@ -7,27 +7,39 @@
...
-			currency: productVariations.base_product.currency,
+			amount: baseProductAmount * quantity,
+			currency: productVariations.base_product?.currency || 'USD',
 		} );
 	};
 
 	$( '.quantity input[type=number]' ).on( 'change', function ( event ) {
-		const newQuantity = event.target.value;
-		const price = productVariations[ productId ].amount;
+		const newQuantity = parseInt( event.target.value, 10 ) || 1;
```
kalessil: Also here: if we couldn't parse new quantity or variation price and currency are not accessible, we shouldn't update the element.

client/product-details/index.js
```diff
@@ -7,27 +7,39 @@
...
+	if ( ! window.wcpayStripeSiteMessaging ) {
+		return;
+	}
 	const bnplPaymentMessageElement = initializeBnplSiteMessaging();
 	const { productVariations } = window.wcpayStripeSiteMessaging;
 	let { productId } = window.wcpayStripeSiteMessaging;
 
 	const resetBnplPaymentMessage = () => {
-		const quantity = $( '.quantity input[type=number]' ).val();
+		const quantityInput = $( '.quantity input[type=number]' );
```
kalessil: Also here: if we couldn't parse new quantity or variation price and currency are not accessible, we shouldn't update the element.

client/product-details/index.js
```diff
@@ -7,27 +7,39 @@
 import { initializeBnplSiteMessaging } from './bnpl-site-messaging';
 
 jQuery( function ( $ ) {
+	// Check for required global variables
```
kalessil: Good one, can we expand a bit what's contained in this variable for readers?

kalessil (changes requested):
Like the direction, but we shouldn't be updating element if we can not reliable calculate price. See my comments.

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/pull/7153/commits/3ad3c2aaf61dcddb961168d7eadfeae460dd7ac9

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/pull/7153/commits/23da3b2d9a1c97f84b4de2c74bcd82774e53d9ef thanks

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/commit/23da3b2d9a1c97f84b4de2c74bcd82774e53d9ef thanks

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/commit/23da3b2d9a1c97f84b4de2c74bcd82774e53d9ef thanks

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
+		return;
+	}
+
+	const { productVariations, productId } = window.wcpayStripeSiteMessaging;
 	const bnplPaymentMessageElement = initializeBnplSiteMessaging();
-	const { productVariations } = window.wcpayStripeSiteMessaging;
-	let { productId } = window.wcpayStripeSiteMessaging;
+
+	// Utility function to safely parse integers and handle NaN
+	const safeParseInt = ( value ) => {
```
kalessil: Naming suggession: `parseIntOrReturnZero`

kalessil: How much efforts is it to convert file into TS?

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
 	const resetBnplPaymentMessage = () => {
-		const quantity = $( '.quantity input[type=number]' ).val();
-		productId = 'base_product';
-		bnplPaymentMessageElement.update( {
-			amount:
-				parseInt( productVariations.base_product.amount, 10 ) *
-				quantity,
-			currency: productVariations.base_product.currency,
-		} );
+		if ( ! quantityInput.length || ! productVariations.base_product )
```
kalessil: We are missing `{}` here.

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
+			return;
+
+		const quantity = safeParseInt( quantityInput.val() );
+		const baseProductAmount = safeParseInt(
+			productVariations.base_product.amount
+		);
+
+		const amount = baseProductAmount * quantity;
+		const currency = productVariations.base_product.currency;
```
kalessil: Would it make sense using `productVariations.base_product?.currency` here?

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
-			} );
-		} );
+		// Update BNPL message based on product variation
+		$( '.single_variation_wrap' ).on(
+			'show_variation',
+			( event, variation ) => {
+				if (
+					! quantityInput.length ||
+					! productVariations[ variation.variation_id ]
+				)
```
kalessil: We are missing `{}` here.

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
-		const newQuantity = event.target.value;
-		const price = productVariations[ productId ].amount;
-		bnplPaymentMessageElement.update( {
-			amount: parseInt( price, 10 ) * newQuantity,
-			currency: productVariations[ productId ].currency,
-		} );
+	// Update BNPL message based on the quantity change
+	quantityInput.on( 'change', ( event ) => {
+		const newQuantity = safeParseInt( event.target.value );
+		const price = safeParseInt( productVariations[ productId ].amount );
```
kalessil: Would it make sense using `productVariations[ productId ]?.amount` here?

client/product-details/index.js
```diff
@@ -7,51 +7,95 @@
...
+			( event, variation ) => {
+				if (
+					! quantityInput.length ||
+					! productVariations[ variation.variation_id ]
+				)
+					return;
+
+				const quantity = safeParseInt( quantityInput.val() );
+				const variationPrice = safeParseInt(
+					productVariations[ variation.variation_id ].amount
```
kalessil: Would it make sense using `productVariations[ productId ]?.amount` here?

kalessil (approved):
Pre-approving, as handed further reviews to @KarlisJ.

KarlisJ (commented):
Overall, when looking strictly at the changes and comparing old vs new code it's a nice improvement, I like the changes.

But my first though seeing the `resetBnplPaymentMessage` and `quantityInput.on( 'change',` was "why can't we trigger the `change` event from `resetBnplPaymentMessage`?" . I believe that's the "jQuery approch" for less code duplication. 

On the surface level it seems that both blocks operate with different values, but digging deeper it seems to me that it's not true. 

In `resetBnplPaymentMessage` we access `productVariations.base_product.amount` but in the change handler we access `productVariations[ productId ].amount`. Looking [at the source ](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-method-messaging-element.php#L83) `productId` that we define at the top will always be `base_product` so essentially both blocks perform the same action. 

Or have I misunderstood something?

client/product-details/index.js
KarlisJ: I'm in favor of existing `safeParseInt` naming

mgascam: I finally renamed to `parseIntOrReturnZero` for being more explicit of what the function does. https://github.com/Automattic/woocommerce-payments/pull/7153/commits/15d2d7d9d496dd3f5f60de7c92f249cf56e6c348 thanks!

> How much efforts is it to convert file into TS?

We are dealing with jQuery so we would need to install the type definitions and also modify the `tsconfig.json`. Nothing major, but I'd rather work on it in a different pull request.

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/pull/7153/commits/15d2d7d9d496dd3f5f60de7c92f249cf56e6c348 thanks!

client/product-details/index.js
mgascam: Done https://github.com/Automattic/woocommerce-payments/pull/7153/commits/15d2d7d9d496dd3f5f60de7c92f249cf56e6c348 thanks!

client/product-details/index.js
mgascam: Yep, done here https://github.com/Automattic/woocommerce-payments/pull/7153/commits/768537975c36b22b97068a2d490fde8b270b1b12 thanks!

client/product-details/index.js
mgascam: Done here https://github.com/Automattic/woocommerce-payments/pull/7153/commits/15d2d7d9d496dd3f5f60de7c92f249cf56e6c348

client/product-details/index.js
mgascam: Done https://github.com/Automattic/woocommerce-payments/pull/7153/commits/15d2d7d9d496dd3f5f60de7c92f249cf56e6c348

mgascam:
> Overall, when looking strictly at the changes and comparing old vs new code it's a nice improvement, I like the changes.
> 
> But my first though seeing the `resetBnplPaymentMessage` and `quantityInput.on( 'change',` was "why can't we trigger the `change` event from `resetBnplPaymentMessage`?" . I believe that's the "jQuery approch" for less code duplication.
> 
> On the surface level it seems that both blocks operate with different values, but digging deeper it seems to me that it's not true.
> 
> In `resetBnplPaymentMessage` we access `productVariations.base_product.amount` but in the change handler we access `productVariations[ productId ].amount`. Looking [at the source ](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-method-messaging-element.php#L83) `productId` that we define at the top will always be `base_product` so essentially both blocks perform the same action.
> 
> Or have I misunderstood something?

Thanks @KarlisJ I looked into your comment about variations and refactored the code. Also fixed some buggy scenarios where the messaging element was not being updated. Please take a look and let me know if you like it better.

> But my first though seeing the `resetBnplPaymentMessage` and `quantityInput.on( 'change',` was "why can't we trigger the `change` event from `resetBnplPaymentMessage`?" . I believe that's the "jQuery approch" for less code duplication.

Apologies but I'm not following you here. `resetBnplPaymentMessage` is a function we use as event handler, not sure what you mean when you suggest to trigger the change event from there?

KarlisJ:
@mgascam 
> Apologies but I'm not following you here. resetBnplPaymentMessage is a function we use as event handler, not sure what you mean when you suggest to trigger the change event from there?

Before your last commits it seemed that both `resetBnplPaymentMessage` and event handler perform the same action. So it seemed to me that we are duplicating logic I suggested to simply call `resetBnplPaymentMessage` in the event handler to avoid duplication.

Now I see that they do not perform same action thus can't be reused is such way.

client/product-details/index.js
```diff
@@ -7,51 +7,124 @@
...
+		// Update BNPL message based on product variation
+		$( SINGLE_VARIATION_SELECTOR ).on(
+			'show_variation',
+			( event, variation ) => {
+				if ( ! productVariations[ variation.variation_id ] ) {
+					return;
+				}
+
+				updateBnplPaymentMessage(
+					productVariations[ variation.variation_id ]?.amount,
```
KarlisJ: Given the check above I think it's safe to remove the `?`

KarlisJ (approved):
LGTM, approving with one minor comment.

client/product-details/index.js
mgascam: Yep, removed it https://github.com/Automattic/woocommerce-payments/pull/7153/commits/1cac1e43bbc5ef2ab4354126e123fd2f0a47be19 thanks!

PR #7174, opened 2023-09-12
Improve disputed order notice load time by directly fetching dispute rather than charge

### LINKED ISSUE #7076: Fetching the charge on each order screen is causing excessive `resource_missing (charge id)` errors
### Describe the bug

On each order screen where WooPayments is the payment method, [we fetch the charge](https://github.com/Automattic/woocommerce-payments/blob/850b3b891033c8573ebbada1582f05fb6540b5dc/client/order/index.js#L130) to determine if the charge is disputed and render a [disputed order notice](https://github.com/Automattic/woocommerce-payments/blob/850b3b891033c8573ebbada1582f05fb6540b5dc/client/order/index.js#L129).

This is contributing to excessive `Invalid request error: resource_missing (charge id)` errors occurring (see logstash), due to:
* the charge ID stored as order meta may be for a live transaction, and the site is in test mode (or vice versa).
* the charge ID stored as order meta may not exist anymore (I'm unsure if this is a possibility).

### To Reproduce

1. Use a WooPayments account that can switch between live/test mode (e.g. onboard with real bank account details).
1. Enable test mode.
1. Create a disputed order in test mode using the card `4000000000000259`.
1. View the order. See the dispute notice introduced in 6.2.
1. Switch to live mode.
1. View the order. See request to `/charges` responds `500` with message
	```
	Error: No such charge: 'ch_abcde'; a similar object exists in test mode, but a live mode key was used to make this request.
	```

### Additional context

* Slack discussion Aug 26 2023: p1692999889340419-slack-CGGCLBN58
* Slack discussion Jul 24 2023: p1690190397004289-slack-CGGCLBN58
* The disputed order notice was introduced in WooPayments 6.2:
   #6354

### LINKED ISSUE #7163: Dispute notice on order edit page is slow
### Describe the bug

As reported here: p1694347746794979-slack-C03LA109BKM, stores with high number of disputes experience long delay before showing the dispute notice.

I think we need to investigate whether `useCharge()` call [here](https://github.com/Automattic/woocommerce-payments/blob/69a25f7a57c13ca517c29d600d92c393791cd011/client/order/index.js#L135) results in slow database query in server.

Also, if `useCharge()` can potentially cause slowness, we need to make sure such call in other places does not cause the same problem.

### To Reproduce

0. Have high number of disputes.
1. Purchase a product with card `4000000000000259` that will generate dispute right after.
2. Open the order and notice the long delay before the dispute notice shows up.

Jinksi:
Fixes #7163
Fixes #7076

This PR is a proposal to improve the performance of the disputed order notice by doing the following:
1. When a dispute created or closed webhook is received
	1. Store the dispute ID as order meta
1. On order screen load
	1. Get the dispute ID from the order meta
		2. If no dispute ID order meta found, request the disputes API filtered by order_number to check for disputes created/received before this PR's release
	4. If the order's dispute ID exists, fetch the single dispute's data during page load (PHP)
	5. Dispute data is then provided to JS immediately on page load and the notice is rendered

* This results in an average **13.5% improvement in rendering time of this notice (refresh → notice visible)** on a WooExpress environment: _see metrics below_.
* Additionally, **testing WCCOM endpoints with live disputes**, I'm expecting this PR to result in a **4-5x improvement** in the dispute notice loading time:
	* the `/wc/v3/payments/charges/{charge_id}` fetch request takes **~6.5s** to respond (currently used for the disputed order notice).
	* the `/wc/v3/payments/disputes/{dispute_id}` fetch request takes **~1.5s** to respond (proposed to use in this PR).
* ~I've also added a `wcadmin_wcpay_single_order_view` tracks event to aid in tracking the "view disputed order" → "clicks disputed order notice button" funnel.~ removed – handled in #7437.
* This PR fixes #7076 since the `charge` endpoint is no longer utilised on the order edit screen.

**TODO:**
- [x] re-introduce `$dispute_id` argument to dispute webhook handler functions to add dispute id to order meta.
- [x] Remove tracks events from this PR (redundant, handled in #7437)

### Performance metrics

Method: Use [the browser `PerformanceElementTiming` API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceElementTiming) to measure the timestamp at which the disputed order notice is rendered. (see commit https://github.com/Automattic/woocommerce-payments/pull/7174/commits/d7ed392a1404d0f04c40eed80e0436cdae76d16a). Record results from 3x full-page refreshes with _Browser devtools -> Network -> Disable cache_ enabled.

For `develop` tests, I created a PR/branch with element performance timing measurements: #7175.

**On a live WooExpress site (Helix Demo site [Order 149](https://woo-delectably-sparkly-ninja.wpcomstaging.com/wp-admin/post.php?post=149&action=edit))**:

`develop`: average 6518ms
This branch average: 5632ms 
**13.5% improvement**

**On my local env with local WCPay-server running and proxied**:

`develop`: average 5375ms
This branch average: 4030ms 
**25% improvement**

### Testing instructions

* Create a disputed transaction using the card `4000000000000259` at checkout
* Navigate to the order edit screen
	* Ensure the disputed order notice is rendered
	* Ensure the tracks event `wcadmin_wcpay_single_order_view` is called with the properties `is_disputed: 1` & `dispute_status: needs_response`
		* Tip: check devtools network requests for `t.gif` with param `_en: wcadmin_wcpay_single_order_view`
	* Click the dispute notice and follow the steps to **Accept** the dispute
	* Ensure the tracks event `wcadmin_wcpay_single_order_view` is called with the properties `is_disputed: 1` & `dispute_status: lost`
		* Tip: check devtools network requests for `t.gif` with param `_en: wcadmin_wcpay_single_order_view`
* Navigate to the order edit screen for a non-disputed order
	* Ensure the disputed order notice is not rendered.
	* Ensure the tracks event `wcadmin_wcpay_single_order_view` is called with the properties `is_disputed: 0` & no `dispute_status`

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7174` or branch name `fix/7163-disputed-order-note-performance` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7163-disputed-order-note-performance) :rocket:

---

Build info: 

- Latest commit: 3b6f89266022ea10bc161025d1293a2baccd51db
- Build time: 2023-10-25 06:56:21 UTC

Jinksi:
Marking as ready for review to test via Jetpack Beta. Still a WIP.

client/order/index.js
```diff
@@ -22,14 +22,13 @@ import {
...
 	isUnderReview,
 } from 'wcpay/disputes/utils';
-import { useCharge } from 'wcpay/data';
 import wcpayTracks from 'tracks';
 import './style.scss';
 
 jQuery( function ( $ ) {
 	const disableManualRefunds = getConfig( 'disableManualRefunds' ) ?? false;
 	const manualRefundsTip = getConfig( 'manualRefundsTip' ) ?? '';
-	const chargeId = getConfig( 'chargeId' );
```
Jinksi, line 32: No need for frontend fetching of data. The dispute data will be fetched via PHP if the order is disputed and provided to the frontend, resulting in component rendering as soon as the page is loaded and JS is executed.

includes/class-wc-payments-order-service.php
```diff
@@ -520,6 +520,36 @@ public function get_charge_id_for_order( $order ) : string {
 		return $order->get_meta( self::CHARGE_ID_META_KEY, true );
 	}
 
+	/**
+	 * Get the dispute id from an order's notes if it exists.
```
Jinksi: Ideally, we would have stored this as order meta, similar to `_charge_id`.

I considered introducing an order meta solution. However, it would not be a backwards-compatible solution, as any previous disputed orders will not have this meta.

includes/class-wc-payments-order-service.php
```diff
@@ -520,6 +520,36 @@ public function get_charge_id_for_order( $order ) : string {
...
+		// Get current notes of the order.
+		$current_notes = wc_get_order_notes(
+			[ 'order_id' => $order->get_id() ]
+		);
+
+		foreach ( $current_notes as $current_note ) {
+			if ( $current_note->content ) {
+				$matches = [];
+				// Regex pattern to match dispute id e.g. "dp_1NpKDUC15a3WbJ1tRUP8UorZ".
+				preg_match( '/dp_[a-zA-Z0-9]{24}/', $current_note->content, $matches );
```
Jinksi: This is an initial idea of how to find the dispute ID for orders that have not stored the dispute id as meta – there may be an opportunity to make this more robust and/or more performant.

includes/class-wc-payments-order-service.php
Jinksi: Added this in f1f7dcc

client/order/index.js
Jinksi, line 1: Note for reviewer:

Another approach to improve the "perceived" load time of this notice could be to move all of the rendering of this component to PHP.

I say "perceived" because the dispute data fetch will still occur, but it will occur before the page has loaded. Therefore, it is likely that the overall load time will be similar.

We could consider iteratively attempting a PHP-only approach if this PR hasn't solved the problem.

Jinksi:
Since webhook handler functions (e.g. `mark_payment_dispute_created()`) have been changed to no longer provide/require the `$dispute_id` as an argument, I'll have to make some changes to re-introduce `$dispute_id` to add it to order meta.

includes/class-wc-payments-order-service.php
shendy-a8c: This won't work after we remove the dispute details page as the link will go to the transaction details page, so no `dp_` won't be found.

I'm trying to think of an alternative...

shendy-a8c (changes requested):

shendy-a8c:
I have not fully completed my review but found a problem where a disputed order without the dispute id in the order meta data isn't showing the dispute order notice.

includes/class-wc-payments-order-service.php
Jinksi: Good catch!

Jinksi:
> I have not fully completed my review but found a problem where a disputed order without the dispute id in the order meta data isn't showing the dispute order notice.

Good catch @shendy-a8c. I've moved this issue to `in-progress` while I work on a solution to this.

Jinksi:
I've proposed a solution for a dispute ID "fallback" for legacy disputes where the ID is not stored in order meta: get the dispute ID via a fetch request in (90e9205 and 996feaf).

It's not ideal, since it will require a network request to get the ID and then another to get the dispute based on this ID.

It also requires adding support for an [`order_number_is` filter on server](https://github.com/Automattic/woocommerce-payments-server/compare/trunk...add/disputes-order-number-filter).

client/order/index.js
haszari: If the issue is that the notice renders way after the rest of the page, then rendering server side could address that directly.

If you were coming at this clean, which approach would you recommend?

includes/class-wc-payments-order-service.php
```diff
@@ -520,6 +534,56 @@ public function get_charge_id_for_order( $order ) : string {
...
+		// get the dispute id via API.
+		$response = $this->api_client->get_disputes(
+			[
+				'pagesize'        => 1,
+				'order_number_is' => $order->get_id(),
+			]
+		);
+
+		// If a dispute exists, return the dispute id, otherwise return null.
+		$dispute_id = $response['data'][0]['dispute_id'] ?? null;
```
haszari, line 583: Does this need error handling? That's a deep reference!

haszari:
TLDR: do we need to proceed with this PR? If it's complicated we can pull back and reset (e.g. focus on conversion tracks and/or render notice server side).

Long version:

I saw a regex in the (old) diff for this PR and wanted to check-in that the effort / complexity here is worth the payoff. I see the regex is gone (that seemed like a hack?), but I see in the discussion there's still some tradeoffs in the PR. 

> It's not ideal, since it will require a network request to get the ID and then another to get the dispute based on this ID.

> Another approach to improve the "perceived" load time of this notice could be to move all of the rendering of this component to PHP.

Taking it back to the problem… I'm assuming the problem here is that the notice is rendered after the order page (by a few seconds). So it is easy to miss, and merchants are less likely to respond to the dispute. 

Do we think this PR will significantly improve that? I notice the PR title is "Improve" – how much is it improved?

Also related, there's another issue open to attack this problem: 

- #7437

i.e. allow us to directly measure the conversion rate of the notice (once rendered). That's a simpler intervention - adding a new tracks event. 

---

Side note – I am in progress on #7437 and will need to refactor the front end code for the notice. The component is doing a lot (including some hidden side effects); I'm planning to refactor to break it down into ~2 components.

If we do proceed with both PRs there will be some (minor) overhead managing merge conflicts on that component. 

- #7545

includes/class-wc-payments-order-service.php
shendy-a8c: I think the null coalescing operator `??` handles that. When any of the reference chain fails, it will return null.

shendy-a8c:
> do we need to proceed with this PR? If it's complicated we can pull back and reset

I think it's a good idea to reset. This PR gives a lot of learnings around this slowness problem though.

IMO there are 2 good potential solutions, each with its own pros and cons.

### Solution 1

Store dispute id in order meta data. Fetch dispute in the backend.

Pros:
- It's faster than current implementation as we will fetch dispute object only.
- We will only fetch dispute when order is disputed, not for all orders. Disputed orders are very rare.

Cons:
- Orders that are disputed with WooPayments versions before this PR won't display dispute notice.

I prefer solution 1 because orders only spends ~90 days in active dispute phase, so the cons above somewhat acceptable.

### Solution 2

Still keep most of the current implementation but instead of fetching charge, we fetch dispute by order id. This is inspired by @Jinksi's comment https://github.com/Automattic/woocommerce-payments/pull/7174#issuecomment-1778619906 (great solution by the way) but I feel like the fetching needs to be done in the frontend so it does not block page load, also because it needs to be done for all orders, not just disputed orders.

Pros:
- Dispute order notice will display for all disputed orders even though the orders were disputed before this PR's release.
- Likely fetching dispute by order is faster than fetching the charge object along with its related objects (read p1698316892008849-slack-C02BW3Z8SHK).

Cons:
- Wasteful fetching for all orders.

---

Also, I want to point out that we're still not sure why the original approach, ie fetching charge object to get the dispute object, is slow.

I agree with @haszari that we need to put tracking in place.

Originally, I thought fetching charge object is doing a database lookup but turns out, it's just eventually calling Stripe's API to get the charge object. We pass the [`expand` parameter](https://stripe.com/docs/api/expanding_objects), so dispute object and some other related objects to that charge is included in the response. Code reference can be found here p1698316892008849-slack-C02BW3Z8SHK, so the slowness would be just network slowness and Stripe's response time.

What do you all think?

shendy-a8c:
> I think it's a good idea to reset.

I meant to step back and try to understand the problem and come up with a solid agreed solution but if we all like solution 1, that mainly this PR is about, so we just need to remove the hacky parts like the regex order note lookup, which is already removed, and fetching the dispute by order id part.

That is if we all are OK to 'break' dispute order notice for orders that are disputed before this PR is released.

client/order/index.js
shendy-a8c: If we all align to take solution 1 from my comment here https://github.com/Automattic/woocommerce-payments/pull/7174#issuecomment-1780866276, we can render in PHP to even make the dispute notice appears on page load.

haszari:
Thanks for outlining those options @shendy-a8c ! I would only add that we have a third option of leaving the notice as it currently is – the perceived performance issue is not critical, we don't have to improve it :)

shendy-a8c:
> I would only add that we have a third option of leaving the notice as it currently is

That is an option.

I encourage we go with solution 1 though. I think the solution purposed by this PR, ie adding dispute id to the order meta is pretty neat and it'd be quite an improvement. Also, with current implementation, we are fetching charge on every order while with this PR, we will just fetch dispute on disputed orders only.

Jinksi:
Closing this PR while we gather more information about the problem via the tracks event introduced in #7437 – will help us determine if we still need to improve the performance or change how we deliver the message.

The issues this PR planned to resolve (#7163, #7076) will be moved to the backlog.

PR #7199, opened 2023-09-13
Preload checkout data in WooPay

lovo-h:

This PR adds a new WooPay express checkout flow. It does this by preloading checkout data the instance the WooPay Express Checkout button is clicked. At the moment, this new checkout flow is behind a feature flag.

This PR also fixes the following issues:
* Memory leak in `client/checkout/woopay/express-button/index.js`.
* Code cleanup in `clien/checkout/woopay/express-button/woopay-express-checkout-button.js`.

#### Testing instructions

For testing instructions, see 2160-gh-Automattic/woopay.

-------------------

- [x] Run `npm run changelog` to add a changelog file, and choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in the description ☝️)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios): _Add link here / 'QA Testing Not Applicable'_
- [ ] Add what's changed (description, screenshot, demo videos, etc.) to the release announcement post, if applicable.


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7199` or branch name `add/2094-pass-checkout-data-before-auth` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/2094-pass-checkout-data-before-auth) :rocket:

---

Build info: 

- Latest commit: 482acfe36095bda1e95bb5bc8ce03d51df15469b
- Build time: 2023-09-29 18:17:51 UTC

rafaelzaleski (approved):
✅ Code looks good and tests well.

On the product page, there's a noticeable lag before the redirect kicks in, and the button appears to remain active. It might be a good idea to introduce a loading state for better user feedback. To keep things tidy, maybe we can address this in a separate PR instead of bundling it with the current feature.

https://github.com/Automattic/woocommerce-payments/assets/4153905/966d4efb-d93e-44d7-b0d9-d586078fa73a

lovo-h:
> On the product page, there's a noticeable lag before the redirect kicks in, and the button appears to remain active. It might be a good idea to introduce a loading state for better user feedback...

@rafaelzaleski, I agree. I also like the idea of disabling the button and introducing a loading indicator. These UI/UX changes are being discussed in this conversation thread: pdpOw2-3Di-p2#comment-3534.

cc @stephendharmadi, @elizaan36, @pierorocca, @bborman22.

pierorocca:
@lovo-h possible to inspect the network performance to see where the lag is occuring? I'm not against the loader once we've exhausted feasible performance improvements.

lovo-h:
@pierorocca, I created a response to your question in pdpOw2-3Di-p2#comment-3568.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
```diff
@@ -69,7 +79,141 @@ export const WoopayExpressCheckoutButton = ( {
...
+						}
+						request(
+							buildAjaxURL(
+								getConfig( 'wcAjaxUrl' ),
+								'get_woopay_session'
+							),
+							{
+								_ajax_nonce: getConfig( 'woopaySessionNonce' ),
+							}
+						).then( ( response ) => {
```
cesarcosta99: Is there any side effects in case this request fails? Either on a 2xx or a 4xx status.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
```diff
@@ -69,7 +79,141 @@ export const WoopayExpressCheckoutButton = ( {
...
+									action: 'setPreemptiveSessionData',
+									value: response,
+								},
+								getConfig( 'woopayHost' )
+							);
+						} );
+					} );
+				} else {
+					// Non-product pages already have pre-fetched session data.
+					sessionDataPromiseRef.current.then( ( response ) => {
```
cesarcosta99: Perhaps we can be extra careful here and check if the promise exists before calling `then`?

```suggestion
					sessionDataPromiseRef.current?.then( ( response ) => {
```

cesarcosta99 (commented):
Code looks good overall and tests well 👍 

I left one minor suggestion 🤏  and have a concern regarding a request failure we may want to address before merging.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
lovo-h: Good callout. Before, on a 400 HTTP response, the button would fail silently. This has been fixed in https://github.com/Automattic/woocommerce-payments/pull/7199/commits/a9e994aed5191dc02d193778b07e06be712b3174 and https://github.com/Automattic/woocommerce-payments/pull/7199/commits/6e55eda8ebe9045d7b5b307bfcaa19b5f541c887. Now, an error notification appears.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
lovo-h: Although at the moment, `sessionDataPromiseRef.current` is guaranteed to be initialized, there is a chance we may refactor in the future and cause this to crash. Fixed in https://github.com/Automattic/woocommerce-payments/pull/7199/commits/530bc0908f101470adcadd9535cb71d7c4b5f662.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
```diff
@@ -196,68 +219,94 @@ export const WoopayExpressCheckoutButton = ( {
 			const isFromWoopayHost = getConfig( 'woopayHost' ).startsWith(
 				event.origin
 			);
-			const isConfirmSessionData =
-				event.data.action === 'confirmSetPreemptiveSessionData';
-			if ( ! isFromWoopayHost || ! isConfirmSessionData || isPreview ) {
+			const isSessionDataSuccess =
+				event.data.action === 'set_preemptive_session_data_success';
+			const istSessionDataError =
```
rafaelzaleski: Was there a reason to name it `istSessionDataError` instead of `isSessionDataError`? 🤔

client/checkout/woopay/express-button/express-checkout-iframe.js
```diff
@@ -2,8 +2,13 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies
+ */
 import { getConfig } from 'utils/checkout';
 import request from 'wcpay/checkout/utils/request';
+import { showErrorMessage } from 'wcpay/checkout/woopay/express-button/woopay-utils';
```
rafaelzaleski: The `woopay-utils` name doesn't quite capture the essence since the utils in this file are geared towards the express checkout. A straightforward `utils.js` inside the `express-checkout` directory might be clearer.

rafaelzaleski (approved):
@lovo-h, kudos on the feature – it’s looking good. I've dropped two minor comments about a possible typo and a file rename suggestion.

There's a slight hiccup in the error handling that caught my eye, but I believe it might be best to address it in a separate issue and PR. This way, we won't hold up the current PR. The specific error seems unlikely to happen on production and isolating its fix would simplify the review and testing process.

To detail, on cart or product pages, clicking the express checkout button and getting a `set_preemptive_session_data_error` from the auth iframe opens the OTP window as expected. But on the checkout page, this action results in an error that stops the checkout process via the button.

Screenshot 2023-09-29 at 12 29 25

client/checkout/woopay/express-button/express-checkout-iframe.js
lovo-h: Updated in https://github.com/Automattic/woocommerce-payments/pull/7199/commits/1412ee084854d98ee1a3b494d637b1a4382ca28e.

client/checkout/woopay/express-button/woopay-express-checkout-button.js
lovo-h: This was a typo. Good catch. Fixed in https://github.com/Automattic/woocommerce-payments/pull/7199/commits/bbf5c0cc345fcaf5cfef350750d0ae28516af7df

lovo-h:
> To detail, on cart or product pages, clicking the express checkout button and getting a set_preemptive_session_data_error from the auth iframe opens the OTP window as expected. But on the checkout page, this action results in an error that stops the checkout process via the button.

@rafaelzaleski, previously, all pages should have worked like the following:
1. Customer clicks the WooPay express checkout button.
2. Expect a `set_preemptive_session_data_error` error.
3. An error notification should appear.
4. If the customer were to click the button again, we would return to step 2.

I am unsure why the OTP modal appeared in your environment in the `cart` and `product` pages since things were not configured to switch over to the OTP flow. The only time the OTP flow is used is if the iFrame that drives the "seamless" flow fails to load up correctly, for any given reason.

However, after https://github.com/Automattic/woocommerce-payments/pull/7199/commits/badf3c5c43dee37d06a3c0ece865a54fede92266, if we get an error response from WooPay, we will now default to the OTP flow.

I wonder if we should only switch to the default OTP flow based on the type of error we receive from WooPay. For example, if we get a connection error, this tells us that the OTP flow will also fail. Therefore, we should reuse the "seamless" flow so the customer can try again. However, if we get an error response from the WooPay backend, we should switch back to the OTP flow. cc @bborman22.

bborman22:
Just wanted to drop in here that I was able to test the error state on product, cart and checkout pages and upon a second click of the button the OTP modal then appeared for me as expected. 👍

rafaelzaleski:
> I am unsure why the OTP modal appeared in your environment in the cart and product pages since things were not configured to switch over to the OTP flow. The only time the OTP flow is used is if the iFrame that drives the "seamless" flow fails to load up correctly, for any given reason.

I think I clicked on the button too soon when the page loaded. Giving it a moment, I can spot the error message on all pages.

rafaelzaleski (approved):
LGTM and it's all functioning smoothly. 🚀

Issue #6792, opened 2023-07-17
BNPL - Fix the PM conditional in Afterpay shipping address sending check


nagpai:
### Describe the bug

For Afterpay, we have a requirement of sending Shipping Address, in the request. We had introduced [this code](https://github.com/Automattic/woocommerce-payments/pull/6492/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1119) in an earlier PR to to do that. ( [Link in the develop branch ](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L1119)) 

Props to @FangedParakeet who reminded that `$payment_information->get_payment_method()` would not return the name of the payment method  (i.e. `afterpay_clearpay` ). Instead it returns the payment method id similar to `pm_xxxxxyyyy...` . Because of this the code in the if statement will never execute. 

### Additional context
Discussion on slack with hints to resolve this 
p1689060282807879-slack-C02C00HCJ8J
```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/woocommerce-payments/pull/6800
- [ ] https://github.com/Automattic/woocommerce-payments/pull/6838
```

nagpai:
I have removed the PHP snippet under discussion to see if it affects Afterpay checkout. 

The transaction completed successfully after removing the PHP snippet for classic as well as block checkout, and within each - for Legacy as well as Split UPE. However, for checkout cases where there was a separate shipping address specified, the address submitted to Afterpay was the billing address. 

I will be creating two PR's to fix this issue in Classic as well as Block checkout. (check tasklist in the main message of the issue )

nagpai:
2. Legacy UPE - Classic checkout - I tested with separate Shipping and billing address.  Checkout completes successfully, however, the billing address is submitted as shipping address. The payment completes successfully and the billing and shipping address appear correctly on the order.

KarlisJ:
@nagpai @mgascam I see that both associated PRs are merged. Can we close this issue?

kalessil:
Yes, we can =)

PR #6492, opened 2023-06-09
Add Affirm and Afterpay to Block Checkout

nagpai:
Fixes #6231 

#### Changes proposed in this Pull Request

* Provide option of Affirm and Afterpay at checkout for eligible merchants on a block-based checkout page

##### Local server preparation
* Check in to any test account of an eligible country. Eg. US - `acct_1LqHU4QocNIKfAAZ` . 
* Add dummy BNPL related fees to the local setup with these commands 
```
wp wcpay add_new_fee base base-us-affirm-fee usd --percentage_rate=0.042 --fixed_rate=42 --countries=US --payment_methods=affirm --mode=test --description="Affirm test"
wp wcpay add_new_fee base base-us-afterpay-fee usd --percentage_rate=0.042 --fixed_rate=42 --countries=US --payment_methods=afterpay_clearpay --mode=test --description="Afterpay test"
```
Add these for any country that you wish to test with by modifying the command accordingly 
* Set the BNPL flag to true
`wp wcpay set_account_meta acct_1LqHU4QocNIKfAAZ --key=bnpl_affirm_afterpay_enabled --value=1`

Make sure you use your test account number in the command.

##### Test
* Browse to `Payments > Settings` in WP Admin and make sure you turn on Affirm and Afterpay. 
* If not done already - create a block based checkout page, and set it as the checkout page from `WooCommerce > Settings > Advanced tab > Page setup` 
* Add items to the cart with total amount within the required threshold for the method you are testing ( this check is being addressed in a [separate PR](https://github.com/Automattic/woocommerce-payments/pull/6540) ) . 
    *  Affirm threshold - [For US](https://stripe.com/docs/payments/affirm#payment-plans-us) | [For Canada](https://stripe.com/docs/payments/affirm#payment-plans-ca)
    * [Afterpay thresholds](https://stripe.com/docs/payments/afterpay-clearpay#collection-schedule)
* Watch for any errors in console or in the front-end flow. 

Use the WCPay Dev tools and test the scenarios above for both UPE Classic and Split. 

-------------------

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

[...]

PR #7200, opened 2023-09-13
Fix checkout error with Afterpay

### LINKED ISSUE #7187: Error in checkout with Afterpay payment method due to missing shipping information
### Describe the bug

There's an `We're not able to process this request. Please refresh the page and try again.` error in the checkout with Afterpay payment method selected.

### To Reproduce

_You need to test in a store where Afterpay is enabled_

1. As a shopper, add a product to the cart and proceed to the checkout. 
2. Fill the form and select Afterpay from Payment options
3. Place order

### Actual behavior

An error appears after the checkout form is submitted:
`We're not able to process this request. Please refresh the page and try again.`

### Screenshots

Screenshot 2023-09-12 at 16 57 32

### Expected behavior

Order is placed without errors when Afterpay method is selected
### Additional context

- We can see this error in the Stripe dashboard, when we POST to `payment_intents` endpoint:
```
{
  "error": {
    "code": "parameter_missing",
    "doc_url": "https://stripe.com/docs/error-codes/parameter-missing",
    "message": "Missing required param: shipping[address][line1].",
    "param": "shipping[address][line1]",
    "request_log_url": "https://dashboard.stripe.com/acct_xxxxxx/test/logs/req_xxxxxx?t=1694520304",
    "type": "invalid_request_error"
  }
}
```
- According to my tests it affects Split and Defferred UPE implementations, the Legacy one is not affected.
- Can be reproduced in both block and shorcode based checkouts
- I can only reproduce the issue locally in v6.4.x, previous versions don't seem to be affected

mgascam:
Fixes #7187

This PR fixes the error in the checkout due to missing shipping information in the request to create and confirm a payment intent.

I'm adding the shipping information to the request in the `process_payment_for_order` method based on two conditions:

1. The selected payment method is `AFTERPAY`, since no other payment methods require shipping information at the moment.
2. The request is being handled by an instance of `UPE_Split_Payment_Gateway`, to avoid any risks of adding the shipping information to Legacy card gateway.

#### Testing instructions

_Prerequisites_

You need to have Afterpay enabled in the test site. If testing locally, make sure that your local server instance has the latest changes from `trunk`.  You'll also need a merchant account that complies with Afterpay requirements (i.e. US based merchant with USD currency). If you are still not able to see Afterpay in the admin, please try to clear the account cache. And finally make sure to have shortcode and block based checkout pages available.

1. Make sure that Enable Split UPE checkout with deferred intent creation is enabled.
2. Add a product to the cart and proceed to shortcode checkout.
3. Fill in the details, select Afterpay and submit the form.
4. Check that there are no errors in the checkout.
-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -3763,7 +3763,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 *
 	 * @return void
 	 */
-	protected function modify_create_intent_parameters_when_processing_payment( Create_And_Confirm_Intention $request, Payment_Information $payment_information, $order ): void {
+	protected function modify_create_intent_parameters_when_processing_payment( Create_And_Confirm_Intention $request, Payment_Information $payment_information, $order ) {
 		// Do nothing.
 	}
 }
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1235,6 +1235,10 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				$request->set_payment_methods( $payment_methods );
 				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
 
+				if ( Payment_Method::AFTERPAY === $this->get_selected_stripe_payment_type_id() && $this instanceof UPE_Split_Payment_Gateway ) {
+					$request->set_shipping( $this->get_shipping_data_from_order( $order ) );
+				}
+
 				// The below if-statement ensures the support for UPE payment methods.
 				if ( $this->upe_needs_redirection( $payment_methods ) ) {
 					$request->set_return_url(
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1235,9 +1235,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				$request->set_payment_methods( $payment_methods );
 				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
 
-				if ( Payment_Method::AFTERPAY === $this->get_selected_stripe_payment_type_id() && $this instanceof UPE_Split_Payment_Gateway ) {
-					$request->set_shipping( $this->get_shipping_data_from_order( $order ) );
-				}
+				// Add specific payment method parameters to the request.
+				$this->modify_create_intent_parameters_when_processing_payment( $request, $payment_information, $order );
 
 				// The below if-statement ensures the support for UPE payment methods.
 				if ( $this->upe_needs_redirection( $payment_methods ) ) {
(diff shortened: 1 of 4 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7200` or branch name `fix/7187-checkout-error-with-afterpay` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7187-checkout-error-with-afterpay) :rocket:

---

Build info: 

- Latest commit: c35cbf3c60ee107735ac298eac3bfd6d5bbd81ac
- Build time: 2023-09-14 10:27:58 UTC

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1235,6 +1235,10 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
 				$request->set_payment_methods( $payment_methods );
 				$request->set_cvc_confirmation( $payment_information->get_cvc_confirmation() );
 
+				if ( Payment_Method::AFTERPAY === $this->get_selected_stripe_payment_type_id() && $this instanceof UPE_Split_Payment_Gateway ) {
+					$request->set_shipping( $this->get_shipping_data_from_order( $order ) );
+				}
```
FangedParakeet: Just poking my nose in to mention that this field _was_ actually included in the original Affirm/Afterpay implementation here -> https://github.com/Automattic/woocommerce-payments/pull/6492/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1120. However, this specific piece of the implementation was inert (see #6792 for more details) and appeared to have no effect on completing transactions, so it was decided to remove it. 

I think it might not have been required with the legacy UPE element (used with both legacy UPE and split UPE), because when the payment element confirms the intent, Stripe tries its hardest to add any missing fields that might be required. That probably also explains this comment (here -> https://github.com/Automattic/woocommerce-payments/issues/6792#issuecomment-1640054759), which states that the billing address is used for this parameter, even when there is a separate shipping address, because the payment element doesn't know any better.

The deferred intent implementation is more manual, because we confirm the intent ourselves, rather than allowing the payment element to perform this task. I think it probably couldn't hurt to add this parameter for all UPE implementations, though it is only _required_ for the deferred intent UPE. 

I think checking whether the gateway is an instance of `UPE_Split_Payment_Gateway` is accurate enough, since the deferred intent UPE is presented to all split UPE users, as of the 6.4.0 release. Alternatively, you could just check `WC_Payments_Features::is_upe_deferred_intent_enabled()` to be as accurate and possibly a bit more explicit. But I'm not sure which conditional will honestly have greater longevity in the grand scheme of things, so no biggie. 🤷

KarlisJ (approved):
LGTM and tests well.

kalessil (changes requested):
We should move this into a protected method: empty method for base gateway class and overridden with the logic in `UPE_Split_Payment_Gateway`.

```
protected function modify_create_intent_parameters_when_processing_payment( Create_And_Confirm_Intention $request, Payment_Information $payment_information, $order) {
    if ( Payment_Method::AFTERPAY === $this->get_selected_stripe_payment_type_id() ) {
        $request->set_shipping( $this->get_shipping_data_from_order( $order ) );
    }
}
```

PS: this method should handle PMs-specific modifications and perhaps more edge-cases.
PPS: this way we should be more resilent in terms of changes in gateways.

frosso (approved):
Changes read well to me 👍

kalessil (approved):
LGTM!

mgascam:
I had to create a new PR https://github.com/Automattic/woocommerce-payments/pull/7214 because I mistakenly branched from `develop` instead of release branch.

PR #7202, opened 2023-09-13
Add helper function/method for raw amount conversion

### LINKED ISSUE #5524: Multi-Currency Compatibility: Price conversion helper function
### Description

I was revisiting the [compatibility Module for Name Your Price](https://github.com/Automattic/woocommerce-payments/commit/55a751bed1ca82c92f831d359a10516b14947d9a). If you add an item to the cart in X currency. And then go to the cart and click "edit" you are taken back to the single product page. _Now_ if you switch the currency the pre-filled amount is converted from the old currency decimal separators to the new.

https://woo-helgatheviking.codeanyapp.com/product/simple-nyp/

It is not converted, but worse the commas/decimals can get borked depending on which currencies you are switching to/from. Example: 55,55EUR (55 eur and 55 cents) becomes 5555.00 USD (five thousand five hundred fifty five) dollars.

I've found a way around this by adding a "Raw" price and a currency into the edit link. And then on the single product page I can prefill the price reliably, by converting that raw price from the link currency into the current currency. 

This was working well with some of my other multi-currency integrations. However, I was going to write that for this plugin too, when I remembered that the approach to conversion is rather complicated. Is there a straight currency conversion helper yet? I think you could benefit (or at least I would benefit :) ) from a helper function that could take any price from X to Y currency. I think that would make integrations a lot easier. 

```
$converted_price = WCPay\MultiCurrency\MultiCurrency::convert_price( $amount, $from_currency, $to_currency )
```

or something like

```
$converted_price = apply_filters( $amount, $from_currency, $to_currency );
```

Or failing that... what's the best way to approximate that assuming `$x = 44.44' where it's always in the database decimal format. 

### Testing instructions

1. Create an NYP simple product
2. Add that product to the cart
3. Go to the cart and click on "Edit" to edit the price
4. Now back on the simple product page, switch the currency using a widget/switcher/whatever - Especially noteworthy if switching from say EUR to USD where the decimal separators are different

jessepearson:
Fixes #5524 

This PR adds the `get_raw_conversion` method that allows for conversion of an amount from one currency to another without any sort of rounding or price charm added. This helps with compatibility issues related to third party developer plugins.

#### Testing instructions

* The tests assume that your store's currency is USD.
* Navigate to WooCommerce > Settings > Multi-Currency and then enable CAD and EUR if not enabled.
  * CAD's conversion rate should be the local default: 1 USD = 1.288345 CAD
  * EUR's conversion rate should be the local default: 1 USD = 0.913827 EUR
* Add the below snippet to your site.
* After the snippet is added, any page on the front end.
* Check your `debug.log` to find output similar to the below.
  * CAD $amount: 12.88345
  * EUR CAD $amount: 7.0930302054186

Snippet:
```
add_action( 'wp_footer', function() {
	if ( ! function_exists( 'WC_Payments_Multi_Currency' ) ) {
		return;
	}
	$amount = WC_Payments_Multi_Currency()->get_raw_conversion( 10.00, 'CAD' );
	error_log( 'CAD $amount: ' . print_r( $amount, true ) );
	$amount = WC_Payments_Multi_Currency()->get_raw_conversion( 10.00, 'EUR', 'CAD' );
	error_log( 'EUR CAD $amount: ' . print_r( $amount, true ) );
});
```
-------------------


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7202` or branch name `fix/5524-price-conversion-helper-function` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/5524-price-conversion-helper-function) :rocket:

---

Build info: 

- Latest commit: 4186c57e059fac94d5aaa8f9a72d841e3d7aaeee
- Build time: 2023-10-04 11:34:15 UTC

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,45 @@ public function get_price( $price, string $type ): float {
...
+	public function get_raw_conversion( float $amount, string $to_currency, string $from_currency = '' ): float {
+		$enabled_currencies = $this->get_enabled_currencies();
+
+		// We return the amount if the to_currency is not enabled.
+		$to_currency = strtoupper( $to_currency );
+		if ( ! isset( $enabled_currencies[ $to_currency ] ) ) {
+			return $amount;
+		}
+
+		if ( '' !== $from_currency ) {
```
tpaksu: This could be simplified with setting `$from_currency_rate` to `1` when the `$from_currency` is empty. And then the converted `$amount` would be `$amount * ($to_currency_rate / $from_currency_rate)`. 

Also we could add a `0` check to the `$from_currency_rate` to prevent fatal errors.

tpaksu (approved):
Tested it on my local, echoed the results on the footer directly and got the numbers. I've checked the amounts on Google with `10 CAD to EUR` and it gave a close number, Also tested in my local currency. It works. The one thing confusing here is returning the amount when one of the currencies don't exist. Maybe we can throw some exception there, so the user can catch and act accordingly? Otherwise it seems the user won't know about the same number being returned as an error, or the rate of the target currency is really "1". It may happen on currencies that are pegged to each other, so they have the same rate. Other than that, I left a nitpick in the comments, feel free to address it or not. Your code is much cleaner, but I needed some time to understand why you wrote it like this :) LGTM, :shipit:

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,45 @@ public function get_price( $price, string $type ): float {
...
+	public function get_raw_conversion( float $amount, string $to_currency, string $from_currency = '' ): float {
+		$enabled_currencies = $this->get_enabled_currencies();
+
+		// If the from_currency is not set, use the store currency.
+		if ( '' === $from_currency ) {
+			$from_currency = $this->get_default_currency()->get_code();
+		}
+
+		// We throw an exception if either of the currencies are not enabled.
+		$to_currency   = strtoupper( $to_currency );
```
tpaksu, line 920: Thanks for doing this!

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,45 @@ public function get_price( $price, string $type ): float {
...
+			if ( ! isset( $enabled_currencies[ $code ] ) ) {
+				$message = 'Invalid currency passed to get_raw_conversion: ' . $code;
+				Logger::error( $message );
+				throw new InvalidCurrencyException( $message, 'wcpay_multi_currency_invalid_currency', 500 );
+			}
+		}
+
+		// Get the rates and convert the amount.
+		$to_currency_rate   = $enabled_currencies[ $to_currency ]->get_rate();
+		$from_currency_rate = $enabled_currencies[ $from_currency ]->get_rate();
```
tpaksu, line 930: Can we add a `0` check here and throw if the from currency rate somehow returns falsy? It'll already throw a "division to zero" exception, but IMO it'll be good to own it.

includes/multi-currency/MultiCurrency.php
jessepearson: No problem. You were right with what you said in your previous comment about how things were written, they were really simple and basic with no real forethought on errors. I added in the exceptions when the currencies are not about to be found because it makes sense since we can be dealing with anyone's input.

includes/multi-currency/MultiCurrency.php
jessepearson: Added in https://github.com/Automattic/woocommerce-payments/pull/7202/commits/d9545cd1d884d1dae3b06634491ef2e29af14d56 . I had made the previous update, read the notes again, and realized I forgot to add it. Getting the testing to work was a pain, I kept trying to pass `BAN` as a currency into the mock arrays, then found that it had to be an actual currency code, so I changed it to `BAM` and it worked 🤦

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,54 @@ public function get_price( $price, string $type ): float {
...
+				throw new InvalidCurrencyException( $message, 'wcpay_multi_currency_invalid_currency', 500 );
+			}
+		}
+
+		// Get the rates.
+		$to_currency_rate   = $enabled_currencies[ $to_currency ]->get_rate();
+		$from_currency_rate = $enabled_currencies[ $from_currency ]->get_rate();
+
+		// Thrown an exception in case from_currency_rate is less than or equal to 0.
+		if ( 0 >= $from_currency_rate ) {
```
jessepearson, line 933: Noting that `0 === $from_currency_rate` does not work here since `$from_currency_rate` is a float. I also figured it would be better to make sure we weren't doing any sort of negative conversion.

tpaksu: Makes sense, `get_rate` will always return a float, so no need for checking the other types. Thanks for also adding the negatives to the check!

tpaksu (approved):
No need to test the latest change, the PHP test you've written explains it all. And the code makes sense. Thanks for fixing it, the previous approval is still good :) :shipit:

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,54 @@ public function get_price( $price, string $type ): float {
...
+		if ( '' === $from_currency ) {
+			$from_currency = $this->get_default_currency()->get_code();
+		}
+
+		// We throw an exception if either of the currencies are not enabled.
+		$to_currency   = strtoupper( $to_currency );
+		$from_currency = strtoupper( $from_currency );
+		foreach ( [ $to_currency, $from_currency ] as $code ) {
+			if ( ! isset( $enabled_currencies[ $code ] ) ) {
+				$message = 'Invalid currency passed to get_raw_conversion: ' . $code;
```
fashuvo: I was thinking whether we can use early return for the validations checks? This may improve the readability and also avoid duplication. I am thinking of something like this:

```
foreach ( [ $to_currency, $from_currency ] as $code ) {
			if ( ! isset( $enabled_currencies[ $code ] ) ) {
				$this->handle_currency_error( 'Invalid currency' );
			}
}

...

// Validate the rate.
    if ( 0 >= $from_currency_rate ) {
        $this->handle_currency_error( 'Invalid rate' );
    }

...

private function handle_currency_error( string $message ): void {
    $log_message = 'Error in get_raw_conversion: ' . $message;
        
    Logger::error( $log_message );
    throw new InvalidCurrencyException( $log_message, 'wcpay_multi_currency_invalid_currency', 500 );
}
```

What do you think about this change?

includes/multi-currency/MultiCurrency.php
```diff
@@ -845,6 +845,54 @@ public function get_price( $price, string $type ): float {
...
+				Logger::error( $message );
+				throw new InvalidCurrencyException( $message, 'wcpay_multi_currency_invalid_currency', 500 );
+			}
+		}
+
+		// Get the rates.
+		$to_currency_rate   = $enabled_currencies[ $to_currency ]->get_rate();
+		$from_currency_rate = $enabled_currencies[ $from_currency ]->get_rate();
+
+		// Thrown an exception in case from_currency_rate is less than or equal to 0.
```
fashuvo: Did we mean `Throw` instead of `Thrown`? 🤔

fashuvo (commented):
_This is not a blocker for the PR. I stumbled upon this while looking for PRs from other teams to review._

I like how the newly introduced `get_raw_conversion` method is well-documented, which makes it very easy to understand what it does. And adding tests for all possible scenarios is a great thing 😃 

I shared some thoughts, would love to heard your feedback on that.

includes/multi-currency/MultiCurrency.php
jessepearson: If we were going to create a method that handled logging and throwing the exception, it would need to be more generic so that it could be used for other instances where `InvalidCurrencyException` is thrown within the class. Something like this could work:

```php
	private function log_and_throw_invalid_currency_exception( $method, $currency_code, $code ) {
		$message = 'Invalid currency passed to ' . $method . ': ' . $currency_code;
		Logger::error( $message );
		throw new InvalidCurrencyException( $message, 'wcpay_multi_currency_invalid_currency', $code );
	}
```

includes/multi-currency/MultiCurrency.php
jessepearson: Updated in https://github.com/Automattic/woocommerce-payments/pull/7202/commits/fe10372c5c0922409a11ee964beac0a96fdb3b6e

includes/multi-currency/MultiCurrency.php
jessepearson: Added the new method in https://github.com/Automattic/woocommerce-payments/pull/7202/commits/fe10372c5c0922409a11ee964beac0a96fdb3b6e

jessepearson:
@tpaksu Would you mind taking a look at the latest commit to see if this is something we'd like to keep. I feel like it helps create a standard for throwing the exception in the class.

includes/multi-currency/MultiCurrency.php
```diff
@@ -871,17 +865,15 @@ public function get_raw_conversion( float $amount, string $to_currency, string $
...
 		}
 
 		// Get the rates.
 		$to_currency_rate   = $enabled_currencies[ $to_currency ]->get_rate();
 		$from_currency_rate = $enabled_currencies[ $from_currency ]->get_rate();
 
-		// Thrown an exception in case from_currency_rate is less than or equal to 0.
+		// Throw an exception in case from_currency_rate is less than or equal to 0.
 		if ( 0 >= $from_currency_rate ) {
 			$message = 'Invalid rate for from_currency in get_raw_conversion: ' . $from_currency_rate;
```
tpaksu, line 934: I think you missed this one.

tpaksu: Sorry, this was a rate exception and you grouped the invalid currency ones. I think what @fashuvo meant at first was removing the duplicate code in this method when throwing errors. Maybe you can convert it to a more generic one that also accepts rate errors?

jessepearson: That one is an `InvalidCurrencyRateException`. I thought about adding a helper method for that exception, as well, but there were only two instances and the messages were unique to where they were thrown.

jessepearson: Not sure how I missed this comment before, it wasn't here when I replied previously.

> I think what @fashuvo meant at first was removing the duplicate code in this method when throwing errors. Maybe you can convert it to a more generic one that also accepts rate errors?

My reply above kind of covers this. If we were to create a handler for throwing the `rate` exception, then we would need to change the messages thrown, unless we allowed to pass messages in. Doing that wouldn't reduce any code or clarify anything, unless I am misunderstanding.

tpaksu: Gotcha. I was also fine with the previous approach, so maybe you can merge this one, and let it be handled by some other issue, if it's needed. The only thing that's concerning is the dynamicity of the error log, but I guess that was an issue for the server code, not the client code, right? You can merge this IMO.

jessepearson:
Waiting on a review of https://github.com/Automattic/woocommerce-payments/pull/7269 before merging.

PR #7205, opened 2023-09-13
Load MultiCurrency class on Multi-currency setup page

### LINKED ISSUE #7151: Multi-Currency onboarding is broken
### Describe the bug
There is a note that a merchant receives in the dashboard directing them to onboard Multi-Currency, but the onboarding wizard currently does not work. When accessing the wizard you get two errors and cannot use the wizard:

* Error retrieving currencies.
* Error retrieving store settings.

### To Reproduce

1. Go to Payments > Settings > Advanced and confirm Multi-Currency is disabled.
2. Go to `yourstoreurl/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fmulti-currency-setup`
3. Receive errors.

### Additional context
This was found by Taha in https://github.com/Automattic/woocommerce-payments/pull/7141

eduardoumpierre:
Fixes #7151

This PR aims to fix the issues after accessing the Multi-currency setup page while having Multi-currency disabled. As described by Jesse in https://github.com/Automattic/woocommerce-payments/issues/7151#issuecomment-1712016217, the issue is that the Multi-currency endpoints aren't loaded when the feature is disabled. The proposed fix is to check the URL parameters in order to define if the Multi-currency code should be included in the page load.

##### Changes:
- Enable the multi-currency feature if it was previously disabled
- Continue setup without selecting extra currencies if already configured
- Always load multi-currency code on setup page

#### Testing instructions

**Load errors**

* **Before checking out to this branch**
  * Go to **Payments > Settings > Advanced** and confirm Multi-Currency is disabled.
  * Go to `yourstoreurl/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fmulti-currency-setup`
  * Notice the toasts with error messages.
  * The content doesn't load.
* **Checkout to this branch**
  * Refresh the page
  * There shouldn't appear any toasts and the content should load normally.

**Enable multi-currency and "Continue" button**

  * Go to **Payments > Settings > Advanced** and confirm Multi-Currency is disabled.
  * Go to `yourstoreurl/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fmulti-currency-setup`
  * Select at least one additional currency
  * Go through the onboarding wizard and finish the setup
  * Go to **Payments > Settings > Advanced** and confirm Multi-Currency is enabled.
  * Go to `yourstoreurl/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fmulti-currency-setup`
  * Deselect any preselected currency
  * The button should be enabled and contain the text "Continue" to allow the merchant to go through the setup without having to add additional currencies
  * Go through the onboarding wizard and finish the setup

-------------------

**Post merge**


eduardoumpierre:
@jessepearson Multi-currency isn't enabled even after going through the setup process. Should we create another issue to tackle that?

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7205` or branch name `fix/7151-multi-currency-onboarding` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7151-multi-currency-onboarding) :rocket:

---

Build info: 

- Latest commit: 47cd9bc25b97a6341621d3de5556162c34beacfb
- Build time: 2023-10-05 13:14:58 UTC

tpaksu:
@eduardoumpierre I feel like this solution will work, but can't escape the feeling that it may have some consequences. Like what will happen if a WP Cron task is run, which should be running normally with MC disabled, but instead running with MC enabled because of this?

jessepearson:
> @jessepearson Multi-currency isn't enabled even after going through the setup process. Should we create another issue to tackle that?

We had a discussion in Slack about this whole issue: p1694692579488859-slack-C01B8KNUYSW

What we're thinking is that we haven't run into this issue due to MC is enabled by default, which means that the wizard just guided the merchant to set up MC and not enable it. I started another thread about disabling the note completely if MC is disabled: p1694693449259509-slack-C01B8KNUYSW

I think the note removal process would be better than trying to force MC to work when it's likely been manually turned off. 

If we do go with the fix you've created here, I think adding the db update to enable MC should be added in this PR so it fully solves the issue. The estimate may need to be increased because of that.

eduardoumpierre:
> I feel like this solution will work, but can't escape the feeling that it may have some consequences. Like what will happen if a WP Cron task is run, which should be running normally with MC disabled, but instead running with MC enabled because of this?

Thanks for raising this concern, @tpaksu! I've updated the code on https://github.com/Automattic/woocommerce-payments/pull/7205/commits/65d13e354b74e6dec3050931bc7a038c160ce05a to ensure we're not running the setup page check on the CRON environment. I also wrapped the check logic inside a function, so it should only run if Multi-Currency is disabled.

eduardoumpierre:
> If we do go with the fix you've created here, I think adding the db update to enable MC should be added in this PR so it fully solves the issue. The estimate may need to be increased because of that.

Thanks for the extra context @jessepearson! I updated the code to:
- Enable multi-currency if necessary (https://github.com/Automattic/woocommerce-payments/pull/7205/commits/72a7a3403b51441d0fed646f877a8c4af65c3cf0)
- Allow the merchant to continue the setup process without having to add another currency (https://github.com/Automattic/woocommerce-payments/pull/7205/commits/93257957857a40222fa278499e1b21d1bb5d9669)

client/multi-currency-setup/tasks/store-settings-task/index.js
```diff
@@ -65,10 +70,18 @@ const StoreSettingsTask = () => {
...
+		if ( ! isMultiCurrencyEnabled ) {
+			updateIsMultiCurrencyEnabled( true );
+			saveSettings();
+		}
+
 		submitStoreSettingsUpdate(
 			isAutomaticSwitchEnabledValue,
-			isStorefrontSwitcherEnabledValue
+			isStorefrontSwitcherEnabledValue,
+			! isMultiCurrencyEnabled
```
allie500, line 82: Just want to make sure I'm understanding this correctly. `isMultiCurrencyEnabled` is a boolean constant. Inside `handleContinueClick` on line 74, if `isMultiCurrencyEnabled` is `false`, we call `updateIsMultiCurrencyEnabled` and pass `true` to enable multi-currency. Then on line 82, we are passing the opposite boolean value of `isMultiCurrencyEnabled` to `submitStoreSettingsUpdate`. Is that correct?

allie500 (commented):
Thanks for working on this multi-currency fix, Eduardo. The manual testing worked for me, but I did need to run `npm run build:client` after checking out this PR branch to get the tests on this branch to work. _Most_ of the code makes sense to me and looks good. I left an inline comment where I could use a bit more enlightenment. Could you share some more information there when you have a moment?

Lastly, it does look like you addressed both Taha's and Jesse's feedback, but I'd like for them to review the PR's code changes to make sure they are happy that the changes address their concerns. 

Thanks again for the work completed so far. I'm looking forward to your feedback on my question.

client/multi-currency-setup/tasks/store-settings-task/index.js
eduardoumpierre: That's correct. We just want to suppress the `submitStoreSettingsUpdate` toast if we are also triggering the `saveSettings` function to update the multi-currency setting, as it will trigger a toast as well.

The third argument of `submitStoreSettingsUpdate` is `suppressNotices` which aims to update the multi-currency store settings silently.

This is how the toasts look without suppressing one of them:
image

eduardoumpierre: If multi-currency is enabled, then we don't need to call `saveSettings` or suppress the toast from `submitStoreSettingsUpdate`, that's why they're related.

eduardoumpierre: Here is the suppressing mechanism on `submitStoreSettingsUpdate`:

https://github.com/Automattic/woocommerce-payments/blob/977ee204c76c90d2e8a3f33608040f542481b25f/client/data/multi-currency/actions.js#L119-L150

eduardoumpierre: Do you think the code would be clearer if we moved the suppress notice mechanism to the `saveSettings` function?

That's how it would look like:

```js
	const handleContinueClick = () => {
		setPending( true );

		if ( ! isMultiCurrencyEnabled ) {
			updateIsMultiCurrencyEnabled( true );
			saveSettings( true );
		}

		submitStoreSettingsUpdate(
			isAutomaticSwitchEnabledValue,
			isStorefrontSwitcherEnabledValue
		);

		setPending( false );
		setCompleted( true, 'setup-complete' );
	};
```

allie500: Thanks for the explanation. It makes sense. I guess I was just not fully focused when I was looking at it the first time and became a bit confused as a result. I think it is fine to keep the current implementation.

allie500 (approved):
Thanks for your explanation in my previous comment, and thanks for working on this fix. I have successfully completed the manual testing steps. The code is the same since the last time I reviewed it, except for the small typo fix. I have checked and everything looks fine with that change.

I think this is good to ship. :shipit:

PR #7215, opened 2023-09-14
Fix WooPay Multiple Redirects

bborman22:

This is one piece of the fix for https://github.com/Automattic/woopay/issues/2125. This fix prevents cases where WooPay sending multiple redirect messages will trigger too many WooPay session init calls and create too many draft orders causing issues with WooPay checkout.

This change was needed because there was a scenario where 3rd party cookies were not blocked and the shopper was already logged into WooPay where the `redirect_to_woopay` post message was getting posted to the merchant store multiple times. This triggered multiple session init calls which created an issue with draft orders being created on the merchant store and at times preventing WooPay checkout.

This PR addresses that issue by checking if we are already in the process of redirecting the shopper and prevents any further redirects. It does this check by using a flag within the WooPay front end specific code.

Below are the network logs from dev tools after clicking the WooPay button while already logged into WooPay.

**Before**
too-many-redirects

**After**
Screenshot 2023-09-15 at 10 28 35 AM

#### Testing instructions

1. Make sure 3rd party cookies are allowed (or if you're using the same TLD for both merchant and platform, it will work either way).
2. Go to the merchant store and checkout using WooPay.
3. Stay logged into WooPay and immediately go to checkout again.
4. This time when the OTP modal opens it should begin an immediate redirect.
5. You should only see 1 wcpay_init_woopay call.
6. At this point check in the merchant store admin area and confirm this is only one draft order created.
7. Continue checkout on WooPay successfully.
8. Confirm that the only draft order from before is now completed and there were no additional draft orders created for that session.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7215` or branch name `fix/woopay-multiple-redirects` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/woopay-multiple-redirects) :rocket:

---

Build info: 

- Latest commit: 010dcc65352b3e97e99ef6da7d67841263f861c2
- Build time: 2023-09-25 16:44:26 UTC

ricardo:
I find we can still trigger multiple requests to `wcpay_init_woopay` by clicking the WooPay express checkout button while the automatic redirect is in progress. This results in multiple draft orders as well.

@bborman22 maybe we can tackle this in a separate issue by introducing some global flag which prevents multiple modals from stacking on top of each other?

https://github.com/Automattic/woocommerce-payments/assets/5509901/dfc6fa69-e6d5-4b12-b59a-bb0063829f3e

ricardo (approved):
Code LGTM and tests well. Nice fix! 💯 

I'm able to replicate the behavior of multiple requests to `wcpay_init_woopay` in `develop`, and not in this branch. This behavior only happens to me when I click the WooPay button from the cart page, not the checkout page.

However, as I mentioned in [my previous comment](https://github.com/Automattic/woocommerce-payments/pull/7215#issuecomment-1721864969), if I click the WooPay button from the checkout page while the automatic redirect is already in progress, I still face the same issue with multiple requests and multiple draft orders.

I guess we can tackle this in a separate PR, totally up to you. 👍

bborman22:
Thanks for taking a look at this. After our discussion I did some experimenting and actually was able to reproduce the issue on product and checkout pages with just the button (ignoring the email input flow for that test). I also checked this against cart and checkout blocks as well.

Also from our discussion I explored moving this check into the `api` object. While that didn't solve the issue with the email input since they are using two different instances, I did feel like the code was a lot cleaner this way and a bit easier to maintain / understand. 

To truly solve the email input + button click issue we would have to resort to some sort of global flag on the page. I would personally consider the failing scenario an edge case (shopper loads checkout, email lookup triggers auto redirect, and shopper clicks on WooPay) and would be hesitant to add a global for just that case. Overall, I think this area needs a bit of a refactor to get more of the code shared and that would be the proper way to solve this edge case in my opinion.

@ricardo if you wouldn't mind taking another look at this PR and let me know what you think of this `api` approach and confirm it works for you as expected and we can address the other case in a separate PR?

ricardo (approved):
>  I would personally consider the failing scenario an edge case (shopper loads checkout, email lookup triggers auto redirect, and shopper clicks on WooPay) and would be hesitant to add a global for just that case.

I agree. It's probably better to explore improving the overall integration of express checkout button + email input than introducing a global flag right now.

> I explored moving this check into the api object.

This approach looks much cleaner 👏.

The latest changes tested well for me. Code LGTM. ✅

PR #7107, opened 2023-08-31
Temporarily disable saving SEPA

gpressutto5:
Fixes #6205

#### Changes proposed in this Pull Request

Temporarily disables saving/reusing SEPA Direct Debit payment methods because we don't want to enable it for subscriptions and currently they share the same flag `is_reusable`. Check discussion p1693511033290239/1693490563.321839-slack-C048UEB37MZ

#### Testing instructions

* Enable SEPA and ensure it is not possible to save it for later or use it for new subscriptions

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7232, opened 2023-09-19
Enable deferred intent creation UPE for SEPA

### LINKED ISSUE #7113: Make stores eligible for deferred intent UPE when SEPA is active
### Description

Now that #7107 has been merged--disabling saved payment and subscription usage for SEPA--we can include it as an available payment method in our deferred intent UPE release. Users should be able to complete checkout using SEPA when the deferred intent UPE is enabled. 

Since deferred intent UPE is automatically enabled when either the deferred intent or split UPE flag are present, I think we can just remove the `WC_Payments_Features::is_upe_split_enabled()` function entirely, so that we no longer fall back to the legacy UPE when either split UPE or deferred intent UPE flags are present...but this should be tested properly, as I'm merely suggesting this off the top of my head. 😅 

### Acceptance criteria

- When the deferred intent UPE is enabled and SEPA is active, we should no longer fall back to the legacy UPE.
- User should be able to complete checkout using SEPA and the deferred intent UPE.
- User should still not be able to save SEPA payment methods and SEPA should not be present at checkout when subscription products are in cart. 

### Additional context

#6205

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/7113

After SEPA has been disabled as a reusable payment method (including Subscriptions) in https://github.com/Automattic/woocommerce-payments/pull/7107, this PR removes the fallback from the deferred intent creation UPE to legacy UPE so that SEPA can be seamlessly used with the deferred intent creation UPE. 

#### Testing instructions

_Legacy UPE_
1. Navigate to WCPay DevTools and enable legacy UPE
2. Enable SEPA in `Payments -> Settings`
3. Confirm that shortcode & Blocks checkout both work as expected when paying with SEPA for a physical product (check the order status in WC Core and Stripe Dashboard)
4. Confirm that shortcode & Blocks checkout don't list SEPA as an option for subscription-based carts nor allow saving SEPA payment method for future usage (more details in: https://github.com/Automattic/woocommerce-payments/pull/7107)
5. Enable Stripe Link and perform two payments (shortcode checkout & Blocks) to confirm that Stripe Link works as expected after changes that touched this area a bit

_Deferred intent creation UPE_
1. Navigate to WCPay DevTools and enable deferred intent creation UPE
2. Enable SEPA in `Payments -> Settings`
3. Open the shortcode checkout page and confirm that there's no fallback to the legacy UPE and instead SEPA is listed as part of the deferred UPE gateways list
4. Perform steps 3-5 from the previous testing block above on the legacy UPE

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7232` or branch name `feature/enable-sepa-for-deferred-intent-upe` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=feature/enable-sepa-for-deferred-intent-upe) :rocket:

---

Build info: 

- Latest commit: ad9ca3c2fc292ca90e5b5f27b44b07deee54c2a9
- Build time: 2023-09-20 10:37:52 UTC

includes/payment-methods/class-upe-split-payment-gateway.php
```diff
@@ -377,6 +377,20 @@ public function create_token_from_setup_intent( $setup_intent_id, $user ) {
...
 
+	/**
+	 * Mandate must be shown and acknowledged by customer before deferred intent UPE payment can be processed.
+	 * This applies to SEPA and Link payment methods.
+	 * https://stripe.com/docs/payments/finalize-payments-on-the-server
+	 *
+	 * @return boolean True if mandate must be shown and acknowledged by customer before deferred intent UPE payment can be processed, false otherwise.
+	 */
+	public function does_payment_method_require_mandate_data() {
+		$is_link_payment       = Payment_Method::CARD === $this->get_selected_stripe_payment_type_id() && in_array( Payment_Method::LINK, $this->get_upe_enabled_payment_method_ids(), true );
```
frosso: Is it accurate to say that `$is_link_payment` could be `true` if the payment method was CC and the merchant enabled Stripe Link? (i.e.: does this check accurately tell us if the customer paid with Stripe Link, and not just with CC?)
And would it matter?

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1483,6 +1479,17 @@ public function process_payment_for_order( $cart, $payment_information, $schedul
...
 	}
 
+	/**
+	 * The parent method which allows to modify the child class implementation, while supporting the current design where the parent process_payment method is called from the child class.
+	 * Mandate must be shown and acknowledged under certain conditions for Stripe Link and SEPA.
+	 * Since WC_Payment_Gateway_WCPay represents card payment, which does not require mandate, we return false.
+	 *
+	 * @return boolean False since card payment does not require mandate.
+	 */
+	protected function does_payment_method_require_mandate_data() {
```
frosso: (minor) just as an alternative with the `is_`/`has_` prefixes convention that is present in _most_ of the codebase, what are your thoughts on naming this new method to `is_mandate_data_required`?

frosso (approved):
Tested, all the changes look good 👍 just a minor question on the mandate check

includes/payment-methods/class-upe-split-payment-gateway.php
timur27: > Is it accurate to say that $is_link_payment could be true if the payment method was CC and the merchant enabled Stripe Link?

Yes. But it'd be inaccurate to say that `$is_link_payment` could be `true` **if and only if** the payment method was CC and the merchant chose Stripe Link. Because for the use-case, when someone's paying with CC w/o choosing Stripe Link (but having it enabled), `$is_link_payment` will be `true` as well.

> does this check accurately tell us if the customer paid with Stripe Link, and not just with CC?

Unfortunately, no. 

The check will be `true` for Stripe Link enabled, even if it's not used. Consequently, the `mandate_data` will be sent for plain CC payments as well. [This logic](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L1540C3-L1545) is being followed to support the process. I think that the Stripe Link initialisation logic on the back-end has a room for improvement and should be further improved/refined.

> And would it matter?

`mandate_data` is added to the payment intent request only in case we're confirming the intent, and Stripe uses it to validate that we've shown the terms to a user before the order placement. I think Stripe takes this parameter only if it's needed (e.g. when SEPA payment is used) and checks if we've sent it. For card payments, this parameter will be redundant and in my opinion won't be checked by Stripe. 

I think I need to change `$is_link_payment` name to something like `$is_link_enabled` to better reflect the reality. (done in 361cef7f534a4e9768715a61ed8fdbb49635448c) Other than that, I'm happy to refine this existing Stripe Link logic further if needed. Let me know your thoughts!

includes/class-wc-payment-gateway-wcpay.php
timur27: Sure, makes total sense, thank you! 765db74015e9af7ec324fb34e3d04fdb9ef7ed5a

includes/payment-methods/class-upe-split-payment-gateway.php
frosso: Thank you for the explanation! I like the rename you've done, I think it'll provide better clarity for future wanderers and adventurers 👍

frosso (approved):
Looking good, thanks for the changes!

PR #7233, opened 2023-09-19
RPP PoC: Base States

### LINKED ISSUE #6686: Payment Process: Payment object and states skeleton
### Description

Lay down the foundations (skeleton) for the payment object, and states, and call them from the payment service.

`Payment` should receive an initial state, and allow switching between states.

### Acceptance criteria

- A payment object exists, can receive an initial state, and can switch state.
- A base state _class_ exists, its constructor does **not** require a context (ready for DI).
- The context calls the necessary setter of the state when the state is provided.

### Testing instructions

All new methods should be covered by unit tests.

### Dev notes

Previously implemented in https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-f7e8f56394e718e1f5e0ba59e9f09d527e9c031122b31c2f3d20b74ac8b6e606 and https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-0f25837dc7983e0ac6dc5d794387e326fc9a376f7a2efa102585347425a32e32

### Additional context

Project: paJDYF-9hL-p2

RadoslavGeorgiev:
Fixes #6686

#### Testing instructions

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7233` or branch name `rpp/6686-payment-states` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/6686-payment-states) :rocket:

---

Build info: 

- Latest commit: 102f03d930c003a56d8dde5bc43a9d6e9fe39065
- Build time: 2023-09-27 07:48:37 UTC

src/Internal/Payment/Payment.php
```diff
@@ -0,0 +1,137 @@
...
+		return $this->order;
+	}
+
+	/**
+	 * Stores data within the payment.
+	 *
+	 * @param string $key   Key of the property to store.
+	 * @param mixed  $value Value to store.
+	 */
+	private function store( string $key, $value ) {
```
RadoslavGeorgiev: I'm thinking of the internal `store` and `read` methods to allow logging when data changes, instead of using properties, which is why those methods are here. Of course, logging is not in place yet.

src/Internal/Payment/Payment.php
```diff
@@ -0,0 +1,137 @@
...
+	 * processing, rather than controlling the flow.
+	 *
+	 * @param Flag $flag The flag to add.
+	 */
+	public function add_flag( Flag $flag ) {
+		$flags = $this->read( 'flags' ) ?? [];
+
+		if ( ! in_array( $flag, $flags, true ) ) {
+			$flags[] = $flag;
+			$this->store( 'flags', $flags );
```
RadoslavGeorgiev: This looks a bit ugly, but I wanted to make use of the `Base_Constant` here. Alternatively, we could use binary flags, similarly to [the previous PoC](https://github.com/Automattic/woocommerce-payments/blob/spike/payment-process-states/includes/payment/class-payment.php#L131).

src/Internal/Payment/Payment.php
```diff
@@ -0,0 +1,137 @@
...
+
+		$this->store( 'flags', $new_flags );
+	}
+
+	/**
+	 * Sets the used payment method.
+	 *
+	 * @param PaymentMethodInterface $payment_method The payment method to use.
+	 */
+	public function set_payment_method( PaymentMethodInterface $payment_method ) {
```
RadoslavGeorgiev, line 127: The context (`Payment` class) would still be used as the main storage of data, related to the payment. The payment method is an example payment property.

src/Internal/Payment/Response/ResponseInterface.php
```diff
@@ -0,0 +1,46 @@
...
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Payment\Response;
+
+/**
+ * Payment processing response interface, used to
+ * describe the response of processing a payment.
+ */
+interface ResponseInterface {
```
RadoslavGeorgiev: I'm showcasing the idea of a few small classes, representing the `ResponseInterface`, that will be used to communicate the result of processing with the gateway. This is caused by my desire to make the gateway unaware of payment states.

src/Internal/Payment/State/CompletedState.php
```diff
@@ -0,0 +1,25 @@
...
+ * Represents the payment in complete (successful) state.
+ */
+class CompletedState extends State {
+	/**
+	 * Returns the response, expected from a gateway's `process_payment()` method.
+	 *
+	 * @return ProcessingCompleted Whenever the current state implements the method.
+	 */
+	public function get_processing_response() {
+		$url = $this->get_context()->get_order()->get_checkout_order_received_url();
```
RadoslavGeorgiev: Kind of a side note: `get_checkout_order_received_url` can easily replace the usage of `get_return_url()`! 🥳

src/Internal/Payment/State/InitialState.php
```diff
@@ -0,0 +1,30 @@
...
+	 * Processes a new payment.
+	 *
+	 * @param PaymentRequest $request Payment request.
+	 */
+	public function process( PaymentRequest $request ) {
+		$context = $this->get_context();
+
+		// Transfer the necessary data from the request into the payment object.
+		$payment_method = $request->get_payment_method();
+		$context->set_payment_method( $payment_method );
```
RadoslavGeorgiev: Considering we'd have just the `process` method accessible externally, this might be the best place to transfer data from the `PaymentRequest` to the payment. WDYT?

src/Internal/Payment/State/InitialState.php
```diff
@@ -0,0 +1,30 @@
...
+
+		return $this->create_state( CompletedState::class );
```
RadoslavGeorgiev: This is not showcased in the example, but we could easily do the following:

```php
```

This could allow us to split states a bit internally, while still allowing the single external method to be used.

This might be useful for a couple of scenarios:

1. Existing successful payment intent, but incomplete payment. We'd be able to load the intent into the `Processed` state, and use its `complete()` method.
2. Something very similar for WooPay.

src/Internal/Payment/State/State.php
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	/**
+	 * Creates a new state, associated with the current payment.
+	 *
+	 * @param string $state_class Classname of the next state.
+	 * @return State
+	 */
+	protected function create_state( string $state_class ) {
+		$state = $this->state_factory->create_state( $state_class );
```
RadoslavGeorgiev, line 67: By storing the state factory here, we'd be able to not reference it anywhere else within the payment process.

src/Internal/Payment/Storage.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @var PaymentMethodService
+	 */
+	private $payment_method_service;
+
+	/**
+	 * Controls whether storage is actually enabled.
+	 *
+	 * @var bool
+	 */
+	private $enabled = false;
```
RadoslavGeorgiev, line 53: I think there's a mutual agreement that we wouldn't use payment storage from the start, but I'd like it to be baked in as early as possible.

src/Internal/Payment/Storage.php
```diff
@@ -0,0 +1,175 @@
...
+		if ( $this->enabled ) {
+			$data = $order->get_meta( self::ORDER_META_KEY );
+			if ( is_array( $data ) && ! empty( $data ) ) {
+				$state_class = $data['state'];
+				$this->import_payment_data( $payment, $data );
+			}
+		}
+
+		$state = $this->state_factory->create_state( $state_class );
+		$state->set_context( $payment );
```
RadoslavGeorgiev, line 103: This is how I imagine states being created, at least when we're always starting with `InitialState` or loading payments from storage.

src/Internal/Payment/Storage.php
```diff
@@ -0,0 +1,175 @@
...
+	 * @return State          Current state of the order payment.
+	 */
+	public function get_order_payment( WC_Order $order ) {
+		$payment     = new Payment( $order );
+		$state_class = InitialState::class;
+
+		if ( $this->enabled ) {
+			$data = $order->get_meta( self::ORDER_META_KEY );
+			if ( is_array( $data ) && ! empty( $data ) ) {
+				$state_class = $data['state'];
```
RadoslavGeorgiev, line 97: `$data['state']` is only read here. The idea is that while storing the payment, the class name of the state would be included, but it would not be stored within the payment otherwise, avoiding a link bank from `Payment` to `State`.

src/Internal/Payment/Storage.php
```diff
@@ -0,0 +1,175 @@
...
+
+		$data['flags'] = [];
+		foreach ( Flag::get_all() as $flag ) {
+			if ( $payment->has_flag( $flag ) ) {
+				$data['flags'][] = $flag->__toString();
+			}
+		}
+
+		return $data;
+	}
```
RadoslavGeorgiev: Serializing and deserializing payment data might make sense outside the payment object, keeping it as lightweight as possible.

src/Internal/Service/GatewayService.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * The gateway here is a dynamic dependency, accessed through the
+ * container, instead of a constructor parameter. That's because
+ * the gateway depends on various services, and there would be a
+ * circular dependency otherwise.
+ *
+ * Before adding the method you need here, please consider
+ * whether it could be extracted out of the gateway instead.
+ */
+class GatewayService {
```
RadoslavGeorgiev: For now this service is not needed, and I'd consider removing it. However, I don't want to delete it yet.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -8,19 +8,129 @@
...
+
+			$post_data = [
+				'payment_method'                             => WC_Payment_Gateway_WCPay::GATEWAY_ID,
+				'wcpay-payment-method'                       => 'pm_XYZ',
+				'wc-woocommerce_payments-new-payment-method' => true,
+			];
+
+			// Initialize the request and payment.
+			$request = new PaymentRequest( $this->legacy_proxy, $post_data );
+			$payment = $this->storage->get_order_payment( $order );
```
RadoslavGeorgiev: How about considering the current state as "payment" throughout methods like this one (`process_payment`)? It makes sense to me that externally, the process should be concerned about payment vs. state as little as possible?

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -8,19 +8,129 @@
...
+
+			// Setup flags.
+			if ( $request->should_save_payment_method() ) {
+				$payment->get_context()->add_flag( Flag::SAVE_PAYMENT_METHOD_TO_STORE() );
```
RadoslavGeorgiev: So far, flags seem like the only thing that would need to be provided straight to the context. We could consider a proxy method within states, but I'm not certain of its usefulness.

Alternatively, those could be provided as a parameter to `process`, considering that (at least without UPE), they are only required for payment processing, and should not change after calling it.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -8,19 +8,129 @@
...
+			}
+			if ( $manual_capture ) {
+				$payment->get_context()->add_flag( Flag::MANUAL_CAPTURE() );
+			}
+			if ( $this->woopay_utilities->should_save_platform_customer() ) {
+				do_action( 'woocommerce_payments_save_user_in_platform_checkout' );
+				$payment->get_context()->add_flag( Flag::SAVE_PAYMENT_METHOD_TO_PLATFORM() );
+			}
+			if ( $this->subscriptions_service->is_payment_recurring( $order ) ) {
```
RadoslavGeorgiev, line 100: Good news: Plenty of subs-related methods are contained within the utilities' trait, and accessible through the subscriptions service. This way, we can use the service until getting rid of the trait within the gateway.

RadoslavGeorgiev (commented):
Hey @marcinbot,

I've pushed some changes here. The main purpose was to see how things would work without the state pattern, wrapping states around the context. I am fully aware that the PR is in a very messy state, and I apologize for that, but I'd appreciate it if you could take a look.

To help navigation, there are a few key places and items worth looking into, mentioned in comments. I wouldn't mind jumping on a call to discuss them quickly.

src/Internal/Payment/Flag.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * All constants, used for payment flags.
+ *
+ * When adding constants here, increase multiply the last flag by two.
+ *
+ * This could live in the `Payment` class as well, but we
+ * should try to keep the main class undiluted.
+ */
+final class Flag extends Base_Constant {
```
marcinbot: What's the difference between the `Flag` and the already-existing `Factor`? They appear to overlap a lot.

src/Internal/Payment/Payment.php
```diff
@@ -0,0 +1,137 @@
...
+	private $data = [];
+
+	/**
+	 * Payment constructor.
+	 *
+	 * @param WC_Order $order Order requiring payment.
+	 */
+	public function __construct( WC_Order $order ) {
+		$this->order = $order;
+	}
```
marcinbot: Is `Payment` like a `payment_context` we discussed? If so, it might be worth to store scalar data only.

Especially here, PHP will pass objects by reference, but it's a bit of a hidden feature of the language, and the order object can't be serialized. I think it would be better to just use order ID. Alternatively, we could have our own data structure describing the order.

src/Container.php
```diff
@@ -67,12 +70,21 @@ public function __construct(
...
+	 * (it does not extend Throwable), and Psalm does not accept it.
+	 * @psalm-suppress MissingThrowsDocblock
 	 */
 	public function get( $id ) {
-		return $this->container->get( $id );
+		try {
+			return $this->container->get( $id );
+		} catch ( Throwable $e ) {
+			throw new ContainerException( $e->getMessage() );
+		}
```
marcinbot: I think this is a good candidate to take out to a separate, much smaller PR and to be discussed there.

I remember you mentioning that the container doesn't throw Throwables, and yet here you catch one and wrap it in a custom exception. Not quite sure what's going on, and probably this PR is not a place to discuss it.

src/Internal/Payment/Payment.php
marcinbot: I think that's fine, but `get/set` would be better - `store/read` makes me think of actually writing the data to a DB.

src/Internal/Payment/Payment.php
marcinbot: I'm still wondering why we need flags. Also, what do you think about having an array of booleans?

src/Internal/Payment/Payment.php
marcinbot: That's fine, and I understand it's just an example. I think we need to agree on terminology regarding "payment methods" before progressing here. It does appear that "payment instruments" is the most industry-accepted term.

src/Internal/Payment/State/CompletedState.php
marcinbot: This looks like a circular reference - the state or response shouldn't know about the gateway. It's the gateway that should look at the state/response and determine what kind of url to return.

A separate scenario is 3DS or anything that requires a special redirect. But even then, the gateway should be able to tell what a state/response is, and get the data from it necessary to construct the return url (again, in case of a 3ds it could even be a whole url, or url chunks).

It all comes down to the idea of wanting the State Machine to be abstract enough to be used in other places than just the gateway (REST API for example)

src/Internal/Payment/Response/ResponseInterface.php
marcinbot: Why would we want the gateway to be unaware of the states? The "awareness" of things should only go one way (the gateway calls the states, so is aware of them, but states are not aware of the gateway). With responses, we make the states aware of the gateway.

And I'm using the term "gateway" here as "anything that initiates the payment" (could be a service), which I think is also your meaning.

src/Internal/Payment/State/InitialState.php
marcinbot: What is the `PaymentRequest` here? I thought `Payment` is supposed to be a context, so why not transfer the data via the context?

src/Internal/Payment/State/InitialState.php
marcinbot: I think if it's not represented in the State Machine diagram, then it shouldn't be hacked into the code. I'm not sure I understand where the scenario 1 would happen (the successful intent should end up in a complete payment), but in general, such paths should be represented in proper states. There's no need to force a single external method. By saving a couple of lines of code in the gateway we'd introduce spaghetti in the State.

Another thing are the states I called "discrete" in my post and Matt D described as "internal" (I like that term). For example, `process` transition would result in a `Processing` state. Still, it should be actions/side-effects that trigger the transitions.

src/Internal/Payment/State/State.php
marcinbot: I'm still not sure about using DI for states at all. States are fairly simple objects, defining the transitions and storing the context. Triggering the side effects will be a problem, but how about we move the factory/DI stuff out of the PoC PRs and solve it separately when the time comes to emit events?

FWIW I think that even then, the states shouldn't take any dependencies in the constructor other than the context. Basically, it's the State driving the application flow and therefore the DI, not the other way around.

src/Internal/Payment/Storage.php
marcinbot: I think that's also fine, but we should also consider splitting the PRs and adding the storage separately.

src/Internal/Payment/Storage.php
marcinbot: Reading states will be tricky, but storing the class name or some other representation of the state will help. I think we might not need the State Factory, as mentioned in my other comment.

There's still possibility of bugs, like there not being enough data for a state to be initialized, but if we make all saving/reading go through one place and manage the context correctly, we should be fine.

Again, let's discuss in a separate PR, once we agree on and hopefully the basic State Machine approach.

src/Internal/Payment/Storage.php
```diff
@@ -0,0 +1,175 @@
...
+
+	/**
+	 * Loads or creates the payment object for an order,
+	 * returning the one true payment object.
+	 *
+	 * @param WC_Order $order Order that requires/has payment.
+	 * @return State          Current state of the order payment.
+	 */
+	public function get_order_payment( WC_Order $order ) {
+		$payment     = new Payment( $order );
```
marcinbot: This is the example of what's wrong with passing the order as an object I mentioned above. Also, it's a bit of a circular reference again: the `Payment` knows about the order and contains it, but is also stored in the order, so the order is more of a "parent" to the Payment, not the other way around. With scalar values we could just serialize and store the payment anywhere.

src/Internal/Payment/Payment.php
```diff
@@ -0,0 +1,137 @@
...
+
+use WC_Order;
+use WCPay\Internal\Payment\PaymentMethod\PaymentMethodInterface;
+
+/**
+ * Payment object.
+ *
+ * This is a data object, used as context and storage for payment states.
+ */
+class Payment {
```
marcinbot, line 17: If this is a payment context then we should call it `PaymentContext`. In general, we should start naming things properly in the PRs, i.e. states should be suffixed with `State` etc.

src/Internal/Service/PaymentProcessingService.php
marcinbot: It's probably fine, I'm hoping places like this one will not operate on these objects too much, just instantiate and call the transitions.

But the classes themselves should mention states, see my other comment.

src/Internal/Service/PaymentProcessingService.php
marcinbot: Ok, I think I understand what the `PaymentRequest` is here - it's an object representing the request made to the client backend, right? In this case, I think we should deserialize the request into the context, so it could be something like:
```
$payment_context = PaymentContext::from_request( $request );
```
This would help avoid all the ifs and add_flags here.

src/Internal/Payment/Flag.php
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	const MANUAL_CAPTURE = 'MANUAL_CAPTURE';
+
+	/**
+	 * Indicates if this payment will be recurring.
+	 * If not, it is just a single payment.
+	 *
+	 * @todo: Check if this flag is only required for the initial payment.
+	 */
+	const RECURRING = 'RECURRING';
```
marcinbot, line 38: In the interest of splitting the PRs we should probably not concern ourselves with flows other than a simple card payment for now. We should keep them in mind, but having them half-implemented is only confusing.

marcinbot (commented):
Thanks for working on this!

I left some comments, but I'm not sure what state this PR is at. This also seems similar to the example I provided (and the team agreed to spike on), but not quite, and I'm not sure if the differences are because this is a completely different approach, because it's half-done, or because there are valid reasons for them.

We should start thinking about splitting the PRs into shippable sizes, which will help with reviews and iterating on them. I think we should even consider closing this one and opening a bunch of smaller ones.

src/Internal/Payment/Payment.php
RadoslavGeorgiev: > I'm still wondering why we need flags.

A couple of reasons:

1. Flags represent minor deviations in the payment process, which do not warrant having separate code flows for them.
2. If those are not fags, but rather simple parameters, they would need to be passed around quite a bit until they reach the method/service, which actually cares about them. Also, it might make calling methods like `->process( $manual_capture, true, false, false, true )`.
3. By defining some of those flags in advance, and storing them within the context, we'd be able to reuse them later. For example, checking whether to save the payment method within the post-3DS-authentication flow.

> Also, what do you think about having an array of booleans?

For me, a flag is either present or not. If it is not present, keeping it in the array doesn't make too much sense. Maybe using `Base_Contant` here was useless though, I'll just switch to a plain class with constants.

src/Internal/Payment/Payment.php
RadoslavGeorgiev: This topic still needs some discussion and exploration. I think we also want to look into using `WC_Payment_Token` ([#](https://woocommerce.github.io/code-reference/classes/WC-Payment-Token.html)) and `WC_Payment_Token_CC`?

src/Internal/Payment/Response/ResponseInterface.php
RadoslavGeorgiev: > Why would we want the gateway to be unaware of the states? 

I think that ideally the gateway should care only about the response from the service, rather than its internals and states. For me, this is a way to decouple them.

> With responses, we make the states aware of the gateway.

I cannot agree with this statement. They are an independent set of classes, representing the outcome of the payment process, and are unaware both of the gateway, and of the internal workings of the payment process/context/state.

src/Internal/Payment/State/CompletedState.php
RadoslavGeorgiev: > It all comes down to the idea of wanting the State Machine to be abstract enough to be used in other places than just the gateway.

I agree with this statement, which is why the response shape is not specific to the gateway (array with `result` and `redirect`). Other places simply receive the URL for the next actionable screen, and it's up to them whether and how to use it. Even if the REST API is used to create the order, they could need the same URL of the 3DS verification screen, for example.

In either case, we could easily remove this URL from the response, and let the gateway figure it out.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: > What is the `PaymentRequest` here? 

It's [defined here](https://github.com/Automattic/woocommerce-payments/blob/a92200c01c9af1588c74ac8eeb88d44f1f04b207/src/Internal/Payment/PaymentRequest.php#L22). It's a layer outside the process, which simply accesses `$_POST`, escaping and sanitizing the provided input.

> I thought Payment is supposed to be a context, so why not transfer the data via the context?

The context should simply act as storage for working data, and should not know what to extract from the request. That's to say that data should be transferred _to_ the context. The question here is whether to do that outside of the state, or within the state.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: > I think if it's not represented in the State Machine diagram, then it shouldn't be hacked into the code.
> ... but in general, such paths should be represented in proper states.

Actually, we probably over-simplified the state diagram during our discussions throughout the meetup. While can drop the _Prepared_ and _Verified_ states, we will likely still need to have the `Processed` (not `Processing`) state, and be able to pick up the payment from it.

>  I'm not sure I understand where the scenario 1 would happen (the successful intent should end up in a complete payment)

It should, yet we ended up with duplicate payments. Let's use a network error on the of the intent way back from the server to the client as an example.

> There's no need to force a single external method. By saving a couple of lines of code in the gateway we'd introduce spaghetti in the State.

What's wrong with moving the state machine forward at the end of one of its transition? AFAICT the state machine we're planning to use is non-deterministic anyway.

> Another thing are the states I called "discrete" in my post and Matt D described as "internal" (I like that term). For example, process transition would result in a Processing state.

If we'd like to have discrete states here, wouldn't we in fact need one state to "automatically" proceed to the next one?

src/Internal/Payment/State/State.php
RadoslavGeorgiev: > Triggering the side effects will be a problem, but how about we move the factory/DI stuff out of the PoC PRs and solve it separately when the time comes to emit events?

This started as a PR, which was supposed to be merged, but is now indeed a PoC for multiple ideas at once. If we decide to remove dependencies in this discussion, they will not be included in the next PR.

> FWIW I think that even then, the states shouldn't take any dependencies in the constructor other than the context. Basically, it's the State driving the application flow and therefore the DI, not the other way around.

I am still not in favor of completely decoupling transitions, and the code driving them. I understand the desire to make the state machine as pure as possible, but doing so will make navigating and following the code much, much harder. Is it worth over-complicating things to avoid providing dependencies to states?

src/Internal/Payment/Storage.php
RadoslavGeorgiev: Yes, this PR ended up as a PoC, combining too many things into one.

src/Internal/Payment/Storage.php
RadoslavGeorgiev: > Reading states will be tricky

Why? If we are sure to store the state class, it should be easy to construct it afterwards.

> if we make all saving/reading go through one place and manage the context correctly, we should be fine.

That's the point of this class. Storage ended up being a secondary responsibility.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: > But the classes themselves should mention states, see my other comment.

I'm still to get there, but they do. 😕

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: > Ok, I think I understand what the PaymentRequest is here - it's an object representing the request made to the client backend, right?

Yes.

> In this case, I think we should deserialize the request into the context, so it could be something like...

Most flags (except the ones for saving a payment method) do not originate from the request, but are rather provided to the gateway.

Also, we'd be adding a responsibility to the payment context, which seems to be outside its scope. That's keeping in mind that we decided that static methods should be pure, and `from_request` is far from pure.

On the other hand, some flags (if we keep them as such, which is worth IMO) might change between requests. In those cases, we should strive to reuse the same payment context, meaning that we cannot use flags to construct it, but should add them once the object is available.

src/Internal/Payment/Flag.php
RadoslavGeorgiev: `Factor`s are disposable, and used to decide whether the new payment process should be entered at all. I intentionally want the two to be separate, so we can just go ahead and remove the routing functionality when the payment process is ready.

src/Internal/Payment/Payment.php
RadoslavGeorgiev: > Is `Payment` like a `payment_context` we discussed? 

Yes. It is the context.

> If so, it might be worth to store scalar data only.
> Especially here, PHP will pass objects by reference, but it's a bit of a hidden feature of the language, and the order object can't be serialized

No. We will need to store various objects within the context, incl. payment intents (or whatever abstracts/wraps them) to make them accessible between states. Storing an ID, and loading whatever it is (intent, order, payment, etc.) whenever needed, is not optimal.

It's worth considering that what we have here is just working data. That's why the `Storage` class has methods to import/export the necessary data from the payment, making them scalar.

With certain objects, especially orders, we should likely strive to be consistent, and intentionally try to use the same instance. Modifying different objects, and saving them individually, might be dangerous.

src/Container.php
RadoslavGeorgiev: This was supposed to be a small side fix in the original PR. That's because currently we don't have much capacity to quickly iterate over PRs, but we still need to maintain project velocity.

> I remember you mentioning that the container doesn't throw Throwables, and yet here you catch one and wrap it in a custom exception

It does indeed always throw throwables, but that's not how it is declared. That's why I'm catching what's partially undefined, and throwing another, properly traceable exception.

Anyway, that's the topic for a different discussion.

src/Internal/Payment/Storage.php
RadoslavGeorgiev: Again, the context here is working data, and not something that goes straight into the database. The `Storage` class makes sure that the order ID is not stored in the payment data (array) that goes into the database, and uses the order as a base for loading the payment.

Also, storing the order ID in the payment context, instead of storing the order, doesn't make it less of a circular reference, does it?

src/Internal/Payment/Payment.php
RadoslavGeorgiev: Yes, as far as states are concerned, the `Payment` object is their context. However, this doesn't mean that it's just a concept when looking at it globally.

Naming this `Payment` instead of `PaymentContext` is intentional. If we have a `PaymentState` and a `PaymentContext`, it sounds like there should be another class called `Payment` to link them, but there would be none.

And states are suffixed by `State`, am I missing something?

RadoslavGeorgiev:
> I'm not sure what state this PR is at.

It was meant to be the PR that introduced states to `src`, but altered to accommodate our discussions from the meetup. I wanted to put a few of the core concepts in the same place, allowing us to reach an agreement on the general approach first. Otherwise we might end up merging PRs, which will need to be overwritten later.

> This also seems similar to the example I provided (and the team agreed to spike on), but not quite, and I'm not sure if the differences are because this is a completely different approach, because it's half-done, or because there are valid reasons for them.

Though many things are half-baked in here, this is supposed to be what we agreed to spike, but with some further details to see what the full picture would look like. 

> We should start thinking about splitting the PRs into shippable sizes, which will help with reviews and iterating on them. I think we should even consider closing this one and opening a bunch of smaller ones.

Added `PoC` to the title to clarify that this is the plan. This PR is way past the shippable size already.

Anyway, I've responded to all threads, waiting for the next round of discussion.

RadoslavGeorgiev:
Base states have already been implemented within the payment process.

PR #7246, opened 2023-09-21
Add WooCommerce Multi-Currency support on WooPay

alefesouza:

Send the current currency to WooPay when WooCommerce Multi-Currency is installed.

#### Testing instructions

* Follow 2172-gh-Automattic/woopay testing instructions.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7246` or branch name `fix/woopay-woocommerce-multicurrency-support` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/woopay-woocommerce-multicurrency-support) :rocket:

---

Build info: 

- Latest commit: e75deaed47b8823aadef4c77ab3d3fd976801474
- Build time: 2023-11-01 22:41:41 UTC

tests/unit/woopay/test-class-woopay-adapted-extensions.php
```diff
@@ -139,4 +139,24 @@ public function test_get_gift_cards_data_while_logged_out_with_zero_balance() {
 
 		$this->assertEquals( $this->woopay_adapted_extensions->get_gift_cards_data( $this->test_user ), $expected );
 	}
+
+	public function test_get_extension_data() {
```
hsingyuc: Should we update the option `'woocommerce_currency'` to USD in case it's not USD so that we won't break the test in the future if the currency changes?

includes/woopay/class-woopay-adapted-extensions.php
```diff
@@ -19,6 +19,9 @@ class WooPay_Adapted_Extensions extends IntegrationRegistry {
 	const GIFT_CARDS_API            = 'woocommerce-gift-cards';
 	const GIFT_CARDS_BLOCKS         = 'wc-gift-cards-blocks';
 
+	const WOOCOMMERCE_MULTICURRENCY      = 'woocommerce-multicurrency';
+	const WOOCOMMERCE_MULTICURRENCY_PATH = 'woocommerce-multicurrency/woocommerce-multicurrency.php';
```
hsingyuc: Should we check if a constant in the plugin is defined so if the folder name is different then we can prevent critical error?

hsingyuc (commented):
@alefesouza Thank you for working on this.  As mentioned in the other PR, tests are as expected, but I couldn't place the order, so I also assigend @malithsen to review the PR.  I left two questions below.

includes/woopay/class-woopay-adapted-extensions.php
alefesouza: Good catch, fixed on https://github.com/Automattic/woocommerce-payments/pull/7246/commits/f3c67b1c24d42bd0496f94bafd753fd8b681b213.

tests/unit/woopay/test-class-woopay-adapted-extensions.php
alefesouza: On tests USD is the default one, I've tried using these codes but no success changing the currency during the tests:

```
add_filter(
	'pre_option_woocommerce_currency',
	function () {
		return 'BRL';
	}
);

add_filter(
	'woocommerce_currency',
	function () {
		return 'BRL';
	}
);

update_option( 'woocommerce_currency', 'BRL' );
```

includes/woopay/class-woopay-session.php
```diff
@@ -453,6 +453,7 @@ public static function ajax_init_woopay() {
 
 			$woopay_adapted_extensions  = new WooPay_Adapted_Extensions();
 			$body['adapted_extensions'] = $woopay_adapted_extensions->get_adapted_extensions_data( $email );
+			$body['extension_data']     = $woopay_adapted_extensions->get_extension_data();
```
malithsen: Do we need to set this and `adapted_extensions` here or do you think we can pass it in the `store_data` in https://github.com/Automattic/woocommerce-payments/blob/c76a59c993fb0bfe836281d819bef614d86194b7/includes/woopay/class-woopay-session.php#L389

alefesouza: We could pass it on store data, but this would not be backwards compatible as a version with `adapted_extensions` was already released, the WooPay side would need to have a condition for both cases until we remove support for older versions, what do you think?

malithsen: ah true. I like the idea of consolidating data passing to WooPay in a single place. If you agree, we can perhaps move `extension_data` data into `store_data` in this PR and migrate `adapted_extensions` in a separate PR.

tests/unit/woopay/test-class-woopay-adapted-extensions.php
hsingyuc: Gatcha

includes/woopay/class-woopay-session.php
hsingyuc: I agree with Malith that the `store_data` seems the right place to be.

alefesouza:
As discussed via Slack, currently on `trunk` the right currency is selected when redirecting on checkout page but it does not happen when using the express checkout button.

I've done some tests and noticed that if we reload the WooPay checkout page or open the same checkout url on a new tab it will load with the fallback currency.

This PR fixes this problem too by forcing the currency before the redirection, @malithsen could you review it again?

malithsen (commented):
@alefesouza I've been testing these changes with the first-party auth flow, which is now the default on `trunk`, and haven't been able to get this to work.
`session_init`, which the associated WooPay PR utilizes is not called during first-party auth flow, instead `create_unclaimed_session` is used. Although I haven't tested it, this is likely affecting adapted extensions as well.

Can you please have a look?

hsingyuc:
@alefesouza  I tested with the first-party auth flow and it doesn't work for me either.


https://github.com/Automattic/woocommerce-payments/assets/56378160/f9a65229-4c8a-4c1d-acdb-922ec1b2ea77

alefesouza:
@malithsen @hsingyuc fixed on https://github.com/Automattic/woocommerce-payments/pull/7246/commits/8305bfe0343b1003b81b91c59d2109cc959474b6 , the problem with adapted extensions already has a workaround implemented on https://github.com/Automattic/woocommerce-payments/pull/7455, which disables first party auth flow when one adapted extension is enabled, fixing it will be a little bit trickier as it depends on knowing the user email address before the redirect, for the `extension_data` field used by this PR there's no need for it.

malithsen (approved):

PR #7249, opened 2023-09-21
Add DisallowHooksInConstructor sniff and migrate some classes to use it

marcinbot:
Follow-up to https://github.com/Automattic/woocommerce-payments/pull/6326

Partial fix for https://github.com/Automattic/woocommerce-payments/issues/2769 and https://github.com/Automattic/woocommerce-payments/issues/5698

The aim is to automatically prevent the developers from using `add_filter` or `add_action` in class constructors. For now, most of the files are excluded from the sniffer, but new ones shouldn't be. There will be follow-up issues to migrate the remaining areas once this is shipped.

- Adds `dev/phpcs/Sniffs/DisallowHooksInConstructorSniff.php`
- Adds rules excluding most of the files 
- Migrated a couple of low-level classes as a proof-of-concept

#### Testing instructions

- Unit tests should pass
- PHPCS should pass
- Light smoke testing in the admin, especially after clearing the account cache. Everything should work

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -601,7 +601,7 @@ class WC_Payments {
 		}
 
 		if ( is_admin() && current_user_can( 'manage_woocommerce' ) ) {
-			new WC_Payments_Admin(
+			$admin = new WC_Payments_Admin(
 				self::$api_client,
 				self::get_gateway(),
 				self::$account,
@@ -610,16 +610,20 @@ class WC_Payments {
 				self::$incentives_service,
 				self::$database_cache
 			);
+			$admin->init_hooks();
 
-			new WC_Payments_Admin_Settings( self::get_gateway() );
+			$admin_settings = new WC_Payments_Admin_Settings( self::get_gateway() );
+			$admin_settings->init_hooks();
 
 			// Use tracks loader only in admin screens because it relies on WC_Tracks loaded by WC_Admin.
 			include_once WCPAY_ABSPATH . 'includes/admin/tracks/tracks-loader.php';
 
 			include_once __DIR__ . '/admin/class-wc-payments-admin-sections-overwrite.php';
-			new WC_Payments_Admin_Sections_Overwrite( self::get_account_service() );
+			$admin_sections_overwrite = new WC_Payments_Admin_Sections_Overwrite( self::get_account_service() );
+			$admin_sections_overwrite->init_hooks();
 
-			new WC_Payments_Status( self::get_gateway(), self::get_wc_payments_http(), self::get_account_service() );
+			$wcpay_status = new WC_Payments_Status( self::get_gateway(), self::get_wc_payments_http(), self::get_account_service() );
+			$wcpay_status->init_hooks();
 
 			new WCPay\Fraud_Prevention\Order_Fraud_And_Risk_Meta_Box( self::$order_service );
 		}
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -297,6 +297,7 @@ class WC_Payments {
 
 		include_once __DIR__ . '/class-database-cache.php';
 		self::$database_cache = new Database_Cache();
+		self::$database_cache->init_hooks();
 
 		include_once __DIR__ . '/class-wc-payments-dependency-service.php';
 
--- a/includes/class-wc-payments.php
+++ b/includes/class-wc-payments.php
@@ -302,6 +302,7 @@ class WC_Payments {
 		include_once __DIR__ . '/class-wc-payments-dependency-service.php';
 
 		self::$dependency_service = new WC_Payments_Dependency_Service();
+		self::$dependency_service->init_hooks();
 
 		if ( false === self::$dependency_service->has_valid_dependencies() ) {
 			return;
@@ -950,6 +951,7 @@ class WC_Payments {
 
 		if ( ! $http_class instanceof WC_Payments_Http_Interface ) {
 			$http_class = new WC_Payments_Http( new Automattic\Jetpack\Connection\Manager( 'woocommerce-payments' ) );
+			$http_class->init_hooks();
 		}
 
 		return $http_class;
--- a/changelog/add-constructor-hooks-sniffer
+++ b/changelog/add-constructor-hooks-sniffer
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Migrate away from hooking into actions in certain classes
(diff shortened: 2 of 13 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7249` or branch name `add/constructor-hooks-sniffer` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/constructor-hooks-sniffer) :rocket:

---

Build info: 

- Latest commit: 28e83e9f365f613a2dffe2337ce47ff5419205c0
- Build time: 2023-09-29 13:56:18 UTC

allie500:
Hey @marcinbot, is it okay to merge latest changes from `develop` into this branch?

allie500:
> > Hey @marcinbot, is it okay to merge latest changes from develop into this branch?
> 
> Sure, merged the changes, but reviewing the PR and landing it would be even better 😅

Thanks for merging them. I'm working on a PR on top of this one. It looks like there are two pending requests, waiting for Gamma and Transact Architecture to take a look at this PR.

jrodger (changes requested):
Adding a forbidden function to a constructor is being picked up, but at least on my machine it's generating a fatal error.

```
PHP Fatal error:  Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/File.php:1056
Stack trace:
#0 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/File.php(1056): vsprintf('Usage of add_fi...', 'function')
#1 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/File.php(672): PHP_CodeSniffer\Files\File->addMessage(true, 'Usage of add_fi...', 28, 9, 'WCPay.CodingSta...', 'function', 5, false)
#2 /Users/james/Projects/woocommerce-payments/dev/phpcs/Sniffs/DisallowHooksInConstructorSniff.php(64): PHP_CodeSniffer\Files\File->addError('Usage of add_fi...', 127, 'WCPay.CodingSta...', 'function')
#3 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/File.php(498): WCPay\CodingStandards\Sniffs\DisallowHooksInConstructorSniff->process(Object(PHP_CodeSniffer\Files\LocalFile), 103)
#4 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/LocalFile.php(92): PHP_CodeSniffer\Files\File->process()
#5 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Runner.php(632): PHP_CodeSniffer\Files\LocalFile->process()
#6 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Runner.php(504): PHP_CodeSniffer\Runner->processFile(Object(PHP_CodeSniffer\Files\LocalFile))
#7 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Runner.php(116): PHP_CodeSniffer\Runner->run()
#8 /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/bin/phpcs(18): PHP_CodeSniffer\Runner->runPHPCS()
#9 /Users/james/Projects/woocommerce-payments/vendor/bin/phpcs(120): include('/Users/james/Pr...')
#10 {main}
  thrown in /Users/james/Projects/woocommerce-payments/vendor/squizlabs/php_codesniffer/src/Files/File.php on line 1056
```

It looks like the last argument in the `addError` call is expected to be an array. This fixes it but I'm not sure if it's completely correct (wrapped the token content argument in an array):
```
$phpcsFile->addError(
	"Usage of $currentTokenContent in __construct() is not allowed",
	$i,
	'WCPay.CodingStandards.DisallowHooksInConstructor',
	[ $token['content'] ]
);
```
That argument looks like it's for providing values to a templated error message, does that mean we might not need to add the `$currentTokenContent` string into the error message ourselves?

We could fix this in a future PR, especially since it doesn't occur in the files you've included, but having a local function call also fails:

```
$this->add_filter();
```

How hard would that be to catch?

That also got me thinking, if that private `add_filter` function used the global `add_filter` the new sniff wouldn't pick that up right? And it's a bit beyond PHPCS to handle that sort of analysis?

includes/class-wc-payments.php
```diff
@@ -945,6 +951,7 @@ private static function get_wc_payments_http() {
 
 		if ( ! $http_class instanceof WC_Payments_Http_Interface ) {
 			$http_class = new WC_Payments_Http( new Automattic\Jetpack\Connection\Manager( 'woocommerce-payments' ) );
+			$http_class->init_hooks();
```
jrodger, line 954: This feels a little risky since there's a chance the `wc_payments_http` filter could be used somewhere and that instance won't have `init_hooks` called on it. It'll only be a problem if the class being used instead extends `WC_Payments_Http` and is expecting the hooks to be registered in the constructor, but that sounds possible?

jrodger:
I've run the linter, unit tests pass.

Clicking around the admin area doesn't generate any issues, clearing the account cache works fine. Settings can be changed and save correctly.

Just a few other questions:

- Do you see calling an `init_hooks` function on the class as the optimal solution? Or is it just a stepping stone that allows us to do something centralised if we wanted to?
- How are you handling unit tests? Just adding the init hooks call if the tests need the hooks wired up?

includes/class-wc-payments.php
marcinbot: I agree the HTTP class is a bit weird here, but this approach feels more right to me.

The override is created like this (a few lines above the highlighted line here):
```
$http_class = apply_filters( 'wc_payments_http', null );
```

It's this filter's responsibility to instantiate the class, and I think it should be its responsibility to init the hooks. Or the override class would create hooks in its constructor, which is really beyond our control since it's not the WCPay code anymore.

jrodger: Yep makes sense. I found an example of the filter being used and the returned class just implements the interface - which feels like the correct way to do this.

jrodger:
Thanks, I agree with all those points.

Let's get this merged so we can continue to work through the related issues.

jrodger (approved):

PR #7172, opened 2023-09-11
Add support for switching currencies when editing NYP price from cart

helgatheviking:
slight enhancement to NYP compatibility module

#### Changes proposed in this Pull Request

Adds param to the "Edit price" link for a NYP product in the cart. When clicked the user is redirected back to the product page with the current price pre-filled. If the user switches the currency at this point the pre-filled price will not change until this patch is implemented. 

#### Testing instructions
Going to need a slight patch in NYP 3.5.6 before can be tested.

Create simple NYP product
Add it to the cart in one currency
Go to cart and click on edit link, should redirect to single product page with price pre-filled
use a currency switcher to change currencies...

-------------------

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7269, opened 2023-09-22
Update Name Your Price compatibility to use new Compatibility methods

### LINKED ISSUE #5524: Multi-Currency Compatibility: Price conversion helper function

jessepearson:
Fixes #5524 #7172

This PR is the second part of #7202 that uses the new methods in order to allow Name Your Price amounts to be able to be switched on the cart and checkout pages.

Adds param to the "Edit price" link for a NYP product in the cart. When clicked the user is redirected back to the product page with the current price pre-filled. If the user switches the currency at this point the pre-filled price will not change until this patch is implemented.

#### Testing instructions

* Name Your Price can be obtained in the all-plugins repo.
* If you do not have more than one currency in your store, then navigate to WooCommerce > Settings > Multi-Currency and add another currency.
* Create simple NYP product.
* View the product, add an amount, and add it to the cart.
* Go to cart and click on edit link, should redirect to single product page with price pre-filled.
* Use the currency switcher to change currencies, the amount in the NYP field should update accordingly.

-------------------


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7269` or branch name `fix/7172-nyp-cart-editing` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7172-nyp-cart-editing) :rocket:

---

Build info: 

- Latest commit: 8f96960d8542cf0f9b04c2e24c7921cab2ecde07
- Build time: 2023-10-04 12:56:43 UTC

jessepearson:
@helgatheviking  I just sent you a DM in slack with a link to a zip file, can you check it out and see if all the expected functionality is there? It takes the edits in the PR (which are copied from your PR #7172 ) and adds them on top of my other changes I've done.

helgatheviking:
@jessepearson lgtm

includes/multi-currency/Compatibility/WooCommerceNameYourPrice.php
```diff
@@ -140,4 +137,43 @@ public function should_convert_product_price( bool $return, $product ): bool {
...
+	 *
+	 * @param string $initial_price The initial price.
+	 * @param mixed  $product       The product being queried.
+	 * @param string $suffix        The suffix needed for composites and bundles.
+	 *
+	 * @return float|string
+	 */
+	public function get_initial_price( $initial_price, $product, $suffix ) {
+
+		if ( isset( $_REQUEST[ 'nyp_raw' . $suffix ] ) && isset( $_REQUEST['nyp_currency'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
fashuvo, line 165: I was thinking if we can combine the isset checks and it turns we can do this actually. So, instead of 

```
if ( isset( $_REQUEST[ 'nyp_raw' . $suffix ] ) && isset( $_REQUEST['nyp_currency'] ) )
```

We could write 

```
isset($_REQUEST['nyp_raw' . $suffix], $_REQUEST['nyp_currency'])
```

And this works the same. The `isset` still returns true if both `$_REQUEST['nyp_raw' . $suffix]` and `$_REQUEST['nyp_currency']` are set and not null. 

Do you think we could implement this?

fashuvo, line 165: This is for my own learning - why did we decide do skip the nonce check here?

fashuvo (commented):
_This is not a blocker for the PR. I stumbled upon this while looking for PRs from other teams to review._

Thank you for your work on this, Jesse. Please feel free to respond to my comments whenever it's convenient for you.

includes/multi-currency/Compatibility/WooCommerceNameYourPrice.php
jessepearson: Keeping the separate `isset` requests makes the code easier to read when skimming through it. Either would work, though.

jessepearson: The nonce isn't needed since this process comes from the cart to the product with the values. If someone were trying to exploit this, the worst they could do is greatly increase the cost of a product that is then editable by the customer anyway.

fashuvo: I agree. It doesn't make sense to do the nonce check in this case. Thanks for the clarification 🙇

tpaksu (approved):
Tested various scenarios with the NYP plugin, and all tests passed: 
- Added only the NYP product to the cart
- Added another regular product alongside the NYP product
- Tested changing currencies on the cart
- Tested changing currencies on the edit price page
- Tested changing currencies after editing the price
- Tested changing currencies on the checkout page

And all showed the correct price. Thanks for working on this! 

LGTM, 🚢 it!

PR #7282, opened 2023-09-25
Introduce Klarna PM

### LINKED ISSUE #7243: [Klarna] Introduce necessary assets to enable Klarna in WooPayments interfaces
#### Merchant's scenarios

 - As a merchant I want to be able to manage Klarna payment method on the WooPayments settings screen. I want to see it in the list of BNPL providers to choose from, see its status and enable or disable it in my store.
 - As a merchant I want to be able to see which transactions were made with Klarna payment method on the transactions list and in the list of transactions included with deposit.
 - As a merchant I want to be able to refund orders paid with Klarma payment method from the order details page, same as with other payment methods.

#### Acceptance criteria

 - Klarna payment method is visible on the WooPayments settings screen for an enrolled merchant with a checkbox indicating its current status.
 - It is possible to enable or disable payment method and setting persists upon saving and reloading the page.
 - Payments made with Klarna payment method have an appropriate payment method icon in the transactions list and deposit details screens.
 - Orders paid with Klarna can be refunded from the Order details screen.

#### Shopper's scenario

As an eligible shopper I want to be able to use Klarna payment method on checkout, if it is supported by merchant.

#### Acceptance criteria

 - Klarna payment method is available for domestic shopper if merchant has it enabled.
 - Checkout with Klarna takes shopper to a Klarna webpage to work out details of the payment.
 - Upon completion payment is recorded with an appropriate status and shopper sees an "order received" page if it is successful.

KarlisJ:
Fixes #7243 

Introduces Klarna payment method. 

- https://github.com/Automattic/woocommerce-payments/issues/7319#top
- We're waiting for the icon from the Design team

#### Testing instructions

- Enable Klarna payments Capability for your test account ( use https://github.com/Automattic/woocommerce-payments-server/pull/3968#top)
- Refresh account cache
- Go to `payments`->`settings` and enable Klarna PM
- Make a test payment (you'll be asked to create a test account. 
   - Use CC from [here](https://docs.klarna.com/resources/test-environment/sample-payment-data/))
   - Use a valid phone number for the country. For example, 8332806812 for the US
- Repeat test payment from Blocks checkouts 
- Repeat test payment from Legacy, Split, Deferred UPE 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7282` or branch name `add/7243-introduce-klarna-pm` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/7243-introduce-klarna-pm) :rocket:

---

Build info: 

- Latest commit: 1cec35ea2c13129320710ce3291a00071cf730df
- Build time: 2023-10-05 08:46:32 UTC

includes/payment-methods/class-klarna-payment-method.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	public function __construct( $token_service ) {
+		parent::__construct( $token_service );
+		$this->stripe_id                    = self::PAYMENT_METHOD_STRIPE_ID;
+		$this->title                        = __( 'Klarna', 'woocommerce-payments' );
+		$this->is_reusable                  = false;
+		$this->icon_url                     = plugins_url( 'assets/images/payment-methods/klarna.svg', WCPAY_PLUGIN_FILE );
+		$this->currencies                   = [ 'AUD', 'CAD', 'CHF', 'CZK', 'DKK', 'EUR', 'GBP', 'NOK', 'NZD', 'PLN', 'SEK', 'USD' ];
+		$this->accept_only_domestic_payment = true;
+		$this->countries                    = [ 'AU', 'AT', 'BE', 'CA', 'CZ', 'DK', 'FI', 'FR', 'GR', 'DE', 'IE', 'IT', 'NL', 'NZ', 'NO', 'PL', 'PT', 'ES', 'SE', 'CH', 'GB', 'US' ];
```
kalessil: Countries need to be confirmed yet - we don't know which of them  has fees  finalized yet.

includes/class-wc-payments.php
```diff
@@ -505,6 +507,7 @@ public static function init() {
 			Affirm_Payment_Method::class,
 			Afterpay_Payment_Method::class,
 			JCB_Payment_Method::class,
+			Klarna_Payment_Method::class,
```
kalessil, line 520: Can we move it one line above please, so it was grouped with other BNPLs?

includes/class-wc-payments.php
```diff
@@ -396,6 +397,7 @@ public static function init() {
 		include_once __DIR__ . '/payment-methods/class-affirm-payment-method.php';
 		include_once __DIR__ . '/payment-methods/class-afterpay-payment-method.php';
 		include_once __DIR__ . '/payment-methods/class-jcb-payment-method.php';
+		include_once __DIR__ . '/payment-methods/class-klarna-payment-method.php';
```
kalessil, line 408: Can we move it one line above please, so it was grouped with other BNPLs?

client/payment-methods-map.tsx
```diff
@@ -247,6 +248,37 @@ const PaymentMethodInformationObject: Record<
...
 	},
+	klarna: {
+		id: 'klarna',
+		label: __( 'Klarna', 'woocommerce-payments' ),
+		brandTitles: {
+			affirm: __( 'Klarna', 'woocommerce-payments' ),
+		},
+		description: __(
+			// translators: %s is the store currency.
+			// TODO: Is "Available to all customers paying in %s." correct for BNPLs given the introduction of domestic check for PMs?.
```
kalessil: IDK, WDYT @dmvrtx?

client/payment-methods-map.tsx
```diff
@@ -19,6 +19,7 @@ import BankDebitIcon from 'assets/images/payment-methods/bank-debit.svg?asset';
 import AffirmIcon from 'assets/images/payment-methods/affirm.svg?asset';
 import AfterpayIcon from 'assets/images/payment-methods/afterpay.svg?asset';
 import JCBIcon from 'assets/images/payment-methods/jcb.svg?asset';
+import KlarnaIcon from 'assets/images/payment-methods/klarna.svg?asset';
```
kalessil, line 22: - Is it empty now? 
- Did we update the assets needed for transaction listing?
- Do we need to update onboarding-related banners?

kalessil (changes requested):
I did quick-check, leaving nit-picks and questions. But overall looks straightforward and clean!

includes/payment-methods/class-klarna-payment-method.php
KarlisJ: Yes, I've noted that to myself in the description of PR as well.

client/payment-methods-map.tsx
KarlisJ: > Is it empty now?

Yes, we're waiting for icon from Design team

> Did we update the assets needed for transaction listing?
Do we need to update onboarding-related banners?

No and No, both are handled as soon as we have the icon in place. We only need to double-check that the icon is aligned correctly in all cases which I'll do once we have the icon.

kalessil: Perfect, thanks for clarifying!

KarlisJ:
Starting from 6.4.0 version deferred Split UPE is the default UPE type but we haven't implemented the logic to restrict payment method to specific countries in it.

I've created a related bug https://github.com/Automattic/woocommerce-payments/issues/7313 that needs to be addressed as part of Klarna launch but to keep things separate, I'll tackle that as separate task.

mgascam (commented):
Code changes read well and I manually tested and was able to complete the checkout with Klarna. I noticed a couple of things:

- Since we still need to include Klarna assets, we can take the chance to add Klarna logo in the banner that appears when UPE is not enabled in the payment settings. I checked with designers in the slack thread: p1695896798254629/1695318697.479009-slack-C04VCT34PCP
- Noticed an error when the Legacy UPE is enabled and the shopper does not fill the phone number. I created this ticket https://github.com/Automattic/woocommerce-payments/issues/7319 to look into it.

client/payment-methods-map.tsx
dmvrtx: I'd change it to something like "Available for orders placed in the following currency: %s".

dmvrtx:
Did testing and the success path worked well with block checkout. Will do another round next week.

client/payment-methods-map.tsx
kalessil: @KarlisJ: did we clarify the TODO here?

client/payment-methods-map.tsx
```diff
@@ -247,6 +248,37 @@ const PaymentMethodInformationObject: Record<
...
+			affirm: __( 'Klarna', 'woocommerce-payments' ),
+		},
+		description: __(
+			// translators: %s is the store currency.
+			// TODO: Is "Available to all customers paying in %s." correct for BNPLs given the introduction of domestic check for PMs?.
+			'Allow customers to pay over time with Klarna. Available to all customers paying in %s.',
+			'woocommerce-payments'
+		),
+		icon: iconComponent( KlarnaIcon, 'Klarna' ),
+		currencies: [
```
kalessil: This is probably is outdated now.

KarlisJ: Yes, good catch, thank you. Another example of why need that "one source-of-truth" implemenation

KarlisJ:
@dmvrtx 
> Did testing and the success path worked well with block checkout. Will do another round next week.

As far as happy path testing, I've tested from different merchant countries and currencies for all (flavors of UPE) x blocks/classic.

As for unhappy path, the biggest problem is that we always show generic error message upon redirecting back from provider page. I'll create a ticket for that as it impacts not only Klarna and should be tackled separately.

client/payment-methods-map.tsx
kalessil: @dmvrtx: I was referencing to TODO. I expect it to be dropped from PR and converted into a dedicated issue instead if applicable.

kalessil (approved):
LGTM (code review).

Notes:
- I tested against prod-alike server state (no fees, trunk) - settings/checkout works
- I noted before, but we might need to update a banner as part of GA:
Screenshot from 2023-10-02 15-50-17

Pre-approving.

mgascam (commented):
Did a new manual testing round of happy path with shortcode based checkout + Legacy and Deferred UPE. Works as expected (I'm using US based merchant account for my tests).

But I noticed that the Klarna icon is missing in the Source column of the transaction list and also not showing in payment details page:

Screenshot 2023-10-02 at 18 45 41

Screenshot 2023-10-02 at 18 46 08

@KarlisJ @dmvrtx are we adding those icons in a separate PR?

dmvrtx:
> @KarlisJ @dmvrtx are we adding those icons in a separate PR?

No, the intent was to add all assets in one PR. Unless it turns out to be a complex issue, I'd rather like to see it here.

KarlisJ:
@mgascam The asset is there. The problem seems that the payment method is not recognised in those views. I'll investigate, thanks for finding this.

mgascam:
> @mgascam The asset is there. The problem seems that the payment method is not recognised in those views. I'll investigate, thanks for finding this.

Thanks @KarlisJ  I think we need to add the CSS rule for the icon, like https://github.com/Automattic/woocommerce-payments/blob/c8190e0efd693faa8b948279b26d0148badcf72d/assets/css/admin.css#L127

KarlisJ:
@mgascam Thank you, that was the missing thing. I've added the class. 

See my nest commit where I add the whole "Payment Method" block to the transaction details page. We missed that for Affirm/Afterpay. https://github.com/Automattic/woocommerce-payments/issues/7343

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+		formattedAddress,
+		paymentMethodCategory,
+		preferredLocale,
+	};
+};
+
+/**
+ * Placeholders to display while loading.
+ */
+const paymentMethodPlaceholders = {
```
mgascam, line 53: The keys in `paymentMethodPlaceholders` do not match with the `details` object we use below. I believe we can remove `fingerprint` and include `paymentMethodCategory` and `preferredLocale `.

EDIT: `verifiedName` and `country` can be removed as well.

client/payment-details/payment-method/index.js
```diff
@@ -20,6 +20,7 @@ import IdealDetails from './ideal';
 import P24Details from './p24';
 import SepaDetails from './sepa';
 import SofortDetails from './sofort';
+import KlarnaDetails from 'wcpay/payment-details/payment-method/klarna';
```
dmvrtx: Why path is so different here? Is it possible to make it consistent with other details components?

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+				>
+					{ id }
+				</Detail>
+
+				<Detail
+					isLoading={ isLoading }
+					label={ __( 'Category', 'woocommerce-payments' ) }
+				>
+					{ paymentMethodCategory }
+				</Detail>
```
dmvrtx, line 91: Nice touch! I wonder if we should add mapping somewhere to show possible values of `pay_later`, `pay_now`, `pay_with_financing`, or `pay_in_installments` as a readable/translatable text?

client/payment-methods-map.tsx
dmvrtx: +1. Agree with moving it into a separate issue.

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+					label={ __( 'ID', 'woocommerce-payments' ) }
+				>
+					{ id }
+				</Detail>
+
+				<Detail
+					isLoading={ isLoading }
+					label={ __( 'Category', 'woocommerce-payments' ) }
+				>
+					{ paymentMethodCategory }
```
mgascam, line 90: I wonder if we should translate the paymentMethodCategory? According to Stripe [docs](https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-klarna-payment_method_category) this can be `pay_later`, `pay_now`, `pay_with_financing`, `pay_in_installments `.

client/types/charges.d.ts
```diff
@@ -41,7 +41,8 @@ export interface PaymentMethodDetails {
 		| 'ideal'
 		| 'p24'
 		| 'sepa_debit'
-		| 'sofort';
+		| 'sofort'
+		| 'klarna';
```
dmvrtx: Will you please create an issue for a similar change for Affirm and Afterpay once we are clear with this one?

dmvrtx (approved):
Provisionally approved, as comments are non-blocking.

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+ */
+const formatPaymentMethodDetails = ( charge ) => {
+	const { billing_details: billingDetails, payment_method: id } = charge;
+
+	const {
+		country: countryCode,
+		verified_name: verifiedName,
+		payment_method_category: paymentMethodCategory,
+		preferred_locale: preferredLocale,
+	} = charge.payment_method_details.klarna;
```
mgascam, line 25: Looking at Stripe [docs](https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-klarna), the `country` and `verified_name` are not informed for Klarna so I believe we can remove them.

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+	formattedAddress: 'address placeholder',
+	country: 'country placeholder',
+};
+
+const KlarnaDetails = ( { charge = {}, isLoading } ) => {
+	const details = charge.payment_method_details
+		? formatPaymentMethodDetails( charge )
+		: paymentMethodPlaceholders;
+
+	const {
```
mgascam, line 67: Following my previous comment about Klarna payment method details, we could remove `verifiedName` and `country`

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+
+				<Detail
+					isLoading={ isLoading }
+					label={ __( 'Preferred Locale', 'woocommerce-payments' ) }
+				>
+					{ preferredLocale }
+				</Detail>
+			</div>
+
+			<div className="payment-method-details__column">
```
mgascam, line 101: I think we can remove this section

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,146 @@
...
+					label={ __( 'Address', 'woocommerce-payments' ) }
+				>
+					<span
+						dangerouslySetInnerHTML={ {
+							__html: formattedAddress,
+						} }
+					/>
+				</Detail>
+
+				<Detail
```
mgascam: And this section as well

mgascam (commented):
Leaving some comments after reviewing the payment method details section (great catch!)

client/payment-details/payment-method/klarna/index.js
KarlisJ: We should probably TS it. But I'm not sure about translating them. At the end of the day, those are codes not messages. If merchant tries to figure out what it means, and it's translated, then it will be hard to Google it.

client/payment-details/payment-method/index.js
KarlisJ: My bad, IDE...

client/payment-details/payment-method/klarna/index.js
KarlisJ: See https://github.com/Automattic/woocommerce-payments/pull/7282#discussion_r1343950205

client/payment-methods-map.tsx
KarlisJ: I looked more into the wording here and it actually seems to be intentional and correct as we allow only domestic payments.

client/types/charges.d.ts
KarlisJ: I don't follow you. Affirm/Afterpay is defined as first in the list.

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,123 @@
...
+const formatPaymentMethodDetails = ( charge ) => {
+	const { billing_details: billingDetails, payment_method: id } = charge;
+
+	const {
+		payment_method_category: paymentMethodCategory,
+		preferred_locale: preferredLocale,
+	} = charge.payment_method_details.klarna;
+
+	const { name, email, formatted_address: formattedAddress } = billingDetails;
+	// Use the full country name.
```
mgascam: It seems we no longer need this comment

client/payment-details/payment-method/klarna/index.js
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Internal dependencies.
+ */
+import PaymentDetailsPaymentMethodDetail from '../detail';
```
mgascam: Since `PaymentDetailsPaymentMethodDetail` is the default export you can import it as `Detail`. This way you don't need to re-assign it later.

```suggestion
import Detail from '../detail';
```

mgascam (approved):
LGTM! 

I left a couple of minor comments.

client/types/charges.d.ts
dmvrtx: My bad, I over scrolled. I think this list is sorted, though, can we keep that approach and place `klarna` after `ideal`, where it belongs?

client/payment-details/payment-method/klarna/index.js
dmvrtx: Stripe translates those in their dashboard, I don't think we should hesitate to do the same.

includes/payment-methods/class-klarna-payment-method.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	public function __construct( $token_service ) {
+		parent::__construct( $token_service );
+		$this->stripe_id                    = self::PAYMENT_METHOD_STRIPE_ID;
+		$this->title                        = __( 'Klarna', 'woocommerce-payments' );
+		$this->is_reusable                  = false;
+		$this->icon_url                     = plugins_url( 'assets/images/payment-methods/klarna.svg', WCPAY_PLUGIN_FILE );
+		$this->currencies                   = [ 'USD', 'GBP', 'EUR' ];
+		$this->accept_only_domestic_payment = true;
+		$this->countries                    = [ 'US', 'GB', 'AT', 'DE', 'NL', 'BE', 'ES', 'IT' ];
```
kalessil, line 34: We miss ranges per currency here, here's [a sample](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-afterpay-payment-method.php#L33).

kalessil (changes requested):
Something needs to be added.

includes/payment-methods/class-klarna-payment-method.php
KarlisJ: There's nothing about ranges in [Stripe's docs for Klarna](https://stripe.com/docs/payments/klarna). 

There are ranges in [Stripe's docs for Afterpay](https://stripe.com/docs/payments/afterpay-clearpay#collection-schedule). 

So, I left ranges out purposely.

kalessil: It's defined here: https://stripe.com/docs/payments/buy-now-pay-later#product-support

In the case of Klarna: 10 - 5000 for each currency.

kalessil: 5000 is the baseline, as we don't have tech. ability to fetch the one for the buyer we stick to the baseline.

KarlisJ: Thank you for pointing this out, implementing right on

KarlisJ: @kalessil Added in https://github.com/Automattic/woocommerce-payments/pull/7282/commits/1cec35ea2c13129320710ce3291a00071cf730df

kalessil: Fantastic, thank you @KarlisJ !

kalessil (approved):
LGTM (left a couple of comments about comments tweaks).

Issue #7289, opened 2023-09-26
Transaction Details → Dispute Details – Remove feature flag


Jinksi:
Part of #6883

We need to remove the feature flag `_wcpay_feature_dispute_on_transaction_page` to allow merchants to access dispute details on the Transaction Details screen.

- Create changelog entry
- [Update any affected critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows)
- [Create release testing instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions)

Jinksi:
I've moved the task "clean up unused code" to #7363

Issue #6928, opened 2023-08-04
Remove View dispute link in Transaction Details page timeline.


shendy-a8c:
![Screenshot 2023-08-04 at 09 27 23](https://github.com/Automattic/woocommerce-payments/assets/73803630/d28ee056-ac01-4809-b0cc-0ed43127d084)

Remove `View dispute` link at timeline.

While in there, please also change all timeline text's font size to 13px. Context: p1692552508057969-slack-C03LA109BKM

Jinksi:
I haven't progressed on this issue for a while, so I'll unassign myself from this issue to allow others to pick it up.

PR #7310, opened 2023-09-27
add a new page component for redirecting disputes => details:

### LINKED ISSUE #6891: Redirect legacy dispute details URL(s) to transaction details screen
We're moving all the relevant UI and actions for disputes from a dedicated screen (aka legacy dispute details) to the transaction details screen.

As part of that we'll need to ensure any links, bookmarks, navigations to legacy dispute details land on the relevant transaction details.

The URL(s) we need to redirect are:

- example URL: `http://{domain}/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes%2Fdetails&id={dispute_id}`
  - route: `wp-admin/admin.php?page=wc-admin`
  - path: `/payments/disputes/details`
  - params: `id`

Any navigation to that url needs to land on the relevant transaction details screen.

#### implementation tips from @Jinksi 
I think we can rename the [‘legacy’ dispute details page](https://github.com/Automattic/woocommerce-payments/blob/feda536268e6254fd63167b23e842d3e2e336ca0/client/disputes/details/index.tsx) and change its behavior to do a Javascript redirect to the transaction details page. 

All the information we need to get the transaction details link is available in the dispute object:

- Payment intent id (pi_abc)
- Transaction id (txn_abc)

The rename needs to be applied [here](https://github.com/Automattic/woocommerce-payments/blob/feda536268e6254fd63167b23e842d3e2e336ca0/client/index.js#L154) as well.

haszari:
Fixes #6891

~~This is ready for review though I haven't fully updated the PR description. Reviewer(s) please ping me / review harshly if any info is missing or not clear!~~  

This PR changes behaviour with `_wcpay_feature_dispute_on_transaction_page` enabled, and should preserve existing behaviour with flag off or missing.

- registers a new handler component for dispute details - `RedirectToTransactionDetails`
- that handler generates the relevant url and redirects there
- uses a notice and spinner component to reassure the user while the redirect happens. This takes 2-10 seconds. In my testing this delay only applies to first nav to that specific dispute. I'm assuming there's some caching somewhere.

Screenshot 2023-09-29 at 11 44 58 AM

https://github.com/Automattic/woocommerce-payments/assets/4167300/4f88eeba-c03e-4dee-b7c2-ff8b9a39238d

Still to do:
- ~~get the redirect working~~
- ~~get the user experience working smoothly~~
  - currently takes time to generate the URL
  - we may need an interstitial "one moment please" UI
- ~~ensure works correctly with feature flag~~ `_wcpay_feature_dispute_on_transaction_page` on/off
- make it robust – we dig deep into dispute object for required url params, so we need to fail gracefully if anything's missing
  - I don't think there are any blockers here - look out when reviewing please

#### Testing instructions

- As shopper, purchase products with dispute cards so merchant has disputed payments.
- Test navigating to dispute list view and clicking disputes with feature flag on and off.

Feature flag: `_wcpay_feature_dispute_on_transaction_page`

e.g. set with WP CLI `docker-compose exec wordpress wp --allow-root option set _wcpay_feature_dispute_on_transaction_page 1`

###### `_wcpay_feature_dispute_on_transaction_page` enabled
- Clicking any dispute should ultimately land on relevant transaction details page, including info about the dispute.
- You may see an interstitial redirect progress UI in the middle. 

###### `_wcpay_feature_dispute_on_transaction_page` disabled or missing
- Clicking any dispute should link direct to the relevant (legacy) dispute detail page. 

-------------------

**Post merge**


haszari:
I'm in the middle of this but I have no objections to anyone taking over at any time 😁

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7310` or branch name `fix/6891-redirect-dispute-detail-to-transaction` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/6891-redirect-dispute-detail-to-transaction) :rocket:

---

Build info: 

- Latest commit: 6216ae0d7b6d6ee020862b2c48eeca16b645265f
- Build time: 2023-10-03 22:43:17 UTC

client/disputes/redirect-to-transaction-details.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const RedirectToTransactionDetails = ( {
+	query: { id: disputeId },
+}: {
+	query: { id: string };
+} ): JSX.Element => {
+	const { dispute, isLoading } = useDispute( disputeId );
+	const disputeObject = dispute || ( {} as Dispute );
+	const disputeIsAvailable = ! isLoading && dispute && disputeObject.id;
+	// Why would dispute.charge ever be a string?
```
Jinksi: Our interfaces for Disputes are not 100% accurate here.:
* When fetching a charge, `charge.dispute.charge: string` 
* When fetching a dispute, `dispute.charge: Charge`

Currently, we use `Dispute` as the interface for both of these response types.

I've had a first crack at solving this in #7311 – unsure if it's the right way to go.

client/disputes/redirect-to-transaction-details.tsx
```diff
@@ -0,0 +1,51 @@
...
+	if ( disputeIsAvailable ) {
+		const paymentIntentId = disputeObject.charge;
+		transactionDetailsUrl = getAdminUrl( {
+			page: 'wc-admin',
+			path: '/payments/transactions/details',
+			id: chargeObject.payment_intent,
+			transaction_id: chargeObject.balance_transaction,
+			type: 'dispute',
+		} );
+		// window.location = transactionDetailsUrl;
```
Jinksi: To improve the UX, instead of a full-page reload, I'd recommend we use `@woocommerce/navigation` client-side routing, similar to the Transaction Details screen redirect.

https://github.com/Automattic/woocommerce-payments/blob/ee35b7685964a53db150de342d1f690598f2e736/client/payment-details/charge-details/index.tsx#L49-L59

client/disputes/redirect-to-transaction-details.tsx
haszari: Thanks for the tip - will take a look!

client/disputes/redirect-to-transaction-details.tsx
haszari: Nice – I can add that in right now to get this closer.

I think there will still be some slowness experienced by the user, as we need to pull the dispute (AJAX). That's ok though I wonder if we can avoid it. 

I guess that will get addressed by other PRs fixing the links elsewhere in admin, so maybe not a concern in practice.

haszari: Added here: 1409a1050c5273e21aeb13c00627a90948bf228f

The redirect is now working well, just need a little interstitial UI.

haszari:
> get the user experience working smoothly
> (currently takes time to generate the URL - we may need an interstitial "one moment please" UI)

@nikkivias do you know how we handle things like this? 

- When the merchant lands on the old (legacy) dispute details page, we're going to redirect them to the transaction details page. 
- The redirect takes a moment or two (seconds) so the dispute details page, or some page/UI, will flash up briefly.
- We need that UI to bridge the gap between what they clicked and where they will end up (transaction details) – i.e. interstitial UI.

I'm thinking we need a message like `Loading transaction…` or `Redirecting to transaction…` with a spinner of some sort.

nikivias:
I will look into how we might handle this @haszari  how many seconds is it currently taking to redirect?

haszari:
> how many seconds is it currently taking to redirect?

Good question @nikkivias! 2-5 seconds approximately. Here's a video :) 


https://github.com/Automattic/woocommerce-payments/assets/4167300/14ee52af-ca6e-4331-9526-1c2424b167da

Jinksi:
😆 Let's just stick with this, @haszari.

image

haszari:
FYI @nikkivias there's an updated prototype now using standard `<Notice><Spinner>` components. See video in PR description. 

I think this is almost shippable, especially since this UI will be very rarely seen (only on first nav to each unique dispute, and we are planning on fixing most links anyway). ~~The text and spinner alignment needs tweaking though~~ - now looking good using `Flex` components.

Let me know what you think.

client/disputes/redirect-to-transaction-details.tsx
haszari: Added some comments to the PR. This feels a bit like typescript chicanery at first glance. Hoping there's a simple non-hack solution! If we need to start sketching out hierarchical types that implies maintenance risk IMO (object oriented vibes).

haszari: I've updated my comment, since the type is correct in my view. The part I don't understand – is casting the correct approach an "OR" type like this? I.e. let me know if `as Charge` is a hack, or if I need a runtime type check.

client/disputes/details/index.tsx
```diff
@@ -27,11 +27,19 @@ const DisputeDetails = ( {
 	query: { id: disputeId },
 }: {
 	query: { id: string };
-} ): JSX.Element => {
+} ): JSX.Element | null => {
```
haszari: I'm (now) in the habit of declaring React components as `Element | null`, consistent with classic React. As I understand, returning null is totally fine way to not render anything. 

Question for front end experts – is this a good approach? Should we have a standard type for this? What does the TypeScript React community do?

client/disputes/redirect-to-transaction-details.tsx
```diff
@@ -0,0 +1,58 @@
...
+		<Page>
+			<Notice status="info" isDismissible={ false }>
+				<Flex justify="left">
+					<FlexItem>
+						<Spinner />
+					</FlexItem>
+					<FlexItem>
+						<span>Redirecting to payment details…</span>
+					</FlexItem>
+				</Flex>
```
haszari: Happy days using core components to get a nice, consistent look out of the box!

client/disputes/details/index.tsx
```diff
@@ -27,11 +27,19 @@ const DisputeDetails = ( {
...
 	const disputeIsAvailable = ! isLoading && dispute && disputeObject.id;
 
+	// Bail early and return nothing if the feature flag is not enabled.
+	// Here as a hint/reminder to delete this file when the feature flag is removed.
+	const isDisputeOnTransactionPageEnabled =
+		wcpaySettings.featureFlags.isDisputeOnTransactionPageEnabled;
+	if ( isDisputeOnTransactionPageEnabled ) {
+		return null;
+	}
+
```
haszari: Added this so when we're working on #7289 we find this class (search for feature flag) and remember to delete it. 🐘

client/disputes/details/index.tsx
```diff
@@ -27,11 +27,19 @@ const DisputeDetails = ( {
...
 	query: { id: string };
-} ): JSX.Element => {
+} ): JSX.Element | null => {
 	const { dispute, isLoading, doAccept } = useDispute( disputeId );
 	const disputeObject = dispute || ( {} as Dispute );
 	const disputeIsAvailable = ! isLoading && dispute && disputeObject.id;
 
+	// Bail early and return nothing if the feature flag is not enabled.
+	// Here as a hint/reminder to delete this file when the feature flag is removed.
+	const isDisputeOnTransactionPageEnabled =
```
shendy-a8c: Hmm.. I think this is not necessary because this component does not need to know the business logic of when it will be rendered or not. You already check the feature flag when registering the page, ie when feature flag is on, you register `RedirectToTransactionDetails` instead of `DisputeDetails`, so that itself guarantees this page will not be rendered at any case.

I cannot find any reference to this `DisputeDetails` other than in `client/index.js`.

If the purpose of this change is as a reminder to remove this legacy dispute details component, I'd say we can just mentioned it in https://github.com/Automattic/woocommerce-payments/issues/7289.

What do you think?

client/disputes/details/index.tsx
shendy-a8c: Not a frontend expert but since as of currently, there is a possibility this component will return null, this `| null` addition is necessary.

However, I have my opinion that checking for feature flag and returning null is not necessary here. Read along =)

client/disputes/details/index.tsx
shendy-a8c: To be explicit, I just added one check item in https://github.com/Automattic/woocommerce-payments/issues/7289:
> - [ ] Remove the [legacy dispute details page](https://github.com/Automattic/woocommerce-payments/blob/2eb4de5d4bca3573339e15197987a3943ffeddf4/client/disputes/details/index.tsx).

client/disputes/details/index.tsx
```diff
@@ -27,11 +27,19 @@ const DisputeDetails = ( {
...
+		wcpaySettings.featureFlags.isDisputeOnTransactionPageEnabled;
```
shendy-a8c: Maybe better to be explicit reading `window.wcpaySettings`.

```suggestion
		window.wcpaySettings.featureFlags.isDisputeOnTransactionPageEnabled;
```

client/index.js
```diff
@@ -150,24 +151,51 @@ addFilter(
...
+							],
+							__( 'Dispute details', 'woocommerce-payments' ),
+						],
+						navArgs: {
+							id: 'wc-payments-disputes-details',
+							parentPath: '/payments/disputes',
+						},
+						capability: 'manage_woocommerce',
+				  }
+		);
```
shendy-a8c: I was expecting only the `container` that needs to conditionally change between `RedirectToTransactionDetails` and `DisputeDetailsPage` based on whether `isDisputeOnTransactionPageEnabled` is on or not.

Why `navArgs.id` needs to be different?

Also, not sure what the correlation between this page registration in Javascript and [in PHP](https://github.com/Automattic/woocommerce-payments/blob/2eb4de5d4bca3573339e15197987a3943ffeddf4/includes/admin/class-wc-payments-admin.php#L468-L475).

Per the [docs](https://developer.woocommerce.com/extension-developer-guide/working-with-woocommerce-admin-pages/#registering-a-react-powered-page), react powered page needs to be registered both in Javascript and PHP.

shendy-a8c (dismissed):
I tested in places where it sends user to the dispute details page:
- On the transaction details page, there is a 'View dispute' link (that will be removed in https://github.com/Automattic/woocommerce-payments/issues/6928) in the timeline.
- Dispute list page.
- Accessing the URL (eg `/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes%2Fdetails&id=dp_abc`) directly (in the case of link from email).

Functionality wise, it works and UI looks good!

My feedbacks:
1. What happens if dispute id is invalid? For example `/wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes%2Fdetails&id=dp_abc`.

This is related to your point the the PR body:
> make it robust – we dig deep into dispute object for required url params, so we need to fail gracefully if anything's missing

IMO, either: a) Redirect to dispute list page maybe, or b) Show that it's an invalid link.


2. From your video you mentioned at https://github.com/Automattic/woocommerce-payments/pull/7310#issuecomment-1738428660, there is a 'transaction details' link. I kind of like that but acknowledge that it might tempt the user to click it making the UX a bit weird but having the link is good in case the redirect somehow does not happen. I wonder what your reason is taking that link out.

3. Some GH checks failed: need changelog and JS linting problem. I think the JS linting problem is caused by your direct read of `window.wcpaySettings.featureFlags.isDisputeOnTransactionPageEnabled`. Maybe the correct way is to use `useContext()` like [here](https://github.com/Automattic/woocommerce-payments/blob/2eb4de5d4bca3573339e15197987a3943ffeddf4/client/payment-details/summary/index.tsx#L172). I think that way, in unit test, we can set the feature flag on / off like [here](https://github.com/Automattic/woocommerce-payments/blob/2eb4de5d4bca3573339e15197987a3943ffeddf4/client/payment-details/summary/test/index.test.tsx#L466).

nikivias:
> FYI @nikkivias there's an updated prototype now using standard <Notice><Spinner> components. See video in PR description.

@haszari I don't think we're meant to use the banner component as a loading component. I was looking and asked around and there doesn't seem to be a known internal pattern for redirects from legacy.

I think the normal loading screen like this would suit best
Transactions List@2x

[See on Figma](https://www.figma.com/file/RiEQmaKRI7u9USAI3lyZz0VX/Dispute-Flows?type=design&node-id=7522%3A72790&mode=design&t=7DwUCScmfKuE1F25-1)


> this UI will be very rarely seen (only on first nav to each unique dispute, and we are planning on fixing most links anyway). 

Do you mean this only occurs during the first visit of each new dispute, and not thereafter? And when fixes are done the redirect will no longer happen?

haszari:
> I don't think we're meant to use the banner component as a loading component. I was looking and asked around and there doesn't seem to be a known internal pattern for redirects from legacy. I think the normal loading screen like this would suit best

Thanks @nikkivias – good idea!

I've updated to use this approach. Note I went with sentence case (no capital P for payment details). 

I didn't find any existing component / instance of this UI, so I've implemented as a one-off with some tweaks to ensure it's visible no matter what screen/device merchant is using.

Looks like this:

Screenshot 2023-10-02 at 11 43 28 AM
Screenshot 2023-10-02 at 11 43 16 AM

> > this UI will be very rarely seen (only on first nav to each unique dispute, and we are planning on fixing most links anyway).

> Do you mean this only occurs during the first visit of each new dispute, and not thereafter? And when fixes are done the redirect will no longer happen?

Yes – the main point is that this UI will be seen very rarely, so don't want to invest too much in it – just make sure it doesn't confuse merchant if they see it. I'm not 100% certain if the redirect is only on the first visit, but in my testing it's very hard to see the redirect UI after the first time 🤷🏻

haszari:
Remaining todos on this PR. FYI @Jinksi - these are the things I was going to look at before re-requesting review.

- ~Linter issues (haven't rechecked since last round of changes).~ Fixed in 54ac2ce
- Use of React context to access feature flag. I decided against this, I don't think it's a good use case for context (feature flags are global and don't change, window state is appropriate). However if the WCPay convention is to use context, or it's better for whatever reason, happy for that to change. _I agree with using `window` rather than `Context` here, the failing JS test seems to have been resolved too_
- ~I'm seeing a `Warning: Cannot update a component (hr) while rendering a different component (RedirectToTransactionDetails).` when the redirect happens. It seems likely that was introduced in this PR, if so need to fix.~ Fixed in 45349c1

client/disputes/redirect-to-transaction-details.tsx
Jinksi: I see type-casting as a risk in this scenario since it makes assumptions about the code that may not always be true. E.g. we may in the future change the response type of this endpoint to always return `dispute.charge: string` and not realise that this has been overridden (`as Charge`) throughout the codebase.

I think a more robust way to assert that `dispute.charge === Charge` would be via a [type guard function](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) e.g.:

```ts
const isCharge = ( charge: Charge | string ): charge is Charge => {
	return typeof charge !== 'string';
};
const { charge } = dispute;
if ( isCharge( charge ) ) {
	// charge.payment_intent exists!
```

However, [since we know that `GET disputes/{id}` endpoint will return a `dispute` with `dispute.charge: Charge`](https://github.com/Automattic/woocommerce-payments/blob/1a76dcefde87b8ae6f8e2bfefa5dad01b76ddba5/client/data/disputes/resolvers.js#L52), it would be better to fix this upstream (the `Dispute` interface) since we have full control here.

I think it's OK to follow this up in a future PR, however.

client/disputes/details/index.tsx
Jinksi: Note that when we use `React.FC`/`React.FunctionComponent`, the return type is implicit and we don't have to define it (see [TS guidelines doc](https://github.com/Automattic/woocommerce-payments/blob/develop/docs/typescript/react-components.md#declare-the-component-using-the-attribute-type-you-declared-in-your-interface)).

I've switched this to `React.FC` in 2c4dbfc.

Jinksi:
> What happens if dispute id is invalid? For example /wp-admin/admin.php?page=wc-admin&path=%2Fpayments%2Fdisputes%2Fdetails&id=dp_abc.

In addition, we aren't handling network errors, see screen recording:

https://github.com/Automattic/woocommerce-payments/assets/3147296/c432417e-f0e2-494e-a7dd-a66d5af84c94

I'm looking into fixing this. PS, this wasn't caught by TS because [we're currently casting `dispute: Dispute` as a return value of `useDispute`](https://github.com/Automattic/woocommerce-payments/blob/54ac2ce36a7ef086eeb073f77d297ebf5df9444b/client/data/disputes/hooks.ts#L37), whereas it is actually `dispute: Dispute | undefined`

client/disputes/redirect-to-transaction-details/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+				{ error ? (
+					<>
+						<FlexItem>
+							<Icon icon={ warning } type="warning" size={ 32 } />
+						</FlexItem>
+						<FlexItem>
+							<div>
+								<b>Error retrieving dispute</b>
+							</div>
+							<div>Please check your network and try again.</div>
```
Jinksi, line 55: We could consider giving more accurate error messages, e.g. looking at `error.code` to determine a suitable messages for `resource_missing` and `fetch_error`.

Jinksi:
I've handled network errors in bec0846, 5e006f8, 6111fac, 6b00887.

See screen recording example of using an unknown dispute id below.

https://github.com/Automattic/woocommerce-payments/assets/3147296/db07c02a-7b17-4c4f-843e-df324ee129b9

Another edge case that may or may not require investigating – what will happen if a live-mode dispute email is clicked and the site is in test-mode?

Jinksi (approved):
This tests well @haszari 🙌 

I've made some relatively significant additions to this PR, `getDispute` error handling in particular – I've approved this PR on the assumption that you're happy with the changes I've introduced 😅

client/data/disputes/hooks.ts
```diff
@@ -44,7 +49,7 @@ export const useDispute = (
 	const { acceptDispute } = useDispatch( STORE_NAME );
 	const doAccept = () => acceptDispute( id );
 
-	return { dispute, isLoading, doAccept };
+	return { dispute, isLoading, error, doAccept };
```
haszari, line 52: Adding support for handling dispute load errors is possibly overengineering for this use case (fallback redirect), but is useful in general so let's keep it :) 

I do wonder how we got to this point in the project without adding this!

client/data/disputes/test/resolvers.js
```diff
@@ -68,6 +69,9 @@ describe( 'getDispute resolver', () => {
 					expect.any( String )
 				)
 			);
+			expect( generator.next().value ).toEqual(
+				updateErrorForDispute( 'dp_mock1', undefined, errorResponse )
```
haszari, line 73: Curious what regressions tests like this would catch.

Jinksi: I guess it could help ensure that we’re calling the update function in the case of a refactor but it feels a bit like we’re testing implementation details here.

Note that this isn’t a new test, I added this here to fix a failing test after adding updateErrorForDispute to the getDispute resolver.

client/data/disputes/hooks.ts
Jinksi: The resolver shows a snackbar on error: “Error retrieving dispute”. I’d guess that this has been a sufficient way to handle errors until now.

client/data/disputes/test/resolvers.js
haszari: Gotcha, figured there would be a thread pulling you here 😁

client/data/disputes/hooks.ts
haszari: Right, so is there a chance we don't need this extra support? Happy for it to merge, just clarifying the value.

client/disputes/redirect-to-transaction-details/index.tsx
haszari: Not needed IMO, this redirect is a fallback so will hopefully not get much use. If merchant lands here and it doesn't work for any reason, they should be able to get back on track :)

client/disputes/redirect-to-transaction-details/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import './style.scss';
+import warning from 'wcpay/components/icons/warning';
+
+const RedirectToTransactionDetails: React.FC< { query: { id: string } } > = ( {
+	query: { id: disputeId },
+} ) => {
+	const { dispute, error, isLoading } = useDispute( disputeId );
+
+	useEffect( () => {
```
haszari, line 24: Doh! Is this the fix for my react issue! Obvious in hindsight that I would need to redirect as a side effect. Thanks for sorting this @Jinksi , major oversight on my side.

client/disputes/redirect-to-transaction-details.tsx
haszari: Thanks - I look forward to understanding this better in the upstream PR. 

This kind of thing feels very heavy compared to classic loose-duck-types JavaScript, so I want to really understand how to use TypeScript to strengthen things in a lean/elegant way.

client/disputes/details/index.tsx
haszari: Aha! Now I'm a fan of `React.FC` and will use from here on in. Can we encourage that pattern with a linter or tooling?

haszari: Looking at the doc, we might be able to distil this to a principle:

- `Use React.FC type for React components`

client/data/disputes/hooks.ts
Jinksi: Yep the value is low at the moment, we’re only checking for a fetch error in this component. Alternatively, we could move the error handling to the component, but I think it’s more predictable to handle dispute fetch errors in a way that is consistent with other wp.data resolvers.

haszari:
Thank you @Jinksi for getting this PR shipshape. Especially for `useEffect` & fixing the react console error 🤦

> Another edge case that may or may not require investigating – what will happen if a live-mode dispute email is clicked and the site is in test-mode?

I don't think we need to worry about any more edge cases, the only real goal here is that if merchant navs to a dispute, they get redirected to the correct transaction. If there are issues that mean we can't redirect, we have to trust that the merchant will be able to use the UI to get back on track – loading this component with low-risk contingencies is not worth the effort IMO.

haszari:
Merging!

PR #7311, opened 2023-09-27
Add correct TS interface for `charge.dispute`

Jinksi:

Our interfaces for Disputes are not 100% accurate:

* When fetching a charge, `charge.dispute.charge: string` 
* When fetching a dispute, `dispute.charge: Charge`

Currently, we use `Dispute` as the interface for both of these response types.

This PR introduces a `ChargeDispute` interface for `charge.dispute` objects.

#### Testing instructions

*

-------------------

**Post merge**

### DIFF
--- a/client/data/disputes/hooks.ts
+++ b/client/data/disputes/hooks.ts
@@ -52,7 +52,7 @@ export const useDispute = (
  * Does not return or fetch the dispute object.
  */
 export const useDisputeAccept = (
-	dispute: Dispute
+	dispute: Pick< Dispute, 'id' | 'payment_intent' >
 ): {
 	doAccept: () => void;
 	isLoading: boolean;
--- a/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
+++ b/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
@@ -22,8 +22,7 @@ import {
 /**
  * Internal dependencies
  */
-import type { Dispute } from 'wcpay/types/disputes';
-import type { ChargeBillingDetails } from 'wcpay/types/charges';
+import type { ChargeBillingDetails, ChargeDispute } from 'wcpay/types/charges';
 import wcpayTracks from 'tracks';
 import { useDisputeAccept } from 'wcpay/data';
 import { getDisputeFeeFormatted, isInquiry } from 'wcpay/disputes/utils';
@@ -36,7 +35,7 @@ import InlineNotice from 'components/inline-notice';
 import './style.scss';
 
 interface Props {
-	dispute: Dispute;
+	dispute: ChargeDispute;
 	customer: ChargeBillingDetails | null;
 	chargeCreated: number;
 }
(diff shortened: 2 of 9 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7311` or branch name `dev/discriminatory-type-for-charge-dispute` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=dev/discriminatory-type-for-charge-dispute) :rocket:

---

Build info: 

- Latest commit: b8f05fa00692a9a83663a77d0d36ea6bba4922aa
- Build time: 2023-09-28 23:41:49 UTC

haszari:
> When fetching a charge, charge.dispute.charge: string

> When fetching a dispute, dispute.charge: Charge


Is that the only difference between the two types?

Are they two variations of the same type ("Dispute") or are they distinct types?

At first glance, I'd look at two options:

- Dispute type is flexible enough to handle both variations – string or nested object for charge.
  - In this case, I'd wonder why the prop uses the same key name. 
  - `chargeId` and `charge` (each is optional / one is required) would be simple and clear.
- OR
- Charge type owns the nested dispute type - don't "borrow" from the real dispute type since it's a different thing. 

Curious where the difference comes from too. Is this how Stripe structures this data (and it transparently passes through WCPay Server), or is this defined in our local data structures?

Jinksi:
> Is that the only difference between the two types?
> Are they two variations of the same type ("Dispute") or are they distinct types?

Yes, the dispute that is part of the `Charge` is otherwise identical to a `dispute` fetched via the `/disputes/{id}` endpoint.

This is due to [our use](https://github.com/Automattic/woocommerce-payments-server/blob/9ffb16012f243989759b2ace33ab6bb49ee089b4/server/wp-content/rest-api-plugins/endpoints/wcpay/class-disputes-controller.php#L275) of [Stripe's `expand` param](https://stripe.com/docs/api/expanding_objects), which is also what causes the dispute `charge` value to change from a `string` to a `Charge` object with the same `charge` key.

image

> Dispute type is flexible enough to handle both variations – string or nested object for charge.

This is the current situation (`Dispute.charge: string | Charge`) but the downside is that we have to unnecessarily check if `dispute.charge !== string` when fetching `/disputes/{id}`.

I'd prefer us to aim for accurate TS interfaces so we can trust and rely on them and not have to find workarounds for them (e.g. `dispute.charge as Charge`), which was the catalyst for this discussion (https://github.com/Automattic/woocommerce-payments/pull/7310#discussion_r1343473582). But maybe this is a nice-to-have and not a priority. 😊

> `chargeId` and `charge` (each is optional / one is required) would be simple and clear.

This could be the best solution, and we can introduce this change server-side.

dmvrtx:
Thanks for the original work here. The underlying issue is still valid, but this draft had drifted too far from current develop and the surrounding dispute UI and types have changed enough that carrying this branch forward would add more churn than value.

We have opened #11607 as a fresh replacement built on top of the current codebase, with the same goal of separating the charge-embedded dispute shape from the fetched dispute resource shape. Closing this one in favor of that updated PR keeps the review surface smaller and makes the follow-up easier to validate.

PR #7314, opened 2023-09-28
Remove passing filter hook name from send() method in request classes

### LINKED ISSUE #6980: Platform - remove the filter-like signature from Request::send
As discussed, thinking of the filters puts the burden on all developers all the time to handle only a fraction of use cases. It also requires the devs to learn about the inner workings of the `Request` classes before using them. It can also promote bad patterns as workarounds.

A plain `send()` after setting all the params should be sufficient. Where necessary for the Platform project, we can wrap them in the old `apply_filters` or `do_action`.

Edit: we could still have a generic filter inside the `Request` class, with a constant name per request type and having passed in the request data as the context.

cc @Automattic/gamma @jrodger

zmaglica:
Fixes #6980 

 This PR address a problem of adding filter hook name as a argument to the send function. The majority of the request classes has their own hook name predefined in request class so we don't have to manually set each time. Also new methods has been added where you pass hook name and arguments and also hook aguments only if filter is already defined.

#### Testing instructions

* Make sure that all tests pass

-------------------

**Post merge**


includes/core/server/class-request.php
```diff
@@ -21,7 +21,7 @@
 /**
  * Base for requests to the WCPay server.
  */
-abstract class Request {
+abstract class Request{
```
RadoslavGeorgiev: I'm not sure how this change got added, let's fix the whitespace.

includes/core/server/class-request.php
```diff
@@ -369,6 +379,26 @@ protected function set_request_route_id_parameter( $id ) {
 			}
 		}
 	}
+	/**
+	 * Assign the WordPress filter hook and the arguments specific to the previously assigned hook.
+	 *
+	 * @param string $hook WordPress filter hook name.
+	 * @param mixed  ...$args Arguments for the filter hook.
+	 */
+	public function apply_hook( string $hook, ...$args ) {
```
RadoslavGeorgiev: Maybe change this to something like `assign_hook`? `apply_hook` sounds like something that would be executed immediately, and that's not the case here.

Also, the description arguments says "filter hook", which is wrong. I think "filter/hook" or just "hook" would be better.

RadoslavGeorgiev: Furthermore, considering that `set_hook_args` is a separate method, let's not try to combine things within this one, and let it focus on the hook name. Otherwise we might have a silent side-effect of `apply_hook( 'some_hook' )` cleaning up args without a warning.

includes/core/server/class-request.php
```diff
@@ -379,7 +409,7 @@ protected function set_request_route_id_parameter( $id ) {
 	 * @param string $key   The name of the parameter.
 	 * @param mixed  $value And the value to set.
 	 */
-	final protected function set_param( string $key, $value ) {
+	final public function set_param( string $key, $value ) {
```
RadoslavGeorgiev: This looks wrong. Making this method public makes all other protections effectively pointless.

includes/core/server/class-request.php
```diff
@@ -408,7 +438,7 @@ final protected function unset_param( string $key ) {
 	 *
 	 * @param array $params The new parameters to use.
 	 */
-	private function set_params( $params ) {
+	public function set_params( $params ) {
```
RadoslavGeorgiev: Same commend about making this one public.

includes/core/server/request/class-refund-charge.php
```diff
@@ -25,6 +25,13 @@ class Refund_Charge extends Request {
 
 	const REQUIRED_PARAMS = [ 'charge' ];
 
+	/**
+	 * Used to set WordPress filter hook that will be executed when send() function is called.
```
RadoslavGeorgiev: I see that "filter hook" is used thoughout all request classes. Let's mass-replace with. Also, there is a "the" missing before "WordPress".

includes/core/server/request/class-refund-charge.php
```diff
@@ -25,6 +25,13 @@ class Refund_Charge extends Request {
...
+	 *
+	 * @var string $hook
```
RadoslavGeorgiev: Also, no need to mention `$hook` here. Just `@var string`

includes/core/server/request/class-get-setup-intention.php
```diff
@@ -16,6 +16,13 @@
  * Request class for getting setup intents.
  */
 class Get_Setup_Intention extends Request {
+
+	/**
+	 * Used to set WordPress filter hook that will be executed when send() function is called.
+	 *
+	 * @var string $hook
+	 */
+	protected $hook = 'wcpay_get_setup_intent_request';
```
RadoslavGeorgiev, line 25: There is an empty line before this declaration, which should be after it.

includes/core/server/request/class-list-deposits.php
```diff
@@ -17,6 +17,12 @@ class List_Deposits extends Paginated {
 
 	use Date_Parameters;
 
+	/**
+	 * Used to set WordPress filter hook that will be executed when send() function is called.
+	 *
+	 * @var string $hook
+	 */
+	protected $hook = 'wcpay_list_deposits_request';
```
RadoslavGeorgiev, line 25: Missing empty line.

includes/core/server/class-request.php
zmaglica: Strange. I thought that I removed this line. I will undo this.

tests/WCPAY_UnitTestCase.php
```diff
@@ -48,16 +48,19 @@ protected function mock_wcpay_request( string $request_class, int $total_api_cal
...
 			->setConstructorArgs( [ $api_client_mock, $http_mock, $request_class_constructor_id ] )
+			->onlyMethods( $methods_to_mock )  // Mock all methods except set_hook_args and apply_hook to accommodate filter args when apply_filters is called.
 			->getMock();
 
 		$api_client_mock->expects( $this->exactly( $total_api_calls ) )
 			->method( 'send_request' )
 			->with(
 				$this->callback(
-					// With filters there is a chance that mock will be changed. With this code we are sure that it belongs to same class.
+							// With filters there is a chance that mock will be changed. With this code we are sure that it belongs to same class.
```
RadoslavGeorgiev: Something happened with tabs here. Great approach otherwise!

RadoslavGeorgiev (changes requested):
Great work with a lot of changes! I have a couple of nitpicks regarding naming and spacing, otherwise it looks good.

One note is that there are a few calls to `$request->apply_hook( 'filter_name' )` without any parameters. Is this because a non-default was used?

zmaglica:
> One note is that there are a few calls to `$request->apply_hook( 'filter_name' )` without any parameters. Is this because a non-default was used?

The main purpose is to set filters for the Get_Requests where you don't predefine filter name inside the request class.

includes/core/server/class-request.php
zmaglica: Fixed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/b72ea986fdd961f8586f2173aa7c8bf0087cbeb8

includes/core/server/class-request.php
zmaglica: Good point. I refactored the method signature by making only filter or hook name as an argument and changed the name of the function to assign_hook.

Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/b72ea986fdd961f8586f2173aa7c8bf0087cbeb8

includes/core/server/class-request.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/6a78ee9776f280116ebe4519e649cb2d692f0eeb

includes/core/server/class-request.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/6a78ee9776f280116ebe4519e649cb2d692f0eeb

tests/WCPAY_UnitTestCase.php
zmaglica: Thank you! On another note, the issue with tabbing here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/e39ac1fe049de9ff790af034762fead702a8fbe9

includes/core/server/request/class-refund-charge.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/e39ac1fe049de9ff790af034762fead702a8fbe9

includes/core/server/request/class-refund-charge.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7314/commits/e39ac1fe049de9ff790af034762fead702a8fbe9

includes/core/server/request/class-get-setup-intention.php
zmaglica: Addressed [here](https://github.com/Automattic/woocommerce-payments/commit/e39ac1fe049de9ff790af034762fead702a8fbe9#diff-b9bbe01f6a7ecad3690036ffe391eebd60297fe950c118f8d2916dd4f3c1f345R26).

includes/core/server/request/class-list-deposits.php
zmaglica: Addressed [here](https://github.com/Automattic/woocommerce-payments/commit/e39ac1fe049de9ff790af034762fead702a8fbe9#diff-990c7853485850aec856db1c1d59df288805cfc303d1c25e584ece5348f8471bR26).

zmaglica:
@RadoslavGeorgiev Addressed all comments.

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7314` or branch name `fix/6980-remove-passing-filter-name-from-request-class-send-method` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/6980-remove-passing-filter-name-from-request-class-send-method) :rocket:

---

Build info: 

- Latest commit: 09ba5a2ed7c3c50e00db1d32182eb8b63b6dce09
- Build time: 2023-09-29 15:37:35 UTC

RadoslavGeorgiev (approved):
LGTM :shipit:

PR #7315, opened 2023-09-28
Restrict PMs to location in deferred UPE implementation

### LINKED ISSUE #7313: Hide PM from deferred UPE if country isn't supported
### Describe the bug
In #6604 we introduced logic to conditionally provide a list of countries where the PM is available to account for differences in BNPL PMs. 

That PR handles legacy and Split UPE checkouts but since then we've made deferred UPE the default and we don't have this logic implemented in deferred checkout JS files. 

### To Reproduce

1. Enable any BNPL payment method, for example, Affirm
2. Go to checkout and enter US address
3. Observe that Affirm is available in checkout, which is expected since US is supported country
4. Update the billing country to any EU country, for example Germany
5. Observe that Affirm is still available in checkout, which is **not** expected 
6. Try to checkout. Observe that we get an error because Germany is not supported location and payments will fail.

### Actual behavior
PMs are displayed in checkout for unsupported countries leading to inevitable checkout error.

### Expected behavior
PMs with `$this->currencies` set in their PHP definition class should be restricted to be displayed in checkout only for those countries

### Additional context
See linked PR for implementation in other UPE flavors.

KarlisJ:
Fixes #7313 

In https://github.com/Automattic/woocommerce-payments/pull/6604 we introduced logic to conditionally provide a list of countries where the PM is available to account for differences in BNPL PMs.

That PR adds the same logic to deferred UPE, which is now what's being used in the checkout since 6.4.0 version came out. 

The PR addresses only classic checkout as block checkout still uses the same files that Split UPE used and has this functionality

#### Testing instructions

- Make sure deferred UPE is enabled
- Enable Affirm Payment method
- Add more than 50USD in cart
- Go to checkout and enter US country in billing address -> the Affirm option should be available
- Enter any country outside US or CAN -> the Affirm option should not be available

*

-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
+++ b/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
@@ -115,6 +115,11 @@ jQuery( function ( $ ) {
 		}
 	}
 
+	/**
+	 * Hides payment method if it has set specific countries in the PHP class.
+	 *
+	 * @param {Object} upeElement The selector of the DOM element of particular payment method to mount the UPE element to.
+	 **/
 	function restrictPaymentMethodToLocation( upeElement ) {
 		const paymentMethodsConfig = getUPEConfig( 'paymentMethodsConfig' );
 		const paymentMethodType = upeElement.dataset.paymentMethodType;
@@ -135,6 +140,11 @@ jQuery( function ( $ ) {
 			} );
 		}
 	}
+
+	/**
+	 * @param {string} paymentMethodType The payment method type.
+	 * @param {Array}  supportedCountries A list of countries that given payment method is supported in.
+	 **/
 	function togglePaymentMethodForCountry(
 		paymentMethodType,
 		supportedCountries
--- a/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
+++ b/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
@@ -8,7 +8,9 @@ import {
 	generateCheckoutEventNames,
 	getSelectedUPEGatewayPaymentMethod,
 	isLinkEnabled,
+	isPaymentMethodRestrictedToLocation,
 	isUsingSavedPaymentMethod,
+	togglePaymentMethodForCountry,
 } from '../../utils/upe';
 import {
 	processPayment,
@@ -115,49 +117,12 @@ jQuery( function ( $ ) {
 		}
 	}
 
-	/**
-	 * Hides payment method if it has set specific countries in the PHP class.
-	 *
-	 * @param {Object} upeElement The selector of the DOM element of particular payment method to mount the UPE element to.
-	 **/
 	function restrictPaymentMethodToLocation( upeElement ) {
-		const paymentMethodsConfig = getUPEConfig( 'paymentMethodsConfig' );
-		const paymentMethodType = upeElement.dataset.paymentMethodType;
-		const isRestrictedInAnyCountry = !! paymentMethodsConfig[
-			paymentMethodType
-		].countries.length;
-
-		if ( isRestrictedInAnyCountry ) {
-			togglePaymentMethodForCountry(
-				paymentMethodType,
-				paymentMethodsConfig[ paymentMethodType ].countries
-			);
+		if ( isPaymentMethodRestrictedToLocation( upeElement ) ) {
+			togglePaymentMethodForCountry( upeElement );
 			$( '#billing_country' ).on( 'change', function () {
-				togglePaymentMethodForCountry(
-					paymentMethodType,
-					paymentMethodsConfig[ paymentMethodType ].countries
-				);
+				togglePaymentMethodForCountry( upeElement );
 			} );
-		}
-	}
-
-	/**
-	 * @param {string} paymentMethodType The payment method type.
-	 * @param {Array}  supportedCountries A list of countries that given payment method is supported in.
-	 **/
-	function togglePaymentMethodForCountry(
-		paymentMethodType,
-		supportedCountries
-	) {
-		const billingCountry = document.querySelector( '#billing_country' )
-			.value;
-		const upeContainer = document.querySelector(
-			'.payment_method_woocommerce_payments_' + paymentMethodType
-		);
-		if ( supportedCountries.includes( billingCountry ) ) {
-			upeContainer.style.display = 'block';
-		} else {
-			upeContainer.style.display = 'none';
 		}
 	}
 } );
--- a/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
+++ b/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
@@ -109,8 +109,45 @@ jQuery( function ( $ ) {
 		) {
 			for ( const upeElement of $( '.wcpay-upe-element' ).toArray() ) {
 				await mountStripePaymentElement( api, upeElement );
+				restrictPaymentMethodToLocation( upeElement );
 			}
 			maybeEnableStripeLink( api );
+		}
+	}
+
+	function restrictPaymentMethodToLocation( upeElement ) {
+		const paymentMethodsConfig = getUPEConfig( 'paymentMethodsConfig' );
+		const paymentMethodType = upeElement.dataset.paymentMethodType;
+		const isRestrictedInAnyCountry = !! paymentMethodsConfig[
+			paymentMethodType
+		].countries.length;
+
+		if ( isRestrictedInAnyCountry ) {
+			togglePaymentMethodForCountry(
+				paymentMethodType,
+				paymentMethodsConfig[ paymentMethodType ].countries
+			);
+			$( '#billing_country' ).on( 'change', function () {
+				togglePaymentMethodForCountry(
+					paymentMethodType,
+					paymentMethodsConfig[ paymentMethodType ].countries
+				);
+			} );
+		}
+	}
+	function togglePaymentMethodForCountry(
+		paymentMethodType,
+		supportedCountries
+	) {
+		const billingCountry = document.querySelector( '#billing_country' )
+			.value;
+		const upeContainer = document.querySelector(
+			'.payment_method_woocommerce_payments_' + paymentMethodType
+		);
+		if ( supportedCountries.includes( billingCountry ) ) {
+			upeContainer.style.display = 'block';
+		} else {
+			upeContainer.style.display = 'none';
 		}
 	}
 } );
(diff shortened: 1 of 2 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7315` or branch name `fix/7313-restrict-pm-to-location` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7313-restrict-pm-to-location) :rocket:

---

Build info: 

- Latest commit: 75695f88476a5bb71c0f14a6c250ff3c48193c19
- Build time: 2023-10-04 13:25:45 UTC

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -109,8 +109,45 @@ jQuery( function ( $ ) {
...
+			);
+			$( '#billing_country' ).on( 'change', function () {
+				togglePaymentMethodForCountry(
+					paymentMethodType,
+					paymentMethodsConfig[ paymentMethodType ].countries
+				);
+			} );
+		}
+	}
+	function togglePaymentMethodForCountry(
```
kalessil: Can you please add JS-doc so types get documented (both functions)?

KarlisJ: Added in https://github.com/Automattic/woocommerce-payments/pull/7315/commits/6df5adeb792e62a3a2a7fbd66e85fe95ef45c0a0

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -109,8 +109,55 @@ jQuery( function ( $ ) {
...
+		const isRestrictedInAnyCountry = !! paymentMethodsConfig[
+			paymentMethodType
+		].countries.length;
+
+		if ( isRestrictedInAnyCountry ) {
+			togglePaymentMethodForCountry(
+				paymentMethodType,
+				paymentMethodsConfig[ paymentMethodType ].countries
+			);
+			$( '#billing_country' ).on( 'change', function () {
```
mgascam, line 123: If I understand correctly there's a chance that we add this event listener more than once. I'd suggest to attach the event listener only once, maybe using a flag that can be flipped when the first event listener is attached.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -109,8 +109,55 @@ jQuery( function ( $ ) {
...
+
+	/**
+	 * @param {string} paymentMethodType The payment method type.
+	 * @param {Array}  supportedCountries A list of countries that given payment method is supported in.
+	 **/
+	function togglePaymentMethodForCountry(
+		paymentMethodType,
+		supportedCountries
+	) {
+		const billingCountry = document.querySelector( '#billing_country' )
```
mgascam: nitpick: In the file, we are using jQuery and vanilla JS to query the DOM elements, I'd stick with one

mgascam (commented):
I followed testing instructions and works as expected with block and shortcode checkouts. Leaving a comment about attaching event handler for change event that I think we need to address.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
KarlisJ: Yes, there is that chance. When we initially implemented this in Split UPE files, I specifically asked whether we should do this optimization or stick with a more straightforward solution, and the consensus was to attach the listener multiple times since the performance gain is so tiny compared to the effect on the codebase.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
```diff
@@ -109,8 +109,55 @@ jQuery( function ( $ ) {
 		) {
 			for ( const upeElement of $( '.wcpay-upe-element' ).toArray() ) {
 				await mountStripePaymentElement( api, upeElement );
+				restrictPaymentMethodToLocation( upeElement );
```
timur27, line 114: Initially, I thought the code was invoked on the _[Add payment method](https://github.com/Automattic/woocommerce-payments/pull/7315/files#diff-cbe82d3f6b904df134f8e7eae4284f6713c802cf837a28b74440bcde0712836fR69)_ page unnecessarily since Affirm isn't a reusable payment method there. However, Affirm can be used for [paying orders](https://github.com/Automattic/woocommerce-payments/pull/7315/files#diff-cbe82d3f6b904df134f8e7eae4284f6713c802cf837a28b74440bcde0712836fR70) too, making it challenging to improve without splitting the event handlers for these pages. I'm open to insights or ideas, but leaving it as is is also fine.

timur27 (commented):
Works as expected and BNPL method is filtered out on checkout pages well, thanks @KarlisJ! I'm unsure how BNPL methods should be handled on the _Pay for order_ page, as currently they're being shown always, but at the same time, there is no space to change the country, other than in the credit card payment element. However, I'd like to confirm if the below use-case is something we expect: 

* Prepare a `>= $50` cart
* On the checkout page, choose a billing country other than the US or Canada for the BNPL to be filtered out
* Pay with a 3DS card and decline it on the authorization step
* Navigate to _My account -> Orders -> Pay for order_ and BNPL method is visible there

In the above scenario, we created an order with a billing country other than the supported one, and when trying to pay for it in a non-checkout environment, we're seeing the BNPL method (I tested with Affirm). You definitely know more about BNPLs and I might be missing something. 

Another thing is the suggestion which could be considered, given that deferred UPE code is organized slightly different than split UPE/legacy UPE.

When we first implemented deferred intent creation UPE, we decoupled jQuery from vanilla JS and created two separate scripts (which are now `event-handlers.js` and `payment-processing.js`). The main reason is we'd like to avoid using jQuery wherever possible (but unfortunately, we can't fully avoid using it due to WC Core events being fired using jQuery) to be able to unit test easily, and it was also great to see how scripts are less heavy compared to the legacy UPE/split UPE with.

`restrictPaymentMethodToLocation` and `togglePaymentMethodForCountry` both seem to be the functions that can be moved to the script other than `event-handlers.js` and be imported here instead. Currently, `payment-processing.js` is the only another place, and from its name, it doesn't seem like the best candidate, but maybe we could re-name `payment-processing.js` to reflect all the logic it handles. We could also create a new script responsible for payment method filtering, or use `client/checkout/utils/upe.js` to put the filtering logic of the HTML there.

We could then migrate 

```js 
$( '#billing_country' ).on( 'change', (){} );
```

event handler inside `restrictPaymentMethodToLocation` to be written in vanilla JS and then cover both functions with a couple of unit tests as we currently do in e.g., `payment-processing.test.js`. Let me know what do you think of this idea! Thanks.

KarlisJ:
@timur27 Thank you for your review and pointers!

> In the above scenario, we created an order with a billing country other than the supported one, and when trying to pay for it in a non-checkout environment, we're seeing the BNPL method (I tested with Affirm). You definitely know more about BNPLs and I might be missing something.

Firstly, it's good that you brought up this scenario - we had overlooked it previously. 

The problem in general is not about "seeing" PM, it's about making a payment. If Affirm is available in checkout and the customer tries to make a payment from a country outside the supported US/CAN country, the payment will always fail, thus we hide it from checkout. 

So, naturally, my first though was that you've found an issue, but upon testing, I'm actually able to make an Affirm payment from "Pay for order" page. Which is not what I expected. 

Digging into the differences between `checkout` and `pay for order` in the checkout we send billing data to the `
https://api.stripe.com/v1/payment_methods` but in `pay for order` all billing fields are absent. That seems to be enough for Stripe to let the payment pass (at least in dev mode).

I then tried removing billing fields from `checkout` but I'm getting 
> You specified "never" for fields.billing_details.name when creating the payment Element but did not pass params.payment_method_data.billing_details.name when calling stripe.createPaymentMethod(). If you opt out of collecting data via the payment Element using the fields option, the data must be passed in when calling stripe.createPaymentMethod().

I lack the expertise to understand if this is an effect or defect, and how to proceed. What worries me is the inconsistencies and whether they are coming from our implementations or Stripe's side, and whether one day things might brake as Stripe changes some validation rules.

KarlisJ:
@timur27 
Given our need for this PR to land in code freeze, we'll leave the "pay for order" flow untouched since it's accepting the payments no matter the order country. We'll investigate the inners afterward. 

I'll put here for visibility that we have a Slack discussion that we can't use vanilla JS for the `change` event on the billing country because of [an issue with the select2 component](https://github.com/select2/select2/issues/5971). 

I've refactored the methods as per your suggestions. TBH, I'm not entirely happy with the end result, but given our constraints I don't see a better alternative.

mgascam (approved):
LGTM! 
I suggest addressing the `pay for order` flow, which currently doesn't send the billing address, in a separate pull request.
I'm approving this as it successfully resolves the issue with Affirm payments failing during checkout due to unsupported shopper locations.

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
mgascam: Got it thanks, marking this as resolved

client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
mgascam: Marking as resolved as per the issue with using vanilla JS mentioned in https://github.com/Automattic/woocommerce-payments/pull/7315#issuecomment-1746879451

timur27 (approved):
Thanks for improvements here, @KarlisJ 👍 I appreciate you following up the pay-for-order flow in a separate issue. By the way, since `client/checkout/utils/upe.js` now has two new functions that can be unit-tested, I'd test them. Leaving it up to you to decide if and when, approving in the meantime not to block this enhancement.

KarlisJ:
@timur27 Thank you for your guidance here! JS part of checkout flows is still something causing headaches in team Pulsar so it's great that you spent the time here educating us.

> I appreciate you following up the pay-for-order flow in a separate issue. 

The question here is, what should be the primary goal of such issue? Is missing billing information expected in "pay for order"? I think we should first understand whether there's a problem at all before trying to address something. Cause from user point of view the `pay for order` works with Affirm. 

> new functions that can be unit-tested, I'd test them. Leaving it up to you to decide if and when

We have a cooldown week coming, great time for such activity

PR #7316, opened 2023-09-28
RPP: Container Exceptions

RadoslavGeorgiev:

**Container Exceptions**

This PR changes the `WCPay\Container::get()` method by catching all `Throwable`s, thrown by the League container, and generating a `ContainerException` instead.

This change is required because the currently used PSR-11 package defines interfaces for exceptions, but those interfaces do not implement the `Throwable` interface. On the other hand, the League container does extend the `Exception` class, so in practice, all exceptions will implement the `Throwable` interface.

Without this change, Psalm will either detect uncaught exceptions from the container, or we wouldn't be catching all possible League exceptions.

Here's PSR-11's code, to save you some digging:

PSR-11's `ContainerInterface` (simplified)

```php
namespace Psr\Container;

interface ContainerInterface {
    /**
     * @throws NotFoundExceptionInterface  No entry was found for **this** identifier.
     */
    public function get(string $id);
}
```

PSR-11's `NotFoundExceptionInterface`

```php
namespace Psr\Container;
interface NotFoundExceptionInterface extends ContainerExceptionInterface {}
```

PSR-11's `ContainerExceptionInterface`

```php
namespace Psr\Container;
interface ContainerExceptionInterface {}
```

☝️ this is where the `Throwable` interface should have been defined.

**`createMock()` return type**

I used this PR to sneak something in: A better PHPDoc comment of the `createMock` method. The new doc describes how the return type is a union of `PHPUnit\Framework\MockObject\MockObject` and the requested class. This way tests would not need to always define `@var MockObject|MyClassName` whenever `$this->createMock()` is called for local variables, rather than class properties.

## Testing instructions

Make sure that nothing breaks (all tests here are green). Shortly, there will be follow-up PRs, which utilize both of those, and will be linked here.

-------------------


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7316` or branch name `rpp/container-exceptions` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/container-exceptions) :rocket:

---

Build info: 

- Latest commit: 538564acaed2baa2f27a5173ecd79d5b985c87f7
- Build time: 2023-09-28 20:00:19 UTC

src/Container.php
```diff
@@ -67,12 +68,21 @@ public function __construct(
...
+	 * PSR-11 containers declares to throw an un-throwable interface
+	 * (it does not extend Throwable), and Psalm does not accept it.
+	 * @psalm-suppress MissingThrowsDocblock
 	 */
 	public function get( $id ) {
-		return $this->container->get( $id );
+		try {
+			return $this->container->get( $id );
+		} catch ( \Throwable $e ) {
+			throw new ContainerException( $e->getMessage() );
```
marcinbot: We can preserve the previous exception by passing it to the constructor like this:
```
new ContainerException( $e->getMessage(), 0, $e );
```
See [here](https://www.php.net/manual/en/exception.construct.php). Could be beneficial if we ever need to debug things or preserve stack traces.

src/Container.php
```diff
@@ -8,6 +8,7 @@
 namespace WCPay;
 
 use Psr\Container\ContainerInterface;
+use WCPay\Internal\DependencyManagement\ContainerException;
```
marcinbot: Why do we need this definition implemented in our code in `src` if the League container also defines a `ContainerException`? And the League's exception extends a `RuntimeException` which might actually be more valid, but that's me being nitpicky again.

src/Container.php
marcinbot: So this is something that doesn't quite grok for me. You mention that:
1. The PSR-11 class doesn't throw objects implementing `Throwable`
2. The League container, on the other hand, does

But then catching a `Throwable` here won't help with scenario 1. It will still be a problem.

And in case of scenario 2, we already have exceptions from League that should be fine to use. And we seem to be ultimately using a League container (via `ExtendedContainer`) so there should be no issue.

So is this just to keep Psalm happy?

I guess I'm still not sure when and why we're using PSR rather than League 🤔

marcinbot (commented):
The code is quite straightforward and looks good to me. Asked a few questions so I understand the problem a bit better.

src/Container.php
RadoslavGeorgiev: > I guess I'm still not sure when and why we're using PSR rather than League 🤔

League uses PSR, and their code works with both `psr/container` 1.* and 2.*. However, one of our libraries is stuck with 1.1 (I think it's Psalm), and we're using 1.1, which is the one that doesn't implement `Throwable`. [2.0 does](https://github.com/php-fig/container/blob/master/src/ContainerExceptionInterface.php).

> 1. The PSR-11 class doesn't throw objects implementing Throwable ... But then catching a Throwable here won't help with scenario 1. It will still be a problem.

It's the PSR-11 interface that doesn't implement `Throwable`. League's exceptions, as you noticed, are based on `RuntimeException`, which implements `Throwable`.

> So is this just to keep Psalm happy?

Yes.

src/Container.php
RadoslavGeorgiev: Looking at https://github.com/Automattic/woocommerce-payments/pull/6862/commits/e153b8081c413db5ba676daf7fa34168a124bcad where the exception got introduced, the answer is that then Psalm exception checking was not introduced, and I didn't even realize that the League package provided exception classes.

FWIW, League has two exceptions: 

- `ContainerException extends RuntimeException implements ContainerExceptionInterface` that you are referencing.
- `NotFoundException extends InvalidArgumentException implements NotFoundExceptionInterface` which is essentially the same as our `ContainerException`.

This PR treats both the same, and converts them into an `InvalidArgumentException`, which might not be ideal, but at the same time, it might be easier to have the one `ContainerException` to worry about.

src/Container.php
RadoslavGeorgiev: TIL. Done in https://github.com/Automattic/woocommerce-payments/pull/7316/commits/85a9d4c32f21a92487b25f65328cb98db4eb42c9

marcinbot: Thanks for explaining!

src/Container.php
marcinbot: What do you think about just using the League exception, maybe `ContainerException` and getting rid of our implementation? It would reduce the extraneous code, and ultimately comes down to working the same way.

RadoslavGeorgiev: Done in: https://github.com/Automattic/woocommerce-payments/pull/7316/commits/5a27500a11380d8d993459e2af2f0e043393d847

marcinbot (approved):
Looks good, :shipit:

Issue #6689, opened 2023-07-05
Payment Process: Create empty states


RadoslavGeorgiev:
### Description

Lay down the initial set of states, with empty methods, except for transitions.

### Acceptance criteria

All states from the state diagram (except for UPE-specific ones) are laid out.

States should be as simple as this:

```php
// Add hollow states like this:
class Prepared_State extends State {
    public function verify() {
        // Nothing but this at first.
        $this->payment->switch_state( new Verified_State( $this->payment ) );
    }
}
 
```

### Testing instructions

We should be able to go through the full cycle without errors.

```php
$payment = new Payment();
$payment->prepare();
$payment->verify();
$payment->process();
$payment->complete();
return $payment->get_response();
```

### Dev notes

Previously implemented in https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-f3e93d7abed593f8ffd32081592094bd60a0680cbb0c0b10c5ed533e6b7bd4cb.

### Additional context

Epic: https://github.com/Automattic/woocommerce-payments/issues/6673
Project: paJDYF-9hL-p2

PR #7318, opened 2023-09-28
RPP: Payments skeleton

### LINKED ISSUE #6686: Payment Process: Payment object and states skeleton

RadoslavGeorgiev:
Fixes #6686

This PR introduces the skeleton of the state mechanism for payments, as well as all basics, required to allow transitions between states.

- The `WCPay\Internal\Payment\PaymentContext` class, which is hollow for now. It will be used to contain the working payment data.
- A set of classes for states:
	- The abstract `WCPay\Internal\Payment\State\PaymentState` class as a base for all states.
		- This class dependens on the `StateFactory)` (see below), as states will be able to generate new states.
		- Receives and contains the `PaymentContext` object. This is not done through the constructor, as states will be managed by DI.
		- Has the base definition of the `process` method, which throws an exception wen called. This can be changed in concrete states.
	- The `InitialState` class, which has a basic `process()` implementation that simply transitions to `CompletedState`.
	- Hollow `CompletedState` class, which for now does nothing more than serve as means of communication with the gateway.
	- A `StateFactory` class that should be the only place where states get instantiated. It has access to (and dependency of) the container, but allows nothing except states to be instantiated. To ensure that, it also requires the `PaymentContext` object.
- The gateway uses the returned/current payment state to determine what to return, namely a `[ 'success', 'redirect' ]` array upon the `CompletedState`.

## Testing instructions

Everything added here has 100% test coverage.

States are not really functional at this point, which leaves little to test. However, submitting the checkout form with the new payment process enabled, should redirect to the Order Received page, even though the order would not be updated.

-------------------


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7318` or branch name `rpp/6686-payments-skeleton` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/6686-payments-skeleton) :rocket:

---

Build info: 

- Latest commit: 1697789fdf497a4ce752a0abf9893252f7d72c1c
- Build time: 2023-10-05 09:01:01 UTC

src/Internal/Payment/State/StateFactory.php
```diff
@@ -0,0 +1,79 @@
...
+					__( 'The class %1$s either does not exist, or is not a subclass of %2$s', 'woocommerce-payments' ),
+					$state_class,
+					PaymentState::class
+				)
+			);
+		}
+
+		$state->set_context( $context );
+
+		return $state;
```
marcinbot: I think this function readability could benefit from early returns/throws:

```
if ( ! is_subclass_of( $state_class, PaymentState::class ) ) {
    // throw an exception with a message informing of this
}

try {
	// Load the state from the container.
	$state = $this->container->get( $state_class );
} catch ( ContainerException $e ) {
	// wrap the exception in something more appropriate and rethrow
    // or maybe even get rid of try/catch altogether since it shouldn't happen often enough and especially not in production
}

$state->set_context( $context );
return $state;
```

No need for nulls and all the nesting :)

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	/**
+	 * Throws an exception, indicating that a given method is not available.
+	 *
+	 * @param string $method_name The name of the called method.
+	 * @throws MethodUnavailableException
+	 */
+	private function throw_unavailable_method_exception( string $method_name ) {
+		throw new MethodUnavailableException(
```
marcinbot: I think there's some mixup with the exception names. The methods on states are transitions, so I feel like `StateTransitionException` would be appropriate here. IMO we could get rid of `MethodUnavailableException`.

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,116 @@
...
+			sprintf(
+				// translators: %s is the name of a method of the payment object.
+				__( 'The %s method is not available in the current payment state.', 'woocommerce-payments' ),
```
marcinbot: Would be good to mention the current class name instead of `the current payment state`. AFAIR this could be achieved with `static::class` which should evaluate to the topmost class name.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -809,7 +810,14 @@ function_exists( 'wcs_order_contains_subscription' )
...
 		$service = wcpay_get_container()->get( PaymentProcessingService::class );
-		return $service->process_payment( $order->get_id() );
+		$state   = $service->process_payment( $order->get_id() );
+
+		if ( $state instanceof CompletedState ) {
+			return [
+				'result'   => 'success',
+				'redirect' => $this->get_return_url( $order ),
+			];
+		}
```
marcinbot, line 822: We'll need to return or throw something at the end of this function in case the state is not a completed state or any other expected state.

marcinbot (commented):
This is looking really good! Great approach to iteration. The only comments I have are pretty minor.

src/Internal/Payment/State/StateFactory.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7318/commits/56933778b2c3bf6cfb97fad2e4e5204e4a74fd07. The nesting was there to avoid multiple exceptions within such a simple method.

FWIW, I'm using the same exception class (`StateTransitionException` now) even for container issues. Ideally, this could bubble up the container exception, but as you said, it should not happen often. Moreover, we'd need to add an additional `@throws` line to each transition method of the state class, and we've already got long doc blocks there.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7318/commits/497b3aedf2179f5ece8ea5694dd611b61fbb68ee. A simple `get_class( $this )` should suffice, as we care about the _current_ state.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Yes. Let's discuss that when we have more states, especially erroneous ones. That should be done in https://github.com/Automattic/woocommerce-payments/issues/6689. The issue's definition still has to be updated, but that's when we'll add and discuss erroneous states.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: > The methods on states are transitions

I'm not ready to close that door yet. Payments should be able to communicate certain pieces of information only from specific states. An example that comes to mind is redirection URLs (or details) whenever a payment requires either 3DS authentication, or a redirect to an external site, but I'm sure we'll have further examples.

Overall, I'm not certain that "transition" is an appropriate name even for the methods, which do trigger transitions. Given the naming in this PR, a transition would occur whenever `create_state` is called. Keeping the definition of a state machine in mind, those methods are either events or inputs, but not transitions.

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	private $state_factory;
+
+	/**
+	 * Class constructor, only meant for storing dependencies.
+	 *
+	 * @param StateFactory $state_factory Factory for payment states.
+	 */
+	public function __construct( StateFactory $state_factory ) {
+		$this->state_factory = $state_factory;
```
zmaglica: Can we store state factory in payment context? WDYT?

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,117 @@
...
+	 *
+	 * States control the payment flow, and allow transitions to the next state.
+	 * This method should only be called whenever the process is ready to transition
+	 * to the next state, as each new state will be considered the payment's latest one.
+	 *
+	 * @param string $state_class The class of the state to crate.
+	 * @return PaymentState
+	 * @throws StateTransitionException In case the new state could not be created.
+	 */
+	protected function create_state( string $state_class ) {
```
zmaglica: I suggest refactoring this function to something like `transition_to(string $state_class)` and retaining the `state_factory` within the context. However, feel free to disregard this comment if you prefer an alternative approach.

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,117 @@
...
+		$state = $this->state_factory->create_state( $state_class, $this->context );
```
zmaglica: IMHO, the state should not set context. Context should be a godlike class that shifts states and keeps all data in between. However, feel free to disregard this comment if you prefer a current approach.

src/Internal/Payment/State/StateFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * Creates a new state based on class name.
+	 *
+	 * @param string         $state_class Name of the state class.
+	 * @param PaymentContext $context     Context for the new state.
+	 * @return PaymentState               The generated payment state instance.
+	 * @throws StateTransitionException   When the class name is not a state.
+	 */
+	public function create_state( string $state_class, PaymentContext $context ): PaymentState {
```
zmaglica: I believe we can eliminate this class and transfer its logic to the `PaymentContext` class. By enforcing that state classes implement a specific interface or contract, we can force certain class behaviors if needed.

zmaglica:
I've included some comments, and I want to express my appreciation for your work. However, I'd like to suggest a slightly different approach to the problem. One of the main distinctions in my approach is that I would utilize the context to set the state, as opposed to the current implementation. Additionally, I'd propose adding a transition_to (or a similar) function within the state class. This function would be used to facilitate transitions between states.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: Based on @marcinbot's suggestion (I'm still to post those), the context is a simple object, containing either scalar or at most serializable data. It should not have dependencies.

As an additional reason, the state factory is needed mostly because states (will) have dependencies as well, see the follow-up PRs. Since they have dependencies, and need the state factory anyway, it seems cleaner to store keep the state factory there.

In simpler terms: States control which state is next. The context only contains context. With that in mind, it makes sense to store the state factory in the place where states are determined.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: I'd prefer the name `transition_to` if we were using the state pattern. With the current state machine, it's not 100% clear what a transition is:

- From within the states, transitions happen exactly in this method. We will probably also use it to store the payment and/or context in the database.
- However, externally, a transition method (ex. `process`) returns the next state, and one could argue that the transition happens exactly when the state is returned there. If you call `$state->process()` without storing the resulting reference (the one to the next state), the payment would essentially not have transitioned.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: You are continuing the "state pattern vs. state machine" discussion we were having at the meetup.  At the end, we agreed to try the new approach and avoid having `Payment` as the god-like class, working with states directly.

So far, that seems like it's working, but I'm curious to see whether something in this PR or the follow-up ones seems like a bad idea to you.

src/Internal/Payment/State/StateFactory.php
RadoslavGeorgiev: I think that this is the same discussion as https://github.com/Automattic/woocommerce-payments/pull/7318#discussion_r1342726060, and as your top-level comment (https://github.com/Automattic/woocommerce-payments/pull/7318#issuecomment-1742990517).

src/Internal/Payment/State/StateFactory.php
marcinbot: > FWIW, I'm using the same exception class (StateTransitionException now) even for container issues. Ideally, this could bubble up the container exception, but as you said, it should not happen often. 

I think we should just allow the ContainerException to bubble up. Then, the process entry point would just be able to handle any throwables and return a generic error.

Basically, if we end up in a situation where a ContainerException is thrown, then something went very wrong, an equivalent of an HTTP 500, a critical error caused by a developer. By wrapping it in an extra exception we risk losing that context as devs, and at the same time, the context is not important to the customer at the checkout (hence the generic error at the entry point).

So basically we save a couple of extra lines here, and there are advantages to doing it this way.

src/Internal/Payment/State/PaymentState.php
marcinbot: Ok, I'm not 100% set on this so that's fine. I just think there could be an opportunity to not over-engineer this thing by reducing the number of exceptions, if, ultimately, it comes down to the same thing: we tried to provide an input to trigger a transition, and then something goes wrong. That would fall under a transition exception.

> An example that comes to mind is redirection URLs (or details) whenever a payment requires either 3DS authentication, or a redirect to an external site, but I'm sure we'll have further examples.

I think it makes sense for states to have those, but there's no need to have them declared on the abstract class. The abstract class should just declare the methods that are triggering transitions (if we're precious and don't want to call them transitions). In this case, yeah, what you described would be just "methods", but there would be no need to have an exception.

includes/class-wc-payment-gateway-wcpay.php
marcinbot: I'm just being a bit nitpicky about all code paths returning values (or throwing) - nothing specific about the states or the process.

The `if ( $state instanceof CompletedState )` is basically an early return, and otherwise we don't define a return value. Even a simple `throw new Exception()` would suffice for the time being.

src/Internal/Payment/State/PaymentState.php
marcinbot: To further explain:
- States are like Controllers in a REST API, controlling which logic is triggered at which point, and also controlling the application flow (by declaring the dependencies)
- The PaymentContext is like a Request, an input for the States to determine what should happen at the given moment

src/Internal/Payment/State/PaymentState.php
marcinbot: IMHO having a godlike object is an anti-pattern. That's how we ended up with the current implementation of the Gateway - a mess of a spaghetti code.

src/Internal/Payment/State/StateFactory.php
RadoslavGeorgiev: Letting the container exception bubble up seems like the way to go. Because of the Psalm rule, it will require a `@throws` tag for every transition method, but that should be fine. Will do tomorrow.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: > I think it makes sense for states to have those, but there's no need to have them declared on the abstract class. 

Let's stick to the Liskov substitution principle (SO**L**ID), and not do that. However, to streamline things, I will switch to the `StateTransition` exception for transition methods, and leave the `MethodUnavailableException` reserved for such methods in the future. How does that sound?

For what it's worth, it might also make sense to declare separate interfaces for the base transition methods (ex. `StateMachineState`) and another interface for specific methods.

I'm not sure what's the right balance here.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: > The PaymentContext is like a Request, an input for the States to determine what should happen at the given moment.

But states should be able to also add to the context, right?

marcinbot: Yes, I think this should be possible

src/Internal/Payment/State/StateFactory.php
RadoslavGeorgiev: Done: https://github.com/Automattic/woocommerce-payments/pull/7318/commits/abc81746f8f9bfd7cf57f398af707204242e6c0d

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Done here: https://github.com/Automattic/woocommerce-payments/pull/7318/commits/f67e84b024a2e3a2a9d173da3bd08b35cd114c6b

src/Internal/Payment/State/PaymentState.php
marcinbot: Well, since we're following SOLID, then I think this could fall under interface segregation, as you describer here:

> For what it's worth, it might also make sense to declare separate interfaces for the base transition methods (ex. StateMachineState) and another interface for specific methods.

Either way, it's probably too early to spend too much time discussing that at this point.

> However, to streamline things, I will switch to the StateTransition exception for transition methods, and leave the MethodUnavailableException reserved for such methods in the future. How does that sound?

It's a compromise, thank you. I would however caution against shipping dead code. The exception file is just a couple of lines, why not re-add it when it's needed?

RadoslavGeorgiev: I was clinging onto the concept, I'm ok removing the exception for now. Done in https://github.com/Automattic/woocommerce-payments/pull/7318/commits/3b3955b18c39620d50c257579ad10a607b534fad

src/Internal/Payment/State/PaymentState.php
zmaglica: Ok, it makes sense now. Thank you for clarifying.

src/Internal/Payment/State/PaymentState.php
zmaglica: Ok, that makes sense. Thank you for clarifying it. I think I mixed both concepts in this PR.

marcinbot (approved):
Looks good to me :shipit:

RadoslavGeorgiev:
> Just a note that I think there's still some of the base PR in this. So we should rebase/merge before merging.

Which one is that? The base PR was `rpp/container-exceptions`, and I don't see any left-overs from it here.

src/Internal/Payment/State/PaymentState.php
```diff
@@ -0,0 +1,117 @@
...
+namespace WCPay\Internal\Payment\State;
+
+use WCPay\Vendor\League\Container\Exception\ContainerException;
+use WCPay\Internal\Payment\Exception\StateTransitionException;
+use WCPay\Internal\Payment\PaymentContext;
+
+/**
+ * Base class for payment states.
+ */
+abstract class PaymentState {
```
htdat: Nitpick: Might use naming `PaymentStateAbstract` or `AbstractPaymentState`? Similar to `PaymentMethodInterface` or `AbstractServiceProvider` in the current `src` folder. 

I think it will help aid reading code in the future to quickly understand what type of name we're going to call (concrete class, abstract class, or interface)

Anyway, this convention is also common in Woo core.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -8,19 +8,46 @@
...
+	 * @throws StateTransitionException In case a state cannot be initialized.
+	 * @throws ContainerException       When the dependency container cannot instantiate the state.
 	 */
 	public function process_payment( int $order_id ) {
-		throw new Exception( 'Re-engineering payment process is in-progress. Sit tight, and wait more!' );
+		$context = new PaymentContext();
+		$state   = $this->state_factory->create_state( InitialState::class, $context );
+		$state   = $state->process();
+
+		return $state;
```
htdat, line 51: Now, it's simple and straightforward. But what if this goes through a few internal states, how will this can be handled? 

I think this question is somewhat mentioned here https://github.com/Automattic/woocommerce-payments/pull/7318#discussion_r1342721335 but it's still not clear to me how we're going to do that.

htdat (commented):
Great work on this. My only concern is how we're going to handle "transitions" and "side effects" for each state?

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: Works for me, will do shortly.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: Actually, the beauty here is that if necessary, we can do that from state to state:

```php
class InitialState {
	public function process() {
	}
}
```

This allows us to define some discrete states, which the payment can go through, without those being visible to the outside.

RadoslavGeorgiev:
> Great work on this. My only concern is how we're going to handle "transitions" and "side effects" for each state?

@htdat with this series of PRs, we're moving away from side effects as small additions, happening upon transitions. Everything should be contained within states, which work as controllers.

src/Internal/Payment/State/PaymentState.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7318/commits/052748eb63db150eed9ca42a843388feaa3223b0

src/Internal/Service/PaymentProcessingService.php
htdat: Gotcha! I understand that now. But I hope that we keep using `$next_state->process()`, which has been described in `AbstractPaymentState`, rather than a new method such as `$next_state->complete()`.

htdat (approved):
Thanks for the explanation. LGTM now! 

@RadoslavGeorgiev - I've slightly edited your recent comment https://github.com/Automattic/woocommerce-payments/pull/7318#issuecomment-1746456434 to correct my GitHub username ;)

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: We'll discuss that when we get there, but why not? The next state would not be doing the same thing, and considering that most of those would be called internally, it would make sense to name the method as what it does.

PR #7321, opened 2023-09-28
Update pay-for-order js config billing email

hsingyuc:
This PR updates the pay-for-order js config billing email to the session email

#### Testing instructions
1. Go to WCPay Dev -> Enable the Pay-for-order flow for WooPay
2. Create a pay-for-order order
3. Go to the pay-for-order page
4. Check in the console if your `billing_email` in the `window.wcpayConfig` is the correct one.

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7321` or branch name `update/js-config-session-email` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/js-config-session-email) :rocket:

---

Build info: 

- Latest commit: 4a9b85223e35bb0604f5fe9457cbc0d27200ce85
- Build time: 2023-11-07 14:02:27 UTC

malithsen (commented):
Hi @hsingyuc Thanks for working on this

This approach does not work if the initial order was placed as a guest user. 
These are my steps.

1. As a guest logged out customer, place an order using WooPay
2. As an admin, change the order status in the merchant site to "Pending payment"
3. Copy the "Customer payment page" link
4. As a customer visit the link in 3 and try to go through the WooPay payment flow
5. WooPay fails to initialize because the `billing_email` captured during step 1 is empty.

While I haven't tested, I believe the same would happen if a logged-out user uses Bookings or one of the extensions that use pay for order flow. Please correct me if I am wrong.

Additionally, it feels a bit weird to use session_email as billing_email in the first place. They are not necessarily the same always.

hsingyuc:
@malithsen Thank you for catching that! Fixed here [5e22de1](https://github.com/Automattic/woocommerce-payments/pull/7321/commits/5e22de198ebc8f6ded5ccfa9a93102c8cad4b959)

hsingyuc:
@malithsen Thanks again and this PR is ready for review

includes/class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -116,18 +116,32 @@ public function is_woopay_enabled() {
 	 * Add the Pay for order params to the JS config.
 	 */
 	public function add_pay_for_order_params_to_js_config() {
+		global $wp;
+		$order_id = $wp->query_vars['order-pay'];
```
malithsen: We need to make sure `order-pay` exists. Otherwise it will result in an `undefined index` error.

Steps:
Go to `My Accounts` page
Notice the following error
 
SCR-20231025-lmkt

includes/class-wc-payments-express-checkout-button-display-handler.php
```diff
@@ -116,18 +116,32 @@ public function is_woopay_enabled() {
...
+					$session       = wc()->session;
+					$session_email = '';
+
+					if ( is_a( $session, WC_Session::class ) ) {
+						$customer      = $session->get( 'customer' );
+						$session_email = is_array( $customer ) && isset( $customer['email'] ) ? $customer['email'] : '';
+					}
+
+					$user_email = sanitize_email( wp_unslash( filter_input( INPUT_POST, 'email' ) ) ) ?? $session_email;
+
```
malithsen, line 142: Do we still need to set `session_email` after the latest changes? or can we just set it to an empty string if the user does not have the correct access rights?

malithsen (commented):
@hsingyuc I missed your ping. Apologies for the delayed review here.

I can confirm that billing_email is no longer revealed. Just one more warning to fix and left a question about the use of `session_email` in the comments.

includes/class-wc-payments-express-checkout-button-display-handler.php
hsingyuc: Good catch!  Fixed here [0593b5e](https://github.com/Automattic/woocommerce-payments/pull/7321/commits/0593b5ee805c5eb0906b9651edb559ce163bd37f).

includes/class-wc-payments-express-checkout-button-display-handler.php
hsingyuc: We are not checking the access rights here.  This function only adds the email to the `user_email`.  Let me know if I understood you correctly.

includes/class-wc-payments-express-checkout-button-display-handler.php
malithsen: Thanks!

includes/class-wc-payments-express-checkout-button-display-handler.php
malithsen: Hi Emma, what I don't quite understand is at occasions do we fallback to `user_email`. Could you give me a test where `user_email` is used instead of the `get_billing_email` in lines 146-149?
I've been testing with guest and logged in orders but have not been successful in finding an occasion where `user_email` is required. I am likely missing something and will appreciate your help in testing it out.

malithsen (approved):
Thanks for your patience with the delayed review here @hsingyuc 

I can confirm this fixes the initial issue of billing_email visibility before the pay for order screen. I have a question about user_email in https://github.com/Automattic/woocommerce-payments/pull/7321#discussion_r1374650710 but I am pre-approving as this is functional as-is.

includes/class-wc-payments-express-checkout-button-display-handler.php
hsingyuc: This is used whenever we have guest orders.  To reproduce:

1. Create a guest order
2. Use incognito or log out of WP
3. Add `error_log($user_email);` right after `$user_email`
4. Visit the pay for order page.  You should see the email input form
5. Enter your email
6. See the debug log with the user email

Screen Shot 2023-10-30 at 2 42 02 PM

malithsen: Thanks. But I am more confused now 😅 
The verification in your screenshot is done by WC-core. It will always check against the order's billing email. 

`js_config['billing_email']` we are setting in this PR is only for WooPay's consumption.

Once a user gets pass the verification in the screenshot above, we can be reasonably confident that it's the correct shopper.  So shouldn't we always be setting `$js_config['billing_email']  = $order->get_billing_email()`?

Is there a particular reason why WooPay require `user_email` as the `billing_email` for logged out users?

---
On a separate note, following your test instructions, `$user_email` is an empty string for me when logged out. This is because `sanitize_email( wp_unslash( filter_input( INPUT_POST, 'email' ) ) )` is an empty string and the null coalescent operator is happy to use it without falling back to the value of `$session_email`

hsingyuc: >The verification in your screenshot is done by WC-core. It will always check against the order's billing email.

This is true.  But we don't have any way of knowing if they are past this check or not.

>Once a user gets pass the verification in the screenshot above, we can be reasonably confident that it's the correct shopper. So shouldn't we always be setting $js_config['billing_email']  = $order->get_billing_email()?

We are not sure if we are through this check or not, they are both on the same page and the core function for this check is not exposed.  You can see that `wcpayConfig ` is visible in your console even before a logged-out user enters their email.

The core check also just uses `sanitize_email( wp_unslash( filter_input( INPUT_POST, 'email' ) ) )`.

>Is there a particular reason why WooPay require user_email as the billing_email for logged-out users?

I'm not quite sure I understand what you mean.  We are using the same POST input from the form that core uses to check for logged-out users.

>On a separate note, following your test instructions, $user_email is an empty string for me when logged out.

Is this after you entered your email in the above form and clicked "Verify"?  I'm able to see the email I entered.

hsingyuc: @malithsen Is this line why you think we are past the user email check?  

https://github.com/Automattic/woocommerce-payments/blob/0593b5ee805c5eb0906b9651edb559ce163bd37f/includes/class-wc-payments-express-checkout-button-display-handler.php#L129

If so, this is not related to the core email verification.  `current_user_can( 'pay_for_order', $order_id )` always returns `true` for guest orders even when a user is logged out and not verified.

malithsen: Thanks for walking me through this @hsingyuc this makes sense to me now. I misunderstood how the verification works. This looks good :shipit:

hsingyuc: Thank you for double-checking on this, @malithsen! 👍

PR #7324, opened 2023-09-29
RPP: Working Data

### LINKED ISSUE #6688: Payment Process: Add a mechanism/trait for working data
### Description

Add an entity mechanism/trait for working data (ex. intent ID, metadata, etc.), separating the model from orchestration logic.

Ideally, the payment object should only be concerned about managing the current state of the payment, and proxying methods to the right state. A trait could be useful here, but we should avoid it.

### Acceptance criteria

The trait/class has proper methods for setting and getting objects, with the right types.

### Testing instructions

Nothing specific, but requires full coverage.

### Dev notes

Previously implemented in https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-8999304bd00deaea378fd22f07dcb833039f55805136ad52a5195fa77ade6894.

### Additional context

Project: paJDYF-9hL-p2

RadoslavGeorgiev:
Fixes #6688. Blocked by https://github.com/Automattic/woocommerce-payments/pull/7318

This PR makes the `PaymentContext` class functional, and adds its basic usage throughout the payment process.

- Order Service:
	- Add the `get_order_service()` method to `WC_Payments` to expose the legacy `WC_Payments_Order_Service` instance, which is also added to the `LegacyContainer`.
	- Create a new `WCPay\Internal\Service\OrderService` class. For now, we should try to keep this class from accepting `WC_Order` objects as parameters, leaning on order IDs instead. Initially, it will just have the right method signatures, but in the future we should migrate everything from `WC_Payments_Order_Service` to it. Currently, the service only contains a single method, which should be used as an example to follow throughout the payment process re-engineering.
	- Introduce a new `WCPay\Internal\DependencyManagement\ServiceProvider\GenericServiceProvider`, which will provide all generic services.
- Saved payment method: The `SavedPaymentMethod` class now stores the IDs of both the payment method, and associated token. This contrasts with the `WC_Payment_Token` instance that it used to wrap. That's done to make the object plain and serializable, forcing delegation to a service in the future, instead of using methods directly.
- Payment Request class:
	- The class is modified to provide a token ID instead of an object to the `SavedPaymentMethod` constructor (see the previous point).
	- A new `populate_context` method was introduced, which will transfer all necessary details into the `PaymentContext` class. For now, the only needed detail is the `PaymentMethodInterface` instance, and further details should be added when needed.
- `PaymentContext` class:
	- The class receives a set of private `set` and `get` methods, which will store the data in an array. This is because in the future, those methods will contain the necessary logic for logging.
	- There are a few setters and getters, showcasing how they will be used.
- The `PaymentProcessingService` is changed to accommodate all changes.
- `InitialState` is the link between all data:
	- It uses `get_order_id()` and `get_payment_method()` from the context.
	- It uses the `OrderService` to attach the payment method to the order.

#### Testing instructions

Try checking out with the new payment process enabled, and check the order meta. `_payment_method_id` should contain a payment method ID.

As always, everything is covered with tests. There's just one place where feedback is required, and I've added a comment to it.

-------------------

- [x] Covered with tests (100% coverage of `src`)
- [x] Psalm has no warnings
- [x] Tested on mobile (Does not apply)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7324` or branch name `rpp/6688-working-data` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/6688-working-data) :rocket:

---

Build info: 

- Latest commit: fd9aa371e448e7e1b305d4031efa1a5bd8af298f
- Build time: 2023-10-05 09:44:20 UTC

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -61,4 +100,30 @@ public function test_process_payment_happy_path() {
...
+	 * Test the basic happy path of processing a payment.
+	 */
+	public function test_process_payment_happy_path_without_mock_builder() {
+		$sut = new PaymentProcessingService( $this->mock_state_factory, $this->mock_legacy_proxy );
+
+		$mock_initial_state   = $this->createMock( InitialState::class );
+		$mock_completed_state = $this->createMock( CompletedState::class );
+
+		// Prepare a payment method for the context.
+		$_POST['payment_method']       = WC_Payment_Gateway_WCPay::GATEWAY_ID;
```
RadoslavGeorgiev, line 114: Should we modify `$_POST` directly here? AFAICT, there are a few approaches:

- Just like this method, modify `$_POST` directly.
- Allow the request array to be provided to the `process_payment` method, which would relay it to `PaymentRequest`.
- `test_process_payment_happy_path` from this class, which mocks `create_payment_context` and `create_payment_request`. This method seems the cleanest here, but has two issues: Missing test coverage for those methods, and the fact that they need to be `protected` instead of `private`.

Which option seems best? Is there a better alternative that I'm missing?

src/Internal/Payment/PaymentContext.php
```diff
@@ -7,9 +7,92 @@
...
 /**
  * A context object, which is shared between payment states.
  */
 class PaymentContext {
+	/**
+	 * Contains all the context's data.
+	 *
+	 * @var array
+	 */
+	private $data = [];
```
zmaglica, line 28: Could we consider using well-defined variables in place of the `$data` array? Using well-defined variables offers several advantages, including type-safety, immutability (if necessary), and the ability to provide type-hinting. With `$data`, one can store anything, which may lead to unintended consequences. I believe it's important to be stringent in defining the `PaymentContext`, as it encompasses a critical piece of logic—data. Allowing arbitrary data storage in this context could pose potential challenges in the future.

src/Internal/Payment/PaymentContext.php
```diff
@@ -7,9 +7,92 @@
...
+
+	/**
+	 * Stores an internal value.
+	 * Use this method for changes to allow logging in the future.
+	 *
+	 * @param string $key   Property name.
+	 * @param mixed  $value Value to store.
+	 */
+	private function set( string $key, $value ) {
```
zmaglica, line 31: Nitpick. Do we plan to use PSR12 for function/variable naming as we use PSR4 for class naming (for the sake of autoloader)?

src/Internal/Payment/PaymentContext.php
```diff
@@ -7,9 +7,92 @@
...
+	private function get( string $key ) {
+		return $this->data[ $key ] ?? null;
+	}
+
+	/**
+	 * Stores the order requiring payment.
+	 *
+	 * @param int $order_id ID of the order.
+	 */
+	public function set_order_id( int $order_id ) {
```
zmaglica: Nitpick. Do we plan to use PSR12 for function/variable naming as we use PSR4 for class naming (for the sake of autoloader)?

src/Internal/Payment/PaymentContext.php
```diff
@@ -7,9 +7,92 @@
...
+
+	/**
+	 * Stores an internal value.
+	 * Use this method for changes to allow logging in the future.
+	 *
+	 * @param string $key   Property name.
+	 * @param mixed  $value Value to store.
+	 */
+	private function set( string $key, $value ) {
+		$this->data[ $key ] = $value;
```
zmaglica, line 56: Are we planning to add immutability to the certain properties? For example, you should not be able to change order_id once it is set, but with this approach, you can change it and cause wired cases during the execution of next states.

src/Internal/Payment/PaymentRequest.php
```diff
@@ -144,4 +144,14 @@ public function get_payment_method(): PaymentMethodInterface {
 
 		throw new PaymentRequestException( __( 'No valid payment method was selected.', 'woocommerce-payments' ) );
 	}
+
+	/**
+	 * Populates a payment context before processing a payment.
```
zmaglica, line 149: Maybe we can explain what needs to be populated and why so anyone who sees this function and comments are familiar with the whole flow.

src/Internal/Payment/State/InitialState.php
```diff
@@ -8,18 +8,47 @@
...
 	/**
 	 * Initialtes the payment process.
 	 *
 	 * @return CompletedState The next state.
-	 * @throws StateTransitionException In case the completed state could not be initialized.
+	 * @throws StateTransitionException Completed state could not be initialized.
+	 * @throws Order_Not_Found_Exception Order could not be found.
 	 */
 	public function process() {
+		$order_id       = $this->get_context()->get_order_id();
```
zmaglica, line 50: Wouldn't be wise here to store the whole instance of WP_Order and prevent calling the order from the database on every step?

src/Internal/Service/OrderService.php
```diff
@@ -0,0 +1,51 @@
...
+	/**
+	 * Set the payment metadata for payment method id.
+	 *
+	 * @param int    $order_id          ID of the order.
+	 * @param string $payment_method_id The value to be set.
+	 *
+	 * @throws Order_Not_Found_Exception
+	 */
+	public function set_payment_method_id( int $order_id, string $payment_method_id ) {
+		$this->legacy_service->set_payment_method_id_for_order( $order_id, $payment_method_id );
```
zmaglica, line 49: Do we plan to extract this function from a legacy part of code into its own service class?

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -40,12 +55,40 @@ public function __construct( StateFactory $state_factory ) {
...
-		$state   = $state->process();
+		// Start with a basis context.
+		$context = $this->create_payment_context( $order_id );
+
+		// Add details from the request.
+		$request = $this->create_payment_request();
+		$request->populate_context( $context );
+
+		$state = $this->state_factory->create_state( InitialState::class, $context );
+		$state = $state->process();
```
zmaglica, line 71: We are not planning to handle any exceptions upon this point?

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: IMHO, modifying `$_POST` is something we should avoid. I would create a DTOish class (or pass an array as you suggested) and pass it to the process_payment method. We can even store WP request object inside the payment context, so we can use it anywhere if needed.

includes/class-wc-payments.php
```diff
@@ -1257,6 +1257,15 @@ public static function get_customer_service_api(): WC_Payments_Customer_Service_
 		return self::$customer_service_api;
 	}
 
+	/**
+	 * Returns the order service instance.
+	 *
+	 * @return WC_Payments_Order_Service
+	 */
+	public static function get_order_service(): WC_Payments_Order_Service {
+		return self::$order_service;
```
zmaglica, line 1290: I noticed in PR description that we still use the legacy Order Service. I think that mixing from the 'old' way and the 'new' way could harm us eventually. Is there a chance to migrate this into its own dedicated service without relaying on legacy one?

src/Internal/Payment/PaymentContext.php
```diff
@@ -7,9 +7,92 @@
 
 namespace WCPay\Internal\Payment;
 
+use Automattic\WooCommerce\Blocks\Payments\PaymentMethodTypeInterface;
+use WCPay\Internal\Payment\PaymentMethod\PaymentMethodInterface;
+
 /**
  * A context object, which is shared between payment states.
  */
 class PaymentContext {
```
zmaglica, line 16: Do you maybe consider exposing an interface to the `PaymentContext` class?

zmaglica, line 15: Do you maybe consider exposing an interface to the PaymentContext class?

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -24,13 +27,25 @@ class PaymentProcessingService {
...
+
 	/**
 	 * Service constructor.
 	 *
 	 * @param StateFactory $state_factory Factory for payment states.
+	 * @param LegacyProxy  $legacy_proxy  Legacy proxy.
 	 */
-	public function __construct( StateFactory $state_factory ) {
+	public function __construct(
+		StateFactory $state_factory,
```
zmaglica, line 45: Will every State have the same dependencies as these? Can we expose those into payment context and call them within the state if needed via payment context call?

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
RadoslavGeorgiev: I'll go with the parameter to `process_payment` approach. Keep in mind that `populate_context` might be a suitable candidate to being the one responsible for validating if all necessary data is present.

Providing the request object inside the context might hide issues until they appear later during the runtime.

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: This comment seems to ignore the fact that `set()`, `get()` and `$data` are all private and can only be accessed through the strongly typed getters and setters.

> immutability (if necessary)

I don't think we require anything here to be immutable, do we? At some point, the context object will be shared between requests, and follow-up requests might provide new details. We want to use `set()` to make sure that those changes are logged, but we'll trust the latest input anyway.

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: I know this feels weird at first, but it is just something to get used to, and that should not be hard, considering we're already using snake case within `includes`.

Also, here we're following the example of Woo core. Random file example: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/Orders/TaxesController.php

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: I think the order ID is the only thing that might make sense as immutable, but everything else could change between requests. And the order ID is a specific case itself, as orders and payments should (probably?) have a 1:1 relationship. With that in mind, I'd put the order ID in the constructor, and remove the setter. WDYT?

src/Internal/Payment/PaymentRequest.php
RadoslavGeorgiev: Done https://github.com/Automattic/woocommerce-payments/pull/7324/commits/2100acc40c7794715516fbf9ecaa4fc8edd0474f

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: I'm 50/50 here. @marcinbot can join the conversation, as this is done per his suggestion. IMO it makes sense to keep the `PaymentContext` simple and easily serializable, but at the same time, now we need to mock a ton of `get_order()` calls within simple methods, and there are `Order_Not_Found` exceptions that need to be documented throughout states and services. Providing an order in advance would eliminate that,

src/Internal/Service/OrderService.php
RadoslavGeorgiev: Yes, but that's out of scope for the process. In fact, there's a comment about that here: https://github.com/Automattic/woocommerce-payments/blob/2100acc40c7794715516fbf9ecaa4fc8edd0474f/src/Internal/Service/OrderService.php#L23-L24

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: Not yet, or at least not here. There's an example in a different PR:

https://github.com/Automattic/woocommerce-payments/blob/a044d358cab255ce8cae5ce4a7a72e57ee586f6f/src/Internal/Payment/State/InitialState.php#L100-L104)

Right now, I'm only trying to cover the bare minimum, to allow multiple people to continue working on this. Erroneous states and exception handling will be a part of that effort.

includes/class-wc-payments.php
RadoslavGeorgiev: I'm glad you caught this, as it is one of the topics I'd like to discuss. Now, the approach here is objective. There are a few scenarios:

1. "Just" migrate the whole service to `src`. However, if we want to stick to order IDs instead of objects, as you've noticed we're doing in this series of PRs, we'd need to change quite a bit of code, and that would be quite the effort, which we don't have time for presently.
2. Use the new service for calls with `$order_id` and new code, leaving the migration to legacy code for some point in the future. I'm ok with this approach.
3. Copy code from the existing service into the new one. This will lead to code duplication, which until we get to migrate the whole service, might become unmanageable.

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: Note: The `Automattic\WooCommerce\....` dependency was not used, and I removed it in https://github.com/Automattic/woocommerce-payments/pull/7324/commits/86f09467217768a9b9ed3f141ce844f3584c5855.

> Do you maybe consider exposing an interface to the PaymentContext class?

What interfaces to use based on SOLID principles is one of the things to discuss. I, personally, wouldn't start with the `PaymentContext` because as a DTO-like object, it's got too many methods that would need to be defined in the interface.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: > Will every State have the same dependencies as these? 

No, states will be like controllers, and they will depend on varying services.

> Can we expose those into payment context and call them within the state if needed via payment context call?

We don't want the payment context to have any dependencies. Also, that would force all dependencies through the `PaymentContext`, and at some point, adding/modifying a dependency, would require fixing hundreds of tests. This is one of the issues with the current gateway classes, and one we definitely need to fix. I don't want to have to spend 2 days just fixing tests, only because I dared to add a simple dependency.

includes/class-wc-payments.php
marcinbot: There's also an option 4:
Have a new `OrderService` which is actually just a wrapper for the old one. It's methods would look like this:
```
public function attach_intent_info_to_order( $order_id, /* all the args */ ) {
    $order = wc_get_order( $order_id );
    return $legacy_service->attach_intent_info_to_order( $order, /* args */ );
}
```

It's a bit more typing, but sets us up for not being too tied to the old service and even refactoring/migrating the old methods at any time.

EDIT: looks like that's what you're intending to do, after scrolling down in this PR

src/Internal/Payment/PaymentContext.php
marcinbot: I agree with Rado. I think having defined properties is better than the internal array (the server Models are a lesson learned), but as long as we use proper getters and setters with type hints then I think it works out to about the same.

src/Internal/Payment/State/InitialState.php
marcinbot: The WC order is neither a service, nor a model, so it doesn't fit easily into our patterns.

> now we need to mock a ton of get_order() calls within simple methods

This is by design and should happen only in the `OrderService`. Everything else should use IDs passed to the service and manipulate/read order data through it.

> and there are Order_Not_Found exceptions that need to be documented throughout states and services.

That's also fine, it's just comments. Also I think we'd run into those potential exceptions anyway (maybe only once at the beginning of the request but still).

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7324/commits/014155f907b6a09c53ef279c3baded1a446ec033. @zmaglica do you approve it this way?

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: I think we could keep it this way for now, and only consider moving back to an object in case it gets too overwhelming.

includes/class-wc-payments.php
RadoslavGeorgiev: It kind of is, but also not: I am using the new `OrderService` as a wrapper for the old one, but it's not just that: New functionality is added to the new service. That's option 2.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: @zmaglica should I consider this one resolved?

includes/class-wc-payments.php
marcinbot: > I am using the new OrderService as a wrapper for the old one, but it's not just that: New functionality is added to the new service. That's option 2.

Meaning that the new `OrderService` is both a wrapper and new functionality is added to it? That's absolutely fine too :)

src/Internal/Payment/PaymentContext.php
htdat: > Done in https://github.com/Automattic/woocommerce-payments/commit/014155f907b6a09c53ef279c3baded1a446ec033

Looks fine to me. But I think it's worth adding a check in `set()` and avoid re-setting value for the `order_id` key. I.e. if this key `order_id`, throws an exception to tell developers something is wrong!

src/Internal/Payment/PaymentRequest.php
```diff
@@ -144,4 +144,19 @@ public function get_payment_method(): PaymentMethodInterface {
...
+	 *
+	 * This method is the link between the payment request, and the payment process.
+	 * Use it to make sure that all necessary parameters are provided in advance,
+	 * or throw an exception otherwise. Once done, the payment process would rely
+	 * on all needed parameters being in place.
+	 *
+	 * @param PaymentContext $context Context to populate.
+	 * @throws PaymentRequestException When data is not available or invalid.
+	 */
+	public function populate_context( PaymentContext $context ) {
```
htdat, line 159: This method here does not feel right to me. I don't think `PaymentRequest` needs to be aware or act on anything related to `PaymentContext` in its own class.

IMO, we would move this method to `PaymentProcessingService`, where it interacts with all relevant payment stuff: states (in a state machine), PaymentContext, and PaymentRequest. We may try to separate these stuff as much as possible.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -41,13 +56,39 @@ public function __construct( StateFactory $state_factory ) {
...
 	public function process_payment( int $order_id ) {
-		$context = new PaymentContext();
-		$state   = $this->state_factory->create_state( InitialState::class, $context );
-		$state   = $state->process();
+		// Start with a basis context.
+		$context = $this->create_payment_context( $order_id );
+
+		// Add details from the request.
+		$request = $this->create_payment_request();
+		$request->populate_context( $context );
```
htdat, line 68: Nitpick: Should these steps be merged into one method in `PaymentProcessingService`? I don't think we need to be so details into the `process_payment` method to get the right PaymentContext.

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
```diff
@@ -7,26 +7,33 @@
...
-	 * @param WC_Payment_Token $token The WooCommerce token.
+	 * @param string $id       External ID of the payment method.
+	 * @param int    $token_id Internal WooCommerce token ID.
 	 */
-	public function __construct( WC_Payment_Token $token ) {
-		$this->token = $token;
+	public function __construct( string $id, int $token_id ) {
+		$this->id       = $id;
+		$this->token_id = $token_id;
 	}
```
htdat, line 37: I understand we'd like to simplify `SavedPaymentMethod` so that it can be easily serializable in `PaymentContext`. 

But these construct parameters lose the tight contract with `WC_Payment_Token`. IMO, we should still `WC_Payment_Token` as a single ~contract~ _construct_ parameter, but does not save it within the `SavedPaymentMethod` object.

```suggestion
	public function __construct( \WC_Payment_Token $woo_token ) {
		$this->id       = $woo_token->get_id();
		$this->token_id = $woo_token->get_token();
	}
```

htdat (commented):
The direction looks good to me. 
I have just some comments to improve the readability and contract tight.

_Update_: manual test runs as expected too.

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: @htdat, in the last commit (you quoted it), I moved `$order_id` out of the `$data` array. So far, it looks like payments will belong to orders, and with that in mind, there should be a connection from the order to the payment, but not the other way around. This way, when we start saving the payment context, we would only care about serializing the `$data` array.

src/Internal/Payment/PaymentRequest.php
RadoslavGeorgiev: @htdat please check this conversation: https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1342767258

I'll be moving this context population to `InitialState` to keep all context population in the same place. I got hung up on some tests last evening, but I'm planning to basically move the `populate_context` method to the state, and to provide the `PaymentContext` to it.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: This is only done the way it is, to allow mocks in tests. Keeping the instantiation of both `PaymentContext` and `PaymentRequest` within `process_payment` would not only make the tests for `process_payment` enormous, but also force us to test everything together, instead of individually (units).

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
RadoslavGeorgiev: Please see https://github.com/Automattic/woocommerce-payments/pull/7324#discussion_r1344193946. The reason to avoid carrying the `WC_Token` object around is the same as orders, and interact through a token (could be called something else) service. It seemed a bit weird at first, but realistically we don't do much with tokens throughout the payment process. As long as I can call `$context->get_payment_method()->get_id()`, I'm happy. :)

src/Internal/Payment/PaymentRequest.php
htdat: Ah, I did not notice that conversation. I will follow up in that convo.

htdat: And resolving this conversation now.

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
htdat: Agreed that we would avoid carrying Woo core objects. However, in my suggestion, we're NOT carrying `WC_Payment_Token`, we just simply use it as a construct parameter to populate the private properties for `SavedPaymentMethod`.

src/Internal/Payment/PaymentContext.php
htdat: >  I moved $order_id out of the $data array.

Aha, it's my oversight. If so, it looks good now IMO.

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
RadoslavGeorgiev: I have considered this option, the problem is that we would require the token service every time we want to re-construct the object because we'd need to load a `WC_Payment_Token` object. AFAIK, tokens would not change: I'm not aware of any flows, which would change the token within a `WC_Payment_Token` object. With that in mind, do you think that there is an upside to having the object?

htdat: > the problem is that we would require the token service every time we want to re-construct the object because we'd need to load a `WC_Payment_Token` object.

It's the problem if we want to re-construct the `SavedPaymentMethod` object often. But I think when it's attached to `PaymentContext`, we mostly only care `$context->get_payment_method()->get_id()` (similar to what you mentioned above). 

Anyway, this is just a suggestion rather than a blocker. I also agree that we do not often instantiate `SavedPaymentMethod`.

RadoslavGeorgiev:
@marcinbot, @htdat, and @zmaglica: All feedback so far has been addressed here, and the PR is ready for another review round.

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
marcinbot: I think we could do a middle ground, keep the constructor as is, and then have a static constructor like `SavedPaymentMethod::from_wc_payment_token` which could also perform any checks if the token is valid. 

I like that the current actual constructor takes only simple data, though :)

RadoslavGeorgiev: > `SavedPaymentMethod::from_wc_payment_token` which could also perform any checks if the token is valid.

What would it validate, and as a static method, could it do it without a dependency?

marcinbot (approved):

src/Internal/Payment/PaymentMethod/SavedPaymentMethod.php
marcinbot: > What would it validate

I'm assuming the Token can have some data missing, or be for a wrong gateway. Not a big deal, just mentioned it in case we wanted to do such a validation.

> as a static method, could it do it without a dependency?

Why not, the method body would be essentially what Dat has proposed in his first comment here. Just "unpacking" the token object. The validation, if we even need it, would also be a bunch of pure functionality, like checking the string structure or comparing against null.

RadoslavGeorgiev: We could, but we have not done anything like this so far, and I think we're on the safe side. Keep in mind that the `PaymentRequest` class already checks if the gateway is correct, and I wouldn't assume an empty token can be stored.

I was thinking that we'd need to validate whether the token still exists, but that will happen during reconstruction anyway. If we get to a point with duplicate code, we could always create the static method. Presently, it seems not needed, right?

marcinbot: Right, it was just a suggestion, not a requirement.

htdat (approved):

PR #7330, opened 2023-09-29
Move hooks out of MultiCurrency class constructors into own init_hooks methods

### LINKED ISSUE #7258: MultiCurrency - remove hooks from constructors
### Description
See base PR: https://github.com/Automattic/woocommerce-payments/pull/7249

We need to move the hooks from all class constructors to `init_hooks` and then call that method whenever the class is instantiated (or whenever we need the hooks to be attached, separately from instantiation).

The goal of this is to remove the side effects from constructors and allow us to instantiate classes freely.

### Acceptance criteria

- Base the PR on #7249 if not merged yet
- Remove the relevant exceptions from `dev/phpcs/ruleset.xml`
- Move the hooks to the new method and call them. Update the code and unit tests accordingly
- PHPCS should pass

jessepearson:
Fixes #7258

* Add new `init_hooks` methods to Multi-Currency related classes.
* Add calls to the new `init_hooks` methods when classes are created and/or initialized.
* Update tests to account for new `init_hooks` methods.

#### Testing instructions

* Tests should pass.
* You should be able to use Multi-Currency with no issues:
  * Access Multi-Currency settings in the admin.
  * Use the Multi-Currency widget on the front end.

-------------------


jessepearson:
@marcinbot MultiCurrency is unique in how it is loaded due to it was build with the possibility for it to be removed from WCPay in the future. Everything is auto loaded and there is an action hook that creates the initial instance of the `MultiCurrency` class. In order to make sure everything loads correctly, I had to add another action in order to call the `init_hooks` method. Can I get your thoughts on this?

_Edit: 2023-10-02_
I really liked how @allie500 called `init_hooks` in his PR, so I updated this PR do to the same. The `init_hooks` method is called when the main class itself has its instance created, that way the hooks are always added when the instance gets created.

https://github.com/Automattic/woocommerce-payments/pull/7330/files#diff-c722c76bb4254c6fe1a32099d4ea5fa641addb44c5cd685a164e8baaff879e1aR199

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7330` or branch name `dev/7258-remove-hooks-from-multi-currency-constructor` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=dev/7258-remove-hooks-from-multi-currency-constructor) :rocket:

---

Build info: 

- Latest commit: b14cba66bcaa5950da7c92e5f0dbb9bdc0300f41
- Build time: 2023-10-03 13:11:28 UTC

includes/multi-currency/AdminNotices.php
```diff
@@ -21,9 +21,11 @@ class AdminNotices {
 	private $notices = [];
 
 	/**
-	 * Constructor
+	 * Initializes this class's WP hooks.
+	 *
+	 * @return void
 	 */
-	public function __construct() {
+	public function init_hooks() {
```
jessepearson, line 28: This class only had hooks in the `__construct` method, so it was replaced with `init_hooks`.

includes/multi-currency/Helpers/OrderMetaHelper.php
```diff
@@ -39,7 +39,18 @@ class OrderMetaHelper {
...
 		$this->payments_api_client = $payments_api_client;
-		$this->add_actions();
+	}
+
+	/**
+	 * Initializes this class' WP hooks.
+	 *
+	 * @return void
+	 */
+	public function init_hooks() {
```
jessepearson, line 49: This class already had a separate method to load hooks/filters, so just renamed it and made it public so it could be called externally outside of the constructor.

includes/multi-currency/MultiCurrency.php
```diff
@@ -240,6 +247,8 @@ public function __construct( WC_Payments_API_Client $payments_api_client, WC_Pay
 			add_action( 'init', [ $this, 'possible_simulation_activation' ], 13 );
 			add_action( 'woocommerce_created_customer', [ $this, 'set_new_customer_currency_meta' ] );
 		}
+
+		$this->currency_switcher_block->init_hooks();
```
jessepearson, line 253: I was unsure of how to include the `init_hooks` call for the `currency_switcher_block` due to it is created in the constructor of this class. Including here causes no side effects.

includes/multi-currency/wc-payments-multi-currency.php
```diff
@@ -17,6 +17,13 @@ function WC_Payments_Multi_Currency() { // phpcs:ignore WordPress.NamingConventi
 }
 
 add_action( 'plugins_loaded', 'WC_Payments_Multi_Currency', 12 );
+add_action(
+	'plugins_loaded',
+	function() {
+		WC_Payments_Multi_Currency()->init_hooks();
+	},
+	12
+);
```
jessepearson: Added the call to `init_hooks` via an action here due to that's how MC is initialized in the first place. The action is added after the one that initializes MC with the same priority so that it is done immediately after.

includes/multi-currency/MultiCurrency.php
marcinbot: This makes sense - I did the same in a few other classes.

includes/multi-currency/AdminNotices.php
marcinbot: It's great that we're catching inconsistencies like that :)

jessepearson:
> Looks good to me so far - let's rebase it now that the base PR has been merged.

Thanks for taking a look, just merged `develop` into this branch.

marcinbot (approved):
:shipit:

PR #7332, opened 2023-09-30
RPP: Minimal Checkout

### LINKED ISSUE #7497: RPP: Minimal Checkout
See https://github.com/Automattic/woocommerce-payments/pull/7332

RadoslavGeorgiev:
Fixes #7497

This PR adds everything needed for a happy-path basic payment to the new checkout process.

- Empty `SystemErrorState` and `PaymentErrorState` classes are added.
- `PaymentContext` receives support for the following props: `currency`, `manual_capture`, `metadata`, `level3_data`, `metadata`, `cvc_confirmation`, `fingerprint`, `user_id` and `customer_id`
- `PaymentRequest`:
	- Has new getters for `cvc_confirmation` and `fingerprint`.
	- Populates the two new proeprties into the context.
- The `process` method of `InitialState` now calls all needed services to process a simple payment:
	- Import data from the order to the context through the order service.
	- Generate metadata through the order service.
	- Generate level3 data through the new Level3 service.
	- Create a customer (no updates yet).
	- Create an intent.
	- Deletegate to the order service to add all necessary details to the order, and update its status.
- `OrderService` is now responsible for both loading data, and post-processing. We will need to split it in a meaningful way, suggestions are welcome. Keep in mind that we'll probably work on them later through a separate issue.
- The new `PaymentRequestService` is responsible for generating API/server requests, keeping things composed instead of inheriting or huge classes.
- The legacy `WC_Payments_Customer_Service` class has a new method to create a customer based on order details. Should we keep it there, or add it to a new service?

#### Testing instructions

As always, check tests first, and make sure that they are sound.

With this PR, checkout with a saved/new payment method should work, adding the necessary order status, data and notes. Sounds simple, but just try it.

-------------------

- [x] Covered with tests (100% coverage 😎)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7332` or branch name `rpp/basic-process` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/basic-process) :rocket:

---

Build info: 

- Latest commit: 25fd14455b572dfcd47e4df0071360c6d852ad1a
- Build time: 2023-10-06 07:11:07 UTC

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -810,9 +810,11 @@ function_exists( 'wcs_order_contains_subscription' )
 	 * @return array|null     Array if processed, null if the new process is not supported.
 	 */
 	public function new_process_payment( WC_Order $order ) {
+		$manual_capture = $this->get_capture_type() === Payment_Capture_Type::MANUAL();
+
 		// Important: No factors are provided here, they were meant just for `Feature`.
 		$service = wcpay_get_container()->get( PaymentProcessingService::class );
```
zmaglica, line 819: Nitpick: Maybe we could use a better variable name here.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -810,9 +810,11 @@ function_exists( 'wcs_order_contains_subscription' )
 	 * @return array|null     Array if processed, null if the new process is not supported.
 	 */
 	public function new_process_payment( WC_Order $order ) {
+		$manual_capture = $this->get_capture_type() === Payment_Capture_Type::MANUAL();
```
zmaglica, line 816: Nitpick: We don't have to define a new variable. We could simply copy this statement inside brackets on line 817.

src/Internal/Payment/State/InitialState.php
```diff
@@ -45,10 +81,32 @@ public function __construct(
...
 	 */
 	public function process() {
-		$order_id       = $this->get_context()->get_order_id();
-		$payment_method = $this->get_context()->get_payment_method();
-		$this->order_service->set_payment_method_id( $order_id, $payment_method->get_id() );
+		$context  = $this->get_context();
+		$order_id = $context->get_order_id();
+
+		// Start by setting up all local objects.
+		$this->order_service->import_order_data_to_payment_context( $order_id, $context );
```
zmaglica: Do you think that this method needs to be inside payment context since it is related to payment context, and not from the order itself?

src/Internal/Payment/State/InitialState.php
```diff
@@ -45,10 +81,32 @@ public function __construct(
...
+		// Customer management involves a remote call.
+		$customer_id = $this->customer_service->get_or_create_customer_id_from_order(
+			$context->get_user_id(),
+			$this->order_service->get_order( $order_id )
+		);
+		$context->set_customer_id( $customer_id );
+
+		// Payments are currently based on intents, request one from the API.
+		try {
+			$intent = $this->payment_request_service->create_intent( $context );
```
zmaglica, line 102: In light of the new payment gateway, do you think that we could rename this function to something like create_new_payment_request (or something similar) and not to be coupled with the Stripe implementation and naming?

src/Internal/Payment/State/InitialState.php
```diff
@@ -45,10 +81,32 @@ public function __construct(
...
+		} catch ( Invalid_Request_Parameter_Exception | Extend_Request_Exception | Immutable_Parameter_Exception $e ) {
+			return $this->create_state( SystemErrorState::class );
+		}
+
+		// Intent available, complete processing.
+		$this->order_service->update_order_from_successful_intent( $order_id, $intent, $context );
```
zmaglica, line 108: maybe we could rename this function to `update_order_from_successful_intent( $order_id, $payment_object, $contect)`

src/Internal/Payment/State/SystemErrorState.php
```diff
@@ -0,0 +1,18 @@
...
+
+namespace WCPay\Internal\Payment\State;
+
+/**
+ * Erroneous state, caused by a bug or an edge case.
+ *
+ * Consider this state an exception: Ideally it would never be entered,
+ * but in case any sort of software issue arrises, it will be used.
+ */
+class SystemErrorState extends PaymentState {
```
zmaglica: I like custom exceptions, but I think that we could add custom error codes and messages, so we don't have to manually add them when the exception is handled.

src/Internal/Service/OrderService.php
```diff
@@ -126,4 +142,105 @@ public function get_payment_metadata( int $order_id, Payment_Type $payment_type
...
+	 * @param WC_Payments_API_Abstract_Intention $intent   Remote object. To be abstracted.
+	 * @param PaymentContext                     $context  Context for the payment.
+	 * @throws Order_Not_Found_Exception
+	 */
+	public function update_order_from_successful_intent(
+		int $order_id,
+		WC_Payments_API_Abstract_Intention $intent,
+		PaymentContext $context
+	) {
+		$order = $this->get_order( $order_id );
```
zmaglica, line 189: I have noticed that we keep getting order from DB. Can we store order in Payment Context and use it when needed?

src/Internal/Service/PaymentRequestService.php
```diff
@@ -0,0 +1,45 @@
...
+	 * Creates a payment intent. To be replaced by an abstraction soon.
+	 *
+	 * @param PaymentContext $context Context for the payment.
+	 * @return WC_Payments_API_Payment_Intention
+	 * @throws Invalid_Request_Parameter_Exception
+	 * @throws Extend_Request_Exception
+	 * @throws Immutable_Parameter_Exception
+	 */
+	public function create_intent( PaymentContext $context ) {
+		$request = Create_And_Confirm_Intention::create();
```
zmaglica, line 31: Nitpick: Perhaps we can create a function inside the request class called :from_payment_context($context) and call these setters if we are planning to use it in multiple places.

src/Internal/Service/PaymentRequestService.php
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Creates a payment intent. To be replaced by an abstraction soon.
+	 *
+	 * @param PaymentContext $context Context for the payment.
+	 * @return WC_Payments_API_Payment_Intention
+	 * @throws Invalid_Request_Parameter_Exception
+	 * @throws Extend_Request_Exception
+	 * @throws Immutable_Parameter_Exception
+	 */
+	public function create_intent( PaymentContext $context ) {
```
zmaglica, line 30: What do you think that we should create a payment-agnostic function name like create_payment()... ?

src/Internal/Service/PaymentRequestService.php
```diff
@@ -0,0 +1,45 @@
...
+		$request->set_payment_method( $context->get_payment_method()->get_id() );
+		$request->set_customer( $context->get_customer_id() );
+		$request->set_capture_method( $context->should_capture_manually() );
+		$request->set_metadata( $context->get_metadata() );
+		$request->set_level3( $context->get_level3_data() );
+		$request->set_payment_methods( [ 'card' ] ); // Initial payment process only supports cards.
+		$request->set_cvc_confirmation( $context->get_cvc_confirmation() );
+		$request->set_fingerprint( $context->get_fingerprint() );
+
+		return $request->send( 'wcpay_create_and_confirm_intent_request_2' ); // @todo: This is weird.
```
zmaglica: Maybe you can add a brand-new filter for the state pattern if needed.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: The service class is on an adjacent line to where `$service` is used. This, combined with the next comment, would end up in this:

```php
// Important: No factors are provided here, they were meant just for `Feature`.
$payment_processing_service = wcpay_get_container()->get( PaymentProcessingService::class );
$state                      = $payment_processing_service->process_payment( $order->get_id(), $this->get_capture_type() === Payment_Capture_Type::MANUAL() );
```

I'd prefer to keep it as-is for readability. This is in combination with https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1342761826.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Answered here: https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1342853886

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Right now, we're trying to keep the payment context dependency-free, which is why this is not there.

However, I'm looking for a suitable name for a service, which links orders and payments. Something like `OrderPaymentService`? We'll have many small methods like this that we'll need to put there.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: My plan is for this topic to become a whole stream of work, focused on abstraction. We'll need to do quite a bit there, and I presume that this method would be gone by that point.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: This is the same as the definition, I think you forgot to change the thing you had in mind.

src/Internal/Payment/State/SystemErrorState.php
RadoslavGeorgiev: This will also be a stream of work.

Keep in mind, that I have questions here myself:

1. How do we transfer the exception to all necessary layers, including the context and the caller?
2. Which exceptions, and how, would be directed to one state vs the other?
3. Do we store the exception in the database, or simply log the type of error it is (payment vs system)?

We'll get there :)

src/Internal/Service/OrderService.php
RadoslavGeorgiev: I've responded in a different PR: https://github.com/Automattic/woocommerce-payments/pull/7324#discussion_r1342773519

src/Internal/Service/PaymentRequestService.php
RadoslavGeorgiev: One of the ideas within `src` is that all static methods must be pure. Your suggestion would be harder to test. However, we must also consider how to construct those requests without the static `create` method, through a dependency instead.

src/Internal/Service/PaymentRequestService.php
RadoslavGeorgiev: As responded here, this will be a full stream of work: https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1342856660.

src/Internal/Service/PaymentRequestService.php
RadoslavGeorgiev: We'll create a new issue to make the WooPay service compatible with both formats. I didn't want to dilute this PR with that, as there are a few key aspects to discuss.

src/Internal/Payment/State/InitialState.php
zmaglica: `OrderPaymentService` looks nice solution to me. Thank you for providing a context regarding the dependency-free payment context.

src/Internal/Service/PaymentRequestService.php
zmaglica: Creating a request class via DI, would be a great improvement, IMHO. We can keep it as it is now, and later on update it.

src/Internal/Payment/State/InitialState.php
zmaglica: I think it is related to make function abstract which you addressed in previous comments. You can ignore the original comment.

src/Internal/Payment/State/InitialState.php
marcinbot: I think this should even be in the same place where you instantiate and populate the context (currently in the `PaymentProcessingService`). Seems like this is base data, similar to the `PaymentRequest`. It shouldn't be the context populating it, but it also shouldn't be the responsibility of the State, especially if everything else was populated in the other place.

src/Internal/Payment/State/InitialState.php
```diff
@@ -47,10 +83,32 @@ public function __construct(
...
+		$this->order_service->import_order_data_to_payment_context( $order_id, $context );
+		$context->set_metadata( $this->order_service->get_payment_metadata( $order_id ) );
+		$context->set_level3_data( $this->level3_service->get_data_from_order( $order_id ) );
+
+		// Customer management involves a remote call.
+		$customer_id = $this->customer_service->get_or_create_customer_id_from_order(
+			$context->get_user_id(),
+			$this->order_service->get_order( $order_id )
+		);
+		$context->set_customer_id( $customer_id );
```
marcinbot: Same here re: [context population](https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1344236115).

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: We've definitely got a mixed approach currently, and need to unify it by populating all the context in the same place. We need to consider whether that should be inside or outside the state, and to me, it seems like we have to do it **within the state**.

Here's why: Taking a look at the `get_or_create_customer_id_from_order` a couple of lines below. It is one that involves external calls, and should not be performed until verification has been performed. Looking at the verification within `Prepared_State` of the previous PoC (https://github.com/Automattic/woocommerce-payments/blob/spike/payment-process-states/includes/payment/state/class-prepared-state.php#L195), it seems like customer data is not required there, so at least for Stripe, we should:

1. Populate the context with request and order data.
2. Verify that everything is alright before proceeding.
3. Populate the context with data, which requires API calls.

`InitialState` seems like the place to do that, and the `process` method might just receive the request as a parameter. WDYT?

---

Another topic, data validation: The state (transition) methods are the entry point for the state machine, and different payment process entry points might have varying input data. It makes sense for the state to try to get it, and handle transitions to failure states, rather than throwing external exceptions. We'll discuss this in length when it comes to error/exception handling, but for now I prefer more logic to be orchestrated by states, rather than outside of them.

marcinbot: > InitialState seems like the place to do that, and the process method might just receive the request as a parameter. WDYT?

Sure, let's then commit to populating all of the context in the initial state. The initial state would start with an empty context, maybe containing the order ID only. Or the initial state could even take care of instantiating the context object.

RadoslavGeorgiev: > Sure, let's then commit to populating all of the context in the initial state.

Will do shortly.

> The initial state would start with an empty context, maybe containing the order ID only. Or the initial state could even take care of instantiating the context object.

In the future, I imagine something loading a previous context from the database, so I'll leave the instantiation external.

htdat: > In the future, I imagine something loading a previous context from the database, so I'll leave the instantiation external.

Yeah, I think so. An example is the required authentication flow (such as using 3DS cards). We would need to stop the state machine, save PaymentContext somewhere, wait for more input for shoppers, and then continue the state machine by re-loading PaymentContext and adding more information to it from PaymentRequest.

RadoslavGeorgiev: @marcinbot done: https://github.com/Automattic/woocommerce-payments/pull/7332/commits/43b9f48fe0fb5e596f9d8db3428648a9bb5a43ed

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Addressed in https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1347011963

src/Internal/Payment/State/InitialState.php
```diff
@@ -23,34 +32,118 @@ class InitialState extends AbstractPaymentState {
...
-		$this->order_service->set_payment_method_id( $order_id, $payment_method->get_id() );
+	public function process( PaymentRequest $request ) {
+		$context  = $this->get_context();
+		$order_id = $context->get_order_id();
 
+		// Populate basic details from the request.
+		$this->populate_context_from_request( $request );
+
+		// Start by setting up all local objects.
+		$this->order_service->import_order_data_to_payment_context( $order_id, $context );
```
marcinbot: Can we have something similar to `populate_context_from_request` with the order? Something like `populate_context_from_order`? I think it would help with readability.

src/Internal/Payment/State/InitialState.php
```diff
@@ -23,34 +32,118 @@ class InitialState extends AbstractPaymentState {
...
 	 * @throws StateTransitionException  In case the completed state could not be initialized.
 	 * @throws ContainerException        When the dependency container cannot instantiate the state.
 	 * @throws Order_Not_Found_Exception Order could not be found.
+	 * @throws PaymentRequestException   When data is not available or invalid.
 	 */
-	public function process() {
-		$order_id       = $this->get_context()->get_order_id();
-		$payment_method = $this->get_context()->get_payment_method();
-		$this->order_service->set_payment_method_id( $order_id, $payment_method->get_id() );
+	public function process( PaymentRequest $request ) {
```
marcinbot, line 90: Now that we have the request as a param passed in here, it almost feels like the order ID should be a param as well. WDYT?

marcinbot (commented):
Looks good to me and almost ready to be shipped. Two small suggestions below.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: On it... ⌛

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: The context should be tied to the order, and generated/restored based on it. I imagine the `create_payment_context` method evolving to something like `load_or_create_payment_context_from_order` at some point.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Done: https://github.com/Automattic/woocommerce-payments/pull/7332/commits/c2d822f7037a8751dfbdf3a79dad5077151cafee

src/Internal/Payment/State/InitialState.php
marcinbot: Ok, not a blocker, we can see how we progress. It makes sense for the context to be tied to an order.

src/Internal/Payment/State/InitialState.php
marcinbot: Sorry, I meant that this call and all the other calls below unpacking the order into context be moved to a private method in this class. So this:
```
$context->set_metadata( $this->order_service->get_payment_metadata( $order_id ) );
$context->set_level3_data( $this->level3_service->get_data_from_order( $order_id ) );

// Customer management involves a remote call.
$customer_id = $this->customer_service->get_or_create_customer_id_from_order( // ...
```
Would be packed into this:
```
$this->populate_context_from_order( $order_id );
```

RadoslavGeorgiev: I like the idea, the related test (https://github.com/Automattic/woocommerce-payments/blob/rpp/basic-process/tests/unit/src/Internal/Payment/State/InitialStateTest.php#L114) is unmanageably big. Will do shortly.

By the way, I'd need those methods to be at least `protected` to test them nicely. Is there a unit testing technique I'm not aware of? I know how to use reflection to test a private method, but testing the caller would still require setting everything up.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -52,43 +52,35 @@ public function __construct(
...
 	 * @throws Exception
 	 * @throws StateTransitionException In case a state cannot be initialized.
 	 * @throws PaymentRequestException  When the request is malformed. This should be converted to a failure state.
 	 * @throws ContainerException       When the dependency container cannot instantiate the state.
 	 */
-	public function process_payment( int $order_id ) {
+	public function process_payment( int $order_id, bool $manual_capture = false ) {
 		// Start with a basis context.
-		$context = $this->create_payment_context( $order_id );
+		$context = $this->create_payment_context( $order_id, $manual_capture );
```
htdat, line 65: Just a note, not something to be fixed. 

Passing $manual_capture from the gateway class seems not ideal, it would be better if we had something like `OptionService`. With a service like that, we do not need to pass multiple options like the code here.

But I think it's OK in this specific situation as in WooPayments settings, this seems the only option is passed to the payment processing.

src/Internal/Payment/State/InitialState.php
```diff
@@ -23,34 +32,118 @@ class InitialState extends AbstractPaymentState {
...
+		$customer_id = $this->customer_service->get_or_create_customer_id_from_order(
+			$context->get_user_id(),
+			$this->order_service->_deprecated_get_order( $order_id )
+		);
+		$context->set_customer_id( $customer_id );
+
+		// Payments are currently based on intents, request one from the API.
+		try {
+			$intent = $this->payment_request_service->create_intent( $context );
+		} catch ( Invalid_Request_Parameter_Exception | Extend_Request_Exception | Immutable_Parameter_Exception $e ) {
```
marcinbot, line 103: Not sure if this falls under the scope of this PR, but ideally we might want to catch all the exceptions here and transition to SystemError

src/Internal/Payment/State/InitialState.php
marcinbot: I think ideally we would have those methods in a separate service(s) that we could mock properly. Right now we're just improving reliability, but this might be an early sign that something is not right with our approach to how we split the responsibilities. Let's just keep that in mind as we work on more payment flows, but IMO it's fine to keep it as it is for now. Let's also make the methods protected if it will help with the test, although that's not ideal either.

RadoslavGeorgiev: Separated and tested in https://github.com/Automattic/woocommerce-payments/pull/7332/commits/4b173d5d1d12c65f9541e547772d42984541a1e8.

Let's see how things progress, and where we'll have too many responsibilities. For example, we could call `set_metadata` and `set_level3_data` from within `import_order_data_to_payment_context`, as it makes sense. But then we'd have too much in `import_order_data_to_payment_context`. For now, this seems good enough.

marcinbot (approved):
:shipit: either with or without the generic exception handling mentioned [here](https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1347205497) but let's make sure we can handle errors properly in the future.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Yes, now that we've got all the context population in the same place, this would be the place to handle exceptions. We'll also need to discuss (and implement) the separation between types of errors, how the exception is logged (incl. the logger class), and what is displayed to the buyer.

I have this in my list, and there will be a few separate issues related to it.

RadoslavGeorgiev:
@htdat the gateway type is now added to the metadata: https://github.com/Automattic/woocommerce-payments/pull/7332/commits/a72e816657a5f91ca2c26d557639ccd8b9259e89

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -48,40 +74,198 @@ class InitialStateTest extends WCPAY_UnitTestCase {
...
+
+	/**
+	 * Calls a protected/private method of the object.
+	 *
+	 * @param object  $object      The object with the method.
+	 * @param string  $method_name Name of the method.
+	 * @param mixed[] ...$args     Arguments for the invokation.
+	 * @return mixed               Result of the call.
+	 */
+	private function call_protected_method( $object, $method_name, ...$args ) {
```
htdat: We have a less well-know helper `PHPUnit_Utils::call_method` for the same purpose. 
https://github.com/Automattic/woocommerce-payments/blob/e4fce5af5fe6d5d83f0001bf13986b2f7e7193f3/tests/unit/helpers/class-phpunit-utils.php#L23-L23

src/Internal/Payment/State/InitialState.php
```diff
@@ -23,34 +32,137 @@ class InitialState extends AbstractPaymentState {
...
+	 *
+	 * This method is the link between the payment request, and the payment process.
+	 * Use it to make sure that all necessary parameters are provided in advance,
+	 * or throw an exception otherwise. Once done, the payment process would rely
+	 * on all needed parameters being in place.
+	 *
+	 * @param PaymentRequest $request The request to use.
+	 * @throws PaymentRequestException When data is not available or invalid.
+	 */
+	protected function populate_context_from_request( PaymentRequest $request ) {
```
htdat, line 125: `populate_context_from_request` and `populate_context_from_order` are only called within this state class. So let's make them `private` by default. Furthermore, I don't think we have any plan to inherit this state, therefore `protected` does not carry much meaning here.

src/Internal/Service/OrderService.php
```diff
@@ -121,7 +137,7 @@ public function get_payment_metadata( int $order_id, Payment_Type $payment_type
 
 		if (
 			'recurring' === (string) $payment_type
-			&& $this->legacy_proxy->call_function( 'function_exists', 'wcs_order_contains_subscription' )
+			&& $this->legacy_proxy->function_exists( 'wcs_order_contains_subscription' )
```
htdat: This seems a left-over from one of merges. 

Context: https://github.com/Automattic/woocommerce-payments/pull/7333#discussion_r1347045317

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -14,6 +14,18 @@
...
+	 * Checks if a function exists.
+	 *
+	 * Use this for WP, WC, and other generic non-native PHP functions.
+	 *
+	 * @param string $name Name of the function.
+	 * @return bool
+	 */
+	public function function_exists( string $name ): bool {
+		return function_exists( $name );
+	}
```
htdat: Seems a left-over from one of merges. 

Context: https://github.com/Automattic/woocommerce-payments/pull/7333#discussion_r1347045317

htdat (commented):
Besides my inline comments (most of which are nitpick), I've tried to test this PR in my local dev site. iI turns that there are some errors. @RadoslavGeorgiev - could you have a look at them?

1. After enabling enough factors in WCPay Dev Tools for my testing flow, I could test it, and face this error `Too few arguments to function WCPay\Internal\Payment\State\InitialState::__construct(), 4 passed and exactly 5 expected`. This can be easily fixed by updating `PaymentsServiceProvider` https://gist.github.com/htdat/5d15b253341e8c0e2e8f304ad40de47d

2. After applying the diff in the gist link above, I got another error, which I have not been able to figure out yet.  

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to WCPay\Core\Server\Request\Create_Intention::set_fingerprint() must be of the type string, null given, called in /var/www/html/wp-content/plugins/woocommerce-payments/src/Internal/Service/PaymentRequestService.php on line 41 and defined in /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/request/class-create-intention.php:156
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-payments/src/Internal/Service/PaymentRequestService.php(41): WCPay\Core\Server\Request\Create_Intention->set_fingerprint(NULL)
#1 /var/www/html/wp-content/plugins/woocommerce-payments/src/Internal/Payment/State/InitialState.php(102): WCPay\Internal\Service\PaymentRequestService->create_intent(Object(WCPay\Internal\Payment\PaymentContext))
#2 /var/www/html/wp-content/plugins/woocommerce-payments/src/Internal/Service/PaymentProcessingService.php(69): WCPay\Internal\Payment\State\InitialState->process(Object(WCPay\Internal\Payment\PaymentRequest))
#3 /var/www/html/w in /var/www/html/wp-content/plugins/woocommerce-payments/includes/core/server/request/class-create-intention.php on line 156
```

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: I have this in my list. There are more instances where not only we need to pass gateway options as parameters, but the gateway is required as a dependency just to get options.

Marking this as resolved.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: They are `protected` in order to allow mocks in tests. Not ideal, but something I have in mind. More context in https://github.com/Automattic/woocommerce-payments/pull/7332#discussion_r1347235754

src/Internal/Service/OrderService.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7332/commits/f2412c1eb6c18f43ad7e22c547998fa4fe53a886

RadoslavGeorgiev:
@htdat I pushed https://github.com/Automattic/woocommerce-payments/pull/7332/commits/f2412c1eb6c18f43ad7e22c547998fa4fe53a886, fixing most of your comments, and more importantly, making it work again.

Tomorrow I'll check for more left-overs from all the merges.

src/Internal/Service/PaymentRequestService.php
```diff
@@ -0,0 +1,51 @@
...
+		$request->set_capture_method( $context->should_capture_manually() );
+		$request->set_metadata( $context->get_metadata() );
+		$request->set_level3( $context->get_level3_data() );
+		$request->set_payment_methods( [ 'card' ] ); // Initial payment process only supports cards.
+		$request->set_cvc_confirmation( $context->get_cvc_confirmation() );
+
+		$fingerprint = $context->get_fingerprint();
+		if ( ! is_null( $fingerprint ) ) {
+			$request->set_fingerprint( $fingerprint );
+		}
```
htdat:
```suggestion
		$request->set_fingerprint( $context->get_fingerprint() ?? '' );

```

I think we would fallback to empty string `''`. Check other usages of fingerprint in the old gateway classes: 

- https://github.com/Automattic/woocommerce-payments/blob/f2412c1eb6c18f43ad7e22c547998fa4fe53a886/includes/class-wc-payment-gateway-wcpay.php#L1277
- https://github.com/Automattic/woocommerce-payments/blob/f2412c1eb6c18f43ad7e22c547998fa4fe53a886/includes/payment-methods/class-upe-payment-gateway.php#L360-L360

htdat (approved):
- LGTM, pre-approving! Only one minor comment, then it's good to ship 🚢. 
- I've run another manual testing too. Everything works like a charm 🎉

src/Internal/Service/PaymentRequestService.php
RadoslavGeorgiev: Wow, this one was too easy to underestimate. I thought that we were passing an empty string just to avoid `if()`s down the payment process, but it turns out that even with an empty string, we need to call `set_fingerprint()` because it is the link between the request, and the `Buyer_Fingerprinting_Service` class. See https://github.com/Automattic/woocommerce-payments/blob/rpp/basic-process/includes/core/server/request/trait-intention.php#L23 and https://github.com/Automattic/woocommerce-payments/blob/rpp/basic-process/includes/core/server/request/trait-intention.php#L24.

Defaulting to an empty string shortly...

RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7332/commits/8eb8bc11f30daec65e10614b7b511e6f32407885

PR #7333, opened 2023-10-01
RPP: Extract gateway methods

### LINKED ISSUE #7496: RPP: Extract gateway methods
See https://github.com/Automattic/woocommerce-payments/pull/7333

RadoslavGeorgiev:
Fixes #7496

While working on a minimum checkout solution (https://github.com/Automattic/woocommerce-payments/pull/7332) for the new payment process, it turned out that intent **metadata** and **level 3 data** are generated within in the gateway class, but independent of it, and required for the payment process. This PR focuses on moving:

1. Metadata generation to the new `OrderService`: It's a simple 32-line method.
2. Level 3 data generation to a new `Level3Service`, as it's much more elaborate.

After extracting those from the gateway, their usages were replaced with calls to the services through `wcpay_get_container()`. This is not ideal, as they should be proper dependencies, but once the payment process refactoring is done, those calls would be deleted.

#### Testing instructions

Perform test checkout on a US store, and verify in the Stripe dashboard that all data is correct.

Tests have been changed as well:

- There are new ones (100% coverage, as always) for the services.
- Existing gateway tests do not focus on level 3 and meta data anymore, instead only ensuring that the right request setter has been called. 

This is the reason the PR has so many changes.

-------------------

- [X] Tested on mobile (does not apply)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7333` or branch name `rpp/extract-gateway-methods` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/extract-gateway-methods) :rocket:

---

Build info: 

- Latest commit: d0f255454aec88942ddb384dc59acb87e2878afa
- Build time: 2023-10-05 11:58:09 UTC

src/Internal/Service/Level3Service.php
```diff
@@ -0,0 +1,192 @@
...
+			$level3_data['shipping_address_zip'] = $shipping_address_zip;
+		}
+
+		// The merchant’s U.S. shipping ZIP code.
+		$store_postcode = get_option( 'woocommerce_store_postcode' );
+		if ( WC_Payments_Utils::is_valid_us_zip_code( $store_postcode ) ) {
+			$level3_data['shipping_from_zip'] = $store_postcode;
+		}
+
+		return $level3_data;
```
zmaglica, line 115: I understand that you're developing an L3 service using the current code. Would it be possible to enhance its error resilience by returning a DTO instead of an array? Please note, I'm not suggesting there's anything wrong with the existing code; I'm merely proposing an alternative approach. What are your thoughts?

src/Internal/Service/Level3Service.php
```diff
@@ -0,0 +1,192 @@
...
+		];
+
+		// The customer’s U.S. shipping ZIP code.
+		$shipping_address_zip = $order->get_shipping_postcode();
+		if ( WC_Payments_Utils::is_valid_us_zip_code( $shipping_address_zip ) ) {
+			$level3_data['shipping_address_zip'] = $shipping_address_zip;
+		}
+
+		// The merchant’s U.S. shipping ZIP code.
+		$store_postcode = get_option( 'woocommerce_store_postcode' );
```
zmaglica: Could we create GetOptionService or something similar that will mimic this behavior when we need it? It could help us in unit testing, and we could 'cache' this function to avoid trip to the database. Perhaps we can implement it in another PR if you are ok with it.

src/Internal/Service/Level3Service.php
```diff
@@ -0,0 +1,192 @@
...
+		if ( $subtotal >= 0 ) {
+			$unit_cost       = $this->prepare_amount( $subtotal / $quantity, $currency );
+			$discount_amount = $this->prepare_amount( $subtotal - $item->get_total(), $currency );
+		} else {
+			// It's possible to create products with negative price - represent it as free one with discount.
+			$discount_amount = abs( $this->prepare_amount( $subtotal / $quantity, $currency ) );
+			$unit_cost       = 0;
+		}
+
+		return (object) [
```
zmaglica, line 152: Same idea as https://github.com/Automattic/woocommerce-payments/pull/7333/files#r1342602551

src/Internal/Service/Level3Service.php
```diff
@@ -0,0 +1,192 @@
...
+			array_map(
+				function( $cost, $qty ) {
+					return $cost * $qty;
+				},
+				array_column( $items, 'unit_cost' ),
+				array_column( $items, 'quantity' )
+			)
+		);
+
+		return (object) [
```
zmaglica, line 182: Same idea as https://github.com/Automattic/woocommerce-payments/pull/7333/files#r1342602551

src/Internal/Service/OrderService.php
```diff
@@ -48,4 +85,45 @@ public function __construct( WC_Payments_Order_Service $legacy_service ) {
...
+			&& $this->legacy_proxy->call_function( 'wcs_order_contains_subscription', $order, 'any' )
+		) {
+			$use_stripe_billing = $this->legacy_proxy->call_static( WC_Payments_Features::class, 'should_use_stripe_billing' );
+			$is_renewal         = $this->legacy_proxy->call_function( 'wcs_order_contains_renewal', $order );
+
+			$metadata['subscription_payment'] = $is_renewal ? 'renewal' : 'initial';
+			$metadata['payment_context']      = $use_stripe_billing ? 'wcpay_subscription' : 'regular_subscription';
+		}
+
+		return apply_filters( 'wcpay_metadata_from_order', $metadata, $order, $payment_type );
```
zmaglica: I'm concerned about combining filters and services, as it can obscure the original behavior of the service and make it challenging to trace and debug. During the RPP project, is there a plan to separate filters and actions from services, or we don't plan to implement such a feature?

src/Internal/Service/Level3Service.php
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Service for generating Level 3 data from orders.
+ */
+class Level3Service {
+	/**
+	 * Order service.
+	 *
+	 * @var OrderService
+	 */
+	private $order_service;
```
zmaglica, line 29: Since we are using PHP 7.4 now, can we maybe type hint `OrderService` here (and in similar places)? @htdat can you perhaps pull the data for us and tell how many merchants are using WooCommerce Payments for PHP version below 7.4?

zmaglica:
Great work on the PR! I'm truly impressed with the changes you've made and your strategic approach. I've left some comments, mostly minor suggestions, and a few ideas that we might consider for future use.

I do have a question regarding the tests. I've noticed that every time we use the get_metadata_from_order function in the tests, it seems to return an empty array instead of an array containing some data. I'm uncertain about the reason behind this behavior. Could you please provide further clarification on this issue?

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: For me, a DTO is supposed to help transfer data between different pieces of the same software. In this case, level 3 data is simply generated, and shipped to the server, without its contents being used by a different piece of the client. With this in mind, a DTO would likely not bring much in this case.

Also, for the changes in this particular PR, I'm migrating live and working code. I wouldn't want to risk changing it, at least not in the near future.

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: In this case I simply moved the code to the service, and didn't focus on this, which is my mistake. For simple methods like this, we're currently using the `LegacyProxy` class ([docs](https://github.com/Automattic/woocommerce-payments/blob/rpp/extract-gateway-methods/src/README.md#legacy-proxy)), which I will do here shortly.

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: And the same comment :)

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: And same comment :)

src/Internal/Service/OrderService.php
RadoslavGeorgiev: I don't think we've discussed such plans, though in this case, I think this is a very meaningful filter, and it makes sense to keep it.

I wouldn't mind opening another issue, in case you have anything specific in mind. For this instance, and especially for now, we should avoid breaking changes, caused by the payment process.

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: Considering that we'd still need PHPDoc comments (unless we change our PHPCS rules), and IDEs understand them well, is this worth doing for now? With Psalm in place, we cannot "hack" this anyway, we need the right types.

RadoslavGeorgiev:
> I do have a question regarding the tests. I've noticed that every time we use the get_metadata_from_order function in the tests, it seems to return an empty array instead of an array containing some data. I'm uncertain about the reason behind this behavior. Could you please provide further clarification on this issue?

There's a lot of redundant code being removed in this PR, and I decided to remove it. That's because we've already got 100% test coverage for the service, and as far as other classes are concerned, we just need to know that the service is being called, and whatever is returned, is being used. Is there any point in keeping those tests?

src/Internal/Service/Level3Service.php
RadoslavGeorgiev: Done: https://github.com/Automattic/woocommerce-payments/pull/7333/commits/2e95e2a3d0e411c0bc9fffded9c93a1e5adf24b1

src/Internal/Service/OrderService.php
RadoslavGeorgiev: By the way, we have the `HooksProxy`, and at least now the call is unit-tested: https://github.com/Automattic/woocommerce-payments/pull/7333/commits/4a1b58eb783a7584ae2b0b8e7e9251fb7ea3ee43 and https://github.com/Automattic/woocommerce-payments/pull/7333/commits/de977fef419d9d29a7ca3e787f2b742e26e4f13f

src/Internal/Service/Level3Service.php
zmaglica: While it's not a priority right now, we could consider adding it later as a minor enhancement or during a hack week. I believe it's not necessary to include it in this PR or even this project phase.

src/Internal/Service/OrderService.php
```diff
@@ -28,13 +33,56 @@ class OrderService {
...
+	 * Retrieves the order object.
+	 *
+	 * Unlike the legacy service, this one only accepts integer IDs,
+	 * and returns only the `WC_Order` object, no refunds.
+	 *
+	 * @param int $order_id ID of the order.
+	 * @return WC_Order Order object.
+	 * @throws Order_Not_Found_Exception If the order could not be found.
+	 */
+	public function get_order( int $order_id ): WC_Order {
```
marcinbot: We really should not allow anything outside of this service to get the order. If we need to manipulate or read the order data it should happen through this service.

The reason is mentioned [here](https://github.com/Automattic/woocommerce-payments/pull/7324#discussion_r1344193946):
> The WC order is neither a service, nor a model, so it doesn't fit easily into our patterns.

Having access to `WC_Order` objects allows other code to mess with the state of the WooCommerce system and make it unpredictable.

If we really want to return some object for readability, we could translate `WC_Order` to some sort of internal model, only carrying data, but not manipulating/saving it.

RadoslavGeorgiev: I understand your reasoning, and in general, the logic has been followed. However, there are some existing pieces of code, which warrant an exception, at least temporarily.

Take a look at `WC_Payments_Customer_Service::map_customer_data`: https://github.com/Automattic/woocommerce-payments/blob/2a93bcd4cbc410bb10202d5625f47f37dc3278c7/includes/class-wc-payments-customer-service.php#L312-L360

I know that `get_order(): WC_Order` here is a shortcut, but in order to achieve what you'd like, we'd need to do something regarding all of those methods, both in the context of the customer, and the order:

1. `get_billing_first_name()`
1. `get_billing_last_name()`
1. `get_billing_email()`
1. `get_billing_phone()`
1. `get_billing_address_1()`
1. `get_billing_address_2()`
1. `get_billing_postcode()`
1. `get_billing_city()`
1. `get_billing_state()`
1. `get_billing_country()`
1. `get_shipping_postcode()`
1. `get_shipping_first_name()`
1. `get_shipping_last_name()`
1. `get_shipping_address_1()`
1. `get_shipping_address_2()`
1. `get_shipping_city()`
1. `get_shipping_state()`
1. `get_shipping_country()`

I'd prefer to keep this specific part as-is for now, opening an issue to eventually tackle it earlier. I know we're using the project to do things differently in `src`, but this falls quite out of our initial scope.

---

FYI, `map_customer_data()` is used through this call in the follow-up PR: https://github.com/Automattic/woocommerce-payments/pull/7332/files#diff-40a5cd10ee78e0dddd7df5d2d9e9c4d746843c998d0fe38acf5b0911c2271e33R95

marcinbot: Yeah, I understand that, and understand that this would fall out of scope of this PR.

I think maybe two things would be good:
- Name the function something like `__deprecated_get_order` (following the React's `unsafe` functions) and add a comment why it shouldn't be used in new places.
- Open an issue to fix this, probably with our own wrapper for order data. We could even try to do something clever with `__call` magic method that will allow to access only getters on the underlying wc_order, but not the setters or the save method (or preferably declare the getters manually).

The better abstraction from Woo Core we achieve here, the more confident we can be that the future devs will follow our patterns.

RadoslavGeorgiev: Done in 9e5da7299 and here is the issue: https://github.com/Automattic/woocommerce-payments/issues/7367.

marcinbot: Let's just put the private method towards the bottom of the file, after all the public ones. Maybe in general, the `src` folder could use a PHPCS rule similar to what server has? IMO it makes the class files much more readable to have the public methods first. WDYT? We could add the rule in a separate PR.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1613,34 +1615,19 @@ public function set_payment_method_title_for_order( $order, $payment_method_type
 	 * @return array Array of keyed metadata values.
 	 */
 	protected function get_metadata_from_order( $order, $payment_type ) {
+		$service  = wcpay_get_container()->get( OrderService::class );
+		$metadata = $service->get_payment_metadata( $order->get_id(), $payment_type ) ?? [];
```
htdat: Nitpick: I don't think `$service->get_payment_metadata()` can return `null`. So using null coalescing operator here is a bit extreme. In case, it can return `null`, let's declare that in phpdoc `@return`. 

```suggestion
		$metadata = $service->get_payment_metadata( $order->get_id(), $payment_type )
```

src/Internal/Proxy/LegacyProxy.php
```diff
@@ -14,6 +14,18 @@
...
+	 * Checks if a function exists.
+	 *
+	 * Use this for WP, WC, and other generic non-native PHP functions.
+	 *
+	 * @param string $name Name of the function.
+	 * @return bool
+	 */
+	public function function_exists( string $name ): bool {
+		return function_exists( $name );
+	}
```
htdat: If I understand correctly, there is a need to mock `function_exists` (PHP core function) in PHPUnit tests. 
However, putting this method here does not sound right as itself is not a legacy of own code in `includes`. May better move it to a separate util class, such as `Util` or `PhpCoreUtil`?

src/Internal/Service/OrderService.php
```diff
@@ -28,13 +33,77 @@ class OrderService {
...
+	 *
+	 * @param int $order_id ID of the order.
+	 * @return WC_Order Order object.
+	 * @throws Order_Not_Found_Exception If the order could not be found.
+	 */
+	private function get_order( int $order_id ): WC_Order {
+		$order = $this->legacy_proxy->call_function( 'wc_get_order', $order_id );
+		if ( ! $order instanceof WC_Order ) {
+			throw new Order_Not_Found_Exception(
+				__( 'The requested order was not found.', 'woocommerce-payments' ),
```
htdat: Nitpick: May add $order_id in the exception message too?

src/Internal/Service/Level3Service.php
htdat: >  @htdat can you perhaps pull the data for us and tell how many merchants are using WooCommerce Payments for PHP version below 7.4?

See here <strike>32027-pb/#bash</strike> 32029-pb/#bash

As long as WooPayments still supports PHP 7.3, we should not use new syntaxes. But we can consider doing that during a HACK week or a cooldown week when we upgrade our requirement for PHP to 7.4

htdat (commented):
Overall, it looks good for me, just a few comments.

src/Internal/Proxy/LegacyProxy.php
RadoslavGeorgiev: > If I understand correctly, there is a need to mock function_exists (PHP core function) in PHPUnit tests.

Correct, mocking the function would allow us to have better code coverage, including conditions.

> However, putting this method here does not sound right as itself is not a legacy of own code in includes. May better move it to a separate util class, such as `Util` or `PhpCoreUtil`?

Maybe we'd need to update some definitions, but so far, I consider "legacy" everything outside `src`. So the legacy proxy is used even for Woo methods and functions, ex. `wc_get_order()`. With that in mind, do you still think it needs to be moved?

src/Internal/Service/OrderService.php
RadoslavGeorgiev: Good point, done here: https://github.com/Automattic/woocommerce-payments/pull/7333/commits/37d2a2d34e6d0336f23f6c7211b8de65574c6b83

src/Internal/Service/OrderService.php
RadoslavGeorgiev: > Let's just put the private method towards the bottom of the file, after all the public ones.

Done in https://github.com/Automattic/woocommerce-payments/pull/7333/commits/37d2a2d34e6d0336f23f6c7211b8de65574c6b83

> Maybe in general, the src folder could use a PHPCS rule similar to what server has? IMO it makes the class files much more readable to have the public methods first. WDYT? We could add the rule in a separate PR.

Added it to my list. Let's do that once we have the current set of PRs merged.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Fixed in https://github.com/Automattic/woocommerce-payments/pull/7333/commits/27859cab56c3ee53e1134ad2ffdbc261c0e4caea. Initially, I did this because I was planning to skip calls in tests, then I realized that it wouldn't be right.

src/Internal/Proxy/LegacyProxy.php
htdat: > Maybe we'd need to update some definitions, but so far, I consider "legacy" everything outside src. 

That's reasonable with this definition. Just improve the phpdoc a bit to tell that LegacyProxy would cover for PHP core functions if needed too.

However, that also means, you can use `LegacyProxy->call_function('function_exists', 'wcs_order_contains_subscription')` and there is no need to add this separate method.

RadoslavGeorgiev: Great point! Both are done in https://github.com/Automattic/woocommerce-payments/pull/7333/commits/ad55bcf1ed108dbc4588096429a92c818593ba30

marcinbot (approved):
Looks good to me! :shipit:

htdat (approved):

src/Internal/Service/OrderService.php
```diff
@@ -48,4 +92,74 @@ public function __construct( WC_Payments_Order_Service $legacy_service ) {
...
+	 * @param Payment_Type $payment_type Type of the payment (recurring or not).
+	 * @return array                     The metadat athat will be sent to the server.
+	 * @throws Order_Not_Found_Exception
+	 */
+	public function get_payment_metadata( int $order_id, Payment_Type $payment_type = null ) {
+		$order = $this->get_order( $order_id );
+
+		$name     = sanitize_text_field( $order->get_billing_first_name() ) . ' ' . sanitize_text_field( $order->get_billing_last_name() );
+		$email    = sanitize_email( $order->get_billing_email() );
+		$metadata = [
```
htdat, line 109: Oh, I think we missed one thing here. 
Should we add a new meta value to differentiate between the old processing and the new one, on which we're working here? 

Not necessarily to do that in this PR, we can create an issue to enhance this later.

RadoslavGeorgiev: It's not too late. I'll add this when fixing conflicts with the base PR.

RadoslavGeorgiev: Actually, I'll do that shortly in https://github.com/Automattic/woocommerce-payments/pull/7332, as it needs to be done outside of the service.

PR #7352, opened 2023-10-03
Fixed the version check regex to allow major versions > 9

### LINKED ISSUE #5706: Release workflows - allow major versions > 9
### Description
Today, we only allow `x.y.z(-test-n)` format for a release (or prerelease) version, with x from 0 to 9.

https://github.com/Automattic/woocommerce-payments/blob/fdaa6c6edd11579469248a0e3a005c3cb9329628/.github/actions/version-check/action.yml#L31-L35

That means: `9.0.0` or `9.0.0-test-2` versions are allowed, while `10.0.0` or `10.0.0-test-2` are not

We should allow the major version (`x`) to be any number, not just from 0 to 9, to prevent any future problems.

**Note**: version `7.0.0` is reached by the end of 2023, so we have some time before running into any issue related to this format check.

### Acceptance criteria

* We can create a prerelease package with version `9.0.0-test-1` or `10.0.0-test-1`
* We can create a release package with version `9.0.0` or `10.0.0`

leonardola:
Fixes #5706

This PR fixes the Regex used to check the release version to allow major versions > 9

#### Testing instructions
The easiest way is to simply check the regexes match given versions. 
- Copy one of them and paste in https://regexr.com/
- Type 9.9.0
- Make sure it matches
- Type 10.0.0 make sure it matches
- Try other versions that you think might break the regex but should actually work like 123.0.0

-------------------

**Post merge**

### DIFF
--- a/.github/actions/version-check/action.yml
+++ b/.github/actions/version-check/action.yml
@@ -29,9 +29,9 @@ runs:
         TRIMMED_VERSION=$(echo "$VERSION" | xargs) 
 
         if ${{ env.IS_PRERELEASE == 'true' }}; then
-          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]-test-[1-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]+-test-[1-9]$"
         else
-          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]+$"
         fi
 
         if [[ $TRIMMED_VERSION =~ $VERSION_FORMAT ]]; then
--- a/.github/actions/version-check/action.yml
+++ b/.github/actions/version-check/action.yml
@@ -29,9 +29,9 @@ runs:
         TRIMMED_VERSION=$(echo "$VERSION" | xargs) 
 
         if ${{ env.IS_PRERELEASE == 'true' }}; then
-          VERSION_FORMAT="^[0-9]\.[0-9]\.[0-9]-test-[1-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]-test-[1-9]$"
         else
-          VERSION_FORMAT="^[0-9]\.[0-9]\.[0-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]$"
         fi
 
         if [[ $TRIMMED_VERSION =~ $VERSION_FORMAT ]]; then
--- a/changelog/fix-major-version-check
+++ b/changelog/fix-major-version-check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Fixed version check regex
--- a/changelog/fix-major-version-check
+++ b/changelog/fix-major-version-check
@@ -1,4 +1,0 @@
-Significance: patch
-Type: fix
-
-Fixed version check regex
--- a/changelog/fix-major-version-check
+++ b/changelog/fix-major-version-check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed version check regex


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7352` or branch name `fix/major-version-check` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/major-version-check) :rocket:

---

Build info: 

- Latest commit: 427a205a4a8266297618fba529a9d7b76f1532c8
- Build time: 2023-10-10 14:22:24 UTC

jbordonado (approved):
Theoretically, we could also end up doing >= 10 patch releases for a version, e.g. 6.5.11
Maybe we could also use this PR for the patch version part of the regex?

Anyway, approving, it should work as expected for major versions

Thanks for taking up this task

changelog/fix-major-version-check
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Fixed version check regex
```
jbordonado, line 4: Just noticed it's not an insignificant changelog, and it will appear in the next release's changelog if it stays like that.

Could you redo it so it's marked as a "comment"? It should, in the end, be written as:

> Comment: Fixed version check regex

It's just a matter of leaving the first question empty after selecting "patch" I think (you'll see a text mentioning "insignificant", "leave empty" ...)

jbordonado (changes requested):

.github/actions/version-check/action.yml
```diff
@@ -29,9 +29,9 @@ runs:
         TRIMMED_VERSION=$(echo "$VERSION" | xargs) 
 
         if ${{ env.IS_PRERELEASE == 'true' }}; then
-          VERSION_FORMAT="^[0-9]\.[0-9]\.[0-9]-test-[1-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]-test-[1-9]$"
```
hsingyuc: Assuming we won't add the test version more than 3, besides that last one, we might want to add `+` after each of the `[0-9]`, or it could break with the number over 9. For example, `10.10.10-test-1`

.github/actions/version-check/action.yml
```diff
@@ -29,9 +29,9 @@ runs:
...
         else
-          VERSION_FORMAT="^[0-9]\.[0-9]\.[0-9]$"
+          VERSION_FORMAT="^[0-9]+\.[0-9]\.[0-9]$"
```
hsingyuc: Same as above

hsingyuc (commented):
@leonardola Thank you for working on this! I left two comments below.

.github/actions/version-check/action.yml
hsingyuc: Sorry, I didn't notice Jordan's comment.

jbordonado: I don't think `x.10.z` would ever happen given the way we follow the versioning in Transact, but `x.y.10` can happen yes

leonardola: Fixed with 51f4640bbed0b0730850add21c0ace3a976baae4

.github/actions/version-check/action.yml
leonardola: Fixed with 51f4640bbed0b0730850add21c0ace3a976baae4

changelog/fix-major-version-check
leonardola: Fixed with ad04862fcc10ea2e204dce88f7dfb8f248d5c7b8

hsingyuc (approved):
Thank you for the changes. LGTM

jbordonado (approved):

PR #7402, opened 2023-10-06
Disputes: Use existing tracks events where possible

### LINKED ISSUE #7395: Check & adjust disputes related tracks events impacted by shifting main dispute UI to transaction detail screen
### Describe the bug
We're in the process of moving the main dispute related UI into the transaction screen (#6883).

There are various flows and events that will change as part of this. E.g. flow traverses different screens, existing event is fired from different screen.

As part of this project we'll need to ensure all the existing tracks events still work as before (or deprecate/remove if no longer relevant. And also consider if there are any areas we need to add events.

[I've made a sheet (open access) here with all the dispute events I can find](https://docs.google.com/spreadsheets/d/1GD_m-CiKDno2eNHftWlnR2tdQ15eh5DF2w1bDhRfC6s/edit#gid=0), and investigated which might need changes. This is not complete, as not all dispute events include `dispute` in event name.

These events have shifted, or are new, and might need checking or tweaks to metadata / registration:

- `wcpay_dispute_challenge_clicked`
- `view_submitted_evidence_clicked`
- `dispute_challenge_click`
- `dispute_challenge_clicked`
- `dispute_accept_modal_view`
- `dispute_accept_click`
- `dispute_inquiry_refund_modal_view`
- `dispute_inquiry_refund_click`
- `payment_details_view_dispute_evidence_button_click`

#### why / outcome
The goals of this issue are:

- Review/test the affected events to confirm they are firing correctly and naming & metadata is consistent and correct.
- To check any existing funnels (especially if in use/current) and ensure they will work after 6.6 (events mean same thing before/after). 
  - Some funnels might no longer be valid (i.e. screen gone/assumptions changed), and if so we need to be aware of that, and possibly reach out to funnel user/owner.
  - Can focus on recent projects and planned impact measures for this project – anything else is lower priority.
- Do a first pass on ensuring our events are registered with correct metadata (this is not a blocker, but this is a convenient time to do that work).

brucealdridge:
Partially fixes #7395 

This PR will use current tracking events rather than new tracking events for disputes. 

There were some new tracking events added as part of moving dispute details from a dedicated page to the transaction detail page. I found that there were some events that are currently in use that can be used instead.

While the buttons are on a different page, the user intent is the same so keeping the event name and any historical data makes sense here.

| Button | Transaction Details (current event) | Dispute Details (previous event) |
|--------|--------|--------|
| `Challenge dispute` | `dispute_challenge_click` | `dispute_challenge_clicked` |
| `View submitted evidence` | `payment_details_view_dispute_evidence_button_click` | `view_submitted_evidence_clicked` |

#### Testing instructions

-------------------

**Post merge**

### DIFF
--- a/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
+++ b/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
@@ -258,7 +258,7 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 												.DISPUTE_CHALLENGE_CLICKED,
 											{
 												dispute_status: dispute.status,
-												on_page: 'transactions_details',
+												on_page: 'transaction_details',
 											}
 										);
 									} }
@@ -280,7 +280,7 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 										disputeAcceptAction.acceptButtonTracksEvent,
 										{
 											dispute_status: dispute.status,
-											on_page: 'transactions_details',
+											on_page: 'transaction_details',
 										}
 									);
 									setModalOpen( true );
@@ -340,7 +340,7 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 														dispute_status:
 															dispute.status,
 														on_page:
-															'transactions_details',
+															'transaction_details',
 													}
 												);
 												setModalOpen( false );
--- a/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
+++ b/client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
@@ -255,9 +255,10 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 									onClick={ () => {
 										wcpayTracks.recordEvent(
 											wcpayTracks.events
-												.DISPUTE_CHALLENGE_CLICK,
+												.DISPUTE_CHALLENGE_CLICKED,
 											{
 												dispute_status: dispute.status,
+												on_page: 'transactions_details',
 											}
 										);
 									} }
@@ -279,6 +280,7 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 										disputeAcceptAction.acceptButtonTracksEvent,
 										{
 											dispute_status: dispute.status,
+											on_page: 'transactions_details',
 										}
 									);
 									setModalOpen( true );
@@ -337,6 +339,8 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 													{
 														dispute_status:
 															dispute.status,
+														on_page:
+															'transactions_details',
 													}
 												);
 												setModalOpen( false );
--- a/client/tracks/index.js
+++ b/client/tracks/index.js
@@ -68,7 +68,7 @@ const events = {
 	DEPOSITS_ROW_CLICK: 'wcpay_deposits_row_click',
 	DEPOSITS_DOWNLOAD_CSV_CLICK: 'wcpay_deposits_download',
 	DISPUTES_ROW_ACTION_CLICK: 'wcpay_disputes_row_action_click',
-	DISPUTE_CHALLENGE_CLICK: 'wcpay_dispute_challenge_click',
+	DISPUTE_CHALLENGE_CLICK: 'wcpay_dispute_challenge_clicked',
 	DISPUTE_ACCEPT_CLICK: 'wcpay_dispute_accept_click',
 	DISPUTE_ACCEPT_MODAL_VIEW: 'wcpay_dispute_accept_modal_view',
 	DISPUTE_INQUIRY_REFUND_CLICK: 'wcpay_dispute_inquiry_refund_click',
@@ -84,7 +84,7 @@ const events = {
 		'wcpay_overview_deposits_change_schedule_click',
 	OVERVIEW_TASK_CLICK: 'wcpay_overview_task_click',
 	PAYMENT_DETAILS_VIEW_DISPUTE_EVIDENCE_BUTTON_CLICK:
-		'wcpay_payment_details_view_dispute_evidence_button_click',
+		'wcpay_view_submitted_evidence_clicked',
 	SETTINGS_DEPOSITS_MANAGE_IN_STRIPE_CLICK:
 		'wcpay_settings_deposits_manage_in_stripe_click',
 	MULTI_CURRENCY_ENABLED_CURRENCIES_UPDATED:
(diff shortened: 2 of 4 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7402` or branch name `fix/7395` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7395) :rocket:

---

Build info: 

- Latest commit: eacaf41fe37af544e20ff7c264ce37befec2f37f
- Build time: 2023-10-08 09:44:20 UTC

client/tracks/index.js
```diff
@@ -84,7 +84,7 @@ const events = {
 		'wcpay_overview_deposits_change_schedule_click',
 	OVERVIEW_TASK_CLICK: 'wcpay_overview_task_click',
 	PAYMENT_DETAILS_VIEW_DISPUTE_EVIDENCE_BUTTON_CLICK:
-		'wcpay_payment_details_view_dispute_evidence_button_click',
+		'wcpay_view_submitted_evidence_clicked',
```
haszari, line 87: We should rename the constant too. 

BUT I really don't think constants are helping here, I'd prefer to see the actual event names in code. 

So… pragmatic steps:

- Minimum - ensure the constant name 100% matches the event name.
- Better but optional – remove the events you're touching from the constants enum, move them inline.

haszari (approved):
Thanks Bruce! I know that the research underneath this PR is the real value, this small safe change is the tip of an iceberg.

One blocker:

- If we stick with a constant, the constant name must match the event name.

If you find it easier to just remove the constant (use the event name string in code direct – a totally fine thing to do IMO) - then feel free to do that. 

If these event names are already in use, the constants should be removed (is the old code / UI using the constant?). If we're converging I thought we'd be deleting a line or two from the constants array.

client/tracks/index.js
haszari: Why do I care? Constants are an unnecessary layer of indirection between an event name and how it is used. When I'm investigating why a funnel isn't working, where an event is firing, etc, I want to search the code with the event name (short version, without `wcadmin_`) and see all the relevant UI.

shendy-a8c:
> the user intent is the same so keeping the event name and any historical data makes sense here.

Would it make it harder to distinguish between merchants using older version and newer version with new UI if event name is the same? Client version will be incorporated in analyzing which is more complicated IMO.

I feel like we need new event name since it's different UI despite the user intent is the same.

Wait... you are changing the event name.

From your table, it's not consistent with the code diff.

| Button | Transaction Details (current event) | Dispute Details (previous event) |
|--------|--------|--------|
| `Challenge dispute` | `dispute_challenge_click` | `dispute_challenge_clicked` |
| `View submitted evidence` | `payment_details_view_dispute_evidence_button_click` | `view_submitted_evidence_clicked` |

The code diff suggests that you are changing from 'click' to 'clicked', not the other way around as indicated by the above table.

haszari:
> Would it make it harder to distinguish between merchants using older version and newer version with new UI if event name is the same? Client version will be incorporated in analyzing which is more complicated IMO.

I don't think we need to concern ourselves with that. Adding a funnel param for software version is the correct way to handle that. 

> Wait... you are changing the event name.

I was wondering this too, thanks for checking @shendy-a8c ! I would expect that `_click` is more conventional, so was suprised that `_clicked` was the "old" event.

shendy-a8c:
I continue on this PR work:
- Added event prop `on_page: 'transaction_details'`, so we can distinguish event that is using an old name, for example `wcpay_dispute_challenge_clicked`.
- Changed constant name to be consistent with constant value.

shendy-a8c (approved):

shendy-a8c:
Merging for now.

haszari:
Thanks @shendy-a8c ! We can continue reviewing/investigating tracks events and tweak/adjust in future if needed 🙌

client/tracks/index.js
Jinksi: >  remove the events you're touching from the constants enum, move them inline.

This is interesting, I've been doing the opposite! I assumed that the enum was the preferred method to help prevent misspellings.

haszari: That is likely to be the intention behind doing it that way (maybe also "move all strings out into a strings file").

haszari: Another way to look at it: Constants, types etc all make it a little harder to iterate on the event – they assume that we got it perfect upfront and it will never change.

client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
```diff
@@ -337,6 +339,8 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 													{
 														dispute_status:
 															dispute.status,
+														on_page:
+															'transaction_details',
 													}
```
Jinksi, line 344: @brucealdridge Can I get your opinion on whether this event property is necessary for the "dispute accept" button event?

It seems redundant since both accept events are new to this page (`wcpay_dispute_inquiry_refund_click` and `wcpay_dispute_accept_click`).

client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
```diff
@@ -279,6 +280,7 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 										disputeAcceptAction.acceptButtonTracksEvent,
 										{
 											dispute_status: dispute.status,
+											on_page: 'transaction_details',
```
Jinksi, line 283: @brucealdridge Also for this event prop: the modal and "view" events are new to this screen, so this event prop may be redundant.

client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
```diff
@@ -255,9 +255,10 @@ const DisputeAwaitingResponseDetails: React.FC< Props > = ( {
 									onClick={ () => {
 										wcpayTracks.recordEvent(
 											wcpayTracks.events
-												.DISPUTE_CHALLENGE_CLICK,
+												.DISPUTE_CHALLENGE_CLICKED,
 											{
 												dispute_status: dispute.status,
```
Jinksi, line 260: I realise that `dispute_status` is also redundant, since this should only every be `needs_response` (inquiries have a different event, `wcpay_dispute_inquiry_refund_click`)

client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
Jinksi: I've created an issue for this #7474.

client/payment-details/dispute-details/dispute-awaiting-response-details.tsx
Jinksi: I've created an issue for this #7474.

PR #7441, opened 2023-10-09
Prevent deleting wrong order in the duplicate payment prevention service

gpressutto5:
#### Changes proposed in this Pull Request

This PR disables the create and update intent endpoints for deferred intent UPE as discussed in the thread p1696515826158069/1696515464.728129-slack-C06088FAWBS
And ensures the order that is deleted by the the duplicate payment prevention service is pending and not the same as the session order as discussed in the thread p1696576238509689/1696455674.066479-slack-C06088FAWBS

#### Testing instructions

* Smoke test all three UPE variations
* Follow the steps here p1696455674066479-slack-C06088FAWBS

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

Issue #7428, opened 2023-10-06
RPP / Housekeeping / Logger


RadoslavGeorgiev:
### Description

The `WCPay\Logger` class is fully static right now, and impossible to mock. It needs to be converted into a regular class.

### Acceptance criteria

- The logger is managed through the dependency container, and has no static methods.
- To minimize changes, the existing logger can remain in place, but instead of logging directly, it should call `wcpay_get_container()->get( WCPay\Internal\Logger::class )->method_name( $message )`. This will allow mocks for includes when needed.
- The logger retains the same methods as the existing one, but non-static.

### Testing instructions

Trigger logs, and check them in WP admin.

### Dev notes

The logger must be managed through the dependency container, and depends on the `WC_Logger` class. Retrieving it can be done through `wc_get_logger()`. Here's how to manage it in the service provider:

```php
$container->add( 'wc_get_logger', 'wc_get_logger' );
$container->addShared( Logger::class )
	->addArgument( 'wc_get_logger' );
```

### Additional context

Epic: https://github.com/Automattic/woocommerce-payments/issues/6673
Project: paJDYF-9hL-p2

Issue #6712, opened 2023-07-07
Refactor needed for methods `check_payment_intent_attached_to_order_succeeded` and `check_against_session_processing_order`


jessepearson:
### Description

Currently the `check_payment_intent_attached_to_order_succeeded` and `check_against_session_processing_order` of class `Duplicate_Payment_Prevention_Service` 
 return both an array or void|null, which does not line up with the expected results of the method name. The methods also perform other side actions while checking, such as removing the order from the current session and updating the order based on the intent status.

Relevant lines: 

https://github.com/Automattic/woocommerce-payments/blob/38c63a1c7bec594b9f6a51e3d6e1ed15275a6761/includes/class-duplicate-payment-prevention-service.php#L133-L133
https://github.com/Automattic/woocommerce-payments/blob/38c63a1c7bec594b9f6a51e3d6e1ed15275a6761/includes/class-duplicate-payment-prevention-service.php#L81-L81

### Acceptance criteria

* If the method is checking for something, it should return a bool value. The calling code should then be able to progress with that value.
* The side actions should be removed and should be handled in the progression of the calling code.

### Additional context
This was brought up in the discussion here: p1688627901421529-slack-CGGCLBN58

htdat:
@jessepearson - thanks for your follow-up and create this issue. I've updated the title and description of the issue to add another `check_against_session_processing_order` method with the same concern. 

--- 

I agree with the argument that we might refactor these methods to make it clear what they do. However, IMO, we just need to rename those methods from the current prefix `check_` to something more correctly `process_`. The main reason is that we do not want to add more procedure steps into the already huge `payment_process` of gateway classes, and offload those steps into service classes or private methods. 

@RadoslavGeorgiev - what do you think about this issue and my proposal?

vbelolapotkov:
@htdat I'd suggest to keep single responsibility for the method. Instead of renaming, I think there is an opportunity for introducing a new method with suggested name. This method would perform check and other necessary actions to complete processing.

PR #7450, opened 2023-10-11
RPP - Connect the duplicate payment prevention service (Verification)

### LINKED ISSUE #7411: RPP / Verification / Connect the duplicate payment prevention service
## Description

We should connect the new payment process with the `WCPay\Duplicate_Payment_Prevention_Service` ([code](https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-duplicate-payment-prevention-service.php)).

To begin with, the service is relatively simple, and mostly independent. With that in mind, let's **duplicate** it to `src` under a new namespace, `WCPay\Internal\Service\DuplicatePaymentPreventionService`. There are too many small tweaks that need to be done on a relatively small and active piece of functionality, which is why it makes more sense to do it cleanly and independently.

1. Copy the service to `src` and the new namespace.
2. Add the service to `WCPay\Internal\DependencyManagement\ServiceProvider\PaymentsServiceProvider`.
3. Remove the dependency on the gateway. `$gateway->get_return_url()` can be safely replaced with `$order->get_checkout_order_received_url()`. If coded properly, this should not be needed, as the gateway would be the one to generate those URLs.
4. Instead of working directly with order objects, use the `OrderService` class. This will require new methods in the order service, seems like those would be:
    - `get_intent_id( $order_id )` instead of accessing meta directly.
    - `get_cart_hash( $order_id )` to compare orders directly.
    - `is_paid( $order_id )` instead of `$order->has_status( wc_get_is_paid_statuses() )`
    - `add_note( $order_id, $note )`
    - `delete( $order_id )`
5. Instead of using the legacy order service, add a proxy to `update_order_status_from_intent` to the new order service.

Now that the service is suitable for `src`, it will need to be implemented within the new payment process. This involves a few moving pieces:

1. Linking all needed methods like `maybe_update_session_processing_order`, `remove_session_processing_order`, and `clear_session_processing_order_after_landing_order_received_page` in the right place. Those are essentially hooks to allow the service to store/clear the needed data.
2. Detect and handle previously paid orders.
3. Detect and handle successful intents.

### Handling previously paid orders

Use `check_against_session_processing_order` to bail early if a similar order was already paid. This will require creating a new `SimilarOrderDetectedState`, and transitioning to it before any actions are performed. The gateway should know how to handle this state, especially regarding return URLs.

Ideally, the gateway should be able to do something like this:

```php
// Entry into the new payment process.
$state = $state->process();

if ( $state instanceof SimilarOrderDetectedState ) {
	// Consider whether to use `wc_get_order()` directly here.
	$return_url = add_query_arg(
		DuplicatePaymentPreventionService::FLAG_PREVIOUS_ORDER_PAID,
		'yes',
		$this->get_return_url( wc_get_order( $context->get_previous_order_id() ) )
	);

	return [
		'result'   => 'success',
		'redirect' => $return_url,
	];
}

// This is already in place, handling the happy path.
if ( $state instanceof CompletedState ) {
	return [
		'result'   => 'success',
		'redirect' => $this->get_return_url( $order ),
	];
}

throw new Exception( 'Unknown state' );
```

### Handling existing successful intents

Use `check_payment_intent_attached_to_order_succeeded` to use an existing intent, rather than creating a new one. 

This one is a bit more complicated, as it will require splitting `InitialState` into `InitialState` (where verification and preparation are done), a `PreparedState` or `VerifiedState` where intents are created (this one will be skipped), and a `Processed` state. If there is an intent already, it should be loaded from the API, stored in the context, and then we'd transition straight to `ProcessedState` to complete post-processing.

Example:

```php
class InitialState {
	public function process() {
		// ... preparation, other verifications.

		$previous_intent = $this->duplicate_payment_prevention_service->get_intent_attached_to_order_succeeded();
		if ( ! is_null( $previous_intent ) ) {
			$context->set_intent( $previous_intent );
			$next_state = $this->createState( ProcessedState::class );
		}

		// ... Continue with customer creation/update, and intent creation.
		$next_state = $this->createSate( PreparedState::class );
		return $next_state->request_remote_object(); // name TBD.
	}
}
```

## Acceptance criteria

1. The DPPS lives in `src`.
2. Both branches of the duplicate payment prevention service work.
3. Everything fits into the right payment state.
4. 100% unit tests coverage.

## Testing instructions

Same as the standard duplicate payment duplication prevention service.

## Dev notes

Something similar was already implemented in the states PoC: https://github.com/Automattic/woocommerce-payments/pull/6287/files#diff-3bb8d0eedea7f88f28b34f5a291c5bf8c29621f783eadbf0cde7498589569a77

## Additional context

Project: paJDYF-9hL-p2

htdat:
Fixes #7411

- Create new `DuplicatePaymentPreventionService` class in `src` and keep most of logics from the legacy one.
- Hooks: remove the hook from old legacy service, and register hooks from the new `src` service. 
- Introduce only one new state `DuplicateOrderDetectedState` when a duplicate order is detected.
- Introduce `Intent_Status::AUTHORIZED_STATUSES` and `is_authorized()` for intent abstraction.
- Add `SessionService` in `src` with the purpose of easily handling session data (sometimes, it's null while other times, it's really the expected instance of `WC_Session`). This will also be useful when other features will use it, such as Fraud relevant features.
- Add tests, cover 100% for `src` files.

## Testing instructions 

- At this stage, we can only test for: 
    - type of card: basic. 3DS is not yet supported: though #7471 is merged but it's still not possible to finish checkout with a 3DS card - see more details p1698122299758599-slack-C05997J5PCG
    - checkout page type: both classic and block.
    - method: both new and saved method.
    - gateway type: does not matter as we only have one type for the new payment process. 

- Ensure that you're using the latest versions of Dev Tools and enable **New Payment Process Factors**.
- Ideally, add Xdebug breakpoints in the new code in InitialState.php and the new service php file to confirm new code really executed. 

### For testing duplicate orders 

Basically, follow test instruction and result of this PR https://github.com/Automattic/woocommerce-payments/pull/5247 with a note: 
- Use this updated snippet https://gist.github.com/htdat/757b1375146584e63ebad80c5d02f1a1 (rather than the one in PR 5247) 

### For testing duplicate payments 

Basically, follow test instruction and result of this PR https://github.com/Automattic/woocommerce-payments/pull/5346 with a note: 

- For the diff, please use this new one https://gist.github.com/htdat/14715ca445cea3d4d704d5d1a924b25f
- For the snippet, it's OK to use the one mentioned in PR 5346 https://gist.github.com/htdat/6a52eec1be77b8442ac05401908a3627

## Remaining TODO(s) for this PR: 

N/A

## Finished TODOs for this PR

- [x] Consider port this change https://github.com/Automattic/woocommerce-payments/pull/7441 - We may not need this as the fix for PR 7441 is for legacy UPE, and split UPE (not dUPE), which are going to be deprecated. 
- [x] Use `Logger` as dependency per https://github.com/Automattic/woocommerce-payments/issues/7428
- [x] Consider including this refactoring https://github.com/Automattic/woocommerce-payments/issues/6712
- [x] Register hooks but avoid duplicate hooks https://github.com/Automattic/woocommerce-payments/pull/7450#discussion_r1356675850
- [x] Update to use a post-procesing state https://github.com/Automattic/woocommerce-payments/pull/7450#discussion_r1363170224

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7450` or branch name `rpp/7411-duplicate-prevention` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7411-duplicate-prevention) :rocket:

---

Build info: 

- Latest commit: f273a60e92c46d6f04278f2d334c33031f74ff31
- Build time: 2023-10-26 08:58:51 UTC

includes/constants/class-intent-status.php
```diff
@@ -26,4 +26,15 @@ class Intent_Status extends Base_Constant {
 	const REQUIRES_CAPTURE        = 'requires_capture';
 	const CANCELED                = 'canceled';
 	const SUCCEEDED               = 'succeeded';
+
+	/**
+	 * Stripe intents that are treated as successfully created.
+	 *
+	 * @type array
+	 */
+	const SUCCESSFUL_STATUSES = [
```
RadoslavGeorgiev: @htdat now that we have the abstract intention, what are your thoughts on adding an `is_successful()` method to it, and using it to check the status there? By doing so, we would not need this constant, and less code would be repeated.

src/Internal/Service/DuplicatePaymentPreventionService.php
```diff
@@ -0,0 +1,220 @@
...
+	 * @var OrderService
+	 */
+	private $order_service;
+
+	/**
+	 * Initializes all dependencies and hooks, related to the service.
+	 *
+	 * @param OrderService $order_service The order service instance.
+	 */
+	public function init( OrderService $order_service ) {
```
RadoslavGeorgiev: 1. The order service should be provided in the class constructor.
2. `init` should be `init_hooks`, and we can add it in `class WC_Payments` for now.

htdat: 😀 Good point to notice. Anyway, I am still in the progress of working on this one. I will take care of this one.

includes/constants/class-intent-status.php
htdat: This is a good point too. I will implement that one.

htdat: @RadoslavGeorgiev - good idea, however, I think, we might think more about the method name. `is_successful()` can be confusing with "this intent status is `succeeded`". What if we come with something like `is_authorized`? Otherwise, the plural noun `statuses` forces other developers to double check what's they are.

src/Internal/Service/DuplicatePaymentPreventionService.php
```diff
@@ -0,0 +1,237 @@
...
+		$this->order_service = $order_service;
+		$this->hooks_proxy   = $hooks_proxy;
+	}
+
+	/**
+	 * Initializes this class's hooks.
+	 *
+	 * @return void
+	 */
+	public function init_hooks(): void {
```
htdat, line 103: Not yet register this `init_hook` in `WC_Payments` as that can make this action `clear_session_processing_order_after_landing_order_received_page` is registered twice. 

Still pending to think how to approach it.

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: Updated the code but there is a new concern
https://github.com/Automattic/woocommerce-payments/pull/7450/files#r1356675850

includes/constants/class-intent-status.php
RadoslavGeorgiev: Yes, that makes sense to me. 👍

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -808,6 +810,14 @@ public function new_process_payment( WC_Order $order ) {
 		$service = wcpay_get_container()->get( PaymentProcessingService::class );
 		$state   = $service->process_payment( $order->get_id(), $manual_capture );
 
+		if ( $state instanceof PreviousPaidOrderDetectedState ) {
+			return [
+				'result'   => 'success',
+				'redirect' => $this->get_return_url( $order ), // TODO: This should be PaymentContext::get_previous_order_id(). But how to get it as we do expose PaymentContext here at the moment.
```
htdat: @RadoslavGeorgiev - See 15dc9679b5eacf97c8e51f356eef1c50f3237ed7 - this is a tricky part. If we had a storage for PaymentContext, this would be easier.

htdat: Handled here a8c478bc2e1d6c5a3f0df1dc0e5ca0e27eda65f3

htdat: Resolving this now.

includes/constants/class-intent-status.php
htdat: Fixed 2a3d46285c3744ba9440248394da0bcdb32886bb

htdat:
I am going to add tests as well as fix Psalm errors. But it's ready to continue having more reviews. @RadoslavGeorgiev @Automattic/gamma @marcinbot

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -47,6 +51,7 @@ class PaymentsServiceProvider extends AbstractServiceProvider {
 		ExampleService::class,
 		ExampleServiceWithDependencies::class,
 		PaymentRequestService::class,
+		DuplicatePaymentPreventionService::class,
```
RadoslavGeorgiev, line 60: Let's also add the new state here.

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -64,12 +69,21 @@ public function register(): void {
 
 		$container->addShared( PaymentRequestService::class );
 
+		$container->addShared( PaymentRequestService::class );
```
RadoslavGeorgiev: Somehow this is added twice now.

src/Internal/Payment/State/DuplicateOrderDetectedState.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Class PreviousPaidOrderDetectedState
```
RadoslavGeorgiev: Comment from another file.

src/Internal/Payment/State/InitialState.php
```diff
@@ -97,6 +108,21 @@ public function process( PaymentRequest $request ) {
...
 
+		$duplicate_order_id = $this->dpps->get_previous_paid_duplicate_order_id( $order_id );
+		if ( ! is_null( $duplicate_order_id ) ) {
+			$this->dpps->clean_up_when_detecting_duplicate_order( $duplicate_order_id, $order_id );
+			$context->set_duplicate_order_id( $duplicate_order_id );
+			return $this->create_state( DuplicateOrderDetectedState::class );
+		}
+		$this->dpps->update_session_processing_order( $order_id );
+
+		$authorized_intent_id = $this->dpps->get_authorized_payment_intent_attached_to_order( $order_id );
```
RadoslavGeorgiev: We can directly use the intent object here, rather than just the ID.

src/Internal/Payment/State/InitialState.php
```diff
@@ -97,6 +108,21 @@ public function process( PaymentRequest $request ) {
...
+		if ( null !== $authorized_intent_id ) {
+			$this->dpps->remove_session_processing_order( $order_id );
+			$context->set_authorized_intent_id( $authorized_intent_id );
+			return $this->create_state( CompletedState::class ); // TODO. This will be updated to a new post-processing state.
```
RadoslavGeorgiev: Adding a comment for the open ToDo.

src/Internal/Payment/PaymentContext.php
```diff
@@ -225,6 +225,53 @@ public function get_customer_id(): ?string {
...
+	public function set_authorized_intent_id( string $authorized_intent_id ) {
+		$this->set( 'authorized_intent_id', $authorized_intent_id );
+	}
+
+	/**
+	 * Gets the detected authorized intent ID.
+	 *
+	 * @return string|null Payment intent ID.
+	 */
+	public function get_authorized_intent_id(): ?string {
```
RadoslavGeorgiev: Instead of using the intent ID for logic here, let's just use a boolean flag. The ID of the authorized intent will be stored in the `intent` property already. Then this method can be removed.

src/Internal/Service/DuplicatePaymentPreventionService.php
```diff
@@ -0,0 +1,237 @@
...
+
+	/**
+	 * Initializes all dependencies.
+	 *
+	 * @param  OrderService   $order_service  The order service instance.
+	 * @param  SessionService $session_service Woo core Session instance.
+	 * @param  HooksProxy     $hooks_proxy  The hooks proxy instance.
+	 * @param  LegacyProxy    $legacy_proxy  The legacy proxy instance.
+	 */
+	public function __construct( OrderService $order_service, SessionService $session_service, HooksProxy $hooks_proxy, LegacyProxy $legacy_proxy ) {
```
RadoslavGeorgiev: For some reason `SessionService` is not detected by my editor.

src/Internal/Service/OrderService.php
```diff
@@ -254,6 +254,87 @@ public function attach_exchange_info_to_order( int $order_id, WC_Payments_API_Ch
...
+	 * Updates the order status from the given intent object.
+	 *
+	 * @param  int                                $order_id Order ID.
+	 * @param  WC_Payments_API_Abstract_Intention $intent Intent object.
+	 *
+	 * @return void
+	 * @throws Order_Not_Found_Exception
+	 */
+	public function update_order_status_from_intent( int $order_id, WC_Payments_API_Abstract_Intention $intent ): void {
+		$this->legacy_service->update_order_status_from_intent( $this->get_order( $order_id ), $intent );
```
RadoslavGeorgiev: See the `update_order_from_successful_intent` method of this class. Let's at least here, create separate methods for separate outcomes. This one should likely make sure to update all needed meta as well.

src/Internal/Service/OrderService.php
```diff
@@ -254,6 +254,87 @@ public function attach_exchange_info_to_order( int $order_id, WC_Payments_API_Ch
...
+	/**
+	 * Checks if the order has one of paid statuses.
+	 *
+	 * @param  int $order_id  ID of the order.
+	 *
+	 * @return bool True if the order has one of paid statuses, false otherwise.
+	 * @throws Order_Not_Found_Exception
+	 */
+	public function is_paid( int $order_id ): bool {
+		return $this->get_order( $order_id )->has_status( wc_get_is_paid_statuses() );
```
RadoslavGeorgiev: `LegacyProxy`` needed here.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -64,11 +64,11 @@ public function process_payment( int $order_id, bool $manual_capture = false ) {
 		// Start with a basis context.
 		$context = $this->create_payment_context( $order_id, $manual_capture );
 
-		$request         = new PaymentRequest( $this->legacy_proxy );
-		$initial_state   = $this->state_factory->create_state( InitialState::class, $context );
-		$completed_state = $initial_state->process( $request );
+		$request       = new PaymentRequest( $this->legacy_proxy );
+		$initial_state = $this->state_factory->create_state( InitialState::class, $context );
+		$new_state     = $initial_state->process( $request );
```
RadoslavGeorgiev: Warning: @zmaglica's PR already returns `$initial_state->process()`, there will be a conflict here.

RadoslavGeorgiev (commented):
Great job so far, @htdat!

I have a few minor comments, but the general approach looks good. A couple of extra notes:

- I'm curious where `init_hooks` of the DPPS will be called. WDYT about https://github.com/Automattic/woocommerce-payments/pull/7479?
- It's late for me, and I haven't thoroughly checked the DPPS, but I'm confident that if I have comments there, they would be nitpicks.

src/Internal/Service/PaymentProcessingService.php
htdat: Yes, I expect that this will happen, and ready to fix the conflict.

src/Internal/Service/OrderService.php
htdat: Not sure why we also need `LegacyProxy` here while `$this->get_order()` is actually using `LegacyProxy`?

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: May you reload your editor? There is no issue in my PHPStorm.
Markup on 2023-10-18 at 11:33:46

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
htdat: Good point, I missed it somehow. Addressed 9b414720c672ea0fcae13b64da3a2c0166e7efbe

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
htdat: Seems a leftover. Fixed 9b414720c672ea0fcae13b64da3a2c0166e7efbe

src/Internal/Payment/State/DuplicateOrderDetectedState.php
htdat: Good spotting! I changed the state class name in the middle, but did not update this. 
Also fixed here 9b414720c672ea0fcae13b64da3a2c0166e7efbe

src/Internal/Service/OrderService.php
htdat: I am going to remove this method. It was a part of `check_payment_intent_attached_to_order_succeeded` for updating orders after the currenlty attached intent has one of successful ("authorized" is more correct, with our new glossary) statuses.

https://github.com/Automattic/woocommerce-payments/blob/9b414720c672ea0fcae13b64da3a2c0166e7efbe/includes/class-duplicate-payment-prevention-service.php#L117

Eventually, this action would be handed in a new post-processing state, which is currently a TODO https://github.com/Automattic/woocommerce-payments/pull/7450#discussion_r1362622441

htdat: Removed here d09d1e03d7e60e3bfde8a2dc31b14abd4f88799f

src/Internal/Payment/PaymentContext.php
htdat: Agreed that we would use a boolean flag here. 
<strike>But it's not clear which `intent` property you mean here?</strike>

But anway, since the intent ID is attached to the order already, we can retrieve it anytime.

htdat: > But it's not clear which intent property you mean here?

Oh, I've just found out it https://github.com/Automattic/woocommerce-payments/pull/7471/files#diff-35fbbfe7a2e6ba1cab161794c92b9001d634b741162fb3f22106c3f03cfec83cR255-R266. This makes sens now.

_Update_: addressed 9cb3e3a8e524e94342ad50a645871d600ce03e72

src/Internal/Payment/State/InitialState.php
htdat: Together with this discussion, https://github.com/Automattic/woocommerce-payments/pull/7450#discussion_r1363181063, I am switching to your suggestion approach in 9cb3e3a8e524e94342ad50a645871d600ce03e72
Note: I am aware of the potential conflict with `get_intent` and `set_intent` with Zvnonimir's PR.

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: @RadoslavGeorgiev 

> I'm curious where `init_hooks` of the DPPS will be called. WDYT about https://github.com/Automattic/woocommerce-payments/pull/7479 (Hooks Registry)?

I like the concept of `Hooks Registry` for new `src` folder. But it's still not getting out of how to solve potential duplicate hook registers in this specific case.

src/Internal/Service/OrderService.php
htdat: Oh, sorry, I missed `wc_get_is_paid_statuses`. Going to fix it now.

htdat: 01e204ce9d4671655cec53453ddbb069898f58f6 would fix it.

src/Internal/Service/DuplicatePaymentPreventionService.php
RadoslavGeorgiev: All good now, maybe it was some one-time editor bug.

src/Internal/Service/DuplicatePaymentPreventionService.php
RadoslavGeorgiev: Do you see any reason to keep the hooked methods in the legacy service? AFAICT we can remove them, and have a single hook?

htdat: No reason to keep it! If we can guarantee that the hook is always loaded, so I think we're all good. I will dive into this idea, and see how it looks like.

src/Internal/Payment/State/InitialState.php
```diff
@@ -97,6 +108,22 @@ public function process( PaymentRequest $request ) {
...
+			return $this->create_state( DuplicateOrderDetectedState::class );
+		}
+		$this->dpps->update_session_processing_order( $order_id );
+
+		$authorized_intent = $this->dpps->get_authorized_payment_intent_attached_to_order( $order_id );
+		if ( null !== $authorized_intent ) {
+			$this->dpps->remove_session_processing_order( $order_id );
+			$context->set_intent( $authorized_intent );
+			$context->set_detected_authorized_intent();
+			return $this->create_state( CompletedState::class ); // TODO. This will be updated to a new post-processing state.
```
marcinbot: It would be nice for this call to DPPS and the ones below to be wrapped in a single method, for clarity. Not sure how to do it elegantly though. Maybe the new method would return a nullable state? So something like this:
```
private function process_duplicate_payment() : ?AbstractPaymentState {
    // do all the dpps calls and conditionally set everything
    // return null if no duplicate payment
}
```
and then in this function we'd just:
```
$duplicate_payment_result = $this->process_duplicate_payment();
if ( null !== duplicate_payment_result ) {
    return $duplicate_payment_result;
}
```
This would discourage people from adding more nested conditionals and introducing messy code paths.

RadoslavGeorgiev: I wrote everything below only to realize that you are suggesting the same 🤣. I'm leaving it here for posterity.

---

This is not something I have not considered while defining the issue, and for the most part, my thinking is that we should not allow states to be "injected" into the payment process from outside.

At first, when I saw the number of lines within the conditional blocks, it seemed a bit much for me as well, but I still prefer it this way:

```php
$reason_found = $some_service->check_something();
if ( $reason_found ) {
    $some_service->perform_action();
    return $this->createState( SomeState );
}
```

I found that the `maybe_perform_some_action_elsewhere_and_return_result()` way of coding obfuscates what's going on, and makes navigating code more difficult.

I'm willing to meet somewhere in the middle: Isolating this functionality within a method, which looks like what you described, but lives in the state.

htdat: Thanks for the great idea. 
I've moved these lines to private methods, and reduce the number of lines in `InitialState::process` - b75d401fc17a67c47bd04f37ba8696c9a4f023a6

marcinbot: > I'm willing to meet somewhere in the middle: Isolating this functionality within a method, which looks like what you described, but lives in the state.

Yeah, not a huge fan of the `maybe_` methods either. It's an indication they are not pure. In this case I'm just advocating for moving them to a private helper for clarity of the main one. I'm quite happy with Dat's changes.

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: Remove the hook from the legacy service, and register the hook from the new `src` service - 95ed11be21ad5cd011ebb4f4c4c91151a7e92369

I am going to mark this as resolved. Please re-open if needed.

htdat:
Hi @marcinbot  and @Automattic/gamma  👋 PR is official ready to review, after adding the PR description and tests (to cover 100% for new changes in the `src` folder).

htdat:
FYI. I have already merged `develop` (including this PR https://github.com/Automattic/woocommerce-payments/pull/7471), addressed conflict, and used `ProcessedState` for the final todo beafe0c1ae748a8662517e7abc441aa42c9f2236

src/Internal/Service/SessionService.php
```diff
@@ -0,0 +1,80 @@
...
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Service;
+
+use WC_Session;
+use WCPay\Internal\Proxy\LegacyProxy;
+
+/**
+ * Session service. Use as a translation layer between `src` class and Woo core WC_Session class.
```
RadoslavGeorgiev:
```suggestion
 * Session service. Use as a translation layer between `src` classes and Woo core WC_Session class.
```

src/Internal/Service/SessionService.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @param  LegacyProxy $legacy_proxy Legacy proxy instance.
+	 */
+	public function __construct( LegacyProxy $legacy_proxy ) {
+		$this->legacy_proxy = $legacy_proxy;
+	}
+
+	/**
+	 * Getter.
+	 *
+	 * @param  string     $key Session key.
```
RadoslavGeorgiev: RadoCS: 
```suggestion
	 * @param string     $key Session key.
```

src/Internal/Service/SessionService.php
```diff
@@ -0,0 +1,80 @@
...
+		} else {
+			return $default;
+		}
+	}
+
+	/**
+	 * Setter.
+	 *
+	 * @param  string $key Session key.
+	 * @param  mixed  $value Value to set.
```
RadoslavGeorgiev: RadoCS:
```suggestion
	 * @param string $key   Session key.
	 * @param mixed  $value Value to set.
```

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -123,5 +123,4 @@ protected function create_payment_context( int $order_id, bool $manual_capture =
 
 		return $context;
 	}
-
```
RadoslavGeorgiev, line 126: DatCS: 💯

src/Internal/Service/OrderService.php
```diff
@@ -254,6 +254,104 @@ public function attach_exchange_info_to_order( int $order_id, WC_Payments_API_Ch
 		$order->save_meta_data();
 	}
 
+	/**
+	 * Gets currently attached intent ID of the order.
+	 *
+	 * @param  int $order_id Order ID.
```
RadoslavGeorgiev: RadoCS:
```suggestion
	 * @param int $order_id Order ID.
```

src/Internal/Service/OrderService.php
```diff
@@ -254,6 +254,104 @@ public function attach_exchange_info_to_order( int $order_id, WC_Payments_API_Ch
...
+		if ( empty( $intent_id ) ) {
+			return null;
+		}
+		return $intent_id;
+	}
+
+	/**
+	 * Gets cart hash for the given order ID.
+	 *
+	 * @param  int $order_id  ID of the order.
```
RadoslavGeorgiev: RadoCS:
```suggestion
	 * @param int $order_id  ID of the order.
```

Note: There are a few more spaces like this in the file, but I won't add comments for them.

src/Internal/Payment/State/InitialState.php
```diff
@@ -169,4 +192,52 @@ protected function populate_context_from_order() {
 		);
 		$context->set_customer_id( $customer_id );
 	}
+
+	/**
+	 * Detects duplicate orders, and run the necessary actions if one is detected.
+	 *
+	 * @return AbstractPaymentState|null The next state, or null if no duplicate order is detected.
```
RadoslavGeorgiev: I think that for this method, and `process_duplicate_payment`, we can declare the concrete return classes, rather than the abstract. It won't matter much outside of this class, but would still be clearer.

src/Internal/Service/DuplicatePaymentPreventionService.php
```diff
@@ -0,0 +1,260 @@
...
+	 */
+	public function get_authorized_payment_intent_attached_to_order( int $order_id ): ?WC_Payments_API_Payment_Intention {
+		$intent_id = $this->order_service->get_intent_id( $order_id );
+		if ( is_null( $intent_id ) ) {
+			return null;
+		}
+
+		// We only care about payment payment_intent.
+		$is_payment_intent = 'pi_' === substr( $intent_id, 0, 3 );
+		if ( ! $is_payment_intent ) {
```
RadoslavGeorgiev: Nitpick: Could we avoid the variable, and use the condition straight within `if()`?

src/Internal/Service/DuplicatePaymentPreventionService.php
```diff
@@ -0,0 +1,260 @@
...
+			return null;
+		};
+
+		if ( ! $payment_intent->is_authorized() ) {
+			return null;
+		}
+
+		$intent_meta_order_id_raw = $payment_intent->get_metadata()['order_id'] ?? '';
+		$intent_meta_order_id     = is_numeric( $intent_meta_order_id_raw ) ? intval( $intent_meta_order_id_raw ) : 0;
+		if ( $intent_meta_order_id !== $order_id ) {
```
RadoslavGeorgiev, line 143: Question: When would this ever happen? I know it was there before, but I don't remember.

I'm asking, because maybe we should kill the payment process, rather than allowing the code to proceed here?

RadoslavGeorgiev (changes requested):
Great work, @htdat! 👏 

I essentially only have nitpicks about some spaces here and there, but code-wise this looks great. I'm about to start testing, but wanted to share those first.

Question: Should we call `remove_session_processing_order` when transitioning from `ProcessedState` to `CompletedState`?

tests/unit/src/Internal/Service/SessionServiceTest.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Set up the test.
+	 */
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->mock_legacy_proxy = $this->createMock( LegacyProxy::class );
+		$this->sut               = new SessionService( $this->mock_legacy_proxy );
+	}
+
```
RadoslavGeorgiev, line 43: RadoCS: Too many empty spaces throughout this file here.

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
```diff
@@ -0,0 +1,480 @@
...
+		bool $is_session_order_paid,
+
+		int $current_order_id,
+		string $current_order_cart_hash,
+		int $current_order_customer_id,
+		bool $is_current_order_pending,
+
+		?int $expected
+	) {
+		$this->sut = $this->getMockBuilder( DuplicatePaymentPreventionService::class )
```
RadoslavGeorgiev: Let's add an inline comment here, defining the type correctly. Right now I'm getting errors in VSCode that DPPS methods are not defined.

Same applies to every instance of `$this->sut = $this->getMockBuilder` in the file.

```suggestion
		/**
		 * @var DuplicatePaymentPreventionService|MockObject
		 */
		$this->sut = $this->getMockBuilder( DuplicatePaymentPreventionService::class )
```

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
```diff
@@ -0,0 +1,480 @@
...
+
+		$this->deps = [
+			$this->mock_order_service,
+			$this->mock_session_service,
+			$this->mock_logger,
+			$this->mock_hooks_proxy,
+			$this->mock_legacy_proxy,
+		];
+
+		$this->sut = new DuplicatePaymentPreventionService( ... $this->deps );
```
RadoslavGeorgiev, line 92: Cool trick with the dependencies!

RadoslavGeorgiev (changes requested):
A couple of comments regarding tests.

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
htdat: Good call. I've created a separated $mocked_sut for tests using SUT with mocked methods.
621dd122fe1bb87df8a7c5677fe237e915896c08

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: Did it c7de0d1adf9a7ac47dc05ffb44cb7d42d4b2aaa9

src/Internal/Payment/State/InitialState.php
```diff
@@ -169,4 +192,52 @@ protected function populate_context_from_order() {
...
+		if ( null === $authorized_intent ) {
+			return null;
+		}
+
+		$this->dpps->remove_session_processing_order( $order_id );
+		$context->set_intent( $authorized_intent );
+		$context->set_detected_authorized_intent();
+
+		$new_state = $this->create_state( ProcessedState::class );
+		return $new_state->complete_processing();
```
htdat: @RadoslavGeorgiev - I am moving your question to here so it's not getting lost. 

> Question: Should we call `remove_session_processing_order` when transitioning from `ProcessedState` to `CompletedState`?

Yes, before we have already set $order_id when no duplicate order is detected.
https://github.com/Automattic/woocommerce-payments/blob/963541d30f6e78653c57325e52712fc417bdf130/src/Internal/Payment/State/InitialState.php#L210

This is also a bit more defensive even though the `clear_session_processing_order_after_landing_order_received_page` hook would also clear it when it arrives the order-received page.

src/Internal/Payment/State/InitialState.php
htdat: Good point. Addressed 1f232642905d05ac1a9fe51baff6473b5c16451a

src/Internal/Service/DuplicatePaymentPreventionService.php
htdat: I think we do this for: 

1. security 09ffd865c635b25d3ecffc976e8070b1f6431711.
2. dealing with the UPE intent ID session caching bug. For example, an intent ID is initially attached to order A, but then it can be reused and succeeded for order B.

Point (2) seems no longer valid. 

> I'm asking, because maybe we should kill the payment process, rather than allowing the code to proceed here?

If we kill the payment process, how the current processing order can be paid? I think letting it continue here is properly.

htdat:
@RadoslavGeorgiev - thanks for your detailed review. I've addressed all your feedback and resolved most conversations, and keep open for two conversations that you might have other feedback.

src/Internal/Service/DuplicatePaymentPreventionService.php
RadoslavGeorgiev: When I wrote the comment, I was considering clearing the cart, and forcing the customer into a new order. However, if we do that, it will probably make more sense as a general part of the payment process, not something deep within the DPPS. Marking this as resolved.

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
RadoslavGeorgiev: Thank you, but we still need to add doc blocks. VSCode is too smart about it, picking up the concrete type from the method: 

Screenshot 2023-10-26 at 8 22 42

With that in mind, and given the fact that `mock_sut` is not defined within `setUp()`, I'd actually prefer to keep it as a local variable, instead of a property.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Sorry @htdat, could you please re-phrase this reply?  Overall, I'm not sure whether you agree we should do it, or not :D

> Yes, before we have already set $order_id when no duplicate order is detected.

Is that the `$order_id` in the session? Because we would overwrite it shortly.

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
htdat: > but we still need to add doc blocks. 

That's because I missed changing `sut` to `mocked_sut` for this test `test_clean_up_when_detecting_duplicate_order`. I've updated it e688822fc

I think it should be OK now. 

> given the fact that mock_sut is not defined within setUp(), I'd actually prefer to keep it as a local variable, instead of a property.

Compared with the benefit that we do not need to declare the mixed (MockObject and concrete type of the mock) type for every `$this->getMockBuilder()` call, IMO, this is acceptable.

htdat: Also, `$this->mocked_sut` can help distinguish it from `$this->sut`. 

That said, I don't mind much if we change all current `$this->mocked_sut` back to `$this->sut` and declare the type of it. Let me know WDYT :D

RadoslavGeorgiev: > Compared with the benefit that we do not need to declare the mixed (MockObject and concrete type of the mock) type for every $this->getMockBuilder() call, IMO, this is acceptable.

Take another look at the screenshot. I (also) didn't notice that `$this->sut` was used instead of `$this->mock_sut`, but the error is still applicable: VSCode does not care of the PHPDoc type of `$this->var`, and infers the type from the assignment earlier in the method. So we do need to define the type regardless, and with that in mind, the private property becomes useless.

src/Internal/Payment/State/InitialState.php
htdat: @RadoslavGeorgiev - Aha, sorry. My previous reply meant "yes, we should do that" and actually, we're doing that around this code (line 236). It's just that we're not doing that exactly when transitioning from `ProcessedState` to `CompletedState`? Instead, it's in the process of transiting from `InitialState` to `ProcessedState` (line 236 here). 

> Is that the `$order_id` in the session? Because we would overwrite it shortly.

Short answer is yes. 

So the process looks like: 

- Run `process_duplicate_order`. If no duplicate order is detected, return `null` as well as call `update_session_processing_order`. Code ref: 
https://github.com/Automattic/woocommerce-payments/blob/e688822fcac55cfd9bb6e66b947c376c092d355b/src/Internal/Payment/State/InitialState.php#L209-L213

- Run `process_duplicate_payment`. If the duplicate payment is detected, remove the `$order_id`, which has just been set.

tests/unit/src/Internal/Service/DuplicatePaymentPreventionServiceTest.php
htdat: I think which IDE being used does matter here :D 

With the current code (declare the mix type for private property `$this->mocked_sut`), PhpStorm can infer the correct type. 
Markup on 2023-10-26 at 14:33:19

Without any declaration, just calling `$this->getMockBuilder`, PhpStorm can also infer the correct type: 

Markup on 2023-10-26 at 14:37:27

--- 

Anyway, I am going to give it another try with your suggestion ff25309b639e7d465e44bc86e458861020def1af

If it's still not working for VSCode properly, might you submit a commit for it? :D

RadoslavGeorgiev: > I think which IDE being used does matter here :D

It does. Seems like PhpStorm uses a combination of the property DocBlock and the assignment, joining them. In VSCode, the assignment takes priority, and the property is ignored.

Anyway, thank you for this! It works for me now.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: Ok, I think there's a gap here. Let me explain my understanding:

- The payment process starts, and in the normal flow (there is no duplicate order detected), we store the order ID in the session.
- If processing fails, we keep the order in session, and use it to prevent duplicates in the future.
- __If processing succeeds, we remove the order ID from the session.__

The last point is what we do on the Order Received page, but we should also do it at the end of the successful process, like presently do in the gateway: https://github.com/Automattic/woocommerce-payments/blob/ea5370017345fc2254f5011e4fdf381b3842e18a/includes/class-wc-payment-gateway-wcpay.php#L1501

This can only happen _after_ successful processing, but not before it, right? So in addition to what you described (`InitialState` -> `ProcessedState`), we should also clear the order from session when transition from `ProcessedState` to `CompletedState`. Am I missing something?

htdat: That's correct 💯 . It was me to miss this fact when merging Zvonimir's PR that introduces the `ProcessedState`. Addressed here c7a932d0f40be7d205d4c06651b794cef06d58f0

RadoslavGeorgiev (approved):
Excellent work, @htdat!

Thank you for addressing my comments! I went through manual tests as well, and everything worked like a charm.

PR #7463, opened 2023-10-12
feat: WooPay button SSR

frosso:

Adding some SSR to pre-render the WooPay button on non-blocks pages.

This is useful to ensure that the page has less of a jarring effect on first load.
The SSR honors the theme/size of the button, but not its contents. The SSR also loads the button in a `disabled` state.

Before:

After:

#### Testing instructions

- Add the WooPay button to any of the pages (product, cart, checkout)
- Load the shortcode version of the pages
- The WooPay button should be pre-rendered in a `disabled` state
- Theme/Size should be honored during the SSR

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7463` or branch name `feat/woopay-button-pre-render` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=feat/woopay-button-pre-render) :rocket:

---

Build info: 

- Latest commit: 7e0fdb7f85f5162ebfb685ac2593d54dd93b20c7
- Build time: 2023-10-16 13:59:47 UTC

LCmry:
Hi @frosso, just taking a look, but everything worked well for me. I did have a question about why the new WooPay svg?

frosso:
Thanks for taking a look @LCmry ! The WooPay SVG isn't new, it's just an adjustment of the markup
Screenshot 2023-10-13 at 10 45 55 PM

`fillRule` and `clipRule` aren't valid SVG attributes (but they are valid attributes in JSX). It seems that the `assets/images/woopay.svg` has been converted from JSX to an SVG file, but whoever did it didn't correctly convert those two attributes.

LCmry:
Gotcha, thanks for the clarification!

timur27 (approved):
Great job and good move forward, in my opinion. Thanks @frosso! The button renders well in a `disabled` state and code-wise, it's great to see output escaping although the settings are some hardcoded strings and are [provided to the front-end](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-woopay-button-handler.php#L132) later on without escaping, from what I understand. **Below are two side notes** from me: I don't feel they should be blocking your PR, but felt it's worth mentioning these. Please let me know your thoughts and in the meantime, I'm approving! 🚀 

1️⃣ From the user perspective, I feel it would be great to unify this approach across all the express checkout buttons, because at the moment I have both WooPay and GPay enabled, and the jarring effect is still there, due to the lack of SSR for GPay button. This is just a proposal and I understand even if agreed, this might not been handled by this PR.

2️⃣ As a side-note, while testing your PR, I've noticed that the express checkout button placement is unexpected as well as the number of `--OR--` separators on the shortcode page. I thought this misbehavior was fixed some time ago, but maybe there is some regression?


|  Cart           | Shortcode checkout |
--------------- | --------------- | 
| image | image

frosso:
@timur27 thanks for the review and the notes!

> 1️⃣ From the user perspective, I feel it would be great to unify this approach across all the express checkout buttons, because at the moment I have both WooPay and GPay enabled, and the jarring effect is still there, due to the lack of SSR for GPay button. This is just a proposal and I understand even if agreed, this might not been handled by this PR.

That's a great observation, but unfortunately we have a lot less control over the Apple Pay/Google Pay PRBs. The same JS API is used to render both Apple Pay/Google Pay by Stripe, which controls which one of the button that should be rendered based on client side information (e.g.: browser, user location, etc.). Even if we try to duplicate this on the server-side for a more generic approach, an additional problem ensues: the buttons' styles are different - so we'd need to also understand on the server-side which one of the buttons that should be displayed. And, additionally, for the eligibility of these buttons is still reliant on client-side information (see [requirements](https://stripe.com/docs/stripe-js/elements/payment-request-button?client=html#html-js-testing)).

> 2️⃣ As a side-note, while testing your PR, I've noticed that the express checkout button placement is unexpected as well as the number of --OR-- separators on the shortcode page. I thought this misbehavior was fixed some time ago, but maybe there is some regression?

Would you please be able to check if the same is also relevant on the `develop` branch? 👀

timur27:
> unfortunately we have a lot less control over the Apple Pay/Google Pay PRBs

Thanks for the detailed explanation, @frosso! I got it - we keep things as they are. 👍 

> Would you please be able to check if the same is also relevant on the develop branch? 👀

Apologies, I mistakenly assumed I mentioned this, but I didn't. It's totally unrelated to your PR and can be replicated on the `develop`.

frosso:
Thanks for confirming @timur27 ! I'll go ahead and merge this PR and log a separate ticket for the issue.

ricardo:
@frosso Thanks for this. I guess this also improves the Cumulative Layout Shift metric in core web vitals for merchants. 👌

> The SSR honors the theme/size of the button, but not its contents.

Are there any cons to rendering the contents of the button as well, besides having to replicate the logic?

I guess people are more used to seeing the jarring effect rather than a delayed content display without loading indicators, so I fear this might cause some confusion. I noticed the difference in behavior after a `pull`, so I found this PR.

frosso:
> Are there any cons to rendering the contents of the button as well, besides having to replicate the logic?
> I guess people are more used to seeing the jarring effect rather than a delayed content display without loading indicators, so I fear this might cause some confusion. I noticed the difference in behavior after a pull, so I found this PR.

My "fear" was exactly that - trying to not replicate the logic and the maintenance of both. An alternative could be to have a skeleton placeholder component for SSR, so that we don't need to replicate the logic 🤷 but we can indeed monitor whether this causes confusion 👍

pierorocca:
Love the initiative @frosso and for all of the incredible thoughtful feedback. The feedback about the contrast between GPay and Apple Pay buttons when loading at the same time as the WooPay button is of interest. I'll connect with you directly to demo and discuss.

PR #7471, opened 2023-10-12
Add AuthorizationRequired State

### LINKED ISSUE #7418: RPP / Processing / AuthenticationRequired state
### Description

Add the necessary logic during payment processing to generate a new `AuthenticationRequiredState` and handle cases when (3DS) authentication is required.

At first, and for now, the new will be final for the re-engineered payment process. The post-authentication flow will fall back to the legacy payment process for now.

### Acceptance criteria

- A new `WCPay\Internal\Payment\State\AuthenticationRequiredState` exists. This state will be able to resume the process in the future, but does not have any functionality for now.
- The gateway is aware of the new state, and handles it accordingly by getting the intent from the context, and forming the necessary redirect URL and response.
- Even though payment processing is incomplete, we need to store all available details to the order. Make sure the necessary services are used before transitioning to the new state.
- The present method, encrypting client secrets, is the static `WC_Payments_Utils::encrypt_client_secret()`. As a part of this issue, create a standard service within `src` and move all related functionality there. The pre-existing method can return something in the liens of `wcpay_get_container( WCPay\Internal\Service\CheckoutEncryptionService::class )->encrypt_client_secret()`.

### Testing instructions

- Manual testing: Checkout with a 3DS-required card works, and all necessary data is stored within order meta.
- Unit testing: 100% test coverage.

### Additional context

Project: paJDYF-9hL-p2

zmaglica:
Fixes #7418 

This PR adds Authentication required state that handles the situation when payment needs to be authorized (3DS). Among others, this PR introduces multiple state changes.

#### Testing instructions

* Make sure that all tests passed
* Make sure to test with a credit card that requires 3DS (4000000000003063)
* Make sure that you are redirected to the authorization URL.

-------------------

**Post merge**


src/Internal/Payment/PaymentContext.php
```diff
@@ -245,4 +246,22 @@ public function set_customer_id( string $customer_id ) {
 	public function get_customer_id(): ?string {
 		return $this->get( 'customer_id' );
 	}
+
+	/**
+	 * Stores the payment intent.
+	 *
+	 * @param WC_Payments_API_Payment_Intention $intent Instance of intent.
+	 */
+	public function set_intent( WC_Payments_API_Payment_Intention $intent ) {
```
RadoslavGeorgiev: Let's use `WC_Payments_API_Abstract_Intention` for now, to support both payment and setup intents.

src/Internal/Payment/State/InitialState.php
```diff
@@ -46,69 +42,45 @@ class InitialState extends AbstractPaymentState {
...
-		} catch ( Invalid_Request_Parameter_Exception | Extend_Request_Exception | Immutable_Parameter_Exception $e ) {
-			return $this->create_state( SystemErrorState::class );
-		}
-
-		// Intent available, complete processing.
-		$this->order_service->update_order_from_successful_intent( $order_id, $intent, $context );
-
-		// If everything went well, transition to the completed state.
-		return $this->create_state( CompletedState::class );
+		return $this->create_state( VerifyState::class );
```
RadoslavGeorgiev: In this case we're not done with processing, so this should be:
```suggestion
		$next_state = $this->create_state( VerifyState::class );
		return $next_state->process();
```

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+class ProcessedState extends AbstractPaymentState {
+
+	/**
+	 * Order service.
+	 *
+	 * @var OrderService
+	 */
+	private $order_service;
+
```
RadoslavGeorgiev, line 25: There are some weird spacings, let's run `npm run lint:php` and remove them from the PR.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -66,7 +66,7 @@ public function process_payment( int $order_id, bool $manual_capture = false ) {
 
 		$request         = new PaymentRequest( $this->legacy_proxy );
 		$initial_state   = $this->state_factory->create_state( InitialState::class, $context );
-		$completed_state = $initial_state->process( $request );
+		$completed_state = $initial_state->start_processing( $request );
```
RadoslavGeorgiev: Seems like here we can immediately return the result of `start_processing`, avoiding a variable. This way there would be no misleading `completed_state`, which might not be completed yet.

src/Internal/Service/CheckoutEncryptionService.php
```diff
@@ -0,0 +1,44 @@
...
+		/**
+		 * Encrypt client secret for the client.
+		 *
+		 * @param string $customer_id Customer id.
+		 * @param string $client_secret Client secret.
+		 *
+		 * @return string
+		 */
+	public function encrypt_client_secret( string $customer_id, string $client_secret ): string {
+		if ( \WC_Payments_Features::is_client_secret_encryption_enabled() ) {
```
RadoslavGeorgiev: Please use the `LegacyProxy`'s `call_static` here.

src/Internal/Service/CheckoutEncryptionService.php
```diff
@@ -0,0 +1,44 @@
...
+			return openssl_encrypt(
```
RadoslavGeorgiev: Same here for `openssl_encrypt`, but with `LegacyProxy`'s `call_function()`.

src/Internal/Payment/State/VerifyState.php
```diff
@@ -0,0 +1,85 @@
...
+use WCPay\Core\Exceptions\Server\Request\Extend_Request_Exception;
+use WCPay\Core\Exceptions\Server\Request\Immutable_Parameter_Exception;
+use WCPay\Core\Exceptions\Server\Request\Invalid_Request_Parameter_Exception;
+use WCPay\Internal\Service\OrderService;
+use WCPay\Internal\Service\PaymentRequestService;
+
+/**
+ * This state is used to create payment intent and verify next actions based on the intent state.
+ */
+class VerifyState extends AbstractPaymentState {
```
RadoslavGeorgiev: Let's make this `VerifiedState`, as verification is already done.

src/Internal/Payment/State/VerifyState.php
```diff
@@ -0,0 +1,85 @@
...
+	 */
+	public function process() {
+		$context = $this->get_context();
+
+		// Payments are currently based on intents, request one from the API.
+		try {
+			$intent = $this->payment_request_service->create_intent( $context );
+
+			// Something went wrong since the intent is not an instance of payment intent class. Transition to error state.
+			if ( ! $intent instanceof WC_Payments_API_Payment_Intention ) {
```
RadoslavGeorgiev: `create_intent` should either return an intent, or throw an exception. This part should not be needed.

src/Internal/Payment/State/VerifyState.php
```diff
@@ -0,0 +1,85 @@
...
+			return $this->create_state( SystemErrorState::class );
+		}
+
+		// Intent requires authorization (3DS check).
+		if ( Payment_Intent_Status::REQUIRES_ACTION === $intent->get_status() ) {
+			return $this->create_state( AuthenticationRequiredState::class );
+		}
+
+		// All good. Proceed to processed state.
+		return $this->create_state( ProcessedState::class );
```
RadoslavGeorgiev: Now that it is processed, we can immediately try to complete it.
```suggestion
		$next_state = $this->create_state( ProcessedState::class );
		return $next_state->complete();
```

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	/**
+	 * Process all needed verifications.
+	 *
+	 * @return AbstractPaymentState
+	 * @throws \WCPay\Exceptions\Order_Not_Found_Exception
+	 * @throws \WCPay\Internal\Payment\Exception\StateTransitionException
+	 */
+	public function complete() {
```
RadoslavGeorgiev: Let's make sure that methods like this one are defined in the base class as well.

RadoslavGeorgiev (changes requested):
Great job so far, here's my initial feedback:

src/Internal/Payment/State/AuthenticationRequiredState.php
```diff
@@ -0,0 +1,75 @@
...
+		CheckoutEncryptionService $checkout_encryption_service
+	) {
+		parent::__construct( $state_factory );
+
+		$this->checkout_encryption_service = $checkout_encryption_service;
+	}
+	/**
+	 * Get response that will be sent to the client.
+	 */
+	public function get_response(): array {
```
RadoslavGeorgiev: This should probably not live in the state, as the redirect (especially 3DS) is very gateway and JS-specific. On the other hand, it is just a string, so maybe it makes sense. I'd like @marcinbot 's opinion here as well.

marcinbot: Yeah, I think it would be better if this was moved to the service/gateway which would take care of everything. Maybe we could just have a helper method to `get_intent`. The main reason is, as you mentioned, abstracting the provider-specific code.

One the topic of abstraction, the use of `intent` in the context is worrying (and there are a few other provider-specific things), but we can figure it out later, once we have something more concrete in place in terms of abstraction.

zmaglica: Good point. Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/044cdf8890f23a366b897fef839b1c28c29bf06a

src/Internal/Payment/State/ProcessedState.php
zmaglica: Adressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/427a6c36a2fbb2bf7f368a10daa4646f4120d3c4

src/Internal/Payment/State/VerifyState.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/582352a1000f75a7d1f6c5b43c1655598557b4d1

src/Internal/Payment/State/VerifyState.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/684b2ad638ebdad6136a352689050e87836c3dd6

src/Internal/Payment/State/VerifyState.php
zmaglica: Good point. Refactored here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/585514936cd51dbe088123bba8615412d3a3272b

src/Internal/Service/CheckoutEncryptionService.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/7296e3be6acb0360e89a3e6d5be2f16868847c0b

src/Internal/Service/CheckoutEncryptionService.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/7296e3be6acb0360e89a3e6d5be2f16868847c0b

src/Internal/Service/PaymentProcessingService.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/ce20eaf4b3aa50f5d0a85af0d1e2d988cc09df10

src/Internal/Payment/State/ProcessedState.php
zmaglica: npm run lint:php didn't linted the files, which is strange, so I manually linted each of these. Commit can be found here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/96812499ac286f9767224082ef325de6e6ff015e

src/Internal/Payment/State/InitialState.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/fa1721e9d3fbd25e4c5b63fb88d5b21bf6d8e1f6

src/Internal/Payment/PaymentContext.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/018d6303d01588e7b589c7af01d6d97303d5dee7

zmaglica:
> Great job so far, here's my initial feedback:

Thank you for the review. It really helped a lot. I have addressed all comments. Please let me know if there is anything to change or need additional feedback.

src/Internal/Payment/State/AuthenticationRequiredState.php
```diff
@@ -0,0 +1,73 @@
...
+		$this->legacy_proxy                = $legacy_proxy;
+		$this->checkout_encryption_service = $checkout_encryption_service;
+	}
+
+	/**
+	 * Get redirect url.
+	 *
+	 * @return string
+	 */
+	public function get_redirect_url() {
```
RadoslavGeorgiev: I'm wondering whether `get_redirect_url()` is the correct name to use here, considering this method stays at all. On one hand, `get_authentication_url` would be more specific to what it does, on the other hand, we could utilize states and the fact that `get_redirect_url()` would be specific to this state.

No need to change anything here immediately, I'm just wondering.

src/Internal/Payment/State/AuthenticationRequiredState.php
```diff
@@ -0,0 +1,73 @@
...
+		if ( isset( $next_action['type'] ) && 'redirect_to_url' === $next_action['type'] && ! empty( $next_action['redirect_to_url']['url'] ) ) {
+			return $next_action['redirect_to_url']['url'];
+		}
+
+		return sprintf(
+			'#wcpay-confirm-%s:%s:%s:%s',
+			$context->get_amount() > 0 ? 'pi' : 'si',
+			$context->get_order_id(),
+			$this->checkout_encryption_service->encrypt_client_secret( $intent->get_customer_id(), $intent->get_client_secret() ),
+			$this->legacy_proxy->call_function( 'wp_create_nonce', 'wcpay_update_order_status_nonce' )
```
RadoslavGeorgiev: There is another thread (https://github.com/Automattic/woocommerce-payments/pull/7471/#pullrequestreview-1674442396 is the link, but GH is removing the hash/deep link) where I mentioned that some of this is too specific to live within the state. What I mean:

1. The `'#wcpay-confirm-%s:%s:%s:%s'` format is tightly tied to our JS implementation. If we wanted to change it, we'd have some very weird `if()`s deep within the payment process. There should be another layer to handle those. I'm not saying it's the gateway class, but this is probably not the right place.
2. Neither the decryption of the client secret, nor the verification of this nonce will be performed within states, but through a controller that will trigger the state machine. As it will be the job of that controller to verify the nonce, its generation should be close to the verification. Presently, that controller is the gateway class, so we should probably keep this there. Once we add 3DS post-verification to the new payment process, we'll create the proper controller, and that code should fit well there.

With that said, this is a lot of Stripe-specific code to put in the gateway. What do you think that creating the new controller(-like) class now, and putting this code there? Then in the gateway we'd be able to use the service, and we'll be able to use it later as well:

```php
public function new_process_payment( WC_Order $order ) {
	// ...
	$state = $service->process_payment( $order->get_id(), $manual_capture );

	if ( $state instanceof AuthenticationRequiredState ) {
		$redirect = $controller->generate_authentication_url( $state->get_context()->get_intent(), $order->get_id() );

		return [
			'result'   => 'success',
			'redirect' => $redirect,
		];
	}
}
```

Note: The amount wouldn't be needed because we can check whether the intent is a setup or payment intent.

src/Internal/Service/CheckoutEncryptionService.php
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * Encrypt client secret for the client.
+	 *
+	 * @param string $customer_id Customer id.
+	 * @param string $client_secret Client secret.
+	 *
+	 * @return string
+	 */
+	public function encrypt_client_secret( string $customer_id, string $client_secret ): string {
```
RadoslavGeorgiev: If we go with the suggestion of creating a controller for managing 3DS verification URLs, maybe we could avoid having a whole service for this, and have `encrypy_client_secret` there?

RadoslavGeorgiev (changes requested):
I like the direction of this PR, great job adjusting it to the concept! 👏 

There are only a couple of topics I'd like to run by you in comments. In the meantime, feel free to proceed with the PR:

1. Basic checkout still works, but it would be nice to see the new `AuthenticationRequiredState` handled as well.
2. Psalm has a few errors, let's address them, and make sure that there are no left-over `@throws` in transition methods, especially within `InitialState`.
3. There are various unused `use` statements. Let's tidy up the code by removing them.

It might make sense to address as much as possible before starting to fix existing tests, and adding new ones, to avoid having to change fresh tests.

src/Internal/Payment/State/AuthenticationRequiredState.php
zmaglica: I agree with your perspective regarding the stripe related code. Relocating this code might be beneficial in the long run.

Moreover, controllers are mainly designed to receive request data, process it by using services for example, and deliver (return) results. Therefore, I suggest moving this code to a service. Since the `CompletedState` and `ErrorState` also need to return responses, it makes sense to me to move them into the same service. This way, we can manage the business logic more effectively and in a centralized way.

~~The ideal candidate, IMHO, could be `PaymentProcessingService` where we already have original state processing and parsing results at the end of the cycle could be a nice way to handle it here.~~

src/Internal/Payment/State/AuthenticationRequiredState.php
zmaglica: Make sense. I addressed it here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/e2189419b383bb031404e3c4f906b26e169caa45

src/Internal/Service/CheckoutEncryptionService.php
zmaglica: Good catch. We don't need that service. Removed it here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/8f1ebd8f2f87833dcccb8539e183ed9e164d3218

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -32,6 +32,8 @@
 use WCPay\Duplicate_Payment_Prevention_Service;
 use WCPay\Fraud_Prevention\Fraud_Prevention_Service;
 use WCPay\Fraud_Prevention\Fraud_Risk_Tools;
+use WCPay\Internal\Payment\Response\AuthenticationRequiredResponse;
```
RadoslavGeorgiev: This line is now not used.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -826,6 +828,15 @@ public function new_process_payment( WC_Order $order ) {
 			];
 		}
 
+		if ( $state instanceof AuthenticationRequiredState ) {
+			$context = $state->get_context();
+			// Maybe we can move this inside the response class and call function like generate_response() that will something like this where we have more control of status code, ...
```
RadoslavGeorgiev: Is this line still needed?

src/Internal/Payment/State/InitialState.php
```diff
@@ -46,69 +42,53 @@ class InitialState extends AbstractPaymentState {
...
-		// Payments are currently based on intents, request one from the API.
-		try {
-			$intent = $this->payment_request_service->create_intent( $context );
-		} catch ( Invalid_Request_Parameter_Exception | Extend_Request_Exception | Immutable_Parameter_Exception $e ) {
-			return $this->create_state( SystemErrorState::class );
-		}
+		/**
+		 * Next state variable.
+		 *
+		 * @var VerifiedState $next_state
```
RadoslavGeorgiev: Instead of defining this here, please make sure that all transition methods (`start_processing`, `process`, and `complete`) are present in the abstract state.

src/Internal/Payment/State/InitialState.php
```diff
@@ -13,11 +13,7 @@
 use WCPay\Internal\Service\OrderService;
 use WCPay\Internal\Service\Level3Service;
 use WCPay\Internal\Service\PaymentRequestService;
```
RadoslavGeorgiev: This is an unused dependency, let's remove it.

src/Internal/Payment/State/VerifiedState.php
```diff
@@ -0,0 +1,84 @@
...
+ * This state is used to create payment intent and verify next actions based on the intent state.
+ */
+class VerifiedState extends AbstractPaymentState {
+
+	/**
+	 * Order service.
+	 *
+	 * @var OrderService
+	 */
+	private $order_service;
```
RadoslavGeorgiev: Let's remove the dependency on the order service from this state. Everything, needed from the order service, is already loaded in `InitialState` or completed in `ProcessedState`.

src/Internal/Payment/State/VerifiedState.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Class VerifiedState
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Payment\State;
+
+use WC_Payments_API_Abstract_Intention;
```
RadoslavGeorgiev: This line is not needed anymore.

src/Internal/Payment/State/VerifiedState.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->order_service           = $order_service;
+		$this->payment_request_service = $payment_request_service;
+	}
+
+	/**
+	 * Process all needed verifications.
+	 *
+	 * @return AbstractPaymentState
+	 *
+	 * @throws \WCPay\Exceptions\Order_Not_Found_Exception
```
RadoslavGeorgiev: Why do we need to refer back to the global namespace instead of adding a `use` statement?

src/Internal/Payment/State/VerifiedState.php
```diff
@@ -0,0 +1,84 @@
...
+		// Payments are currently based on intents, request one from the API.
+		try {
+			$intent = $this->payment_request_service->create_intent( $context );
+			$context->set_intent( $intent );
+		} catch ( Invalid_Request_Parameter_Exception | Extend_Request_Exception | Immutable_Parameter_Exception $e ) {
+			return $this->create_state( SystemErrorState::class );
+		}
+
+		// Intent requires authorization (3DS check).
+		if ( Payment_Intent_Status::REQUIRES_ACTION === $intent->get_status() ) {
```
RadoslavGeorgiev: `Payment_Intent_Status` is now deprecated in favor of `Intent_Status`.

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Class ProcessedState
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Payment\State;
+
+use WC_Payments_API_Payment_Intention;
```
RadoslavGeorgiev: There are a bunch of left-over `use` statements, let's remove them from here.

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,61 @@
...
+use WCPay\Core\Exceptions\Server\Request\Immutable_Parameter_Exception;
+use WCPay\Core\Exceptions\Server\Request\Invalid_Request_Parameter_Exception;
+use WCPay\Internal\Service\OrderService;
+use WCPay\Internal\Service\PaymentRequestService;
+
+/**
+ * This state is used when payment is completed on the server, and we need to update date on the plugin side.
+ */
+class ProcessedState extends AbstractPaymentState {
+
```
RadoslavGeorgiev: Let's not add empty lines in the beginning of code blocks, they only make sense as separators between blocks of  code.

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Order service.
+	 *
+	 * @var OrderService
+	 */
+	private $order_service;
+
+
```
RadoslavGeorgiev: There are still two empty lines here instead of one.

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,61 @@
...
+	 * Process all needed verifications.
+	 *
+	 * @return AbstractPaymentState
+	 * @throws \WCPay\Exceptions\Order_Not_Found_Exception
+	 * @throws \WCPay\Internal\Payment\Exception\StateTransitionException
+	 * @throws \WCPay\Vendor\League\Container\Exception\ContainerException
+	 */
+	public function complete() {
+		$context = $this->get_context();
+		// Complete processing.
```
RadoslavGeorgiev, line 52: One of those empty lines can be used here :p

src/Internal/Payment/State/AbstractPaymentState.php
```diff
@@ -91,17 +91,42 @@ protected function create_state( string $state_class ) {
...
+
+	/**
+	 * Get redirect url.
+	 *
+	 * @psalm-suppress InvalidReturnType
+	 *
+	 * @return string
+	 * @throws \WCPay\Internal\Payment\Exception\StateTransitionException
+	 */
+	public function get_authentication_url() {
```
RadoslavGeorgiev: This method is not used anymore.

src/Internal/Payment/State/AuthenticationRequiredState.php
```diff
@@ -0,0 +1,15 @@
...
+ * Class AuthenticationRequiredState
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Payment\State;
+
+use WCPay\Internal\Proxy\LegacyProxy;
+/**
+ * The state, which indicates that the payment processing has been completed.
```
RadoslavGeorgiev: This comment needs to be updated for the given state.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -7,7 +7,10 @@
 
 namespace WCPay\Internal\Service;
 
-use Exception; // Temporary exception! This service would have its own exception when more business logics are added.
+use Exception;
```
RadoslavGeorgiev: I think that after re-arranging things, nothing here throws a simple `Exception`. Could you please verify if that's the case, and remove it?

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -64,23 +67,62 @@ public function process_payment( int $order_id, bool $manual_capture = false ) {
...
+				'aes-128-cbc',
+				substr( $intent->get_customer_id(), 5 ),
+				0,
+				str_repeat( 'WC', 8 )
+			);
+		}
+
+		return sprintf(
+			'#wcpay-confirm-%s:%s:%s:%s',
+			substr( $intent->get_id(), 0, 2 ), // intents starts with pi_ or si_ so we need only to first two letters.
```
RadoslavGeorgiev: Instead of string operations here, could we use the intent class?
```suggestion
			$intent instanceof WC_Payments_API_Setup_Intention ? 'si' : 'pi',
```

RadoslavGeorgiev (changes requested):
We're getting closer! 🙇

src/Internal/Payment/State/VerifiedState.php
```diff
@@ -0,0 +1,72 @@
...
+use WCPay\Constants\Payment_Intent_Status;
+use WCPay\Core\Exceptions\Server\Request\Extend_Request_Exception;
+use WCPay\Core\Exceptions\Server\Request\Immutable_Parameter_Exception;
+use WCPay\Core\Exceptions\Server\Request\Invalid_Request_Parameter_Exception;
+use WCPay\Internal\Service\PaymentRequestService;
+
+/**
+ * This state is used to create payment intent and verify next actions based on the intent state.
+ */
+class VerifiedState extends AbstractPaymentState {
```
marcinbot: Perhaps we should aim to follow the new glossary.

So the `AuthenticationRequired` would become `PendingAuthentication`.

It would then transition to `PendingCapture` rather than Verified. If autocapture is enabled, then it would transition straight to `Captured`.

There's no complete state, because the payment is never really "complete". The funds might end up in the merchant's bank, but the payment can be refunded later, for example.

marcinbot: also cc @RadoslavGeorgiev

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: Let's remove the comment as well :)

src/Internal/Payment/State/VerifiedState.php
RadoslavGeorgiev: @marcinbot should we follow the naming convention from pc2DNy-3e9-p2#payment-charges?

Otherwise, your comment makes sense, and the earlier we apply the new naming, the less we'd need to change.

> It would then transition to `PendingCapture` rather than `Verified`.

Actually, `VerifiedState` is when we've prepared all the data, and verified that we can continue with processing, or creating a charge/order. At this point, we don't have a charge. I see why we would want to name it a different way, but I'm not sure what would be more suitable.

marcinbot: > should we follow the naming convention from pc2DNy-3e9-p2#payment-charges?

Yes

> Actually, VerifiedState is when we've prepared all the data, and verified that we can continue with processing, or creating a charge/order. At this point, we don't have a charge. I see why we would want to name it a different way, but I'm not sure what would be more suitable.

Ah, I see what you mean. But it feels like a redundant step a little bit. Most state transitions will probably require some sort of data verification as they progress through the process, so I think it might make sense to roll Verified back to Initial.

Alternatively, we could use something like `VerifiedInitialState` but again, I think we create a precedent for more redundant states like that, which would be difficult to manage.

RadoslavGeorgiev: > But it feels like a redundant step a little bit. Most state transitions will probably require some sort of data verification as they progress through the process, so I think it might make sense to roll Verified back to Initial.

The state is not redundant. See https://github.com/Automattic/woocommerce-payments/issues/7411 and specifically "Handling existing successful intents". This is because of the way the duplicate payment prevention service works now, and needs to work, at least until we implement storage for the payment context. If there was an existing intent, detected during the verification stage, payment processing needs to be skipped, and we'd transition straight to the `ProcessedState`.

Now that I'm writing this though, I realize that we can simply do this within `InitialState`:

```php
class InitialState {
	function start_processing() {
		// prepare data...
		
		// verification
		$existing_intent = $this->duplicate_payment_prevention_service->check_for_existing_intent();
		if ( $existing_intent ) {
			$context->set_intent( $existing_intent );
			$next_state->complete();
		}

		// normal processing (what we have in VerifiedState now)
		$intent = $this->payment_request_service->create_intent( $context );
		$context->set_intent( $intent );

		// Intent requires authorization (3DS check).
		if ( Payment_Intent_Status::REQUIRES_ACTION === $intent->get_status() ) {
			return $this->create_state( AuthenticationRequiredState::class );
		}

		// All good. Proceed to processed state.
	}
}
```

@zmaglica splitting `InitialState` into `InitialState` and `VerifiedState` was not a part of the issue definition for this PR. Why did you go ahead and implement it?

zmaglica: > @zmaglica splitting `InitialState` into `InitialState` and `VerifiedState` was not a part of the issue definition for this PR. Why did you go ahead and implement it?

The reason for splitting lies in the introduction of new states. For more details about this decision, please refer to the conversation at: p1697092433168499/1696947391.111769-slack-C05997J5PCG. As you follow the conversation, you will notice that these new states have been discussed/added, and they need to be defined before we can proceed with defining the `AuthenticationRequiredState`. The only state that is not part of the authentication flow, which is part of this PR, is the `ProcessedState`, but I didn't want to leave the processing flow half-done.

I can easily edit the PR and remove the `VerifiedState`.

marcinbot: Ok, the duplicate payment prevention was the missing piece of the puzzle for me. Let's leave the Verified State in for now, but rename other states.

RadoslavGeorgiev: > For more details about this decision, please refer to the conversation at: p1697092433168499/1696947391.111769-slack-C05997J5PCG. As you follow the conversation, you will notice that these new states have been discussed/added, and they need to be defined before we can proceed with defining the `AuthenticationRequiredState`.

I'm too used to having the verified state, and I mistakenly used as a part of the example in this message (p1697092433168499/1696947391.111769-slack-C05997J5PCG). However, it seems like we do not need the state for now, so let's merge it back with `InitialState`, please. @htdat can re-add it in https://github.com/Automattic/woocommerce-payments/pull/7450 if needed, but that will probably not be the case. Sorry for triggering this change, and then requesting it to be reverted.

> the only state that is not part of the authentication flow, which is part of this PR, is the ProcessedState, but I didn't want to leave the processing flow half-done.

There will be more states, as we proceed with the payment process. I understand the temptation to create them all in advance, but that's not needed yet. Let's leave the `ProcessedState`, but try to merge this PR soon, to allow @htdat to use it within the in-progress PR.

--- 

Regarding @marcinbot's request, lets rename:

- `AuthenticationRequiredState` to `PendingAuthenticationState`
- `CompletedState` to `CapturedState`
- Introduce a new (and empty for now) `PendingCaptureState` state. Based on the outcome of the intent (captured or not), `ProcessedState` should transition to either this state, or `CapturedState`.

@marcinbot how would we call the transition? `complete_processing()` instead of `complete()`?

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -132,52 +105,68 @@ public function test_process( Exception $exception = null ) {
...
+					$verified_state = new VerifiedState( $this->mock_state_factory, $mock_payment_request_service );
+					$verified_state->set_context( $this->mock_context );
+
+					return $verified_state;
+				} )(),
+				( function () {
+					$this->mock_order_service->expects( $this->once() )
+						->method( 'update_order_from_successful_intent' )
+						->with( $this->mock_context->get_order_id(), $this->mock_context->get_intent(), $this->mock_context );
+					$processed_state = new ProcessedState( $this->mock_state_factory, $this->mock_order_service );
```
RadoslavGeorgiev: This is not testing the `InitialState`, it's testing the whole process, which is not needed, and over-complicates this test. Let's just return a mock state from the first `create_state` call, assert that the right method is called, and be done with it. Otherwise, once we add all the needed logic to the next state, this test will grow out of control.

src/Internal/Payment/State/VerifiedState.php
marcinbot: There should be `ProcessedState` or a concept of `complete`. Sounds like this is the state machine we need, based on the various discussions:
```mermaid
stateDiagram-v2
    [*] --> InitialState
    InitialState --> PendingAuthentication : process()
    InitialState --> PendingAuthorization : process()
    PendingAuthentication --> PendingAuthorization : authenticate()
    PendingAuthorization --> PendingCapture : ["automatic" transition]
    PendingAuthorization --> Captured : ["automatic" transition]
    PendingCapture --> Captured : capture()
```

zmaglica: I have removed `VerifiedState` here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/7765bee4c4e6a445879d0019d080de8be2660c52

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/776c96fa45b2f0da36dd2483b2d60a645bf04abf

src/Internal/Payment/State/InitialState.php
```diff
@@ -56,10 +56,10 @@ class InitialState extends AbstractPaymentState {
...
 	 * Class constructor, only meant for storing dependencies.
 	 *
-	 * @param StateFactory                 $state_factory           Factory for payment states.
-	 * @param OrderService                 $order_service           Service for order-related actions.
-	 * @param WC_Payments_Customer_Service $customer_service        Service for managing remote customers.
-	 * @param Level3Service                $level3_service          Service for Level3 Data.
+	 * @param StateFactory                 $state_factory Factory for payment states.
+	 * @param OrderService                 $order_service Service for order-related actions.
+	 * @param WC_Payments_Customer_Service $customer_service Service for managing remote customers.
+	 * @param Level3Service                $level3_service Service for Level3 Data.
```
RadoslavGeorgiev: Let's restore the right spacing here.

src/Internal/Payment/State/InitialState.php
```diff
@@ -78,37 +78,42 @@ public function __construct(
...
-		$order_id = $context->get_order_id();
-
+	public function start_processing( PaymentRequest $request ) {
 		// Populate basic details from the request.
 		$this->populate_context_from_request( $request );
 
 		// Populate further details from the order.
 		$this->populate_context_from_order();
 
+		$context = $this->get_context();
```
RadoslavGeorgiev: Now that the context is only used for creating the intent, let's move this line within the `try {}` block.

src/Internal/Payment/State/InitialState.php
```diff
@@ -78,37 +78,42 @@ public function __construct(
...
+		if ( Intent_Status::REQUIRES_ACTION === $intent->get_status() ) {
+			return $this->create_state( AuthenticationRequiredState::class );
+		}
+
+		// All good. Proceed to processed state.
+		$next_state = $this->create_state( ProcessedState::class );
 
-		// If everything went well, transition to the completed state.
-		return $this->create_state( CompletedState::class );
+		return $next_state->complete();
```
RadoslavGeorgiev: Based on conversations so far, wdyt about calling this `complete_processing`?

RadoslavGeorgiev (changes requested):
I still need to check tests, but the rest of the changes look pretty well. I only have a couple of nitpicks.

 know we are also discussing renaming states and etc., but that should be done through another issue & PR. This one is already pretty big.

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -14,11 +14,14 @@
...
 use WCPay\Internal\DependencyManagement\AbstractServiceProvider;
 use WCPay\Internal\Payment\Router;
+use WCPay\Internal\Payment\State\AuthenticationRequiredState;
 use WCPay\Internal\Payment\State\CompletedState;
 use WCPay\Internal\Payment\State\InitialState;
 use WCPay\Internal\Payment\State\PaymentErrorState;
+use WCPay\Internal\Payment\State\ProcessedState;
 use WCPay\Internal\Payment\State\StateFactory;
 use WCPay\Internal\Payment\State\SystemErrorState;
+use WCPay\Internal\Payment\State\VerifiedState;
```
RadoslavGeorgiev: Small left-over here :)

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -41,6 +44,7 @@ class PaymentsServiceProvider extends AbstractServiceProvider {
 		Router::class,
 		StateFactory::class,
 		InitialState::class,
+		ProcessedState::class,
```
RadoslavGeorgiev, line 47: Just like this declaration, we need to declare that the provider provides `AuthenticationRequiredState`. Right now, everything works, because the `register` method is called for another state. At some point in the future, if someone requests the `AuthenticationRequiredState` directly, this could lead to a fatal error.

includes/class-wc-payment-gateway-wcpay.php
RadoslavGeorgiev: Somehow `AuthenticationRequiredResponse` is back :D

src/Internal/Payment/State/AbstractPaymentState.php
```diff
@@ -91,17 +91,43 @@ protected function create_state( string $state_class ) {
...
+	/**
+	 * Process verified state.
+	 *
+	 * @psalm-suppress InvalidReturnType
+	 *
+	 * @return AbstractPaymentState
+	 * @throws \WCPay\Exceptions\Order_Not_Found_Exception
+	 * @throws \WCPay\Internal\Payment\Exception\StateTransitionException
+	 */
+	public function process() {
```
RadoslavGeorgiev: This method is not used anywhere, let's only leave `start_processing` here.

src/Internal/Payment/State/AbstractPaymentState.php
```diff
@@ -91,17 +91,43 @@ protected function create_state( string $state_class ) {
...
+	}
+
+	/**
+	 * Complete processing.
+	 *
+	 * @psalm-suppress InvalidReturnType
+	 *
+	 * @return AbstractPaymentState
+	 * @throws \WCPay\Exceptions\Order_Not_Found_Exception
+	 * @throws \WCPay\Internal\Payment\Exception\StateTransitionException
```
RadoslavGeorgiev: Let's be consistent, and add `use` statements for those, only listing exceptions like `Order_Not_Found_Exception` or `StateTransitionException`, keeping in mind that those are already "imported" into the file. This applies to the whole PR.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: You fixed it for `start_processing`, but it should be consistent for all files within the PR (and `src` for that matter).

src/Internal/Payment/State/InitialState.php
```diff
@@ -120,6 +124,7 @@ public function process( PaymentRequest $request ) {
 	 * on all needed parameters being in place.
 	 *
 	 * @param PaymentRequest $request The request to use.
+	 *
```
RadoslavGeorgiev: More RadoCS: We don't need an empty line between parameters and exceptions. Same for the whole PR (if there are other instances).

tests/unit/src/Internal/Payment/State/AbstractPaymentStateTest.php
```diff
@@ -80,9 +80,22 @@ public function test_create_state_uses_the_factory() {
...
+		$this->expectExceptionMessage( 'The WCPay\Internal\Payment\State\AbstractPaymentState::' . $method . ' method is not available in the current payment state (' . PureState::class . ').' );
+
+		// Dynamically call the method based on $method.
+		$this->sut->{$method}( $this->createMock( PaymentRequest::class ) );
+	}
+
+	public function state_defined_methods_provider() {
+		return [
+			[ 'start_processing' ],
+			[ 'process' ],
```
RadoslavGeorgiev: This one should go away, we only have `start_processing` and `complete_processing` right now.

With that in mind, let's just have 2 separate tests for the two methods. We shouldn't assume every transition method requires either the request as a parameter, or nothing.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -107,11 +109,90 @@ public function test_process_payment_happy_path_without_mock_builder() {
...
+			->method( 'start_processing' )
 			->with( $this->isInstanceOf( PaymentRequest::class ) )
 			->willReturn( $mock_completed_state );
 
 		$result = $sut->process_payment( 1 );
 		$this->assertSame( $mock_completed_state, $result );
 	}
+
+	public function test_get_authentication_redirect_url_will_return_url_from_payment_intent() {
+		$sut = new PaymentProcessingService( $this->mock_state_factory, $this->mock_legacy_proxy );
```
RadoslavGeorgiev, line 118: `$sut` is defined explicitly in the previous test because it tests `create_payment_context` (that's mocked normally). For this and all following tests, you can use `$this->sut` directly.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -107,11 +109,90 @@ public function test_process_payment_happy_path_without_mock_builder() {
...
+					$intent->get_client_secret(),
+					'aes-128-cbc',
+					substr( $intent->get_customer_id(), 5 ),
+					0,
+					str_repeat( 'WC', 8 ),
+				],
+				[ 'wp_create_nonce', 'wcpay_update_order_status_nonce' ]
+			)
+			->willReturnOnConsecutiveCalls( $secret, $nonce );
+		$result = $sut->get_authentication_redirect_url( $intent, $order );
```
RadoslavGeorgiev: RadoCS: In this, and the next test, let's add an empty line before `$result = ...`.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -107,11 +109,90 @@ public function test_process_payment_happy_path_without_mock_builder() {
...
+			->willReturn( false ); // Just to test when encryption is disabled.
+
+		$this->mock_legacy_proxy->expects( $this->once() )
+			->method( 'call_function' )
+			->with( 'wp_create_nonce', 'wcpay_update_order_status_nonce' )
+			->willReturn( $nonce );
+		$result = $sut->get_authentication_redirect_url( $intent, $order );
+		$this->assertSame( "#wcpay-confirm-si:$order:" . $intent->get_client_secret() . ":$nonce", $result );
+	}
+
```
RadoslavGeorgiev, line 199: RadoCS: Too many empty lines.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -107,11 +109,90 @@ public function test_process_payment_happy_path_without_mock_builder() {
...
+					'aes-128-cbc',
+					substr( $intent->get_customer_id(), 5 ),
+					0,
+					str_repeat( 'WC', 8 ),
+				],
+				[ 'wp_create_nonce', 'wcpay_update_order_status_nonce' ]
+			)
+			->willReturnOnConsecutiveCalls( $secret, $nonce );
+		$result = $sut->get_authentication_redirect_url( $intent, $order );
+		$this->assertSame( "#wcpay-confirm-pi:$order:$secret:$nonce", $result );
```
RadoslavGeorgiev: Can we add a data provider here, and test with both `pi` and `si` through it? This way `test_get_authentication_redirect_url_will_return_url_with_secret_and_si_prefix` (is the method name correct there btw?) can focus on what happens when encryption is disabled.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -7,15 +7,16 @@
 
 namespace WCPay\Tests\Internal\Payment\State;
 
-use Exception;
+use WC_Helper_Intention;
+use WCPay\Constants\Intent_Status;
+use WCPay\Internal\Payment\State\AuthenticationRequiredState;
+use WCPay\Internal\Payment\State\ProcessedState;
+use WCPay\Internal\Payment\State\VerifiedState;
```
RadoslavGeorgiev: Unused `use` statement.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -57,12 +58,12 @@ class InitialStateTest extends WCPAY_UnitTestCase {
 	/**
 	 * @var Level3Service|MockObject
 	 */
-	private $mock_level3_service;
+	private $mock_payment_request_service;
```
RadoslavGeorgiev, line 67: Not sure why this got switched with the next property, but the `@var` definition is now incorrect.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -254,4 +253,20 @@ public function test_populate_context_from_order() {
 
 		PHPUnit_Utils::call_method( $this->sut, 'populate_context_from_order', [] );
 	}
+
+	private function mock_initial_state() {
```
RadoslavGeorgiev: Re-defining the service under tests like this:

1. Makes it a bit more hidden, and not so obvious what's being done, as it's in the end of the test.
2. Causes a bunch of errors when editing:

image

As there are a multiple methods that use the pure class as `sut`, and multiple others that need the mocked one, let's just define both of them in the `setUp` method, something like this:

```php
/**
 * Service under test.
 *
 * @var InitialState
 */
private $sut;

/**
 * Service under test with mocked protected methods.
 *
 * @var InitialState|MockObject
 */
private $mocked_sut;

/**
 * Set up the test.
 */
protected function setUp(): void {
	parent::setUp();

	// set up mocks...

	$this->sut = new InitialState(
		$this->mock_state_factory,
		$this->mock_order_service,
		$this->mock_customer_service,
		$this->mock_level3_service,
		$this->mock_payment_request_service
	);
	$this->sut->set_context( $this->mock_context );

	$this->mocked_sut = $this->getMockBuilder( InitialState::class )
		->onlyMethods( [ 'populate_context_from_request', 'populate_context_from_order' ] )
		->setConstructorArgs(
			[
				$this->mock_state_factory,
				$this->mock_order_service,
				$this->mock_customer_service,
				$this->mock_level3_service,
				$this->mock_payment_request_service,
			]
		)
		->getMock();
	$this->mocked_sut->set_context( $this->mock_context );
}
```

Then tests can decide which one to use.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -92,92 +93,90 @@ protected function setUp(): void {
...
-					$this->mock_state_factory,
-					$this->mock_order_service,
-					$this->mock_customer_service,
-					$this->mock_level3_service,
-					$this->mock_payment_request_service,
-				]
-			)
-			->getMock();
-		$this->sut->set_context( $this->mock_context );
+		$this->mock_initial_state();
```
RadoslavGeorgiev: The comment above this line is useless now, let's remove it.

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
```diff
@@ -0,0 +1,84 @@
...
+use WC_Helper_Intention;
+use WCPay\Internal\Payment\PaymentContext;
+use WCPay\Internal\Payment\State\CompletedState;
+use WCPay\Internal\Payment\State\ProcessedState;
+use WCPay\Internal\Payment\State\StateFactory;
+use WCPay\Internal\Service\OrderService;
+use WCPAY_UnitTestCase;
+
+/**
+ * Tests for the initial payment state.
```
RadoslavGeorgiev: Outdated comment.

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Class ProcessedStateTestTest
```
RadoslavGeorgiev: `TestTest` :)

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+		$this->sut = new ProcessedState(
+			$this->mock_state_factory,
+			$this->mock_order_service
+		);
+		$this->sut->set_context( $this->mock_context );
+	}
+
+	public function test_complete_processing_will_transition_to_completed_state() {
+
```
RadoslavGeorgiev: RadoCS: Let's remove this empty line.

RadoslavGeorgiev (changes requested):
Looks pretty well, only a few nitpicks left :)

Also, this would be a good time to update the PR description, and mark it as ready for review (not draft).

src/Internal/Payment/PaymentContext.php
```diff
@@ -245,4 +246,22 @@ public function set_customer_id( string $customer_id ) {
 	public function get_customer_id(): ?string {
 		return $this->get( 'customer_id' );
 	}
+
+	/**
+	 * Stores the payment intent.
```
htdat: Nitpick: 

```suggestion
	 * Stores the payment intent object.
```

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7471` or branch name `rpp/7414-authentication-required-state` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7414-authentication-required-state) :rocket:

---

Build info: 

- Latest commit: 518dff16e722f23d30bb7cc60995d93e75c24f96
- Build time: 2023-10-23 12:38:38 UTC

src/Internal/Payment/PaymentContext.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
zmaglica: FIxed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
zmaglica: Fixed :) https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
zmaglica: Comment is updated now: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: The comment is removed (ref: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d)

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: I have updated the test with the example you provided (https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d)

src/Internal/Payment/State/AbstractPaymentState.php
zmaglica: Used `use` statements as you suggested (https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d)

src/Internal/Payment/State/AbstractPaymentState.php
zmaglica: Removed (https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d)

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
zmaglica: Fixed (https://github.com/Automattic/woocommerce-payments/pull/7471/commits/48da0fc92461c2e53787cf45c2e79dcf0fa5bf8d)

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -38,6 +38,12 @@ class InitialStateTest extends WCPAY_UnitTestCase {
 	 * @var InitialState
 	 */
 	private $sut;
+	/**
```
RadoslavGeorgiev, line 42: RadoCS: Let's add a line before this comment.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -38,6 +38,12 @@ class InitialStateTest extends WCPAY_UnitTestCase {
...
+	 * Service under test.
+	 *
+	 * @var StateFactory|InitialState
```
RadoslavGeorgiev:
```suggestion
	 * @var MockObject|InitialState
```

Just curious, doesn't PHPStorm highlight errors because of this? I'm noticing that @jessy-p's tests have something similar.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
RadoslavGeorgiev: This still needs to be fixed.

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
```diff
@@ -0,0 +1,83 @@
...
+	/**
+	 * Set up the test.
+	 */
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->mock_state_factory = $this->createMock( StateFactory::class );
+		$this->mock_order_service = $this->createMock( OrderService::class );
+		$this->mock_context       = $this->createMock( PaymentContext::class );
```
RadoslavGeorgiev, line 53: The `$mock_context` property is not defined, let's define it.

RadoslavGeorgiev (changes requested):
Only a couple of small (new) things I noticed. There are some comments from previous reviews that are not addressed, probably because they were collapsed:

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: It should, but perhaps it doesn't check tests. I will check that.

tests/unit/src/Internal/Payment/State/ProcessedStateTest.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb

src/Internal/Payment/State/InitialState.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb 

But I will also check the whole PR and fix if I detect some.

tests/unit/src/Internal/Payment/State/AbstractPaymentStateTest.php
zmaglica: This is addressed.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: I used dataprovider as you suggested and changed test names. Now the test for non encrypted secret key is standalone and PI/SI prefix tests are handled by provider.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: Fixed in this commit: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: FIxed. https://github.com/Automattic/woocommerce-payments/pull/7471/commits/97345dbc66b1cc740585d7850147f228ccdad2eb

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/b35ba641977dfd7e93c02b85d336adae06f2dcf7

src/Internal/Payment/State/AuthenticationRequiredState.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @package WooCommerce\Payments
+ */
+
+namespace WCPay\Internal\Payment\State;
+
+/**
+ * The state, which indicates that the payment requires authentication (3DS).
+ */
+class AuthenticationRequiredState extends AbstractPaymentState {
```
htdat, line 13: Based on https://github.com/Automattic/woocommerce-payments/pull/7471#discussion_r1361620696: 

```suggestion
class PendingAuthenticationState extends AbstractPaymentState {
```

src/Internal/Payment/State/ProcessedState.php
```diff
@@ -0,0 +1,58 @@
...
+
+use WCPay\Exceptions\Order_Not_Found_Exception;
+use WCPay\Internal\Payment\Exception\StateTransitionException;
+use WCPay\Internal\Service\OrderService;
+use WCPay\Vendor\League\Container\Exception\ContainerException;
+
+/**
+ * This state is used when payment is completed on the server, and we need to update date on the plugin side.
+ */
+class ProcessedState extends AbstractPaymentState {
```
htdat, line 18: Based on https://github.com/Automattic/woocommerce-payments/pull/7471#discussion_r1361620696: 


```suggestion
class CapturedState extends AbstractPaymentState {
```

tests/unit/src/Internal/Payment/State/AbstractPaymentStateTest.php
RadoslavGeorgiev: @zmaglica I see that the methods are updated, but the bigger point of the comment was not to use a data provider, instead having two separate tests for the two methods.

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -90,41 +97,8 @@ protected function setUp(): void {
...
-		$mock_intent  = $this->createMock( WC_Payments_API_Payment_Intention::class );
-
-		/**
-		 * This test works with the root `process` method, which calls a few
-		 * internal methods. We want to mock them for the purpose of this test.
-		 *
-		 * @var MockObject|InitialState
-		 */
-		$this->sut = $this->getMockBuilder( InitialState::class )
+		$this->mocked_sut = $this->getMockBuilder( InitialState::class )
```
RadoslavGeorgiev, line 107: We still need the comment with the `@var` declaration here, at least for VS Code: 
image

Even though the private property is defined as `InitialState|MockObject`, the definition here makes the editor thing it's just a `MockObject`

You can simply restore the comment that was there:
```suggestion
		/**
		 * This test works with the root `process` method, which calls a few
		 * internal methods. We want to mock them for the purpose of this test.
		 *
		 * @var MockObject|InitialState
		 */
		$this->mocked_sut = $this->getMockBuilder( InitialState::class )
```

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -92,8 +96,6 @@ public function test_process_payment_happy_path() {
 	 * Test the basic happy path of processing a payment.
 	 */
 	public function test_process_payment_happy_path_without_mock_builder() {
-		$sut = new PaymentProcessingService( $this->mock_state_factory, $this->mock_legacy_proxy );
```
RadoslavGeorgiev: This is the one method that should indeed use the `PaymentProcessingService` directly. Now coverage has gone down for the file, because `create_payment_context` is not covered with tests.

RadoslavGeorgiev (changes requested):

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -68,23 +72,19 @@ public function test_process_payment_happy_path() {
 		$mock_initial_state   = $this->createMock( InitialState::class );
 		$mock_completed_state = $this->createMock( CompletedState::class );
```
RadoslavGeorgiev, line 72: The mock context (line 71) is not used, let's remove that line.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: Addressed here. https://github.com/Automattic/woocommerce-payments/pull/7471/commits/2c165178f28b4e6a39827c60fd9e55320d218bff

tests/unit/src/Internal/Payment/State/AbstractPaymentStateTest.php
zmaglica: Good point. I missed that. Corrected here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/2c165178f28b4e6a39827c60fd9e55320d218bff

src/Internal/Payment/State/ProcessedState.php
RadoslavGeorgiev: Sorry for mistakenly giving my 👍 here, but this is not right.

Because of the patch, I am still to share the state graph so we can discuss it, so let's undo this change for now. `CapturedState` should come after `ProcessedState`, it is not the same.

It would be:
1. `InitialState`
2. `PendingAuthenticationState`
3. `ProcessedState`
4. `CapturedState` or `PendingCaptureState`

The point is not to end up in a `CompletedState`, as a payment is never "complete".

Let's undo this change. We'll finalize state names in a different issue/PR.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
zmaglica: Good catch. Fixed here: https://github.com/Automattic/woocommerce-payments/pull/7471/commits/518dff16e722f23d30bb7cc60995d93e75c24f96

src/Internal/Payment/State/ProcessedState.php
zmaglica: I have reverted a commit that changed states name.

RadoslavGeorgiev (approved):
LGTM! :shipit: 

Thank you for bearing with me through all the nitpicks :)

zmaglica:
> LGTM! :shipit:
> 
> Thank you for bearing with me through all the nitpicks :)

I love nitpicks 😃  Thank you for the extended review 🙇

PR #7328, opened 2023-09-29
Enables deferred intent UPE for existing legacy card stores

timur27:
Closes https://github.com/Automattic/woocommerce-payments/issues/7221

❗ **Should be merged between Oct 9 and Oct 27**. Keeping the PR status _Draft_ to avoid an accidental merge.

#### Changes proposed in this Pull Request
This PR is the third and the last step of migration to deferred intent creation UPE. Previously, we migrated split UPE stores in https://github.com/Automattic/woocommerce-payments/pull/7058 and legacy UPE stores in https://github.com/Automattic/woocommerce-payments/pull/7306. During legacy UPE stores migration, we switched to the server-side feature flag, which made the implementation in https://github.com/Automattic/woocommerce-payments/pull/7306 a bit more complex than it could have been because we needed to support both client-side and server-side stores. This PR removes all this complexity since, starting at `6.7.0`, each store should rely only on the server-side feature flag.

#### Testing instructions

Speaking high-level, we need to check that legacy card/legacy UPE/split UPE stores are now all using deferred intent creation UPE, that there's no fallback available for merchants but if we turn off the dUPE toggle for legacy card or legacy UPE, the store should roll back with corresponding notice in the Payments settings. 

> Pre-requisite: In your local server code, change https://github.com/Automattic/woocommerce-payments-server/pull/3963/files#diff-c79fe5913d8e8943eac11e58823d53e42ce3e6234736f0ed0c745fa901b53c17R1844 to compare against 6.5.0 so that your local is treated as eligible, for testing purposes

##### Legacy card stores

1. For [this configuration](https://github.com/Automattic/woocommerce-payments/assets/22319444/40d458b9-571c-4e17-a320-77d6cd3ec2c1) (no UPE is enabled), confirm that `'is_deferred_intent_creation_upe_enabled' => true,` in the account cache contents. This means that by default, even if the feature flag doesn't exist in the server's DB, deferred UPE will be treated as enabled by the server. 
2. Navigate to `Payments -> Settings` and confirm that your settings [look the following](https://github.com/Automattic/woocommerce-payments/assets/22319444/568c869f-f2b1-4d16-b26e-d251f9fa7f42), meaning no notices, banners, but UPE methods
3. Click the kebab menu on the upper right and confirm that only the _Provide feedback_ button is available, but not _Disable_, like it was previously.
[...]

Issue #6882, opened 2023-07-31
Improve integration tests for deferred intent UPE


FangedParakeet:
### Description

Currently, in our deferred intent UPE integration work, we have relied greatly on extending existing functionality. Our checkout code inherits the original UPE payment gateway class and relies on the legacy gateway's `process_payment` function to complete a checkout. Similarly our unit test coverage mostly consists of extending existing UPE unit tests. Consequently our unit test coverage is somewhat lacking in that we neglect to comprehensively the new deferred intent flows of logic from the UPE gateway to the legacy gateway, testing all components in concord concurrently. 

It would be nice if we could create some new unit tests that effectively test the deferred intent UPE checkout by ensuring that relevant functions are only fired when necessary, metadata is always added when relevant, and that redundant UPE function calls are no longer made.

### Acceptance criteria

Here are a few things of the top of my head that would be nice to have specific integration test coverage for.
- Complete deferred intent checkout flow.
- Assert/count calls to payments API (e.g. calls to retrieve payment method, calls to update intent, etc.).
- Assert intent metadata added to WC order.
- Assert WC order metadata added to intent.
- Assert calls to hooks.
- Platform checkout flow.
- Assert additional request fields for Stripe Link (e.g. `payment_method_types`, `mandate_data`).

### Designs

N/A

### Testing instructions

Exactly.

### Additional context

#6391

FangedParakeet:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Heisenberg-63586f2ec60359001310fa6a/issues/Automattic/woocommerce-payments/6882?planning-poker) with Zenhub @timur27

timur27:
Use cases for saving payment methods to check and try to integrate into the testing coverage
https://github.com/Automattic/woocommerce-payments/issues/5022#issuecomment-1663973487

timur27:
* This issue can also include checking and possibly improving E2E & unit test coverage for deferred intent creation UPE. 
* While working on this issue, we should consider what is the status of https://app.zenhub.com/workspaces/heisenberg-63586f2ec60359001310fa6a/issues/gh/automattic/woocommerce-payments/5028, because both issues can have some common areas

PR #3006, opened 2021-09-28
Fix error message when adding new credit card

leonardola:
When a customer adds a new credit card inside My account > Payment Methods using other gateways than WC Payments the error "Your credit card number is invalid" would pop up but the credit card would be added anyways as the error was not related to the payment gateway selected.

Fixes #1149 

#### Changes proposed in this Pull Request
When the customer clicked on Add payment method, WC Payments validated if the information was correctly filled on WC Payments fields, even when the customer selected other Payment Gateway. I've added a conditional to make sure we only validate WC Payments fields if the customer selected a WC Payments Gateway.

Also added tests to guarantee this problem does not happen again in the future.

#### Testing instructions

- Have WC Payments installed and configured
- Have some other Credit Card Payment method installed and configured
- Log in as a customer
- Go to My account > Payment Methods
- Click on Add Payment Method
- Fill the **other gateway** Credit Card details
- Try to submit
- The error "Your card number is incomplete" should not show up

-------------------

- [x] Added changelog entry (or does not apply)
- [x] Covered with tests (or have a good reason not to test in description ☝️)

**Post merge**

- [ ] Added testing instructions to the [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) (or does not apply)

PR #7478, opened 2023-10-13
Fix E2E tests to support deferred intent UPE

timur27:
Post #7328, all stores, including the legacy card store, are treated as parties utilizing deferred intent creation UPE. Consequently, specific E2E tests covering legacy card use cases started failing. This PR fixes those tests.

While fixing tests, it became clear that we also need to clean up the tests' implementation by removing unnecessary feature flag toggling (since stores already use dUPE by default) and eliminate redundant `card/legacy UPE/split UPE` tests to leave only one group of tests ultimately. This is supposed to be done in a separate issue, and I think https://github.com/Automattic/woocommerce-payments/issues/6882 is a good candidate to cover it as well.

Some context on the changes: 
* By adding a check in `event-handlers.js`, we ensure that deferred UPE flow is not executed on the _Add payment method_ page if a gateway other than WooPayments is being used. This was done for other gateways in https://github.com/Automattic/woocommerce-payments/pull/3006, but skipped in deferred UPE initially because the error doesn't appear. However, the tests fail. I thought it was still important to stop the flow execution if another plugin is used, so I re-introduced this check and [the corresponding test](https://github.com/Automattic/woocommerce-payments/blob/edd225a66243123d74364db4c68c27dcc98728ed/tests/e2e/specs/wcpay/shopper/shopper-myaccount-payment-methods-add-fail.spec.js#L59) now passes.
* `tests/e2e/utils/payments.js` now handle Stripe UPE when filling the card details on _Pay for order page_ as well as cleaning up card details
* `await clearCardDetails()` invocation in `/wcpay/shopper/shopper-checkout-failures.spec.js` was moved to the end of every test, because the last 3DS-related test runs the 3DS authentication modal in the `iframe` which then clashes with the payment element `iframe`, making it not possible to clear the card data. Since clearing the card data is an optional step, we now skip it for the last 3DS flow
* [This assertion](https://github.com/Automattic/woocommerce-payments/blob/edd225a66243123d74364db4c68c27dcc98728ed/tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js#L52-L60) was removed because we don't need to test the behavior of the Stripe Payment Element, we should focus on the functionality on our side.

#### Testing instructions

* In GitHub Actions, run [PR flow](https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-pull-request.yml) on this branch twice sequentially and confirm all the tests are passing
* In GitHub Actions, run [E2E Tests - All](https://github.com/Automattic/woocommerce-payments/actions/workflows/e2e-test.yml) flow on this branch and confirm that all the tests pass
* Ensure your local store uses deferred intent creation UPE, follow [these testing steps](https://github.com/Automattic/woocommerce-payments/pull/3006) and confirm they are successful. Right after, confirm that the payment method was successfully added by other gateway from that test.
* Perform a sanity check of the _add new payment method_ flow - add a new card using deferred intent creation UPE and confirm that it was added successfully and you can use it by making following purchases
-------------------

**Post merge**

### DIFF
--- a/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
+++ b/client/checkout/classic/upe-deferred-intent-creation/event-handlers.js
@@ -77,6 +77,13 @@ jQuery( function ( $ ) {
 	}
 
 	$( 'form#add_payment_method' ).on( 'submit', function () {
+		if (
+			$(
+				"#add_payment_method input:checked[name='payment_method']"
+			).val() !== 'woocommerce_payments'
+		) {
+			return;
+		}
 		// WC core calls block() when add_payment_method form is submitted, so we need to enable the ignore flag here to avoid
 		// the overlay blink when the form is blocked twice.
 		$.blockUI.defaults.ignoreIfBlocked = true;
--- a/tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js
+++ b/tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js
@@ -49,15 +49,6 @@ describe( 'Shopper > Checkout > Failures with various cards', () => {
 	it( 'should throw an error that the card expiration date is in the past', async () => {
 		const cardInvalidExpDate = config.get( 'cards.invalid-exp-date' );
 		await fillCardDetails( page, cardInvalidExpDate );
-		await page.waitForSelector(
-			'div#wcpay-errors > ul.woocommerce-error > li'
-		);
-		await expect( page ).toMatchElement(
-			'div#wcpay-errors > ul.woocommerce-error > li',
-			{
-				text: "Your card's expiration year is in the past.",
-			}
-		);
 		await expect( page ).toClick( '#place_order' );
 		await uiUnblocked();
 		await expect(
--- a/tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js
+++ b/tests/e2e/specs/wcpay/shopper/shopper-checkout-failures.spec.js
@@ -23,11 +23,6 @@ describe( 'Shopper > Checkout > Failures with various cards', () => {
 		);
 	} );
 
-	afterEach( async () => {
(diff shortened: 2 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7478` or branch name `fix/e2e-tests-after-legacy-card-migration` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/e2e-tests-after-legacy-card-migration) :rocket:

---

Build info: 

- Latest commit: 4509f6b300c355667e3e894e76d93fb43aefe252
- Build time: 2023-10-16 07:48:58 UTC

frosso (approved):
Thanks for the explanation of the changes, it was really helpful to read! 👍 

Do you have any insight on why `clearCardDetails()` needs to support both UPE and non-UPE flows? 

Other than that, this looks good and the tests are passing :shipit:

timur27:
@frosso Thanks for the review!

> Do you have any insight on why clearCardDetails() needs to support both UPE and non-UPE flows?

Great point. No reason and this support essentially needs to be removed. I've left this effort for https://github.com/Automattic/woocommerce-payments/issues/6882, because during a quick removal, there were some other errors appearing, and given the priority of https://github.com/Automattic/woocommerce-payments/pull/7328 I thought that it's important to merge it sooner and perform a cleanup in #6882. Let me know if this makes sense (a 👍 would be enough) or if there are any suggestions in this context, and I'm happy to align further!



PS I'll make sure to update #6882 within the next 24hrs to reflect the scope change based on this PR description.

frosso:
Thanks for the context @timur27 ! Let's get this in 👍

PR #7494, opened 2023-10-17
fix: Stripe Link eligibility check

frosso:
Reported p1697496446121059-slack-C7U3Y3VMY

In a few, very edgy scenarios, `link` might be an item in the `upe_enabled_payment_method_ids` for the WooPayments gateway.
This can be a problem for non-US accounts, where Link is not available.
We're confirming with Stripe whether this error only appears in live mode (in test mode this doesn't seem to be an issue) p1697634872443619-slack-C9976E5MJ

#### Testing instructions

- Without this fix:
	- Onboard with a non-US account
	- Try and force `link` in the `upe_enabled_payment_method_ids` option (you'll need to do this manually)
	- Add a product to the cart, go to checkout, place the order
	- You should see `link` in the list of payment methods sent to the `POST /v1/payment_intents` endpoint
Screenshot 2023-10-18 at 4 10 45 PM
- With this fix:
	- Onboard with a non-US account
	- Try and force `link` in the `upe_enabled_payment_method_ids` option (you'll need to do this manually)
	- Add a product to the cart, go to checkout, place the order
	- You should not see `link` in the list of payment methods sent to the `POST /v1/payment_intents` endpoint
Screenshot 2023-10-18 at 4 11 45 PM
- Try the same for a US-based account (with Stripe Link enabled) - there should be no issue

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-stripe-link-eligibility
+++ b/changelog/fix-stripe-link-eligibility
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Stripe Link eligibility at checkout
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1560,23 +1560,27 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	public function get_payment_methods_from_gateway_id( $gateway_id, $order_id = null ) {
 		$split_upe_gateway_prefix = self::GATEWAY_ID . '_';
 		// If $gateway_id begins with `woocommerce_payments_` payment method is a split UPE LPM.
-		// Otherwise $gateway_id must be `woocommerce_payments`.
+		// Otherwise, $gateway_id must be `woocommerce_payments`.
 		if ( substr( $gateway_id, 0, strlen( $split_upe_gateway_prefix ) ) === $split_upe_gateway_prefix ) {
-			$payment_methods = [ str_replace( $split_upe_gateway_prefix, '', $gateway_id ) ];
-		} elseif ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
+			return [ str_replace( $split_upe_gateway_prefix, '', $gateway_id ) ];
+		}
+
+		$eligible_payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( $order_id, true );
+
+		if ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
 			// If split or deferred intent UPE is enabled and $gateway_id is `woocommerce_payments`, this must be the CC gateway.
 			// We only need to return single `card` payment method, adding `link` since deferred intent UPE gateway is compatible with Link.
 			$payment_methods = [ Payment_Method::CARD ];
-			if ( in_array( Payment_Method::LINK, $this->get_upe_enabled_payment_method_ids(), true ) ) {
+			if ( in_array( Payment_Method::LINK, $eligible_payment_methods, true ) ) {
 				$payment_methods[] = Payment_Method::LINK;
 			}
-		} else {
-			// $gateway_id must be `woocommerce_payments` and gateway is either legacy UPE or legacy card.
-			// Find the relevant gateway and return all available payment methods.
-			$payment_methods = WC_Payments::get_gateway()->get_payment_method_ids_enabled_at_checkout( $order_id, true );
+
+			return $payment_methods;
 		}
 
-		return $payment_methods;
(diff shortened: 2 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7494` or branch name `fix/stripe-link-eligibility` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/stripe-link-eligibility) :rocket:

---

Build info: 

- Latest commit: 6e7a3362ea267513a813a28979dad6aa02037251
- Build time: 2023-10-18 14:59:39 UTC

timur27 (commented):
I'll post the testing result in a separate commit together with potential approval.

### Reviewing code changes

Code-wise, the change looks good. Given that newest versions of WooPayments can _technically_ run legacy card, legacy UPE and deferred intent creation UPE, I made the following analysis: 

- Store uses _legacy card_
  - The result of the function will be the same as of the function prior to the changes, because [`get_payment_method_ids_enabled_at_checkout` will be returned](https://github.com/Automattic/woocommerce-payments/pull/7494/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1598), similar to [as it was previously](https://github.com/Automattic/woocommerce-payments/pull/7494/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124L1606)
- Store uses _legacy UPE_
    - Same as the above legacy card conclusion
- Store uses _deferred intent creation UPE_
  - No card
    - No change, because we just extend the array with [`Payment_Method::CARD`](https://github.com/Automattic/woocommerce-payments/pull/7494/files#diff-bcfd303d507c33a27671935e724ee64d0df30ff2a18bb6ef4c94f28d48190124R1603)
  - Stripe Link
    - This is changed now and this is the main goal of this PR. We now check if Stripe Link is present in the `get_payment_method_ids_enabled_at_checkout` array, while previously, we checked if `get_upe_enabled_payment_method_ids` contains Stripe Link. This change helps us to perform more check and double-confirm, that Stripe Link is not used on the checkout page. 

### Thinking about the problem root cause
Looking more high-level at the problem we faced, I have the feeling, that our implementation (prior to this fix) is the correct implementation, but the problem is we are dealing with the wrong configuration/state on some subset of the merchant sites, which for some reason, have Stripe Link enabled, although they are not eligible. This requires us to add additional filtering which this PR successfully accomplishes. 

> we are dealing with the wrong configuration/state on some subset of the merchant sites, which for some reason, have Stripe Link enabled, although they are not eligible

Or, alternatively, `get_upe_enabled_payment_method_ids` might give us a result we don't expect. But I doubt it.

### Proposing a follow up item
One thing that requires a follow up in my opinion, is to find a way for WooPayments Client to distinguish if it processes a credit card filled out with the help of Stripe Link, or a normal credit card without Stripe Link. This way, we won't attach `link` to every card payment like we do currently, thus leading us to a more clean logic which (what's more important) is more aligned with our users' actions.

frosso:
Thanks @timur27 ! I logged a separate ticket with your proposal here: https://github.com/Automattic/woocommerce-payments/issues/7507

timur27 (approved):
The changes are testing well. I performed tests/sanity checks on my local with legacy card, legacy UPE and deferred intent creation UPE. I've also performed tests on the JN sites below that were created and provided by @frosso (many thanks for it!)

1️⃣ `https://fantastic-caribbean.jurassic.ninja/` non-US store (not eligible for Stripe Link) which is forced to have Stripe Link as the available payment method for testing purposes.
* [Removes the redundant `link`](https://dashboard.stripe.com/acct_1O2WwbCIIZuGWpba/test/logs/req_UopI43liQJynSA) payment method since it's unavailable for this store.

```json
"payment_method_types": [
    "card"
]
```

2️⃣ `http://arrogant-hector.jurassic.ninja/` US store with Link
* Keeps Link payment method in the request body
```json
"payment_method_types": [
    "card",
    "link"
]
```


3️⃣ `https://beneficial-reindeer.jurassic.ninja/` US store without Link
* Payments are processed successfully
* Doesn't have "link" property in the request body
```json
"payment_method_types": [
    "card"
]
```

frosso:
Confirmed that the e2e failures can be ignored in this case: p1697647246500959/1697635378.339339-slack-CGGCLBN58
Confirmed that the test mode behaves differently than live mode in this case, so that's why the methods were running fine for us - we'd need a live account to make this work accurately: p1697649888244579/1697634872.443619-slack-C9976E5MJ

Merging.

PR #7503, opened 2023-10-18
Tmp/check logging changes

jessy-p:


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7503` or branch name `tmp/check-logging-changes` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=tmp/check-logging-changes) :rocket:

---

Build info: 

- Latest commit: 3a3413e2f7b8cc70277aaf0b20a00b6e7635ef98
- Build time: 2023-10-18 17:52:13 UTC

src/Internal/Payment/Change.php
```diff
@@ -0,0 +1,76 @@
...
+	public function __construct( $key, $old_value, $new_value ) {
+		$this->key       = $key;
+		$this->old_value = $old_value;
+		$this->new_value = $new_value;
+	}
+
+	/**
+	 * Defines which data will be serialized to JSON
+	 */
+	public function jsonSerialize(): array {
```
RadoslavGeorgiev, line 54: Let's avoid JSON as an idea for now, we can remove this method. The `var_export` in `__toString()` should do the job better, because we'd be seeing times as well.

src/Internal/Payment/Change.php
```diff
@@ -0,0 +1,76 @@
...
+	 *
+	 * @return string String representation of the object.
+	 */
+	public function __toString() {
+		return sprintf(
+			'Change: %s from %s to %s',
+			$this->key,
+			var_export( $this->old_value, true ), // phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_var_export
+			var_export( $this->new_value, true ) // phpcs:disable WordPress.PHP.DevelopmentFunctions.error_log_var_export
+		);
```
RadoslavGeorgiev, line 73: Let's wait until https://github.com/Automattic/woocommerce-payments/pull/7471 is merged, and revisit this string generation. The context will also contain serializable objects (intents), and this format might not work, or at least be insufficient.

src/Internal/Payment/PaymentContext.php
```diff
@@ -225,6 +232,30 @@ public function get_customer_id(): ?string {
 		return $this->get( 'customer_id' );
 	}
 
+	/**
+	 * Returns the changes.
+	 *
+	 * @return array
```
RadoslavGeorgiev, line 238: Let's change this to `@return Change[]` to make it more explicit.

src/Internal/Payment/PaymentContext.php
```diff
@@ -225,6 +232,30 @@ public function get_customer_id(): ?string {
...
+	public function get_changes(): array {
+		return $this->changes;
+	}
+
+	/**
+	 * Returns the changes as a string that can be logged.
+	 *
+	 * @return string
+	 */
+	public function get_changes_log(): string {
```
RadoslavGeorgiev, line 249: Let's do this in the `PaymentProcessingService` instead. In the future, when payments are stored in the DB, it will be a class there that stores everything needed, incl. changes/logs.

src/Internal/Payment/PaymentContext.php
```diff
@@ -225,6 +232,30 @@ public function get_customer_id(): ?string {
...
+		$change_strings = array_map(
+			function( $change ) {
+				return (string) $change;
```
RadoslavGeorgiev, line 252: Wouldn't `implode` trigger the conversion to string automatically?

src/Internal/Payment/PaymentContext.php
```diff
@@ -233,6 +264,7 @@ public function get_customer_id(): ?string {
 	 * @param mixed  $value Value to store.
 	 */
 	private function set( string $key, $value ) {
+		$this->changes[]    = new Change( $key, $this->get( $key ) ?? '', $value );
```
RadoslavGeorgiev, line 267: 1. Let's only log _changes_. Presently, everything is only set once, but in the future, we might do the same thing multiple times (ex. extract order data), and we should only log the things that actually changed.
2. `get` would return `null`, let's keep it `null`. In various cases, empty strings are a different thing.

RadoslavGeorgiev (changes requested):
The direction looks great, I left a few minor comments.

In general, I'd also like to make sure that all state changes are logged as well. What are your thoughts on dumping and clearing the log every time the state changes? This way we'd store transitions as well, but we'll also be able to see the timestamps of transitions. Keeping API calls in mind, those might be helpful.

jessy-p:
Refer #7513

PR #7510, opened 2023-10-18
RPP / Add 100% unit test coverage requirement for src

RadoslavGeorgiev:

Allow the `src` directory's unit coverage to be checked separately from `includes`.

1. There is a new `test:php-coverage-src` npm command, which only checks `src`, and is significantly faster than the full `npm run test:php-coverage` command.
2. A new "Code coverage (src) / Code coverage" workflow in GitHub which fails if coverage is less than 100%.

How this was achieved:

- Introduced a custom `phpunit-src.xml.dist` to allow the selection of only the `src` directory. Otherwise it is the same as `phpunit.xml.dist`.
- `bin/phpunit.sh` does not run by with `-c phpunit.xml.dist` by default anymore, instead allowing other commands to define it.
- Introduced 3 new files:
    1. `bin/check-src-test-coverage.sh` for the local command. Identical to `bin/check-test-coverage.sh`, but with the new XML config.
    2. `bin/run-ci-tests-check-src-coverage.bash` for the CI job, almost identical to the non-`src` one as well.
    3. `.github/workflows/src-coverage.yml` is the definition of the GitHub workflow.

#### Testing instructions

1. Run both `npm run test:php-coverage` and `npm run test:php-coverage-src` locally to make sure they work.
2. All CI checks should pass in GitHub.

-------------------


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7510` or branch name `rpp/100-coverage-command` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/100-coverage-command) :rocket:

---

Build info: 

- Latest commit: 107069463846e3664a4005a1299fb460ecc2e2c1
- Build time: 2023-10-27 08:16:39 UTC

bin/check-src-test-coverage.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "Checking coverage..."
+
+docker-compose exec -u www-data wordpress \
+	/var/www/html/wp-content/plugins/woocommerce-payments/vendor/bin/phpunit \
+	--configuration /var/www/html/wp-content/plugins/woocommerce-payments/phpunit-src.xml.dist \
+	--coverage-html /var/www/html/php-test-coverage \
+	--coverage-clover /var/www/html/clover.xml
+	$*
+
+./vendor/bin/coverage-check docker/wordpress/clover.xml 100
```
htdat: This is a nice improvement comparing `check-test-coverage.sh`.

However, I feel that we do not need a new `.sh` file, instead we can use `check-test-coverage.sh` and conditionally use a bash variable flag to run against either `phpunit-src.xml.dist` or `phpunit.xml.dist`.

bin/run-ci-tests-check-coverage.bash
```diff
@@ -11,6 +11,6 @@ composer self-update && composer install --no-progress
 sudo systemctl start mysql.service
 bash bin/install-wp-tests.sh woocommerce_test root root localhost $WP_VERSION $WC_VERSION false
 echo 'Running the tests...'
-bash bin/phpunit.sh --coverage-clover /tmp/clover.xml
+bash bin/phpunit.sh -c phpunit.xml.dist --coverage-clover /tmp/clover.xml
 vendor/bin/coverage-check /tmp/clover.xml 60
```
htdat: Similarly, I don't think we need a new bash file such as `bin/run-ci-tests-check-src-coverage.bash`. 

We can run both in the same bash file:

```bash
echo 'Running the tests for the whole repo...'
bash bin/phpunit.sh -c phpunit.xml.dist --coverage-clover /tmp/clover.xml
vendor/bin/coverage-check /tmp/clover.xml 60

echo 'Running the tests for `src` folder...'
bash bin/phpunit.sh -c phpunit-src.xml.dist --coverage-clover /tmp/clover-src.xml
vendor/bin/coverage-check /tmp/clover-src.xml 100
```

.github/workflows/src-coverage.yml
```diff
@@ -0,0 +1,39 @@
...
+          path: ~/.cache/composer/
+          key:  ${{ runner.os }}-composer-${{ hashFiles('composer.lock') }}
+      # setup PHP, but without debug extensions for reasonable performance
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          tools:       composer
+          coverage:    xdebug2
+      # run CI checks
+      - run: bash bin/run-ci-tests-check-src-coverage.bash
```
htdat: Also, I don't think we need a new workflow. 

With my suggestion on file `bin/run-ci-tests-check-coverage.bash`, checks for `src` folder have been included. 

In case that we still need to run this check separately, I think we can conditionally run `bin/run-ci-tests-check-coverage.bash`, and then add another `run` for the existing workflow file - after this line: 

https://github.com/Automattic/woocommerce-payments/blob/b94ccdff800339d50653a61f137bd93655afd8ad/.github/workflows/coverage.yml#L38-L39

htdat (commented):
Left some comments for re-using current bash and workflow files (i.e. DRY principle), rather than introducing new files with very similar structures of existing files - which can be a burden for Harmony to update them in the future.

RadoslavGeorgiev:
@htdat all good points. I've removed all nearly duplicate files, except for the XMLs. Could you please take another look?

htdat (commented):
@RadoslavGeorgiev - Looks excellent now. 

> I've removed all nearly duplicate files, except for the XMLs. 

Agreed, we need to have a new XML config file to have a separate code coverage report for `src` except that if there is a way that I don't know. However, even if there was a way, it seemed it's more complicated than just changing a few config lines. 

--- 

My last comment is that we might also need to use the `phpunit.xml.dist` config for only `includes`, and leave `src` for `phpunit-src.xml.dist`.
<details>

<summary>This is my suggestion: </summary>

```diff
Index: phpunit.xml.dist
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/phpunit.xml.dist b/phpunit.xml.dist
--- a/phpunit.xml.dist	(revision e5dc5a913730827dfdf40671d0e2788cd528ba02)
+++ b/phpunit.xml.dist	(date 1698307623999)
@@ -11,6 +11,7 @@
 		<testsuite name="WCPay">
 			<directory suffix=".php">./tests/unit</directory>
 			<exclude>./tests/unit/multi-currency</exclude>
+			<exclude>./tests/unit/src</exclude>
 		</testsuite>
 		<testsuite name="WCPayMultiCurrency">
 			<directory suffix=".php">./tests/unit/helpers</directory>
@@ -22,12 +23,6 @@
 	<filter>
 		<whitelist>
 			<directory suffix=".php">includes</directory>
-			<directory suffix=".php">src</directory>
-			<exclude>
-				<!-- Service providers are simple, and ideally should not be used within tests. -->
-				<directory suffix=".php">src/Internal/DependencyManagement/ServiceProvider</directory>
-			</exclude>
 		</whitelist>
 	</filter>
-
 </phpunit>
```

</details>

RadoslavGeorgiev:
@htdat I'd prefer to intentionally keep `src` covered by the standard commands. Most developers are/will not be aware of the difference in `src` requirements at first, and it might be confusing for them. I wanted a separate set of rules and commands:

- To allow quick `src` checks locally, especially for coverage.
- Enforce a 100% coverage in GitHub
- **Not to** run tests separately.

htdat:
@RadoslavGeorgiev - that also makes sense. If so, at least, we would still get rid of directory `src` from the code coverage in `phpunit.xml.dist` to maintain really 60% for `includes`. Otherwise, `includes` can go lower 60% while `src` keeps 100%, but the overall coverage can be still 60% - which is definitely NOT what we'd like to have at the end.

RadoslavGeorgiev:
@htdat done with 3 total files:

- `npm run test:php` through `phpunit.xml.dist` is the standard unit tests for all of the plugin, as devs are used to.
- `npm run test:php-coverage` through `phpunit-includes.xml.dist`, also used for CI checks.
- `npm run test:php-coverage-src` through `phpunit-src.xml.dist`, also used for CI checks.

The `includes` check is now down about 1% because of excluding `src`.

htdat (approved):
I think that makes sense now. Thanks for addressing my review.

RadoslavGeorgiev:
Merging without Harmony's approval for now. If needed, we can change things.

PR #7535, opened 2023-10-23
Always show Express payment buttons for Pay for Order page

### LINKED ISSUE #6650: QuickPay buttons don't fire on Pay For Order page (manual order payment page)
### Describe the bug
When a shop's customer has to pay for an order manually, they receive an email that directs them through to their My Account area on the website, to the Pay For Order page, to pay. This is a different page than the typical Checkout page.

You can also access this page if you are logged into your account and go to the My Account area, under Orders --> find an order in Payment Pending status and click the "Pay" button.

On the Pay For Order page, quick pay buttons (Google Pay and Apple Pay) are no longer visible.

You *can* see them in version 5.9.1 of WCPay
You *cannot* see them in version 6.1.1 of WCPay

Will test further to see if I can pinpoint when the bug was introduced.

### To Reproduce
1. Install and configure WooCommerce Payments and check the option to enable Quickpay buttons in the settings. (Double check: they should show up on the product/cart/checkout page, Google Pay in Chrome and Apple Pay in Safari.)

2. Create a manual order - can be for an existing customer on the site or a guest. It should be in Payment Pending status.
3. Use the Order Actions on the order to send an invoice email to the customer, like this:

4. Alternately --> if it is for an existing customer, you can log into their account and go under My Account > Orders, and click the "Pay" button next to the pending order, like this:

5. Either way, this loads the Pay For Order page --> the QuickPay buttons are missing (when testing with version 6.1.1 of WCPay).

6. Testing this with version 5.9.1 shows that the quick pay buttons do show up on this page, as shown here:

### Expected behavior
The quick pay buttons show show here as shown in the last screenshot above. Not sure exactly when this stopped working, somewhere between 5.9.1 and 6.1.1.

* Browser: tested in both Chrome for Google Pay, and Safari for Apple Pay

### Additional context
Wordpress version 6.2.2
WooCommerce version 7.8.1

elazzabi:
Fixes #6650

A while ago, a regression was introduced making Express pay buttons not showing in the Pay for Order page. This fixes the issue.

#### Testing instructions

You'll need these requirements to test this PR:
- Apple Pay/Google Pay setup
- An SSL-accessible website for your local env (like Jurassic tunnel)

Testing instructions:
* Take a Needs Payments order (you can create a new order and select cash)
* Go to WooCommerce > Orders > Your order. Copy the link in "Customer Payment page"

Screenshot 2023-10-23 at 12 13 48

* Open the link in either Safari/Chrome. There should not be a button for Google Pay/Apple Pay
* Get this PR and refresh the page
* The button for Google Pay/Apple Pay should be showing now next to WooPay (and after the order summary)

-------------------

**Post merge**

### DIFF
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -499,11 +499,6 @@ class WC_Payments_Payment_Request_Button_Handler {
 			return false;
 		}
 
-		// Order total doesn't matter for Pay for Order page. Thus, this page should always display payment buttons.
-		if ( $this->is_pay_for_order_page() ) {
-			return true;
-		}
-
 		// Page not supported.
 		if ( ! $this->is_product() && ! $this->is_cart() && ! $this->is_checkout() ) {
 			return false;
@@ -534,6 +529,11 @@ class WC_Payments_Payment_Request_Button_Handler {
 		if ( ( $this->is_checkout() || $this->is_cart() ) && ! $this->has_allowed_items_in_cart() ) {
 			Logger::log( 'Items in the cart have unsupported product type ( Payment Request button disabled )' );
 			return false;
+		}
+
+		// Order total doesn't matter for Pay for Order page. Thus, this page should always display payment buttons.
+		if ( $this->is_pay_for_order_page() ) {
+			return true;
 		}
 
 		// Cart total is 0 or is on product page and product price is 0.
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -381,8 +381,6 @@ class WC_Payments_Payment_Request_Button_Handler {
 		];
 
 		wp_localize_script( 'WCPAY_PAYMENT_REQUEST', 'wcpayPaymentRequestPayForOrderParams', $data );
-
-		$this->display_payment_request_button_html();
 	}
 
 	/**
--- a/includes/class-wc-payments-payment-request-button-handler.php
+++ b/includes/class-wc-payments-payment-request-button-handler.php
@@ -499,6 +499,11 @@ class WC_Payments_Payment_Request_Button_Handler {
 			return false;
 		}
 
+		// Order total doesn't matter for Pay for Order page. Thus, this page should always display payment buttons.
+		if ( $this->is_pay_for_order_page() ) {
+			return true;
+		}
+
 		// Page not supported.
 		if ( ! $this->is_product() && ! $this->is_cart() && ! $this->is_checkout() ) {
 			return false;
--- a/changelog/fix-express-pay-pay-for-order-page
+++ b/changelog/fix-express-pay-pay-for-order-page
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Show Google Pay/Apple Pay buttons in the Pay for Order page


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7535` or branch name `fix/express-pay-pay-for-order-page` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/express-pay-pay-for-order-page) :rocket:

---

Build info: 

- Latest commit: d5e949b50797963e223095fd0248998e46e83944
- Build time: 2023-10-31 14:58:14 UTC

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -499,6 +499,11 @@ public function should_show_payment_request_button() {
 			return false;
 		}
 
+		// Order total doesn't matter for Pay for Order page. Thus, this page should always display payment buttons.
+		if ( $this->is_pay_for_order_page() ) {
```
dpaun1985: shouldn't we check if all the products are supported by the payment request button ( most of the conditions from `has_allowed_items_in_cart` method ) ?

dpaun1985 (commented):
LGTM
Screenshot 2023-10-23 at 16 08 34

I tested and works as described, however I have the question above.

includes/class-wc-payments-payment-request-button-handler.php
elazzabi: Hmmm, I think you are right! I'm unfamiliar with the logic and helpers here, but it looks like `$this->is_product_supported()` does that check right?

dpaun1985: yes. I think so.

elazzabi: Moved the check a little down as discussed.

hsingyuc (commented):
@elazzabi LGTM

@pierorocca @elizaan36 Pinging you here for the design input.  Below is what I see on the pay-for-order page with the Google pay button.  Does the button separation look good or should the button display next to the WooPay button?
Screen Shot 2023-10-27 at 4 10 21 PM

includes/class-wc-payments-payment-request-button-handler.php
hsingyuc: I agree with @elazzabi, moving the check down LGTM.

elazzabi:
Thank you @hsingyuc for looking! I agree that it should probably show close to the WooPay button. I'll take a look and see why it's showing there

pierorocca:
> Thank you @hsingyuc for looking! I agree that it should probably show close to the WooPay button. I'll take a look and see why it's showing there

Most likely because we were under the assumption Pay for Order didn't support GPay and Apple Pay so we chose a button location based what we felt was the optimal location - after the invoice, and before the payment form. That pattern best matches the pattern found on cart and checkout.

pierorocca:
For clarity the GPay and Apple Pay button should render below the WooPay button.

dpaun1985 (approved):
✅ LGTM

elazzabi:
@pierorocca @elizaan36 I had some problems replicating this, but with help from @hsingyuc I got it working.

So the problem is the Woo Pay button is showing only if WooCommerce Blocks is installed (at least in the Pay for Order page). This means, if a user doesn't have WooCommerce Blocks, it will only show Apple/Google Pay at the top. If they have the WC Blocks, it will show Apple/Google Pay at the top, and Woo Pay and the button like @hsingyuc shared ^.

I've taken a look into the code and it still seems unclear to me why the handling is different for this page. So I'll need some help from folks working on WooCommerce Blocks to help with that.

That being said, is this issue out of the scope of this PR, or would you like this to be fixed part of this PR? (Also, I'll be on a long AFK, so maybe someone else should pick it over if we decide to go with the second option)

pierorocca:
Thanks Ahmed. We definitely cannot go to market with a split scenario where buttons are showing up in multiple locations. Additionally I'd argue the existing button location at the top is not ideal placement. So we'd want 1) all buttons after the invoice 2) consistent behavior regardless if the merchant has shortcode or block checkout.

elizaan36:
+1 to what Piero [shared here](https://github.com/Automattic/woocommerce-payments/pull/7535#issuecomment-1785688005).

elazzabi:
I agree, we should have everything in one place to deliver a good UX.

I have a proposal for a solution, but I'll need to check with other folks to make sure I'm not breaking something else 😅 

Screenshot 2023-10-31 at 10 43 02

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -381,8 +381,6 @@ public function display_pay_for_order_page_html( $order ) {
 		];
 
 		wp_localize_script( 'WCPAY_PAYMENT_REQUEST', 'wcpayPaymentRequestPayForOrderParams', $data );
-
-		$this->display_payment_request_button_html();
```
elazzabi, line 385: For future eyes: This is removed from here because it injects a div in the page with the same id expected for Apple/Google Pay creating the button in the top of the page instead of close to WooPay and other express payments.

hsingyuc (approved):
Looks good to me and tested well.

[I created a follow-up PR ](https://github.com/Automattic/woocommerce-payments/pull/7594)to restructure the pay-for-order check for express buttons so that it works with older WC Blocks versions.

elazzabi:
Thank you for testing, reviewing, and creating a follow-up change @hsingyuc 🙇 

Given this has enough reviews now, I'll go ahead and merge it.

PR #7547, opened 2023-10-25
RPP - Add phone number check

### LINKED ISSUE #7412: RPP / Verification / Phone number check
### Description

Carry over the phone number verification to the new payment process.

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L850-L854

### Acceptance criteria

- There is a new method in an `src` service.
- 100% tests coverage.

### Testing instructions

Checkout with a phone number that is too long should fail.

### Additional context

Project: paJDYF-9hL-p2

htdat:
Fixes #7412

- Add the check of phone number before processing anything else in `InitialState`. 
- Add tests to cover 100% for new code. 

#### Testing instructions

* Try to check out an order.
* In the billing phone, choose this number `242342342324234234232423423423` or any number with length > 20. 
* See this error message in the checkout page: 

Markup on 2023-10-25 at 22:59:07

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7547` or branch name `rpp/7412-validate-phone-number` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7412-validate-phone-number) :rocket:

---

Build info: 

- Latest commit: 605ec24d8605ee310a795a76358d1d9d6f3670ad
- Build time: 2023-10-26 16:53:16 UTC

htdat:
Ready for review!

changelog/rpp-7412-validate-phone-number
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: add
+
+Validate order phone number to InitialState (RPP).
```
RadoslavGeorgiev: I guess that's either "add phone number validaiton to" or "validate in"?

src/Internal/Payment/PaymentContext.php
```diff
@@ -281,6 +281,27 @@ public function get_intent(): ?WC_Payments_API_Abstract_Intention {
...
 	}
 
+	/**
+	 * Setter for the payment error message.
+	 *
+	 * @param  string $message Error message.
+	 *
+	 * @return void
+	 */
+	public function set_error_message( string $message ): void {
```
RadoslavGeorgiev: All the necessary data to determine the context of the error, and why it happened, should already be in the context. Instead of storing errors within it, let's just let the exception bubble up.

src/Internal/Payment/State/InitialState.php
```diff
@@ -194,6 +199,28 @@ protected function populate_context_from_order() {
...
+	 * @throws Order_Not_Found_Exception
+	 * @throws StateTransitionException
+	 * @throws ContainerException
+	 */
+	protected function process_order_phone_number(): ?PaymentErrorState {
+		$context  = $this->get_context();
+		$order_id = $context->get_order_id();
+
+		if ( ! $this->order_service->is_valid_phone_number( $order_id ) ) {
+			$context->set_error_message(
```
RadoslavGeorgiev: The exception handling for intent requests in `start_processing` was just an excuse to use the state, and show it can be generated, but it will probably not remain this way, and is a bad example. Right now, it does not make much sense to add all the additional logic to transition to a new payment state, store the message, etc, because those would not be persisted anyway. An exception, thrown here, can simply bubble up to the gateway without much logic required, and would achieve the exact same result. Error handling additions will be managed as a part of a future stream.

For now, let's throw an exception here immediately, and remove the `if()` from `start_processing`.

RadoslavGeorgiev (changes requested):
Great work so far! I'm glad to see a 🟢 PR right from the start.

changelog/rpp-7412-validate-phone-number
htdat: Good point. Addressed c403f9a64de567b42d7ce4883595915cf5f94370

src/Internal/Payment/PaymentContext.php
htdat: I am living a bit in the future here and thought that we eventually need to set error message to PaymentContext, similar to what PaymentIntent has - `last_payment_error`
https://github.com/Automattic/woocommerce-payments/blob/ea5370017345fc2254f5011e4fdf381b3842e18a/includes/wc-payment-api/models/class-wc-payments-api-payment-intention.php#L95

But yeah, for now, we do not need it. Just throwing an exception is good enough.

src/Internal/Payment/State/InitialState.php
htdat: Agreed with more explanation above https://github.com/Automattic/woocommerce-payments/pull/7547#discussion_r1372856077. 

For now, I will follow your suggestion. Just have done it here 4c635e0a3437349b2ac8bc03bd4e9ffa069431df

src/Internal/Payment/PaymentContext.php
RadoslavGeorgiev: We'll probably have something similar, but it would be in the payment processing service, and likely when we start storing the context. Something like:

```php
try {
	return $initial_state->start_processing( $request );
} catch ( Exception $e ) {
	$state = $this->state_factory->createState( $e->is_payment_error() ? PaymentErrorState::class : SystemErrorState::class );
	$state->set_context( $context );

	// Save the state and context.
	$this->storage->save( $context, $state );

	// Log it.
	$this->logger->log( $logs );

	throw $e;
}
```

But until then, errors go straight to the gateway.

src/Internal/Payment/State/InitialState.php
```diff
@@ -194,6 +196,28 @@ protected function populate_context_from_order() {
...
+		$context  = $this->get_context();
+		$order_id = $context->get_order_id();
+
+		if ( ! $this->order_service->is_valid_phone_number( $order_id ) ) {
+			throw new StateTransitionException(
+				__(
+					'Please enter a valid phone number, whose length is less than 20.',
+					'woocommerce-payments'
+				)
+			);
```
RadoslavGeorgiev, line 217: While all exceptions within states are state transition exceptions, we should have a specific exception for this. We'll likely also need to have exceptions declare whether they are system or payment errors. But all of that will be handled later. The PR is good for now.

RadoslavGeorgiev (approved):
LGTM! :shipit:

PR #7545, opened 2023-10-24
fire an event when disputed order notice is shown

haszari:
Fixes #7437

In progress. The event is working but I'll need to refactor things to fire it in a mount effect, once only on render. The current component is doing a lot (including… some jQuery style DOM manipulation as a side effect!) so the refactor is worthwhile IMO. 

- fire a new event with relevant props
- refactor code to ensure props are available
- store isInquiry in a local variable to save multiple calls (not sure if this is necessary?)
- move `click` event name inline - keep name close to use for clarity
- remove parseInt for click event countdownDays – seems unnecessary

Fixes #

#### Testing instructions

*

-------------------

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7043, opened 2023-08-22
Disable refund button when transaction disputed

shendy-a8c:
Fixes https://github.com/Automattic/woocommerce-payments/issues/6378

#### Changes proposed in this Pull Request

- Disable refund button when there is active or lost dispute.
- Show explanation of to why refund button is disabled:

Untitled

I refactored `DisputeNotice` a bit so it can show the explanation when refund button is disabled.
Follow the commits in this PR to understand the diff better.

When dispute is closed (status `warning_closed`), [per Stripe's docs](https://stripe.com/docs/disputes/how-disputes-work#:~:text=If%20an%20inquiry%20has%20been%20open%20for%20120%20days%20without%20escalation%20to%20a%20chargeback%2C%20Stripe%20marks%20it%20as%20closed%20in%20the%20Dashboard%20and%20API.%20At%20this%20point%2C%20you%20can%20be%20confident%20the%20card%20network%20won%E2%80%99t%20escalate%20it%E2%80%94they%20don%E2%80%99t%20provide%20an%20explicit%20%E2%80%9Cwin%E2%80%9D%20message%20for%20inquiries.), looks like refund should be allowed although I cannot test this for real. I do not disable the refund button in such case. Worst case, if Stripe does not allow refund, user will see the error.

Screenshot 2023-07-27 at 09 14 41

#### Testing instructions

Either create transaction with different state of disputes or alter the dispute object [here](https://github.com/Automattic/woocommerce-payments/blob/5be7cb9e3485cf59915a3a562dc9305e51c5a63a/client/order/index.js#L138). I tested by altering the dispute object [here](https://github.com/Automattic/woocommerce-payments/blob/5be7cb9e3485cf59915a3a562dc9305e51c5a63a/client/order/index.js#L138), eg:
```
dispute.status = 'won';
```

- When there is no dispute:
Dispute notice should not show. Refund button should not be disabled.

- When there is a dispute needing a response:
Warning text and CTA button should show. Refund button should be disabled and there should be the explanation of to why the refund button is disabled.

Screenshot 2023-08-23 at 01 57 23

- When there is a dispute that does not need a response, eg when its status is `under_review`, or when due date has passed:
Warning text and CTA button should not show. Refund button should be disabled and there should be the explanation of to why the refund button is disabled.

Screenshot 2023-08-23 at 02 00 46

- When there is a dispute won / closed (status `warning_closed`):
Warning text and CTA button should not show. Refund button should not be disabled.
[...]

Issue #6354, opened 2023-05-23
Epic: WCPay Disputes – Highlight active dispute in Woo order details screen


haszari:
### Description
The goal of this epic is to improve visibility of active disputes in the Woo orders screen, by adding a new notice (and improving clarity of the order note).

<img width="978" alt="Screenshot 2023-05-23 at 1 08 42 PM" src="https://github.com/Automattic/woocommerce-payments/assets/4167300/c5dae493-6736-478d-a2c7-5d1a8ca5f775">

![order-details-screen](https://github.com/Automattic/woocommerce-payments/assets/3147296/770b9d1f-c74c-4388-8670-16e2347a63c5)

⚠️ **Note** – the wording of the notice/note in the screenshot is out of date. Updated wording below.

* Colour and wording:
	* \> 7 days – 🟡 Yellow, `This order has a chargeback dispute of {AMOUNT} labeled as {REASON}. Please respond to this dispute before {DEADLINE}`
	* < 7 days – 🟡 Yellow, `Please resolve the dispute on this order for {AMOUNT} labeled {REASON} by {DEADLINE}. ({N} days left)`
	* < 72hours – 🔴 Red, `Please resolve the dispute on this order for {AMOUNT} labeled {REASON} by {DEADLINE}. ({N} days left)`
	* < 24hours – 🔴 Red, `Please resolve the dispute on this order for {AMOUNT} labeled {REASON} by {DEADLINE}. (Last day today)`

### Steps
- [x] Clearer & more informative order note when order is disputed - see prototype / WIP PRs:
  - https://github.com/woocommerce/woocommerce/pull/38443
  - #6535
  - #6373
- [x] #6379 
- [x] Register Tracks events
	- [x] #6742
- [x] Add testing instructions to this description for impacted flows

The above is a high-level summary of the changes in the design. These can be split off as separate issues. Can add / adapt if details change.

### Additional context (internal links)
- Project thread: pdjTHR-2Hv-p2

## Testing flows

- Install and activate WC 7.9+
- Set up a dispute by purchasing with a [dispute test card](https://stripe.com/docs/testing#disputes).
- View the order edit page and confirm the alert shows correct info and button links to dispute detail. 
- Try various combinations of dispute amount, date, and store settings (e.g. multicurrency) to ensure the notice appears when necessary and renders correctly.
- Ensure the notice is not displayed if the dispute has been lost or otherwise resolved, or the merchant has already submitted evidence (responded).

[...]

PR #7557, opened 2023-10-26
refactor disputed order notice to separate side effects & add a `_view` tracks event

### LINKED ISSUE #7437: Fire a `order dispute notice render` tracks event so we can measure click-through rate for disputed order notice
### Describe the bug
In #6354 and particularly #6373 we implemented a new notice in the WooCommerce core order screen, to alert the merchant that the buyer has disputed the payment.

We want to know how effective this notice is at driving action, and ultimately, improving response to disputes. 

However, we don't have an event for when the notice is shown. This means our conversion funnel starts earlier (e.g. view order screen (any type of order) or view order screen (order is `on-hold` due to dispute or various other reasons).

The goal of this issue is to add an event when the disputed order notice is rendered. Then we can measure the % of notices that merchants click on (to view or respond to dispute), and ultimately better understand how this notice helps merchants.

##### Option 1: use generic notice event(s)
There may be a generic `inline notice` event already firing. If so, need to check this will work for our needs (e.g. has a prop to filter only dispute notices). 

##### Option 2: fully custom event for our notice only
We can fire an event just as the notice is rendered, e.g. here:

https://github.com/Automattic/woocommerce-payments/blob/703835a096cb03ed8b7595b9d13059d1dabae1d9/client/order/index.js#L332-L347

Might be worth adding some props for the notice colour / urgency.

haszari:
Fixes #7437

- ~~alternative to #7545~~ now closed
- see comment on issue for rationale https://github.com/Automattic/woocommerce-payments/issues/7437#issuecomment-1780237772

Please also review the registration PR: Automattic/tracks-events-registration#1921 (private repo)

- Refactor disputed order notice code to simplify logic. 
- Move notices to a dedicated module, away from the old school jQuery code (simplify the main file).
- Use explicit side effects (`useEffect`) for:
  - manipulating DOM to disable refund button
  - firing a tracks event for notice render `wcpay_order_dispute_notice_view`
- Removes the info about why refund is disabled from the notice. This clutters the notice and may be distracting / confusing for merchants. The purpose of the notice is to drive merchant to respond to dispute.

#### Testing instructions
Test viewing order page with all dispute scenarios: 

- no dispute
- dispute with various days remaining
- inquiry with various days remaining
- dispute with various status (pending, lost, won)
 
Notice should work as before (#6354) and refund behaviour should be disabled appropriately when order is disputed (#7043).

When the order is locked because of a pending dispute, the merchant should have a clue about why – tooltip at bottom of order table should be relevant. The dispute notice should focus on the dispute and the fact that merchant should take action.

When the order has a resolved or pending dispute, there should be a notice explaining that the order is locked and that this is due to the dispute.

-------------------

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_ na - behaviour is unchanged
- [x] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable. na - no change to flows
- [x] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable. na


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7557` or branch name `fix/7437-refactor-order-dispute-side-effects` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7437-refactor-order-dispute-side-effects) :rocket:

---

Build info: 

- Latest commit: 0877c19b9472cf6b53d6423a48747a098b7d70fb
- Build time: 2023-10-31 22:35:48 UTC

client/order/index.js
```diff
@@ -135,221 +135,209 @@ jQuery( function ( $ ) {
...
+	disputeReason,
+	formattedAmount,
+	isPreDisputeInquiry,
+	dueBy,
+	countdownDays,
+	disputeDetailsUrl,
+} ) => {
+	useEffect( () => {
+		wcpayTracks.recordEvent( 'wcpay_order_dispute_notice_view', {
+			is_inquiry: isPreDisputeInquiry,
```
Jinksi: What do you think about providing `dispute_status` rather than `is_inquiry`, in order to capture `under_review`.

I see that this PR has removed the notice for `under_review`, so this may not be required, although it's not clear.

Jinksi (commented):
This refactor is an improvement and the code looks good overall.

I've tested the following and I think this PR might be ready to merge, but I'm unsure if this PR intends to make the changes to the `Under review` disputed order notice (the PR description mentions "Notice should work as before").

| Dispute status         | develop                      | this branch                                      |
|------------------------|------------------------------|--------------------------------------------------|
| Lost                   | No notice / no refund button | No notice / no refund button ✅                   |
| Needs response         | Notice / refund disabled     | Notice (removed refund text) / refund disabled ✅ |
| Under review           | Notice / refund disabled     | No notice ❌ / refund disabled ✅                  |
| Won                    | No notice / refund enabled   | No notice / refund enabled ✅                     |
| Inquiry needs response | Notice / refund enabled      | Notice / refund enabled ✅                        |
| Inquiry closed         | No notice / refund enabled   | No notice / refund enabled ✅                     |


As a side note, I've discovered that the tooltip text is not rendered for `Under Review` disputes – this may be a separate bug, but I'm unsure if this PR intends to remove all communication about disabling refunds or not.

**Under review** notice on `develop`:
image


IMO this PR could benefit from splitting up some of these changes into smaller stacked PRs to help ease the cognitive load of reviewers and clarify the intent of each code change, i.e.:
1. Refactor the disputed order notice.
2. Add view disputed order notice tracks event.
3. Remove disputed order notice text about refunds / remove Under Review notices.

shendy-a8c:
> IMO this PR could benefit from splitting up some of these changes into smaller stacked PRs to help ease the cognitive load of reviewers and clarify the intent of each code change, i.e.:
> 
> Refactor the disputed order notice.
> Add view disputed order notice tracks event.
> Remove disputed order notice text about refunds / remove Under Review notices.

I agree to have separate PRs for each will make it easier to review.

client/order/index.js
haszari: I suspect I will need to do this to fix the "dispute under review" scenario – thanks for the tip 🙌🏻

haszari:
Thanks for the feedback @Jinksi @shendy-a8c – good catalyst to make the refactor more readable / decoupled. Ready for more reviews (though I am still testing).

- The `under review` case is now working correctly.
- I have refactored the components to a separate module. No particular reason, but might help with reading the code, reviewing. Also might help us run more tooling on this code (… I'm wondering if eslint react is running somewhere, which picks up rules of hooks etc) … or migrating to TypeScript. 
- Components have new names that I think better reflect the purpose (from merchant point of view). 

I still need to test urgency colours - should be unchanged but I need to ensure that a few cases are tested.

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+import {
+	isAwaitingResponse,
+	isInquiry,
+	isUnderReview,
+} from 'wcpay/disputes/utils';
+import { useCharge } from 'wcpay/data';
+import wcpayTracks from 'tracks';
+import './style.scss';
+
+const DisputedOrderNoticeHandler = ( { chargeId, onDisableOrderRefund } ) => {
```
haszari, line 23: Renamed this – it's a "worker", handle ssome logic to potentially show notices and/or disable UI elements. 

Takes a callback to handle the disabling (this is not ideal, but at least is explicit).

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+	const { dispute } = charge;
+	if ( ! charge?.dispute ) {
+		return null;
+	}
+
+	// Special case the "under review" notice which is much simpler.
+	// (And return early.)
+	if ( isUnderReview( dispute.status ) ) {
+		return (
+			<DisputeUnderReviewNotice disputeDetailsUrl={ disputeDetailsUrl } />
```
haszari: Single purpose notice component – this one was easy to name 😁

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+	const now = moment();
+	const dueBy = moment.unix( dispute.evidence_details?.due_by );
+
+	// If the dispute is due in the past, don't show notice.
+	if ( ! now.isBefore( dueBy ) ) {
+		return null;
+	}
+
+	return (
+		<DisputeNeedsResponseNotice
```
haszari, line 100: I better understand the role of this component now. It is still quite complex, but the common thing for all the different flavours is that we want the merchant to click `Respond now`.

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+			countdownDays
+		);
+	}
+
+	const title = sprintf(
+		titleText,
+		formattedAmount,
+		reasons[ disputeReason ].display,
+		dateI18n( 'M j, Y', dueBy.local().toISOString() )
+	);
```
haszari, line 186: The logic for generating the copy is still a little convoluted and complex, but leaving as is since that code is well tested, keep the refactor contained to the most pressing issue (manage side effects).

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+							{
+								due_by_days: countdownDays,
+							}
+						);
+						window.location = disputeDetailsUrl;
+					},
+				},
+			] }
+		>
+			{ <strong>{ `${ title } ${ suffix }` }</strong> }
```
haszari, line 213: We could have another look at this if/when we work on #7568.

- #7568

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+				),
+				{
+					// createInterpolateElement is incompatible with this eslint rule as the <a> is decoupled from content.
+					// eslint-disable-next-line jsx-a11y/anchor-has-content
+					a: <a href={ disputeDetailsUrl } />,
+				}
+			) }
+		</InlineNotice>
+	);
+};
```
haszari, line 232: The joy of a simple component that just does what it says, and can be easily used whereever!

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,210 @@
...
+		<InlineNotice status="warning" isDismissible={ false }>
+			{ createInterpolateElement(
+				__(
+					'This order has an active payment dispute. ' +
+						'Refunds and order editing are disabled. <a>View details</a>',
+					'woocommerce-payments'
+				),
+				{
+					// createInterpolateElement is incompatible with this eslint rule as the <a> is decoupled from content.
+					// eslint-disable-next-line jsx-a11y/anchor-has-content
```
haszari, line 226: This is my crack at a standard rationale for why we ignore this rule.

client/order/index.js
```diff
@@ -1,30 +1,58 @@
...
+	const refundButton = document.querySelector( 'button.refund-items' );
+	if ( ! refundButton ) {
+		return;
+	}
+
+	refundButton.disabled = true;
+
+	// Show helpful info in order edit lock icon tooltip.
+
+	let tooltipText = '';
```
haszari, line 27: Could probably refactor the help icon tooltip into another function …

client/order/index.js
```diff
@@ -135,221 +163,12 @@ jQuery( function ( $ ) {
 			return;
 		}
 
-		ReactDOM.render( <DisputeNotice chargeId={ chargeId } />, container );
+		ReactDOM.render(
+			<DisputedOrderNoticeHandler
```
haszari, line 167: Curious to see if this name works for y'all in context here.

client/order/index.js
```diff
@@ -135,221 +163,12 @@ jQuery( function ( $ ) {
...
+				chargeId={ chargeId }
+				onDisableOrderRefund={ disableWooOrderRefundButton }
+			/>,
```
haszari, line 170: Curious to see if this name (and the props interface) works for y'all in context here.

client/order/index.js
```diff
@@ -135,221 +163,12 @@ jQuery( function ( $ ) {
...
-			status={ urgency }
-			isDismissible={ false }
-			actions={ actions }
-		>
-			{ showWarning && <strong>{ warningText }</strong> }
-
-			{ disableRefund && <div>{ refundDisabledNotice }</div> }
-		</InlineNotice>
-	);
-};
```
haszari, line 355: 💣

client/order/index.js
haszari: Actually, can render that simple case a whole lot earlier.

haszari:
I discovered that there's now a missing "this order is not editable because of what happened with a dispute" notice on this branch, vs develop. So adding that back in.

refund disabled  notice is shown on develop for past-due dispute (gone in branch)

Dispute lost  refund disabled  notice MIA on branch

client/components/disputed-order-notice/index.js
haszari: ~~The joy of a simple component that just does what it says, and can be easily used whereever!~~

Is now slightly more complex … but I'm ok with that as the purpose is clear: explain to the merchant why the order is not editable.

client/order/index.js
```diff
@@ -1,30 +1,58 @@
...
+
+	let tooltipText = '';
+	if ( isAwaitingResponse( disputeStatus ) ) {
+		tooltipText = __(
+			'Refunds and order editing are disabled during disputes.',
+			'woocommerce-payments'
+		);
+	} else if ( isUnderReview( disputeStatus ) ) {
+		tooltipText = __(
+			'Refunds and order editing are disabled during an active dispute.',
```
haszari, line 35: Note I tweaked this wording to remove redundant "at this time". 

Screenshot 2023-10-30 at 2 33 24 PM

client/components/disputed-order-notice/index.js
```diff
@@ -0,0 +1,225 @@
...
+	if ( ! charge?.dispute ) {
+		return null;
+	}
+
+	// Special case the "under review" notice which is much simpler.
+	// (And return early.)
+	if ( isUnderReview( dispute.status ) ) {
+		<DisputeOrderLockedNotice
+			message={ __(
+				'This order has an active payment dispute. Refunds and order editing are disabled.',
```
haszari: Note I tweaked this wording to remove redundant "at this time". 

Screenshot 2023-10-30 at 2 33 24 PM

haszari: Was: `This order has an active payment dispute. Refunds and order editing are disabled at this time.`

haszari:
I've tested a range of cases now and am happy with this (pending review feedback / testing). 

The one thing I haven't been able to test is different urgency levels. I hacked `daysRemaining` prop to check this code is firing and has no issues, but I don't have any disputes that are old enough and still open. 

**If anyone has disputes with 3-1 days to respond on their store**, I'd really appreciate some extra testing. E.g. view order page for an order with a dispute with < 3 days remaining, check notice is rendered with appropriate urgency!

haszari:
Side note – do we have [eslint react hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) running? During development I saw a bunch of runtime hooks errors, which possibly can be caught by linter.

client/order/index.js
```diff
@@ -1,30 +1,58 @@
...
-import { getDetailsURL } from 'wcpay/components/details-link';
-import {
-	isAwaitingResponse,
-	isInquiry,
-	isUnderReview,
-} from 'wcpay/disputes/utils';
-import { useCharge } from 'wcpay/data';
-import wcpayTracks from 'tracks';
-import './style.scss';
+import DisputedOrderNoticeHandler from 'components/disputed-order-notice';
```
shendy-a8c: I remember @Jinksi usually would suggest that we prefix the import path with `wcpay/`.

```suggestion
import DisputedOrderNoticeHandler from 'wcpay/components/disputed-order-notice';
```

shendy-a8c (changes requested):

shendy-a8c:
Looks great but there is something needs fixing, @haszari.

Foremost, I like to mention that there are times where in the notice we show that order is not editable and refunds are disabled and yes the refund button is disabled but order is still editable.

For example, when dispute status is `under_review`, we show
Screenshot 2023-10-31 at 01 02 12

but order is editable, since status is `on-hold` ([ref](https://github.com/woocommerce/woocommerce/blob/d27c3941ea23c449c4e8066c642c54f32af77068/plugins/woocommerce/includes/class-wc-order.php#L1588)), so there is no `This order is no longer editable.`, so no tooltip either.

Screenshot 2023-10-31 at 01 57 21


I think that's OK for now as we are aware that we would start by disabling the refund button for now even though we would eventually want to disable order editing in the future.

ps: to make order not editable, we might just need to add [`wc_order_is_editable` filter](https://github.com/woocommerce/woocommerce/blob/d27c3941ea23c449c4e8066c642c54f32af77068/plugins/woocommerce/includes/class-wc-order.php#L1588) but since it depends whether order is disputed, we need to do a fetch in PHP, which might slow down order edit page load, but that's OK if we only do it for disputed order, which is related to https://github.com/Automattic/woocommerce-payments/pull/7174. Adding dispute id to the order meta might be the cleanest way to mark if order is disputed.

shendy-a8c:
I did these tests. The ❌'s need fixing.

When dispute status is:

- `warning_needs_response`

✅ Notice shows.
Screenshot 2023-10-31 at 01 40 53
✅ Refund button is not disabled.

- `warning_under_review`

❌ Notice should not show. In `develop`, no notice shows but in this PR branch, this notice shows:
Screenshot 2023-10-31 at 01 45 17
✅ Refund button is not disabled.

- `warning_closed`

✅ No notice.
✅ No refund button.

- `needs_response`

✅ Notice shows.
Screenshot 2023-10-31 at 01 31 01
✅ Refund button is disabled. No tooltip though because order is editable, so tooltip and `This order is no longer editable.` text are not rendered. See my comment https://github.com/Automattic/woocommerce-payments/pull/7557#issuecomment-1785860344.

- `under_review`

✅ Notice shows.
Screenshot 2023-10-31 at 01 02 12 (1)
✅ Refund button is disabled. No tooltip though because order is editable, so tooltip and `This order is no longer editable.` text are not rendered. See my comment https://github.com/Automattic/woocommerce-payments/pull/7557#issuecomment-1785860344.

- `won`

✅ No notice.
✅ No refund button.

- `lost`

❌ Notice should not show. In `develop`, no notice shows but with PR branch, this notice shows up:
Screenshot 2023-10-31 at 01 35 55
✅ No refund button, so message in notice is conflicting.

shendy-a8c:
I can confirm that tracks are firing correctly, ie
`wcadmin_wcpay_order_dispute_notice_view` when dispute notice is showing and `wcadmin_wcpay_order_dispute_notice_action_click` when the CTA button is clicked.

haszari:
Thanks for testing @shendy-a8c !

> incorrect behaviour for dispute `lost`
> ❌ Notice should not show. In develop, no notice shows but with PR branch, this notice shows up: `Refunds and order editing have been disabled as a result of a lost dispute.`

As I understand it, this is correct behaviour. If the merchant loses the dispute, they won't want to refund – they will lose 2x the revenue. So order is locked and we show the (sad) notice so it's clear why.

I tested this on develop and saw the same notice too actually. If you think this behaviour's not quite right, let me know and I can retest (maybe I have a different scenario).

> incorrect behaviour for `warning_under_review` - inquiry which merchant has challenged (and bank is considering)
> ❌ Notice should not show. In develop, no notice shows but in this PR branch, this notice shows: `This order has an active payment dispute. Refunds and order editing are disabled.`

🧤 Good catch! In this scenario, it's just a regular order that the customer or bank feels iffy about. The merchant should be able to refund as normal. I've fixed this by tightening up the if conditions - only show this for actual disputes (not inquiries) that are under review. 4792d3362d438964f99bb76824a9d4e608181839

> when dispute status is under_review, we show `This order has an active payment dispute. Refunds and order editing are disabled.`

As you've noted in your comment, our current implementation of `locked order` or `disable order editing` is piecemeal – there are gaps. This is something we'd need more work to improve, likely involving core changes. There's an issue open for this potential future maintenance:

- #7567 

In the meantime, the notice and the disabling will help _remind_ merchant not to fiddle with disputed orders.

haszari:
Some of my test disputes are old enough, so I've now also confirmed at least one `urgency` case: 

Screenshot 2023-10-31 at 2 09 56 PM

I think we're good to merge! @shendy-a8c @Jinksi let me know if you have any concerns or feedback.

client/order/index.js
haszari: @Jinksi any ideas if there are tools we can use to standardise our imports (order, relative, roots etc)? Would be cool to get a robot on that.

haszari: PS I like `wcpay/components` as it disambiguates which components we are talking about (vs `@wordpress` and `@woocommerce`).

Jinksi: > any ideas if there are tools we can use to standardise our imports

I'm not sure about this one. We could change our config to remove the [`wcpay/` alias](https://github.com/Automattic/woocommerce-payments/blob/4e872a9e1bce8729f2eae97be1c131df555c17a8/webpack/shared.js#L115-L117) or remove each [`client/subdir` alias](https://github.com/Automattic/woocommerce-payments/blob/4e872a9e1bce8729f2eae97be1c131df555c17a8/webpack/shared.js#L114).

> I like wcpay/components as it disambiguates which components we are talking about

There was a discussion in Slack and it seems most were in favour of omitting `wcpay/`, although it was a small sample size: p1696552621397799-slack-C053VQPD74J.

haszari: 🤷🏻 I guess this is pretty minor so let's not worry about it too much

Jinksi:
> Side note – do we have [eslint react hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) running? During development I saw a bunch of runtime hooks errors, which possibly can be caught by linter.

Checking.... yes. I see both `react-hooks/exhaustive-deps` and `react-hooks/rules-of-hooks` eslint errors when I violate the rules.

https://github.com/Automattic/woocommerce-payments/blob/ad935b02ddb234d48441817b80477648b4c36020/.eslintrc#L89-L90


image

image

haszari:
> Checking.... yes. I see both react-hooks/exhaustive-deps and react-hooks/rules-of-hooks eslint errors when I violate the rules.

Nice! Do I need to install / configure this (in VScode in my case) or does it come for free with the repo/environment? Would be good to confirm we're all running the best tooling (or intentionally not, vs. accidental!). In the meantime I'm installing [ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) … I bet this will help.

haszari:
> In the meantime I'm installing [ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) … I bet this will help.

✅ … and now I get hooks rules errors in my IDE too.

shendy-a8c:
> As I understand it, this is correct behaviour. If the merchant loses the dispute, they won't want to refund – they will lose 2x the revenue. So order is locked and we show the (sad) notice so it's clear why.
> 
> I tested this on develop and saw the same notice too actually. If you think this behaviour's not quite right, let me know and I can retest (maybe I have a different scenario).

~Did you build client after switching branch, ie `npm run build:client` or `npm run watch`?~ We discussed this just now that you did build client after switching branch.

That's weird. I do *not* see a notice when I'm on `develop`.

Screenshot 2023-10-31 at 09 25 04

It should not show a notice since this [if condition](https://github.com/Automattic/woocommerce-payments/blob/ad935b02ddb234d48441817b80477648b4c36020/client/order/index.js#L232-L236) won't be satisfied when dispute status is `lost.

I have made sure with a JN site too. Read p1698719942987859-slack-C02BW3Z8SHK.

Anyway, I'm just pointing out the behavior discrepancy between this PR branch and `develop` but since the order status is `Refunded` when dispute is lost, having that notice is OK. In a way, it's better because merchant has more context and can go to see the dispute details.

On a side note, a bit random, this reminds me of this issue we discussed last week.

https://github.com/Automattic/woocommerce-payments/issues/6196

shendy-a8c (approved):
> I've fixed this by tightening up the if conditions - only show this for actual disputes (not inquiries) that are under review. https://github.com/Automattic/woocommerce-payments/commit/4792d3362d438964f99bb76824a9d4e608181839

Looks good.

Other than the discussion about the notice showing when dispute status is `lost`, which I think we all agree that it's OK, this PR is good to merge.

haszari:
> Other than the discussion about the notice showing when dispute status is lost, which I think we all agree that it's OK, this PR is good to merge.

I don't agree this is ok 😁 so I'm fixing it too. Now this notice is only shown if **WCPay** disables refund button due to dispute. d9e0bdd23dea99292bd5672e69edfbd13377cbde

Why?

[The `charge.dispute.closed` webhook handler refunds the order](https://github.com/Automattic/woocommerce-payments/blob/b011b37dd8c7bb16ef7c75d12ad681f2d3585135/includes/class-wc-payments-order-service.php#L311-L331). That shows up in the order UI and it's confusing / incorrect to say we have disabled refunds.

Screenshot 2023-11-01 at 11 25 56 AM

https://github.com/Automattic/woocommerce-payments/blob/b011b37dd8c7bb16ef7c75d12ad681f2d3585135/includes/class-wc-payments-order-service.php#L311-L331

--- 

I've left the notice code in place in the PR. However I believe it will rarely (never?) show, since the dispute lost status will be closely followed by the webhook-powered refund †

https://github.com/Automattic/woocommerce-payments/blob/d9e0bdd23dea99292bd5672e69edfbd13377cbde/client/components/disputed-order-notice/index.js#L62-L79

† – in most stores/environments - on dev sites webhook handling may be impacted, it is on my site. I need to manually run action scheduler actions (`wcpay_webhook_fetch_events` `wcpay_webhook_process_event`) to process webhooks.

PR #7570, opened 2023-10-27
Display express payment buttons on checkout blocks pay-for-order page

hsingyuc:
When the page has blocks we don't initialize the express payment buttons, so if it's a pay-for-order page but the checkout page uses blocks, we won't see the express payment buttons even though pay-for-order never shows the checkout blocks.  This PR adds the logic to display the buttons on pay-for-order pages when the checkout page is using checkout blocks.
 
[Design discussion in this PR](https://github.com/Automattic/woocommerce-payments/pull/7535).

Screen Shot 2023-10-27 at 4 10 21 PM

#### Testing instructions
1. Make sure your merchant is online so the express payment buttons (ex. Google pay button) would show
2. Use blocks checkout on your checkout page that is set under WooCommerce settings -> Advanced
3. Create a pay-for-order order
4. Go to the pay-for-order page and see the express payment buttons

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7570` or branch name `fix/express-payment-buttons-on-blocks-pay-for-order-page` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/express-payment-buttons-on-blocks-pay-for-order-page) :rocket:

---

Build info: 

- Latest commit: 8fe168047dfa17d2428320519e5c602792602188
- Build time: 2023-10-30 06:40:53 UTC

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -499,6 +499,11 @@ public function should_show_payment_request_button() {
 			return false;
 		}
 
+		// Order total doesn't matter for Pay for Order page. Thus, this page should always display payment buttons.
+		if ( $this->is_pay_for_order_page() ) {
+			return true;
+		}
+
```
malithsen: nit: I feel like it would be more appropriate to update the logic in line 541 to exclude pay-for-orders instead of a top-level `return true`. Excluding the pay-for-order page in line 541 would make it easier to reason out the logic.

It will also match the existing pattern of this function, where we bail out if conditions are met and only return true at the very end.

That said, I don't have any strong objections to the current implementation.

Suggested change for line 541 instead of above:
```
( ! $this->is_product() && ! $this->is_pay_for_order_page() && 0.0 === (float) WC()->cart->get_total( 'edit' ) ) ||
```

malithsen (approved):
Thanks @hsingyuc This tests well. Left a comment on the order of logic but it's nothing major and you can disregard it if you'd like.

includes/class-wc-payments-payment-request-button-handler.php
hsingyuc: Good idea!  Updated here [ab2c28f](https://github.com/Automattic/woocommerce-payments/pull/7570/commits/ab2c28f9ee6203a0bd1afd6e54f2406c342d88eb).

PR #7581, opened 2023-10-30
Avoid returning Stripe error message when capturing more than order amount

### LINKED ISSUE #7029: Avoid returning Stripe error message when capturing more than order amount
If I try to capture an order after editing it to total to more than the original authorized amount, I get an error in the order notes that tells me to contact Stripe. This is not ideal.

Screenshot taken on 2023-08-18 at 08 02 48 UTC@2x

To reproduce:

1. WooPayments test site with an example product.
2. Under Payments > Settings, enable the authorize and capture later setting.
3. On the frontend, place an order for the example product.
4. Under WooCommerce > Orders, edit the order to add a fee, or any other way of making the order total more than it was originally.
5. Using order actions box, try to capture the order.
6. You'll get the above error telling you to contact Stripe.

This shouldn't happen. Merchants should always be directed to our support channels, but IMHO contacting support isn't really needed here. Simply displaying the error is enough:

> Error: The payment could not be captured because the requested capture amount is greater than the amount you can capture for this charge.

mgascam:
Fixes #7029 

* Redact the prompt in order notes to contact Stripe support when a merchant attempts to capture more than the authorized amount. Contacting support is unnecessary in this case.
* Introduce `Amount_Too_Large_Exception` API exception

**Before**

261541727-6f87fd67-9afa-462b-a7fa-1a693c0b2339

**After**

Screenshot 2023-11-02 at 10 11 47

#### Testing instructions

- As a merchant, enable the authorize and capture later setting under Payments > Settings.
- As a shopper, place an order
- As a merchant, edit the order to add a fee or use any other method to increase the order total beyond its original amount.
- Using order actions box, try to capture the order.
- Ensure there's no prompt advising the merchant to contact Stripe support (refer to the "After" screenshot above).

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7581` or branch name `update/7029-avoid-returning-stripe-error-message-when-capturing-more-than-order-amount` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/7029-avoid-returning-stripe-error-message-when-capturing-more-than-order-amount) :rocket:

---

Build info: 

- Latest commit: f5662c7b20b27c1a415cea596930343a8c264bca
- Build time: 2023-11-02 15:44:19 UTC

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -2785,6 +2785,11 @@ public function capture_charge( $order, $include_level3 = true, $intent_metadata
 			try {
 				$error_message = $e->getMessage();
 				$http_code     = $e->get_http_code();
+				$error_code    = $e->get_error_code();
+
+				if ( 'amount_too_large' === $error_code ) {
```
kalessil: That's a simple change, but I have mixed feelings about where it has been placed.
Are there any alternatives where this message normalization can be located?

kalessil (changes requested):

includes/class-wc-payment-gateway-wcpay.php
mgascam: Indeed @kalessil, there's `generate_capture_failed_note` method, seems a more suitable place.  https://github.com/Automattic/woocommerce-payments/blob/ea5370017345fc2254f5011e4fdf381b3842e18a/includes/class-wc-payments-order-service.php#L1142

This method is called in `mark_payment_capture_failed` https://github.com/Automattic/woocommerce-payments/blob/ea5370017345fc2254f5011e4fdf381b3842e18a/includes/class-wc-payments-order-service.php#L204

kalessil: There is also one more opportunity: https://github.com/Automattic/woocommerce-payments/blob/develop/includes/wc-payment-api/class-wc-payments-api-client.php#L1914

Can you compare 2 of them weighing the pros and cons a comment?

mgascam: > Can you compare 2 of them weighing the pros and cons a comment?

Sure! But before getting into the details please note that I already explored one option by the time I noticed your comment (https://github.com/Automattic/woocommerce-payments/pull/7581/commits/5630836f515770995ae422620efb36a816350b46) so I explored your proposal in an additional commit https://github.com/Automattic/woocommerce-payments/pull/7581/commits/d8bbe34c08708bd64c46c2c09e8c376a04b30901

### A. Modify order service (https://github.com/Automattic/woocommerce-payments/pull/7581/commits/5630836f515770995ae422620efb36a816350b46)

**Pros**

- It overrides the message in the context of capturing a charge, where the error may happen.

**Cons**

- I'm passing the `error_code` from the gateway down to the `generate_capture_failed_note` in the order service, going through `mark_payment_capture_failed`. This method does nothing with this parameter, only passes it to `generate_capture_failed_note`.

### B. Introduce Amount_Too_Large_Exception (https://github.com/Automattic/woocommerce-payments/pull/7581/commits/d8bbe34c08708bd64c46c2c09e8c376a04b30901)

**Pros**
- Centralized solution.
- Does not require changes to the current flow of capturing a charge

**Cons**
- If I am not wrong, the `amount_too_large` error is not specific to capturing charges. I'm checking that the payment intent status  is `requires_capture`  before throwing the `Amount_Too_Large_Exception`. But it feels like dealing with a specific issue of the capture flow in a more generic place such as the API client.

I'm leaning towards option B, as it seems a cleaner solution to me. What do you think?

kalessil: Well, the whole point of asking to compare the options is to double-check if you notice that option A would require a change in public method `mark_payment_capture_failed` (passing in an error code along with the message).

The problem with public methods is that they might be consumed already by in-house or custom merchant integrations.  This means it's a no-go from the beginning (for the draft, probably ok, but still a point to concern).

Option B is an isolated solution, which is flexible so yes, I think it's the best shot for our needs.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1919,6 +1920,13 @@ protected function check_response_for_errors( $response ) {
 					$response_code
 				);
 			} elseif ( isset( $response_body['error'] ) ) {
+				if ( isset( $response_body['error']['code'] ) && 'amount_too_large' === $response_body['error']['code'] && 'requires_capture' === $response_body['error']['payment_intent']['status'] ) {
+					throw new Amount_Too_Large_Exception(
+						// translators: This is an error API response.
+						__( 'Error: The payment could not be captured because the requested capture amount is greater than the amount you can capture for this charge.', 'woocommerce-payments' ),
```
kalessil, line 1931: Just looking into the code, aren't we getting `Error: Error:` in the order note?

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1919,6 +1920,13 @@ protected function check_response_for_errors( $response ) {
 					$response_code
 				);
 			} elseif ( isset( $response_body['error'] ) ) {
+				if ( isset( $response_body['error']['code'] ) && 'amount_too_large' === $response_body['error']['code'] && 'requires_capture' === $response_body['error']['payment_intent']['status'] ) {
```
kalessil: An additional check is a good one, but it needs hardening a bit. And I'd:
- move it into a dedicated local variable (so we'll get one more nested if)
- add a comment explaining that we are redacting the message and the reasoning behind

kalessil (changes requested):
It looks better, we can now polish details and reliability aspects.

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Good call, but not in this case. I added screenshot above of how the order note looks like with this change.

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Thanks @kalessil, moved error code and payment intent status to local variables and added comment in this commit  https://github.com/Automattic/woocommerce-payments/pull/7581/commits/6f9a3c446cd44af8382242bce9606a78b4a6b541

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1919,10 +1920,23 @@ protected function check_response_for_errors( $response ) {
 					$response_code
 				);
 			} elseif ( isset( $response_body['error'] ) ) {
+				$response_body_error_code = $response_body['error']['code'] ?? null;
+				$payment_intent_status    = $response_body['error']['payment_intent']['status'] ?? null;
+
+				// Check for an 'amount_too_large' error code and a 'requires_capture' payment intent status.
+				// We redact the API error message to prevent prompting the merchant to contact Stripe support
+				// when attempting to manually capture an amount greater than what's authorized. Contacting support is unnecessary in this scenario.
+				if ( isset( $response_body_error_code ) && 'amount_too_large' === $response_body_error_code && 'requires_capture' === $payment_intent_status ) {
```
kalessil: We can omit `isset( $response_body_error_code )` - it's being covered by ` ?? ` in earlier assignment.

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1919,10 +1920,23 @@ protected function check_response_for_errors( $response ) {
 					$response_code
 				);
 			} elseif ( isset( $response_body['error'] ) ) {
+				$response_body_error_code = $response_body['error']['code'] ?? null;
+				$payment_intent_status    = $response_body['error']['payment_intent']['status'] ?? null;
+
+				// Check for an 'amount_too_large' error code and a 'requires_capture' payment intent status.
```
kalessil: We can skip this specific line - if-condition stating the same =)

includes/wc-payment-api/class-wc-payments-api-client.php
kalessil: To double-check: is `elseif ( isset( $response_body['error'] ) )` the correct branch for the redacting message? Asking as I don't recall exactly if we should aim `$response_body['code']` or `$response_body['error']['code']`.

kalessil (changes requested):

includes/wc-payment-api/class-wc-payments-api-client.php
```diff
@@ -1919,10 +1920,23 @@ protected function check_response_for_errors( $response ) {
...
+				// We redact the API error message to prevent prompting the merchant to contact Stripe support
+				// when attempting to manually capture an amount greater than what's authorized. Contacting support is unnecessary in this scenario.
+				if ( 'amount_too_large' === $response_body_error_code && 'requires_capture' === $payment_intent_status ) {
```
kalessil: We a constant for this intent status in `\Intent_Status`

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Yep, you beat me to it, removed in this commit https://github.com/Automattic/woocommerce-payments/pull/7581/commits/6b36ddf58fdfce78cc5d2b78af64fa9fc88509d0

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Removed here  https://github.com/Automattic/woocommerce-payments/pull/7581/commits/3d4250aa6e97c15a36e66d7eb8b5eddfd19d6504 thanks!

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Yes, we should look into $response_body['error'] because the error has this shape:

Screenshot 2023-11-02 at 10 56 56

Thanks for double checking.

includes/wc-payment-api/class-wc-payments-api-client.php
mgascam: Using the constant here https://github.com/Automattic/woocommerce-payments/pull/7581/commits/f8ad535d33cf07ff1856a682bea48145c2c45e19 thanks!

kalessil (approved):
LGTM (code review only)!

dmvrtx (approved):
Looks good to me code-wise.

PR #7582, opened 2023-10-31
Don't load Apple/Google Pay for synced subscriptions that don't have payment due today and also require shipping

### LINKED ISSUE #3742: Synchronized Subscriptions Requiring Shipping Can't Be Purchased with Apple Pay
### Describe the bug
This is using the main WooCommerce Subscriptions extension and not WooCommerce Payments subscriptions. 

If a customer tries to purchase a synchronized subscription that requires shipping with Apple Pay, the transaction will be canceled and an "Invalid recurring shipping method" notice added.  

This works as expected for virtual subscription products.  

### To Reproduce
1.  Enable the WooCommerce Subscriptions extension on a site using WooCommerce Payments.
2. Create a simple subscription product with "Synchronise renewals" set for a day other than today.
3. Add the item to the cart and try to pay with Apple Pay.
4. Notice how the transaction is canceled and the error message appears. 

### Actual behavior
The payment can not be completed.

### Screenshots
apple-pay

invalid-shipping-method

### Expected behavior
I expect to be able to purchase a synchronized subscription with Apple Pay.

* OS: MacOS
* Browser Safari
* Version 15.2

### Additional context
4741498-zen

mattallan:
Fixes #3742

When customers attempt to use Apple Pay or Google Pay to purchase a synchronized subscription product that has no upfront costs (i.e. trial period or delayed initial payment due to product being synced) and also requires shipping, they're met with an error on the page:

image

Inside our `WC_Payments_Payment_Request_Button_Handler` class, we have code that prevents non-virtual (physical) products with free trials from being purchased with Apple/Google Pay, however there isn't any code to handle products that imitate a free trial period like what we do to handle synchronised subscription purchased outside of the synced date with no initial payment.

In this PR, I've introduced a new function to check if the product is a subscription product and has a free trial period (either trial set on the product or is a synced product), and requires shipping, and if so, prevent loading the Apple Pay/Google Pay buttons on the page.

#### Testing instructions

1. Enable Apple Pay and Google Pay payment methods
2. Install and active Woo Subscriptions
3. Enable synchronised subscriptions in **WooCommerce > Settings > Subscriptions > Synchronise renewals**
image
4. Create a synced product that is not virtual and is not synced to today:
image
5. While on the `develop` branch, view the product page and note the Payment Request buttons (PRB) are loaded. If you attempt to purchase this product with PRB, you will see the above error.
6. Check out this branch and confirm the PRB are no longer loaded on product pages for physical synced subscription products that

Other test cases to run:
 - Physical (non-virtual) variable synced subscriptions. The PRB shouldn't show on the product page if any variation isn't supported.
 - Virtual synced subscriptions. The PRB should show.
 - Physical synced products with a sync date set to today. The PRB should show
 - Physical synced subscription that is synced to another day but proration is enabled in **WC > Settings > Subscriptions**

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7582` or branch name `fix/3742-synced-subscriptions-payment-requests` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/3742-synced-subscriptions-payment-requests) :rocket:

---

Build info: 

- Latest commit: d9cc5617ac1a21495041c1c897cfea260853388c
- Build time: 2024-02-06 05:40:08 UTC

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1700,4 +1700,44 @@ private function get_taxes( $product, $price ) {
...
+			// If the product has a trial period or is synchronised and the first payment is not today.
+			if ( WC_Subscriptions_Product::get_trial_length( $product ) > 0 ) {
+				return true;
+			} elseif ( WC_Subscriptions_Synchroniser::is_product_synced( $product ) && ! WC_Subscriptions_Synchroniser::is_payment_upfront( $product ) && ! WC_Subscriptions_Synchroniser::is_today( WC_Subscriptions_Synchroniser::calculate_first_payment_date( $product, 'timestamp' ) ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
```
james-allan, line 1544: @mattallan & I had a look at this and think there may be a less heavy handed approach to fix this issue. This function currently makes the whole variable product unpurchasable using express payment methods if 1 of the variations requires shipping. 

We agreed that the better way to go would be to use the `ajax_get_selected_product_data()` function or similar to limit the express payment method to eligible products. 

ie:

1. Change this `product_has_trial_and_needs_shipping()` function (and possibly its name) to check if **all** variations require shipping and have a free trial. 
2. Change `ajax_get_selected_product_data()` to throw an exception if the product is ineligible. 

That would have the effect of only loading the express payment method buttons if there is at least 1 eligible variation and hiding the express payment buttons if you choose a variation which is ineligible.

james-allan (changes requested):

Mayisha (approved):
Thanks @mattallan. The fix is working as described :tada:

Mayisha:
@mattallan I have reviewed the changes and approved the PR as things are fixed. But you may need to consider the scenario @james-allan raised about variations.

includes/class-wc-payments-payment-request-button-handler.php
Mayisha: Asking out of curiosity, do you have any type of product in mind that may have some variations that need shipping and some that do not need shipping? I did not think of this scenario initially :thinking:

james-allan: It could be like a variable product where a customer can sign up to the virtual copy or a physical copy of a magazine. But because this logic includes shipping and free trial this could also apply to stores who offers a free trial option on a variable product. eg subscribe now or sign up to our free trial.

mattallan:
Thanks @Mayisha and @james-allan. I've refactored the `product_has_trial_and_needs_shipping()` function so that if this function is passed a variable subscription, it will only return true if all the variations have a trial and require shipping.

This means if you have a subscription that has one variation that is valid for express payment methods and another that is not, the button will still appear on the product page for it and only show for variations that are supported, here's a quick screencast showcasing this:

Screen Capture on 2023-11-03 at 12-48-54

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1703,4 +1707,46 @@ private function get_taxes( $product, $price ) {
...
+
+		foreach ( $products as $product ) {
+			// Return early if the product doesn't require shipping.
+			if ( ! $product->needs_shipping() ) {
+				return false;
+			}
+
+			// If product is synced, check if the first payment is upfront or today (i.e. no trial period). If product is not synced, check if it has a trial period.
+			if ( WC_Subscriptions_Synchroniser::is_product_synced( $product ) ) {
+				if ( WC_Subscriptions_Synchroniser::is_payment_upfront( $product ) || WC_Subscriptions_Synchroniser::is_today( WC_Subscriptions_Synchroniser::calculate_first_payment_date( $product, 'timestamp' ) ) ) {
```
james-allan:
```suggestion
				if ( WC_Subscriptions_Synchroniser::is_payment_upfront( $product ) ) {
```

You can get rid of the `WC_Subscriptions_Synchroniser::is_today()` check because the `is_payment_upfront()` function checks if the product is synced to today [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/5.5.0/includes/class-wc-subscriptions-synchroniser.php#L548-L549).

james-allan (changes requested):
Hey @mattallan.

Changes look good and works. I left a comment suggesting that we can remove the `is_today()` check. I believe that isn't necessary given the `is_payment_upfront()` function includes an equivalent check. :)

includes/class-wc-payments-payment-request-button-handler.php
mattallan: Thanks @james-allan. Looks like `is_payment_upfront` does infact also call `calculate_first_payment_date` and compares the `'ymd'` of this date and current time to check if it's the same ✋

mattallan:
Thanks for the suggestion @james-allan. I had to fix the psalm issues and unit tests but this is good for a final review

Edit: I've ran through the standard tests on the latest changes

mattallan:
While copying these changes over to our Stripe gateway, I found a few more issues over there that are also issues in WooPayments.

Here are the individual fixes I've pushed up to this PR:
- [x] Don't display the PRB on the product page for virtual synced subscriptions that have no upfront payment. This is to prevent incorrect totals that don't account for things like a trial period or a prorated initial payment. Fixed by de80fb0fc
- [x] If a variation is already selected on page load (i.e. using `/?attribute_{name}=value` in URL), don't show the PRBs on page load if that variation isn't supported. Fixed by b9d2ba59c
- [x] After adding a variation to the cart that doesn't support PRBs, don't show PRBs on the page load that occurs while the unsupported variation is still selected. Fixed by 47374a0b4

@james-allan did you mind taking this PR for another spin? 🙏 Thanks!

dmvrtx:
Closing, ECE has changed much since then the PR was open and has a better support of subscription products of different types and options.

PR #7608, opened 2023-11-02
Handle minimum amount error in InitialState (project RPP).

### LINKED ISSUE #7415: RPP / Verification / Minimum amount
### Description

Implement the minimum amount caching and verification in the new payment process.

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L1211-L1216

### Acceptance criteria

1. It is done through a dependency, instead of static `WC_Payments_Utils` methods, probably `MimimumAmountService`. The dependency should have functionality to cache the amount, and verify it. We can do this in parallel to `WC_Payments_Utils` and use the same transient.
2. The minimum amount is cached upon an exception from the server.
3. The minimum amount is verified before processing a new payment.

### Testing instructions

1. Add a debug point where the amount is cached.
2. Try checking out with an insufficient amount, making sure the break point is reached.
3. Try checking out again, and make sure that verification happens before reaching the API/server.

### Additional context

Project: paJDYF-9hL-p2

htdat:
Fixes #7415

- Create a new service `MinimumAmountService` to handle set and get transient, and get error message.
- Integrate this service into `InitialState`. 

#### Testing instructions

* Store with free shipping and flat rate (with something like over 1$)
* Create a product with 10 cents USD.
* Add this product to the checkout with free shipping.
* Try to make payment, see error message regarding the small amount. 
* Choose flat rate, which allows the order larger than minimum amount, and try to make payment again.
* No longer see the error.

Other tests:

* Add breakpoints to the new code in InitialState, and ensure they're run through. 

-------------------

**Post merge**


src/Internal/Payment/State/InitialState.php
```diff
@@ -218,6 +234,42 @@ protected function process_order_phone_number(): void {
...
+	 * @throws StateTransitionException
+	 */
+	protected function process_minimum_amount_exception( Amount_Too_Small_Exception $e ): void {
+		$minimum_amount = $e->get_minimum_amount();
+		$currency       = $e->get_currency();
+
+		$this->minimum_amount_service->set_cache( $currency, $minimum_amount );
+		$error_message = $this->minimum_amount_service->get_error_message_for_shoppers( $currency, $minimum_amount );
+
+		throw new StateTransitionException( $error_message );
```
htdat: Note: now this is throwing an exception. `PaymentErrorState` would be more appropriate but it seems we need to leave this for a later stream.

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7608` or branch name `rpp/7415-verification-minimum-amount` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7415-verification-minimum-amount) :rocket:

---

Build info: 

- Latest commit: 68c40f439ba2b3591ffbb70e6a34a68a0d42a11a
- Build time: 2023-11-14 08:56:29 UTC

htdat (commented):
Ready to get reviews.

src/Internal/Service/MinimumAmountService.php
```diff
@@ -0,0 +1,94 @@
...
+		$this->legacy_proxy = $legacy_proxy;
+	}
+
+	/**
+	 * Saves the minimum amount required for transactions in a given currency.
+	 *
+	 * @param string $currency The currency.
+	 * @param int    $amount   The minimum amount in cents.
+	 */
+	public function set_cache( string $currency, int $amount ): void {
```
RadoslavGeorgiev: I'd call this `set_cached_amount`, and `get_cache` could be `get_cached_amount`. WDYT?

src/Internal/Payment/State/InitialState.php
```diff
@@ -218,6 +234,42 @@ protected function process_order_phone_number(): void {
...
+	protected function verify_minimum_amount(): void {
+		$context        = $this->get_context();
+		$currency       = $context->get_currency();
+		$order_amount   = (int) $context->get_amount();
+		$minimum_amount = $this->minimum_amount_service->get_cache( $currency );
+
+		if ( $minimum_amount > $order_amount ) {
+			$error_message = $this->minimum_amount_service->get_error_message_for_shoppers( $currency, $minimum_amount );
+
+			throw new StateTransitionException( $error_message );
```
RadoslavGeorgiev: As we've already discussed, we need to use specific exceptions rather than `StateTransitionException`s in future streams. In this case, I think it makes sense to throw the same `Amount_Too_Small_Exception` that would be thrown by the API.

RadoslavGeorgiev: As a matter of fact, I'd say that this method can go straight to the new service, which is not aware of `StateTransitionExceptions` at all. You can pass the amount and currency to that method, and use the exception as means of communication.

src/Internal/Payment/State/InitialState.php
RadoslavGeorgiev: We'll need to heavily look into `get_filtered_error_message` as a part of the errors stream, and it will likely be implemented somewhere else. For now, I'd be fine with simply re-throwing the original `Amount_Too_Small_Exception` after caching the amount.

RadoslavGeorgiev (changes requested):
Code looks pretty good, except for a couple of remarks regarding exceptions. I prefer something like this for the public signature of the service:

```php
<?php
class MinimumAmountService {
	/**
	 * Extract and stores the amount, provided by the API through an exception.
	 *
	 * @param Amount_Too_Small_Exception $exception The exception that was thrown.
	 */
	public function store_amount_from_exception( Amount_Too_Small_Exception $exception ): void;

	/**
	 * Checks if there is a minimum amount required for transactions in a given currency.
	 *
	 * @param string $currency The currency to check for.
	 *
	 * @throws Amount_Too_Small_Exception
	 */
	public function verify_amount( int $amount, string $currency ): void;
}
```

WDYT?

htdat:
@RadoslavGeorgiev - Thanks a lot for detailed reviews. All of them make a lot of sense, and simplify code a lot. I've followed your suggestion, and marked this PR ready to review again.

src/Internal/Payment/State/InitialState.php
```diff
@@ -118,13 +131,21 @@ public function start_processing( PaymentRequest $request ) {
 		if ( null !== $duplicate_payment_result ) {
 			return $duplicate_payment_result;
 		}
+
+		$context = $this->get_context();
+		$this->minimum_amount_service->verify_amount(
+			(string) $context->get_currency(),
+			(int) $context->get_amount()
```
RadoslavGeorgiev: This looked weird at first, and now I see that it is done in order to allow tests to pass. Let's remove this (unsafe) casting, and add simple mocks to the context in tests:

```php
// Make sure that the minimum amount verification is properly called.
$this->mock_context->expects( $this->once() )->method( 'get_amount' )->willReturn( 1 );
$this->mock_context->expects( $this->once() )->method( 'get_currency' )->willReturn( 'usd' );
```

If this causes an issue in the real world, we should be able to catch it and fix it. Otherwise we might have some silent issues.

RadoslavGeorgiev (changes requested):
The code is almost 💯 ready to go, just one nitpick.

src/Internal/Payment/State/InitialState.php
htdat: > This looked weird at first, and now I see that it is done in order to allow tests to pass. 

@RadoslavGeorgiev - My original intention was not to pass tests. Instead, that's because `$context->get_curreny()` and `$context->get_amount()` can return `null` based on their signatures. 

However, in the code flow of `InitialState`, that's NOT possible since their respective setters are always called in advance as a part of the `populate_context_from_order` call.

Given that, if both of them are null, then this call would be like the following, which is still acceptable: 

```php
		$this->minimum_amount_service->verify_amount('', 0)
```

WDYT?

RadoslavGeorgiev: > However, in the code flow of `InitialState`, that's NOT possible

I'd rephrase this as "NOT **currently** possible". If someone alters this fact in the future, they would be less likely to realize it, as there would be no errors. Removing the casting, and adding tests, would force them to realize that something might be wrong in that theoretical future scenario.

htdat: Agreed that this can be changed in the future.
Addressed here 68c40f439ba2b3591ffbb70e6a34a68a0d42a11a

htdat:
@RadoslavGeorgiev - thanks for feedback. Ready to review again.

RadoslavGeorgiev (approved):
LGTM, thank you for addressing my feedback! :shipit:

PR #7575, opened 2023-10-27
Remove UGX from zero decimal currencies

naman03malhotra:
Fixes #6946 

#### Changes proposed in this Pull Request
Removed UGX currency from ZDC `zero_decimal_currencies` array.

#### Testing instructions
- Setup a store with base currency UGX. Ensure that the products are setup with correct UGX prices (e.g. 18 UGX is not a valid price. The UGX to USD conversion rate is 4000)
- Use multi-currency module to checkout with UGX/ISK.
- Test checkout for different types of currencies - zero-decimal, special case (UGX, ISK) and two decimal currencies.
- Test Transaction, Order Edit and other screens if the values shown are correct.

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [x] Tested on mobile (or does not apply)

**Post merge**

- [x] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7620, opened 2023-11-03
Add fix for special case zero decimal currencies

### LINKED ISSUE #6946: UGX generates zero decimal currency errors because should not be zero based
### Describe the bug

UGX is now a zero decimal currency, and placing orders in UGX generates errors. 

### To Reproduce

1. Go to shop base settings and set UGX as base currency
2. Set up a product using UGX and checkout
3. Checkout

### Actual behavior

With UPE disabled a general error occurs `We cannot process at this time`

### Expected behavior

UGX is supported, as it shows as available as a base currency

### Additional context
6626828-zd-a8c

jessy-p:
Fixes #6946 

- Modified `prepare_amount` to handle special case zero-decimal currencies 

UGX and ISK are the special cases of zero decimal currencies as described [here](https://stripe.com/docs/currencies#special-cases) by Stripe. These currencies, though technically zero-decimal, have to be sent with two decimals to Stripe.

#### Testing instructions
- Setup a store with base currency UGX. Ensure that the products are setup with correct UGX prices (e.g. 18 UGX is not a valid price. The UGX to USD conversion rate is 4000)
- Use multi-currency module to checkout with UGX/ISK.
- Test checkout for different types of currencies - zero-decimal, special case (UGX, ISK) and two decimal currencies.
- Test Transaction, Order Edit and other screens if the values shown are correct.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7620` or branch name `fix/6946-zero-decimal-special-case` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/6946-zero-decimal-special-case) :rocket:

---

Build info: 

- Latest commit: 6d11eeb7ccee4108d5de148ade6b81962dfe9e07
- Build time: 2023-11-09 17:29:49 UTC

tpaksu (approved):
Thanks for working on this and fixing it! The code is clear and understandable. 

I've tested with several currencies. Here are some screenshots:

image

image

image

These are working correctly. But I found one minor issue on the frontend: 

image

image

Only the conversion tooltip and transaction details for UGX show the amount multiplied by 100. ISK and the others are showing fine. Maybe there's another part in the frontend code that checks for ZDC, and doesn't have UGX on the list? 

Pre-approving this, as it fixes the main issue, which is not being able to take payments via UGX.

> [!IMPORTANT]
> If you're not going to handle the UGX display problem on the frontend in this PR, please don't forget to create another issue for it. Thanks in advance!

includes/class-wc-payments-utils.php
```diff
@@ -119,7 +119,7 @@ public static function esc_interpolated_html( $string, $element_map ) {
 	public static function prepare_amount( $amount, $currency = 'USD' ): int {
 		$conversion_rate = 100;
 
-		if ( self::is_zero_decimal_currency( strtolower( $currency ) ) ) {
+		if ( ! self::is_special_case_zero_decimal_currency( strtolower( $currency ) ) && self::is_zero_decimal_currency( strtolower( $currency ) ) ) {
```
naman03malhotra, line 122: Wouldn't we want to supplement this change in other places where we are dealing with `conversion_rate` and formatting the currency? For example, `interpret_stripe_amount` is one such function.

naman03malhotra: image

naman03malhotra: It seems like the reasoning behind @tpaksu 's https://github.com/Automattic/woocommerce-payments/pull/7620#pullrequestreview-1714770579.
>Only the conversion tooltip and transaction details for UGX show the amount multiplied by 100

jessy-p:
@tpaksu
>UGX show the amount multiplied by 100. ISK and the others are showing fine. Maybe there's another part in the frontend code that checks for ZDC, and doesn't have UGX on the list?

Looks like the issue is because `zero_decimal_currencies` https://github.com/Automattic/woocommerce-payments/blob/e3b86b4e0bd9fe9311dcc801d9b79c5f709f9acc/includes/class-wc-payments-utils.php#L193 does not list ISK. Maybe all we need to do is to remove UGX also from the list, if ISK is functioning fine?

tpaksu:
@Automattic/sigma can you check?

jessepearson:
> @Automattic/sigma can you check?

I've been looking into this, and I noticed that the transaction fee in the order is incorrect, as well. 

Screenshot 2023-11-08 at 2 37 55 PM

I feel like we've just gotten lucky with ISK at this point and that the fact that it's not in our zero decimal currency list, and that it's decimals default to 0 via the locale config, has allowed it to slide under the radar.

I also want to note that Stripe's docs technically are not correct. They have the [zero decimal currencies list](https://stripe.com/docs/currencies#zero-decimal), and then just below they have the [special cases list](https://stripe.com/docs/currencies#special-cases), which includes _additional_ zero decimal currencies.

I'll look more into the issue tomorrow to see if we should split up the problems into additional issues or if we can make some more changes here to fix everything together.

jessy-p:
@jessepearson  if ISK is working perfectly fine, just by being left out of zero-decimal currencies, could I just skip the changes in this PR and instead remove UGX too from the zero-decimal currency list?

naman03malhotra:
Stripe's [docs](https://stripe.com/docs/currencies#special-cases) are incorrect as @jessepearson mentioned, as they list UGX as a zero decimal currency and then the opposite is mentioned in the Special cases. IMO, for now as customers are blocked on this, we can remove UGX from zero decimal currency and go ahead with rest of the changes and open a separate issue for ISK.

Releasing it with 6.8.0 would allow the blocked customers to take advantage of BFCM time :)

jessy-p:
Added formatting fixes for UGX in case we want to move this issue in this release

image

jessepearson:
> @jessepearson if ISK is working perfectly fine, just by being left out of zero-decimal currencies, could I just skip the changes in this PR and instead remove UGX too from the zero-decimal currency list?

ISK seems functional for payments, but it's got problems in Multi-Currency. If you add it as an additional currency it is treated as a decimal currency and not a zero-decimal currency for rounding and charm purposes. As far as I know this isn't transaction breaking, so a new issue could be created and fixed later.

For UGX, if it is removed from the zero-decimal currencies list the same thing will happen.

I am still seeing the amounts in the transaction timeline multiplied by 100, so 1,000 becomes 100,000. I believe this is due to `WC_Payments_API_Client->get_timeline` gets the data from Stripe, and it's returned multiplied times 100 (like we have to pass it). We can add a hacky fix to divide these amounts by 100 if the currency is UGX just so we can get this out for merchants to use throughout the holiday season.

From there, I think better handling of zero-decimal currencies should happen via another issue. The `locale-info.php` file that is included in WC core is what is used for formatting of currencies, it may be possible to get rid of the `zero_decimal_currencies` array and use that as a reference instead. In a quick search it looks like quite a few more currencies are zero-decimal than what are listed in the `zero_decimal_currencies` array, which seems to only be just the ones Stripe accepts. From there, we could set up special handling for currencies like ISK and UGX.

jessy-p:
So the question is, should I add all this handling for `UGX` which is already in this PR, or should I just remove `UGX` from the zero decimal currency list as a quick fix, if our immediate priority is to stop transaction breaking for `UGX`?
@naman03malhotra  @jessepearson

naman03malhotra:
IMO, we should temporarily remove UGX from ZDC to solve the issue and unblock customers from sales for now during the high sales season.

However, this issue warrants a proper fix as @jessepearson suggested in his approach. Ref https://github.com/Automattic/woocommerce-payments/pull/7620#issuecomment-1804337691

I suggest we go ahead with removing UGX from ZDC and release it after proper testing. Also, I suggest that we add the testing instructions in critical flow to identify bugs.

jessy-p:
Closed in favor of https://github.com/Automattic/woocommerce-payments/pull/7575

PR #7640, opened 2023-11-06
Add customer details flow to inital state

### LINKED ISSUE #7417: RPP / Processing / Manage customer details
### Description

Implement customer management within the new payment process, and do it as close to the payment request as possible.

As a part of the minimal checkout through the new payment process, this piece of code is in place:

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payments-customer-service.php#L157-L171

However, everything from the following method of gateway needs to be handled, including customer update:

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L1074-L1092

### Acceptance criteria

1. Customer management during the payment process is a part of providing the necessary data for a successful payment to the API, and is very API-specific. Because of this, it should be grouped with the remote payment processing call, `$this->payment_request_service->create_intent()`. Once we switch to a separate API version, it will likely work differently.
2. Customers are both created and updated, depending on their prior existence.
3. The customer ID is stored within the order before the payment processing request gets sent.
4. The existing customer service seems to be organized well enough, and it does not require refactoring yet. The only current downside is that it requires the `WC_Order` object, which we're trying to avoid within `src`. For now, we've made an exception for it, but if any other principle is broken, it might warrant creating a new order service within `src`, which either contains the necessary functionality, or delegates it to the legacy service.

### Testing instructions

- Manual testing: Checkout as a guest should update a new customer. Checkout as an existing user should update the existing customer.
- Unit tests: 100% coverage of everything within `src`.

### Additional context

Project: paJDYF-9hL-p2

zmaglica:
Fixes #7417 

This PR ads a customer detail flow to the initial state where we create a customer or update the customer's data before we proceed with the payment processing

#### Testing instructions

* Make sure that the customer is created if it's not existed
* Make sure that the customer is updated if it exists
* Make sure that customer is added to the order before payment is processed

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7640` or branch name `rpp/7417-manage-customer-details` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7417-manage-customer-details) :rocket:

---

Build info: 

- Latest commit: f329ca091c191c4e0d0a21384812cb1b6c01f7c5
- Build time: 2023-11-15 16:34:46 UTC

changelog/rpp-7417-manage-customer-details
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: dev
+
+Add customer details flow to the initial state
```
RadoslavGeorgiev: Imagine reading the changelog without context. We should mention something about managing customer details within the re-engineered payment process, rather than focusing on a state.

src/Internal/Payment/State/InitialState.php
```diff
@@ -109,6 +114,9 @@ public function start_processing( PaymentRequest $request ) {
 		// Start multiple verification checks.
 		$this->process_order_phone_number();
 
+		// Manage customer details.
+		$this->manage_customer_details();
```
RadoslavGeorgiev: This method needs to be called as closely to `$this->payment_request_service->create_intent( $context )` as possible, as it's essentially a pair.

src/Internal/Payment/State/InitialState.php
```diff
@@ -109,6 +114,9 @@ public function start_processing( PaymentRequest $request ) {
 		// Start multiple verification checks.
 		$this->process_order_phone_number();
 
+		// Manage customer details.
```
RadoslavGeorgiev: Let's mention that customer details are used for the payment, rather than just a side detail. One thing I dislike about the current process is that the link between the two is not as apparent as it should be.

src/Internal/Payment/State/InitialState.php
```diff
@@ -187,12 +194,34 @@ protected function populate_context_from_order() {
...
+		// Determine the customer making the payment, create one if we don't have one already.
+		$customer_id = $this->customer_service->get_customer_id_by_user_id( $user_id );
+		if ( null === $customer_id ) {
+			$customer_id = $this->customer_service->create_customer_for_user( $user, $customer_data );
+		} else {
+			$metadata = $context->get_metadata();
+			// We don't need to apply a filter here as it currently is.
+			if ( filter_var( $metadata['paid_on_woopay'] ?? false, FILTER_VALIDATE_BOOLEAN ) ) {
+				$customer_data['email'] = $user->user_email;
+			}
```
RadoslavGeorgiev: I think we can skip this part for now. When it comes to WooPay, the process will likely look differently, and we will have proper ways to detect the flow.

FWIW using metadata to transfer flags between various places is one of the hardest to understand things in the current payment process.

src/Internal/Payment/State/InitialState.php
```diff
@@ -187,12 +194,34 @@ protected function populate_context_from_order() {
...
+	 *
+	 * @throws Order_Not_Found_Exception In case the order could not be found.
+	 * @throws API_Exception
+	 * @throws Exception
+	 */
+	protected function manage_customer_details() {
+		$context       = $this->get_context();
+		$user_id       = $context->get_user_id();
+		$user          = get_user_by( 'id', $user_id );
+		$customer_data = WC_Payments_Customer_Service::map_customer_data( $this->order_service->_deprecated_get_order( $context->get_order_id() ), new WC_Customer( $user_id ) );
```
RadoslavGeorgiev: Can we move most of this logic out of the state, and into the service? As far as the payment process is concerned, there is `( user_id, order|wc_customer )  => customer_id`, and that can apply to other situations. IMO this could live in the `get_or_create_customer_id_from_order` I created for the new payment process (and is otherwise unused right now).

src/Internal/Payment/State/InitialState.php
```diff
@@ -187,12 +194,34 @@ protected function populate_context_from_order() {
...
-			$context->get_user_id(),
-			$this->order_service->_deprecated_get_order( $order_id )
-		);
+	/**
+	 * Parse and manage customer details.
+	 * This includes the update/creation of a customer and link it to the order.
+	 *
+	 * @throws Order_Not_Found_Exception In case the order could not be found.
+	 * @throws API_Exception
+	 * @throws Exception
```
RadoslavGeorgiev: The API exception makes sense, and we can handle it together with intent-related API exceptions in the future, but where does this one come from? I don't see any methods in here that throw it. Let's see where it originates, and decide how to handle it, as states should not throw plain `Exception`s.

RadoslavGeorgiev (changes requested):

src/Internal/Payment/State/InitialState.php
zmaglica: I removed this function and simplified a code where I used the existing `get_or_create_customer_id_from_order` function and added missing logic for updating a customer data. Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7640/commits/e86e5ef2e21cbe1588a18de580827515557ac946

src/Internal/Payment/State/InitialState.php
zmaglica: You are right on this one. I refactored this part by using the already existing method(s) from customer service. Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7640/commits/e86e5ef2e21cbe1588a18de580827515557ac946

src/Internal/Payment/State/InitialState.php
zmaglica: Make sense. Metadata is a tricky one. I wasn't sure will I get the data I need there. This is removed.

src/Internal/Payment/State/InitialState.php
zmaglica: This is refactored and simplified (look for previous comments). Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7640/commits/e86e5ef2e21cbe1588a18de580827515557ac946

src/Internal/Payment/State/InitialState.php
zmaglica: This plan Exception is no longer thrown here since this part of code is moved into the customer service. The `new WC_Customer( $user_id ) `can throw plan `Exception`,  so I had to add it.

changelog/rpp-7417-manage-customer-details
zmaglica: Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7640/commits/3503d9beea9aa377568c494e02878d163d720c0c

src/Internal/Payment/State/InitialState.php
```diff
@@ -120,18 +122,25 @@ public function start_processing( PaymentRequest $request ) {
...
-		// Payments are currently based on intents, request one from the API.
+		// Payments are currently based on intents, request one from the API, and before that make sure that customer is created or updated.
 		try {
-			$context = $this->get_context();
-			$intent  = $this->payment_request_service->create_intent( $context );
+			$context  = $this->get_context();
+			$order_id = $context->get_order_id();
+
+			// Create or update customer and customer details.
+			$customer_id = $this->customer_service->get_or_create_customer_id_from_order( $context->get_user_id(), $this->order_service->_deprecated_get_order( $order_id ) );
```
RadoslavGeorgiev: Let's make this one multi-line:


```suggestion
			$customer_id = $this->customer_service->get_or_create_customer_id_from_order(
				$context->get_user_id(),
				$this->order_service->_deprecated_get_order( $order_id )
			);
```

It's by far the longest line in this file.

src/Internal/Payment/State/InitialState.php
```diff
@@ -120,18 +122,25 @@ public function start_processing( PaymentRequest $request ) {
 		}
 		// End multiple verification checks.
 
-		// Payments are currently based on intents, request one from the API.
+		// Payments are currently based on intents, request one from the API, and before that make sure that customer is created or updated.
```
RadoslavGeorgiev:
```suggestion
		/**
		 * Payments are based on intents, and intents use customer objects for billing details.
		 *
		 * The customer is created/updated right before requesting the creation of
		 * a payment intent, and the two actions must be adjacent to each-other.
		 */
```

tests/unit/src/Internal/Payment/State/InitialStateTest.php
```diff
@@ -176,8 +199,32 @@ public function test_start_processing() {
...
+			->method( '_deprecated_get_order' )
+			->with( $order_id )
+			->willReturn( $mock_order );
+		$this->mock_customer_service->expects( $this->once() )
+			->method( 'get_or_create_customer_id_from_order' )
+			->with( $user_id, $mock_order )
+			->willReturn( $customer_id );
+		$this->mock_context->expects( $this->once() )
+			->method( 'set_customer_id' )
+			->with( $customer_id );
```
RadoslavGeorgiev: This is quite a complicated setup, and blurs the purpose of the specific test. Could you please move it to a separate private method?

includes/class-wc-payments-customer-service.php
```diff
@@ -165,22 +165,21 @@ public function create_customer_for_user( WP_User $user, array $customer_data ):
...
 	 * @return string           WooPayments customer ID.
-	 */
+	 * @throws API_Exception    Throws when server API request fails.
+*/
 	public function get_or_create_customer_id_from_order( int $user_id, WC_Order $order ): string {
 		// Determine the customer making the payment, create one if we don't have one already.
-		$customer_id = $this->get_customer_id_by_user_id( $user_id );
+		$customer_id   = $this->get_customer_id_by_user_id( $user_id );
+		$customer_data = self::map_customer_data( $order, new WC_Customer( $user_id ) );
+		$user          = get_user_by( 'id', $user_id );
```
RadoslavGeorgiev: Seems like this doesn't work when creating customers in incognito (w/o user logged in) because `create_customer_for_user()` requires a `WP_User` object, and this line returns `false` when `$user_id` is 0 (which it is when no customer is logged in).

The current implementation in the gateway is slightly weird: It's either `$order->get_user()` (🆗) or `wp_get_current_user()` (⛔) otherwise. Then we're working with a dummy `WP_User` object, which requires the weird checks here: https://github.com/Automattic/woocommerce-payments/blob/e3b86b4e0bd9fe9311dcc801d9b79c5f709f9acc/includes/class-wc-payments-customer-service.php#L151-L153C4

I'm marking the PR as blocked for now, let's discuss how to solve this tomorrow.

RadoslavGeorgiev (changes requested):

src/Internal/Service/OrderService.php
```diff
@@ -170,7 +170,8 @@ public function import_order_data_to_payment_context( int $order_id, PaymentCont
 
 		$context->set_currency( $currency );
 		$context->set_amount( $amount );
-		$context->set_user_id( $user->ID );
+		// In case we don't have user, we are setting user id to be 0 which could cause more harm since we don't have a real user.
+		$context->set_user_id( $user->ID > 0 ? $user->ID : null );
```
RadoslavGeorgiev: Now that we have this line, we do not need the `$user = $this->legacy_proxy->call_function( 'wp_get_current_user' )` a few lines above. Seems like it was only there to make sure a `WP_User` instance is available, even if the user is null. Now that we actually allow the code to work with `null`, we can get rid of it.

RadoslavGeorgiev: Also, if it is `null`, there is an exception that `set_user_id` of the context requires an integer. Please modify the context method to allow for null users.

RadoslavGeorgiev (changes requested):
Almost there! A couple of nitpicks, and I'll do a final test.

Please make sure to test in incognito mode, whithout a customer!

src/Internal/Service/OrderService.php
zmaglica: > Also, if it is `null`, there is an exception that `set_user_id` of the context requires an integer. Please modify the context method to allow for null users.

I am sorry. I haven't committed that file (it was stashed for unknown reason to me). I have committed it now.

zmaglica: > Now that we have this line, we do not need the `$user = $this->legacy_proxy->call_function( 'wp_get_current_user' )` a few lines above. Seems like it was only there to make sure a `WP_User` instance is available, even if the user is null. Now that we actually allow the code to work with `null`, we can get rid of it.

Addressed here: https://github.com/Automattic/woocommerce-payments/pull/7640/commits/7677e2b93910ece4f5903976220b9422fa61aae3

RadoslavGeorgiev (approved):
LGTM :shipit: 🎉

PR #7643, opened 2023-11-06
Update links across the plugin to woo.com

### LINKED ISSUE #7637: Update references to woocommerce.com to point to woo.com
### Description

As part of the branding update, we should update any hyperlinks in the plugin pointing to `woocommerce.com` to `woo.com` - this isn't urgent because they will redirect anyway, but it would be nice to tackle for conistency.

dmallory42:
Fixes #7637 

- Update hyperlinks across the plugin to refer to `woo.com` rather than `woocommerce.com`
- Update plugin URL to point to `woo.com` in the `woocommerce-payments.php` file.

#### Testing instructions

* Spot check a few of the URLs in the diff to ensure they go to the correct woo.com documentation page.
* Check a few of the links in the plugin (for example, there are some hyperlinks on the `Payments > Settings` page like the Multi-Currency documentation link).

-------------------

**Post merge**

### DIFF
--- a/client/utils/format-ssr.js
+++ b/client/utils/format-ssr.js
@@ -125,7 +125,7 @@ Taxonomies: Product Types: ${ printTerms( systemStatus.settings.taxonomies ) }
 Taxonomies: Product Visibility: ${ printTerms(
 		systemStatus.settings.product_visibility_terms
 	) }
-Connected to Woo.com: ${
+Connected to WooCommerce.com: ${
 		systemStatus.settings.woocommerce_com_connected === 'yes'
 			? CHECK_MARK
 			: DASH_MARK
--- a/client/utils/format-ssr.js
+++ b/client/utils/format-ssr.js
@@ -125,7 +125,7 @@ Taxonomies: Product Types: ${ printTerms( systemStatus.settings.taxonomies ) }
 Taxonomies: Product Visibility: ${ printTerms(
 		systemStatus.settings.product_visibility_terms
 	) }
-Connected to WooCommerce.com: ${
+Connected to woo.com: ${
 		systemStatus.settings.woocommerce_com_connected === 'yes'
 			? CHECK_MARK
 			: DASH_MARK
--- a/client/utils/format-ssr.js
+++ b/client/utils/format-ssr.js
@@ -125,7 +125,7 @@ Taxonomies: Product Types: ${ printTerms( systemStatus.settings.taxonomies ) }
 Taxonomies: Product Visibility: ${ printTerms(
 		systemStatus.settings.product_visibility_terms
 	) }
-Connected to woo.com: ${
+Connected to Woo.com: ${
 		systemStatus.settings.woocommerce_com_connected === 'yes'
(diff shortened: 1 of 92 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7643` or branch name `dev/woo.com-update` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=dev/woo.com-update) :rocket:

---

Build info: 

- Latest commit: 8fcb7fc3ddbdddfb8106d3454abaf9673cc2f86c
- Build time: 2023-11-08 10:53:56 UTC

client/utils/format-ssr.js
```diff
@@ -125,7 +125,7 @@ Taxonomies: Product Types: ${ printTerms( systemStatus.settings.taxonomies ) }
 Taxonomies: Product Visibility: ${ printTerms(
 		systemStatus.settings.product_visibility_terms
 	) }
-Connected to WooCommerce.com: ${
+Connected to Woo.com: ${
 		systemStatus.settings.woocommerce_com_connected === 'yes'
```
daquinons: Umm, I guess this is similar to the "WCpay" mentions in code after the branding rollout, that we don't want to touch them? 🤔

daquinons (approved):
Thank you, Dan! Approving but I'd suggest to wait for the response on the last point that I'm going to mention!

I've left a comment with a question, but I believe I know the answer is we probably don't want to touch a settings name!

Also, I think this is the only thing missing. It's a documentation that I don't think it's very important, but it's the only place outside of changelogs where woocommerce.com still appears https://github.com/Automattic/woocommerce-payments/blob/af25053380d03eeb03a4bc3f30bf87a7279f24e7/docs/version-support-policy.md#L3 but the redirection seems to not be applied to the Woo Developers blog yet. Asked marketing here paJDYF-b5P-p2#comment-21649

client/utils/format-ssr.js
dmallory42: So, I think this is a string that will be added to the `SSR` - however, I think you are probably right because maybe people have macros etc with the search for this string built into them, so I've reverted this change in db2cf9c67.

dmallory42:
Thanks for the review @daquinons!

> Also, I think this is the only thing missing. It's a documentation that I don't think it's very important, but it's the only place outside of changelogs where woocommerce.com still appears 

I didn't change this one deliberately because the `developers.woo.com` link was broken - thanks for asking marketing about it, I should have done that! Once we get confirmation if that URL is going to exist I can either merge or fix the link and merge. :)

dmallory42:
@daquinons I updated that URL to `developer.woo.com` as a result of the discussion you started in P2! The URL should be propogated soon, so it will be in place by the time of the next WooPayments release.

PR #7651, opened 2023-11-07
Changes to Payment Process to store the working mode of the gateway

### LINKED ISSUE #7421: RPP / Completion / Store the working mode of the gateway
### Description

Previously implemented here in the existing payment process, we need to make sure that the mode gets updated within the new payment process as well.

https://github.com/Automattic/woocommerce-payments/blob/43433555c9f48abde8f8a98570ae3833e272934a/includes/class-wc-payment-gateway-wcpay.php#L1229

### Acceptance criteria

- Done through a setter/getter pair within the new order service.
- To allow storing the mode within the context in the future, let's also store it in the context.
- We'll be calling the right completion methods at the end of the PP, even upon failure, so this can be done while completing the process.

### Testing instructions

1. Checkout.
2. Make sure that the right mode is stored against the order.

### Additional context

Project: paJDYF-9hL-p2

jessy-p:
Fixes #7421 

- Stores the mode in `PaymentContext`
- updates the mode into `PaymentContext` in `PaymentProcessingService`
- Updates the mode into `_wcpay_mode` order metadata when the payment process finishes.

#### Testing instructions
- Place an order with the new payment process and check that `_wcpay_mode` is populated in order metadata.

-------------------

- [x] N/A - Tested on mobile (or does not apply)

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7651` or branch name `add/7421-store-working-mode` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/7421-store-working-mode) :rocket:

---

Build info: 

- Latest commit: 6adc8dc70af5c39a241e2ec1c31885f0b4d542d3
- Build time: 2023-11-15 12:02:17 UTC

src/Internal/Service/OrderService.php
```diff
@@ -421,4 +422,16 @@ protected function get_order( int $order_id ): WC_Order {
...
 		return $order;
 	}
+
+	/**
+	 * Updates the '_wcpay_mode' meta data on an order.
+	 *
+	 * @param WC_Order $order The order.
+	 * @param string   $mode  Mode from the context.
+	 */
+	private function update_mode_meta_data( WC_Order $order, string $mode ) : void {
```
RadoslavGeorgiev: To keep things consistent with the other methods, let's:

1. Make this method public.
2. Switch from using the `WC_Order` object to an order ID.
3. Add a getter for the meta, even though it would not be used for now.
4. Maybe rename this from `update_mode_meta_data` to `set_mode`? From outside the service, the mode should be only a property of the order, regardless of how it is stored.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -45,22 +48,42 @@ class PaymentProcessingService {
...
+	 */
+	private $logger;
+
+	/**
+	 * Mode
+	 *
+	 * @var Mode
+	 */
+	private $mode;
+
```
RadoslavGeorgiev: Let's make it one empty line here instead of 2.

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -134,6 +157,11 @@ public function get_authentication_redirect_url( $intent, int $order_id ) {
 	protected function create_payment_context( int $order_id, bool $manual_capture = false ): PaymentContext {
 		$context = new PaymentContext( $order_id );
 		$context->toggle_manual_capture( $manual_capture );
+		try {
+			$context->set_mode( $this->mode->is_test() ? 'test' : 'prod' );
+		} catch ( Exception $e ) {
+			$this->logger->error( 'Failed to fetch the mode: ' . $e );
```
RadoslavGeorgiev: This exception is almost impossible to appear as late as the beginning of the payment process, as the mode will surely be initialized by this point. With that in mind, let's just `->set_mode( 'unknown' )` here, and add a comment that this is not an expected scenario.

RadoslavGeorgiev (changes requested):
Works like a charm! 🎉 

I only have a couple of nitpicks:

src/Internal/Service/OrderService.php
jessy-p: > Switch from using the WC_Order object to an order ID.

Could you explain why?

src/Internal/Service/PaymentProcessingService.php
jessy-p: The `is_test` method throws `Exception` so I either have to deal with it at this point or keep bubbling it up, which looked really odd.

src/Internal/Service/OrderService.php
RadoslavGeorgiev: Because that is the way the service works. We don't want to carry `WC_Order` objects around, but work with IDs instead. Because of that, you'll notice how `_deprecated_get_order` is not "just" a method of the service, but is mostly used internally.

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: I'm not saying to bubble the exception up:
```suggestion
			$context->set_mode( 'unknown' );
```

A simple setter to make sure something was set should be sufficient here.

src/Internal/Service/OrderService.php
jessy-p: >We don't want to carry WC_Order objects around, but work with IDs

Internally when using these private functions, isn't that an overhead, passing around IDs and constructing WC_Order again and again?

RadoslavGeorgiev: We've already discussed this, and the conclusion is explained here: paJDYF-9hL-p2#comment-21012

If we see that there's additional overhead, we could introduce a small caching layer within the internal `get_order()` method, but the point is that the order service should be the link between WooPayments and Woo core, and nothing outside the service should be aware of `WC_Order` instances. 

> when using these private functions,

I do not see a reason to keep the mode setter+getter pair private.

jessy-p: Thank you for explaining it.

src/Internal/Service/PaymentProcessingService.php
jessy-p: Done

src/Internal/Service/OrderService.php
jessy-p: Changes made.

src/Internal/Service/PaymentProcessingService.php
jessy-p: Done

src/Internal/Service/PaymentProcessingService.php
```diff
@@ -45,22 +48,41 @@ class PaymentProcessingService {
 	 */
 	private $context_logger_service;
 
+	/**
+	 * Logger instance.
+	 *
+	 * @var Logger
+	 */
+	private $logger;
```
RadoslavGeorgiev: The logger is not used here anymore.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
```diff
@@ -126,8 +143,40 @@ public function test_process_payment_happy_path_without_mock_builder() {
...
+			->with( InitialState::class, $this->isInstanceOf( PaymentContext::class ) )
+			->willReturn( $mock_initial_state );
+
+		$mock_initial_state->expects( $this->once() )
+			->method( 'start_processing' )
+			->with( $this->isInstanceOf( PaymentRequest::class ) )
+			->willReturn( $mock_completed_state );
+
+		// process completes as mode is used only for logging and meta.
+		$this->assertSame( $mock_completed_state, $sut->process_payment( 1 ) );
```
RadoslavGeorgiev: Is there a reason to test all of `process_payment` just to see whether `create_state` handles the mode correctly? IMO as long as the exception is thrown, and `create_payment_context` handles it, that should be good enough.

RadoslavGeorgiev (approved):
Great job. Code looks good, and works like a charm! 🎉 

I only have a couple of minor comments. Feel free to address them as you see fit, and merge the PR without further review.

src/Internal/Service/PaymentProcessingService.php
jessy-p: I had debated removing it, but considering this is the core of payment process, do you think we might start using it soon?

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
jessy-p: Any suggestion for how to examine that the un-mocked protected method `create_payment_context` handles the exception without errors?

src/Internal/Service/PaymentProcessingService.php
RadoslavGeorgiev: It's pretty easy to add it later if needed, but I don't think we will. AFAIAC, the service will not use the logger directly, and rely on `PaymentContextLoggerService` instead.

tests/unit/src/Internal/Service/PaymentProcessingServiceTest.php
RadoslavGeorgiev: It's relatively easy to test `protected` or `private` methods. You can:

1. Create a temporary instance of the service, like how it's done on line 99 (within the `test_process_payment_happy_path` test).
2. Use `PHPUnit_Utils::call_method`, see https://github.com/Automattic/woocommerce-payments/blob/942dfb0579f0a9d27af1f409ff179ea403339031/tests/unit/src/Internal/Payment/State/InitialStateTest.php#L520 for reference.

jessy-p: Changed

PR #7665, opened 2023-11-08
Fixes to the account status logic

dmallory42:
Fixes server/3899

Note: this should be tested in conjunction with the server PR.

* Improvement to the payments status logic, will return `Active` rather than pending if the account is pending verification but payments are enabled.
* Refactored the deposits logic to simplify slightly.

#### Testing instructions

* Following the instructions in the server/3899 PR, onboard a new PO US account with the tax capabilities, then go to **Payments > Overview** and ensure that the account status is `Pending`, the payments status is `Active`, and the deposits status is `Pending Verification`.
* Make a test transaction, clear the account cache and check the account status again, verifying that it did not change.

-------------------

**Post merge**

### DIFF
--- a/client/components/payments-status/index.tsx
+++ b/client/components/payments-status/index.tsx
@@ -60,11 +60,12 @@ interface Props {
 const PaymentsStatus: React.FC< Props > = ( props ) => {
 	const { paymentsEnabled, accountStatus } = props;
 
-	if ( 'pending_verification' === accountStatus ) {
-		return <PaymentsStatusPending iconSize={ props.iconSize } />;
+	if ( paymentsEnabled ) {
+		return <PaymentsStatusEnabled iconSize={ props.iconSize } />;
 	}
-	return paymentsEnabled ? (
-		<PaymentsStatusEnabled iconSize={ props.iconSize } />
+
+	return accountStatus === 'pending_verification' ? (
+		<PaymentsStatusPending iconSize={ props.iconSize } />
 	) : (
 		<PaymentsStatusDisabled iconSize={ props.iconSize } />
 	);
--- a/changelog/update-account-status-logic
+++ b/changelog/update-account-status-logic
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Updates to the account status logic to reflect status more accurately in some cases.
--- a/client/components/account-status/test/__snapshots__/index.js.snap
+++ b/client/components/account-status/test/__snapshots__/index.js.snap
@@ -246,7 +246,7 @@ exports[`StatusChip renders pending verification status 1`] = `
 </div>
 `;
 
-exports[`StatusChip renders pending verification status for po 1`] = `
+exports[`StatusChip renders pending verification status for progressive onboarding 1`] = `
 <div>
   <button
     class="wcpay-tooltip__content-wrapper"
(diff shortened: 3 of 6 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7665` or branch name `update/account-status-logic` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/account-status-logic) :rocket:

---

Build info: 

- Latest commit: 23b13580d36823af885dff210bcaf3abbff8a404
- Build time: 2023-11-09 19:29:21 UTC

client/components/payments-status/index.tsx
```diff
@@ -60,11 +60,12 @@ interface Props {
 const PaymentsStatus: React.FC< Props > = ( props ) => {
 	const { paymentsEnabled, accountStatus } = props;
 
-	if ( 'pending_verification' === accountStatus ) {
-		return <PaymentsStatusPending iconSize={ props.iconSize } />;
+	if ( paymentsEnabled ) {
```
oaratovskyi, line 63: This change created a bit of confusion IMO
This is how it looks when Stripe returns status pending within first minute
image
And after making transaction and clearing the cache
image

it used to be both payments and deposits`Pending verification`, now it's only deposits
WDYT?

oaratovskyi (approved):
Big thanks for refactoring deposits status, it's now way clearer! 
Tested with the server PR, works according to testing instructions 🟢 
Pre-approving this, I have only one question before it becomes ready to merge, left it as a review comment 🙂

client/components/payments-status/index.tsx
dmallory42: Thanks @oaratovskyi , great catch! 

So, the change to payments being `Enabled` is what I wanted 🙂  previously, it would show as `pending verification` in some cases which I think is misleading because the merchant may think payments are not enabled even though they are actually able to accept payments.

However, the deposits change from `pending verification` to `not connected` is not intended - I think it should stay `pending verification` in both cases. I'll take another look at that logic and see why that happens!

oaratovskyi: > However, the deposits change from pending verification to not connected is not intended - I think it should stay pending verification in both cases. I'll take another look at that logic and see why that happens!

That's because the status that Stripe returns changes from `pending_verification` to something else! Though it shows as Pending on Stripe's side, I know it's confusing 😕

dmallory42: Ah yes, you're right! I think I can fix it by adding a check if the user is in PO for the "disabled" deposits status.

dmallory42: This should be addressed by the latest two commits: https://github.com/Automattic/woocommerce-payments/pull/7665/files#diff-6de3ec2c3cd386ba46aca51c5d7df994a402e52fea7b3ffcbcc1c1d6d887f2fe

Could you take another look and make sure it looks okay to you? Thanks! 🙂

oaratovskyi: Now it works precisely with testing instructions 🚀 Thanks for addressing that!

PR #7671, opened 2023-11-09
Add Playwright e2e test suite ready for incremental migration and visual regression testing

### LINKED ISSUE #8177: Set up Playwright e2e test framework to improve test authoring and maintainability
### Description

This issue aims to set up the fundamental Playwright set up that will allow us to incrementally migrate Puppeteer e2e tests to Playwright.

Migrating our e2e tests to use Playwright instead of Puppeteer will provide us benefits including, but not limited to:
* Reduced test flakiness/increased test stability with [auto-waiting](https://playwright.dev/docs/actionability) and configurable test [retries](https://playwright.dev/docs/test-retries).
* Easier to write robust and maintainable e2e tests with [locators](https://playwright.dev/docs/locators#quick-guide) for targeting DOM elements using testing-library APIs (consistent with integration test APIs).
* [Visual regression testing](https://playwright.dev/docs/test-snapshots) to help us update component library dependencies with more confidence.

### Acceptance criteria

* Playwright e2e tests should be able to be run locally
* Playwright e2e tests should run on GH actions against PRs merged to `develop` or `trunk`
* Documentation in a README.md that will help WooPayments devs get started writing and running tests
* Some example e2e tests and visual comparison tests
* Puppeteer tests should continue to run on GH actions and locally

### Additional information

* See the [WooCommerce Puppeteer to Playwright migration guide](https://github.com/woocommerce/woocommerce/wiki/Puppeteer-to-Playwright-Migration-Guide-for-e2e-Testing) which adds a Playwright test suite alongside the existing Puppeteer test suite.

Jinksi:
Resolves #8177

This PR lays the foundation for Playwright e2e and visual regression testing, ready for incremental migration of existing Puppeteer e2e tests.

Migrating our e2e tests to use Playwright instead of Puppeteer will provide us benefits including, but not limited to:
* Reduced test flakiness/increased test stability with [auto-waiting](https://playwright.dev/docs/actionability) and configurable test [retries](https://playwright.dev/docs/test-retries).
* Easier to write robust and maintainable e2e tests with [locators](https://playwright.dev/docs/locators#quick-guide) for targeting DOM elements using testing-library APIs (consistent with integration test APIs).
* [Visual regression testing](https://playwright.dev/docs/test-snapshots) to help us update component library dependencies with more confidence.

To aid in migration, this PR follows the [WooCommerce Puppeteer to Playwright migration guide](https://github.com/woocommerce/woocommerce/wiki/Puppeteer-to-Playwright-Migration-Guide-for-e2e-Testing) which adds a Playwright test suite alongside the existing Puppeteer test suite.

**TODO**

- [x] Migrate example tests
- [x] Setup GH check
- [x] Add basic tips on how to write PlayWright tests in `tests/e2e-pw/README.md`
- [x] #8160 
- [x] Review and merge this PR 🚀
- [x] Discuss handover with Harmony – FE guild will remain caretakers until a future time TBD
- [x] Post the draft P2 post (paJDYF-cjc-p2) to inform stakeholder teams and encourage adoption (link to `tests/e2e-pw/README.md` for getting started)

**Notes/inclusions:**

* The fundamental Playwright boilerplate and config.
* Authentication setup step for admin & customer with caching for local dev envs.
* Some initial basic tests with visual regression snapshots.
* Uses the existing e2e docker environment setup.
* Uses many of the same conventions from the [WC Puppeteer to Playwright migration guide](https://github.com/woocommerce/woocommerce/wiki/Puppeteer-to-Playwright-Migration-Guide-for-e2e-Testing) (e.g. `e2e-pw` directory).
* A GH action that runs e2e-pw on PRs to `develop` or `trunk` and outputs the Playwright HTML report as a GH action artefact ([see this workflow run](https://github.com/Automattic/woocommerce-payments/actions/runs/7218882881?pr=7922))

image

#### Testing instructions
1. `npm install`
1. Setup e2e environment
	1. `npm run test:e2e-setup`
	1. `npm run test:e2e-up`
1. Run e2e tests with Playwright
	-   `npm run test:e2e-pw` headless run from within a linux docker container.
	-   `npm run test:e2e-pw-ui` runs tests in interactive UI mode from within a linux docker container – recommended for authoring tests and re-running failed tests.
	-   `npm run test:e2e-pw keyword` runs tests only with a specific keyword in the file name, e.g. `dispute` or `checkout`.
	-   `npm run test:e2e-pw --update-snapshots` updates snapshots.
1. Ensure the GH check `E2E Playwright Tests` passes
	- View the summary of the GH check run and download the `playwright-report.zip`
	- Open the `index.html` locally in your browser to view the test report
1. Ensure Puppeteer tests pass

-------------------

- [x] Covered with tests
- [x] Tested on mobile **N/A**

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7671` or branch name `try/playwright-e2e-tests` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=try/playwright-e2e-tests) :rocket:

---

Build info: 

- Latest commit: 9cd9095f5f452f988676fe413251ce9060476533
- Build time: 2024-04-19 01:24:10 UTC

haszari:
> This PR is a prototype of using Playwright for e2e and visual regression testing.

@Jinksi can you add a little summary of the rationale, pitch or benefits of the switch (Puppeteer => Playwright) for WooPayments?

Likely some crossover with Woo core's rationale, but I'm keen to understand how this benefits us (e.g. maybe aligning with Woo is part of the appeal!).

> There were a number of reasons for this migration:

> Increase the reliability/decrease the flakiness of the end-to-end test suite
Simplify the code base, eliminate a good deal of boilerplate code in favour of built-in functions/conventions
Speed up the test suite execution
Position the test suite to use some of the advanced features available in Playwright (cross browser/device, visual testing) - [Woo Puppeteer to Playwright migration guide](https://github.com/woocommerce/woocommerce/wiki/Puppeteer-to-Playwright-Migration-Guide-for-e2e-Testing)

tests/e2e-pw/playwright.config.ts
```diff
@@ -0,0 +1,88 @@
...
+		trace: 'retain-on-failure',
+		video: 'on-first-retry',
+		viewport: { width: 1280, height: 720 },
+		// storageState: 'tests/e2e-pw/storage/state.json',
+	},
+	expect: {
+		toHaveScreenshot: { maxDiffPixels: 100 },
+	},
+	snapshotPathTemplate:
+		'{testDir}/__snapshots__/{testFilePath}/{arg}_{projectName}_{platform}{ext}',
```
Jinksi: `{platform}` results in snapshots that aren't compatible across darwin/linux.

E.g. on macOS, the filename is `Merchant-deposits-Load-the-deposits-list-page-1_chromium_darwin.png`
When the test runs on GH actions, it expects a file `Merchant-deposits-Load-the-deposits-list-page-1_chromium_linux.png`

Should we remove the `{platform}` part of snapshot filenames? This would require screenshots being close enough across both darwin/linux.

Jinksi: I'll test this in 6897f72180a6babfe9ecf584d386dd6e3a9ba239 & 97606bf to see if the e2e-pw GH check on #7922 pass

haszari: Could also recommend/require running tests in known env (e.g. docker container running linux).

Jinksi: It might work, but the GH check e2e env has deposit history that the local env doesn't have:

image

haszari: Aha, so ideally that history would be standardised in the tests container (sounds like this snapshot depends on leaky state and could be more self-contained).

ismaeldcom: >Should we remove the {platform} part of snapshot filenames? This would require screenshots being close enough across both darwin/linux.

Yep, I think we can get rid of the platform part aiming for a common look. We can always adjust the pixel diff to fit slight differences by fonts or similar things.

>It might work, but the GH check e2e env has deposit history that the local env doesn't have

🤔 That makes it trickier, ideally, we should try to have the same env. In the past I used a fully mocked server for visual regression testing, to remove it from the equation, but that will require another big effort. We can try to limit screenshots to more generic parts, for now.

Jinksi: > We can try to limit screenshots to more generic parts, for now.

Yeh, agree we can keep it simple for the MVP of the Playwright PR.

Jinksi:
I've merged the GH action PR #7922 into this branch to simplify the work-in-progress into a single PR.

tests/e2e-pw/playwright.config.ts
Jinksi: I've replaced the deposit list visual snapshot test with a non-state-dependant UI test in a25a555 – mainly to see how different this will be cross-platform, but also this UI was an element that broke in the past due to a wp/components update IIRC.

image

.github/workflows/e2e-pw-pull-request.yml
```diff
@@ -0,0 +1,77 @@
...
+
+            - name: Setup E2E environment
+              uses: ./.github/actions/e2e/env-setup
+
+            - name: Install Playwright
+              shell: bash
+              run: |
+                  echo "::group::Install Playwright"
+                  npx playwright install
+                  echo "::endgroup::"
```
Jinksi: This `npx playwright install` step currently installs more browsers than just Chromium – might be wasteful unless we intend to test against the other browsers.

image

tests/e2e-pw/playwright.config.ts
Jinksi: > see how different this will be cross-platform

The [snapshot test failed](https://github.com/Automattic/woocommerce-payments/actions/runs/7498234661?pr=7671) due to small font-rendering differences. 

```
3533 pixels (ratio 0.01 of all image pixels) are different.
```

image

Linux:
image

Darwin:
image

Jinksi: I've adjusted the `maxDiffPixelRatio` to `0.025` in 889b5b5 (tried `0.0005` in b9cc207 but it failed cross-platform). If it becomes flakey we should run these visual regression tests within a linux container as @haszari mentioned.

image

ismaeldcom: Yes, I was expecting fonts to be an issue between platforms. I usually solve that by using a specific font instead of the system one. Do you know if we could use a linux container and still use the Playwright UI? 🤔

.github/workflows/e2e-pw-pull-request.yml
ismaeldcom: I also noticed it on local. We can be more specific and run it like `npx playwright install chromium` to ensure it only installs a single browser.

tests/e2e-pw/playwright.config.ts
Jinksi: [Running in UI mode from within a Docker container is mentioned in the docs](https://playwright.dev/docs/test-ui-mode#docker--github-codespaces), so might be worth a shot if we think it'd be more reliable – hopefully it doesn't add too much complexity.

There is a [docs page with the recommended path for running tests from browsers in a Docker container](https://playwright.dev/docs/docker).

.github/workflows/e2e-pw-pull-request.yml
Jinksi: Resolved in fd60c3780d461d408b1f7789ed3f16e0002960fe

image

tests/e2e-pw/playwright.config.ts
Jinksi: Getting the network right is proving painful with Docker:

Running this command is almost working, and the PW UI is available at `localhost:8077`.

```sh
docker run -v -p 8077:8077 $PWD:/woopayments -w /woopayments -e BASE_URL=http://host.docker.internal:8084 -it --rm --ipc=host mcr.microsoft.com/playwright:v1.39.0-jammy npm run test:e2e-pw -- --ui-host=0.0.0.0 --ui-port=8077
```

But then PW only gets a buggy version of WP and cannot login to wp-admin. I think we need to tell WP that the site's hostname is `host.docker.internal:8084`.


image

Jinksi: As mentioned in Slack, I think we leave the Docker env part out of this PR's scope and handle it as a follow-up issue. Everything works without running in Docker, although it may make the visual regression tests more reliable.

p1704929040720799/1702046252.009489-slack-C053VQPD74J

tests/e2e-pw/utils/shopper.ts
```diff
@@ -0,0 +1,131 @@
...
+		'iframe#challengeFrame'
+	);
+	let challengeFrame = await challengeFrameHandle.contentFrame();
+	if ( ! challengeFrame ) return;
+
+	// 3DS 1 cards have another iframe enclosing the authorize form
+	if ( cardType.toUpperCase() === '3DS' ) {
+		const acsFrameHandle = await challengeFrame.waitForSelector(
+			'iframe[name="acsFrame"]'
+		);
```
Jinksi: @ismaeldcom  This selector is causing this test to fail. I think it has changed to `name="stripe-challenge-frame"` but yet to verify.

```
Error: frame.waitForSelector: Page closed
=========================== logs ===========================
waiting for locator('iframe[name="acsFrame"]') to be visible
============================================================
```

```suggestion
		const acsFrameHandle = await challengeFrame.waitForSelector(
			'iframe[name="stripe-challenge-frame"]'
		);
```

View test report and trace [here](https://lighthearted-baklava-76fd5a.netlify.app/#?testId=d5355fd83eb21895cd99-11495a16b0f15e6259cc) 

🌟 Note: playwright-report.zip GH check artefact makes this such an easy way to investigate test failures.

Jinksi:
Note @ismaeldcom: I've added todos to this PR description, I think this covers what we need for this FE guild project launch but feel free to edit ✅

Jinksi:
I've added the `github` reporter to the PW config in a30a227. This adds annotations to the "Files changed" tab which may help us identify the cause of test failures more quickly.

tests/e2e-pw/utils/shopper.ts
Jinksi: Welp, after reading the 3DS test util I realise `stripe-challenge-frame` isn't going to solve the maze of nested iframes.

Jinksi: I've attempted to resolve this in 896c367, which fixes the failing test when run locally.

Feel free to revert my changes if not suitable.

Jinksi: 🙈 Nope, looks like more tests are failing now...

Although they are failing due to WooPayments being disabled in `should disable WCPay after confirming, then enable again without confirming` and not re-enabling.

Jinksi: They passed on re-run!

I guess the test `should disable WCPay after confirming, then enable again without confirming` might be a bit flaky.

Jinksi: I notice a related problem. Since PW is running everything in [parallel](https://playwright.dev/docs/test-parallel) when run locally, some merchant tests (e.g. disabling WooPayments) will cause others to fail (e.g. trying to complete checkout while the plugin is disabled).

Due to this, I ~think we need to~ have disabled parallelism locally as well as in CI.

tests/e2e-pw/README.md
```diff
@@ -0,0 +1,88 @@
...
+	const { shopperPage } = await getShopper( browser );
+	const { merchantPage } = await getMerchant( browser );
+
+	// do things as shopper
+	await shopperPage.goto( '/cart/' );
+
+	// do things as merchant
+	await merchantPage.goto( '/wp-admin/admin.php?page=wc-settings' );
+} );
+```
```
Jinksi: While migrating the merchant disputed order notice test, I noticed the problem of switching between `shopper` and `merchant` within a test.

I added util functions to access `shopperPage` and `merchantPage`. What do we think about this pattern?

tests/e2e-pw/playwright.config.ts
Jinksi: Revisiting visual regression testing, I added one for this dispute notice component – font rendering differences will make visual testing unreliable until we fix this.

Should we fix this for this MVP PR or leave it as a followup issue?

image

Jinksi: 🏗️ I've got to a point where running in a Docker env locally works, and the visual snapshots match with the GH Actions snapshot.

To do this:

1. I added the following to `tests/e2e/docker/wordpress/wp-config.php`:

	```php
	define('DOCKER_REQUEST_URL', (!empty($_SERVER['HTTPS']) ? 'https://' : 'http://') . (!empty($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : 'localhost'));
	define('WP_SITEURL', DOCKER_REQUEST_URL);
	define('WP_HOME', DOCKER_REQUEST_URL);
	```

2. Then Run the following commands:
	- UI mode: `docker run -v $PWD:/woopayments -w /woopayments -p 8077:8077 -e BASE_URL=http://host.docker.internal:8084 -it --rm --ipc=host mcr.microsoft.com/playwright:v1.39.0-jammy npm run test:e2e-pw -- --ui-host=0.0.0.0 --ui-port=8077`
	- CLI run: `docker run -v $PWD:/woopayments -w /woopayments -e BASE_URL=http://host.docker.internal:8084 -it --rm --ipc=host mcr.microsoft.com/playwright:v1.39.0-jammy npm run test:e2e-pw`


So if we tidy this up into the npm setup & run scripts, we'll have a better setup for robust visual regression tests.

Jinksi: This thread is getting long, and the Docker part of this may not make it into this MVP PR.

So I've opened #8160 to implement running PW locally from a linux Docker container.

Jinksi: Resolved now that #8160 has been merged into this branch.

tests/e2e-pw/specs/merchant/merchant-payment-gateways-confirmation.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+		// and refreshing the page should show WCPay become disabled
+		await waitForToggleLoading( page );
+		await getSaveButton( page ).click();
+
+		// now we can re-enable it with no issues
+		await getToggle( page ).click();
+		await waitForToggleLoading( page );
+		await getSaveButton( page ).click();
+		await expect( getToggle( page ) ).toHaveText( 'Yes' );
```
Jinksi, line 96: @ismaeldcom this test is being reported as flaky by PW, works on retry.

```
  Locator: getByRole('link', { name: '"WooPayments" payment method is currently' })
    Expected string: "Yes"
    Received string: ""
```

ismaeldcom: Hey @Jinksi can you provide a link to the flaky report? I've tried to find it, but seems to be fine in the latest GH runs.

Jinksi: Yeh, you're right – no flakiness on recent runs 👏 😉

.github/workflows/e2e-pw-pull-request.yml
```diff
@@ -0,0 +1,77 @@
...
+
+            - name: Setup E2E environment
+              uses: ./.github/actions/e2e/env-setup
+
+            - name: Install Playwright
+              shell: bash
+              run: |
+                  echo "::group::Install Playwright"
+                  npx playwright install chromium
+                  echo "::endgroup::"
```
achyuthajoy: @Jinksi Any particular reason to use groups here? Since this step doesn't use composite actions and the step is doing only one thing, I believe we can remove the group.

tests/e2e-pw/playwright.config.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const { BASE_URL } = process.env;
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig( {
+	testDir: './specs/',
+	/* Run tests in files in parallel */
+	fullyParallel: true,
```
achyuthajoy: Shouldn't this be `false` since we're not running tests in parallel and limiting the number of workers to 1?

tests/e2e-pw/test-e2e-pw-ui.sh
```diff
@@ -0,0 +1,14 @@
...
+echo "Open http://localhost:8077 in your browser to see the UI.";
+
+# When updating the Playwright version in the image tage below, make sure to update the npm `@playwright/test` package.json version as well.
+
+docker run -v "$PWD:/woopayments" -w /woopayments \
+	-p 8077:8077 \
+	-e BASE_URL=http://host.docker.internal:8084 \
+	-it --rm --ipc=host \
+	mcr.microsoft.com/playwright:v1.39.0-jammy \
+	npx playwright test --config=tests/e2e-pw/playwright.config.ts --ui --ui-host=0.0.0.0 --ui-port=8077 "$@"
```
achyuthajoy: Do you think it's better to have the docker config outside the shell script? Like the ones we have for Puppeteer E2E tests?

tests/e2e-pw/test-e2e-pw.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo "🎭 Running Playwright e2e tests in default headless mode.";
+
+# When updating the Playwright version in the image tage below, make sure to update the npm `@playwright/test` package.json version as well.
+
+docker run -v "$PWD:/woopayments" -w /woopayments \
+	-e BASE_URL=http://host.docker.internal:8084 \
+	-it --rm --ipc=host \
+	mcr.microsoft.com/playwright:v1.39.0-jammy \
+	npx playwright test --config=tests/e2e-pw/playwright.config.ts "$@"
```
achyuthajoy: Do you think it's better to have the docker config outside the shell script? Like the ones we have for Puppeteer E2E tests?

achyuthajoy (changes requested):
Thanks for working on this @Jinksi. I'm yet to fully test the changes, but left a few comments.

.github/workflows/e2e-pw-pull-request.yml
Jinksi: Thanks, this was probably just a symptom of copy-pasta.

tests/e2e-pw/test-e2e-pw.sh
Jinksi: 👍 I agree, I've had a crack at this via a `docker-compose.yml` file in 660c077, let me know what you think.

tests/e2e-pw/playwright.config.ts
Jinksi: 💯 resolved in d8940c1

.github/workflows/e2e-pw-pull-request.yml
Jinksi: Resolved in b09866a

tests/e2e-pw/test-e2e-pw.sh
Jinksi: ~I assume this has broken the PW GH action – I'll look into it.~

All E2E tests are failing due to #8291.

Jinksi: Now resolved 🙂

Jinksi:
E2E failures due to #8317 – I'll update the card details in the e2e-pw config file once changes have been merged to `develop`.

vbelolapotkov:
@Jinksi thanks for all the work here! I've increased default timeout to 60 secs due to one of the tests failing locally. I checked the UI mode and noticed that setup takes 15-18s out of default 30 for the whole test, and 2-3s extra was needed to complete `should navigate to dispute details when disputed order notice button clicked` test. 

The tests passed locally like a charm.

I've also refactored a bit auth tests hope you don't mind. If you don't like refactors, feel free to revert them.

Jinksi:
> I've also refactored a bit auth tests hope you don't mind. If you don't like refactors, feel free to revert them.

Thanks for taking this for a spin and making these additions @vbelolapotkov 🙌

The auth setup tests was mostly copy-pasta from [WC core](https://github.com/woocommerce/woocommerce/blob/b10be3d8269dfa22a1be21e4da89cc4979553dac/plugins/woocommerce/tests/e2e-pw/global-setup.js#L64-L70) and your changes to using user-facing locators are an improvement.

tests/e2e-pw/README.md
reykjalin: Love this, it can be really tricky to manage the roles in a test and making that straight-forward is essential ❤️

tests/e2e-pw/test-e2e-pw-ui.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
reykjalin: > [!NOTE]
> This is a nitpick.

I think it's typically better to set the crunch bang line to `#!/usr/bin/env bash` (i.e. to use `env`) to make sure `bash` is run in your typical shell environment 🤔 

To be fair though, this didn't really affect the testing for me, but it's what we've done for all our other bash scripts, so I think it'd be better to do this here as well, even if only for consistency 👍

tests/e2e-pw/specs/shopper/shopper-checkout-purchase.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Internal dependencies
+ */
+
+import { config } from '../../config/default';
+import * as shopper from '../../utils/shopper';
+
+test.describe( 'Successful purchase', () => {
+	test.beforeEach( async ( { page } ) => {
+		await page.goto( '/cart/' );
+		await shopper.addCartProduct( page );
```
reykjalin: Shouldn't `shopper.addCartProduct( page )` come first here? I suppose it doesn't _really_ matter, but I think that's a more typical flow? It's a bit odd to go first to the cart page, then to the product page, and then directly to checkout :)

tests/e2e-pw/test-e2e-pw.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo "🎭 Running Playwright e2e tests in default headless mode.";
+
+docker compose -f ./tests/e2e-pw/docker-compose.yml run --rm -it --service-ports playwright \
+	npx playwright test --config=tests/e2e-pw/playwright.config.ts "$@"
```
reykjalin, line 6: > [!NOTE]
> This is a suggestion for future improvement and doesn't necessarily have to be a part of this PR.

One thing I would loooove to see here is a way to run the tests against some arbitrary URLs, if possible. In other words, to remove the dependency on the Docker containers.

As an example, you could run the `npm task` as something like

```sh
npm run test:e2e-pw -- --store_url=http://wcpay.local --server_url=http://wcpayserver.local
```

This enables 3 things:

- No dependency on Docker
- The ability to switch in a different server, e.g. prod, your sandbox, or someone else's exposed environment.
- The ability to switch in a different store, e.g. a production store, or someone else's exposed dev environment.

We have a setup like this for WooPay so we can run the e2e tests there against our local environments and staging without having separate npm tasks.

reykjalin (commented):
I have some observations, but haven't been able to run the tests yet because `npm run test:e2e-setup` fails at the following stage:

```
$ npm run test:e2e-setup

# ...

===> Configuring server with stripe account
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

I don't think that's related to this PR so it's probably something in my environment but I haven't been able to figure out what it is. Maybe y'all have run into this error before?

tests/e2e-pw/test-e2e-pw-ui.sh
Jinksi: Nice catch! Resolved in e808ea2

Jinksi:
Thanks for taking a look @reykjalin 🙌

> `npm run test:e2e-setup` fails
> I don't think that's related to this PR so it's probably something in my environment 

It shouldn't be impacted by this PR, how does this script go on `develop`?

I'd suggest running `npm run test:e2e-reset` and `npm run test:e2e-setup` if you haven't already.

tests/e2e-pw/specs/shopper/shopper-checkout-purchase.spec.ts
Jinksi: cc @ismaeldcom as the author of shopper tests.

ismaeldcom: Good catch! It's a leftover from a different code approach, we don't need to visit the cart with this one 🙂

_Addressed in 7b88a7c_

reykjalin (commented):
This is all looking really good! Got this working in [Rancher desktop](https://rancherdesktop.io/) too, so you don't even need Docker 🙌 

`npm run test:e2e-pw-ui` only showed the `auth` tests though, not sure why that's happening?

Jinksi:
> `npm run test:e2e-pw-ui` only showed the auth tests though, not sure why that's happening?

@reykjalin This _might_ happen when restarting the container (re-running `npm run test:e2e-pw-ui`). A new `ws` key is generated for the UI each time.

E.g. `localhost:8077/trace/uiMode.html?ws=7bebab541dcfed2157412f1461a5bca7`

You can fix this by visited `http://localhost:8077/` without the `ws` key.

reykjalin:
> You can fix this by visited http://localhost:8077/ without the `ws` key

Hmm, that doesn't seem to work 🤔 . `npm run test:e2e-pw` still runs all 13 tests just fine, but `npm run test:e2e-pw-ui` only has the auth tests. Not sure why tbh, but might not be super important?

ismaeldcom:
Hey @reykjalin, that's probably because [Playwright UI only shows the first project by default](https://github.com/microsoft/playwright/issues/21952#issuecomment-1483349947), which is the setup part in our case.

reykjalin (changes requested):
> Hey @reykjalin, that's probably because [Playwright UI only shows the first project by default](https://github.com/microsoft/playwright/issues/21952#issuecomment-1483349947), which is the setup part in our case.

Oh, that's exactly it! I didn't even think to click that, and even if I had I wouldn't have thought to click "chromium". Is there a way we can change the name of the project?

image

***

Also, it seems that `basic.spec.ts` and `auth.setup.ts` serve a very similar purpose. Or at least they're basically the same tests. It's probably enough to keep only one of those? Or somehow merge them?

ismaeldcom:
> Is there a way we can change the name of the project?

I think we could [rename it here](https://github.com/Automattic/woocommerce-payments/blob/7b88a7c2f283e9ca39f91fc7ef009d44bcf5f437/tests/e2e-pw/playwright.config.ts#L51), AFAIK that's the actual name, the browser comes from the `use` property.

Jinksi:
Reading the [docs](https://playwright.dev/docs/api/class-testproject#properties), PW projects lean towards being used to designate testing for multiple browsers, hence the default `chromium` project.

But I've attempted a `merchant`/`shopper`/`setup`/`basic` project outline in 07e31b64351bb79a92ed1466761885e2568d35d0. This results in the snapshots moving into the `/specs/merchant|shopper` dir, which is all good.

image

>  It's probably enough to keep only one of those? Or somehow merge them?

I think `basic` is worth keeping since I've found it to be a "canary" that lets me know that something more fundamental is going wrong with my local env (e.g. if I forget `npm run build`).

I tried it merged with `setup`, but then it runs on every test run, which is a bother since `basic` takes a second or so to run while `setup` is instant after the initial auth.

reykjalin (approved):
LGTM, awesome work 👏 🚀

Jinksi:
@achyuthajoy would you still like to re-review this PR?

achyuthajoy (approved):
Thanks for working @Jinksi. Tested the changes and works well.

LGTM 🚢

Jinksi:
Thanks! I'll aim to merge this ASAP (~Wednesday~ Friday 🤷).

Jinksi:
Failing `Compatibility` and `Code coverage` GH checks are unrelated – due to #8662.

PR #7672, opened 2023-11-09
RPP: Handle failed transaction rate limiter

### LINKED ISSUE #7414: RPP / Verification / Link the transaction limiter
### Description

The existing failed rate transaction limiter should be linked within the payment processing service. This includes both checking the limiter, and incrementing it.

Here is where the current check is performed:

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L868-L873

And it also needs to be bumped upon payment (not system) errors:

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L921-L923

### Acceptance criteria

1. The transaction limiter is bumped and checked.
2. The limiter has 100% test coverage.
3. Move to `src` at your own discretion.

### Testing instructions

1. Enable the limited through the option.
2. Fail enough transactions to reach the threshold.
3. Make sure that the exception is thrown and handled.

### Additional context

Project: paJDYF-9hL-p2

htdat:
Fixes #7414

- Migrate legacy `Session_Rate_Limiter` to `SessionRateLimiter` in `src`. 
- Integrate this new class into `InitialState`.
- Add tests.
- Prepare #7734 to clean up in Jan 2024. 

## Todos

- [x] Get feedback regarding dealing with generating a new class for `SessionRateLimiter`.
- [ ] Get feedback in using `DG\BypassFinals` for testing `final` methods/classes. 

#### Testing instructions

* Use declined cards for checkout https://woo.com/document/woopayments/testing-and-troubleshooting/testing/#declined-test-cards
* After 5 times of failure, see the error message `Your payment was not processed.` in the checkout page. Tip: can change this number `5` to `2` [here](https://github.com/Automattic/woocommerce-payments/blob/f48f4fcde4ce0c481c1996e5d06d84988c63d81f/src/Internal/Payment/FailedTransactionRateLimiter.php#L29-L29) so that it's faster to see the expected result.
* Check the note section of this order, see a relevant note that the rate limiter has been enabled. 
* Put breakpoints in the new code to ensure they're run through during manual tests.

-------------------

**Post merge**


src/Internal/Payment/FailedTransactionRateLimiter.php
```diff
@@ -0,0 +1,47 @@
...
+	 * @param LegacyProxy    $legacy_proxy    LegacyProxy instance.
+	 */
+	public function __construct(
+		SessionService $session_service,
+		LegacyProxy $legacy_proxy
+	) {
+		parent::__construct(
+			'wcpay_card_declined_registry',
+			5,
+			10 * MINUTE_IN_SECONDS,
```
htdat, line 31: For ref https://github.com/Automattic/woocommerce-payments/blob/6750444118550c2c3ef6a5cd4ab00f6f2372196a/includes/class-wc-payments.php#L493

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7672` or branch name `rpp/7414-transaction-limiter` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7414-transaction-limiter) :rocket:

---

Build info: 

- Latest commit: 18113a5f320fcc3724bee23b11448bd6d6496216
- Build time: 2023-11-22 04:45:02 UTC

src/Internal/Payment/FailedTransactionRateLimiter.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace WCPay\Internal\Payment;
+
+use WCPay\Internal\Proxy\LegacyProxy;
+use WCPay\Internal\Service\SessionService;
+
+/**
+ * A wrapper class for keeping track of events in registries, and to trigger a rate limiter after a threshold.
+ */
+class FailedTransactionRateLimiter extends AbstractSessionRateLimiter {
```
htdat, line 16: I'd like to receive the feedback on this inheritance. To be honest, I myself don't like this approach but given that, `SessionRateLimiter` (or in the current PR, it's `AbstractSessionRateLimiter`) depends on `SessionService` and `LegacyProxy`, this approach is allowed to use DI nicely with `PaymentsServiceProvider`. Otherwise, it will be something like this: 

https://github.com/Automattic/woocommerce-payments/pull/7673/files#diff-bd48d09e9d1c96e424dd2cf820a10f9aba6b0b2c9752bf7e9e9fb82c33a69889R74-R85

htdat:
👋 @Automattic/gamma  I'd like to get feedback regarding the redirection to handle this issue first before writing tests for it.

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -70,6 +71,10 @@ public function register(): void {
 		$container->addShared( StateFactory::class )
 			->addArgument( Container::class );
 
+		$container->addShared( FailedTransactionRateLimiter::class )
```
RadoslavGeorgiev: We also need to add this class to `$provides`.

src/Internal/Payment/State/InitialState.php
```diff
@@ -109,6 +121,13 @@ public function start_processing( PaymentRequest $request ) {
 		// Start multiple verification checks.
 		$this->process_order_phone_number();
 
+		if ( $this->failed_transaction_rate_limiter->is_limited() ) {
+			throw new StateTransitionException(
+				__( 'Your payment was not processed.', 'woocommerce-payments' ),
+				400
+			);
```
RadoslavGeorgiev, line 162: In addition to throwing the exception, which will reach the customer, please implement this code from the `catch()` block of the gateway here as well: https://github.com/Automattic/woocommerce-payments/blob/a19f2b0f38abdbce2946103fc9cfabe2d42684da/includes/class-wc-payment-gateway-wcpay.php#L1033-L1051

It's an action that we can perform immediately, and should not be a part of the catch block.

RadoslavGeorgiev (changes requested):

src/Internal/Payment/State/InitialState.php
htdat: Handled c0a8ddff5a3f02dfeb90950d99e90e6861a9c788 - note that I moved most of the logic to generate the note content to `OrderService` to simplify code in `InitialState`. Otherwise, it would be very verbose in `InitialState` and this state requires an additional dependency (`LegacyProxy`).

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
htdat: Handled in 907e4368daf00dca71a1cb996556d4c05b3e0179

htdat:
👋 Ready for another review. Meanwhile, I am going to add PHP unit tests. 

Also, @RadoslavGeorgiev and I had some discussions regarding the approach via DM, I would like to summarize for future reference. 

- ✅ Continue using the current PR and discard PR #7673. Reasons: 
    - Using aliases in the container (PR 7673) would definitely work, but both of us have the feeling that this is too much functional definition for the contaienr to handle. 
    - Use the current approach and mark the public methods of the abstract class as final to clear up my concern regarding inheritance usage.

- ✅ Add a new PR to handle the removal of legacy service after 2023 to avoid any other point releases during the upcoming high-season for ecommmerce industry.
    - Update: PR: #7734

RadoslavGeorgiev (commented):
Looks good so far, let's see it with tests :)

tests/unit/bootstrap.php
```diff
@@ -98,6 +98,13 @@ function() {
 	require_once $_plugin_dir . 'includes/admin/class-wc-rest-payments-payment-intents-controller.php';
 	require_once $_plugin_dir . 'includes/class-woopay-tracker.php';
 	require_once $_plugin_dir . 'includes/admin/class-wc-rest-payments-customer-controller.php';
+
+	DG\BypassFinals::enable( false, true );
+	DG\BypassFinals::setWhitelist(
+		[
+			'*/AbstractSessionRateLimiter.php',
+		]
+	);
```
htdat, line 112: @RadoslavGeorgiev - what do you think of introducing this package to overcome the limitation of PHPUnit and `final` keywords? 

> Please note that final, private, and static methods cannot be stubbed or mocked. 
> -- Source: https://docs.phpunit.de/en/9.6/test-doubles.html

FYI. Woo core is literally using the same approach and package https://github.com/woocommerce/woocommerce/blob/70154d5476254a061734c0bf47259d9a71156aca/plugins/woocommerce/tests/Tools/CodeHacking/CodeHacker.php#L18-L19

htdat: Another approach is to extract an interface for `AbstractSessionRateLimiter` and we're going to provide this interface in `PaymentsServiceProvider` rather than the concrete class `FailedTransactionRateLimiter`.

htdat: The interfact approach is possible, however the downside is that every time we add new methods into `AbstractSessionRateLimiter`, we also need to add it to the interface. It's not a specific issue here as I don't oversee that we will frequently add more methods into  `AbstractSessionRateLimiter`. Nonetheless, that can happen for other future classes using `final` keywords. 

In general, I prefer the approach of using the new package `DG\BypassFinals`. Otherwise, at least we can ask Woo core team to extract their CodeHacker to a PHP composer package so that we can re-use the code in WooPayments.

htdat:
👋 Get ready for next round of review. 
❓ The most outstanding question is the new package for writing tests against `final` methods/classes https://github.com/Automattic/woocommerce-payments/pull/7672#discussion_r1395105789

src/Internal/Payment/FailedTransactionRateLimiter.php
htdat: Resolved, see https://github.com/Automattic/woocommerce-payments/pull/7672#issuecomment-1809921599

tests/unit/bootstrap.php
RadoslavGeorgiev: The `dg/bypass-finals` package looks good to me, excellent job figuring this out!

We might also keep the `CodeHacker` class in mind for the future, at some point, something similar would be good for `src`.

RadoslavGeorgiev (approved):
Flawless work, @htdat! 🙇 :shipit:

PR #7686, opened 2023-11-09
Enable deferred intent creation when initialization process encounters cache unavailability

### LINKED ISSUE #7482: Call to a member function is_reusable() on bool in class-wc-payments.php:768 + e is null
### Describe the bug
After updating WooCommerce Payments, I get an error the first time I land on the shop or checkout page:
Call to a member function is_reusable() on bool in class-wc-payments.php:768
It doesn't trigger again, but the payments modes show empty frames, and trying to pay triggers a notice "e is null".
Only 3 plugins are active: Woocommerce, Woocommerce Payments, Woocommerce Blocks. Theme is 2023.

### To Reproduce

1. Update to WooCommerce Payments 6.6.0
2. Go to the shop or checkout page; it will say something about a critical error; checking the error log will show the error mentioned above.
3. Reload the page; it will work again but will now display empty payment methods.

### Actual behavior
Fatal error

### Screenshots
screenshot-woo-payments

### Expected behavior
No errors and no empty payment method.

* OS: Windows 10
* Browser Firefox 118.0.2

timur27:
Fixes https://github.com/Automattic/woocommerce-payments/issues/7482

When checking the state of the feature flag for the deferred intent creation UPE, sometimes the cache isn't available which makes the [`WC_Payments_Features` return false](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-features.php#L81). If with this setup, a store has legacy UPE enabled, [legacy UPE takes over the initialization](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L548) (instead of [initializing deferred UPE](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments.php#L527) as one would expect), eventually leading to unpredictable state of the registered gateways. This is what the author of https://github.com/Automattic/woocommerce-payments/issues/7482 faced. 

The main change in this PR is to enable deferred UPE when the cache is unavailable for some reason. Cache being unavailable is a rare/edge-case scenario, and the changes in this PR are not negatively impacting the scenario when we disable the dUPE feature flag for a merchant. The flag state is correctly propagated to the plugin if we do so.

While the unit test suite is initialized, it checks UPE flags. Previously, while checking the flag, all the UPEs were disabled, and thus, the legacy gateway was initialized for tests. Now, when we run dUPE in case of a non-available cache, tests start to run deferred intent creation UPE, which led to multiple failures that this PR fixes. I encourage to review the tests in more detail. Some tests were removed, some were replaced, and some context is left in the comments. 

#### Testing instructions

On `develop`
1. Ensure that the server-side flag for dUPE on your local store is enabled
2. Enable client-side FF for legacy UPE
3. In WCPay Dev, refresh the cache and confirm you see a fatal error like described https://github.com/Automattic/woocommerce-payments/issues/7482#issuecomment-1784989547
4. Checkout `fix/more-use-cases-for-enabled-dupe` 
5. Repeat steps 1-3 and confirm that you don't see the error from the third step anymore
6. Perform a sanity check by purchasing a product and confirm that deferred intent creation UPE is being used (e.g. by checking the `gateway_type` field in the Metadata of payment details on the Stripe dashboard)
7. Disable dUPE FF in the server's database and confirm that legacy UPE is used if legacy UPE FF is on, and legacy card is used when no UPE FF is enabled)
8. Run `npm run test:php` on your local to ensure that the changes in `bootstrap.php` don't impact the setup negatively. 
-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7686` or branch name `fix/more-use-cases-for-enabled-dupe` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/more-use-cases-for-enabled-dupe) :rocket:

---

Build info: 

- Latest commit: a540e9d36947e2dbb610a7bf86c15b7ae3e2eb5c
- Build time: 2023-11-17 16:36:22 UTC

tests/unit/migrations/test-class-track-upe-status.php
```diff
@@ -59,21 +59,6 @@ public function test_track_enabled_on_upgrade() {
...
-		$this->assertEquals(
-			[
-				'wcpay_upe_disabled' => [],
-			],
-			Tracker::get_admin_events()
-		);
-
-		$this->assertSame( '1', get_option( Track_Upe_Status::IS_TRACKED_OPTION ) );
-	}
-
```
timur27, line 76: No need to check for `wcpay_upe_disabled` anymore since UPE is always disabled now

tests/unit/notes/test-class-wc-payments-notes-additional-payment-methods.php
```diff
@@ -24,78 +24,6 @@ public function tear_down() {
 		delete_option( '_wcpay_feature_upe' );
 	}
 
-	public function test_get_note() {
```
timur27, line 27: All the non-UPE related tests have been removed from this class, since the plugin now executes UPE by default

tests/unit/notes/test-class-wc-payments-notes-set-up-stripelink.php
```diff
@@ -55,14 +55,6 @@ public function test_stripelink_setup_get_note() {
...
 	}
 
-	public function test_stripelink_setup_note_null_when_upe_disabled() {
-		$this->mock_gateway_data( '0', [ 'card', 'link' ], [ 'card' ] );
-
-		$note = \WC_Payments_Notes_Set_Up_StripeLink::get_note();
-
-		$this->assertNull( $note );
-	}
-
```
timur27, line 65: This is irrelevant going forward due to UPE being always enabled

tests/unit/payment-methods/test-class-upe-payment-gateway.php
```diff
@@ -1888,19 +1888,6 @@ function( $argument ) {
 		}
 	}
 
-	/**
-	 * @dataProvider maybe_filter_gateway_title_data_provider
-	 */
-	public function test_maybe_filter_gateway_title_with_no_additional_feature_flags_enabled( $data ) {
```
timur27, line 1894: [Filtering occurs for non-deferred UPE only](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/payment-methods/class-upe-payment-gateway.php#L1152), so this test is not needed anymore as it will fail because of UPE enabled by default.

tests/unit/test-class-wc-payment-gateway-wcpay.php
```diff
@@ -364,12 +365,6 @@ public function test_attach_exchange_info_to_order_with_different_order_currency
 		$this->assertEquals( 0.853, $order->get_meta( '_wcpay_multi_currency_stripe_exchange_rate' ) );
 	}
 
-	public function test_payment_fields_outputs_fields() {
-		$this->wcpay_gateway->payment_fields();
-
-		$this->expectOutputRegex( '/<div id="wcpay-card-element"><\/div>/' );
```
timur27, line 370: No `wcpay-card-element` is rendered anymore. Instead, [the UPE gateway test tests the correct input element.](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/payment-methods/test-class-upe-payment-gateway.php#L371)

tests/unit/admin/test-class-wc-rest-payments-settings-controller.php
```diff
@@ -451,22 +451,14 @@ public function test_upe_update_settings_saves_enabled_payment_methods() {
...
-			$request->set_param( 'enabled_payment_method_ids', [ Payment_Method::CARD, Payment_Method::GIROPAY ] );
-
-			$this->upe_split_controller->update_settings( $request );
-
-			$this->assertEquals( [ Payment_Method::CARD, Payment_Method::GIROPAY ], $this->mock_upe_split_payment_gateway->get_option( 'upe_enabled_payment_method_ids' ) );
-	}
+		$request = new WP_REST_Request();
+		$request->set_param( 'enabled_payment_method_ids', [ Payment_Method::CARD, Payment_Method::GIROPAY ] );
 
-	public function test_update_settings_validation_fails_if_invalid_gateway_id_supplied() {
```
timur27, line 464: This test was removed because it's an integration test that works only with the legacy gateway. It's not easy nor elegant to mock gateway in `WC_Payments` because methods/objects are static. 

Since the tested method relies solely on [`get_upe_available_payment_methods`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/admin/class-wc-rest-payments-settings-controller.php#L94C40-L94C73), `get_upe_available_payment_methods` is now fully tested by `test-class-upe-payment-gateway.php::test_get_upe_available_payment_methods()`

tests/unit/bootstrap.php
```diff
@@ -98,6 +98,9 @@ function() {
 	require_once $_plugin_dir . 'includes/admin/class-wc-rest-payments-payment-intents-controller.php';
 	require_once $_plugin_dir . 'includes/class-woopay-tracker.php';
 	require_once $_plugin_dir . 'includes/admin/class-wc-rest-payments-customer-controller.php';
+
+	// Load currency helper class early to ensure its implementation is used over the one resolved during further test initialization.
+	require_once __DIR__ . '/helpers/class-wc-helper-site-currency.php';
```
timur27, line 103: More context on this change in p1700139623363929-slack-CU6SYV31A

includes/class-wc-payments-features.php
```diff
@@ -78,6 +78,9 @@ public static function is_upe_split_enabled() {
 	 */
 	public static function is_upe_deferred_intent_enabled() {
 		$account = WC_Payments::get_database_cache()->get( WCPay\Database_Cache::ACCOUNT_KEY, true );
+		if ( null === $account ) {
+			return true;
+		}
 		return is_array( $account ) && ( $account[ self::DEFERRED_UPE_SERVER_FLAG_NAME ] ?? false );
```
frosso: Side question - would it also make sense to return `true` in case `$account[ self::DEFERRED_UPE_SERVER_FLAG_NAME ]` is not set?

i.e.:
```suggestion
		return is_array( $account ) && ( $account[ self::DEFERRED_UPE_SERVER_FLAG_NAME ] ?? true );
```

Asking because we're basically saying: "if there's no account data - force dUPE". But if `$account` data is an array without the `is_deferred_intent_creation_upe_enabled` key, we're saying that dUPE is disabled. Am I getting my neurons too twisted? 😵

frosso (approved):
Changes look good, thanks for your diligence with the UTs! Just a minor, non-blocking question.

includes/class-wc-payments-features.php
timur27: You're right and https://github.com/Automattic/woocommerce-payments/pull/7686/commits/abc7347b03a246c0caf99daf5973562cc5f4bea7 improves this together with removing one redundant test which came up after the change, thank you!

The scenario with returning disabled dUPE state wouldn't happen because [server always returns a value](https://github.com/Automattic/woocommerce-payments-server/blob/trunk/server/wp-content/rest-api-plugins/endpoints/wcpay/service/class-account-service.php#L1976-L1978), but this doesn't mean client can rely on this so I agree with you.

frosso (approved):

timur27:
Performed a quick successful sanity check after https://github.com/Automattic/woocommerce-payments/commit/abc7347b03a246c0caf99daf5973562cc5f4bea7. Merging.

Issue #7425, opened 2023-10-06
RPP / Failures / Catch exceptions and generate erroneous states


RadoslavGeorgiev:
### Description

The current payment process throws exceptions upon all sorts of failure, and has a single block to catch all of them.

The new payment process should handle exceptions at a more granular level, and switch to a failure state immediately. 

### Acceptance criteria

- Generic exceptions (`Exception`, `Order_Not_Found`, etc.) are handled within the payment processing service, rather than within a state, entering `SystemErrorState`.
- Specific exceptions (ex. those around requests) should be handled immediately.
- The error states should contain the needed functionality to log the errors, and mark the order as failed if needed.

### Testing instructions

All thrown exceptions should be caught.

### Dev notes

For now, we can focus only on the exceptions we are already aware of through Psalm.

### Additional context

Epic: https://github.com/Automattic/woocommerce-payments/issues/6673
Project: paJDYF-9hL-p2

RadoslavGeorgiev:
Marking as blocked because of Gamma's shift of current priorities.

frosso:
See closing reason in the parent epic: https://github.com/Automattic/woocommerce-payments/issues/6673

PR #7696, opened 2023-11-12
RPP: Handle fraud prevention service in InitialState

### LINKED ISSUE #7413: RPP / Verification / Link the existing fraud prevention service
### Description

Adapt the new payment process to the fraud prevention service. This code should work within the new payment process:

https://github.com/Automattic/woocommerce-payments/blob/c1f7b63d4102703841099f5e2234b720e431a483/includes/class-wc-payment-gateway-wcpay.php#L857-L866

### Acceptance criteria

1. The `is_card_testing_protection_eligible` gateway method is moved to the service. The service can depend on the account class, and work directly with it.
2. We can get rid of the static `get_instance()` and `set_instance()` pair. Looks like `get_instance()` is used as a shortcut to avoid proper constructor dependencies. It also accepts a session, but that's just for testing, and can remain optional as `null`. `set_instance()` on the other hand, is there just for testing. Moving the service to `src` would allow us to use `wcpay_get_container()` and `wcpay_get_test_container()` mocks.
3. Tests of the service should be improved by using the legacy proxy.

### Testing instructions

This one should be figured out. Seems like tampering with the token should allow use to test functionality.

### Additional context

Project: paJDYF-9hL-p2

htdat:
Fixes #7413

1 - Duplicate legacy `Fraud_Prevention_Service` to `src` with the following changes - commit d09463f984861ab01fd546982534956a4d435543, and then integrate it into `InitialState`:  

- Remove `set_instance` and `get_instance` as we no longer need singleton pattern.
- Change constructor args to `SessionService` and `WC_Payments_Account`.
- Simplify `get_token`.
- Correct param type for `verify_token`.

2 - The removal part of the legacy will be handled in Jan 2024 to avoid any potential issue for merchants during high-season in Nov and Dec 2023. PR here: https://github.com/Automattic/woocommerce-payments/pull/7697

3 - Introduce `PaymentProcessingException`, which would be used for other exceptions during the verification state.

#### Testing instructions

- Enable `card_testing_prevention_enabled` on server to activate this feature.
- Add a breakpoint to the new code in `InitialState` to confirm they're really run into.
- Ideally test for both classic and block checkouts.
- To throw exception/error, a simple trick is to run this JS code in browser console: 

```js
document.querySelector('#wcpay-fraud-prevention-token').value += '_change';
``` 

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7696` or branch name `rpp/7413-fraud-service` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp/7413-fraud-service) :rocket:

---

Build info: 

- Latest commit: 0e1f6121906ac46d1876e27519bfed5bf4bc9c71
- Build time: 2023-11-22 03:41:44 UTC

htdat:
👋 I make this PR is ready for review to get initial feedback before adding and fixing PHP tests.

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
```diff
@@ -84,13 +85,18 @@ public function register(): void {
 			->addArgument( HooksProxy::class )
 			->addArgument( LegacyProxy::class );
 
+		$container->addShared( PaymentFraudPreventionService::class )
```
RadoslavGeorgiev: Don't forget to add this one to `$provides` :)

src/Internal/Payment/State/InitialState.php
```diff
@@ -109,6 +121,13 @@ public function start_processing( PaymentRequest $request ) {
 		// Start multiple verification checks.
 		$this->process_order_phone_number();
 
+		if ( $this->fraud_prevention_service->is_enabled()
+			&& ! $this->fraud_prevention_service->verify_token( $this->get_context()->get_fraud_prevention_token() ) ) {
```
RadoslavGeorgiev: This is a pet peeve of mine, but could you please integrate `is_enabled` within `verify_token` of the service? I don't see why external callers would need to check both. Then this would be a bit cleaner:



```suggestion
		if ( ! $this->fraud_prevention_service->verify_token( $this->get_context()->get_fraud_prevention_token() ) ) {
```

src/Internal/Payment/State/InitialState.php
```diff
@@ -109,6 +121,13 @@ public function start_processing( PaymentRequest $request ) {
...
+			throw new PaymentProcessingException(
```
RadoslavGeorgiev: Alright, you got me here. Let's use `StateTransitionException` for now, to avoid adding a new exception class. We'll do better with @zmaglica in https://github.com/Automattic/woocommerce-payments/issues/7425.

src/Internal/Service/PaymentFraudPreventionService.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+namespace WCPay\Internal\Service;
+
+use WC_Payments_Account;
+
+/**
+ * Class PaymentFraudPreventionService
+ */
+class PaymentFraudPreventionService {
```
RadoslavGeorgiev: Any reason not to just call it `FraudPreventionService`?
```suggestion
class FraudPreventionService {
```

src/Internal/Service/PaymentFraudPreventionService.php
```diff
@@ -0,0 +1,92 @@
...
+	 * @var SessionService
+	 */
+	private $session_service;
+
+	/**
+	 * Instance of WC_Payments_Account.
+	 *
+	 * @var WC_Payments_Account
+	 */
+	private $legacy_account_service;
```
RadoslavGeorgiev: Within the payment process so far, we have not used `legacy_`, unless it's the "same" service, ex. `legacy_service` within `WCPay\Internal\Service\OrderService`. Any reason to explicitly mention it here?

RadoslavGeorgiev (changes requested):
Looks pretty good so far, I only have a couple of comments:

src/Internal/Service/PaymentFraudPreventionService.php
htdat: Nothing specific, just my feeling that is better to carry the meaning. I will update it back to `$account_service`.

_Update: addressed 5502a670557fc49b0d4ca78aa5f03dc718645517_

src/Internal/Service/PaymentFraudPreventionService.php
htdat: Just my own preference to let it a bit more specific when seeing a few other services in `includes/fraud-prevent` folder. 

I am not too adherent here so will make change except that makes sense to you too.

_Update: Resolved 96509c9acc58fa41706547206b5abe08a29efae1_

src/Internal/Payment/State/InitialState.php
htdat: yeah, that's something a bit strange under my eyes too. Will update it.

_Update_: addressed c712223b046e67de08ca6094f505ec7394a9a231

src/Internal/DependencyManagement/ServiceProvider/PaymentsServiceProvider.php
htdat: 🤦 yeah, will do it. 

In my manual tests, there is no issue at all. Wonder that makes any difference? 🤔 Without providing this class in `$provides`, I can not statically get it in `includes`, something like the following?

```php
 wcpay_get_container()->get( PaymentFraudPreventionService::class );
```

RadoslavGeorgiev: So the logic is as follows:

1. Class is needed.
2. The container goes through all service providers, and checks if they provide the class.
3. If a service provider provides the class, its `register()` method is called once to register all the classes it provides.
4. Class is used.

> Wonder that makes any difference? 🤔

In most cases, we end up using another service from the same provider as the root. In this case, the `PaymentProcessingService` is always first, and it forces everything else to be loaded. In short: There is not much of a practical difference. It could still cause weird side-effects someday, so I guess it's better to have it.

htdat: Resolved 96509c9acc58fa41706547206b5abe08a29efae1

src/Internal/Payment/State/InitialState.php
htdat: Resolved a50c4a40993a7929e7e0c16e31b655eca82aaaa6

htdat:
👋 @RadoslavGeorgiev - Addressed all feedback, and added tests. Get ready for next review round.

tests/unit/src/Internal/Service/FraudPreventionServiceTest.php
```diff
@@ -0,0 +1,160 @@
...
+			->method( 'get' )
+			->with( FraudPreventionService::TOKEN_NAME )
+			->willReturn( $token_stub );
+
+		$this->assertSame( $token_stub, $this->sut->get_token() );
+	}
+
+	public function test_get_token__on_first_page_load() {
+		$new_token_stub = 'new-token';
+		$this->mock_sut = $this->getMockBuilder( FraudPreventionService::class )
```
RadoslavGeorgiev, line 91: Could you please define `mock_sut` as a property? VSCode does not understand this :D

RadoslavGeorgiev (approved):
Another (essentially) flawless one! 💯 :shipit: 

Only one small nitpick, no further review required.

tests/unit/src/Internal/Service/FraudPreventionServiceTest.php
htdat: Yup, good point. Addressed 0e1f612

PR #5964, opened 2023-04-04
Fix #5963. GPay shows $0 for subscription_variation products

ecairol:
Fixes #5963

#### Changes proposed in this Pull Request

This change takes into account `subscription_variation` when re-adding products to the cart in the `ajax_add_to_cart()` function.

A similar fix was applied on the function `WC_Payments_Payment_Request_Button_Handler::get_product_price()` on the same file.

#### Testing instructions

1. Go to the product page of a subscription. Make sure the product type is subscription_variation and has free shipping
2. Click on Google Pay button
3. Check the totals, should have the Product price
4. Select a shipping method
5. Check the totals again, the product price is $0

Screen Shot 2023-04-03 at 13 41 58

Screen Shot 2023-04-03 at 14 10 52

-------------------

- [ ] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7702, opened 2023-11-13
Support "variation" products on ajax_add_to_cart for Google/Apple Pay checkout

### LINKED ISSUE #7695: Google/Apple Pay total is $0 if product type is variation
### Describe the bug
If a Product type is `variation`, the totals is reset to $0. The same behavior happens on Apple Pay.

### To Reproduce
1. Go to this site: https://novoslabs-dev-optimization.mystagingwebsite.com/product/novos-core/
2. Click on `Individual`
3. Click on `Google Pay` or `Apple Pay` button
5. If there is not selected a shipping method, enter or select one
6. The total price is changed to $0

### Actual behavior
Total amount on the Google/Apple Pay modal is $0

### Screenshots
Screenshot 2023-11-10 at 14 10 08

### Expected behavior
The total amount calculation should take into account products with the type `variation`. This will ensure that the product is re-added to the cart during the `ajax_add_to_cart()` implementation.

* OS: MacOS
* Browser Chrome
* Version 118.0

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Browser [e.g. stock browser, safari]
* Version [e.g. 22]

### Additional context

- This is pretty much the same issue as https://github.com/Automattic/woocommerce-payments/issues/5963, just a different product type to give support. 

- This is also related to https://github.com/Automattic/woocommerce-payments/issues/7588

ecairol:
Support "variation" products on ajax_add_to_cart for Google/Apple Pay checkout

Fixes #7695

Products of type `variation` are now supported on the `ajax_add_to_cart()` implementation, for Google/Apple Pay checkout.

The function add_to_cart empties the cart and then re-adds the products one by one. If the product however, is not in a whitelist, it won't be added back. This is happening to products of type `variation`, which are now included in the array of supported product types.

Before:
Screenshot 2023-11-10 at 09 49 11

After:
Screenshot 2023-11-13 at 09 06 39

#### Testing instructions
- Go to https://novoslabs-dev-optimization.mystagingwebsite.com/product/novos-core/
- Click on `Individual" which is a `variation` product
- Click on the Google/Apple Pay button below

-------------------

**Post merge**


includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1129,7 +1129,9 @@ public function ajax_add_to_cart() {
 			define( 'WOOCOMMERCE_CART', true );
 		}
 
-		$subscription_types = [
+		$supported_product_types = [
+			'simple',
+			'variation',
 			'subscription',
 			'subscription_variation',
```
ecairol: I just wonder if we should get rid of this array altogether and replace it with `$this->supported_product_types()` which includes additional product types like `booking`, `bundle` and `composite`. 

However, I'm not sure whether these product types are supported by Google/Apple Pay or if they were excluded for a reason.

botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7702` or branch name `fix/7695-total-0-for-variation-products-gpay` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7695-total-0-for-variation-products-gpay) :rocket:

---

Build info: 

- Latest commit: 4a19ce980732597ef45fea132490e21bd0009255
- Build time: 2023-11-20 15:46:21 UTC

includes/class-wc-payments-payment-request-button-handler.php
frosso: I don't think we can use `$this->supported_product_types()` directly, in this scenario. I see that [in this code block](https://github.com/Automattic/woocommerce-payments/blob/fix/7695-total-0-for-variation-products-gpay/includes/class-wc-payments-payment-request-button-handler.php#L1149-L1156), we are handling the logic slightly differently than [in this other code block](https://github.com/Automattic/woocommerce-payments/blob/fix/7695-total-0-for-variation-products-gpay/includes/class-wc-payments-payment-request-button-handler.php#L1158-L1160).

If this array is renamed to `$supported_product_types` from `$subscription_types`, I think we should clean up a few things try and find a way to handle both scenarios.

I see that `$subscription_types` was added here: https://github.com/Automattic/woocommerce-payments/pull/5964 , with the intent of targeting "simpler" types (i.e.: products that don't have child variations).

I don't have a better naming for this array - maybe it could also be inlined, since it's not used anywhere else.

frosso (commented):
Thanks for contributing to the fix! I added a small suggestion. You'll also need to add a changelog entry - otherwise the GH checks won't be happy :D

frosso (approved):
Thank you for the changes! Looks good!

ecairol:
Thanks for the suggestion @frosso – I agree that an inline array makes sense here, I've added that and the changelog entry in this commit here: https://github.com/Automattic/woocommerce-payments/pull/7702/commits/c6445fa61aa5d5c6646517d765a99ebd7292bb12

Please let me know if you have any more feedback.

--- 

Sharing an idea here: what if inside `ajax_add_to_cart()` we count how many items are there on the cart, and then after the cart is emptied and all the items are re-added, we count again how many items we have so that if there's a mismatch we log it somewhere for the dev team? Ideally, we shouldn't get any entries on the log, but if there are any we would get visibility of which other product types are being affected.

My reasoning here is that this $0 issue has happened 2 times for one of our partners (Team51 "customers"), so I wonder how common are other scenarios that we are unaware of. Is there anything similar to a "global log" where we can send this kind of stuff?

frosso:
@ecairol sorry I missed your comments, last week.
I don't have further feedback.

> Is there anything similar to a "global log" where we can send this kind of stuff?

I like your idea to ensure this issue doesn't happen again. I am not aware of logs on the client-side (plugin-side) that can be sent to our systems. If the error happened on the server-side (WC Payments Server), we could log to logstash, but this isn't the case.
You could use the "usual" `Logger` utility, or the order notes. But data extrapolation would require `wp-admin` access.

ecairol:
@frosso no worries.

> I am not aware of logs on the client-side (plugin-side) that can be sent to our systems. If the error happened on the server-side (WC Payments Server), we could log to logstash, but this isn't the case.

Ok, that makes sense, I thought it was worth asking. Thanks!

PR #7739, opened 2023-11-19
RPP: Added error states

### LINKED ISSUE #7425: RPP / Failures / Catch exceptions and generate erroneous states

zmaglica:
Fixes #7425 

This PR introduces error states and their management. It ensures that all exceptions, including minor ones, are effectively handled. The core principle is to avoid returning exceptions from states. Instead, states should consistently return their designated state. This approach eliminates the need for exception handling in services utilizing these states or in the gateway, where they were previously handled.

Error states:

- `AbstractPaymentErrorState` : Abstract error class that holds common or reusable logic for all error states
- `PaymentErrorState`: Error state that is used when an error occurs on the buyer's side, which cannot be fixed through code.
- `PaymentRequestErrorState`: Handles `PaymentRequestException` that can occur when context is populated from the request.
- `SystemErrorState`: Equivalent of \Exception we have in PHP but used for our state machine. Look at this state as a default error state when an exception or any unwanted flow occurs, where we don't have a specific error state to transition.
- `WooPaymentsApiServerErrorState`: Error state that is used when server API throws an exception. I used WooPayments prefix to make it more clearer that this is the error state for API server errors on WooCommerce Payments server.

#### Testing instructions

* Make sure that all exceptions are handled by the error states. In other words, if an exception is triggered in state, the state needs to handle it and return error state. The only exception that can be handled without the error state is `StateTransitionException` which can occur outside state processing or within the catch block, where we transition from state to new error state.
* All error states should extend the base or abstract error class.
* Make sure that all error states have their logging and mark order as failed function configured property, and according to the business logic. What that means is that each error state can have disabled logging or disabled marking order as failed.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7739` or branch name `rpp-7425-handle-error-states` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=rpp-7425-handle-error-states) :rocket:

---

Build info: 

- Latest commit: 64683232977c02bc6012f133a10c63faa4e5c880
- Build time: 2023-11-21 13:23:57 UTC

src/Internal/Payment/State/AbstractPaymentErrorState.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+		throw new Exception( __( 'The payment process could not be completed.', 'woocommerce-payments' ) );
+	}
+
+	/**
+	 * Determines whether an error should be logged.
+	 *
+	 * @return bool True if the error should be logged, otherwise false.
+	 */
+	protected function should_log_error(): bool {
```
jessy-p, line 79: Shouldn't all errors be logged by default?

jessy-p (commented):
Under testing instructions, it would be nice to have a listing of the errors and maybe a one liner of the scenarios to test.

src/Internal/Payment/State/AbstractPaymentErrorState.php
zmaglica: I think they need to, but I wasn't 100% certain, so I made it to false. We can easily set that all error states needs to be logged, by the default.

zmaglica:
> Under testing instructions, it would be nice to have a listing of the errors and maybe a one liner of the scenarios to test.

I have updated the test instructions and added a small description to each error state.

src/Internal/Payment/PaymentContext.php
```diff
@@ -37,6 +37,13 @@ class PaymentContext {
 	 */
 	private $transitions = [];
 
+	/**
+	 * Stores exception that is triggered before error state is created.
+	 *
+	 * @var \Exception
+	 */
+	private $exception;
```
RadoslavGeorgiev, line 45: Like request classes, there is no need to define a property of the object: It's all stored within `data`. With that in mind, this one is not used, and can be removed.

src/Internal/Payment/PaymentContext.php
```diff
@@ -300,6 +307,24 @@ public function get_transitions(): array {
 		return $this->transitions;
 	}
 
+	/**
+	 * Stores state exception.
+	 *
+	 * @param \Exception $exception Exception to store.
+	 */
+	public function set_exception( $exception ) {
```
RadoslavGeorgiev, line 315: I'm not insisting on this, but I'd prefer to call those `set_last_exception` and `get_last_exception`, similarly to how there's a `$wpdb->last_error`.

Also, please add a proper type hint to the method:
```suggestion
	public function set_exception( Exception $exception ) {
```

Similar,y `get_(last_)exception()` should return `: ?Exception`

src/Internal/Payment/State/InitialState.php
```diff
@@ -106,49 +106,44 @@ public function __construct(
...
+			if ( null !== $duplicate_payment_result ) {
+				return $duplicate_payment_result;
+			}
+
+			/**
+			 * Payments are based on intents, and intents use customer objects for billing details.
+			 *
+			 * The customer is created/updated right before requesting the creation of
+			 * a payment intent, and the two actions must be adjacent to each-other.
+			 */
```
RadoslavGeorgiev, line 138: This docblock is completely out of place now.

src/Internal/Payment/State/InitialState.php
```diff
@@ -161,23 +156,34 @@ public function start_processing( PaymentRequest $request ) {
...
-		// All good. Proceed to processed state.
-		$next_state = $this->create_state( ProcessedState::class );
+			// All good. Proceed to processed state.
+			$next_state = $this->create_state( ProcessedState::class );
 
-		return $next_state->complete_processing();
+			return $next_state->complete_processing();
+		} catch ( Amount_Too_Small_Exception $e ) {
+			$this->minimum_amount_service->store_amount_from_exception( $e );
+			return $this->create_error_state( SystemErrorState::class, $e );
```
RadoslavGeorgiev, line 172: The amount being too small is not a system exception, it happens whenever the amount to pay is less than the required minimum. This is not an indication that there's something wrong with the system, just a payment error.

src/Internal/Payment/State/InitialState.php
```diff
@@ -161,23 +156,34 @@ public function start_processing( PaymentRequest $request ) {
...
+				return $this->create_state( AuthenticationRequiredState::class );
+			}
 
-		// All good. Proceed to processed state.
-		$next_state = $this->create_state( ProcessedState::class );
+			// All good. Proceed to processed state.
+			$next_state = $this->create_state( ProcessedState::class );
 
-		return $next_state->complete_processing();
+			return $next_state->complete_processing();
```
RadoslavGeorgiev, line 169: By positioning this line within the `catch` block of this state, we're essentially trying to handle all exceptions from the `ProcessedState` here as well. This goes against the idea of granularly handling exceptions.

src/Internal/Payment/State/AbstractPaymentErrorState.php
```diff
@@ -0,0 +1,91 @@
...
+		return false;
+	}
```
RadoslavGeorgiev, line 81: Logging of the whole context is already built, and specific errors (ex. API errors) are logged already. I'd remove the logging functionality from erroneous states altogether.

RadoslavGeorgiev (changes requested):
I have not reviewed all the details regarding the PR, but I think we need a significant change in direction. Keep this line from the issue in mind:

> The new payment process should handle exceptions **at a more granular level**, and switch to a failure state immediately.

The current implementation in the PR goes against granularity, as we're back to the same state as the current payment process, where all exceptions are caught in the same place.

A few notes:
- `StateTransitionException`s were a placeholder for exceptions in specific scenarios. We didn't want to create too many separate exception classes, in case they needed to be replaced by this stream. However, all but the `create_state()`'s `StateTransitionException`s should be replaced by something else.
- Please merge `develop` into the PR, there are some more exceptions after Dat merged some PRs.

Presently we're not storing the context anywhere, but keep this question in mind: If we were to store payments, how would we approach erroneous states when we need to retry those payments?

dmvrtx:
There has been no progress on this pull request since November 2023. While it suggests an approach to improve the payment flow, it is now outdated with numerous conflicts. I'm closing this pull request. If the issue will be worked on, it should be done anew.

PR #7742, opened 2023-11-20
Refund transaction from details page

### LINKED ISSUE #7248: Refund transaction from the transaction details screen
### Description

As a merchant I need to be able to refund transactions directly from the WooPayments transaction details screen:
 - If transaction can be refunded, I need to clearly see how to start a refund.
 - I want a linked order (if present) to change status to refunded if I refund all connected transactions, i.e. there are no other completed and non-refunded transactions with the same order_id.
 - I want to see information about the issued refund in the order notes if transaction is linked to an order.

### Acceptance criteria

 - Refund action is presented on the Transaction Details screen if transaction can be refunded.
 - Refund action asks for refund reasons and creates a refund request.
 - Order changes status to "refunded" if all the linked transactions have been fully refunded.
 - Order notes indicate information about issued refunds.

### Designs

 - paJDYF-9sf-p2

KarlisJ:
Fixes #7248 

Adds a proxy endpoint to call server's refund endpoint directly from transaction details page bypassing.

- If there is no associated order for the transaction -> display a notice with the option to do a full refund. This option proxies directly to Stripe, bypassing WC's built-in refunding logic since there's no order to add any notes.
- Dropdown menu to do a full refund. If we have associated order, then we'll call WC build-in refund logic and will add notes to order
- Dropwon menu option to do a partial refund. Redirects to "edit order" page to perform partial refund.

#### Testing instructions

- Make payment and check it's details page
- Choose "Refund in full"
- The order and transaction should be refunded

- Make payment and check it's details page
- Choose "Partial refund"
- Assert you can do partial refund from "edit order" page

- Make payment and check it's details page
- add `delete charge.order` at the top of `PaymentDetailsSummary` component to simulate missing order case.
- Assert that there's no option to do a partial refund
- Assert that there's "missing order" notice
- Assert can refund from notice and dropdown.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7742` or branch name `add/7248-refund-transaction-from-details-page` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=add/7248-refund-transaction-from-details-page) :rocket:

---

Build info: 

- Latest commit: e18e4d46fed1fec24a45b7faf8734790360229ec
- Build time: 2023-12-20 14:16:45 UTC

KarlisJ:
It's not clear from the designs how we should handle the feedback from refund response. 

In case of error, things are more straightforward. We'll show an error notice in the frontend.

In case of success, there are more questions. 
- How does the UI of the whole page changes after the transaction is refunded? 
   - Do we only disable the "Refund" button?
   - Do we need to update more information in the details page?
   - Can we even display refunded transaction in the details page? If not, should we redirect to transactions list with success notice?
- ~~How can we mark the transaction as refunded in the backend so that we don't show the "refund" button on page refresh?~~  ?~~It's already available in the timeline notes?~~ It's on the charge object

kalessil:
> How does the UI of the whole page changes after the transaction is refunded? 

@KarlisJ: same behaviour as for "Capture" (cc @mgascam to ensure I'm not mistaken):
- prompt success (left bottom)
- hide the notice (you might have to trigger refreshing transaction summary via React)

mgascam:
> > How does the UI of the whole page changes after the transaction is refunded?
> 
> @KarlisJ: same behaviour as for "Capture" (cc @mgascam to ensure I'm not mistaken):
> 
> * prompt success (left bottom)
> * hide the notice (you might have to trigger refreshing transaction summary via React)

Thanks folks, adding my 2 cents:

- It would be nice to have a loading indicator while the refund is processed. The Capture button uses an animation we might be able to use in the refund modal.
- I'd dismiss the modal upong sucessful refund.
- Concur with @kalessil, there should be 'toast' notification for success and error and  the UI components ( Payment info  timeline and details) in this screen should reload upon success.

kalessil:
@KarlisJ : updating PR status, it doesn't look like ready for a review (previous clarifications, also I don't transaction reason is being transferred to the server).

KarlisJ:
Thank you both @mgascam and @kalessil for input. It's much clearer to me how to proceed with UI/UX now

KarlisJ:
@mgascam I've implemented all the feedback and PR is ready for proper review. 

One thing I noticed when doing the testing is that the notice says 
> This transaction is not connected to order. Investigate this purchase and refund the transaction as needed.

after the refund which doesn't make sense. We could probably remove 2nd sentence from there if payment is refunded

includes/admin/class-wc-rest-payments-refunds-controller.php
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	public function process_refund( $request ) {
+		$charge_id = $request->get_param( 'charge_id' );
+		$amount    = $request->get_param( 'amount' );
+
+		try {
+			$refund_request = Refund_Charge::create( $charge_id );
+			$refund_request->set_charge( $charge_id );
+			$refund_request->set_amount( $amount );
+			$refund_request->set_reason( $amount );
```
dmvrtx: There's definitely a typo here 😅

client/payment-details/summary/index.tsx
```diff
@@ -11,6 +11,8 @@ import {
 	CardFooter,
 	CardDivider,
 	Flex,
+	DropdownMenu,
```
mgascam, line 14: I'd remove all references to DropdownMenu

client/payment-details/summary/index.tsx
```diff
@@ -225,6 +228,18 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 		balance.currency
 	);
 
+	const dropDownButtons = [
```
mgascam: This can be removed as well

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -18,20 +18,29 @@ import './style.scss';
 import CardNotice from 'wcpay/components/card-notice';
 import ConfirmationModal from 'wcpay/components/confirmation-modal';
```
mgascam: In retrospective, it would have been better to implement the `ConfirmationModal` in the parent, since we'll need the same modal for the refund from dropdown. Can be extracted now or when we implement the refund from dropdown.

mgascam (commented):
Pre-approving as manual tests work as intended. Left some comments, related to the implementation of the "refund from dropdown"  that can be tackled now or on a subsequent pull request.

Re

> We could probably remove 2nd sentence from there if payment is refunded

I guess the message is subject to interpretation. As non-native English speaker, I interpret it like, "refund the transaction if needed". But maybe is easier to just remove the "refund the transactions as needed".

KarlisJ:
@mgascam 
> I guess the message is subject to interpretation. As non-native English speaker, I interpret it like, "refund the transaction if needed". But maybe is easier to just remove the "refund the transactions as needed".

My point was about the fact that after successfully refunding the transaction there's no "refund" button anymore but the text still says ".. and refund the transaction as needed." which is illogical since you can't refund transaction that's already refunded.

includes/admin/class-wc-rest-payments-refunds-controller.php
```diff
@@ -0,0 +1,71 @@
...
+		$amount    = $request->get_param( 'amount' );
+		$reason    = $request->get_param( 'reason' );
+
+		if ( $order_id ) {
+			$order   = wc_get_order( $order_id );
+			$gateway = wc_get_payment_gateway_by_order( $order );
+			// TODO: Correct way to parse amount to WC style floats.
+			$result = $gateway->process_refund( $order_id, floatval( $amount / 100 ), $reason );
+
+			return rest_ensure_response( $result );
```
KarlisJ: I'm unsure as to how to best handle the error message. Right now we display hardcoded (in the JS) generic message regardless of the case. `$gateway->process_refund` handle some cases where we expect to return message to end user. However, other cases involve API failure and shouldn't be displayed to end user.

dmvrtx:
> - Make payment and check it's details page
> - Choose "Refund in full"
> - The order and transaction should be refunded

If I follow these steps on the payment that is linked to an order, order remains in the "Completed" status. The way I read testing instructions order is expected to transition into "Refunded" status, isn't it?

dmvrtx:
When simulating a missing order link, the notice stays the same after refund goes through, but without an action button:

image

It would seem reasonable to me to remove the second sentence if transaction is already refunded.

dmvrtx (changes requested):
Thanks for an elaborate effort to handle this scenario! Looks sane code-wise, but there are two cases I mentioned earlier which I'd like to be addressed.

client/payment-details/summary/missing-order-notice/index.tsx
```diff
@@ -5,57 +5,45 @@
...
 				<CardNotice
 					actions={
-						<Button
-							variant="primary"
-							isSmall={ false }
-							onClick={ handleOnButtonClick }
-						>
-							{ __( 'Refund', 'woocommerce-payments' ) }
-						</Button>
+						! charge.refunded ? (
```
mgascam, line 36: Since charge.refunded is reactive, we can take advantage and remove the "refund the transaction as needed" but we may need two localized strings.

client/payment-details/summary/RefundModal/index.tsx
mgascam, line 1: If I am not wrong, we usually follow snake-case naming convention for the folders, I mean `client/payment-details/summary/refund-modal/index.tsx`

mgascam (commented):
Looking good to me, thanks for the refactor of the modal! I left a couple of minor comments. Will follow-up with additional testing and review.

KarlisJ:
@dmvrtx I was using the wrong WC method to do the refund. It's fixed now and the case you pointed out works as expected. 

Also addressed other comments.

dmvrtx (approved):
Looks good to me!

changelog/2023-11-13-09-52-07-471016
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: dev
+
+
```
kalessil: Can we rename the changelog file, please?

client/data/payment-intents/actions.ts
```diff
@@ -1,5 +1,14 @@
 /** @format */
 
+/**
+ * External dependencies
+ */
+import { apiFetch } from '@wordpress/data-controls';
+import { Charge } from 'wcpay/types/charges';
```
kalessil: `wcpay/*` should probably go into `Internal dependencies`.

client/payment-details/summary/index.tsx
```diff
@@ -55,6 +62,8 @@ import MissingOrderNotice from 'wcpay/payment-details/summary/missing-order-noti
 import DisputeAwaitingResponseDetails from '../dispute-details/dispute-awaiting-response-details';
 import DisputeResolutionFooter from '../dispute-details/dispute-resolution-footer';
 import ErrorBoundary from 'components/error-boundary';
+import { moreVertical } from '@wordpress/icons';
```
kalessil: Not sure, but this import probably belongs to external dependecies.

client/payment-details/summary/index.tsx
```diff
@@ -170,8 +179,12 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 	const renderStorePrice =
 		charge.currency && balance.currency !== charge.currency;
 
+	// // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
kalessil: A bit confused on this comments block, can we do a bit of cleaning here?

client/payment-details/summary/index.tsx
```diff
@@ -460,6 +474,41 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+														)
+													}
+												>
+													Refund in full
+												</MenuItem>
+												{ charge.order && (
+													<MenuItem
+														onClick={ () =>
+															( window.location =
+																// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
kalessil: What TS was complaining about here?

includes/admin/class-wc-rest-payments-refunds-controller.php
```diff
@@ -0,0 +1,76 @@
...
+				]
+			);
+
+			return rest_ensure_response( $result );
+		} else {
+			try {
+				$refund_request = Refund_Charge::create( $charge_id );
+				$refund_request->set_charge( $charge_id );
+				$refund_request->set_amount( $amount );
+				$refund_request->set_reason( $reason );
```
kalessil: Looking at server code, I don't see it processes the `reason` parameter, can you double-check please?

kalessil (approved):
Pre-approving: leaving nits and a couple questions.

client/payment-details/summary/index.tsx
KarlisJ: That `charge.order` can possibly be null but it's it's in practise impossible since we have condition for that.

includes/admin/class-wc-rest-payments-refunds-controller.php
KarlisJ: You're right. I had that change locally but I forgot to create a PR in the server repo. Here it is https://github.com/Automattic/woocommerce-payments-server/pull/4287

Also https://github.com/Automattic/woocommerce-payments/pull/7742/commits/148d32d0594c0a7669a1f780b1dfa45db8129fe2 adds the reason to refund request during the existing refund flow

client/payment-details/summary/index.tsx
KarlisJ: It's a dead code that will be removed from PR before merging. It's used to simulate the "order missing" flow. Without the comments the TS compilers errors when simulating missing order

includes/admin/class-wc-rest-payments-refunds-controller.php
kalessil: Thanks!

kalessil:
Testing in progress, I will add comments as findings surface.

First one: on FF, the drop-down menu alignment does not match the original designs:
Screenshot from 2023-12-08 14-23-44

And:
- The tooltip 'more' is not helpful, it should prompt you to click for more transactions-related actions.
- When clicking on partial refund, redirected to a non-existing page as e.g. `http://localhost:8082/wp-admin/undefined`

kalessil:
Next batch:
- I can submit a refund modal without picking a reason - I believe the submit button shall be disabled until the reason is selected
- After submission, the notification says, 'This transaction is not connected to order.'. I think we need a second sentence, complementary to the one asking to review/refund, but saying it's been refunded and not a subject for disputes.

kalessil:
Testing results for refund reasons:
- Duplicate order: Stripe responds with error `Invalid reason: must be one of duplicate, fraudulent, or requested_by_customer`
- Fraudulent: OK
- Requested by customer: OK
- Other: OK

kalessil (changes requested):
I completed the next review/testing round: some bugs surfaced, and we must address them.

EDIT: tests are missing. We can do snapshot-based for JS and regular for PHP.

KarlisJ:
@kalessil Thank you for testing.

> When clicking on partial refund, redirected to a non-existing page as e.g. http://localhost:8082/wp-admin/undefined

I'm unable to reproduce this. Would you try again and write specific steps? Could it be you did some tampering before seeing this issue?

> I can submit a refund modal without picking a reason - I believe the submit button shall be disabled until the reason is selected

The reason is optional, the same as in regular refunds through the order page. The option "other" might be a bit confusing since it's equivalent to sending no reason to Stripe, but I assume it's provided by designs as a means of "unselecting" reasons in case the user selected any of them by mistake.

kalessil:
> I'm unable to reproduce this. Would you try again and write specific steps? Could it be you did some tampering before seeing this issue?

Yes, I changed how to trigger missing order details - it can be a side effect. We'll re-test a bit later, and I'd like to record a demo when we merge the PR.

> The reason is optional, the same as in regular refunds through the order page. The option "other" might be a bit confusing since it's equivalent to sending no reason to Stripe, but I assume it's provided by designs as a means of "unselecting" reasons in case the user selected any of them by mistake.

Okay, sounds like I missed that conversation. Let's keep this way then.

includes/class-wc-payment-gateway-wcpay.php
```diff
@@ -1816,6 +1816,10 @@ static function ( $refund ) use ( $refund_amount ) {
 				if ( null !== $amount ) {
 					$refund_request->set_amount( WC_Payments_Utils::prepare_amount( $amount, $order->get_currency() ) );
 				}
+				// These are reasons supported by Stripe https://stripe.com/docs/api/refunds/create#create_refund-reason.
+				if ( in_array( $reason, [ 'duplicate', 'fraudulent', 'requested_by_customer' ], true ) ) {
```
kalessil, line 1821: Nice one!

client/payment-details/summary/index.tsx
```diff
@@ -469,6 +483,63 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+									position: 'bottom left',
+								} }
+							>
+								{ () => (
+									<MenuGroup>
+										<MenuItem
+											onClick={ () => {
+												setIsRefundModalOpen( true );
+												wcpayTracks.recordEvent(
+													'payments_transactions_details_refund_model_open',
```
kalessil: model, or modal?

client/payment-details/summary/index.tsx
```diff
@@ -496,14 +567,20 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+					charge={ charge }
+					formattedAmount={ formattedAmount }
+					onModalClose={ () => setIsRefundModalOpen( false ) }
+				/>
+			) }
+			{ ! _.isEmpty( charge ) && ! charge.order && ! isLoading && (
+				<MissingOrderNotice
+					charge={ charge }
+					isLoading={ isLoading }
+					onButtonClick={ () => setIsRefundModalOpen( true ) }
```
kalessil, line 593: I think we missed an event tracking here.

kalessil (commented):
Performing one more review/testing round, leaving first comments.

includes/admin/class-wc-rest-payments-refunds-controller.php
kalessil: We should be good as it is (except I think we miss handling cases when `$order` is empty), just ensure that we show a generic error message in UI - that'll be enough, as details likely be in order notes.

includes/admin/class-wc-rest-payments-refunds-controller.php
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+use WCPay\Core\Server\Request\Refund_Charge;
+use WCPay\Exceptions\API_Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * REST controller for the timeline, which includes all events related to an intention.
+ */
```
kalessil, line 15: Can we add something like `@internal Not intended for usage in integrations or outside of WooCommerce Payments.` to flag the controller's intended usage?

includes/admin/class-wc-rest-payments-refunds-controller.php
dmvrtx: Can we add a metadata flag here to indicate that refund was initiated from the transaction page?

client/payment-details/summary/refund-modal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+	) as PaymentChargeDetailsResponse;
+
+	const handleModalCancel = () => {
+		onModalClose();
+	};
+
+	const handleRefund = async () => {
+		wcpayTracks.recordEvent( 'payments_transactions_details_refund_full', {
+			payment_intent_id: charge.payment_intent,
+		} );
```
dmvrtx, line 51: Tracks data is limited to the merchants who have opted in for WooCommerce analytics. Given that we have tens of potential users per month, it could be that those who use the tool overlap very little with those who have analytics enabled. I suggest to keep tracks but to add a more reliable way of knowing that the refund was initiated through the transactions page, e.g. refund metadata.

kalessil: It looks like I missed some conversations, @dmvrtx - which analytics makes sense for this phase of the project? Are tracks on client tracks on the server or via meta-data? I'm asking if we can start with tracks and replace them later in a separate ticket with the desired output - currently, I'm not aware of specific expectations.

client/data/payment-intents/actions.ts
```diff
@@ -32,3 +41,57 @@ export function updateErrorForPaymentIntent(
 		error,
 	};
 }
+
+export function* refundCharge(
```
kalessil, line 45: Can we keep naming consistent here? E.g. `refundChargeByPaymentIntent`

client/payment-details/summary/index.tsx
```diff
@@ -469,6 +483,63 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
 				</div>
+				<div className="payment-details__refund-controls">
+					{ ! charge?.refunded && (
+						<Loadable
+							isLoading={ isLoading }
+							placeholder={ moreVertical }
+						>
+							<DropdownMenu
+								icon={ moreVertical }
+								label="Transaction actions"
```
kalessil: @KarlisJ @mgascam: I guess we have to pass in an already translated string. Can you confirm somehow?

client/payment-details/summary/index.tsx
```diff
@@ -469,6 +483,63 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+												wcpayTracks.recordEvent(
+													'payments_transactions_details_refund_model_open',
+													{
+														payment_intent_id:
+															charge.payment_intent,
+													}
+												);
+											} }
+										>
+											Refund in full
```
kalessil: Also: translations.

client/payment-details/summary/index.tsx
```diff
@@ -469,6 +483,63 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
...
+																	?.number,
+														}
+													);
+													window.location =
+														// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+														// @ts-ignore
+														charge.order.url;
+												} }
+											>
+												Partial refund
```
kalessil: Also: translations.

client/payment-details/types.ts
```diff
@@ -9,6 +9,7 @@ export interface PaymentChargeDetailsResponse {
 	data: PaymentIntent | Charge;
 	error: ApiError;
 	isLoading: boolean;
+	doRefund: ( charge: Charge, reason: string | null ) => void;
```
kalessil, line 12: @mgascam @KarlisJ , this looks a bit weird to me - we don't have `doCapture`, but adding `doRefund`. Any insights, on why we don't have the same approach?

kalessil (changes requested):
Okay, we are making good iterations, but I'm leaving new comments and want to outline topics we need to focus on and don't side-track:
- approach to data collection
- testing approach, as we discussed during the Team chat
- some inconsistency with out approaches compared to Auth&Capture

client/payment-details/summary/refund-modal/index.tsx
dmvrtx: We need metadata, because we might have only few data points and I'd rather have them all, than miss something due to the way Tracks work.

client/payment-details/summary/index.tsx
mgascam: Yes I think it needs to be translated. This `label` prop is meant for accessibility purposes according to the component docs. The usual `_('Translation actions', ''woocommerce-payments')` should work.

mgascam: Yes I think it needs to be translated. This label prop is meant for accessibility purposes according to the component docs. The usual `_('Translation actions', ''woocommerce-payments')` should work.

client/payment-details/summary/index.tsx
```diff
@@ -469,6 +483,63 @@ const PaymentDetailsSummary: React.FC< PaymentDetailsSummaryProps > = ( {
 						</div>
 					</div>
 				</div>
+				<div className="payment-details__refund-controls">
+					{ ! charge?.refunded && (
+						<Loadable
+							isLoading={ isLoading }
+							placeholder={ moreVertical }
+						>
+							<DropdownMenu
```
mgascam, line 488: I believe we need to control the visibility of the dropdown menu. In my local it stays open after clicking in the 'Refund in full' option. Ideally we should close it. I'd say we can leverage the `isOpen` prop for this purpose.

mgascam (commented):
Thanks Karlis it's shaping up nicely. I left a couple of comments related to some findings after testing. 

I also tried to perform a full refund on an uncaptured payment and got an error. Looking at the dev tools, I can see more details in the response:

`{
    "code": "error",
    "message": "This payment is not captured yet. To cancel this order, please go to 'Order Actions' > 'Cancel authorization'. To proceed with a refund, please go to 'Order Actions' > 'Capture charge' to charge the payment card, and then trigger a refund via the 'Refund' button.",
    "data": null
}`

I'd say we should surface this details to the merchant (somehow) or limit this functionality for uncaptured payments? At the same time, this PR is already growing quite big so we may also address in a separate issue?

I intend to continue testing and still got some pings to look into.

client/payment-details/types.ts
mgascam: @kalessil I'll try my best answer based on my understanding but I may be off. The reason why `doCapture` is not there is because the types for the `useAuthorization` hook are defined inline here https://github.com/Automattic/woocommerce-payments/blob/cd45dcb7b620a28d187d9a340c02e59cb3df2d9e/client/data/authorizations/hooks.ts#L79.  IMO the `doCapture` does not belong in the `PaymentChargeDetailsResponse` type because we are talking about two different hooks that are interested in two different parts of the state: PaymentIntents and Charges on one hand and Authorizations on the other.

Having said that, there may be an opportunity to extract the inline types definition of the `useAuthorization` hook into its own interface, same as the `PaymentChargeDetailsResponse`. But truth is it is only used in one place at the moment, so I don't see the point. 

Furthermore, the usage of `PaymentChargeDetailsResponse` in https://github.com/Automattic/woocommerce-payments/blob/e00a4af880a9fa9d1d764e2b9ffe93b8ada10e26/client/payment-details/charge-details/index.tsx#L31 and https://github.com/Automattic/woocommerce-payments/blob/34eb13c4b140cebd233314eb06d4daa750ae86fb/client/payment-details/summary/refund-modal/index.tsx#L42 seem redundant to me, because the TS compiler should already able to infer the type based on the return type of the `usePaymentIntentWithChargeFallback` hook.  In my understanding of TS, the 'as' keyword should be used sparingly, since you are essentially telling the compiler to trust the human. 

Hope to clarify things, please let me know if you have any question.

client/payment-details/summary/refund-modal/index.tsx
KarlisJ: @dmvrtx 
> refund metadata.

Are you referring to Stripe's metadata here?

client/payment-details/types.ts
kalessil: Yes, that gives more clarity. Thanks!

Since there's not much we can do with generalization here, marking the threa as resolved.

client/payment-details/summary/refund-modal/index.tsx
dmvrtx: > Are you referring to Stripe's metadata here?

Yes. I'll check if we need to do anything on the server side to make it readily-available for querying, but I think it might be collected already.

KarlisJ: @dmvrtx It's a bit tricky. 

We now have three execution paths for how the refund can happen. 
1. The "old" from order page
2. (2.1) From the transaction page with order id
3. (2.2.) From the transactions page without order id

We can add the metadata to the request that we make to the server, but for the first two options, we call `wc_create_refund` which internally calls the request to the server, and it doesn't have any arguments that we can use to pass that informations in. We'd have to make changes in the WC core to make this possible ...

dmvrtx: Since we're mostly interested in the third (or 2.2) use-case I'm okay with having metadata just for it. We can look into tracking down full refunds for order transactions later on if the need arises.

KarlisJ: @dmvrtx FYI, the meta would be added to the `refund` object in Stripe, not `payment intent` since the proxied endpoint is for `/refunds`

mgascam (approved):
Pre-approving since works as expected after additional manual tests. I cannot think of any other feedback code wise, thanks for addressing all my comments. I reckon there are some open ends related to tracking, so I'll defer to other folks for additional approvals.

client/payment-details/summary/refund-modal/index.tsx
dmvrtx: Thanks. Since we send `refund_source` as part of the request to the server endpoint, it is possible to catch it there and it won't even require looking into the refund metadata.

includes/core/server/request/class-refund-charge.php
```diff
@@ -59,6 +60,31 @@ public function set_amount( int $amount ) {
...
+	 *
+	 * @param string $source The reason for the refund.
+	 * @throws Invalid_Request_Parameter_Exception
+	 */
+	public function set_source( string $source ) {
+		$this->set_param(
+			'metadata',
+			[
+				'refund_source' => $source,
+			]
```
dmvrtx: Would it make sense to use `array_merge` here for better extensibility and avoiding overwrite of existing metadata in the future?

kalessil: Done in 1a011d3cfa8b039c83292afce97f3d95c952a0af

kalessil (approved):
LGTM and tests well (I pushed a couple of commits to address remaining comments)!

PR #7753, opened 2023-11-20
[WIP] Improvements to E2E setup script

achyuthajoy:
Fixes #

#### Testing instructions

*

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7753` or branch name `enhancement/improve-e2e-setup-script` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=enhancement/improve-e2e-setup-script) :rocket:

---

Build info: 

- Latest commit: 6ce5b586e1cf0dc71208734578bfbff14df58f14
- Build time: 2023-11-21 03:07:18 UTC

tests/e2e/env/setup.sh
```diff
@@ -1,366 +1,520 @@
...
-if ! command -v jq &> /dev/null
-then
-    echo "The script requires jq library to be installed. For more info visit https://stedolan.github.io/jq/download/."
-    exit 1
-fi
-
-. ./tests/e2e/env/shared.sh
+# Import helper functions.
+# shellcheck disable=SC1091
+source "$(pwd)/tests/e2e/env/helpers.sh"
```
nb, line 5: Does this assume that the script is run from the root folder of the plugin?

tests/e2e/env/helpers.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
nb, line 1: What's the difference between `shared` and `helpers`? There seem to be some functions in `shared`, too.

tests/e2e/env/setup.sh
```diff
@@ -1,366 +1,520 @@
 #!/usr/bin/env bash
 
-set -e
```
nb, line 3: It looks like this `set -e` was lost – was this on purpose?

tests/e2e/env/setup.sh
```diff
@@ -1,366 +1,520 @@
...
-	if [[ -n $CI ]]; then
-		step "Disable Xdebug on server container"
-		docker exec "$SERVER_CONTAINER" \
-		sh -c 'echo "#zend_extension=xdebug" > /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini && echo "Xdebug disabled."'
+	# Loop through server variables and check if they are set.
+	check_vars "server_logs" "${server_vars[@]}"
+	check_vars "client_logs" "${client_vars[@]}"
+	check_vars "other_logs" "${other_vars[@]}"
+
+	# Log to terminal if any errors are present and exit.
```
nb, line 104: How does storing all log messages in an array help? Is it only to be able to group errors by category and to be able to have the red message in the beginning, instead of in the end? If yes, is the complexity worth it?

nb (commented):
It was fun pairing yesterday. I know this is still WIP, but left some comments, since I had context about the change.

tests/e2e/env/setup.sh
achyuthajoy: @nb Yes. The setup script is run via npm and it's called from the project root

tests/e2e/env/helpers.sh
achyuthajoy: I'm yet to consolidate other files. I'll be including all the variables & functions inside helpers or the main script

tests/e2e/env/setup.sh
achyuthajoy: Yes. I missed that. But the script is working as expected even without it. 

But I will add to make sure we exit immediately if a step fails.

tests/e2e/env/setup.sh
achyuthajoy: So bash supports associative arrays, which can be used to store the logs, but that doesn't work with Bash versions < 4. 

I used the current approach as an alternative to keep the logs contained to a single place for each category and to keep the code clean. 

Let me know if you have other suggestions?

achyuthajoy: And yes, I wanted to keep the error text order.

nb: Just printing as an error comes up sounds as good and also makes the code so much simpler.

tests/e2e/env/setup.sh
nb: If I run it locally, do I also need to run it from this folder? Is anything else stopping us to run it from anywhere, except the `pwd`?

achyuthajoy: > If I run it locally, do I also need to run it from this folder?

Running locally, you can clone the repo, do the basic setup & then run `npm run test:e2e-setup`, which will run the script from the project root. This is the expected approach.

>Is anything else stopping us to run it from anywhere, except the pwd?

There are variables defined under shared.sh, for which the path is defined from project root. So that would be a blocker.

tests/e2e/env/setup.sh
achyuthajoy: Sure. I'll make modifications :) Thanks

tests/e2e/env/setup.sh
nb: But it's because we're also using `pwd` there. The whole thing assumes we run it from a specific directory and breaks in a weird way if not run from the root – maybe we should either stop execution if we're not in the expected directory, or make it work when it's run from anywhere.

tests/e2e/env/setup.sh
nb: Well, as expected in the happy path :)

tests/e2e/env/setup.sh
nb: Thanks! And wanted to follow up on the changes, I don't see more commits.

achyuthajoy: @nb I had to shift focus to WooPayments MC architecture project which is priority for our team at the moment.

So I'll addressing the reviews sometime next week.

PR #7285, opened 2023-09-25
Fix incorrect payment method title

mdmoore:
Fixes #7272
Fixes #7273 

#### Changes proposed in this Pull Request

This prevents the payment method title from being updated for non-WooPayments gateways and limits those updates to the Legacy UPE gateway.

#### Testing instructions

- Enable Cash On Delivery and WooPayments
- As a customer, add any product to the cart
- Select Cash On Delivery as payment method
- Checkout
- On the Order Confirmation page, ensure the correct payment method title is set

Before:

After:

-------------------

- [x] Run `npm run changelog` to add a changelog file, choose `patch` to leave it empty if the change is not significant. You can add multiple changelog files in one PR by running this command a few times. 
- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

- [ ] Link to testing instructions from [release testing doc](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions) following [these instructions](https://github.com/Automattic/woocommerce-payments/wiki/How-to-write-good-manual-testing-scenarios) : _Add link here / 'QA Testing Not Applicable'_
- [ ] Add or update [critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Critical-flows) and [testing instructions for critical flows](https://github.com/Automattic/woocommerce-payments/wiki/Testing-instructions-for-critical-flows), if applicable.
- [ ] Add what's changed (description, screenshot, demo videos etc.) to the release announcement post, if applicable.

PR #7770, opened 2023-11-22
Streamline UPE & non-UPE branching across the codebase & improve test coverage

### LINKED ISSUE #5028: Streamline UPE/non-UPE code branching and eliminate UPE feature flag checks

timur27:
Closes https://github.com/Automattic/woocommerce-payments/issues/5028

This PR is the first part of the legacy UPE & card gateways cleanup. It eliminates all the branching in code and lets the plugin run deferred intent creation UPE. It also improves the code coverage, striving to full unit test coverage of methods introduced in the UPE. This PR: 
* removes all the `if ... else` statements that checked which UPE/non-UPE type is enabled
* cleans up feature flags for dUPE, split UPE & legacy UPE
* covers more deferred intent creation UPE functionality with unit tests
* removes legacy UPE & split UPE-related logic, which was only executed when the corresponding flag was enabled, which is not the case for any merchants since `6.7.0`

> 💡 UPE feature flag states provided to the front-end were converted to some static values ([#1](https://github.com/Automattic/woocommerce-payments/pull/7770/files#diff-27b0c813f2fb3734bfda326e404028952d157c765615c4e6f6cc4d4efb183b8dR190-R191), [#2](https://github.com/Automattic/woocommerce-payments/pull/7770/files#diff-a441a64d0adecb8d97d15e2ad40f34011f5cb32f50257b6a3d9061cdb5fb83f2R35)). This helps us clean things up on the back-end but don't start the cleanup on the front-end and approach it in a separate PR.

#### Changes provided and testing instructions for them

- [trim down logic for stripe platform account usage check](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/7f9418df847515ffd4c965891f81392b44391da3) 
  - with the deferred UPE feature flag enabled, turn on WooPay and confirm that the default save payment method checkbox disappeared and the WooPay-related one now shows up
  - disable deferred UPE and confirm the previous step once again

- [skip checking dUPE for mandate data attachment with physical products](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/969a71332c5369b53c983ceec49511eae2667ce5) 
   - 🧪 https://github.com/Automattic/woocommerce-payments/pull/7770/commits/9681346e3af9fae2b7e274fc7cba4c84c81427b5
   - for physical product purchases, ensure that regardless of the flag state, `mandate_data` is attached to `POST /v1/payment_intents` request when Stripe Link is enabled (request to be checked in Stripe dashboard)

- [skip checking dUPE for mandate data attachment with zero-total products ](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/05a4336334b64ca62b6e14e9d44f1903d275bb87)
    -  🧪 a4ed36a3c3a7a97c2d5ded8c30f8733d73c1fc67
    - same as for physical products above, but purchasing e.g., a free trial subscription and checking `mandate_data` is part of the body in wp-admin WooCommerce logs

- [eliminate deferred UPE check during object retrieval for Stripe platform usage verification](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/19532b67b26b19d4e80cf584ed2385bae0150369) 
   - 🧪 https://github.com/Automattic/woocommerce-payments/commit/dacbd7329ce37d4f09996835b6d373d6667b6865 & [existing](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/woopay/services/test-checkout-service.php)
   - with disabled WooPay, confirm that card & non-card payments end are processed successfully
   - enable WooPay and verify card & any non-card payment method as well

- [streamline payment method retrieval from the request](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/a307647228c1381064a7957464789ac4d903bb2b)
   - 🧪 dacbd7329ce37d4f09996835b6d373d6667b6865
   - enable Stripe Link, set currency to USD, and when paying with a card, confirm that the create intent request in Stripe contains both payment method types
   - change currency to EUR, pay with card keeping Stripe Link enabled, and confirm Stripe request contains only `card` for the `payment_method_types` array key

- [align SEPA gateway id generation while adding a token](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/e242b048c725c3ae76728d13925895111cf17717)
   - 🧪 [exists](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/test-class-wc-payments-token-service.php#L106-L122)
   - SEPA support has been temporarily [disabled](https://github.com/Automattic/woocommerce-payments/pull/7107/files#diff-8c4486ba4e071c67d482cfc2226228a3e4af5f78a126d46b5db55a41a9870108R28) but you can remove the first part of [the check here](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payment-gateway-wcpay.php#L1450), enable SEPA for your account, place order while paying with SEPA and confirm the order is processed successfully

- [always register deferred UPE scripts for zero order total](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/fc4b338b7abdb23536c35407d83998444acc61f9)
   - test zero order total with free shipping & change the country to non-free shipping country without leaving the checkout page and confirm payment methods are loaded
      - ensure you have two shipping zones in WooCommerce (e.g. one free shipping for US and another one flat rate $10 for e.g. United Arab Emirates)
      - add one free physical product and open the shortcode checkout so that US is chosen by default as shipping address country
      - confirm no payment methods section is loaded
      - change the shipping country to United Arab Emirates and confirm that payment methods are loading properly

- [use deferred UPE checks for valid payment method type for gateway](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/a84f904cada8105a2a77b5b7fedf462b8bed3ea8)
   - 🧪 https://github.com/Automattic/woocommerce-payments/pull/7770/commits/a84f904cada8105a2a77b5b7fedf462b8bed3ea8

- [delete `create_payment_intent_ajax` methods and their attachment to the hooks](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/b9e2971c9932473065aeb2b457405ddb6637bb30)
   - perform a sanity check by placing an order through shortcode checkout
   - no further testing is needed because this commit removes the code that wasn't ever used by the dUPE and instead [was used by legacy UPE & split UPE](https://github.com/Automattic/woocommerce-payments/blob/develop/client/checkout/api/index.js#L416)

- [remove branching to load front end scripts](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/dc48fbedd5f88d5152feb86d01ef1e4b8ca0025d)
   - perform a sanity check by placing an order through shortcode checkout (if you have already done this e.g. as part of previous test scneario, please skip this step)

- [remove branching to load front end scripts](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/dc48fbedd5f88d5152feb86d01ef1e4b8ca0025d)
   - perform a sanity check by placing an order

- [remove branching to make back-end provide only dUPE related fields to the front-end](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/2da2193f8a1266a3d865bd49b351624b570870fd) 
   - perform a sanity check by placing an order

- [always provide additional dUPE setting values to front-end](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/28f0367feae4951e688e97031b19d8c2d4f12c1b)
   - place an order & confirm that testing instructions are displayed for the card (this is one of the fields that are now always provided so checking if it's displayed means there is no regression)

- [streamline upe config object name for front-end](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/52d5671c07ca5d12b4a341e72fc852f144aaec51)
   - open the shortcode checkout page and confirm that payment elements are loaded correctly and you are not seeing the `Uncaught ReferenceError: wc is not defined` error

- [streamline checkout gateway registration for blocks](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/0172328d0b2344f93d3d883694c267aa40859b83)
   - open blocks checkout page and confirm that payment elements are loaded correctly, and you are not seeing any errors (like e.g. `Unexpected error in: woocommerce/checkout-payment-block` or `There are no payment methods available. This may be an error on our side`

- [delete branching in the plugin initialization phase](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/7ae214462c15debd72d628a51402179be2df0abe)
   - with only card enabled, pay on shortcode & blocks checkout 
   - enable more payment methods, confirm they are all displayed as expected and perform a purchase with any non-card gateway

- [remove gateway title filtering which was not used by deferred intent UPE](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/2eefb0adf138e4f37a71f3fcceda741b08c9ed11)
   - install WP Mail Logging plugin & enable card + e.g. Bancontact
   - open the shortcode checkout and confirm that both gateways have the right title (`Credit card / debit card` & `Bancontact`)
   - pay with each of them, navigate to WP Mail Logging, and confirm the gateway title didn't change 
   - repeat the same with blocks checkout (optional & checked by PR author as well)

- [remove legacy UPE flag and its checks across the code](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/5abe673fcc088fc8bce9e75b2da84f52786685aa)
   - nothing to check, one thing to note is that the `set_payment_method_title_for_email` method was removed [as planned](https://github.com/Automattic/woocommerce-payments/pull/7285#discussion_r1342913516)

- small changes that don't require testing
   - [always pass isUPEDeferredEnabled as true to front-end before it is removed there](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/6f97c148cf79292053001dfad966f45cc6cd2b84)
   - [remove irrelevant unit test](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/d96fdd0ba855acf244b4fedaa289376a68aa695b)
   - [remove the non-used deferred UPE feature flag method](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/c837d9e6d941c4f1a7ac147d95674e4eab6f94e1)
   - [remove split UPE flag and its checks accross the code](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/bccfdab7ff1e86f944f03c8c74a3c2d2b97bc351)
   - [remove deferred UPE feature flag constant](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/2a9a8eb8cad5665f9ed7a1ba8978231a5f27b85b)
   - [remove redundant hook](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/9c26d7abbaa5d4e5c56cdff1f658733718733f0c)

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7770` or branch name `cleanup/legacy-upe-and-card` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=cleanup/legacy-upe-and-card) :rocket:

---

Build info: 

- Latest commit: 3d81123c793aae60fdb6a13dc9965f65b422b9c6
- Build time: 2023-11-27 09:58:48 UTC

includes/class-wc-payments-upe-checkout.php
```diff
@@ -137,11 +133,7 @@ public function register_scripts() {
 			$script_dependencies[] = 'woocommerce-tokenization-form';
 		}
 
-		if ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
-			$script = 'dist/upe_with_deferred_intent_creation_checkout';
-		} else {
-			$script = 'dist/upe_checkout';
```
frosso, line 143: Note for our future selves: `dist/upe_checkout` seems to no longer be used. It looks like its reference in `webpack/shared.js` can be removed, as well as `client/checkout/classic/upe.js` and every other frontend dependency imported in `client/checkout/classic/upe.js` that is not shared

includes/class-wc-payments-checkout.php
```diff
@@ -145,14 +145,8 @@ public function register_scripts_for_zero_order_total() {
 			! has_block( 'woocommerce/checkout' )
 		) {
 			WC_Payments::get_gateway()->tokenization_script();
-			$script_handle = 'WCPAY_CHECKOUT';
```
frosso, line 148: Since `WCPAY_CHECKOUT` is not used here anymore, do you think its references can also be removed from anywhere else in the application? 👀

includes/class-wc-payments.php
```diff
@@ -1343,16 +1322,8 @@ public static function get_session_service() {
 	 */
 	public static function register_checkout_gateway( $payment_method_registry ) {
 		require_once __DIR__ . '/class-wc-payments-blocks-payment-method.php';
-		if ( WC_Payments_Features::is_upe_deferred_intent_enabled() ) {
-			require_once __DIR__ . '/class-wc-payments-upe-split-blocks-payment-method.php';
-			$payment_method_registry->register( new WC_Payments_UPE_Split_Blocks_Payment_Method() );
-		} elseif ( WC_Payments_Features::is_upe_legacy_enabled() ) {
-			require_once __DIR__ . '/class-wc-payments-upe-blocks-payment-method.php';
```
frosso, line 1350: It looks like `class-wc-payments-upe-blocks-payment-method.php` is not included anywhere else - can it be deleted as part of this PR?

includes/payment-methods/class-upe-payment-gateway.php
```diff
@@ -1132,58 +1073,6 @@ public function clear_upe_appearance_transient() {
...
-	}
-
-	/**
-	 * Sets the payment method title on the order for emails.
-	 *
-	 * @param WC_Order $order   WC Order object.
-	 *
-	 * @return void
-	 */
-	public function set_payment_method_title_for_email( $order ) {
```
frosso, line 1166: Is it accurate to also remove this filter for dUPE? 👀

frosso (commented):
> 💡 UPE feature flag states provided to the front-end were converted to some static values (https://github.com/Automattic/woocommerce-payments/pull/7770/files#diff-27b0c813f2fb3734bfda326e404028952d157c765615c4e6f6cc4d4efb183b8dR190-R191, https://github.com/Automattic/woocommerce-payments/pull/7770/files#diff-a441a64d0adecb8d97d15e2ad40f34011f5cb32f50257b6a3d9061cdb5fb83f2R35). This helps us clean things up on the back-end but don't start the cleanup on the front-end and approach it in a separate PR.

I think this is a good approach 👍 

I went through some static analysis, but haven't performed a full test yet. So I'm leaving these comments in the meantime. I'm wrapping up a few other things before the weekend, and I'll continue the review by actually testing the changes on Monday. Feel free to reply, in the meantime :)

includes/payment-methods/class-upe-payment-gateway.php
timur27: I understand the concern. I think so, here is some reference from the description

> nothing to check, one thing to note is that the set_payment_method_title_for_email method was removed [as planned](https://github.com/Automattic/woocommerce-payments/pull/7285#discussion_r1342913516)


And I think that one of the test scenarios from above will also cover checking this

>[remove gateway title filtering which was not used by deferred intent UPE](https://github.com/Automattic/woocommerce-payments/pull/7770/commits/2eefb0adf138e4f37a71f3fcceda741b08c9ed11)
install WP Mail Logging plugin & enable card + e.g. Bancontact
open the shortcode checkout and confirm that both gateways have the right title (Credit card / debit card & Bancontact)
pay with each of them, navigate to WP Mail Logging, and confirm the gateway title didn't change
repeat the same with blocks checkout (optional & checked by PR author as well)


Happy to discuss this further!

includes/class-wc-payments.php
timur27: Definitely, thanks! https://github.com/Automattic/woocommerce-payments/pull/7770/commits/564d78cb4379086bb5b111d669fed21d2e9632c7

includes/class-wc-payments-checkout.php
timur27: That's a good suggestion. The main reason I kept this constant for now is that removing it will probably lead to deleting the `WC_Payments_Checkout` class because it is used as the one & only while registering scripts or providing payment fields to JS. 

`WCPAY_CHECKOUT` is also used in Subscriptions for some reason, and I'd like to check it more deeply on why none of the UPE constants is used there before removing it.

I'd probably proceed with removing it within one of the next PRs, if all the above makes sense.

timur27: [Here's where it gets registered](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-checkout.php#L130) and I'd like to start the cleanup of the `WCPAY_CHECKOUT` from there. Probably as part of the issue to streamline `class-wc-payments-checkout.php` & `class-wc-payments-upe-checkout.php`. Let me know how it sounds for you and I'm happy to align further 👍

timur27:
@frosso I'm seeing a slight performance improvement on the checkout with this branch. This is good, but at the same time I want to know what is causing this, which I'll do on Monday. This is not a blocker at all, I'm just letting you know in case you would also want to consider the performance during the testing.

includes/class-wc-payments-upe-checkout.php
timur27: Mentioned in the last bullet point in #7780 and will be handled there. Thank you!

timur27:
@frosso regarding performance, I think I experienced some false positives on Friday and there's no real difference between `develop` and this branch as I see now. I also believed in the potential perf improvement, because we removed legacy UPE/dUPE FF checks that retrieved values through `get_option` and those were executed ~20-30 times per request. But now I think `get_option` also caches the result and number of the queries shouldn't have impacted performance on the current `develop`.

Sorry for the early heads-up here and the unnecessary noise.

includes/payment-methods/class-upe-payment-gateway.php
frosso: Thanks for confirming and digging up the conversation. The reason why I was concerned about the removal was because this logic seemed to have been applied to all gateway flavors (i.e.: I didn't see a feature-flag check for when this logic was applied).
The only check (below on L1169) was the early return when UPE Legacy was _not_ enabled (`WC_Payments_Features::is_upe_legacy_enabled()`).

frosso (approved):
I checked the conditions in the description + a few more from the critical flows.
I logged https://github.com/Automattic/woocommerce-payments/issues/7792, which is not related to this branch (I confirmed it happening in `develop` as well).

Thanks for your diligence with the testing and for answering my questions.

Good job! Looking forward to less branching logic!

timur27:
[Added new instructions](https://github.com/Automattic/woocommerce-payments/wiki/Release-testing-instructions-for-WC-Payments-6.9.0#tests-from-different-areas-to-be-done-separately-from-the-critical-flows) to the release testing docs. The rest should be covered by the critical flows testing.

PR #7775, opened 2023-11-24
Allow requests with IDs to be extended

### LINKED ISSUE #7508: Request Class through this exception `this request requires an item ID`
Implement a new class named My_Namespace\Prefix_Update_Intention to extend the Update_Intention class, plan to add some additional metadata, and override existing metadata.  Followed this process

```php
namespace My_Namespace;

use WCPay\Core\Server\Request\Update_Intention;

class Prefix_Update_Intention extends Update_Intention {
  // some custom function
}
```

```php
add_filter( 'wcpay_update_intention_request', [ __CLASS__, 'add_dynamic_statement_descriptor_for_update_intention' ], 20, 2 );
...
public static function add_dynamic_statement_descriptor_for_update_intention( Request Request $base_request, WC_Order $order ) {
        $request = Prefix_Create_And_Confirm_Intention::extend( $base_request );
        // some customization
        return $request;
}
```

However, when attempting to complete the checkout, an error emerged: `This request requires an item ID.`

As I delved into the codebase to identify the source of this error, I discovered that when we utilize the extend method to generate a new instance based on `$base_request`, it neglects to transfer the intent ID during the initialization of the new instance.

`$obj = new $current_class( $base_request->api_client, $base_request->http_interface );`

https://github.com/Automattic/woocommerce-payments/blob/287990726914e93078e5af4207498136206d59a4/includes/core/server/class-request.php#L463-L464

The `set_request_route_id_parameter` method in the request class is searching for the intent_id, but it's unable to locate it, which eventually leads to an exception being thrown.

https://github.com/Automattic/woocommerce-payments/blob/287990726914e93078e5af4207498136206d59a4/includes/core/server/class-request.php#L373-L381

For testing purposes, we added the `intent_id` during the class initialization, and the checkout worked perfectly as expected.

```
$obj = new $current_class( $base_request->api_client, $base_request->http_interface, $base_request->id );
```

RadoslavGeorgiev:
Fixes #7508

Allow requests with IDs to be extended without triggering an exception. Without this PR, this is how requests get created during extension:

https://github.com/Automattic/woocommerce-payments/blob/287990726914e93078e5af4207498136206d59a4/includes/core/server/class-request.php#L463-L464

As you can notice, the third constructor parameter (ID) was being skipped. This causes requests that require IDs to immediately throw an exception, because if a `set_id()` method exists within the class, the parameter is automatically verified.

This PR simply adds the `$id` parameter.

#### Testing instructions

The description in https://github.com/Automattic/woocommerce-payments/issues/7508 is pretty good, and recommended.

The easiest way to understand the issue would be to to undo the fix from `class-request.php` and run the test: You will immediately see the problematic exception.

-------------------

### DIFF
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -460,7 +460,7 @@ abstract class Request {
 				'wcpay_core_extend_class_incorrectly'
 			);
 		}
-		$obj = new $current_class( $base_request->api_client, $base_request->http_interface );
+		$obj = new $current_class( $base_request->api_client, $base_request->http_interface, $base_request->id );
 		$obj->set_params( array_merge( static::DEFAULT_PARAMS, $base_request->params ) );
 
 		// Carry over the base class and protected mode into the child request.
--- a/includes/core/server/class-request.php
+++ b/includes/core/server/class-request.php
@@ -460,7 +460,7 @@ abstract class Request {
 				'wcpay_core_extend_class_incorrectly'
 			);
 		}
-		$obj = new $current_class( $base_request->api_client, $base_request->http_interface, $base_request->id );
+		$obj = new $current_class( $base_request->api_client, $base_request->http_interface, $base_request->id ?? null );
 		$obj->set_params( array_merge( static::DEFAULT_PARAMS, $base_request->params ) );
 
 		// Carry over the base class and protected mode into the child request.
--- a/changelog/fix-7508-request-extension
+++ b/changelog/fix-7508-request-extension
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fix
+
+Allow requests with item IDs to be extended without exceptions.
--- a/tests/unit/core/server/request/test-class-core-request.php
+++ b/tests/unit/core/server/request/test-class-core-request.php
@@ -8,6 +8,7 @@
 use WCPay\Core\Server\Request;
 use WCPay\Core\Server\Request\Paginated;
 use WCPay\Core\Server\Request\List_Transactions;
+use WCPay\Core\Server\Request\Update_Intention;
 
 // phpcs:disable
 class My_Request extends Request {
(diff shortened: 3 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7775` or branch name `fix/7508-request-extension` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7508-request-extension) :rocket:

---

Build info: 

- Latest commit: 4da980d40748d6656ecea999d67e69ea3accbd2a
- Build time: 2023-11-24 09:42:14 UTC

zmaglica (approved):

includes/core/server/class-request.php
```diff
@@ -460,7 +460,7 @@ final public static function extend( Request $base_request ) {
 				'wcpay_core_extend_class_incorrectly'
 			);
 		}
-		$obj = new $current_class( $base_request->api_client, $base_request->http_interface );
+		$obj = new $current_class( $base_request->api_client, $base_request->http_interface, $base_request->id );
```
zmaglica: Nitpick: I would do something like `$base_request->id ?? null` to avoid possible undefined id property since we are checking only if id is not null inside the function that sets id that is passed via constructor.

zmaglica:
Looks good to me. Left one small comment.

includes/core/server/class-request.php
RadoslavGeorgiev: Done in https://github.com/Automattic/woocommerce-payments/pull/7775/commits/4da980d40748d6656ecea999d67e69ea3accbd2a.

PR #7798, opened 2023-11-28
Use supported countries instead of business types for PO

### LINKED ISSUE #6841: Update PO Form to use supported countries on client, with guardrails and fallbacks
### Description

Currently, our logic fetches the scraped data from Stripe from the server for business types, and only shows countries which are returned. This relies on us not forgetting to add new countries when they are activated for WCPay. We should handle this by instead using the supported countries list stored on the client. The data from the server will still be used to allow business type selection per country, with a fallback that if data isn't returned from the server for a particular country, we can just not prompt the user for business information (and possibly alert ourselves to this so we can add the country!)

ismaeldcom:
Fixes #6841 

- Update PO business details step and utils to use countries from [`supported_countries`](https://github.com/Automattic/woocommerce-payments/blob/10d1cb52ed2162800d8a06c5a30632c1cba67538/includes/class-wc-payments-utils.php#L224) instead of [`get_business_types`](https://github.com/Automattic/woocommerce-payments/blob/10d1cb52ed2162800d8a06c5a30632c1cba67538/includes/class-wc-payments-onboarding-service.php#L105), to prevent missing a country while onboarding due to it not being added to the server.
- Ensure that's possible to onboard without providing a business type. The field will be hidden when you choose a country without type information.

Regarding this part of the issue:
>we can just not prompt the user for business information (and possibly alert ourselves to this so we can add the country!)

I've been thinking about it, and trying a couple of approaches, but I don't see a good way to properly warn us from the client.  A track event could be an option, but I don't think it's the way to do it. In the end, it will not alert us directly. It will be probably better to ensure we only have a single source of truth for supported countries in the server, rather than trying to cover these gaps with alerts or fallbacks, WDYT? 🤔 

#### Testing instructions
- If you have an account onboarded, be sure to enable `Force the WCPay plugin to act as disconnected from the WCPay Server` in WCPay Dev, and PO onboarding.
- Edit [`supported_countries`](https://github.com/Automattic/woocommerce-payments/blob/10d1cb52ed2162800d8a06c5a30632c1cba67538/includes/class-wc-payments-utils.php#L224) to add a new country, for example: `'AD' => __( 'Andorra', 'woocommerce-payments' ),`
- Proceed with the PO onboarding form until the business details step.
- Pick `United States` in the country field.
- The business type field should be visible.
- Pick the new country in the country field.
- The business type field shouldn't be visible.
- It should be possible to proceed with the onboarding.

-------------------

**Post merge**

### DIFF
--- a/changelog/update-6841-supported-countries-fallback
+++ b/changelog/update-6841-supported-countries-fallback
@@ -1,0 +1,5 @@
+Significance: patch
+Type: update
+Comment: Update PO onboarding country source to use available countries.
+
+
--- a/client/onboarding/steps/business-details.tsx
+++ b/client/onboarding/steps/business-details.tsx
@@ -14,14 +14,19 @@ import {
 	OnboardingSelectField,
 	OnboardingTextField,
 } from '../form';
-import { getBusinessTypes, getMccsFlatList } from 'onboarding/utils';
+import {
+	getAvailableCountries,
+	getBusinessTypes,
+	getMccsFlatList,
+} from 'onboarding/utils';
 import { BusinessType } from 'onboarding/types';
 
 const BusinessDetails: React.FC = () => {
 	const { data, setData } = useOnboardingContext();
-	const countries = getBusinessTypes();
+	const countries = getAvailableCountries();
+	const businessTypes = getBusinessTypes();
 
-	const selectedCountry = countries.find(
+	const selectedCountry = businessTypes.find(
 		( country ) => country.key === data.country
 	);
 	const selectedBusinessType = selectedCountry?.types.find(
--- a/client/onboarding/steps/test/business-details.tsx
+++ b/client/onboarding/steps/test/business-details.tsx
@@ -12,14 +12,39 @@ import { mocked } from 'ts-jest/utils';
 import BusinessDetails from '../business-details';
 import { OnboardingContextProvider } from '../../context';
 import strings from '../../strings';
-import { getBusinessTypes, getMccsFlatList } from 'onboarding/utils';
+import {
+	getAvailableCountries,
+	getBusinessTypes,
+	getMccsFlatList,
+} from 'onboarding/utils';
 
 jest.mock( 'onboarding/utils', () => ( {
+	getAvailableCountries: jest.fn(),
 	getBusinessTypes: jest.fn(),
 	getMccsFlatList: jest.fn(),
 } ) );
 
 const countries = [
+	{
+		key: 'ES',
+		name: 'Spain',
+		types: [],
+	},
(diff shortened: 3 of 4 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7798` or branch name `update/6841-supported-countries-fallback` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=update/6841-supported-countries-fallback) :rocket:

---

Build info: 

- Latest commit: 10d1cb52ed2162800d8a06c5a30632c1cba67538
- Build time: 2023-11-28 08:53:54 UTC

oaratovskyi (approved):
The code looks good and works well!
Small observation: if we pass Andorra to Stripe, then we lose PO, because Stripe does not support Andorra, so prefill does not work and the merchant needs to fill all the data (including bank details) even after that they pick US that supports PO, but I guess it's known behaviour

> I've been thinking about it, and trying a couple of approaches, but I don't see a good way to properly warn us from the client. A track event could be an option, but I don't think it's the way to do it. In the end, it will not alert us directly. It will be probably better to ensure we only have a single source of truth for supported countries in the server, rather than trying to cover these gaps with alerts or fallbacks, WDYT? 🤔

Regarding this, I agree with single source of truth on our server, but I'll let @dmallory42 the final word!

ismaeldcom:
Thanks!

>but I guess it's known behaviour

Yep, it's intended. We already have all the supported countries on both lists, so any other country we will add will behave like that. The alternative test could be removing one country from the business types list on the local server, clearing the cache, and using it to onboard 🙂

dmallory42:
> I've been thinking about it, and trying a couple of approaches, but I don't see a good way to properly warn us from the client. A track event could be an option, but I don't think it's the way to do it. In the end, it will not alert us directly. It will be probably better to ensure we only have a single source of truth for supported countries in the server, rather than trying to cover these gaps with alerts or fallbacks, WDYT? 🤔

I agree, the original idea was that this issue is mainly just a kind of failsafe to prevent users not being able to onboard if we forgot to add the business types for a country or Stripe made an unexpected change.

Following up to make sure the server supported countries list is a single source of truth sounds good to me - WDYT?

ismaeldcom:
Created the follow-up issue https://github.com/Automattic/woocommerce-payments-server/issues/4237 🙂

PR #7873, opened 2023-12-11
Pass additional data to the intent authentication exception

hsingyuc:
Slack discussion p1702070481324349/1702053239.912509-slack-C022WMN88KG

Currently, we only throw an exception message `We're not able to process this payment. Please try again later.` and don't have further information for debugging.

Throw additional data `intent_meta_data` and `order_id` to the intent authentication exception so we have more information for the failed order. 

#### Testing instructions
1. Test with https://github.com/woocommerce/woocommerce/pull/42685
2. Update the [condition](https://github.com/Automattic/woocommerce-payments/blob/0d3bd80bca2f10a6a75b35b01d0f6004bc88436e/includes/class-wc-payment-gateway-wcpay.php#L1273) to `true`
3. Add a log `error_log(print_r($e,true));` behind the [catch](https://github.com/Automattic/woocommerce-payments/blob/0d3bd80bca2f10a6a75b35b01d0f6004bc88436e/includes/class-wc-payment-gateway-wcpay.php#L947) to see the additional data
4. Place an order via WooPay
5. Check the additional data in the debug log

#### If test with WooPay
See the test instruction https://github.com/Automattic/woopay/pull/2394

-------------------

**Post merge**

### DIFF
--- a/changelog/fix-pass-additional-data-to-intent-authentication-exception
+++ b/changelog/fix-pass-additional-data-to-intent-authentication-exception
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fix
+
+Pass additional data to the intent authentication exception
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -1033,7 +1033,8 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 
 			// Re-throw the exception after setting everything up.
 			// This makes the error notice show up both in the regular and block checkout.
-			throw new Exception( WC_Payments_Utils::get_filtered_error_message( $e ) );
+			$additional_data = method_exists( $e, 'getAdditionalData' ) ? $e->getAdditionalData() : [];
+			throw new Process_Payment_Exception( WC_Payments_Utils::get_filtered_error_message( $e ), 'process_payment_error', 0, null, $additional_data );
 		}
 	}
 
@@ -1272,7 +1273,13 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 				if ( $intent_meta_order_id !== $order_id ) {
 					throw new Intent_Authentication_Exception(
 						__( "We're not able to process this payment. Please try again later.", 'woocommerce-payments' ),
-						'order_id_mismatch'
+						'order_id_mismatch',
+						0,
+						null,
+						[
+							'intent_meta_data' => $intent->get_metadata(),
+							'order_id'         => $order_id,
+						]
 					);
 				}
 			}
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -881,7 +881,6 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 	 *
 	 * @return array|null An array with result of payment and redirect URL, or nothing.
 	 * @throws Process_Payment_Exception Error processing the payment.
-	 * @throws Exception Error processing the payment.
 	 */
 	public function process_payment( $order_id ) {
 		$order = wc_get_order( $order_id );
@@ -944,7 +943,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 
 			$payment_information = $this->prepare_payment_information( $order );
 			return $this->process_payment_for_order( WC()->cart, $payment_information );
-		} catch ( Exception $e ) {
+		} catch ( Throwable $e ) {
 			// We set this variable to be used in following checks.
 			$blocked_due_to_fraud_rules = $e instanceof API_Exception && 'wcpay_blocked_by_fraud_rule' === $e->get_error_code();
 
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -945,6 +945,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			$payment_information = $this->prepare_payment_information( $order );
 			return $this->process_payment_for_order( WC()->cart, $payment_information );
 		} catch ( Exception $e ) {
+			// error_log(print_r($e,true));
 			// We set this variable to be used in following checks.
 			$blocked_due_to_fraud_rules = $e instanceof API_Exception && 'wcpay_blocked_by_fraud_rule' === $e->get_error_code();
 
@@ -1277,6 +1278,7 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 						0,
 						null,
 						[
+							'error_code'       => 'order_id_mismatch',
 							'intent_meta_data' => $intent->get_metadata(),
 							'order_id'         => $order_id,
 						]
--- a/includes/class-wc-payment-gateway-wcpay.php
+++ b/includes/class-wc-payment-gateway-wcpay.php
@@ -945,7 +945,6 @@ class WC_Payment_Gateway_WCPay extends WC_Payment_Gateway_CC {
 			$payment_information = $this->prepare_payment_information( $order );
 			return $this->process_payment_for_order( WC()->cart, $payment_information );
 		} catch ( Exception $e ) {
-			// error_log(print_r($e,true));
 			// We set this variable to be used in following checks.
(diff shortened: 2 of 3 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7873` or branch name `fix/pass-additional-data-to-intent-authentication-exception` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/pass-additional-data-to-intent-authentication-exception) :rocket:

---

Build info: 

- Latest commit: ea9ee608f0742e7c726068569e03a2624ca865ff
- Build time: 2023-12-22 01:08:06 UTC

malithsen (commented):
@hsingyuc  I haven't yet tested the changes, but I have a couple of questions that I wanted to ask.

1. I see a few other instances of the `We're not able to process this payment. Please try again later.` exception in the `update_order_status` function. Do we know if these can be reached in the WooPay flow? If yes, should they be updated as well?

2. As far as I can see, WooPay doesn't display this error to the shopper. What do you think about changing the error message to better indicate the error and include the order_id and meta_order_id in the message itself? This would make it easier to get this onto production as we wouldn't need the WC dependency.

hsingyuc:
>I see a few other instances of the We're not able to process this payment. Please try again later. exception in the update_order_status function. Do we know if these can be reached in the WooPay flow? If yes, should they be updated as well?

I did process of elimination, we are not going through the other exception while testing this error.  Let me know if you found an edge case otherwise I think we don't need to worry about it.

>As far as I can see, WooPay doesn't display this error to the shopper. What do you think about changing the error message to better indicate the error and include the order_id and meta_order_id in the message itself? This would make it easier to get this onto production as we wouldn't need the WC dependency.

I agree with more clear error message but we do display the message to the shopper on the WooPay checkout page(See below).  Since you mentioned, it seems weird we show this message to shoppers.

Screen Shot 2023-12-20 at 4 42 37 PM

malithsen:
> I agree with more clear error message but we do display the message to the shopper on the WooPay checkout page(See below). Since you mentioned, it seems weird we show this message to shoppers.

@hsingyuc Ah thanks, you're right. WDYT of removing the [WCPay error message](https://github.com/Automattic/woopay/blob/e86f4acaa39f95b7e525432ca9766d78f11ed1d4/src/Checkout/PaymentsHandler.php#L239) from WooPay and showing only the generic "Merchant store order creation failed" part. 

If the merchant order fails, I don't think there'll be any actionable input for the shopper in the complete error message. We will still log the [complete error](https://github.com/Automattic/woopay/blob/e86f4acaa39f95b7e525432ca9766d78f11ed1d4/src/Checkout/PaymentsHandler.php#L229) in WooPay logs. I am suggesting this because, this way, we can include `intent_meta_order_id` and `order_id` directly in the error message and avoid introducing changes to the WC core, and associated backward compatibility issues.

hsingyuc:
>WDYT of removing the [WCPay error message](https://github.com/Automattic/woopay/blob/e86f4acaa39f95b7e525432ca9766d78f11ed1d4/src/Checkout/PaymentsHandler.php#L239) from WooPay and showing only the generic "Merchant store order creation failed" part.
If the merchant order fails, I don't think there'll be any actionable input for the shopper in the complete error message. We will still log the [complete error](https://github.com/Automattic/woopay/blob/e86f4acaa39f95b7e525432ca9766d78f11ed1d4/src/Checkout/PaymentsHandler.php#L229) in WooPay logs. I am suggesting this because, this way, we can include intent_meta_order_id and order_id directly in the error message and avoid introducing changes to the WC core, and associated backward compatibility issues.


For a quick fix, that's a good approach.  Do we need to get any input from designer, @pierorocca?

hsingyuc:
@malithsen After thinking about it more, does that mean we also change all the error messages to a generic message like `We're not able to process this request. Please refresh the page and try again.`.  Meaning shoppers won't see `Nonce is invalid` and `Coupon usage limit has been reached` anymore?

malithsen:
@hsingyuc 
> Meaning shoppers won't see Nonce is invalid and Coupon usage limit has been reached anymore?

Nonce is invalid is not actionable to a shopper. "Coupon usage limit..." can be useful though. 

~Just taking a step back, do you think we still need this logging, given that order number mismatches are visible via the Stripe logs? (context: p1703179801368429/1703043494.702709-slack-C02CCT3F1QA)~

~The intention of implementing `getAdditionalData` is to get the mismatched order IDs logged, so that it'd help debug the intermittent order failures further. Since the order ID mismatches are visible on the Stripe logs, it would be enough to continue the investigation without further logs?~

Edit: We seem to have failed orders that go against my initial hypothesis that they are failing because the woopay order id is equal to the merchant order id. So adding logs closer to the check would still be useful. P2: pdpOw2-45Y-p2#comment-3865

pierorocca:
If I'm following along correctly there seem to be different categories of errors - some not actionable, some too obscure for shopper comprehension, some that would be useful.

Could you please summarize all the use cases that need feedback?

malithsen:
> After thinking about it more, does that mean we also change all the error messages to a generic message like We're not able to process this request. Please refresh the page and try again. Meaning shoppers won't see Nonce is invalid and Coupon usage limit has been reached anymore?

That's what I was initially thinking but to avoid a review of error messages that would increase the scope of this PR, perhaps we can handle the `order_id_mismatch` error separately. We append the order IDs to the [error message](https://github.com/Automattic/woocommerce-payments/blob/9aa8390c8b027878c3d1b4492822f37c76cff924/includes/class-wc-payment-gateway-wcpay.php#L1273), and on the WooPay side we extract it and log it in `WooPayLogger`. Finally, to avoid order ids from showing up on the frontend, remove them before throwing the error back. It's not as neat as the current solution, but avoids WC core changes and keeps the error notices visible to shopper as is.

hsingyuc:
>If I'm following along correctly there seem to be different categories of errors - some not actionable, some too obscure for shopper comprehension, some that would be useful.
Could you please summarize all the use cases that need feedback?

That's correct.  We have three types of error messages below showing to shoppers. Initially, Malith mentioned updating the three messages below to one generic message and showing it to shoppers.   But now we might not do that because `Coupon usage limit has been reached` is actually helpful to shoppers.   Since we have you here, do you think we should remove `Merchant store order creation failed with the following error: ` or leave it as it is?
1) Nonce is invalid 
2) Coupon usage limit has been reached
3) Merchant store order creation failed with the following error: We're not able to process this payment. Please try again later.    -----> This happens because the order id is not the same as the intent meta order id

image

hsingyuc:
> We append the order IDs to the [error message](https://github.com/Automattic/woocommerce-payments/blob/9aa8390c8b027878c3d1b4492822f37c76cff924/includes/class-wc-payment-gateway-wcpay.php#L1273), and on the WooPay side we extract it and log it in WooPayLogger. Finally, to avoid order ids from showing up on the frontend, remove them before throwing the error back. It's not as neat as the current solution, but avoids WC core changes and keeps the error notices visible to shopper as is.

Are we confirming doing it this way?  Do you think we should do both or close this PR?   If we confirm doing this way, I can open PRs.

malithsen:
> Are we confirming doing it this way? Do you think we should do both or close this PR? If we confirm doing this way, I can open PRs.

@hsingyuc If you don't have any concerns about the new approach, I think we should do that as it would allow us to get these logging out faster.

hsingyuc:
PR created https://github.com/Automattic/woocommerce-payments/pull/7951

pierorocca:
Thanks @hsingyuc and @malithsen. Besides preventing errors as the first line of defense:

1. For supported errors like coupon code is invalid or limit has been reached, these would be displayed to the user in plain language in a way that provides useful information or further instruction

2. For non-recoverable errors, like "nonce is invalid" unrelated to payments processing, instead of surfacing a technical error to the user I'm thinking a plain language generic like "Oops, we've run into a problem. Please check out as guest." so that the shopper at least gets back to the store and tries again using another payment method. The actual technical message should still be logged.

3. For "Merchant store order creation failed with the following error:" this seems like a serious, critical bug that would be better off resolved than improving the error message. Is there any circumstance where this error would legitimately occur? Agree with Malith's assessment to not include the order id in the shopper message but to include it in the logs.

@gpressutto5 recently updated many notifications to a new component. That work should be a good inventory of existing errors and error messages. Would there be value in taking that inventory of those error codes and come up with a revised list of front-end error messages?

hsingyuc:
Close the PR because [this PR ](https://github.com/Automattic/woocommerce-payments/pull/7951)fixed the issue.

Issue #6774, opened 2023-07-14
Show clear info in UI when there will be no payout because merchant balance is below deposit minimums (e.g. AUD$5)


EjayhanFernandes:
### Description

The estimated deposit date displayed for users in WooCommerce Payments may draw confusion when the minimum deposit amount is not yet met. 

![https://d.pr/i/r3y5wm+](https://d.pr/i/r3y5wm+)
For example: The minimum deposit amount for AUD currency is $5, as stated in the [doceumentation](https://woocommerce.com/document/woocommerce-payments/deposits/deposit-schedule/#minimum-deposit-amounts). However, it is worth noting that the dashboard does not provide a clear indication of this requirement. As a result, users may be left wondering why their deposit did not go through on the estimated date.

To improve the user experience, we can incorporate a tooltip or a similar prompt when the deposit amount is below the minimum requirement. This would explain that the deposit will be processed once the minimum balance is exceeded. Additionally, we can include a note about minimum balances in a prominent location in the user interface

### Additional context

- [Slack conversation](https://a8c.slack.com/archives/C7U3Y3VMY/p1689309505749639)
- 6535410-zen

**Update**
A proposed solution for this issue is a banner on the **Payments → Overview** screen, as per @bradyv's [design](https://www.figma.com/file/aDFNTeogq2cXSr4Nqeni8g/%5BWCPay%5D-Instant-Deposits?type=design&node-id=294-83778&mode=design&t=exSeHwn5tRxr33zg-4).

When the available balance is below the minimum deposit threshold:
* Remove the schedule string as it causes confusion when deposits aren't happening
* Show the banner below
      <img width="500" alt="image" src="https://user-images.githubusercontent.com/889872/282548013-cdba6fc9-faa5-4b0f-8695-844fe3ee6d9e.png">

csmcneill:
Please note that the previously documented $5 minimum deposit amount for AUD amount has since been removed from [our documentation](https://woocommerce.com/document/woocommerce-payments/deposits/deposit-schedule/#minimum-deposit-amounts). 

Discussion in p1689353300538449-slack-C03EE4CPPK5

haszari:
I've clarified the title of the issue to `Show clear info in UI when there will be no payout because merchant balance is below deposit minimums (e.g. AUD$5)` – that seems to be the main problem here! @EjayhanFernandes @csmcneill let me know if I've misunderstood, feel free to tweak the title.

[...]

PR #7875, opened 2023-12-11
Show deposit schedule message if account's deposits are unrestricted

### LINKED ISSUE #7874: Deposits - Overview: When available balance is at or below zero, no schedule is shown.
### Describe the bug

When a merchants available balance is at or below $0, no schedule is shown. This leads merchants to incorrectly assume that deposits are blocked or paused.

The deposit schedule should always show when deposits are not blocked or restricted.

image

Jinksi:
Fixes #7874

1. Renders the deposit schedule message on the Payments → Overview screen unless the account's deposits are blocked or restricted.

This helps communicate to merchants that they have an expected upcoming deposit, even if their available balance <= 0.

2. The "next deposit is scheduled" part of the message will only be rendered if there is available balance. Without available balance, this part of the message will not render. (in the future, this will only render if above the minimum automatic deposit threshold #6774)

image

#### Testing instructions

* Check that the message is rendered for accounts without deposit restrictions (use MC tools to add/remove restrictions to account)
	* With an available balance of 0
	* With a negative available balance
* Check that the message is not rendered
	* Deposits blocked
	* Deposits restricted

-------------------

**Post merge**

### DIFF
--- a/client/components/deposits-overview/index.tsx
+++ b/client/components/deposits-overview/index.tsx
@@ -45,9 +45,8 @@ const DepositsOverview: React.FC = () => {
 
 	const availableFunds = overview?.available?.amount ?? 0;
 
-	// If the account has deposits blocked, there is no available balance or it is negative, there is no future deposit expected.
-	const isNextDepositExpected =
-		! account?.deposits_blocked && availableFunds > 0;
+	// If the account has deposits blocked, there is no future deposit expected.
+	const isNextDepositExpected = ! account?.deposits_blocked;
 	// If the available balance is negative, deposits may be paused.
 	const isNegativeBalanceDepositsPaused = availableFunds < 0;
 	const hasCompletedWaitingPeriod =
--- a/client/components/deposits-overview/index.tsx
+++ b/client/components/deposits-overview/index.tsx
@@ -99,6 +99,7 @@ const DepositsOverview: React.FC = () => {
 				<CardBody className="wcpay-deposits-overview__schedule__container">
 					<DepositSchedule
 						depositsSchedule={ account.deposits_schedule }
+						showNextDepositDate={ availableFunds !== 0 }
 					/>
 				</CardBody>
 			) }
--- a/client/components/deposits-overview/index.tsx
+++ b/client/components/deposits-overview/index.tsx
@@ -54,10 +54,7 @@ const DepositsOverview: React.FC = () => {
 		wcpaySettings.accountStatus.deposits?.completed_waiting_period;
 	// Only show the deposit history section if the page is finished loading and there are deposits. */ }
 	const showRecentDeposits =
-		! isLoading &&
-		deposits?.length > 0 &&
-		!! account &&
-		isDepositsUnrestricted;
+		! isLoading && deposits?.length > 0 && !! account;
 
 	// Show a loading state if the page is still loading.
 	if ( isLoading ) {
--- a/client/components/deposits-overview/index.tsx
+++ b/client/components/deposits-overview/index.tsx
@@ -35,6 +35,9 @@ const DepositsOverview: React.FC = () => {
 		overview,
 		isLoading: isLoadingOverview,
 	} = useSelectedCurrencyOverview();
+	const isDepositsUnrestricted =
+		wcpaySettings.accountStatus.deposits?.restrictions ===
+		'deposits_unrestricted';
 	const selectedCurrency =
 		overview?.currency || wcpaySettings.accountDefaultCurrency;
 	const { isLoading: isLoadingDeposits, deposits } = useRecentDeposits(
@@ -45,8 +48,6 @@ const DepositsOverview: React.FC = () => {
 
 	const availableFunds = overview?.available?.amount ?? 0;
 
-	// If the account has deposits blocked, there is no future deposit expected.
-	const isNextDepositExpected = ! account?.deposits_blocked;
 	// If the available balance is negative, deposits may be paused.
 	const isNegativeBalanceDepositsPaused = availableFunds < 0;
 	const hasCompletedWaitingPeriod =
@@ -56,7 +57,7 @@ const DepositsOverview: React.FC = () => {
 		! isLoading &&
 		deposits?.length > 0 &&
 		!! account &&
-		! account?.deposits_blocked;
+		isDepositsUnrestricted;
 
 	// Show a loading state if the page is still loading.
 	if ( isLoading ) {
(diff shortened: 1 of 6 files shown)


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7875` or branch name `fix/7874-show-deposit-schedule-when-zero-available-balance` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=fix/7874-show-deposit-schedule-when-zero-available-balance) :rocket:

---

Build info: 

- Latest commit: 8550239f77618cf55be33f5b34eadcc32e9739ee
- Build time: 2023-12-13 03:48:39 UTC

csmcneill:
I have a couple of live accounts. I enabled this PR via Jetpack Beta and tested it in a couple of scenarios; it LGTM so far. Sharing these here in case they help.

Live mode with a pending balance, no available balance, and in the new account waiting period:
live no available

Test mode with an available balance:
live test mode with available

Live mode with no transactions:
live no balance

Activating this with the WCPay dev tools _did_ produce a fatal error on one site, but it was because I was running an outdated build of the dev tools — doesn't seem to be an issue with the PR.

client/components/deposits-overview/index.tsx
```diff
@@ -57,7 +57,7 @@ const DepositsOverview: React.FC = () => {
 		! isLoading &&
 		deposits?.length > 0 &&
 		!! account &&
-		! account?.deposits_blocked;
+		isDepositsUnrestricted;
```
brucealdridge: I think we still want to show recent deposits if deposits are restricted. 
Actually recent deposits should show even if they're blocked IMO

Jinksi: Good point, I'll test and remove this requirement.

Jinksi: Updated in 867116c.

**Deposits hard blocked** – Deposits card not rendered anyway

image

**Deposits soft blocked** – Deposits card not rendered anyway

image 2


**Deposits schedule restricted** – recent deposit list shown

image 3

client/globals.d.ts
```diff
@@ -30,7 +30,11 @@ declare global {
 			country?: string;
 			paymentsEnabled?: boolean;
 			deposits?: {
-				status: string;
+				status;
```
Jinksi:
```suggestion
				status: string;
```

Jinksi:
My concern with this one is that the estimated date for daily schedule can be wildly incorrect.

If the account has zero available or pending, there is no way this account will receive a deposit tomorrow.

image

client/components/deposits-overview/index.tsx
```diff
@@ -98,10 +95,11 @@ const DepositsOverview: React.FC = () => {
 			</CardHeader>
 
 			{ /* Deposit schedule message */ }
-			{ isNextDepositExpected && !! account && (
+			{ isDepositsUnrestricted && !! account && (
 				<CardBody className="wcpay-deposits-overview__schedule__container">
 					<DepositSchedule
 						depositsSchedule={ account.deposits_schedule }
+						showNextDepositDate={ availableFunds !== 0 }
```
shendy-a8c: Is this intentional to show next deposit date string if available balance is not 0? That means if it's negative too.

For example, every time I create an instant deposit, not sure why the available balance becomes negative even though the total balance (available + pending) is correct.

In this case (when available balance is negative), I can see the next deposit date string.

Screenshot 2023-12-12 at 23 40 10

Also, per the issue https://github.com/Automattic/woocommerce-payments/issues/7874 the next deposit date string shouldn't show if available balance is at or below zero.

shendy-a8c: Or you meant

```suggestion
						showNextDepositDate={ availableFunds > 0 }
```

client/components/deposits-overview/index.tsx
```diff
@@ -35,6 +35,9 @@ const DepositsOverview: React.FC = () => {
 		overview,
 		isLoading: isLoadingOverview,
 	} = useSelectedCurrencyOverview();
+	const isDepositsUnrestricted =
+		wcpaySettings.accountStatus.deposits?.restrictions ===
```
shendy-a8c, line 39: Why do you check for `wcpaySettings.accountStatus.deposits.restrictions` instead of `account.deposits_blocked`?

If I understand it correctly, `account.deposits_blocked` takes account new account waiting period and `wcpaySettings.accountStatus.deposits.restrictions` does not.

Just curious.

Current code would work because [the deposits card won't show if available balance is 0 and there is no deposits](https://github.com/Automattic/woocommerce-payments/blob/251fb285089a073f6d7f7279f34b56c6352a6ae9/client/components/deposits-overview/index.tsx#L86-L89) anyway. When in waiting period, merchant won't have deposits.

shendy-a8c (approved):
Overall looks good and tested well per testing instructions.

I have a question when available balance is negative, why the next date string shows up.

shendy-a8c (changes requested):
Overall looks good and tested well per testing instructions.

I have a question when available balance is negative, why the next date string shows up.

client/components/deposits-overview/index.tsx
Jinksi: I think you're right, we should go with showing this if there is a positive available balance.

Jinksi: Resolved in b6b2523f798de979fdc446d86d2eda648a23d59d

client/components/deposits-overview/index.tsx
Jinksi: AFAICT `account.deposits_blocked` [only checks for deposits blocked](https://github.com/Automattic/woocommerce-payments-server/blob/4f187b11eb9ca3237e4806a2f7de406a6f0d53a7/server/wp-content/rest-api-plugins/endpoints/wcpay/class-deposits-controller.php#L238), and doesn't take into consideration schedule restrictions.

In this case, we want to check for both schedule restrictions and blocked deposits, which `wcpaySettings.accountStatus.deposits.restrictions === 'deposits_unrestricted'` covers.

I could include a check for `wcpaySettings.accountStatus.deposits.completed_waiting_period === true`, that might be more explicit. What do you think?

Jinksi:
Rebased to `release/6.9.2` and removed all commits carried over from `develop`.

client/components/deposits-overview/index.tsx
shendy-a8c: Thank you for explaining. I don't think it's necessary to check waiting period.

shendy-a8c (approved):
Looks good. Retested to make sure and everything looks good.
Also, I made sure the rebase went OK by comparing with the PR branch before rebased, https://github.com/Automattic/woocommerce-payments/pull/7896/files. I still had it on my local minus your latest commit.

PR #7890, opened 2023-12-12
WooPay Direct Checkout

lovo-h:

This PR introduces the WooPay direct checkout flow on the merchant site. The WooPay direct checkout flow improves the user experience for shoppers who have opted-in for express checkout through WooPay. 

Previously, shoppers who opted-in for express checkout would be navigated to the merchant site's checkout page and then immediately redirected to the express checkout page. Now, this experience has been improved so that these shoppers are instead immediately navigated to the express checkout page.

By removing the unnecessary navigation to the merchant site's checkout page and then to the WooPay site's checkout page, the WooPay direct checkout flow helps reduce the amount of time it takes a WooPay shopper to complete an order.

#### Testing instructions

* See testing instruction sin 2390-gh-Automattic/woopay.

-------------------

**Post merge**


botwoo:
### Test the build 

#### Option 1. Jetpack Beta

- Use this build by searching for PR number `7890` or branch name `poc/woopay-direct-checkout` in your-test.site/wp-admin/admin.php?page=jetpack-beta&plugin=woocommerce-payments          

:rocket: [Launch a JN site with this branch](https://jurassic.ninja/create/?jetpack-beta&shortlived&nojetpack&woocommerce&woocommerce-payments-dev-tools&branches.woocommerce-payments=poc/woopay-direct-checkout) :rocket:

---

Build info: 

- Latest commit: ca0a77eee2ec3ea04627bbd4dac304c85205ff2d
- Build time: 2024-01-29 17:26:26 UTC

ricardo:
@lovo-h FYI I ended up mistakenly pushing a [commit](https://github.com/Automattic/woocommerce-payments/commit/62b24e540f8e895ef7de752edb7cc001498ac77f) into this branch for #8088, so I force-pushed-removed the commit from this branch. The previous commit hashes should be the same.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
```diff
@@ -0,0 +1,167 @@
...
+
+		const { redirect_url: redirectUrl } = await woopaySessionData;
+
+		return redirectUrl;
+	}
+
+	static getCheckoutRedirectElements() {
+		// TODO: Add other checkout elements, if needed.
+		const checkoutButton = document.querySelector(
+			'.wc-proceed-to-checkout .checkout-button'
```
bborman22: Could we add a selector here now to capture block checkout?

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
```diff
@@ -0,0 +1,167 @@
...
+			} );
+		} );
+	}
+
+	static redirectToWooPay( elements ) {
+		elements.forEach( ( element ) => {
+			element.addEventListener( 'click', async ( event ) => {
+				event.preventDefault();
+
+				const woopayRedirectUrl = await this.sendSessionDataToWooPay();
```
bborman22, line 231: This might be a pre-emptive optimization, but I was wondering if this redirect URL could be stored once on the object (or maybe even on a merchant level cookie?) so that we don't have to create a new WooPay session each time? I've confused myself a bit on this topic already so just wanted to ask the question. Same idea could apply to `redirectToWooPaySession`.

bborman22 (commented):
As I was reviewing #8088 I figured I'd get a start on reviewing this whole feature PoC overall as well to get it merged into develop behind the feature flag.

I left a couple notes on the code, but overall it looked really good and seemed pretty extendable to me as well! I'll be reviewing the functionality and edge cases next, but figured I'd start the review process now with some comments.

includes/class-wc-payments.php
```diff
@@ -1484,6 +1487,20 @@ function ( $container ) {
...
+	 *
+	 * @return void
+	 */
+	public static function maybe_enable_woopay_direct_checkout() {
+		if ( ! WC_Payments_Features::is_woopay_enabled() ) {
+			return;
+		}
+
+		$woopay_direct_checkout = new WC_Payments_WooPay_Direct_Checkout();
+		$woopay_direct_checkout->init();
```
bborman22, line 1474: Once the feature is enabled, this ends up adding about 2kB to every page load by including the direct checkout script. Given the mini cart needing to be captured, I'm not sure we can escape this, but wanted to mention it in case there is a good solution as script bloat has been a recent complaint in WC and WooPayments.

client/checkout/woopay/direct-checkout/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+window.addEventListener( 'load', async () => {
+	if ( ! WooPayDirectCheckout.isWooPayEnabled() ) {
+		return;
+	}
+
+	WooPayDirectCheckout.init();
+
+	const checkoutElements = WooPayDirectCheckout.getCheckoutRedirectElements();
+	const isThirdPartyCookieEnabled = await WooPayDirectCheckout.isWooPayThirdPartyCookiesEnabled();
```
bborman22, line 14: We discussed on our call about creating an issue to explore alternative ways to determine third party cookie status. We also discussed that once we add things like the mini-cart to the selectors, this could run on every page load so we should just run this once and store on a session level cookie for example.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
bborman22: Based on our discussion during the call, we should add that selector now as part of this PR to get blocks compatibility right away.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
bborman22: There are a lot of options to go with here, but that discussion fits better under our two optimizations task under milestone 2 in the https://github.com/Automattic/woopay/issues/2440.

includes/class-wc-payments.php
bborman22: Good to note this, but currently there isn't much of an option otherwise. One thing to note is that once we combine the first party auth and direct to woopay connect iframes we might actually come out even as we'll remove some duplicate iframe code.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
```diff
@@ -0,0 +1,167 @@
...
+		iframePostMessage: null,
+	};
+	static listeners = {
+		setIframePostMessage: () => {},
+		setTempThirdPartyCookie: () => {},
+		setIsWooPayThirdPartyCookiesEnabled: () => {},
+		setIsUserLoggedIn: () => {},
+		setWoopaySessionResponse: () => {},
+	};
+	static actionCallback = {
```
bborman22, line 39: Based on our discussion it might be helpful to drop a brief description of this functionality in the PR description for future reference.

bborman22 (changes requested):
I gave this a code review while reviewing the feature flag PR in #8088. I also just completed a functional review as well to confirm several of the following:
- The redirect doesn't apply when the cart is not WooPay eligible
- The redirect doesn't trigger when WooPay is down (earlier requests fail)
- All for scenarios listed in the WooPay PR tested correctly and as expected for me:
    - 3rd-Party Cookies Enabled & Customer Unauthenticated From WooPay Redirects to Merchant Checkout
    - 3rd-Party Cookies Enabled & Customer Authenticated From WooPay Redirects to WooPay Checkout
    - 3rd-Party Cookies Disabled & Customer Unauthenticated From WooPay Redirects to Merchant Checkout
    - 3rd-Party Cookies Disabled & Customer Authenticated From WooPay Redirects to WooPay Checkout
- I focused a lot on the third option there as it will be the primary user experience for most shoppers and confirmed that there were no unnecessary requests or redirects in those scenarios.

I dropped a few notes on this PR that would need an update, but also the obvious note to fully flesh out the PR description to get out of draft state.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
lovo-h: Good call. Added in bcb915b0f165b6f44aa3364512b11eda31e3d8ba.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
lovo-h: Good call. I ended up adding inline comments for easy reference in 446f6ee7a850c51d7e975e2c164b46409959b724.

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
```diff
@@ -0,0 +1,254 @@
...
+
+		const postMessage = await this.state.iframePostMessage;
+		postMessage( { action: 'setTempThirdPartyCookie' } );
+
+		// Since iFrame is refreshed, we need to re-init the iframePostMessage promise.
+		this.state.iframePostMessage = new Promise( ( resolve ) => {
+			this.listeners.setIframePostMessage = resolve;
+		} );
+
+		return await setTempThirdPartyCookie;
```
bborman22, line 117: After thinking this over for the weekend I finally realized what was trippping me up most about this piece and it's the `return await`. I'm more used to seeing something like: `await this.setTempThirdPartyCookie()`, at first I thought this might not work because of the listeners in the React component, but with some looking / testing I found that these are already being called as I was thinking in the `direct-checkout/index.js`. Testing locally with removing the `return await`s and just using `return` from these functions everything still seemed to work as expected.

I wouldn't say this is a blocking change, but wanted to mention it to get your opinion and if that also feels cleaner to you (or if there is a piece of the functionality I'm missing that is reliant on the await within these functions)?

bborman22 (approved):
I was able to test this again with the latest commits and with the feature flag in #8088 all with success. The same test scenarios listed above all passed still. I did drop one more comment on the piece about Promises that was tripping me up originally, let me know what you think.

One note that is outside the scope of this PR as it's behind a feature flag, but the performance on determining third party cookies seemed a lot slower in my testing today and I often never got the redirect to WooPay and only went right to merchant as the `connect` refresh was taking more than 3 seconds. This may have just been an issue with my local set up though, but wanted to monitor this and look for chances to improve that performance (if it's actually an issue that is).

client/checkout/woopay/direct-checkout/woopay-direct-checkout.js
lovo-h: Good call out. We discussed this on Slack and agreed to leave the code as-is since removing the `await` from `return await...` would be equivalent and appear a bit cleaner, but we may [unintentionally bypass the try/catch mechanism discussed in this SO](https://stackoverflow.com/a/42750371).

lovo-h:
> ... the performance on determining third party cookies seemed a lot slower in my testing today and I often never got the redirect to WooPay and only went right to merchant as the connect refresh was taking more than 3 seconds. This may have just been an issue with my local set up though, but wanted to monitor this and look for chances to improve that performance (if it's actually an issue that is).

We discussed this on Slack, and although this was not the case on my local setup, we did observe a significant slowdown on your local setup.

Just to note, we plan on creating an issue to replace the mechanism that detects third-party cookies (See https://github.com/Automattic/woocommerce-payments/pull/7890#discussion_r1467826676). This will make it so that we can do away with the `connect` page "refresh", therefore optimizing the performance some more.

Issue #7577, opened 2023-10-29
Tax is calculated twice for virtual products if Apple/Google Pay is used on product pages


grig-23:
### Describe the bug
With certain tax settings, tax is calculated twice for virtual products if Apple/Google Pay is used on product pages.

### To Reproduce

Settings:

<img width="1607" alt="Screenshot 2023-10-29 at 13 45 52" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/c8569b06-86c0-4f4c-b43d-8d02eb684f20">

<img width="1451" alt="Screenshot 2023-10-29 at 13 46 07" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/14e34f21-19b0-4c7a-b618-34a0b009656c">

<img width="1615" alt="Screenshot 2023-10-29 at 13 46 55" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/b0ff1eaf-5bce-4329-96eb-a6f80c13a427">

Product settings:

<img width="1635" alt="Screenshot 2023-10-29 at 13 48 53" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/50cf667e-cdca-4e4d-8fad-1319396d2d7f">

1. Open a browser that works with Apple or Google Pay set up.
2. Click the Apple or Google Pay button on the product page.
3. The tax is already included into the product price but it's added again in the Apple/Google Pay window on the product page:

<img width="674" alt="Screenshot 2023-10-29 at 13 50 02" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/a781d63a-2dd8-4075-96a3-8492b44f952f">

If a product is not virtual, then tax is added for a couple of seconds and then removed:

<img width="348" alt="Screenshot 2023-10-29 at 13 52 15" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/fc104395-e747-4ed9-b13a-484fde1be329">
<img width="387" alt="Screenshot 2023-10-29 at 13 52 24" src="https://github.com/Automattic/woocommerce-payments/assets/13084901/cdac61d1-2f95-4bb9-95a3-5d045ad83d26">

Also, this doesn't happen on cart or checkout pages.

### Actual behavior

Tax is added on top of the product price. If product is virtual, that price with 2x tax remains displayed in Apple/Google Pay window, while if a product is not virtual - it's only displayed for a few seconds and then correct price is displayed.

### Expected behavior

[...]

PR #7910, opened 2023-12-14
Add support for WooCommerce Deposits when using Apple Pay and Google Pay

### LINKED ISSUE #7907: Express Pay buttons expect extension field names to be prefix with `addon-` which isn't always the case.
### Describe the bug

The JavaScript for the Express Pay button puchases on the product page expects fields added by other extensions to use the prefix `addon-`

https://github.com/Automattic/woocommerce-payments/blob/e5bca8df8ccbb61b1707cbd947f413b1bacb9f27/client/payment-request/index.js#L176-L194

This isn't always the case, as many extensions (possibly most) use the prefix `wc_extensionname` for custom fields on the front end. This is certainly the case with deposits and bookings.

Deposits uses the field `wc_deposit_option` for selecting whether to pay a deposit or the full amount. Bookings uses a number of fields: `wc_bookings_field_start_date_month`, `wc_bookings_field_start_date_day`, `wc_bookings_field_start_date_year`, `min_date`, `max_date` and `timezone_offset` for daily blocks, others exist for part day blocks.

As a default, I think it would be good to include fields that begin with `wc_` in the form values submitted to the express pay buttons. A JavaScript filter could also be made available to allow extensions to add other fields to be included when the express pay buttons add items to carts.

This issue came out of testing deposits compatibility with this extension, see woocommerce-deposits#569.

The result of this assumption for Deposits is that the user always gets charged the full amount when using express pay buttons on the product page. This is regardless of their selection.

### To Reproduce

1. Activate Payments and Deposits
2. Create a product with a deposit: make the deposit optional, ie allow purchesers to chose whether to pay in full or pay a deposit
3. Enable Express Pay buttons
4. Visit the product create above
5. Select "Pay deposit"
6. Click an express pay button (eg, Google Pay)
7. The full amount will be charged to the user.

### Actual behavior

Incomplete form data is used on the product page.

### Screenshots

Screen Shot 2023-12-14 at 10 50 40 am

Screen Shot 2023-12-14 at 10 51 14 am

### Expected behavior

Full form data be used when adding items to cart for express pay buttons

(In case of Deposits, that selecting deposit results in charging the deposit only)

* OS: macOS
* Browser Chrome
* Version 120

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Browser [e.g. stock browser, safari]
* Version [e.g. 22]

### Additional context

woocommerce-deposits#569.

peterwilsoncc:
Fixes #7907
Fixes woocommerce/woocommerce-deposits#569

This modifies the add to cart AJAX request made during express payments to include field names prefixed with `wc_`. This is to account for fields added by extensions, for example the deposits extensions fields `wc_deposit_option` and `wc_deposit_payment_plan`.

The goal is to include extension fields during an express payment request so if an input affects the price, that is reflected in the data added to the cart. 

#### Testing instructions

This has been tested with the deposits extension.

1. Enable Express Pay options (Google and Apple Pay) to be displayed on product pages.
1. Install and enable WooCommerce Deposits
3. Create a virtual simple product with an optional deposit of 50%
6. Visit virtual simple product with optional deposit of 50% 
   1. Select "Pay Deposit"
   2. Click Apple/Google Pay button
   3. The price in the platforms popover should reflect the deposit amount
   4. Proceed with purchase
   5. The deposit amount should be recorded as paid against the order
   6. The remaining amount should be recorded against the order
1. Update the Deposit options ensure it works as expected. Special attention to Enable Deposits and Default Deposit Selected Type.
    image
7. Update the product to Physical (or create a new product) an check the price is not updated after the shipping methods load (more info about that in the second part of https://github.com/Automattic/woocommerce-payments/issues/7577#issuecomment-1847967971)
4. Create a virtual simple product with an optional payment plan 
8. Visit virtual simple product with optional payment plan
   1. Select "Pay Deposit"
   3. Select payment plan
   4. Click Apple/Google Pay button
   5. Proceed with purchase
   6. The first payment amount should be recorded as paid against the order
   7. The remaining amount should be recorded against the order

-------------------

**Post merge**


gpressutto5:
Thank you for starting this PR @peterwilsoncc. I fixed the remaining issues and updated the test instruction.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -1050,6 +1062,7 @@ public function ajax_get_selected_product_data() {
 			$product      = wc_get_product( $product_id );
 			$variation_id = null;
 			$currency     = get_woocommerce_currency();
+			$is_deposit   = isset( $_POST['wc_deposit_option'] ) ? 'yes' === $_POST['wc_deposit_option'] : null;
```
timur27: This is good enough in terms of security, in my opinion, as we're strictly checking the data received against the `yes` string, declining any potentially malicious values. However, having multiple layers of security by adding data sanitization won't hurt and will increase the probability of not missing any important bit in the future if we will expand the usage of the `wc_deposit_option` param. 

To sum this up, could we consider expanding this line with input sanitisation by e.g. adding `wc_clean( wp_unslash( ) )` or similar?

client/checkout/woopay/express-button/use-express-checkout-product-handler.js
```diff
@@ -104,14 +105,23 @@ const useExpressCheckoutProductHandler = ( api ) => {
...
+		}
 
 		if ( addOnForm ) {
 			const formData = new FormData( addOnForm );
 
 			formData.forEach( ( value, name ) => {
 				if (
 					/^addon-/.test( name ) ||
-					/^wc_gc_giftcard_/.test( name )
+					/^wc_/.test( name ) ||
```
timur27: Any objections on aligning this with a more concise version from `/payment-request/index.js`?

```js
/^(addon-|wc_)/.test( field.name )
```

client/payment-request/index.js
```diff
@@ -189,8 +189,19 @@ jQuery( ( $ ) => {
...
+			// Ensure allowedFieldNames is an array.
+			if ( ! Array.isArray( allowedFieldNames ) ) {
+				allowedFieldNames = [ allowedFieldNames ];
+			}
 			$.each( formData, ( i, field ) => {
-				if ( /^addon-/.test( field.name ) ) {
+				if (
+					allowedFieldNames.includes( field.name ) ||
+					/^(addon-|wc_)/.test( field.name )
+				) {
```
timur27: Non-blocking and nice-to-have (or a good candidate for a separate small issue): It'd be beneficial to streamline this `data` variable appending implementation with the identical one from `woopay/express-button/use-express-checkout-product-handler.js`, to avoid code duplication and all the potential future regressions because of it. Both are located in different places which might be complicating things, but just a little bit.

tests/unit/test-class-wc-payments-payment-request-button-handler.php
timur27, line 1: Could we please add a test here that will cover the scenario of [`$base_price`](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/class-wc-payments-payment-request-button-handler.php#L212-L214) being overridden by `WC_Deposits_Product_Manager::get_deposit_amount( $product, 0, 'display,' $base_price );` which was added in this PR? Please let me know if I'm missing something.

client/checkout/woopay/express-button/use-express-checkout-product-handler.js
timur27, line 1: I'd like to ask about two enhancements and the interconnection between them. This PR is:
* adding `wc_` prefix to the allowed field set shows that we are fixing the issue for all the extensions that add fields with this prefix
* creating a filter allows extensions add their fields to the set of allowed fields

Do you agree that maintaining only the filter in the long term is the best approach? This would eliminate the need for third-party extension-related implementation in the plugin. Do you see any drawbacks or am I overlooking something important with this perspective?

timur27 (commented):
Thanks @peterwilsoncc @gpressutto5 for improving this area! 🚀 

I went through the code and am leaving a few comments. The testing result will follow in a separate comment because it takes me a bit longer and I want to post these comments in the meantime to speed up the process.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -202,18 +202,30 @@ public function get_button_height() {
...
+	public function get_product_price( $product, ?bool $is_deposit = null ) {
 		// If prices should include tax, using tax inclusive price.
 		if ( $this->express_checkout_helper->cart_prices_include_tax() ) {
 			$base_price = wc_get_price_including_tax( $product );
 		} else {
 			$base_price = wc_get_price_excluding_tax( $product );
 		}
 
+		// If WooCommerce Deposits is active, we need to get the correct price for the product.
+		if ( class_exists( 'WC_Deposits_Product_Manager' ) && WC_Deposits_Product_Manager::deposits_enabled( $product->get_id() ) ) {
```
timur27, line 219: Are we running the test class code here?

timur27, line 219: Are we running the testing helper class here, or do I have some wrong environment configuration?

client/checkout/woopay/express-button/use-express-checkout-product-handler.js
gpressutto5: The fields from extensions under the WooCommerce domain will most likely start with wc_, but I don't think that's a rule. Third-party exceptions might not follow this pattern, and this filter allows them to add their fields here.
That said, sending these fields to the backend is worth nothing when we don't handle them properly in the backend. So, we will still need to add support for other extensions here manually. I'm removing this filter.

timur27:
I tested deposits with virtual and physical products. I successfully tested physical product payment by paying the deposit first and paying the rest with a pay-for-order flow. 

One thing that doesn't work for me as expected yet is the product price within the Google Pay window for a virtual product with deposits enabled. It isn't updated, although the final order reflects the deposit amount only. @gpressutto5 may I ask you to double-check this if possible? Thanks!

gpressutto5:
Thanks @timur27! It is working fine here, maybe it was fixed by something I changed now. Could you clear cache and test it again, please?

Jan-08-2024 12-40-38

peterwilsoncc:
Thanks for picking this up @gpressutto5! It looks in much better shape now than it did before I went on holiday.

I've tested it with the following and it seems fine:
* percentage, optional
* percentage, required
* fixed amount, optional
* fixed amount, required

For payment plan purchases, I am seeing JS errors on this branch: `wcpayPaymentRequestParams.product` is set to an empty string, so `wcpayPaymentRequestParams.product.total.amount` throws a type error on this line:

https://github.com/Automattic/woocommerce-payments/blob/e37202f27452616347251bdd0d6effc2ad1158e2/client/payment-request/index.js#L597

The key product details are:

* Price: $1
* Deposit: `Optional, payment plan` and `Required, payment plan` both affected
* Payment Plan: Two Easy Payments (but not important)
* Default Deposit Selected Type: Pay deposit

I'm not seeing this bug on trunk so I think it's coming from this branch rather than deposits.

client/payment-request/index.js
```diff
@@ -429,6 +438,14 @@ jQuery( ( $ ) => {
 				wcpayPaymentRequest.addToCart();
 			} );
 
+			// WooCommerce Deposits support.
+			// Trigger the "woocommerce_variation_has_changed" event when the deposit option is changed.
+			$( 'input[name=wc_deposit_option]' ).on( 'change', () => {
```
peterwilsoncc:
```suggestion
			$( 'input[name=wc_deposit_option],input[name=wc_deposit_payment_plan]' ).on( 'change', () => {
```

This causes same to happen when the payment plan is changed if multiple payment plans are available.

client/payment-request/index.js
```diff
@@ -429,6 +438,14 @@ jQuery( ( $ ) => {
...
+				$( 'form:has(input[name=wc_deposit_option])' ).trigger(
```
peterwilsoncc:
```suggestion
				$( 'form' ).has('input[name=wc_deposit_option],input[name=wc_deposit_payment_plan]').trigger(
```

As above. 

I've switched to using the `has()` jQuery method rather than the selector as I don't think some browsers support selector lists in early `:has()` implementations. But I may be hallucinating.

timur27 (approved):
> It is working fine here, maybe it was fixed by something I changed now. Could you clear cache and test it again, please?

@gpressutto5 It works for me now, thank you! I'm approving the PR from my side. Feel free to 🚢 once threads from Peter are resolved.

E2E tests are failing on the env setup stage, I assume it has something to do with how the PR is opened from a fork.

gpressutto5:
It is working now @peterwilsoncc! Can you test it again?

peterwilsoncc:
@gpressutto5 I was still hitting the JavaScript error for payment plans but was able to resolve that in eff996efeafdabb02191ab5a71b96726137174ed.

The product I used for testing was set up as:

* Type: Simple Product
* Price: $2
* Enable Deposits: Yes (either optional or required doesn't matter)
* Deposit Type: Payment Plan
* Default selected: Pay deposit
* Payment plans: Two Easy Payments (again, the details don't matter)

I've tested switching between different plans, switching between full and deposit and it seems to work.

Otherwise this looks to be working great and I appreciate your help with the things I missed over the break. I can't approve the PR (as it's officially mine) but please consider it so for your code.

peterwilsoncc:
Sorry, I didn't notice this until after the comment above.

The Google pay screen looks fine as it doesn't display the unit price on the product page. Apple Pay does and it's showing the full amount but attempting to charge the correct amount (which is an improvement)

Initially it shows the correct itemised amount but for non-virtual products it gets confused once the shipping address is selected.

1. Go to product page of a simple non-virtual product requiring deposit or payment plan
2. Select a shipping address 
3. Select a shipping type (if required)
4. The itemized price will display the full amount, the total the correct amount

Screen Shot 2024-01-11 at 12 15 03 pm

(unredacted address is fake)

I'm investigating now but if I don't push any changes in the next three or four hours then assume I couldn't get it figured out during my day.

gpressutto5:
I found out what is causing this issue and how to fix it in [WC_Payments_Express_Checkout_Button_Helper](https://github.com/Automattic/woocommerce-payments/blob/ba86be4c972e85152774bcf56a484c9a04f9cceb/includes/express-checkout/class-wc-payments-express-checkout-button-helper.php#L120). When paying for a deposit, the `line_subtotal` is the full price of the product, so we need to check `is_deposit` and use `deposit_amount`.
However, while working on that, I found another issue that also impacts the itemized price, but I couldn't fix it because it looks like it is a bug in WooCommerce Deposits. It does not follow the tax rules, so it is impossible to calculate taxes correctly. Is this a known issue?

#### This is the problem:
image

#### WooCommerce Deposits does not take tax into account

| Full Payment | Deposit |
|----|----|
|image|image|

Notice the Full Payment is $10, but the deposit plan totals $9 despite it being a plan that sums to 100%. These are my tax settings:

image

You can check more information about the tax in issue #7577.

When we use Apple Pay, we have the initial settings rendered into the page, and we calculate tax manually at this point, but after selecting an address, we get the updated data from the cart, so we need the correct data in the cart to fix this.
If we discover this is indeed a bug in WC Deposits and decide to fix this there, could we consider also updating the line_subtotal to the deposit amount so we don't need to handle those cases here too?

gpressutto5:
Hi @peterwilsoncc! Have you had a chance to read my comment above?

peterwilsoncc:
I'm unable to reproduce the issue you are seeing on the cart page or checkout pages. I am seeing it in the Apple Pay popup though.

I've changed the tax settings to match those in your screen shot and tested a product on both my dev site with an address in Australia and on my dev site with an address in the USA. In the screen shot below, the tax rate is 10%. I'm seeing the same result with both a deposit and payment plan.

**Deposit purchase - cart**
Screen Shot 2024-01-22 at 11 08 07 am

I'm running WooComerce-Deposts @ trunk.

gpressutto5:
@peterwilsoncc I was using WC Deposits 2.2.4 and updated to 2.2.5 now. I still have the error. I noticed it works as expected when using a percentage deposit, but the issue is always reproduced when using a payment plan. Is there anything wrong with my plans? Can you send me a the settings you are using in your plans and product?

image

image

peterwilsoncc:
@gpressutto5 Your payment plans look fine. I'm following this up with a team mate to see if we can figure it out.

This is the product I've been testing with, using an optional 30/70% payment plan similar to yours.

[wc-product-export-25-1-2024-1706155839195.csv](https://github.com/Automattic/woocommerce-payments/files/14046719/wc-product-export-25-1-2024-1706155839195.csv)

peterwilsoncc:
@gpressutto5 Are you able to send me the following so I can test further:

* export of tax rates in your environment
* Currency in your environment
* Store country in your environment

I can't recall where but I saw some code comments about `Account for Australian GST` relating to some tax settings so I'm wondering if I'm hitting special rules because I'm using my home town for the store's location.

gpressutto5:
@peterwilsoncc 

**WooCommerce** Version 8.3.1
**WooCommerce Deposits** Version 2.2.5
**Country:** United States
**Currency:** I'm using multi currency and tested with USD, GBP, and EUR


<details>
<summary>Tax settings</summary>

image

image

image

image

</details>

gpressutto5:
This is an example of item from cart. Notice `is_deposit` is true, `line_subtotal` is the full amount excluding tax, but it should be the deposit amount excluding tax. `line_subtotal_tax` should be the tax considering the deposit amount, instead it is showing the full amount tax.

#### Payment plan
```json
{
    "is_deposit": true,
    "payment_plan": 2,
    "key": "5ea5441ac283b4f25d32d4aa879cf82f",
    "product_id": 926,
    "variation_id": 0,
    "variation": [],
    "quantity": 1,
    "data_hash": "b5c1d5ca8bae6d4896cf1807cdf763f0",
    "deposit_amount": "3",
    "full_amount": 10,
    "line_tax_data": {
        "subtotal": {
            "1": 0.909091
        },
        "total": {
            "1": 0.909091
        }
    },
    "line_subtotal": 9.090909,
    "line_subtotal_tax": 0.909091,
    "line_total": 9.090909,
    "line_tax": 0.909091,
    "data": {}
}
```

#### Percent

```json
{
  "is_deposit": true,
  "payment_plan": 0,
  "key": "05d28410c987177e1e2916acdb73748c",
  "product_id": 926,
  "variation_id": 0,
  "variation": [],
  "quantity": 1,
  "data_hash": "b5c1d5ca8bae6d4896cf1807cdf763f0",
  "deposit_amount": "5",
  "full_amount": "10",
  "line_tax_data": {
    "subtotal": {
      "1": 0.909091
    },
    "total": {
      "1": 0.909091
    }
  },
  "line_subtotal": 9.090909,
  "line_subtotal_tax": 0.909091,
  "line_total": 9.090909,
  "line_tax": 0.909091,
  "data": {}
}
```

gpressutto5:
I created #8110 to run CI tests, and they passed. The tests that are not passing are the same that are broken in `develop` already.

The payment itself is working as expected. The remaining issue is just the tax price being displayed incorrectly when you click the ℹ button with a deposit in the cart, but we are still investigating if taxes are broken in WooCommerce Deposits.
I added a comment explaining this might not work correctly when using Deposits + Taxes so we can get this released and work on the remaining items in another issue. The payment works successfully with the correct amount.

@timur27 would you please review the new changes?
@peterwilsoncc feel free to review this too and let me know if you find anything that is not working correctly. We can create a new issue (maybe in the WooCommerce Deposits repository) to continue investigating the tax issue if it is reproducible.

peterwilsoncc:
> @peterwilsoncc feel free to review this too and let me know if you find anything that is not working correctly. We can create a new issue (maybe in the WooCommerce Deposits repository) to continue investigating the tax issue if it is reproducible.

I'll take a look.

I've been working through the tax display issues on the Deposits#569. Switching the store to USD and the default tax rates revealed the error for me. I'll need to test any changes against other payment gateways to make sure any changes to the Deposits code doesn't result in undercharging elsewhere.

peterwilsoncc:
@gpressutto5 I've created Deposits#583 for the data storage issue.

Modifying the data as items are added to the cart (be it in session storage or a draft order) makes sense to me but we'll need to identify the risk involved in making the change.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -224,7 +225,16 @@ public function get_product_price( $product, ?bool $is_deposit = null ) {
...
+				$deposit_type = WC_Deposits_Product_Manager::get_deposit_type( $product->get_id() );
+				if ( 'plan' === $deposit_type && 0 === $deposit_plan_id ) {
+					// Default to first (default) plan if no plan is specified.
+					$available_plan_ids = WC_Deposits_Plans_Manager::get_plan_ids_for_product( $product->get_id() );
+					if ( ! empty( $available_plan_ids ) ) {
+						$deposit_plan_id = $available_plan_ids[0];
+					}
+				}
+
+				$base_price = WC_Deposits_Product_Manager::get_deposit_amount( $product, $deposit_plan_id, 'display', $base_price );
```
timur27: Can the deposits-related logic be encapsulated within the deposits plugin any further? In this code, we seem to

1. take deposit type from deposits (dependency on Deposits)
2. get deposit plans for product (dependency on Deposits) and
    * establish deposit plan id based on it (currently within WooPayments)
4. get deposit amount (dependency on Deposits)

We essentially perform one business logic decision (getting the default plan if no plan is specified) within WooPayments, the rest happens by just calling the deposits and thus creating more coupling between the two which increases the probability of the WooPayments implementation breaking once something changes on the deposits side. I'm curious if, as WooPayments, we could just 

```php
// WC_Deposits::get_deposit_amount( $product, $base_price ) is just an example
$base_price = WC_Deposits::get_deposit_amount( $product, $base_price );
```

and this line will perform all the steps within the Deposits plugin scope. 
Is `Default to first (default) plan if no plan is specified` something that should be done by Deposits? I'd expect it to but I believe Deposits plugin developers will have more information on this. 

I'm happy to hear your opinion @peterwilsoncc.

includes/class-wc-payments-payment-request-button-handler.php
```diff
@@ -975,13 +985,14 @@ public function ajax_get_selected_product_data() {
...
-			$currency     = get_woocommerce_currency();
-			$is_deposit   = isset( $_POST['wc_deposit_option'] ) ? 'yes' === sanitize_text_field( wp_unslash( $_POST['wc_deposit_option'] ) ) : null;
+			$product_id      = isset( $_POST['product_id'] ) ? absint( $_POST['product_id'] ) : false;
+			$qty             = ! isset( $_POST['qty'] ) ? 1 : apply_filters( 'woocommerce_add_to_cart_quantity', absint( $_POST['qty'] ), $product_id );
+			$addon_value     = isset( $_POST['addon_value'] ) ? max( (float) $_POST['addon_value'], 0 ) : 0;
+			$product         = wc_get_product( $product_id );
+			$variation_id    = null;
+			$currency        = get_woocommerce_currency();
+			$is_deposit      = isset( $_POST['wc_deposit_option'] ) ? 'yes' === sanitize_text_field( wp_unslash( $_POST['wc_deposit_option'] ) ) : null;
+			$deposit_plan_id = isset( $_POST['wc_deposit_payment_plan'] ) ? absint( $_POST['wc_deposit_payment_plan'] ) : 0;
```
timur27, line 1012: Depending on the outcome of https://github.com/Automattic/woocommerce-payments/pull/7910#discussion_r1477971749, we might remove this parameter (because it's currently only used to check if a deposit plan is specified as `0 === $deposit_plan_id` and is part of the logic I suggest to move to deposits), and let Deposits get it from the request.

Otherwise, it looks good with a caveat that we could add a layer of security by verifying if a deposit plan with `$deposit_plan_id` exists in the system. This would require Deposits with the method like, e.g., `WC_Deposits_Plan::exists( $deposit_plan_id )` if we'll keep this parameter in our code, alternatively if we address https://github.com/Automattic/woocommerce-payments/pull/7910#discussion_r1477971749 then the verification will occur on the Deposits side (I think [currently it might fail for a plan which doesn't exist](https://github.com/woocommerce/woocommerce-deposits/blob/trunk/includes/class-wc-deposits-plan.php#L51-L57))

timur27 (commented):
Hey @gpressutto5 @peterwilsoncc! Thanks again for all your efforts so far! 🚀  I left two comments and look forward to following them up with you all. 

In terms of testing, I've tested deposits and deposit plans, and it works well for me, with the caveat that there are some open issues you moved to separate tickets. 👍 

Overall, I think it'd be great to merge this PR as soon as possible and keep all the non-critical items or items from different areas for follow-up issues due to PR's complexity at this point.

includes/class-wc-payments-payment-request-button-handler.php
gpressutto5: Unfortunately not. While I agree this logic seems to be more related to the Deposits extensions, when we use a payment request button, we skip key parts of the process, so we have to recreate it, [like we do for variations](https://github.com/Automattic/woocommerce-payments/blob/7448731cb746d6e6832d389557ba1fad610f03e0/includes/class-wc-payments-payment-request-button-handler.php#L257-L276).
For example, the method `get_product_data` is used to get the payment's initial data, which will be rendered as JSON with the page load. So, the product can't be added to the cart, but we still need to get the exact same data as if it were. This is something that is only ever done for payment requests, so this logic belongs here. Also notice in this method we call `get_product_price` without any arguments, which mean it will use the default data, so we could let WC_Deposits handle everything and return the default amount, but it is better to do it manually so it can be reused by the second method, `ajax_get_selected_product_data`.
When we call this method, we are creating a new cart manually, not using the "Add to cart" button, so we need to handle all the data we can by ourselves. Because of that, we can't have `get_product_price` always return the default amount for the deposit; we need to handle the selected options. So, the code that returns the price needs to support provided values instead of always using the default.

> Is Default to first (default) plan if no plan is specified something that should be done by Deposits?

Well, it is, but how is it specified if not by us? Updating 
`WC_Deposits::get_deposit_amount()` to read from `$_POST` is a bad pattern. Request params should be read in the controller and then we pass the data to `get_deposit_amount`.
We are passing these variables because we want to be able to overwrite `$is_deposit` and `$deposit_plan_id` as needed.

includes/class-wc-payments-payment-request-button-handler.php
gpressutto5: > because it's currently only used to check if a deposit plan is specified as 0 === $deposit_plan_id

When it is not `0`, we use the value. We only check if it is `0` to overwrite it.

> Otherwise, it looks good with a caveat that we could add a layer of security by verifying if a deposit plan with $deposit_plan_id exists in the system. 

I don't think checking if it exists is a security layer. We would just be replacing an unhandled error from Deposits with a handled error from our side, but the result would be the same and only displayed to someone who is trying to hack the system.
That said, I agree with checking if the plan is available for the product. Someone could buy a product that only accepts Plan 1 and Plan 2 using Plan 3, and I don't think anything is preventing them from doing that in `get_deposit_amount`, so I'm adding a check for that.

peterwilsoncc:
@gpressutto5 @timur27 I was considering this overnight and how it could be fixed in Deposits and worked around in Payments for the time being.

The risk of adding a work around in Payments is that once it's fixed, the maths may be doubled up here and reintroduce a problem.

Until deposits is fixed, would it be possible to remove itemisation from Apple Pay for carts containing a deposit? The total is correct. Provided, of course, that the Apple terms and conditions allow for this.

includes/class-wc-payments-payment-request-button-handler.php
timur27: Thanks for your response, Guilherme! Below, I'm raising a few more thoughts and open questions; sorry for prolonging this topic, I do appreciate your knowledge in this area and find it really helpful to learn, thanks to those discussions. I don't want this thread to block the whole PR, so please feel free to merge the PR, and, in the worst case, we can improve this small area separately if there is such a need. 

> For example, the method get_product_data is used to get the payment's initial data, which will be rendered as JSON with the page load. So, the product can't be added to the cart, but we still need to get the exact same data as if it were. This is something that is only ever done for payment requests, so this logic belongs here

As I understand from your explanation, there are two steps to building a dataset for a PRB, and the one described here is the first step, where we provide initial data, which might (and most probably will) change after running the second step through `wc_ajax_wcpay_get_selected_product_data` hook. Given that it's just initial data, which is subject to change, can we prepare it as if no deposit applies to a product? During the second dataset update, we will provide the real deposit data for a product. If such an approach would make sense, then we could 

1. Exclude deposit-related logic which updates the `$base_price`, from `get_product_price()`
2. Add deposit-related logic within `get_product_data()` right after `get_product_price()`, and make sure this deposit-related logic runs only during the second step (`ajax_get_selected_product_data`) and is skipped on the first step. We could distinguish between steps with `wc_deposit_option/wc_deposit_payment_plan` request params I guess.

This way, we will decouple a PRB-specific logic, which manages the flow for when getting the deposit-related price, from the actual get-deposit-related-price logic. Now, having this, we would know that we only calculate deposits when needed. And now, coming back to how it could make it, I still think (maybe I'm missing something) that Deposits could encapsulate all the steps within lines `228-237` into its own single method. 

> the code that returns the price needs to support provided values instead of always using the default

Yes, and I think the main point of my questions here is exactly if this support of provided deposit values should be within the Deposits area of responsibility. 


> Updating
WC_Deposits::get_deposit_amount() to read from $_POST is a bad pattern. Request params should be read in the controller and then we pass the data to get_deposit_amount.

I completely agree and thank you for bringing this up. Sorry that I missed this obvious thing. So we still are forwarding `$deposit_plan_id` further. Then the single method from Deposits could get the plan id as an argument.

includes/class-wc-payments-payment-request-button-handler.php
timur27: Thanks for mentioning in another comment that getting request params should be done here rather than further during the process: this almost makes my initial message in this thread irrelevant. Also, thanks for adding the check for plan availability for the product. 👍 I'm closing this thread.

includes/class-wc-payments-payment-request-button-handler.php
gpressutto5: > the first step, where we provide initial data, which might (and most probably will) change after running the second step through

The second step never runs for virtual products. Only products that require shipping update after initialized.

> Given that it's just initial data, which is subject to change, can we prepare it as if no deposit applies to a product?

Because of the issue above, if we start it with the incorrect value it will never be updated to the correct value if you're paying for a virtual product. The problem for physical products would be the price flashing wrong for a few seconds.

> Yes, and I think the main point of my questions here is exactly if this support of provided deposit values should be within the Deposits area of responsibility.

I agree it should be that plugin responsibility to set the price, but here we need the price in different circumstances, so it makes sense for me we have to ask for it. In an ideal scenario we would have the item properly added to the cart (not by us), then let the cart do the calculations and use the result. The problem is that in both of the methods we have here, we can't do that, this makes this logic specific to wcpay. So we need to either have wcpay with code to add compatibility to Deposits or the opposite. But I think it is something too specific for Apple Pay and will feel even more out of place if done in WC Deposits. What do you think?

gpressutto5:
Hi @peterwilsoncc. Itemization is not an issue anymore, only taxes, and removing it for carts with a deposit would require even more code specific for Deposits. I suggest one of the following:

1. Leave it as it is, and then when refactoring WC Deposits, consider fixing the possible math conflict.
2. Remove the itemization fix and fix it only in WC Deposits as suggested here: https://github.com/Automattic/woocommerce-payments/pull/7910#issuecomment-1888404095. Meanwhile, itemization being wrong for Deposits is a known conflict.

I like the option 2 better as it does not require updating WCPay, only WC Deposits.

includes/class-wc-payments-payment-request-button-handler.php
timur27: > The second step never runs for virtual products. Only products that require shipping update after initialized.

Ah, this changes a lot. And also, after this, the below makes sense to me.

> In an ideal scenario we would have the item properly added to the cart (not by us), then let the cart do the calculations and use the result. The problem is that in both of the methods we have here, we can't do that, this makes this logic specific to wcpay

I'm fine with keeping things as they are, thanks for the discussion!

gpressutto5:
Hi @peterwilsoncc, did you have a chance to look at the options I suggested? It would be great to merge or close this PR as soon as possible to avoid conflicts.

peterwilsoncc:
@gpressutto5 I think two is best, is it possible to remove itemization when there is a deposit so customers aren't confused.

I looked at doing it in deposits with the `wcpay_payment_request_hide_itemization` filter but it's not possible due to the or condition used in the if statement within `build_display_items()`

peterwilsoncc:
Thanks for changing the filter in 369509e13db8cb22bb1d702087c823079cb5876a, I've tested it with the following and it works as expected:

```php
/**
 * Hide itemization for Apple Pay and Google Pay when deposits are present.
 *
 * Compatibility with WooCommerce Payments. This filter is used to hide itemization
 * due to an issue in which the itemized amount shows the full price of the product
 * rather than the deposit amount.
 *
 * @todo Remove this filter once the issue is resolved in the extension.
 * @see https://github.com/woocommerce/woocommerce-deposits/issues/583
 *
 * @param bool $hide_itemization Whether to hide itemization for Apple Pay and Google Pay.
 * @return bool Modified value based on whether the cart has deposits.
 */
public function wcpay_payment_request_hide_itemization( $hide_itemization ) {
	// Hide itemization for carts containing deposits.
	if ( $this->has_deposit() ) {
		$hide_itemization = true;
	}
	return $hide_itemization;
}
```

timur27 (approved):

gpressutto5:
The build is failing here because it is a fork. I created #8110 only to run CI, and it is passing.

peterwilsoncc:
Thank you so much for your help wrapping this up and catching all the gotchas and items I was unaware of. 

Once this is released, we'll declare compatibility in the Deposits readme for the relevant versions of Payments.

